Issue #1648, opened 2021-12-15
Named parameters should be identified in the error message


seebees:
This yields the error: "incorrect type of datatype constructor argument (expected nat, found uint8)"
But this should include the name e.g.

"incorrect type of datatype constructor argument someValue(expected nat, found uint8)"

```dafny
  datatype Foo = Foo(
    nameonly someValue: nat
  )

  function Bar(baz: uint8): Foo {
    Foo(someValue := baz)
  }
```

keyboardDrummer:
Why should the name be included? There's only 1 parameter so it's unambiguous which one it is.

MikaelMayer:
It is better for names to be included if available because they are more expressive than argument indices, especially in the light of the possibility that it is possible to provide arguments out-of-order. For example:

```
function divide(x: nat, y: nat): nat
requires y != 0
 {
  x/y
}

function divideByTwo(x: string): nat {
  divide(y := 2, x := x) //incorrect type of function argument 0 (expected nat, found string)
}
```
In this case, 0 is the first argument of the function `divide`, but it's not the first argument at the call site. With this PR, the error message becomes

```
  divide(y := 2, x := x) //incorrect type of function argument 0 named 'x' (expected nat, found string)
```

which is arguably easier to debug.

seebees:
Yes, @MikaelMayer is right. The example I gave has only one argument because it was simple to toss in.

But in the case of a more complicated example with many argument,
figuring out which argument is 6 of 12 is work.
And the fact that we added `nameonly` to disambiguate 5 from 6
it seems logical that the error message should reflect that. :)

PR #1650, opened 2021-12-16
fix: Better error reporting for type mismatch

MikaelMayer:
Fixes issue #1648

Before:

```cs
datatype Foo = Foo(nat, x: nat)
function Bar(baz: string): Foo {
  Foo(baz, baz) // incorrect type of datatype constructor argument 0 (expected nat, found string)
                // incorrect type of datatype constructor argument 1 (expected nat, found string)
}
datatype Foo3 = Foo3(x: nat)
function Bar3(baz: string): Foo3 {
  Foo3(baz) // incorrect type of datatype constructor argument (expected nat, found string)
}
```

After:

```cs
datatype Foo = Foo(nat, x: nat)
function Bar(baz: string): Foo {
  Foo(baz, baz) // incorrect type of datatype constructor argument 0 (expected nat, found string)
                // incorrect type of datatype constructor argument 1 named 'x' (expected nat, found string)
}
datatype Foo3 = Foo3(x: nat)
function Bar3(baz: string): Foo3 {
  Foo3(baz) // incorrect type of datatype constructor argument named 'x' (expected nat, found string)
}
```

Unchanged:
```cs
datatype Foo2 = Foo2(nat)
function Bar2(baz: string): Foo2 {
  Foo2(baz) // incorrect type of datatype constructor argument (expected nat, found string)
}
```

I fixed various tests that already cover all of the examples above so there is no need for new tests.

### DIFF
--- a/Source/Dafny/Resolver.cs
+++ b/Source/Dafny/Resolver.cs
@@ -13546,10 +13546,16 @@ namespace Microsoft.Dafny {
           callee.IsStatic ? null : s.Receiver);
         // type check the out-parameter arguments (in-parameters were type checked as part of ResolveActualParameters)
         for (int i = 0; i < callee.Outs.Count && i < s.Lhs.Count; i++) {
-          var it = callee.Outs[i].Type;
+          var outFormal = callee.Outs[i];
+          var it = outFormal.Type;
           Type st = SubstType(it, typeMap);
           var lhs = s.Lhs[i];
-          AddAssignableConstraint(s.Tok, lhs.Type, st, "incorrect type of method out-parameter" + (callee.Outs.Count == 1 ? "" : " " + i) + " (expected {1}, got {0})");
+          AddAssignableConstraint(
+            s.Tok, lhs.Type, st, "incorrect type of method out-parameter" +
+                                 (callee.Outs.Count == 1 ? "" : " " + i) +
+                                 (outFormal is ImplicitFormal || !outFormal.HasName
+                                   ? "" : "named '" + outFormal.Name + "'") +
+                                 " (expected {1}, got {0})");
         }
         for (int i = 0; i < s.Lhs.Count; i++) {
           var lhs = s.Lhs[i];
@@ -16806,7 +16812,7 @@ namespace Microsoft.Dafny {
               formals = new List<Formal>();
               for (var i = 0; i < fnType.Args.Count; i++) {
                 var argType = fnType.Args[i];
-                var formal = new Formal(e.tok, "_#p" + i, argType, true, false, null);
+                var formal = new ImplicitFormal(e.tok, "_#p" + i, argType, true, false);
                 formals.Add(formal);
               }
             }
@@ -16942,7 +16948,7 @@ namespace Microsoft.Dafny {
           if (formals.Count != 1) {
             what += " " + j;
           }
-          if (formal.HasName) {
+          if (formal.HasName && !(formal is ImplicitFormal)) {
             what += " named '" + formal.Name + "'";
           }
 
--- a/Source/Dafny/Resolver.cs
+++ b/Source/Dafny/Resolver.cs
@@ -13554,7 +13554,7 @@ namespace Microsoft.Dafny {
             s.Tok, lhs.Type, st, "incorrect type of method out-parameter" +
                                  (callee.Outs.Count == 1 ? "" : " " + i) +
                                  (outFormal is ImplicitFormal || !outFormal.HasName
-                                   ? "" : "named '" + outFormal.Name + "'") +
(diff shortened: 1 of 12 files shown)


prvshah51 (dismissed):

prvshah51 (approved):

Source/Dafny/Resolver.cs
```diff
@@ -13546,10 +13546,16 @@ private class RBranchExpr : RBranch {
...
         for (int i = 0; i < callee.Outs.Count && i < s.Lhs.Count; i++) {
-          var it = callee.Outs[i].Type;
+          var outFormal = callee.Outs[i];
+          var it = outFormal.Type;
           Type st = SubstType(it, typeMap);
           var lhs = s.Lhs[i];
-          AddAssignableConstraint(s.Tok, lhs.Type, st, "incorrect type of method out-parameter" + (callee.Outs.Count == 1 ? "" : " " + i) + " (expected {1}, got {0})");
+          AddAssignableConstraint(
+            s.Tok, lhs.Type, st, "incorrect type of method out-parameter" +
+                                 (callee.Outs.Count == 1 ? "" : " " + i) +
```
keyboardDrummer, line 13555: It seems redundant to specify both the index and the name of the parameter.

Source/Dafny/Resolver.cs
```diff
@@ -13546,10 +13546,16 @@ private class RBranchExpr : RBranch {
...
         // type check the out-parameter arguments (in-parameters were type checked as part of ResolveActualParameters)
         for (int i = 0; i < callee.Outs.Count && i < s.Lhs.Count; i++) {
-          var it = callee.Outs[i].Type;
+          var outFormal = callee.Outs[i];
+          var it = outFormal.Type;
           Type st = SubstType(it, typeMap);
           var lhs = s.Lhs[i];
-          AddAssignableConstraint(s.Tok, lhs.Type, st, "incorrect type of method out-parameter" + (callee.Outs.Count == 1 ? "" : " " + i) + " (expected {1}, got {0})");
+          AddAssignableConstraint(
+            s.Tok, lhs.Type, st, "incorrect type of method out-parameter" +
```
keyboardDrummer, line 13554: I think this expression could use some variable extraction, and an interpolated string.

```
$"incorrect type of method out-parameter{parameterReference} (expected {{1}}, got {{0}})"
```

Source/Dafny/Resolver.cs
```diff
@@ -16942,6 +16948,10 @@ class MethodCallInformation {
           if (formals.Count != 1) {
             what += " " + j;
           }
+          if (formal.HasName && !(formal is ImplicitFormal)) {
```
keyboardDrummer, line 16951: Similarly, here it seems strange to include both the index and name.

Btw, what's a situation in which there's no name? And what's the implicitFormal situation?

Source/Dafny/Resolver.cs
MikaelMayer: It seems indeed, but in Dafny you can omit the name of datatype arguments, so the index is still needed, and the name helps a lot (especially when checking the meaning).
This PR takes an hybrid approach: As before, do not show the index if there is only one argument; do not show the name if it has been derived implicitly,
Looking at the error messages, I don't personally feel bothered by seeing the argument index (where to look at) and the name (the semantics of this argument) at the same time.

Source/Dafny/Resolver.cs
MikaelMayer: Nice use of double braces. Will do it.

Source/Dafny/Resolver.cs
MikaelMayer: Sure, here are the situations I encountered in the CI.

```
datatype List = Nil | Cons(nat, List)
```
Here the formal associated to "nat" does not have any name, it makes no sense to display the generated name "_p0".

A second example:
```
function f(callback: nat -> nat): nat {
  callback(-1)
}
```
Here a formal is specially created for the application of callback, because call sites need formals. This formal is implicit and has the name "_#p0".

Last example:
```
least lemma G[ORDINAL](x: int)
    ensures x == 8
  {
    H(x);  // error: cannot call from [ORDINAL] to [nat]
  }

  least lemma H[nat](x: int)
    ensures x == 8
  {
    G(x);  // error: cannot call from [nat] to [ORDINAL]
  }
```

Here least lemmas are implicitly provided another first formal, thus an implicit formal, which is named `_k` (see [here](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-friendliness) for an explanation of why these variables are useful) which makes little sense to report since the name is not part of the AST.

keyboardDrummer: Thanks

Issue #149, opened 2017-10-21
Internal error related to type variables


wilcoxjay:
Consider the following program:

```
function Foo<T>(m: map<int, T>): seq<T>
    ensures Foo(m) != []

lemma Bar()
    ensures forall m | 0 in m :: Foo(m)[0] == m[0]
{}
```
This program is malformed because the type of `m` is not inferable. The likely intent of the programmer was for `Bar` to take a type variable `T` which is the range type of `m`.

However, instead of reporting this problem gracefully, Dafny crashes with an internal error:
```
Assumption failed.  DEBUG: Unexpectedly confirmed XConstraint: Equatable: ? ? |||| 
```

alexlesaka:
Hello @wilcoxjay , 

two details. As the lemma calls to a polimolphic function, one of the error is solved if you write <T> after the name (Bar<T>). 

Then, dafny is not able to infer the type of the forall, but you can write it explicitly: `forall m: map<int,T> |
`
After that, the lemma have to be completed.

```
function Foo<T>(m: map<int, T>): seq<T>
    ensures Foo(m) != []

lemma Bar<T>()
    ensures forall m:map<int,T> | 0 in m :: Foo(m)[0] == m[0]
{
	assume forall m:map<int,T> | 0 in m :: Foo(m)[0] == m[0];
}
```

Another reason could be that dafny is not able to ensure that T-type defines equality (realding the error case could be that). Try tu `<T(==)>` insted of `<T>`.

 I'm using Dafny 2.0.0. Are you too?

wilcoxjay:
Hi @alexlesaka,

Yes, the program is definitely malformed. I also agree that there are ways of changing the program so that is not malformed.

The issue I am recording here (so that I or someone else can fix it later) is that Dafny itself crashes on this program. Dafny should never crash on any input program. Instead, it should gracefully report an error and exit if it discovers a problem with the given program.

I ran into this issue on the master branch (currently [749ec852a](https://github.com/Microsoft/dafny/tree/749ec852a6c5429796cc1aea0dfcb7de5ec53da4)). But I wouldn't be surprised if it's also present in the 2.0.0 release, since that is still very recent.

RustanLeino:
[...]

PR #1652, opened 2021-12-16
 fix(lsp): Don't get stuck when onsave verification is cancelled

atomb:
Previously, if an LSP client triggered the verification of a file because of a document save while a previous verification was still running, the process of cancelling the first verification was badly behaved. The thread handling the verification would throw an exception and the client would not get a response. This would lead to the VSCode client getting stuck displaying "Verifying..." forever.

This commit fixes the issue with two changes:

* It adds a catch block in the handler for a save event that behaves similarly to that for a change event in the case where verification is cancelled.
* It makes the check for in-order change messages more liberal, allowing non-consecutive version numbers (as allowed by the LSP standard, and as occur in practice after save events).

It also includes a test that fails on current `master` but succeeds on this branch.


MikaelMayer (dismissed):
I approve these changes because they are conservative and fix this issue, but would love to have the feedback on @camrein as well.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -142,7 +142,11 @@ ISymbolTableRelocator symbolTableRelocator
       if (!VerifyOnSave) {
         return databaseEntry.Document;
       }
-      return VerifyDocumentIfRequiredAsync(databaseEntry);
+      try {
+        return VerifyDocumentIfRequiredAsync(databaseEntry);
+      } catch (OperationCanceledException) {
```
camrein, line 147: This catch-clause won't have any effect since the `OperationCanceledException` is encapsulated in the `Task` returned by `VerifyDocumentIfRequiredAsync`.
You'll need to refactor the method to the async/await pattern of C# and `await` the invocation of `VerifyDocumentIfRequiredAsync`.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -90,8 +90,10 @@ ICompilationStatusNotificationPublisher notificationPublisher
             _ => throw new ArgumentException($"invalid request type ${request.GetType()}")
           };
           request.Document.SetResult(document);
-        } catch (OperationCanceledException e) {
-          request.Document.SetCanceled(e.CancellationToken);
+        } catch (OperationCanceledException) {
```
camrein, line 93: I believe this adaption was made because the catch-clause inside `DocumentDatabase` doesn't have any effect. After applying my proposed refactoring, this error handling should become obsolete.
For design purposes, I refactored the cancellation so that the handling is as close as possible to the issuer of the cancellation. Therefore, `DocumentDatabase` should remain as the ultimate class to decide how to treat a cancellation request and all other classes should bubble the `OperationCancelationException` to the invoker.

camrein (changes requested):
Thanks for nailing down the issue and proposing a fix. It makes sense to restore the original document `Task` when a verification is canceled.

PS: An integration test would be neat. However, since the cancellation of the verification appears to be unreliable (see #149), we have to leave that aside for now.

keyboardDrummer:
I'm a little but confused about the proposed changes.

I'm wondering if we should ever let the Task returned by `Task<DafnyDocument> VerifyAsync(DafnyDocument document, CancellationToken cancellationToken)` be cancelled.

Maybe we should change 

```

    private DafnyDocument VerifyInternal(VerifyRequest verifyRequest) {
      var (document, cancellationToken) = verifyRequest;
      notificationPublisher.SendStatusNotification(document.Text, CompilationStatus.VerificationStarted);
      var verificationResult = verifier.Verify(document.Program, cancellationToken);
      var compilationStatusAfterVerification = verificationResult.Verified
        ? CompilationStatus.VerificationSucceeded
        : CompilationStatus.VerificationFailed;
      notificationPublisher.SendStatusNotification(document.Text, compilationStatusAfterVerification);
      return document with {
        SerializedCounterExamples = verificationResult.SerializedCounterExamples
      };
    }
```

so that in the event `verifier.Verify` throws a cancellation exception, we publish a verification cancelled notification, and return the document.

camrein:
> so that in the event `verifier.Verify` throws a cancellation exception, we still return the document.

I don't like that the internal API no longer exposes the fact that the verification was canceled. That's why I'd prefer the change in the `DocumentDatabase` to handle the cancellation there, similar as it's already handled for document updates in `ApplyChangesAsync`.

> Maybe when a cancellation occurs, we should also publish a (new) verification cancelled notification. The client would handle this notification by removing the "Verifying.." message.

I agree that it makes sense to send a suitable status message here, or at least fall back to `VerificationFailed`.

MikaelMayer:
@keyboardDrummer  If I remember correctly, the problem was that an exception was not caught in the main thread when the cancellation happened during a save even, but it was for a verification during an onchange event.
But there might be more to that.

It is crucial that we first ensure that the fix works (Aaron did that), to optimize its handling based on your comments, and then we can figure out the details later. This problem has been around for three years and it very annoying, as users encounter it daily, multiple times a day.

As a side note it makes little sense currently to remove the "Verifying..." or even display a "Verification cancelled" message because when verification is canceled, it means that another verification has taken place.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
atomb: Good catch. I can confirm that, after removing this change, the the other change is sufficient to fix the issue in my testing.

However, I also agree that the refactoring you describe is cleaner, so I'm going to attempt to get that fully in place today, too. It'll make the `onchange` and `onsave` behavior more symmetrical.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
atomb: I think the root cause you're describing here matches my understanding. However, even after the refactoring you suggested above, this change still seems to be necessary. I'll continue investigating why.

atomb:
> I'm guessing an error is thrown in the call to `verifier.Verify` in `VerifyInternal`, and then no `notificationPublisher.SendStatusNotification(document.Text, compilationStatusAfterVerification);` is sent. What I don't understand is how the action that triggered the cancellation does not send another verification notification once it finishes.

I'm also pretty sure that that's at least half of what's going on here.

> I'm wondering if we should ever let the Task returned by `Task<DafnyDocument> VerifyAsync(DafnyDocument document, CancellationToken cancellationToken)` be cancelled.
> 
> Maybe we should change
> 
> ```cs
>     private DafnyDocument VerifyInternal(VerifyRequest verifyRequest) {
>       var (document, cancellationToken) = verifyRequest;
>       notificationPublisher.SendStatusNotification(document.Text, CompilationStatus.VerificationStarted);
>       var verificationResult = verifier.Verify(document.Program, cancellationToken);
>       var compilationStatusAfterVerification = verificationResult.Verified
>         ? CompilationStatus.VerificationSucceeded
>         : CompilationStatus.VerificationFailed;
>       notificationPublisher.SendStatusNotification(document.Text, compilationStatusAfterVerification);
>       return document with {
>         SerializedCounterExamples = verificationResult.SerializedCounterExamples
>       };
>     }
> ```
> 
> so that in the event `verifier.Verify` throws a cancellation exception, we still return the document.

That was what I tried first, and settled on the solution currently in this PR because it seemed simpler. I'm not entirely sure it is simpler (or at least better) in retrospect, though.

> Maybe when a cancellation occurs, we should also publish a (new) verification cancelled notification. The client would handle this notification by removing the "Verifying.." message.

I particularly like this idea. Having a distinct message for cancellation seems like the "right" approach. Clients can then decide exactly how to handle that message, but at least they'll get the information.

atomb:
I just pushed a change including the refactoring @camrein suggested. I think that further refactoring, including addition of a cancellation message, is probably warranted, but I'd be more inclined to do that as part of a separate PR later. That would allow this fix to be available to users more quickly.

I'd also love to have a unit test for it, but expect it would be somewhat tricky, as @camrein suggests. I hope we can have one eventually, though.

keyboardDrummer:
> I'd also love to have a unit test for it, but expect it would be somewhat tricky, as @camrein suggests. I hope we can have one eventually, though.

Why is the test tricky? We can send LSP client requests to the LSP server in the test suite, so why not have a test that sends the combination of save and change requests that trigger this bug?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -135,29 +135,33 @@ ISymbolTableRelocator symbolTableRelocator
       }
     }
 
-    public Task<DafnyDocument> SaveDocumentAsync(TextDocumentIdentifier documentId) {
+    public async Task<DafnyDocument> SaveDocumentAsync(TextDocumentIdentifier documentId) {
```
keyboardDrummer, line 138: Why change this method to async?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -135,29 +135,33 @@ ISymbolTableRelocator symbolTableRelocator
...
+          return document;
+        }
+        var cancellationSource = new CancellationTokenSource();
+        var updatedEntry = new DocumentEntry(
+          document.Version,
+          Task.Run(() => documentLoader.VerifyAsync(document, cancellationSource.Token)),
+          cancellationSource
+        );
+        documents[document.Uri] = updatedEntry;
+        return await updatedEntry.Document;
```
keyboardDrummer, line 161: Can you put the `try/catch` around  `documentLoader.VerifyAsync(document, cancellationSource.Token)` ?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
atomb: This change is to that it's possible to `await` the result of `VerifyDocumentIfRequiredAsync` as @camrein suggested. Given that I also moved the exception handler into `VerifyDocumentIfRequiredAsync`, it may not strictly be required, but it also makes it more closely match the structure of `UpdateDocumentAsync`, which seems good for making the code easier to understand.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
atomb: I'm not sure that the behavior in the `catch` block should be the same for all calls to `VerifyAsync`, the way things are currently structured. It's possible that we could refactor things so that it would be the same, though.

atomb:
> > I'd also love to have a unit test for it, but expect it would be somewhat tricky, as @camrein suggests. I hope we can have one eventually, though.
> 
> Why is the test tricky? We can send LSP client requests to the LSP server in the test suite, so why not have a test that sends the combination of save and change requests that trigger this bug?

You may be right. It'll be timing-sensitive, but I expect the test script will always be faster than manual typing, which is how I've tested things so far.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: It seems right now, if you cancel verification during `ApplyChangesAsync`, then it'll think the entire document load was cancelled and migrate the old document, even though only the verification was cancelled.

I think the behavior in the catch block should be the same for all calls to `VerifyAsync`, and not doing so may cause bugs.

atomb: That does make a lot of sense to me. However, if I understand correctly, this PR preserves the existing behavior other than keeping verification from blocking. I think I'd prefer to make those changes as part of a separate PR which may be a pure refactor or may fix existing bugs other than the one this PR aims to fix.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: I agree it might be more readable 👍

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: I think if you make the change I suggested, then you won't need the change in `TextDocumentLoader.Run`.

I think you could also only have the change in `TextDocumentLoader.Run`, without this change, but I believe @camrein didn't like that because it moves handling some cancellation from `DocumentDatabase` to `TextDocumentLoader`.

atomb: Ah, but `TextDocumentLoader.Run` calls `VerifyInternal`, not `VerifyAsync`. That is, it does the actual verification work, rather than queuing up a task to run. And, from what I can tell, both the queuing and the execution need to be able to handle cancellation.

I originally tried modifying only `TextDocumentLoader.Run`, but it only halfway solved the problem.

keyboardDrummer: I think both of your changes handle cancellation that occurs during actual verification work. When the call to `VerifyAsync ` returns a Task, that task has already been queued. When you `await` that `Task`, a cancelled exception is thrown only when a cancelled exception is thrown inside `VerifyInternal` and there is caught and used in `request.Document.SetException(e);`.

atomb: Ah, of course you're right. I've refactored with that in mind, according to your suggestions. I think this is the cleanest version of the fix so far, and it still seems to behave well.

atomb: After closer investigation, it actually sometimes still behaves badly, but in a different way. Under some circumstances, it can be possible to get the IDE+server into a state where it responds with "Verification Succeeded" to essentially any change, including obvious syntax errors. This seems to occur after one or more verification runs have been cancelled before one finally succeeds.

My understanding was that returning the original document in the case of an `OperationCanceledException` was the right thing to do, but perhaps it could lead to the behavior I'm now seeing?

One additional data point is the closing and re-opening the file in VSCode seems to reset the state and let things behave correctly for a while.

keyboardDrummer: > After closer investigation, it actually sometimes still behaves badly, but in a different way. Under some circumstances, it can be possible to get the IDE+server into a state where it responds with "Verification Succeeded" to essentially any change, including obvious syntax errors. This seems to occur after one or more verification runs have been cancelled before one finally succeeds.

Do you know how to let VSCode show the LSP logs? Might be useful to get them in this case. @MikaelMayer wrote a nice help here: https://github.com/dafny-lang/ide-vscode/pull/92#issuecomment-996825783

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -144,9 +144,13 @@ bool loadCanceled
...
 
     public async Task<DafnyDocument> VerifyAsync(DafnyDocument document, CancellationToken cancellationToken) {
-      var request = new VerifyRequest(document, cancellationToken);
-      requestQueue.Add(request, cancellationToken);
-      return await request.Document.Task;
+      try {
+        var request = new VerifyRequest(document, cancellationToken);
+        requestQueue.Add(request, cancellationToken);
+        return await request.Document.Task;
+      } catch (OperationCanceledException) {
```
keyboardDrummer, line 151: Personally I like having the `try` around as little code as possible, to make it clear where it's trying to catch. So if you have the try/catch in this file, I would put it around the line `var verificationResult = verifier.Verify(document.Program, cancellationToken);`

However, @camrein preferred not handling `OperationCanceledException` in this file at all, so you could move the try/catch to a `VerifyAsyncHandleCancellation` in `DocumentDatabase`

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
camrein: > But regardless, I do think your current change has improved things, so maybe let that second bug for another PR.

If you could share how I can reproduce this issue, I might better understand what's happening. More specifically, why my assumption about the root cause didn't resolve the issue. But I guess it's an issue that occasionally happens during development.

> I think one problem is that there is a state in the client, the contents of the status bar, that is not stored on the server.

This might be a possible reason. However, it's rather strange to me since the status notifications are sent with each change to the document. Have you guys checked the language server logs when this issue occurs? I'm just wondering if there's an exception thrown, such as this one:
https://github.com/dafny-lang/dafny/blob/189f4aef9ee49c80941d0cb9c3d20709a03a08da/Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs#L97-L99
I was thinking if we ran in a race condition somewhere. However, all the queued `Task` instances that represent a change wait for their predecessor to complete (or handle its cancellation). Moreover, outside the `DocumentDatabase` changes cannot run in parallel since `TextDocumentLoader` just has a single thread to handle them. On the client side, VSCode is mostly an asynchronous event queue receiving the status notifications.

> In `HandleUpdateAndPublishDiagnosticsAsync` we can then send a notification for the current status bar state, both before and after `await documentAction`, and we won't need to send any notification in inner code like `VerifyInternal`.

If I understand you correctly, this will no longer allow showing intermediate states and will affect #1655. Or do want to send this information redundantly?

> I also think we should store a separate `Task<Document> ResolvedDocument` and `Task<Document> VerifiedDocument` in `DocumentEntry`, so that handlers for things like `go to definition` can execute while a document is still resolving.

Great idea +1
We could also experiment with running/awaiting the verification on a different thread to not block the document change queue.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
camrein: Indeed, I'd prefer to leave the cancellation logic centralized within `DocumentDatabase`.
Therefore, I propose adding a `private` method inside `DocumentDatabase` that takes responsibility for the verification cancellation. This method can then be used by `VerifyDocumentIfRequiredAsync` and `LoadAndVerifyAsync` in place of `documentLoader.VerifyAsync`.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: > If I understand you correctly, this will no longer allow showing intermediate states and will affect #1655. Or do you want to send this information redundantly?

The idea would be something like this:

```csharp
// Handles an LSP request
async Task ProcessRequest(Request request) {
  // One update for parsing+resolution, one for verification
  Func<State, Task<State>> stateUpdates = SplitIntoUpdates(request); 
  foreach(var stateUpdate in stateUpdates) {
    UpdateState(await stateUpdate(state));
  }
}

// Can also be called from internal state updates such as in #1655
private void UpdateState(State newState)
{
  if (!newState.StatusBar.Equals(state.StatusBar)) {
    publisher.SendStatusNotification(document, newState.StatusBar.Status, newState.StatusBar.Message);
  }
  state = newState;
}
```

camrein: Ahh I see. You're suggesting that each step in the pipeline can be treated as a separate task within the request handlers. I like the approach of moving the LSP-related stuff more to the endpoints. But I think we should leave this architectural change to another PR (and maybe post Dafny 3.4.0).

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
atomb: That makes a lot of sense. I'll do that along with a deeper investigation of the false reports of verification success described above.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: Sounds good :)

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
atomb: That refactoring is now applied, along with a fix to avoid discarding non-consecutive document updates (which are allowed by the protocol, and can show up when saving documents, but were previously rejected by the server).

It's not a completely bug-free experience yet. The document can get into an inconsistent state where the parser or name resolver think there's an error and can't be convinced otherwise. But that's better than incorrectly reporting verification success. Reloading the document always seems to fix the issue.

keyboardDrummer (commented):
Changes look good. I will approve once we've tried adding a test.

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -73,6 +84,70 @@ public class ConcurrentInteractionsTest : DafnyLanguageServerTestBase {
       Assert.AreEqual(1, diagnostics.Length);
     }
 
+    [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+    public async Task VerificationErrorDetectedAfterCanceledSave() {
+      // Create a document that'll be slightly slow to verify
```
keyboardDrummer, line 89: Would it help if we add an option to Boogie that'll make the verifier take forever? Similar to this https://github.com/boogie-org/boogie/pull/472

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -73,6 +84,70 @@ public class ConcurrentInteractionsTest : DafnyLanguageServerTestBase {
...
+              Text = text
+            }
+          }
+        };
+      }
+
+      // Add a space in the document
+      var change1 = makeChange(documentItem.Version + 1, new Range((0, 6), (0, 7)), " ");
+      client.DidChangeTextDocument(change1);
+      // Save and don't wait, so the next save will interrupt and cancel it
```
keyboardDrummer, line 128: Could you change `it` into `verification` ? That's what's being interrupted right, not the saving? Same for line 135

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -73,6 +84,70 @@ public class ConcurrentInteractionsTest : DafnyLanguageServerTestBase {
...
+      client.SaveDocument(documentItem);
+
+      // Remove the space
+      // Use a non-consecutive version to test that the server doesn't drop it
+      var change2 = makeChange(documentItem.Version + 4, new Range((0, 6), (0, 7)), "");
+      client.DidChangeTextDocument(change2);
+      // Save and don't wait, so the next save will interrupt and cancel it
+      client.SaveDocument(documentItem);
+
+      // Make a verification-breaking change: replace == with !=
```
keyboardDrummer, line 138: It says `replace == with !=`, but that's not the change you're making right?

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -73,6 +84,70 @@ public class ConcurrentInteractionsTest : DafnyLanguageServerTestBase {
...
+      }
+
+      // Add a space in the document
+      var change1 = makeChange(documentItem.Version + 1, new Range((0, 6), (0, 7)), " ");
+      client.DidChangeTextDocument(change1);
+      // Save and don't wait, so the next save will interrupt and cancel it
+      client.SaveDocument(documentItem);
+
+      // Remove the space
+      // Use a non-consecutive version to test that the server doesn't drop it
```
keyboardDrummer, line 132: Could you replace `it` with `the change` ?

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -73,6 +84,70 @@ public class ConcurrentInteractionsTest : DafnyLanguageServerTestBase {
       Assert.AreEqual(1, diagnostics.Length);
     }
 
+    [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+    public async Task VerificationErrorDetectedAfterCanceledSave() {
```
keyboardDrummer, line 88: Test looks great 🥇

keyboardDrummer (dismissed):

camrein (dismissed):

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
atomb: The latest version of this PR now leaves `TextDocumentLoader` unchanged.

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
atomb: That could be useful, though I don't think it's urgent.

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
atomb: Good point. Updated.

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
atomb: Good catch! The original version of the code did that, but now it doesn't. Fixed.

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
atomb: Done.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -94,14 +102,20 @@ ISymbolTableRelocator symbolTableRelocator
       if (!documents.TryGetValue(documentUri, out var databaseEntry)) {
         throw new ArgumentException($"the document {documentUri} was not loaded before");
       }
-      if (documentChange.TextDocument.Version != databaseEntry.Version + 1) {
-        throw new InvalidOperationException($"the updates of document {documentUri} are out-of-order");
+
+      var oldVer = databaseEntry.Version;
+      var newVer = documentChange.TextDocument.Version;
+      if (oldVer > newVer) {
```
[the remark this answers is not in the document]
atomb: Good point. I've changed the conditional to catch that case.

atomb:
> I feel now a bit uncomfortable with this. I know that you found some cases in which the version bumping would would increment by one, but now what about "lost edits" in between?

My understanding is that a save operation updates the version but doesn't send the new version to the server. So a sequence of "change", "save", "change" will result in the two "change" operations having versions n and n+2, respectively. So I don't think there's anything we can do to detect "lost edits" when doing verification on save.

One option, perhaps, would be to retain the old behavior when doing verification on change, and only adopt the new behavior when doing verification on save.

atomb:
While I think this PR is close to finished, I'm still dissatisfied with the test I've come up with so far. Although it seems to encapsulate the situation that I've seen lead to problems, it turns out it doesn't actually fail on `master`, so it's not fully testing what it should be. I'm going to keep investigating.

atomb:
Alright, now I think this is really done. The current test fails with the code that existed prior to this PR and succeeds now.

I think there are still some potential bugs with synchronization of text changes between client and server, but I don't think there's anything that could be considered a soundness issue now (i.e, saying that something's verified when it isn't). Fixing any other synchronization issues can probably be the subject of a separate PR.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -94,9 +102,14 @@ ISymbolTableRelocator symbolTableRelocator
       if (!documents.TryGetValue(documentUri, out var databaseEntry)) {
         throw new ArgumentException($"the document {documentUri} was not loaded before");
       }
-      if (documentChange.TextDocument.Version != databaseEntry.Version + 1) {
-        throw new InvalidOperationException($"the updates of document {documentUri} are out-of-order");
+
+      var oldVer = databaseEntry.Version;
+      var newVer = documentChange.TextDocument.Version;
+      if (oldVer >= newVer) {
```
keyboardDrummer, line 108: Might be nice to add a link to LSP documentation about how the version should be handled. https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md?plain=1#L1195

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -73,6 +84,86 @@ public class ConcurrentInteractionsTest : DafnyLanguageServerTestBase {
...
+}".TrimStart();
+      var failSource = @"method Contradiction() { assert false; }";
+      await SetUp(new Dictionary<string, string>() {
+        { $"{DocumentOptions.Section}:{nameof(DocumentOptions.Verify)}", nameof(AutoVerification.OnSave) }
+      });
+      var documentItem = CreateTestDocument(source);
+      await client.OpenDocumentAndWaitAsync(documentItem, CancellationTokenWithHighTimeout);
+
+      DidChangeTextDocumentParams MakeChange(int? version, Range range, string text) {
+        return new DidChangeTextDocumentParams {
```
keyboardDrummer, line 111: Why not make the call `client.DidChangeTextDocument` here already?

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -73,6 +84,86 @@ public class ConcurrentInteractionsTest : DafnyLanguageServerTestBase {
...
+      client.DidChangeTextDocument(change1);
+      // Save and don't wait, so the next save will interrupt and cancel verification
+      Thread.Sleep(1);
+      client.SaveDocument(documentItem);
+
+      // Remove the space, and use a non-consecutive version to test that the server doesn't drop the change
+      var change2 = MakeChange(documentItem.Version + 4, new Range((0, 6), (0, 7)), "");
+      client.DidChangeTextDocument(change2);
+      // Save and don't wait, so the next save will interrupt and cancel verification
+      Thread.Sleep(1);
```
keyboardDrummer, line 136: Could you replace calls to `Thread.Sleep` with `await Task.Delay` ?

Also, can we avoid using sleep by waiting on something else? What do you want to happen during the sleep? Tests that depend on sleeps are often unstable.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
atomb: Good idea. Done.

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
atomb: It would slightly shorten the code, but I tend to like to keep pure computation or data construction separate from effect-producing operations. I don't have a super strong opinion about that, though.

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
atomb: The issue here is that cancelling the operation too quickly leads to an exception within the OmniSharp JSON parsing code. It isn't prepared for an operation to be cancelled before it has even finished parsing it. Fixing that would be ideal, but would probably require a change to the OmniSharp libraries. I'll make a note of it in a comment, though.

And, yes, I think `await Task.Delay` seems better. I'm still learning the C# concurrency API. :)

keyboardDrummer: > The issue here is that cancelling the operation too quickly leads to an exception within the OmniSharp JSON parsing code

Is that a bug in our code then, for cancelling code which shouldn't be cancelled? Maybe we can use different cancellationTokens to allow only cancelling the part that works well?

Also, do you happen to have more information on this OmniSharp JSON parsing exception? I might have been running into it without understanding it. I've seen this in some build runs:

```csharp
  Standard Output Messages:
 crit: OmniSharp.Extensions.JsonRpc.InputHandler[0]
       unhandled exception
       System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'start')
          at System.Buffers.BuffersExtensions.ToArray[T](ReadOnlySequence`1& sequence)
          at OmniSharp.Extensions.JsonRpc.InputHandler.ProcessInputStream(CancellationToken cancellationToken)
          at OmniSharp.Extensions.JsonRpc.InputHandler.ProcessInputStream(CancellationToken cancellationToken)
          at OmniSharp.Extensions.JsonRpc.InputHandler.<Start>b__26_0()
```

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -103,6 +103,8 @@ ISymbolTableRelocator symbolTableRelocator
         throw new ArgumentException($"the document {documentUri} was not loaded before");
       }
 
+      // According to the LSP specification, document versions should increase monotonically but may not be consecutive.
```
keyboardDrummer, line 106: `may be not consecutive` ?

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
atomb: The exception I had seen was in related code but was a different exception with a different message. In an attempt to get the exact message, I tried to remove the delays and reproduce the issue, and after 10 runs it didn't show up again. There have been a couple of other changes to the test, and changes in the rest of the code base merged in since then, including some in the test infrastructure.

Regardless, since I haven't been able to reproduce it, I'm going to remove the delays altogether. I'm much happier without them. If the issue shows up in the future, we can dig into why it's happening.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
atomb: Yes, much better.

keyboardDrummer (approved):

PR #1653, opened 2021-12-16
chore: Delete InterimKey.snk

cpitclaudel:
It looks like it hasn't been used since it was checked in back in
d3eacad3 (October 2012), and Microsoft doesn't recommend strong-naming any more
anyway (“Strong-names are left over from previous eras of .NET”,
https://github.com/dotnet/runtime/blob/main/docs/project/strong-name-signing.md)


MikaelMayer:
I don't see any reference in the link that you gave to `interimKey` or `snk`. What was exactly the purpose of this file? Couldn't it be that some users are still dependent on it?

cpitclaudel:
The explanation here is pretty good: https://stackoverflow.com/questions/131181/what-is-a-snk-for .  It's a binary signing key that you use to uniquely identify a build artifact.  The whole concept is obsolete AFAIU, and it shouldn't be in the repo anyway.

keyboardDrummer (approved):

PR #1654, opened 2021-12-16
Add an command line flag to DafnyLanguageServer.exe to control caching

cpitclaudel:
I'd like to collect data on how well the caching feature works, and part of that would be making it easy to turn it on and off in the UI.  But to do that I need it to be exposed on the CLI first, hence this PR.


Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -45,16 +45,21 @@ public class DafnyProgramVerifier : IProgramVerifier {
           // https://github.com/boogie-org/boogie/blob/b03dd2e4d5170757006eef94cbb07739ba50dddb/Source/VCGeneration/Couterexample.cs#L217
           DafnyOptions.O.ModelViewFile = "-";
           DafnyOptions.O.VcsCores = GetConfiguredCoreCount(options.Value);
+          DafnyOptions.O.VerifySnapshots = (int)options.Value.VerifySnapshots;
```
MikaelMayer, line 48: Is this PR still a draft? If so, mark is as such. It seems that you are defining an option but never using it anywhere except in loggin.

MikaelMayer (changes requested):

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
cpitclaudel: It's not a draft, it's ready to go.

MikaelMayer: What does it enable that was not possible before?

cpitclaudel: The option is set through the command line, and the command line flag is defined automatically by reflection.

cpitclaudel: It lets you turn on caching in Dafny's LSP server

keyboardDrummer (approved):

keyboardDrummer (approved):

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
MikaelMayer: Oh I see, so it's a Boogie option that needed to be declared by Dafny.

MikaelMayer (approved):

PR #1655, opened 2021-12-16
feat(lsp): Report tracing messages to language server clients

cpitclaudel:
This is intended to expose to LSP clients a feature that already exists in
Emacs' dafny-mode (coarse-grained progress reporting).

In the long term the proper way to do this is to have Dafny report fine-grained
progress as it runs verification tasks, but for now we can leverage the
existing API to display simple progress messages.

* Workspace/ICompilationStatusNotificationPublisher.cs,
Workspace/CompilationStatusNotificationPublisher.cs (SendStatusNotification):
Accept an additional parameter for progress messages.

* Language/IProgramVerifier.cs, Language/DafnyProgramVerifier.cs (Verify):
Accept a IVerificationProgressReporter instance.
* Language/DafnyProgramVerifier.cs (ModelCapturingOutputPrinter): Forward
Inform() messages to it.

* Language/IVerificationProgressReporter.cs: New interface to report
verification progress.
* Workspace/TextDocumentLoader.cs (VerificationProgressReporter): Implement it
by forwarding notifications to an ICompilationStatusNotificationPublisher.

(This still need updated tests)


Source/DafnyLanguageServer/Language/IVerificationProgressReporter.cs
```diff
@@ -0,0 +1,14 @@
+using System.Threading;
+
+namespace Microsoft.Dafny.LanguageServer.Language {
+  /// <summary>
+  /// A callback interface to report verification progress
+  /// </summary>
+  public interface IVerificationProgressReporter {
```
robin-aws, line 7: This smells (in a good way :) ) like something we could make more structured and hence more useful for other purposes. As in, we could have multiple event payloads like `VerificationStarted(name)` and `VerificationComplete(name, result)`. Converting those into string messages would be something one possible reporter would do, but we could also use this as a better communication channel than parsing the Boogle `/xml` output for the `/verificationLogger` feature.

cpitclaudel: Certainly.  That's what I was referring to in he commit message.  I was hoping this would be part of either your work on the verificationLogger feature or Remy's work on individual verification tasks.  If you want we can rescope this feature, but then I fear that the we'll step on each other's toes (?)

Source/DafnyLanguageServer/Workspace/ICompilationStatusNotificationPublisher.cs
```diff
@@ -12,6 +12,7 @@ public interface ICompilationStatusNotificationPublisher {
     /// </summary>
     /// <param name="textDocument">The document which had a compilation status change.</param>
     /// <param name="status"></param>
-    void SendStatusNotification(TextDocumentItem textDocument, CompilationStatus status);
+    /// <param name="message">Additional info about the current status</param>
+    void SendStatusNotification(TextDocumentItem textDocument, CompilationStatus status, string? message = null);
```
keyboardDrummer, line 16: Note that we'll run into some versioning issues when we change this API again.

The Dafny VSCode extension should be compatible with old and new Dafny versions, so it'll have to handle both the version of this API that this PR is adding, and the future API which will be different.

Since we don't have many customer yet, we might be able to make an exception though, and say that from a certain version of the extension it only supports Dafny 3.5 (or whatever version we add the final API in) or higher. Maybe that's the most practical approach here.

keyboardDrummer (dismissed):

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -170,4 +171,19 @@ private record LoadRequest(TextDocumentItem TextDocument, CancellationToken Canc
...
+    public ICompilationStatusNotificationPublisher publisher { get; init; }
+    public TextDocumentItem document { get; init; }
+
+    public VerificationProgressReporter(TextDocumentItem document,
+                                        ICompilationStatusNotificationPublisher publisher) {
+      this.document = document;
+      this.publisher = publisher;
+    }
+
+    public void reportProgress(string message) {
```
camrein, line 185: The usual coding convention is that methods start with an upper case letter in C#.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -152,7 +152,8 @@ bool loadCanceled
     private DafnyDocument VerifyInternal(VerifyRequest verifyRequest) {
       var (document, cancellationToken) = verifyRequest;
       notificationPublisher.SendStatusNotification(document.Text, CompilationStatus.VerificationStarted);
-      var verificationResult = verifier.Verify(document.Program, cancellationToken);
+      var progressReporter = new VerificationProgressReporter(document.Text, notificationPublisher);
```
camrein, line 155: I'd prefer that the instance of `VerificationProgressReporter` is provided through dependency injection. See [LanguageServerExtensions](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Workspace/LanguageServerExtensions.cs). Then, the actual instance can be injected through the constructor of `DafnyProgramVerifier`.
However, this will require that `VerificationProgressReporter` will be a singleton instance; thus, the `reportProgress` method has to receive the actual document.

camrein (changes requested):
There are some unit tests failing because of the newly introduced `VerificationInProgress` status.
Moreover, I think it makes sense to add at least one test that verifies that the `Message` part meets the expectations.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
cpitclaudel: Would that really be the right choice here? VerificationProgressReporter is a purely internal wapper around notificationPublisher to connect the Boogie API to the notifications API: what are cases at this point of the code where we would want to pass in something else than an instance of this specific VerificationErrorReporter? (Originally I thought I wouldn't even have that interface at all, but I introduced it to decouple the two APIs.

In any case I've changed `VerificationErrorReporter` to be a private class, which is certainly better.  Let me know if you think I should change it further.

Source/DafnyLanguageServer/Workspace/ICompilationStatusNotificationPublisher.cs
cpitclaudel: That's right, and that's also why I stuck with very very basic progress report here: since the cost on the IDE side to support this is literally below 10 lines, keeping support for it while supporting a more advanced API in the future will be a very small cost.

cpitclaudel:
OK, added tests, addressed feedback, will now wait for @camrein 's guidance on the tests and on the implementation of the reporter.  Thanks all for the feedback!

camrein (dismissed):

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -142,6 +148,10 @@ private class ModelCapturingOutputPrinter : OutputPrinter {
 
       public void Inform(string s, TextWriter tw) {
         logger.LogInformation(s);
+        var match = Regex.Match(s, "^Verifying .+[.](?<name>[^.]+) ...$");
```
MikaelMayer, line 151: The last three dots are any characters, can you escape them ?.
Also, this parsing of the verification message leaves aside the module and the class name when they are not default, which shouldn't be the case.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -169,5 +170,20 @@ private record Request(CancellationToken CancellationToken) {
     private record LoadRequest(TextDocumentItem TextDocument, CancellationToken CancellationToken) : Request(CancellationToken);
 
     private record VerifyRequest(DafnyDocument OriginalDocument, CancellationToken CancellationToken) : Request(CancellationToken);
+
+    private class VerificationProgressReporter : IVerificationProgressReporter {
+      public ICompilationStatusNotificationPublisher publisher { get; init; }
```
MikaelMayer, line 175: Either make these properties `readonly private`, or capitalize them `Publisher`, `Document`
This is the convention that is being used in similar files.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -142,6 +148,10 @@ private class ModelCapturingOutputPrinter : OutputPrinter {
...
+        if (match.Success) {
+          progressReporter.ReportProgress($"Verifying {match.Groups["name"].Value}...");
```
MikaelMayer, line 153: I don't like it that the message now contains the verification status and overrides it(`Verifying`). To me the compilation status should indicate what is happening, and the "message" should be additional information/arguments, not something that overrides the status itself, e.g. for maintainability or even down the road consistent translations.
After a second look, it looks like @camrein  already had the same remark here:
https://github.com/dafny-lang/ide-vscode/pull/104/files#r772466907
I second it.

MikaelMayer (changes requested):
Great job, I'm eager to see the verificationProgress expand to get other parameters.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
cpitclaudel: Good point re. the `.`! 

I left out the module name on purpose; it's true that it's ambiguous if you have multiple methods in one file with the same name, but it keeps things short in the common case where you don't; the alternative makes for longer progress messages.  WDYT?

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
cpitclaudel: I thought we'd do that with more structured messages: for now we were starting with the prover telling the IDE exactly what to display, and then we'd move to more structured messages.  (If we were to consider translations, I would expect the translation to happen once in Dafny, not separately in each IDE).  But I'm happy o change if you want.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
MikaelMayer: Ok, let's keep it as such, since verification happens only in one file and usually they don't define different modules.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
MikaelMayer: It's not just for translation (there is no need for that) It's more for separation of concerns. The message "Verifying" should be there only once for semantic reasons.
Please do this change so that the message is just an argument.

cpitclaudel: Done

MikaelMayer (approved):
Approved. Would be great down the road to have an option be able to deactivate this eye-catching quickly changing text. It could be beneficial for users with epilepsy, for when we'll have feedback directly in the document or progress bar.

PR #1656, opened 2021-12-16
feat: Log verification of splits

### LINKED ISSUE #1624: Report outcome/duration of splits when they occur in /verificationLogger:trx content
Depends on https://github.com/boogie-org/boogie/pull/461

The data recorded by the `/verificationLogger:trx` option can help to identify Boogie procedures that take much longer to verify, which we hypothesize in #1582 are therefore more likely to fail to verify in the future as code is changed. However, if a procedure is verified in multiple Boogie verification conditions because of directives like `assert {:split_here} true` or `assert {:focus} true`, those "splits" become the actual unit of verification. This means that Dafny codebases should pay attention to the verification time of individual splits whenever splitting occurs, instead of the total time to verify all splits for a particular procedure (especially since they may even be verified in parallel).

Unfortunately there isn't a good way to identify an individual Boogie split yet, even at the Boogie level, which makes it challenging to provide a good user experience (and I'm going to cut a separate issue for this as it is a deeper gap to address). As incremental progress, I propose just using the "split number" to disambiguate, using identifiers such as `CheckWellformed$$MyModule.__default.MyMethod$$7`. The exact format of the identifiers in the TRX output is already flagged as experimental in the help text for the option, and this feature wouldn't affect that status.

### LINKED ISSUE #1640: Lit test runner: support %OutputCheck
This isn't currently used by any integration tests, but would be useful to more thoroughly check the output of CLI behavior such as `/verificationLogger:trx`. It's also a required feature to use this runner for Boogie since it DOES already use `%OutputCheck`.

robin-aws:
Fixes #1624
Fixes #1640

Implementing %OutputCheck in the new test runner allows me to actually test this change properly. :)


atomb:
It appears that this PR includes the mechanism to run a new kind of test (`%OutputCheck`), but doesn't include any tests that use that mechanism. Am I understanding that correctly? And, if so, is that intentional?

robin-aws:
> It appears that this PR includes the mechanism to run a new kind of test (`%OutputCheck`), but doesn't include any tests that use that mechanism. Am I understanding that correctly? And, if so, is that intentional?

It totally does have a test that uses it! :) See https://github.com/dafny-lang/dafny/pull/1656/files?authenticity_token=ozLC4200hFQCF6xns1S2CGIL8KPFgVvEc3HM0Y7tqubhXjK5Lw%2BEF4w2%2F7X9FkWU%2Fnw77hNydRTONNnmcczxDg%3D%3D&file-filters%5B%5D=.dfy#diff-3138a1afb17294036ea0b0eb30c83b5ea4f9a915ce67367d669f727fd4afb98dR2

You can also have a look at some of the Boogie tests for larger examples, such as https://github.com/boogie-org/boogie/blob/master/Test/commandline/SplitOnEveryAssert.bpl

atomb:
> > It appears that this PR includes the mechanism to run a new kind of test (`%OutputCheck`), but doesn't include any tests that use that mechanism. Am I understanding that correctly? And, if so, is that intentional?
> 
> It totally does have a test that uses it! :) See https://github.com/dafny-lang/dafny/pull/1656/files?authenticity_token=ozLC4200hFQCF6xns1S2CGIL8KPFgVvEc3HM0Y7tqubhXjK5Lw%2BEF4w2%2F7X9FkWU%2Fnw77hNydRTONNnmcczxDg%3D%3D&file-filters%5B%5D=.dfy#diff-3138a1afb17294036ea0b0eb30c83b5ea4f9a915ce67367d669f727fd4afb98dR2
> 
> You can also have a look at some of the Boogie tests for larger examples, such as https://github.com/boogie-org/boogie/blob/master/Test/commandline/SplitOnEveryAssert.bpl

Ah, right! I missed that that comment was enough to trigger that test harness.

atomb (approved):
It looks like the style check is failing, but I'm happy with this once that passes.

keyboardDrummer:
Change looks good but there are whitespace (and other..?) errors :)

atomb (approved):

Source/XUnitExtensions/Lit/OutputCheckCommand.cs
```diff
@@ -0,0 +1,115 @@
+using System;
```
MikaelMayer, line 1: @robin-aws what are all these new check warnings I see on PRs now? Is this something related to this branch?

PR #1659, opened 2021-12-17
Git issue 1648 fix

MikaelMayer:
Code reformatting to address comments on this PR:
https://github.com/dafny-lang/dafny/pull/1650


Source/Dafny/Resolver.cs
```diff
@@ -13550,12 +13550,13 @@ private class RBranchExpr : RBranch {
           var it = outFormal.Type;
           Type st = SubstType(it, typeMap);
           var lhs = s.Lhs[i];
+          var outIndex = callee.Outs.Count == 1 ? "" : $" {i}";
```
keyboardDrummer, line 13553: Maybe change `{i]` into `at index {i}` ?

Source/Dafny/Resolver.cs
```diff
@@ -13550,12 +13550,13 @@ private class RBranchExpr : RBranch {
...
+          var outName = outFormal is ImplicitFormal || !outFormal.HasName
+            ? ""
```
keyboardDrummer, line 13555: How do you feel about putting `outIndex` where `""` is in `? ""`? Showing both an index and a name doesn't seem useful, don't you think?

MikaelMayer: The index describes the positional argument in the function, method or datatype definition, but this index might not match the positional argument at the call site, so the index is useful in any case to identify the argument in the definition.

Thus, the question is not whether or not to remove the index, but whether or not to add the name. And it can be argued that the name provide very useful information.

That is how I feel.

Source/Dafny/Resolver.cs
MikaelMayer: I was thinking about it too. It does not make sense to say "argument 0" because usually we don't start counting at zero indeed !
So precising that it's an index is nice so that it explains why the number is zero-based.

Instead of lengthening with "at index 0", why don't we just increase the positional argument by 1, so that it is easier to decipher? Are you ok with this alternative?

cpitclaudel: @RustanLeino will be very sad if you start counting at 1 ^^

robin-aws: And @RustanLeino and I have continued to have the friendliest and slowest argument about this in the entire time I have known him. :D

Source/Dafny/Resolver.cs
keyboardDrummer: > The index describes the positional argument in the function, method or datatype definition, but this index might not match the positional argument at the call site, so the index is useful in any case to identify the argument in the definition.

Why is the index still useful if you already have the name?

MikaelMayer: Because the list of parameters is not sorted alphabetically, so it's faster to lookup for some information using an index (logarithmic time) than a name (linear time).

keyboardDrummer: Don't you think that by the time you need to start counting, you would search by name?

MikaelMayer: Another Dafny user told me about this issue: "I don't have a strong opinion about this but the method names is definitely useful. I do prefer the names but some Dafny users might prefer the index, so having both might be better. I do agree with Remy that I don't think I will need to count the index anymore if the name is provided. Maybe in cases where the argument names are similar or just filled with p q r then looking at the index might be easier."

I'd add to that the providing the argument index is only 2 characters.

keyboardDrummer: I think it should say `at index {index}` if no name is specified, and let's leave it zero based since that's what it has been.

MikaelMayer: There was an ambiguity in the discussion, here it is and how I currently resolve it:
Ambiguity: "no name is specified" could either mean "no name is specified for the function definition (i.e. it's a closure)" or "no name is specified at the call site".
I believe that you point is the second variant, which makes sense. I solve it by now detecting if the name was explicitly provided, and, if it was not, it will include the index if there are more than 1 formal. Here is an example that illustrates all possible cases.

```cs
1: function recurseInto(message: string, nested: bool, acc: int, callback: int -> int, callback2: (int, int) -> int): int {
2:   var x := callback(nested);
3:   callback2(nested, x)
4: }
5: method callMethod() {
6:   var x := recurseInto(true, "abc", acc := "3", callback := x => x, callback2 := (x, y) => x);
7: }
// incorrect type of function application argument (expected int, found bool)  Resolver [2, 20]
// incorrect type of function application argument at index 0 (expected int, found bool)  Resolver [3, 12]
// incorrect type of function argument at index 0 named 'message' (expected string, found bool)  Resolver [6, 23]
// incorrect type of function argument at index 1 named 'nested' (expected bool, found string)  Resolver [6, 23]
// incorrect type of function argument named 'acc' (expected int, found string)  Resolver [6, 23]
```

Let me know if you request any changes.

keyboardDrummer: Ah, I think I didn't understand the importance of the index well. You're saying the index is important because at the call site you might have fewer names than at the definition site, and you would like to understand the exception based on looking just at the call site, right?

PR looks good. I'll approve it. For the future, it might be good to look at compilers that are known for having good error messages, say the Rust compiler, and see how they handle similar cases.

Source/Dafny/Resolver.cs
```diff
@@ -16942,11 +16945,12 @@ class MethodCallInformation {
           actuals.Add(b.Actual);
           substMap.Add(formal, b.Actual);
           var what = whatKind + (context is Method ? " in-parameter" : " argument");
-          if (formals.Count != 1) {
-            what += " " + j;
+          var displayName = formal.HasName && !(formal is ImplicitFormal);
+          if (!namesProvidedExplicitly.Contains(formal.Name) && formals.Count() > 1) {
```
keyboardDrummer, line 16949: I would also be okay with not having the `namesProvidedExplicitly` logic and always showing the index.

Source/Dafny/Resolver.cs
```diff
@@ -16942,11 +16945,12 @@ class MethodCallInformation {
...
+            what += $" at index {j}";
           }
-          if (formal.HasName && !(formal is ImplicitFormal)) {
-            what += " named '" + formal.Name + "'";
+          if (displayName) {
```
keyboardDrummer, line 16952: There's a little duplication of logic between here and `ResolveCallStmt`, it might be nice to put that in a method.

keyboardDrummer (approved):

Source/Dafny/Resolver.cs
MikaelMayer: Me too and that would be my preference, so I removed that logic.

Source/Dafny/Resolver.cs
MikaelMayer: Nice refactoring done.

Source/Dafny/Resolver.cs
amaurremi: Indeed, as @keyboardDrummer points out a nice solution could be to produce error messages like Rust, which are in this case brief in text (expected X, found Y), but conveniently highlight the pieces of code that had the corresponding types:
Screen Shot 2022-01-07 at 8 57 50 PM

MikaelMayer: Indeed, that's even better.

PR #1661, opened 2021-12-20
chore: Documentation about array initialization

### LINKED ISSUE #1660: [Docs] Show inline array value initialiser shortcut
## What
Add inline array intialiser syntax to documentation
```dfy
// This
var a := new int[] [1, 2, 3, 4];
// is better than
var a := new int[4];
a[0], a[1], a[2], a[3] := 1, 2, 3 , 4;
```

## Why
This is an expected feature from the use in other languages such as c#. It should feel more obvious.

## Background
I was trying to find a way to initialise an array to a list of values. I could not find this in the docs easily (cheatsheet, tutorial, quick reference). Except for doing it in spread assignment fashion.

I then found this PR which introduced the feature. However it seems it's undocumented:

https://github.com/dafny-lang/dafny/commit/72f60ac8238b5c10cb32469c1a674ae8f504ac9e

Edit: I can see the syntax listed here:
https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#151-one-dimensional-arrays

MikaelMayer:
This fixes #1660
Array initialization with values was not documented yet.


docs/DafnyRef/Types.md
```diff
@@ -2409,7 +2409,10 @@ A one-dimensional array of `n` `T` elements is created as follows:
 a := new T[n];
 ```
 The initial values of the array elements are arbitrary values of type
-`T`.
+`T`. They can also be initialized with the array display syntax:
```
keyboardDrummer, line 2412: I don't see "array display syntax" defined anywhere in the manual, so I wouldn't introduce this name without defining it. 

Consider changing this to `They can be initialized using an ordered list of expressions enclosed in square brackets, similar to a sequence display expression. Here is an example:`

docs/DafnyRef/Types.md
```diff
@@ -2409,7 +2409,7 @@ A one-dimensional array of `n` `T` elements is created as follows:
 a := new T[n];
 ```
 The initial values of the array elements are arbitrary values of type
-`T`. They can also be initialized with the array display syntax:
+`T`. They can also be initialized using a sequence display expression, as follows:
```
keyboardDrummer, line 2412: I think this is incorrect, since a sequence display expression also had other forms, and this only has the bracket form.

docs/DafnyRef/Types.md
```diff
@@ -2409,7 +2409,7 @@ A one-dimensional array of `n` `T` elements is created as follows:
 a := new T[n];
 ```
 The initial values of the array elements are arbitrary values of type
-`T`. They can also be initialized using a sequence display expression, as follows:
+`T`. They can also be initialized using a sequence display expression delimited by square brackets, as follows:
```
keyboardDrummer, line 2412: I think 'delimited by square brackets' implies that multiple elements are separated by square brackets, like `1 [] 2 [] 3`. I'd use the term 'enclosed' here.

The suggestion I made earlier was largely based on this existing piece from the manual, which reads "A sequence can be formed using a sequence display expression, which is a possibly empty, ordered list of expressions enclosed in square brackets."

What do you have against saying "They can be initialized using an ordered list of expressions enclosed in square brackets" ?

docs/DafnyRef/Types.md
```diff
@@ -2409,7 +2409,7 @@ A one-dimensional array of `n` `T` elements is created as follows:
 a := new T[n];
 ```
 The initial values of the array elements are arbitrary values of type
-`T`. They can also be initialized using a sequence display expression delimited by square brackets, as follows:
+`T`. They can also be initialized using a sequence display expression enclosed in square brackets, as follows:
```
keyboardDrummer, line 2412: I would leave out the 'also', otherwise it's implying the previous sentence already mentioned a way of initializing the values. Also, "a sequence display expression enclosed in square brackets" to me reads like there's an entire sequence display expression in square brackets, even though such as expression already has square brackets.

I think referring to a sequence display expression is a little confusing, since it's not exactly the same as this, so I'd prefer an explicit description like "an ordered list of expressions enclosed in square brackets". If we want to mention "sequence display expression" then I'd like to indicate it's only similar.

keyboardDrummer (approved):

PR #1662, opened 2021-12-20
fix(lsp): Fix a concurrency bug in `AddDiagnosticForFile` (Fixes #1651)

### LINKED ISSUE #1651: Language server seems to return incomplete diagnostics
I'm trying to pinpoint the cause of unexpected results that I observe while interacting with the LSP server.  Specifically:

1. Here's an input file, `false5.dfy`:

```dafny
method t0() { assert true; }
method t1() { assert true; }
method t2() { assert true; }
method t3() { assert true; }
method t4() { assert true; }
method t5() { assert true; }
method t6() { assert true; }
method t7() { assert true; }
method t8() { assert true; }
method t9() { assert true; }
method t10() { assert true; }
method t11() { assert true; }
method t12() { assert true; }
method t13() { assert true; }
method t14() { assert true; }
method t15() { assert true; }
method t16() { assert true; }
method t17() { assert true; }
method t18() { assert true; }
method t19() { assert true; }
method t20() { assert true; }
method t21() { assert true; }
method t22() { assert true; }
method t23() { assert true; }
method t24() { assert true; }
method t25() { assert false; }
method t26() { assert false; }
method t27() { assert false; }
method t28() { assert false; }
method t29() { assert false; }
```

2. Here's the way I feed that to the LSP server (remember that all newlines should be `\r\n`):

```
Content-Length: 272

{
 "jsonrpc": "2.0",
 "method": "initialize",
 "params": {
  "processId": 26828,
  "clientInfo": {
   "name": "Dafny diff tester"
  },
  "rootUri": "file:///c:/Users/cpitcla/wip/difftests",
  "capabilities": {
   "publishDiagnostics": {}
  }
 },
 "id": 0
}
Content-Length: 68

{
 "jsonrpc": "2.0",
 "method": "initialized",
 "params": {}
}
Content-Length: 1155

{
 "jsonrpc": "2.0",
 "method": "textDocument/didOpen",
 "params": {
  "textDocument": {
   "uri": "file:///c:/Users/cpitcla/wip/difftests/false5.dfy",
   "languageId": "dafny",
   "version": 0,
   "text": "method t0() { assert true; }\nmethod t1() { assert true; }\nmethod t2() { assert true; }\nmethod t3() { assert true; }\nmethod t4() { assert true; }\nmethod t5() { assert true; }\nmethod t6() { assert true; }\nmethod t7() { assert true; }\nmethod t8() { assert true; }\nmethod t9() { assert true; }\nmethod t10() { assert true; }\nmethod t11() { assert true; }\nmethod t12() { assert true; }\nmethod t13() { assert true; }\nmethod t14() { assert true; }\nmethod t15() { assert true; }\nmethod t16() { assert true; }\nmethod t17() { assert true; }\nmethod t18() { assert true; }\nmethod t19() { assert true; }\nmethod t20() { assert true; }\nmethod t21() { assert true; }\nmethod t22() { assert true; }\nmethod t23() { assert true; }\nmethod t24() { assert true; }\nmethod t25() { assert false; }\nmethod t26() { assert false; }\nmethod t27() { assert false; }\nmethod t28() { assert false; }\nmethod t29() { assert false; }\n"
  }
 }
}
Content-Length: 76

{
 "jsonrpc": "2.0",
 "method": "shutdown",
 "params": {},
 "id": 3
}
Content-Length: 61

{
 "jsonrpc": "2.0",
 "method": "exit",
 "params": {}
}
```

3. Here are two output traces.  First a good one:

```
Content-Length: 65

{"jsonrpc":"2.0","method":"serverStarted","params":"3.3.0.31104"}Content-Length: 86

{"jsonrpc":"2.0","method":"dafnyLanguageServerVersionReceived","params":"3.3.0.31104"}Content-Length: 350

{"jsonrpc":"2.0","id":0,"result":{"capabilities":{"textDocumentSync":2,"experimental":{},"workspace":{"workspaceFolders":{"changeNotifications":false},"fileOperations":{}},"completionProvider":null,"definitionProvider":null,"documentSymbolProvider":null,"hoverProvider":null,"signatureHelpProvider":null},"serverInfo":{"name":"DafnyLanguageServer"}}}Content-Length: 166

{"jsonrpc":"2.0","method":"dafny/compilation/status","params":{"uri":"file:///c:/Users/cpitcla/wip/difftests/false5.dfy","version":0,"status":"CompilationSucceeded"}}Content-Length: 165

{"jsonrpc":"2.0","method":"dafny/compilation/status","params":{"uri":"file:///c:/Users/cpitcla/wip/difftests/false5.dfy","version":0,"status":"VerificationStarted"}}Content-Length: 164

{"jsonrpc":"2.0","method":"dafny/compilation/status","params":{"uri":"file:///c:/Users/cpitcla/wip/difftests/false5.dfy","version":0,"status":"VerificationFailed"}}Content-Length: 987

{"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///c:/Users/cpitcla/wip/difftests/false5.dfy","version":0,"diagnostics":[{"range":{"start":{"line":26,"character":22},"end":{"line":26,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]},{"range":{"start":{"line":27,"character":22},"end":{"line":27,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]},{"range":{"start":{"line":28,"character":22},"end":{"line":28,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]},{"range":{"start":{"line":29,"character":22},"end":{"line":29,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]},{"range":{"start":{"line":25,"character":22},"end":{"line":25,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]}]}}Content-Length: 38

{"jsonrpc":"2.0","id":3,"result":null}
```

and second a bad one:

```
Content-Length: 65

{"jsonrpc":"2.0","method":"serverStarted","params":"3.3.0.31104"}Content-Length: 86

{"jsonrpc":"2.0","method":"dafnyLanguageServerVersionReceived","params":"3.3.0.31104"}Content-Length: 350

{"jsonrpc":"2.0","id":0,"result":{"capabilities":{"textDocumentSync":2,"experimental":{},"workspace":{"workspaceFolders":{"changeNotifications":false},"fileOperations":{}},"completionProvider":null,"definitionProvider":null,"documentSymbolProvider":null,"hoverProvider":null,"signatureHelpProvider":null},"serverInfo":{"name":"DafnyLanguageServer"}}}Content-Length: 166

{"jsonrpc":"2.0","method":"dafny/compilation/status","params":{"uri":"file:///c:/Users/cpitcla/wip/difftests/false5.dfy","version":0,"status":"CompilationSucceeded"}}Content-Length: 165

{"jsonrpc":"2.0","method":"dafny/compilation/status","params":{"uri":"file:///c:/Users/cpitcla/wip/difftests/false5.dfy","version":0,"status":"VerificationStarted"}}Content-Length: 164

{"jsonrpc":"2.0","method":"dafny/compilation/status","params":{"uri":"file:///c:/Users/cpitcla/wip/difftests/false5.dfy","version":0,"status":"VerificationFailed"}}Content-Length: 821

{"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///c:/Users/cpitcla/wip/difftests/false5.dfy","version":0,"diagnostics":[{"range":{"start":{"line":27,"character":22},"end":{"line":27,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]},{"range":{"start":{"line":28,"character":22},"end":{"line":28,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]},{"range":{"start":{"line":29,"character":22},"end":{"line":29,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]},{"range":{"start":{"line":25,"character":22},"end":{"line":25,"character":27}},"severity":1,"source":"Other","message":"assertion violation","relatedInformation":[]}]}}Content-Length: 38

{"jsonrpc":"2.0","id":3,"result":null}
```

The difference is the number of diagnostics: the first run reports only four (one is missing), while the last run returns 5.  Here is the missing issue:

```json
      {
        "range": {
          "start": {
            "line": 26,
            "character": 22
          },
          "end": {
            "line": 26,
            "character": 27
          }
        },
        "severity": 1,
        "source": "Other",
        "message": "assertion violation",
        "relatedInformation": []
      },
```

Both of these result from calling the verifier as `c:/Apps/dafny-3.3/DafnyLanguageServer.exe "--documents:verify=onchange"`.

I'm guessing I'm doing something wrong with the way I feed this document to language server.  But what?  Help would be much appreciated.

cpitclaudel:
Sending a didOpen message to the server produces roughly the following call tree:

    └┬─DocumentDatabase.OpenDocumentAsync
     └┬─DocumentDatabase.OpenAndVerifyAsync
      └┬─DocumentDatabase.LoadAndVerifyAsync
       └┬─TextDocumentLoader.VerifyAsync
        └┬─TextDocumentLoader.VerifyInternal
         └┬─DafnyProgramVerifier.Verify
          ├┬─DafnyProgramVerifier.VerifyWithBoogie
          │└┬─ExecutionEngine.InferAndVerify
          │ └┬─ExecutionEngine.VerifyImplementation
          │  └─ ExecutionEngine.ProcessErrors
          └┬─(Callback) ModelCapturingOutputPrinter.WriteErrorInformation
           └┬─DiagnosticErrorReporter.ReportBoogieError
            └─ DiagnosticErrorReporter.AddDiagnosticForFile

There can be multiple concurrent calls to `AddDiagnosticForFile`, but the code doesn't protect the datastructures that this function touches from concurrent modifications.  Hence we can observe traces like this:

    [lsp] Verifying t26: Enter WriteErrorInformation
    [lsp] Verifying t27: Enter WriteErrorInformation
    [lsp] Verifying t25: Enter WriteErrorInformation
    [lsp] Verifying t25: Exit WriteErrorInformation
    [lsp] Verifying t26: Exit WriteErrorInformation
    [lsp] Verifying t27: Exit WriteErrorInformation

Which in turn cause issues with the call to `diagnostics.GetOrCreate`, since we are in `didOpen`:

    diagnostics.GetOrCreate(documentUri, () => new List<Diagnostic>());

There's probably also an issue with the increment to count, since it's not atomic:

    counts.TryGetValue(severity, out var count);
    counts[severity] = count + 1;

In fact, even the call to `fileDiagnostics.Add(item)` is probably unsafe, since that's not a concurrent collection.

I'm confident with the diagnosis, but I'm not confident with the fix: @camrein, do you have an intuition for the right fix, and for which other methods of this class should be protected against concurrent access?

I'd also appreciate tips on how we might test this.  It was caught by the differential tester that I'm working on.

* DiagnosticErrorReporter.cs (AddDiagnosticForFile): Wrap function body in a lock to avoid concurrent updates.

### DIFF
--- a/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
+++ b/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
@@ -5,6 +5,7 @@ using OmniSharp.Extensions.LanguageServer.Protocol.Models;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 
 namespace Microsoft.Dafny.LanguageServer.Language {
   public class DiagnosticErrorReporter : ErrorReporter {
@@ -15,7 +16,7 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     private readonly DocumentUri entryDocumentUri;
     private readonly Dictionary<DocumentUri, List<Diagnostic>> diagnostics = new();
     private readonly Dictionary<DiagnosticSeverity, int> counts = new();
-    private System.Threading.ReaderWriterLockSlim rwLock = new();
+    private readonly ReaderWriterLockSlim rwLock = new();
 
     /// <summary>
     /// Creates a new instance with the given uri of the entry document.
@@ -31,7 +32,8 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     public IReadOnlyList<Diagnostic> GetDiagnostics(DocumentUri documentUri) {
       rwLock.EnterReadLock();
       try {
-        // Return a fresh list to avoid concurrency issues
+        // Concurrency: Return a copy of the list not to expose a reference to an object that requires synchronization.
+        // LATER: Make the Diagnostic type immutable, since we're not protecting it from concurrent accesses
         return new List<Diagnostic>(diagnostics.GetValueOrDefault(documentUri) ?? Enumerable.Empty<Diagnostic>());
       }
       finally {
--- a/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
+++ b/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
@@ -4,6 +4,7 @@ using OmniSharp.Extensions.LanguageServer.Protocol;
 using OmniSharp.Extensions.LanguageServer.Protocol.Models;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Dafny.LanguageServer.Language {
   public class DiagnosticErrorReporter : ErrorReporter {
@@ -14,9 +15,20 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     private readonly DocumentUri entryDocumentUri;
     private readonly Dictionary<DocumentUri, List<Diagnostic>> diagnostics = new();
     private readonly Dictionary<DiagnosticSeverity, int> counts = new();
-    private static readonly object syncObject = new();
+    private System.Threading.ReaderWriterLockSlim rwLock = new();
 
-    public IReadOnlyDictionary<DocumentUri, List<Diagnostic>> Diagnostics => diagnostics;
+    public IReadOnlyDictionary<DocumentUri, List<Diagnostic>> Diagnostics {
+      get {
+        rwLock.EnterReadLock();
+        try {
+          // Copy the dictionary to avoid concurrency issues
+          return diagnostics.ToDictionary(e => e.Key,
+                                          e => new List<Diagnostic>(e.Value));
+        } finally {
+          rwLock.ExitReadLock();
+        }
+      }
+    }
 
     /// <summary>
     /// Creates a new instance with the given uri of the entry document.
@@ -93,14 +105,22 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     }
 
     public override int Count(ErrorLevel level) {
-      return counts.GetValueOrDefault(ToSeverity(level), 0);
+      rwLock.EnterReadLock();
+      try {
+        return counts.GetValueOrDefault(ToSeverity(level), 0);
+      } finally {
+        rwLock.ExitReadLock();
+      }
     }
 
     private void AddDiagnosticForFile(Diagnostic item, DocumentUri documentUri) {
-      lock (syncObject) {
+      rwLock.EnterWriteLock();
+      try {
         var severity = item.Severity!.Value; // All our diagnostics have a severity.
         counts[severity] = counts.GetValueOrDefault(severity, 0) + 1;
         diagnostics.GetOrCreate(documentUri, () => new List<Diagnostic>()).Add(item);
+      } finally {
+        rwLock.ExitWriteLock();
       }
     }
 
--- a/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
+++ b/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
@@ -17,19 +17,6 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     private readonly Dictionary<DiagnosticSeverity, int> counts = new();
     private System.Threading.ReaderWriterLockSlim rwLock = new();
 
-    public IReadOnlyDictionary<DocumentUri, List<Diagnostic>> Diagnostics {
-      get {
-        rwLock.EnterReadLock();
-        try {
-          // Copy the dictionary to avoid concurrency issues
-          return diagnostics.ToDictionary(e => e.Key,
-                                          e => new List<Diagnostic>(e.Value));
-        } finally {
-          rwLock.ExitReadLock();
-        }
-      }
-    }
-
     /// <summary>
     /// Creates a new instance with the given uri of the entry document.
     /// </summary>
@@ -39,6 +26,16 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     /// </remarks>
     public DiagnosticErrorReporter(DocumentUri entryDocumentUri) {
       this.entryDocumentUri = entryDocumentUri;
+    }
+
+    public IReadOnlyList<Diagnostic> GetDiagnostics(DocumentUri documentUri) {
+      rwLock.EnterReadLock();
+      try {
+        // Return a fresh list to avoid concurrency issues
+        return new List<Diagnostic>(diagnostics.GetValueOrDefault(documentUri) ?? Enumerable.Empty<Diagnostic>());
+      } finally {
+        rwLock.ExitReadLock();
+      }
     }
 
     public void ReportBoogieError(ErrorInformation error) {
--- a/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
+++ b/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
@@ -33,7 +33,8 @@ namespace Microsoft.Dafny.LanguageServer.Language {
       try {
         // Return a fresh list to avoid concurrency issues
         return new List<Diagnostic>(diagnostics.GetValueOrDefault(documentUri) ?? Enumerable.Empty<Diagnostic>());
-      } finally {
+      }
+      finally {
         rwLock.ExitReadLock();
       }
     }
@@ -105,7 +106,8 @@ namespace Microsoft.Dafny.LanguageServer.Language {
       rwLock.EnterReadLock();
       try {
         return counts.GetValueOrDefault(ToSeverity(level), 0);
-      } finally {
+      }
+      finally {
         rwLock.ExitReadLock();
       }
     }
@@ -116,7 +118,8 @@ namespace Microsoft.Dafny.LanguageServer.Language {
         var severity = item.Severity!.Value; // All our diagnostics have a severity.
         counts[severity] = counts.GetValueOrDefault(severity, 0) + 1;
         diagnostics.GetOrCreate(documentUri, () => new List<Diagnostic>()).Add(item);
-      } finally {
+      }
+      finally {
         rwLock.ExitWriteLock();
       }
     }
--- a/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
+++ b/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
@@ -98,10 +98,9 @@ namespace Microsoft.Dafny.LanguageServer.Language {
 
     private void AddDiagnosticForFile(Diagnostic item, DocumentUri documentUri) {
       lock (syncObject) {
-        var fileDiagnostics = diagnostics.GetOrCreate(documentUri, () => new List<Diagnostic>());
         var severity = item.Severity!.Value; // All our diagnostics have a severity.
         counts[severity] = counts.GetValueOrDefault(severity, 0) + 1;
-        fileDiagnostics.Add(item);
+        diagnostics.GetOrCreate(documentUri, () => new List<Diagnostic>()).Add(item);
       }
     }
 
--- a/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
+++ b/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
@@ -92,10 +92,7 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     }
 
     public override int Count(ErrorLevel level) {
-      if (counts.TryGetValue(ToSeverity(level), out var count)) {
-        return count;
-      }
-      return 0;
+      return counts.GetValueOrDefault(ToSeverity(level), 0);
     }
 
     private void AddDiagnosticForFile(Diagnostic item, DocumentUri documentUri) {
@@ -103,8 +100,7 @@ namespace Microsoft.Dafny.LanguageServer.Language {
         var fileDiagnostics = diagnostics.GetOrCreate(documentUri, () => new List<Diagnostic>());
 
         var severity = item.Severity!.Value; // All our diagnostics have a severity.
-        counts.TryGetValue(severity, out var count);
-        counts[severity] = count + 1;
+        counts[severity] = counts.GetValueOrDefault(severity, 0) + 1;
         fileDiagnostics.Add(item);
       }
     }
--- a/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
+++ b/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
@@ -99,12 +99,14 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     }
 
     private void AddDiagnosticForFile(Diagnostic item, DocumentUri documentUri) {
-      var fileDiagnostics = diagnostics.GetOrCreate(documentUri, () => new List<Diagnostic>());
+      lock (this) {
+        var fileDiagnostics = diagnostics.GetOrCreate(documentUri, () => new List<Diagnostic>());
 
-      var severity = item.Severity!.Value; // All our diagnostics have a severity.
-      counts.TryGetValue(severity, out var count);
-      counts[severity] = count + 1;
-      fileDiagnostics.Add(item);
+        var severity = item.Severity!.Value; // All our diagnostics have a severity.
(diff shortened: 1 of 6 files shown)


MikaelMayer (dismissed):
I don't know if there is a better way, but this should fix the error you mentioned indeed.

keyboardDrummer:
I suggest you use a program like the one you found this bug with as a test-case. It won't be able to catch this bug with certainty but at least it gives us a chance to catch it.

It might help to increase the amount of false assert procedures to a very high number so there's more chance for two of the concurrent Boogie verifications to add an error at a similar time.

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -99,12 +99,14 @@ public class DiagnosticErrorReporter : ErrorReporter {
     }
 
     private void AddDiagnosticForFile(Diagnostic item, DocumentUri documentUri) {
```
keyboardDrummer, line 101: Would be nicer if we can record the errors for each Boogie procedure in a separate non-blocking collection, which is how it works when Dafny is called from the CLI (search for `OutputCollector`), but changing that seems like a lot of hassle, so let's do that as part of bigger Boogie changes :)

keyboardDrummer (changes requested):
Let's add a test-case that tries to find this issue

camrein:
Thanks for catching this issue @cpitclaudel.
I suppose it became apparent when we allowed using multiple cores at once for verification. I didn't think through that the callbacks might be called in parallel as well.

> In fact, even the call to `fileDiagnostics.Add(item)` is probably unsafe, since that's not a concurrent collection.

Honestly, the whole method is broken in terms of parallelism.

Your fix should be sufficient. However, I suggest moving the synchronization to `DafnyProgramVerifier` since it's causing this parallelism (although indirectly through boogie, but it applies the configuration):
https://github.com/dafny-lang/dafny/blob/c02d9efd565c4f2f3cea55924be9b20b9596716b/Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs#L151-L154
As we see, the use of `CaptureCounterExamples` isn't thread-safe as well, but it should be synchronized.

Moreover, in C# it's the usual convention to create a dedicated synchronization object for the monitor rather than using the instance of the enclosing type. The idea behind this pattern is to avoid accidental synchronization with types that may use said instance for synchronization.

For example:
```cs
private readonly object syncObject = new();

// ...
public void WriteErrorInformation(ErrorInformation errorInfo, TextWriter tw, bool skipExecutionTrace) {
  lock(syncObject) {
    CaptureCounterExamples(errorInfo);
    errorReporter.ReportBoogieError(errorInfo);
  }
}
```

Testing for the correct treatment of parallelism is (nearly) impossible. A unit test may show the presence of a race condition occasionally, but it can't prove their absence. The best we could do is run multiple iterations with multiple threads and see if we keep having the expected output. At this time, I'm not aware of a tool that can track low-level data races as a part of a unit test for C#.

For your solution, you could create a unit test similar to this one:
```cs
[TestMethod]
public void MessageIsThreadSafe() {
  for(int i = 0; i < 100; i++) {
    var reporter = new DiagnosticErrorReporter(DocumentUri.From("file://c/test.dfy"));
    Action logMessage = () => reporter.Message(MessageSource.Other, ErrorLevel.Error, Boogie.Token.NoToken, "test");
    Parallel.Invoke(
      Enumerable.Range(0, 100).Select(_ => logMessage).ToArray()
    );
    Assert.AreEqual(100, reporter.Count(ErrorLevel.Error));
    Assert.AreEqual(1, reporter.Diagnostics.Count);
  }
}
```

camrein:
I've just checked if it's possible to reproduce the error using an integration test. You can do the following in `DiagnosticsTest.cs`:

```cs
    [TestMethod]
    public async Task OpeningDocumentWithMultipleVerificationCoresReturnsStableDiagnostics() {
      var source = @"
method t0() { assert true; }
method t1() { assert true; }
method t2() { assert true; }
method t3() { assert true; }
method t4() { assert true; }
method t5() { assert true; }
method t6() { assert true; }
method t7() { assert true; }
method t8() { assert true; }
method t9() { assert true; }
method t10() { assert true; }
method t11() { assert true; }
method t12() { assert true; }
method t13() { assert true; }
method t14() { assert true; }
method t15() { assert true; }
method t16() { assert true; }
method t17() { assert true; }
method t18() { assert true; }
method t19() { assert true; }
method t20() { assert true; }
method t21() { assert true; }
method t22() { assert true; }
method t23() { assert true; }
method t24() { assert true; }
method t25() { assert false; }
method t26() { assert false; }
method t27() { assert false; }
method t28() { assert false; }
method t29() { assert false; }".TrimStart();
      for (int i = 0; i < 100; i++) {
        var documentItem = CreateTestDocument(source, $"test_{i}.dfy");
        client.OpenDocument(documentItem);
        var report = await diagnosticReceiver.AwaitNextNotificationAsync(CancellationToken);
        var diagnostics = report.Diagnostics.ToArray();
        Assert.AreEqual(5, diagnostics.Length);
        Assert.AreEqual("Other", diagnostics[0].Source);
        Assert.AreEqual(DiagnosticSeverity.Error, diagnostics[0].Severity);
      }
    }
```

Checking the counterexamples would make sense too.

camrein (changes requested):
Thanks for identifying this concurrency bug. See my comments on your questions.

cpitclaudel:
One thing that might we worth a double check is the way I wrapped the `Diagnostics` property (I made it copy all diagnostics, since it can be modified concurrently.  WDYT?

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -14,8 +15,20 @@ public class DiagnosticErrorReporter : ErrorReporter {
     private readonly DocumentUri entryDocumentUri;
     private readonly Dictionary<DocumentUri, List<Diagnostic>> diagnostics = new();
```
keyboardDrummer, line 16: Have you considered using a thread-safe dictionary? https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-6.0

cpitclaudel: I thought it wouldn't help, since there's concurrency on the access to the dictionary values as well (they are mutable lists), and since we have two separate dictionaries to update at the same time.

cpitclaudel:
I'm having the test fail with a cancellation error:

```
Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
  Skipped ChangeDocumentRightAfterOpeningCancelsLoad
  Failed OpeningDocumentWithMultipleVerificationCoresReturnsStableDiagnostics [50 s]
  Error Message:
   Test method Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization.DiagnosticsTest.OpeningDocumentWithMultipleVerificationCoresReturnsStableDiagnostics threw exception: 
System.OperationCanceledException: The operation was canceled.
  Stack Trace:
      at System.Threading.CancellationToken.ThrowOperationCanceledException()
   at System.Threading.SemaphoreSlim.WaitUntilCountOrTimeoutAsync(TaskNode asyncWaiter, Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at Microsoft.Dafny.LanguageServer.IntegrationTest.Util.TestNotificationReceiver`1.AwaitNextNotificationAsync(CancellationToken cancellationToken) in /home/runner/work/dafny/dafny/Source/DafnyLanguageServer.Test/Util/TestNotificationReceiver.cs:line 20
   at Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization.DiagnosticsTest.OpeningDocumentWithMultipleVerificationCoresReturnsStableDiagnostics() in /home/runner/work/dafny/dafny/Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs:line 642
```

I will investigate this.

keyboardDrummer:
> I will investigate this, but I'd love to hear if someone has experience with this.

I don't but seems like `JsonRpcIntegrationServerTestBase` cancels after 50 seconds.

camrein:
> One thing that might we worth a double check is the way I wrapped the `Diagnostics` property (I made it copy all diagnostics, since it can be modified concurrently. WDYT?

You're right about creating a copy. Otherwise, you're leaking a reference to the instance that can be used outside of the synchronization efforts, making debugging even harder.
I've quickly checked the usages of `Diagnostics`. Since it's only used to access the diagnostics of a specific file, I'd suggest we simplify the interface by providing a `GetDiagnostics(DocumentUri documentUri)` instead. That way, we only need to create a copy of the requested list.

> I don't but seems like `JsonRpcIntegrationServerTestBase` cancels after 50 seconds.

The test base class provides a `CancellationToken` that is canceled after a certain amount of time. By default, it is 50 seconds (see [JsonRpcTestOptions](https://github.com/OmniSharp/csharp-language-server-protocol/blob/112f24001e9d7c7929c60fd92dcaa3ca2025ba1e/src/JsonRpc.Testing/JsonRpcTestOptions.cs#L32)).
It helps to prevent situations where a test case runs infinitely in case of a bug. Depending on the number of iterations you're using for the concurrency test, the 50 seconds may be too little. Therefore, you could pass a placeholder cancellation token instead (e.g., `default`) and specify a higher threshold through the `Timeout` attribute (in milliseconds):
```cs
[TestMethod, Timeout(10_000)]
public async Task SomeTestMethod() {
  // ...
  var report = await diagnosticReceiver.AwaitNextNotificationAsync(default);
}
```
Of course, you could extend `DafnyLanguageServerTestBase` to provide a constructor that allows the specification of the timeout. However, this will affect all test cases of the test class.

cpitclaudel:
Thanks, the issue was indeed with the timeout.  50 seconds for a single test is too long, so I've made it smaller.  I performed the other API change.

keyboardDrummer (dismissed):

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -14,8 +15,7 @@ public class DiagnosticErrorReporter : ErrorReporter {
...
     private readonly Dictionary<DiagnosticSeverity, int> counts = new();
-
-    public IReadOnlyDictionary<DocumentUri, List<Diagnostic>> Diagnostics => diagnostics;
+    private System.Threading.ReaderWriterLockSlim rwLock = new();
```
camrein, line 18: Mark as `readonly`. And you may add `using System.Threading` in-place of the fully qualified name.

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -28,6 +28,17 @@ public class DiagnosticErrorReporter : ErrorReporter {
       this.entryDocumentUri = entryDocumentUri;
     }
 
+    public IReadOnlyList<Diagnostic> GetDiagnostics(DocumentUri documentUri) {
+      rwLock.EnterReadLock();
+      try {
+        // Return a fresh list to avoid concurrency issues
```
camrein, line 34: We could make this comment more explicit:

```cs
// Concurrency: Return a copy of the list not to expose a reference to an object that requires synchronization.
```

PS: For complete sanity, the `Diagnostic` instances would require a copy too. However, we never mutate any of these instances, so that should be fine for now (there are other locations with a  similar case). One thing I envision is to make `DafnyDocument` and its property types completely immutable.

camrein (changes requested):
Thanks for your effort, that looks fantastic so far.
I've just added a minor change for the sake of consistency.

keyboardDrummer (approved):

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
cpitclaudel: Done

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
cpitclaudel: Done, and I added a `LATER` note about this

camrein (approved):
Looks good, thanks!

PR #1663, opened 2021-12-22
Git issue 1648 fix

MikaelMayer:
Refactoring so that the code looks nicer.


Source/Dafny/Resolver.cs
```diff
@@ -16985,6 +16980,20 @@ class MethodCallInformation {
       bindings.AcceptArgumentExpressionsAsExactParameterList(actuals);
     }
 
+    private static string GetLocationInformation(Formal formal, int formalsCount, int index) {
```
keyboardDrummer, line 16983: Did you consider having the signature be `GetLocationInformation(IReadonlyList<Formal> formals, int index)` ?

MikaelMayer: Yes I did consider the two. However, since we already extracted the formal previously from the enumerable, I don't think we should extract it one more time. Some collections can be enumerated only once.

keyboardDrummer: If you make the argument `IReadonlyList<Formal> formals` then you know the collection supports constant time random access so there's no downside in accessing the element twice.

MikaelMayer: Indeed, that would work. However, with the `IReadonlyList<Formal>` we would further need to check that the index is within the bounds of the list, which we don't with the current method, so that's what I prefer. Despite that, would you still prefer the `IReadOnlyList<Formal>` version ?

keyboardDrummer: > we would further need to check that the index is within the bounds of the list

I don't think we should do that check. The caller is responsible for the index being in bounds, and in our two cases that would be guaranteed.

MikaelMayer: Done

Source/Dafny/Resolver.cs
```diff
@@ -16980,10 +16980,11 @@ class MethodCallInformation {
       bindings.AcceptArgumentExpressionsAsExactParameterList(actuals);
     }
 
-    private static string GetLocationInformation(Formal formal, int formalsCount, int index) {
+    private static string GetLocationInformation(List<Formal> formals, int index) {
+      var formal = formals[index];
       var displayName = formal.HasName && !(formal is ImplicitFormal);
       var description = "";
-      if (formalsCount > 1) {
+      if (formals.Count() > 1) {
```
keyboardDrummer, line 16987: No need to change anything, but I prefer `.Length` over `.Count()` when it is available, to indicate that it's constant time.

keyboardDrummer (approved):

Source/Dafny/Resolver.cs
```diff
@@ -16985,6 +16980,21 @@ class MethodCallInformation {
       bindings.AcceptArgumentExpressionsAsExactParameterList(actuals);
     }
 
+    private static string GetLocationInformation(List<Formal> formals, int index) {
```
keyboardDrummer, line 16983: No need to change anything, but here I would prefer `IReadOnlyList` over `List`, since it's a less specific interface and lets us know the method is pure.

keyboardDrummer (approved):

Source/Dafny/Resolver.cs
MikaelMayer: When I initially wrote `IReadOnlyList`, it complained because it was missing an import.
So I thought about it a second time. Since we know that at both call sites the first arguments are `List<Formal>`(likewise, we know that the index is within bounds), and this method is private, it makes sense to reuse the same time. So I chose it this way.

PR #1664, opened 2021-12-23
fix(lsp): kill language server if parent process dies

keyboardDrummer:
Fixes https://github.com/dafny-lang/ide-vscode/issues/95


Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -19,21 +20,46 @@ public static class DafnyLanguageServer {
...
+      }
+
+      try {
+        var hostProcess = Process.GetProcessById((int)request.ProcessId)!;
+        hostProcess.EnableRaisingEvents = true;
+        hostProcess.Exited += (_, _) => cancelLanguageServer.Cancel();
+      }
+      catch {
+        // If the process dies before we get here then request shutdown immediately
+        cancelLanguageServer.Cancel();
```
cpitclaudel, line 59: This will make it harder to test the server independently, because it will prevent us from replaying previously-recorded LSP traces.  Can we find a trick to have the server ignore this in some cases?  Maybe it's enough to do nothing in this catch case?

Source/DafnyLanguageServer/Program.cs
```diff
@@ -3,26 +3,31 @@
...
           options => options
             .WithInput(Console.OpenStandardInput())
             .WithOutput(Console.OpenStandardOutput())
             .ConfigureConfiguration(builder => builder.AddConfiguration(configuration))
             .ConfigureLogging(SetupLogging)
             .WithUnhandledExceptionHandler(LogException)
-            .WithDafnyLanguageServer(configuration)
+            .WithDafnyLanguageServer(configuration, cancelLanguageServer)
         );
+        cancelLanguageServer.Token.Register(() => server.ForcefulShutdown());
```
cpitclaudel, line 30: I think it would be good to log something here.

Source/DafnyLanguageServer.Test/Various/StartLanguageServer.cs
```diff
@@ -0,0 +1,11 @@
+// using System;
```
MikaelMayer, line 1: Why is this file commented out?

Source/DafnyLanguageServer/DafnyLanguageServer.cs
keyboardDrummer: Would it be possible to change the ParentId in the Initialize request when replaying the messages?

Source/DafnyLanguageServer.Test/DafnyLanguageServer.Test.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.7.0" />
```
MikaelMayer, line 10: Where is this dependency used?

Source/DafnyLanguageServer/Program.cs
keyboardDrummer: Added a log message, although in another location

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -19,21 +20,45 @@ public static class DafnyLanguageServer {
...
         .OnStarted(StartedAsync);
     }
 
-    private static Task InitializeAsync(ILanguageServer server, InitializeParams request, CancellationToken cancellationToken) {
+    private static Task InitializeAsync(ILanguageServer server, InitializeParams request, CancellationToken cancelRequestToken,
+        CancellationTokenSource cancelLanguageServer) {
       var logger = server.GetRequiredService<ILogger<Program>>();
       logger.LogTrace("initializing service");
+
+      KillLanguageServerIfParentDies(request, cancelLanguageServer);
```
MikaelMayer: I'm very new to this code, so I need to better understand it.
Why was the request not used before? Is this code run for every request?

Source/DafnyLanguageServer/DafnyLanguageServer.cs
cpitclaudel: Yes, certainly.

cpitclaudel:
This implementation strategy looks good to me.  I'm surprised that the VSCode folks don't clean up language server processes when the IDE is killed.

keyboardDrummer:
> This implementation strategy looks good to me. I'm surprised that the VSCode folks don't clean up language server processes when the IDE is killed.

The LSP process is not started by VSCode but by the LSP client NPM package that the Dafny extension uses, but that's a Microsoft package so you might be just as surprised :D

There's different ways of connecting to a language server with that client though, not all of them start a fresh language server.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -19,21 +20,51 @@ public static class DafnyLanguageServer {
...
       return Task.CompletedTask;
     }
 
+    /// <summary>
+    /// As part of the LSP spec, a language server must kill itself if its parent process dies
+    /// https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md?plain=1#L1713
+    /// </summary>
+    private static void KillLanguageServerIfParentDies(ILogger<Program> logger, InitializeParams request,
+      CancellationTokenSource cancelLanguageServer) {
+      if (!(request.ProcessId >= 0)) {
```
cpitclaudel, line 49: What is ProcessId set to when the client passes `null`?

cpitclaudel: Also, any reason to write this like this instead of `pid < 0`>

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -19,21 +20,51 @@ public static class DafnyLanguageServer {
...
+      void Kill() {
+        logger.LogWarning("Shutting down language server because parent process died.");
+        cancelLanguageServer.Cancel();
+      }
+
+      try {
+        var hostProcess = Process.GetProcessById((int)request.ProcessId)!;
+        hostProcess.EnableRaisingEvents = true;
+        hostProcess.Exited += (_, _) => Kill();
+      } catch {
```
cpitclaudel, line 62: Can we catch a more specific exception?  And maybe shrink the `try` block to just the GetProcessById?

Source/DafnyLanguageServer/DafnyLanguageServer.cs
keyboardDrummer: > What is ProcessId set to when the client passes null?

null

> Also, any reason to write this like this instead of pid < 0?

No. I copied this code from the C# LSP server

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,182 @@
...
+      } catch (ArgumentException) {
+        // Language server process is not running, pass the test.
+      }
+    }
+
+    private static async Task<Process> StartLanguageServerRunnerProcess() {
+      var languageServerBinary = Path.Join(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "DafnyLanguageServer");
+      var languageServerRunnerPath = await CreateDotNetDllThatStartsGivenFilepath(languageServerBinary);
+
+      var processInfo = new ProcessStartInfo("dotnet", languageServerRunnerPath) {
```
cpitclaudel, line 91: Why do we run `dotnet DafnyLanguageServer.dll` instead of `DafnyLanguageServer.exe` directly?

cpitclaudel: Oh, I see what's happening, sorry.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,181 @@
...
+  [TestClass]
+  public class ShutdownTest {
+
+    [TestMethod]
+    public async Task LanguageServerStaysAliveIfParentDiesButNoParentIdWasProvided() {
+      var process = await StartLanguageServerRunnerProcess();
+      var error = "";
+      process.ErrorDataReceived += delegate (object sender, DataReceivedEventArgs args) { error += args.Data; };
+
+      var languageServerProcessId = await process.StandardOutput.ReadLineAsync();
```
cpitclaudel, line 33: Why do you wrap a custom binary that prints its PID around the server instead of just reading the PID of the regular server?

cpitclaudel (commented):
LGTM, but I'm a bit confused by the testing strategy (I don't see yet why we need a custom binary to wrap the server).

Source/DafnyLanguageServer/DafnyLanguageServer.cs
cpitclaudel: > The value is then null, the >= operator returns false if any of the operands are null.

😱

Source/DafnyLanguageServer.Test/Various/StartLanguageServer.cs
keyboardDrummer: My bad. It's a development relic. Deleting..

Source/DafnyLanguageServer/DafnyLanguageServer.cs
keyboardDrummer: The request is also used in another location, but this method allows you do additional work. An `InitializeParams` request is sent at most once, so this code is run at most once.

Source/DafnyLanguageServer.Test/DafnyLanguageServer.Test.csproj
keyboardDrummer: It's used in the method `CreateDotNetDllThatStartsGivenFilepath` where it compiles the inline C# code.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
keyboardDrummer: Catching `ArgumentException` now. Do you feel the shrinking is worth the extra verbosity?

keyboardDrummer:
> LGTM, but I'm a bit confused by the testing strategy (I don't see yet why we need a custom binary to wrap the server).

It mimics how the LSP server would normally be started, although the only thing we _need_ to do is start a 'parent' process, which doesn't need to actually be the parent of anything, and kill it after the LSP server has received the initialize request passing in the processId of the parent process.

Still though, keeping this test platform independent, what would you start as a parent process? Creating a custom binary for this purpose seems practical to me, and it allows setting up a more realistic scenario.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
keyboardDrummer: We need to pass the PID of a running process to the language server in the initialize request, and later kill that process. This is what the custom binary is for. It also starts the language server to make the test more realistic, since the PID is supposed to be the parent process of the language server.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,181 @@
...
+        languageServer.Kill();
+        Assert.Fail("Language server should have killed itself if the parent is gone.");
+      } catch (ArgumentException) {
+        // Language server process is not running, pass the test.
+      }
+    }
+
+    private static async Task<Process> StartLanguageServerRunnerProcess() {
+      var languageServerBinary = Path.Join(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "DafnyLanguageServer");
+      var languageServerRunnerPath = await CreateDotNetDllThatStartsGivenFilepath(languageServerBinary.Replace("\\", "\\\\"));
```
MikaelMayer, line 89: This is a good place to use uninterpreted strings `"\\"` ==> `@"\"` and  `"\\\\"` ==> `@"\\". Not just to save one char for the second string, but to improve readability .`

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,181 @@
...
+          Capabilities = new ClientCapabilities(),
+        }
+      });
+#pragma warning disable VSTHRD002
+      outputHandler.StopAsync().Wait();
+#pragma warning restore VSTHRD002
+      return Encoding.ASCII.GetString(buffer.ToArray());
+    }
+
+    private static async Task<string> CreateDotNetDllThatStartsGivenFilepath(string filePathToStart) {
```
MikaelMayer, line 124: Since it is not explicit in the method name - and we cannot enlarge the method name - please add a comment to this method such as the following:

    // This method starts a process (like an IDE) that starts the Dafny Language Server but just crashs after reading one line.
    // It enables to test if the DafnyLanguageServer detects that its parent process died and also terminates

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,181 @@
...
+      var code = @$"using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+public class Foo {{
```
MikaelMayer, line 129: I have no strong opinion for this class name, but I would love to see an explicit class name here. What do you think of something such as "TerminatingVisualStudioCode" or "CrashingIDE"?

MikaelMayer (changes requested):
Minor comments, otherwise I now understand your solution and your test, it's fantastic !

fabiomadge (changes requested):

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,181 @@
...
+      await process.StandardInput.WriteAsync(initializeMessage);
+
+      var initializedResponseFirstLine = await process.StandardOutput.ReadLineAsync();
+      Assert.IsFalse(string.IsNullOrEmpty(initializedResponseFirstLine));
+
+      var error = await process.StandardError.ReadToEndAsync();
+      var didExit = process.WaitForExit(-1);
+      Assert.IsTrue(didExit);
+      Assert.IsNotNull(languageServerProcessId, error);
+      Assert.IsTrue(string.IsNullOrEmpty(error), error);
```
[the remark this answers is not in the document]
keyboardDrummer: Why not?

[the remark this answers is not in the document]
keyboardDrummer: Why not? It fails and shows `error` if `error` is not null or empty.

Source/DafnyRuntime/DafnyRuntime.csproj
```diff
@@ -40,8 +40,8 @@
   <Target Name="BuildDafnyRuntimeJar" AfterTargets="ResolveReferences" BeforeTargets="CoreCompile" Inputs="$(MSBuildProjectFile);@(DafnyRuntimeJavaInputFile)" Outputs="$(DafnyRuntimeJar)">
     
     <Message Text="Compiling DafnyRuntimeJava to $(DafnyRuntimeJar)..." Importance="high" />
-    <Exec WorkingDirectory="DafnyRuntimeJava" Condition="!$([MSBuild]::IsOSPlatform('Windows'))" Command="./gradlew -q clean build" />
-    <Exec WorkingDirectory="DafnyRuntimeJava" Condition="$([MSBuild]::IsOSPlatform('Windows'))" Command="./gradlew.bat -q clean build" />
+    <Exec WorkingDirectory="DafnyRuntimeJava" Condition="!$([MSBuild]::IsOSPlatform('Windows'))" Command="./gradlew -q build" />
```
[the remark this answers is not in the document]
keyboardDrummer: Added a comment.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -19,21 +21,51 @@ public static class DafnyLanguageServer {
...
+        return;
+      }
+
+      void Kill() {
+        logger.LogWarning("Shutting down language server because parent process died.");
+        cancelLanguageServer.Cancel();
+      }
+
+      try {
+        var hostProcess = Process.GetProcessById((int)request.ProcessId)!;
```
camrein, line 60: `GetProcessById` doesn't return a nullable reference hence why the null-forgiving operator `!` is unnecessary.

camrein (changes requested):
This looks great. Since it's part of the spec that language server should handle the termination of the parent process, I'm wondering why it's not part of omnisharp.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,181 @@
...
+using OmniSharp.Extensions.JsonRpc.Serialization;
+using OmniSharp.Extensions.LanguageServer.Protocol.Client.Capabilities;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
+
+  [TestClass]
+  public class ShutdownTest {
+
+    [TestMethod]
```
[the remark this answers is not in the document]
keyboardDrummer: > This looks great. Since it's part of the spec that language server should handle the termination of the parent process, I'm wondering why it's not part of omnisharp.

Yes, I don't understand that either.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -19,21 +21,51 @@ public static class DafnyLanguageServer {
       }
     }
 
-    public static LanguageServerOptions WithDafnyLanguageServer(this LanguageServerOptions options, IConfiguration configuration) {
+    public static LanguageServerOptions WithDafnyLanguageServer(this LanguageServerOptions options,
+      IConfiguration configuration, CancellationTokenSource cancelLanguageServer) {
```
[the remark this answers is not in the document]
keyboardDrummer: I wasn't happy with this design but didn't know how to do it better, and was hoping you'd have an idea :D

Could you provide some code for the event alternative you mention? I'm not sure what you're suggesting.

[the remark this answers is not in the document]
camrein: I've been checking omnisharp. Since it isn't using C#'s `event` feature I think we should leave that one aside since we don't need multiple registrations for now and follow omnisharp's `On*` convention with delegates. Therefore, I propose that we simply provide a callback for the `Kill` event. 

For example:
```cs    public delegate void OnDafnyLanguageServerKill();

    public static LanguageServerOptions WithDafnyLanguageServer(
      this LanguageServerOptions options,
      IConfiguration configuration,
      OnDafnyLanguageServerKill? onKill = null
    ) {
      return options
        .WithDafnyLanguage(configuration)
        .WithDafnyWorkspace(configuration)
        .WithDafnyHandlers()
        .OnInitialize((server, request, cancellationToken) => InitializeAsync(server, request, cancellationToken, onKill))
        .OnStarted(StartedAsync);
    }

    private static Task InitializeAsync(
      ILanguageServer server,
      InitializeParams request,
      CancellationToken cancellationToken,
      OnDafnyLanguageServerKill? onKill
    ) {
      void Kill() {
        logger.LogWarning("Shutting down language server because parent process died.");
        onKill?.Invoke();
      }
    }
```

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -19,21 +21,51 @@ public static class DafnyLanguageServer {
...
         .WithDafnyHandlers()
-        .OnInitialize(InitializeAsync)
+        .OnInitialize((server, @params, token) => InitializeAsync(server, @params, token, cancelLanguageServer))
         .OnStarted(StartedAsync);
     }
 
-    private static Task InitializeAsync(ILanguageServer server, InitializeParams request, CancellationToken cancellationToken) {
+    private static Task InitializeAsync(ILanguageServer server, InitializeParams request, CancellationToken cancelRequestToken,
+        CancellationTokenSource cancelLanguageServer) {
       var logger = server.GetRequiredService<ILogger<Program>>();
```
camrein, line 36: That's something I must have missed when I extracted the `WithDafnyLanguageServer` method into a separate class. Could you change the injected logger instance to `ILogger<DafnyLanguageServer>`?

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -24,24 +24,23 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
   [TestClass]
   public class ShutdownTest {
 
-    [TestMethod]
+    [TestMethod, Timeout(10000)]
```
camrein, line 27: For readability, you may want to write the literal as `10_000`.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -158,14 +157,14 @@ public class ShortLivedProcessStarter {{
         "System.ComponentModel.Primitives"
       };
       compilation = compilation.AddReferences(standardLibraries.Select(fileName =>
-        MetadataReference.CreateFromFile(Assembly.Load(fileName).Location)));
-      compilation = compilation.WithOptions(new CSharpCompilationOptions(OutputKind.ConsoleApplication))
-        .AddReferences(
-          MetadataReference.CreateFromFile(typeof(object).GetTypeInfo().Assembly.Location),
-          MetadataReference.CreateFromFile(Assembly.Load("mscorlib").Location));
+            MetadataReference.CreateFromFile(Assembly.Load(fileName).Location))).
+        AddReferences(
```
camrein, line 161: So far I applied the convention to use the dot of method chained calls as a prefix.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -158,14 +157,14 @@ public class ShortLivedProcessStarter {{
...
+            MetadataReference.CreateFromFile(typeof(object).GetTypeInfo().Assembly.Location),
+            MetadataReference.CreateFromFile(Assembly.Load("mscorlib").Location));
+      compilation = compilation.WithOptions(new CSharpCompilationOptions(OutputKind.ConsoleApplication));
```
camrein, line 164: You could continue the method chaining here since you neither explicitly name nor use the intermediate state.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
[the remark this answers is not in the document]
fabiomadge: Right, my mistake.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,187 @@
...
+
+      // Wait for the language server to kill itself by waiting until it closes the output stream.
+      await process.StandardOutput.ReadToEndAsync();
+      Thread.Sleep(200); // Give the process some time to die
+      try {
+        var languageServer = Process.GetProcessById(int.Parse(languageServerProcessId));
+        languageServer.Kill();
+        Assert.Fail("Language server should have killed itself if the parent is gone.");
+      } catch (ArgumentException) {
+        // Language server process is not running, pass the test.
```
camrein, line 80: There's a suitable assertion in MSTest:
```cs
Assert.ThrowsException<ArgumentException>(() => {
  var languageServer = Process.GetProcessById(int.Parse(languageServerProcessId));
  languageServer.Kill();
});
```

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -19,21 +21,51 @@ public static class DafnyLanguageServer {
       }
     }
 
-    public static LanguageServerOptions WithDafnyLanguageServer(this LanguageServerOptions options, IConfiguration configuration) {
+    public static LanguageServerOptions WithDafnyLanguageServer(this LanguageServerOptions options,
+        IConfiguration configuration, Action killLanguageServer) {
```
camrein, line 25: I think it makes more sense to name `killLanguageServer` like the event when it's raised rather than the action it's doing in a concrete implementation.

camrein (commented):
Looks good so far. I've added two more comments to the others.

PS: We may switch to XUnit in the future since the rest of Dafny's test suite is built around it.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
keyboardDrummer: I don't understand. I would name an event that is raised when a language server is killed `OnLanguageServerKilled`. But this argument is not an event, it's a function that when called will kill the language server. Shouldn't a function get the name of the action it performs?

fabiomadge (dismissed):

Source/DafnyLanguageServer/DafnyLanguageServer.cs
MikaelMayer: @camrein  I think that `killLanguageServer` is appropriate as well. This is not a callback about what to do when the language server is killed. It's actually the action that does kill the language server, and that has to be provided by the external object creating the language server.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,185 @@
...
+    /// another .NET project to the Dafny solution, which incurs a build time overhead.
+    /// The downside if this choice is that we have to define the source code of our binary in a string literal,
+    /// which makes it harder to develop.
+    /// </summary>
+    private static async Task<string> CreateDotNetDllThatStartsGivenFilepath(string filePathToStart) {
+      var code = @$"using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+public class ShortLivedProcessStarter {{
```
MikaelMayer, line 133: I like this name.

MikaelMayer (approved):
Well done for this one.
Next one interesting to solve would be the memory leak of a single DafnyLanguageServer with "save on change" that keeps too much in memory and goes to 4Gb of RAM 
in a few minutes.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
camrein: I was thinking about something like `OnShutDownRequested`. However, your arguments make sense including the decision to not make this parameter nullable.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -0,0 +1,181 @@
...
+        Id = 1,
+        Method = "initialize",
+        Params = new InitializeParams {
+          ProcessId = processId,
+          ClientInfo = new ClientInfo(),
+          Capabilities = new ClientCapabilities(),
+        }
+      });
+#pragma warning disable VSTHRD002
+      outputHandler.StopAsync().Wait();
```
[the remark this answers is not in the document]
camrein: Any reason for this? I see that the target buffer is a `MemoryStream`; thus, the actions tend to be synchronous anyway. However, it would make the warning suppression obsolete, as well as the use of `Wait` that would require manual unwrapping of the nested exceptions.

Source/DafnyLanguageServer/Program.cs
```diff
@@ -14,15 +16,17 @@ public class Program {
       var configuration = CreateConfiguration(args);
       InitializeLogger(configuration);
       try {
+        var cancelLanguageServer = new CancellationTokenSource();
```
camrein, line 19: This `CancellationTokenSource` is now unnecessary. The call to `cancelLanguageServer.Cancel()` can be replaced with `server.ForcefulShutdown()`. This change would make my comment regarding the disposal of the registration obsolete too since it can be omitted.

camrein (changes requested):
Thanks for the improvements. Since I belive the use of `CancellationTokenSource` is now unnecessary, I've requested one more change. After that, we're good to go.

Source/DafnyLanguageServer/Program.cs
keyboardDrummer: > Any reason for this? I see that the target buffer is a MemoryStream; thus, the actions tend to be synchronous anyway. However, it would make the warning suppression obsolete, as well as the use of Wait that would require manual unwrapping of the nested exceptions.

The reason is to make the return type a `string` instead of `Task<string>`. I don't see why computing a static value would require a `Task`. But.. I guess this goal is not worth the trouble I'm creating :-)

Source/DafnyLanguageServer/DafnyLanguageServer.cs
cpitclaudel: Yes, but in fact camrein suggested an even better alternative (an explicit assertion)

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
cpitclaudel: Then I think your solution makes sense, thanks!

cpitclaudel (approved):

camrein (approved):

PR #1666, opened 2021-12-24
Separate resolution and verification diagnostics

keyboardDrummer:
### Changes
- Enable go to definition before verification has completed
- Enable hover before verification has completed
- Enable signature help before verification has completed
- Publish parse and resolution diagnostics before verification has completed. Verification diagnostics from previous runs are migrated.

### Testing
- Add several tests that look at the behavior of the migration of verification diagnostics.
- Add test for resolution cancellation
- Add test that checks go to definition requests return before verification diagnostics
- Add test that checks hover requests return before verification diagnostics


Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -99,6 +138,26 @@ Position afterChangeEndOffset
         return new Position(changeStart.Line + changeEof.Line, characterOffset);
       }
 
+      private static Range? MigrateRange(Range rangeToMigrate, Range changeRange, Position afterChangeEndOffset) {
+        if (!rangeToMigrate.Contains(changeRange) && rangeToMigrate.Intersects(changeRange)) {
+          // Do not migrate ranges that partially overlap with the change
+          return null;
```
cpitclaudel, line 144: Don't you want to migrate one of the endpoints here?

Source/DafnyLanguageServer/Workspace/IDocumentDatabase.cs
```diff
@@ -38,21 +38,28 @@ public interface IDocumentDatabase {
...
     /// <summary>
     /// Tries to resolve a document with the specified identifier.
     /// </summary>
     /// <param name="documentId">The ID of the document to resolve.</param>
-    /// <returns>An instance of the managed document, <c>null</c> if the specified document was not found.</param>
+    /// <returns>An instance of the managed document, <c>null</c> if the specified document was not found.</returns>
     Task<DafnyDocument?> GetDocumentAsync(TextDocumentIdentifier documentId);
+
+    /// <summary>
+    /// Tries to resolve a document with the specified identifier.
```
cpitclaudel, line 59: Missing docstring update

cpitclaudel (commented):
Very nice to get this source of confusion addressed.  I guess the only remaining issue is displaying stale verification diagnostics while waiting for new verification results, but I'm not sure how we could solve that particular issue.  Does LSP let us treat some diagnostics differently?  Can we somehow mark these questionable diagnostics differently?

keyboardDrummer:
> Very nice to get this source of confusion addressed. I guess the only remaining issue is displaying stale verification diagnostics while waiting for new verification results, but I'm not sure how we could solve that particular issue. Does LSP let us treat some diagnostics differently? Can we somehow mark these questionable diagnostics differently?

We could prefix the messages from migrated verification diagnostics as with "Stale:". I was't sure whether that would be better so opted not to do it and wait for customer feedback.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
keyboardDrummer: Well, I think if the entire range of a diagnostics has been removed, for example an ensures clause has been removed, then we shouldn't migrate the diagnostic. 

Similarly, if either the start or end of a diagnostic has been removed, I'm guessing it's better not to migrate it.

MikaelMayer:
> > Very nice to get this source of confusion addressed. I guess the only remaining issue is displaying stale verification diagnostics while waiting for new verification results, but I'm not sure how we could solve that particular issue. Does LSP let us treat some diagnostics differently? Can we somehow mark these questionable diagnostics differently?
> 
> We could prefix the messages from migrated verification diagnostics as with "Stale:". I was't sure whether that would be better so opted not to do it and wait for customer feedback.

I'm waiting for their feedback too. Instead adding "stale", one might consider down the road indicators in the gutter, like opacity of 0.5 for verification indicators, indicating that they are obsolete and are going to be recomputed soon.

keyboardDrummer:
> I'm waiting for their feedback too. Instead adding "stale", one might consider down the road indicators in the gutter, like opacity of 0.5 for verification indicators, indicating that they are obsolete and are going to be recomputed soon.

Well these are diagnostics, which are separate from gutter indicators. Diagnostics also show in the problems panel. There's not much to change about diagnostics except the message and the severity level. You could consider setting the severity for stale diagnostics to warning instead of error.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -99,65 +109,107 @@ ISymbolTableRelocator symbolTableRelocator
...
         cancellationSource
       );
       documents[documentUri] = updatedEntry;
-      return await updatedEntry.Document;
+      return resolvedDocumentTask.ToObservable().Concat(verifiedDocument.ToObservable());
+    }
+
+
+#pragma warning disable VSTHRD200
+    private static Task<T> FirstSuccessfulTask<T>(params Task<T>[] tasks) {
```
MikaelMayer, line 127: Please rename this method `FirstSuccessfulAsync<T>` to get rid of the warning removal directives.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -99,65 +109,107 @@ ISymbolTableRelocator symbolTableRelocator
...
+#pragma warning restore VSTHRD200
+      var taskList = tasks.ToList();
+      var tcs = new TaskCompletionSource<T>();
+      int remainingTasks = taskList.Count;
+      foreach (var task in taskList) {
+#pragma warning disable VSTHRD110
```
MikaelMayer, line 133: I believe you can remove this warning by assigning to a forgetting pattern instead of pragma directives, like that:

    _ = task.ContinueWith( ...

See https://stackoverflow.com/questions/14903887/warning-this-call-is-not-awaited-execution-of-the-current-method-continues

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -99,65 +109,107 @@ ISymbolTableRelocator symbolTableRelocator
...
+            }
+          }
+        }, TaskScheduler.Default);
+      }
+      return tcs.Task;
     }
 
-    private async Task<DafnyDocument> ApplyChangesAsync(DafnyDocument oldDocument, DidChangeTextDocumentParams documentChange, CancellationToken cancellationToken) {
+    private async Task<DafnyDocument> ApplyChangesAsync(Task<DafnyDocument> oldDocumentTask, DidChangeTextDocumentParams documentChange, CancellationToken cancellationToken) {
+#pragma warning disable VSTHRD003
```
MikaelMayer, line 151: Could you just await the document at the call site to get rid of this warning? That way you also help reduce the number of created tasks.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -99,65 +109,107 @@ ISymbolTableRelocator symbolTableRelocator
...
           SerializedCounterExamples = null,
-          LoadCanceled = true
+          LoadCanceled = true,
+          OldVerificationDiagnostics = migratedVerificationDiagnotics
         };
       }
     }
 
-    public Task<DafnyDocument> SaveDocumentAsync(TextDocumentIdentifier documentId) {
+    public IObservable<DafnyDocument> SaveDocument(TextDocumentIdentifier documentId) {
```
MikaelMayer, line 188: This change will conflict with @atomb 's changes. @atomb can you comment on this version?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -99,65 +109,107 @@ ISymbolTableRelocator symbolTableRelocator
       }
       databaseEntry.CancelPendingUpdates();
       var cancellationSource = new CancellationTokenSource();
+      var previousDocumentTask = FirstSuccessfulTask(databaseEntry.VerifiedDocument, databaseEntry.ResolvedDocument);
```
MikaelMayer, line 112: Oh, so this is saying "If verification previously succeeded, keep the verified document, else, keep the resolved document". Can you add a comment on that ? because it took me a long time to figure out that these are not concurrent tasks.
Or, rename `FirstSuccessfullTask` to `FirstOrElse` to be more self-explanatory.

MikaelMayer (changes requested):
That's a great start and the idea to migrate to IObservable seems legit. Here are some comments to address.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: I think the conflicts should be fine to resolve :-)

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: I copied the method `FirstSuccessfullTask` from stack overflow, and the name makes sense to me. The method accepts multiple tasks, and returns the task that is the first to complete successfully. The name `FirstOrElse` seems less explicit to me, and seems to imply that the order of the arguments matters. What is unclear about the name `FirstSuccessfullTask` to you?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: I can't because it's not an async method.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -54,42 +56,50 @@ ISymbolTableRelocator symbolTableRelocator
...
         return documentLoader.CreateUnloaded(textDocument, CancellationToken.None);
       }
     }
 
-    private async Task<DafnyDocument> LoadAndVerifyAsync(TextDocumentItem textDocument, bool verify, CancellationToken cancellationToken) {
-      var document = await documentLoader.LoadAsync(textDocument, cancellationToken);
-      if (!verify || document.Errors.HasErrors) {
-        return document;
+    private async Task<DafnyDocument> VerifyAsync(Task<DafnyDocument> documentTask, bool verify, CancellationToken cancellationToken) {
+#pragma warning disable VSTHRD003
```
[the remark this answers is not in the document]
keyboardDrummer: > I think @cpitclaudel has a point though. If a change occurs before and after a border of a diagnostic, that position of the diagnostic won't be relevant anymore and will land in new text.
Hence, I too believe you should move the boundary of the diagnostic, while it's temporary, so that it does not spill out in the new text.

I'm not sure what you mean by "If a change occurs before and after a border of a diagnostic". Do you mean the text deleted by the change covers the entire range of the diagnostic?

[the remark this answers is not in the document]
keyboardDrummer: > Could you remove these warning removal directives by awaiting for the document before calling VerifyAsync ? It is not catching exceptions anyway.

I can't because the caller is not `async`.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
MikaelMayer: Do we agree that `databaseEntry.ResolvedDocument` will always terminate before `databaseEntry.VerifiedDocument` ?
The question on StackOverflow was assuming different tasks producing the same homogeneous result, in which the order did not matter.
`FirstSuccessfullTask` assumes no relationship between the tasks, and offers no guarantee about which one will be returned if both are already completed by the time this function is called. Thus, in some (rare) cases, it might return the Resolution document instead of the Verification document although both of them are available, despite you putting `VerifiedDocument` first.
I believe that the order matter to us, because else you would have just waited for the Resolution task. If Verification was finished, you do want to ensure you get it instead of Resolution, right?

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
MikaelMayer: >I'm not sure what you mean by "If a change occurs before and after a border of a diagnostic". Do you mean the text deleted by the change covers the entire range of the diagnostic? In that case this PR is not migrating the diagnostic, since it seems reasonable to guess that by deleting the offending code the diagnostic will disappear as well.

For example
```
^^^^^^^^^^^^^^^ (diagnostic)
          -------- (text removal)
^^^^^^^^^^ (diagnostic)
```

And in case the change removes the entire diagnostic, you should also delete it if it's not already what you are doing. Since these diagnostics are published and result in red squiggles, you don't want red squiggles in unrelated text.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: Well spotted! Will fix.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
keyboardDrummer: This PR will, if the change overlaps either the beginning or end of a diagnostic, delete instead of migrate the diagnostic.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
MikaelMayer: Yes, it's the most likely, Now, I hope I conveyed the intuition that the use of `FirstSuccessfullTask` is not obvious in this context, and I would like to put a comment about that, so that next time someone looks at this code, he does understand that this is a special case, and we have an empirically verified assumption - that FirstSuccessfullTask will return the result of the first task if both of them completed.

On another thought, if resolution wasn't finished before, I believe it will be canceled by `databaseEntry.CancelPendingUpdates();`, so no task will actually terminate successfully and you'll wait forever, right?

keyboardDrummer: Shall I add a comment to `FirstSuccessfullTask` to clarify what it does?

> On another thought, if resolution wasn't finished before, I believe it will be canceled by databaseEntry.CancelPendingUpdates();, so no task will actually terminate successfully and you'll wait forever, right?

Then `FirstSuccessfullTask` returns a cancelled task.

MikaelMayer: Thanks for the clarification, I wasn't sure.
Either add a comment to FirstSuccessfulTask (and the assumption it always returns the first task if both are completed), or add a comment at the call site.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/IRelocator.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.Dafny.LanguageServer.Language.Symbols;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using System.Threading;
+
+namespace Microsoft.Dafny.LanguageServer.Workspace.ChangeProcessors {
+  /// <summary>
+  /// Implementations of this interface are responsible to relocate symbols and diagnostics
+  /// to their new positions according to the given <see cref="DidChangeTextDocumentParams"/>.
+  /// </summary>
+  public interface IRelocator {
```
camrein, line 11: We may want to consider keeping these two relocation mechanisms in distinct classes since they're unrelated, at least separate interfaces.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -4,47 +4,86 @@
...
 
-      public SymbolTable MigrateSymbolTable() {
+      public IReadOnlyList<Diagnostic> MigrateDiagnostics(IReadOnlyList<Diagnostic> originalDiagnostics) {
+        return contentChanges.Aggregate(originalDiagnostics, MigrateDiagnostics);
+      }
+
+      private IReadOnlyList<Diagnostic> MigrateDiagnostics(IReadOnlyList<Diagnostic> originalDiagnostics, TextDocumentContentChangeEvent change) {
+        var result = new List<Diagnostic>();
+        foreach (var diagnostic in originalDiagnostics) {
+          cancellationToken.ThrowIfCancellationRequested();
```
camrein, line 56: I suggest extracting the body of this foreach-loop. Besides more clarity, it allows omitting the manual filling of the `result` list since it can be replaced with a LINQ statement.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -4,47 +4,86 @@
...
+          }
+
+          var afterChangeEndOffset = GetPositionAtEndOfAppliedChange(change.Range, change.Text);
+          var newRange = MigrateRange(diagnostic.Range, change.Range, afterChangeEndOffset);
+          if (newRange == null) {
+            continue;
+          }
+
+          var newRelated = diagnostic.RelatedInformation?.SelectMany(related => {
+            var migratedRange = MigrateRange(related.Location.Range, change.Range, afterChangeEndOffset);
```
camrein, line 68: I also suggest extracting this lambda expression in a different method.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -114,65 +116,111 @@ ISymbolTableRelocator symbolTableRelocator
...
+      var taskList = tasks.ToList();
+      var tcs = new TaskCompletionSource<T>();
+      int remainingTasks = taskList.Count;
+      foreach (var task in taskList) {
+        if (tcs.Task.IsCompleted) {
+          break;
+        }
+
+        _ = task.ContinueWith(t => {
+          if (task.Status == TaskStatus.RanToCompletion) {
```
camrein, line 147: `t.IsCompletedSuccessfully`

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -114,65 +116,111 @@ ISymbolTableRelocator symbolTableRelocator
...
+            tcs.TrySetResult(t.Result);
+          } else if (Interlocked.Decrement(ref remainingTasks) == 0) {
```
camrein, line 149: Wouldn't the checks of the other tasks be unnecessary if you handle the failure of the current task here?

camrein (changes requested):
Looks great and thanks for your efforts on the test suite.
Would you mind elaborating on the motivation behind the switch to `IObservable`?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -114,65 +116,111 @@ ISymbolTableRelocator symbolTableRelocator
...
-      return await updatedEntry.Document;
+      return resolvedDocumentTask.ToObservable().Concat(verifiedDocument.ToObservable());
+    }
+
+
+    /// <summary>
+    /// Returns the first task that completes successfully.
+    /// If multiple tasks are already completed, the first one is returned.
+    /// </summary>
+    private static Task<T> FirstSuccessfulAsync<T>(params Task<T>[] tasks) {
```
camrein, line 137: Maybe it makes sense to create a separate class for this method so it can be unit tested.

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
```diff
@@ -0,0 +1,57 @@
...
+  protected DiagnosticsReceiver diagnosticReceiver;
+
+  [TestInitialize]
+  public virtual async Task SetUp() {
+    diagnosticReceiver = new();
+    client = await InitializeClient(options => options.OnPublishDiagnostics(diagnosticReceiver.NotificationReceived));
+  }
+
+  protected void ApplyChange(ref TextDocumentItem documentItem, Range range, string text) {
+    documentItem = documentItem with { Version = documentItem.Version + 1 };
```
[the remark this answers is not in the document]
keyboardDrummer: With a ref parameter you can force the version to be updated, otherwise you depend on the caller to update `documentItem`

Source/DafnyLanguageServer/Workspace/ChangeProcessors/IRelocator.cs
keyboardDrummer: Actually, I don't understand why we're using interfaces that have only one implementation. I would understand if we're stubbing the interfaces in testing, but we're not. What benefit is gained from having `IRelocator` ?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: Thanks! I added unit tests and they helped spot a rare concurrency bug that I'm guessing has been the reason the build was failing non-deterministically.

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -61,7 +61,8 @@ public class DiagnosticErrorReporter : ErrorReporter {
           Message = error.Msg,
           Range = error.Tok.GetLspRange(),
           RelatedInformation = relatedInformation,
-          Source = VerifierMessageSource.ToString()
+          Source = VerifierMessageSource.ToString(),
+          Code = "Verification"
```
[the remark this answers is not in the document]
keyboardDrummer: Good suggestion, thanks 👍

Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
```diff
@@ -67,5 +89,19 @@ public class DafnyLanguageServerTestBase : LanguageServerTestBase {
...
     }
+
+    public static string PrintDiagnostics(IEnumerable<Diagnostic> items) {
+      return PrintEnumerable(items.Select(PrintDiagnostic));
+    }
+
+    public static string PrintDiagnostic(Diagnostic diagnostic) {
+      var relatedPrint = string.Join(", ", diagnostic.RelatedInformation?.
+        Select(r => $"at {r.Location} saying '{r.Message}'") ?? Array.Empty<string>());
+      return $"Diagnostic at {diagnostic.Range} saying '{diagnostic.Message}', related: " + relatedPrint;
```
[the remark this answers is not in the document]
keyboardDrummer: None apart from that I was rushing : ) changed.

keyboardDrummer:
> Looks great and thanks for your efforts on the test suite. Would you mind elaborating on the motivation behind the switch to `IObservable`?

What would be the alternative to using `IObservable`? I wanted a return type that could return multiple values of `DafnyDocument`, one after another.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/IRelocator.cs
camrein: I agree that it's over-engineered to have an interface for everything while there is only one implementation.
It's a usual convention for (ASP).NET Core applications to create an interface and only inject these rather than the concrete type. This helps testing since the interfaces can be mocked. However, due to the strong dependence on `Program`, things got difficult to unit test separately hence why there mostly only integration tests.

One thing I have in mind is to create an interface for `DafnyDocument` and its dependencies to make unit testing easier in the future. I didn't do that from the beginning because one of my goals when I initially created the language server was to not touch the compiler code base.

camrein:
> > Looks great and thanks for your efforts on the test suite. Would you mind elaborating on the motivation behind the switch to `IObservable`?
> 
> What would be the alternative to using `IObservable`? I wanted a return type that could return multiple values of `DafnyDocument`, one after another.

Ah I see, I didn't notice this requirement. However, keep in mind that subscriptions usually have to be disposed (unregistered) manually, unless the source completes (i.e., `OnCompleted` signal is sent). Otherwise, we'll have a memory leak of subscriptions.
If I'm right, you're converting the resolution and verification tasks into observables and concatenate these. I guess that both send `OnCompleted` individually to the concatenation stream, but I don't know if the concatenation stream handles these.

If this switch to `IObservable` is only necessary to handle the diagnostics, we could consider relocating these, i.e., no longer handle them inside `DafnyTextDocumentHandler`.

keyboardDrummer:
>  However, keep in mind that subscriptions usually have to be disposed (unregistered) manually, unless the source completes (i.e., `OnCompleted` signal is sent). 

I assume the concatenated stream completes when the individual streams complete. The documentation for Concat reads "Concatenates the second observable sequence to the first observable sequence upon successful termination of the first."

> If this switch to `IObservable` is only necessary to handle the diagnostics, we could consider relocating these, i.e., no longer handle them inside `DafnyTextDocumentHandler`.

What's the goal?

MikaelMayer (approved):
Looks good to me.

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
```diff
@@ -0,0 +1,57 @@
...
+      ContentChanges = new[] {
+        new TextDocumentContentChangeEvent {
+          Range = range,
+          Text = text
+        }
+      }
+    });
+  }
+
+  public async Task AssertNoDiagnosticsAreComing() {
```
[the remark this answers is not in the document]
keyboardDrummer: I'm trying to prove that however long you would wait, no more diagnostics will come. 

By waiting until all documents are verified I'm making sure all the verification diagnostics are ready to be published, then I use the extra document trick to add one more diagnostic to the queue and assume that once that one is published, no more diagnostics are verification and waiting to be published.

By creating a document with a parse error I'm guaranteeing that it only returns resolution and no verification diagnostics, so I have to wait for exactly one diagnostic from that document and don't need to think about what configuration there is for automatic verification.

keyboardDrummer:
@camrein could you comment on how OmniSharp handles concurrency LSP requests? Can requests be handled concurrently even in their synchronous part, before they call `await`?

Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
```diff
@@ -70,19 +70,18 @@ public class DafnyTextDocumentHandler : TextDocumentSyncHandlerBase {
...
     public override Task<Unit> Handle(DidSaveTextDocumentParams notification, CancellationToken cancellationToken) {
       logger.LogTrace("received save notification {DocumentUri}", notification.TextDocument.Uri);
-      HandleUpdateAndPublishDiagnosticsAsync(documents.SaveDocumentAsync(notification.TextDocument));
+      documents.SaveDocument(notification.TextDocument).Subscribe(HandleUpdateAndPublishDiagnostics);
       return Unit.Task;
     }
 
-    private async Task HandleUpdateAndPublishDiagnosticsAsync(Task<DafnyDocument> documentAction) {
+    private void HandleUpdateAndPublishDiagnostics(DafnyDocument document) {
       try {
```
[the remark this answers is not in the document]
cpitclaudel: https://github.com/dafny-lang/dafny/pull/1857

PR #1667, opened 2021-12-25
chore: fix warnings

fabiomadge:
Likely a side effects of moving to `net6.0`. The CI build warnings are caused by the fact that `SelfContained` is no longer implied by specifying a runtime. Making it explicit triggers a problem where the `RuntimeIdentifier` is no longer handed down to included projects. Setting the new environment variable works around that.

### DIFF
--- a/Scripts/package.py
+++ b/Scripts/package.py
@@ -128,18 +128,21 @@ class Release:
             "-f", "net6.0",
             "-o", self.buildDirectory,
             "-r", self.target,
+            "--sc",
             "-c", "Release"])
         run(["dotnet", "publish", path.join(SOURCE_DIRECTORY, "DafnyServer", "DafnyServer.csproj"),
             "--nologo",
             "-f", "net6.0",
             "-o", self.buildDirectory,
             "-r", self.target,
+            "--sc",
             "-c", "Release"])
         run(["dotnet", "publish", path.join(SOURCE_DIRECTORY, "DafnyDriver", "DafnyDriver.csproj"),
             "--nologo",
             "-f", "net6.0",
             "-o", self.buildDirectory,
             "-r", self.target,
+            "--sc",
             "-c", "Release"])
 
     def pack(self):
--- a/Scripts/package.py
+++ b/Scripts/package.py
@@ -128,21 +128,18 @@ class Release:
             "-f", "net6.0",
             "-o", self.buildDirectory,
             "-r", self.target,
-            "--sc",
             "-c", "Release"])
         run(["dotnet", "publish", path.join(SOURCE_DIRECTORY, "DafnyServer", "DafnyServer.csproj"),
             "--nologo",
             "-f", "net6.0",
             "-o", self.buildDirectory,
             "-r", self.target,
-            "--sc",
             "-c", "Release"])
         run(["dotnet", "publish", path.join(SOURCE_DIRECTORY, "DafnyDriver", "DafnyDriver.csproj"),
             "--nologo",
             "-f", "net6.0",
             "-o", self.buildDirectory,
             "-r", self.target,
-            "--sc",
             "-c", "Release"])
 
     def pack(self):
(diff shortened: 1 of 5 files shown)


Source/DafnyDriver/DafnyDriver.csproj
```diff
@@ -14,6 +14,10 @@
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
   </PropertyGroup>
+
+  <PropertyGroup Condition="$(RUNTIME_IDENTIFIER) != ''">
```
keyboardDrummer, line 18: Could you add a comment explaining why we need this? Maybe add a reference to the issue in .NET that causes it: https://github.com/dotnet/sdk/issues/10566

Scripts/package.py
```diff
@@ -122,25 +122,30 @@ def build(self):
 
         if path.exists(self.buildDirectory):
             shutil.rmtree(self.buildDirectory)
+        env = dict(os.environ)
+        env["RUNTIME_IDENTIFIER"] = self.target
```
keyboardDrummer, line 126: Is it not possible to update `os.environ` so you don't need the dictionary copying and additional parameters?

Source/XUnitExtensions/Lit/OutputCheckCommand.cs
```diff
@@ -41,7 +41,7 @@ private abstract record CheckDirective(string File, int LineNumber) {
   }
 
   private record CheckRegexp(string File, int LineNumber, Regex Pattern) : CheckDirective(File, LineNumber) {
-    public static CheckDirective Parse(string file, int lineNumber, string arguments) {
+    public new static CheckDirective Parse(string file, int lineNumber, string arguments) {
```
MikaelMayer, line 44: What's this syntax change doing? I don't know this C# feature.

Scripts/package.py
```diff
@@ -122,25 +122,30 @@ def build(self):
...
         run(["make", "--quiet", "clean"])
         run(["dotnet", "publish", path.join(SOURCE_DIRECTORY, "DafnyLanguageServer", "DafnyLanguageServer.csproj"),
             "--nologo",
             "-f", "net6.0",
             "-o", self.buildDirectory,
             "-r", self.target,
-            "-c", "Release"])
+            "--sc",
```
MikaelMayer, line 133: I had to figure out manually that this option means "self-contained". Since this is our script, can you use the non abbreviate version?

Scripts/package.py
MikaelMayer: Looks like this new `env` argument is not used in all the calls to `run`, only in the three publishing directives. I think it makes sense to control the environment modification to affect only the child processes who need it. It'll also be easier to reason about than changing the environment globally.

Source/DafnyDriver/DafnyDriver.csproj
MikaelMayer: Yes that would be helpful for maintainability.

MikaelMayer (changes requested):
Almost there ! Glad you could fix those warnings.

Source/XUnitExtensions/Lit/OutputCheckCommand.cs
fabiomadge: It's making the hiding of the base class's implementation explicit. ([docs](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/knowing-when-to-use-override-and-new-keywords))

MikaelMayer (approved):
Looks good to me. Great that we are getting rid of those warnings in code review.

Source/DafnyServer/DafnyServer.csproj
```diff
@@ -15,6 +15,11 @@
     <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
   </PropertyGroup>
 
+  <!-- Working around some stange behavior in dotnet publish: https://github.com/dotnet/sdk/issues/10566 -->
```
keyboardDrummer, line 18: Nitpick: "some stRange behavior" -> "a bug" ?

PR #1669, opened 2021-12-27
chore: fix test reports

fabiomadge:
There are error notifications about failures in the CI. The reports of parse errors `on` should go away.

### DIFF
--- a/.github/workflows/release-downloads.yml
+++ b/.github/workflows/release-downloads.yml
@@ -5,7 +5,8 @@ name: Test Release Downloads
 on:
   release:
     types: [ published ]
-  workflow_dispatch: // For manual testing through the Actions UI
+  # For manual testing through the Actions UI
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/test-report.yml
+++ b/.github/workflows/test-report.yml
@@ -1,13 +1,14 @@
-name: 'Test Report'
+name: "Test Report"
 on:
   workflow_run:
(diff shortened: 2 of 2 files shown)


fabiomadge:
This seems to do some of what I'm promising, but might need to be in master to be fully evaluated.

keyboardDrummer:
Seems like the build is quite unstable. Two errors that don't seem to relate to this change:

```
/home/runner/.dotnet/sdk/6.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(172,5): error MSB4018: System.IO.IOException: The process cannot access the file '/home/runner/work/dafny/dafny/dafny/Binaries/net6.0/linux-x64/DafnyServer.deps.json' because it is being used by another process. 
```

and

```
Error: /Users/runner/.dotnet/sdk/6.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(550,5): error NETSDK1177: Failed to sign apphost with error code 0: /Users/runner/work/dafny/dafny/dafny/Source/DafnyServer/obj/Release/net6.0/osx-x64/apphost: No such file or directory [/Users/runner/work/dafny/dafny/dafny/Source/DafnyServer/DafnyServer.csproj]
```

keyboardDrummer (approved):
> This seems to do some of what I'm promising, but might need to be in master to be fully evaluated.

Could you be more specific about what you can see it does and what you're still unsure about?

PR #1673, opened 2021-12-28
fix(lsp): Use VerificationStarted instead of VerificationInProgress

cpitclaudel:
This last published version of the Dafny extension displays nothing if it receives an
unknown compilation status.  Repurposing the existing VerificationStarted status
plays nicer with older versions of the extension.

Reported-by: @keyboardDrummer
Suggested-by: @MikaelMayer


MikaelMayer (approved):
In a second step, we should make the IDE not publish any unknown status, so that one day we can revert this commit.

PR #1674, opened 2021-12-28
chore: enable test reports

fabiomadge:
`test-report.yaml` will not run smoothly until we give it some files. This might do the trick. Otherwise I propose disabling it for the time being.


keyboardDrummer (commented):
What is change you've made that you're hoping will produce an effect?

fabiomadge:
We now actually produce logs again when running the integration tests. We also let the reporting script handle downloading the correct artifact, which didn’t work before. This part isn’t tested, so it still might not work.

keyboardDrummer:
Is it possible to fully test this on your own fork?

fabiomadge:
No, it isn't, but I’ll set up a different testing org for other CI experiments I want to do.

fabiomadge:
Confirmed to work in a separate repo.

keyboardDrummer (approved):

PR #1675, opened 2021-12-30
Hello.dfy Testfile for Python compiler.

prvshah51:


Test/comp/Hello.dfy
```diff
@@ -10,4 +11,4 @@ method Main() {
   print "x is ", x, "\n";
   var y := false;
   print "y is ", y, "\n";
-}
+}
```
fabiomadge, line 14: I think the majority of the existing files have the \n at the end.

PR #1677, opened 2021-12-30
Elephant operator for methods with generic return types, removed compiler-crashing default value for elephant operator

### LINKED ISSUE #1514: Elephant `:-` does not unwrap through a subtype
The line `var res :- foo.Bar(a);` gives the error "The type of the first expression is not a failure type in :- statement"

```dafny
  trait Foo<A, B> {
    method Bar(a: A) returns (r: B)
  }

  type FooWithResult<A, B, E> = Foo<A, Result<B, E>>

  method Bar<A, B, E>(
    a: A,
    foo: FooWithResult<A, B, E>
  ) returns (r: Result<B, E>) {
    var res :- foo.Bar(a);
  }
```

However Dafny is happy with:

```dafny
  method Bar<A, B, E>(
    a: A,
    foo: FooWithResult<A, B, E>
  ) returns (r: Result<B, E>) {
    var res := foo.Bar(a);

    if res.Success? {
      return Success(res.value);
    } else {
      return Failure(res.error);
    }
  }
```

### LINKED ISSUE #1553: C# compiler bug: type descriptor not in scope
Program to reproduce below. 

```dafny
include "../../libraries/src/Wrappers.dfy"
module Actions {
  import opened Wrappers
  trait {:termination false} Action<A, R>
  {
    method Invoke(a: A)
      returns (r: R)
      ensures Ensures(a, r)
    predicate Ensures(a: A, r: R)
  }
  trait {:termination false} ActionWithResult<A, R, E>
    extends Action<A, Result<R, E>>
  {
    method Invoke(a: A)
      returns (res: Result<R, E>)
      ensures Ensures(a, res)
  }
  method Map<A, R>(
    action: Action<A, R>,
    s: seq<A>
  )
    returns (res: seq<R>)
    ensures |s| == |res|
    ensures
      forall i ::
        && 0 <= i < |s|
      ==>
        action.Ensures(s[i], res[i])
  {
    var rs := [];
    for i := 0 to |s|
      invariant |s[..i]| == |rs|
      invariant forall j ::
        && 0 <= j < i
      ==>
        action.Ensures(s[j], rs[j])
    {
      var r := action.Invoke(s[i]);
      rs := rs + [r];
    }
    return rs;
  }
  method MapWithResult<A, R, E>(
    action: ActionWithResult<A, R, E>,
    s: seq<A>
  )
    returns (res: Result<seq<R>, E>)
    ensures
      res.Success?
    ==>
      |s| == |res.value|
    ensures
      res.Success?
    ==>
        (forall i ::
          && 0 <= i < |s|
        ==>
          action.Ensures(s[i], Success(res.value[i])))
  {
    var rs := [];
    for i := 0 to |s|
      invariant |s[..i]| == |rs|
      invariant forall j ::
        && 0 <= j < i
      ==>
        action.Ensures(s[j], Success(rs[j]))
    {
      var r :- action.Invoke(s[i]);
      rs := rs + [r];
    }
    return Success(rs);
  }
```

Compilation error when compiling to C#:

```
(2344,22): error CS0103: The name '_td___R' does not exist in the current context

(2345,106): error CS0103: The name '_td___R' does not exist in the current context
```

### LINKED ISSUE #1679: Compilation error: error CS0103: The name '_td___T' does not exist in the current context
This issue is related to #1514  but concerns the compiler and the elephant operator. Here is a MVE:

```
method s<E>() returns (r: Option<E>)  {
}

method test<T>() returns (r: Option<T>) {
  var x :- s<T>();
  r := Some(x);
}

method Main() {
  var x := test<int>();
  print "ok";
}
```

Although it verifies and compiles to c#, the compilation of c# gives

    (2099,20): error CS0103: The name '_td___T' does not exist in the current context

An easy fix is to write `method test<T(0)>()` instead of `method test<T>()`. Indeed, the translation of `var x :- s<T>();` is:

```
var x: T;
var valueOrError := s<T>();
if(valueOrError.IsFailure()) {
  return valueOrError.PropagateFailure();
}
x := valueOrError.Extract();
```

The problem in the original code is that, if T is not auto-initializable, `var x: T` will try to initialize x but the type descriptor is not available, hence the compilation error.

Two solutions here:
1. Either require that T to be auto-initializable
2. Or in case of declarations, move the _declaration_ `var`  to when x is actually initialized, so that there is no need for auto-initializable descriptor.

I'm trying to implement 2 right now.

MikaelMayer:
This PR will fix #1514  and fix #1679. and fix #1553.

Changes:
* Added a new field [`ResolvedOutparameterTypes`](https://github.com/dafny-lang/dafny/pull/1677/files#diff-fa257574acd41f12645bbc3b82e23fc93ac386d13a9fb58c0f5f54944a8359e3R10226) in [`MemberSelectExpr`](https://github.com/dafny-lang/dafny/pull/1677/files#diff-fa257574acd41f12645bbc3b82e23fc93ac386d13a9fb58c0f5f54944a8359e3R10226) to record the inferred types of the output parameters when they exist 
* Populated this new field at resolution time in an [identical way of how functions are treated](https://github.com/dafny-lang/dafny/pull/1677/files#diff-7d845d452486b0b59700ca5c133f2648f76d44616148081d51efffe32dbc6e7aR16674).
* Use this new field [in place of the return value of a function](https://github.com/dafny-lang/dafny/pull/1677/files#diff-7d845d452486b0b59700ca5c133f2648f76d44616148081d51efffe32dbc6e7aR13220) so that inference works also on methods
* Enhanced the [error message](https://github.com/dafny-lang/dafny/pull/1677/files#diff-7d845d452486b0b59700ca5c133f2648f76d44616148081d51efffe32dbc6e7aR13220) in case the return type is not compatible with the elephant operator
*  [Removed the default value for variables declared using the elephant operator](https://github.com/dafny-lang/dafny/pull/1677/files#diff-998a9162635a910ca366a26c8f679ea96d6bbb3f703a52ca1487fd19599de49dR3079), since this default value is always overridden and furthermore leads to crashes when a generic type is not auto-initializable.
* Added three tests that were not passing before, that are now passing.
  * [git-issue-1514.dfy](https://github.com/dafny-lang/dafny/pull/1677/files#diff-dd6af6bd57af7c3a49fbf398b89826ef8d89eece27b953ee9fa31d8cb0a70d82R1) was crashing at compile time 
  * [git-issue-1514b.dfy](https://github.com/dafny-lang/dafny/pull/1677/files#diff-d1eeb2b737b2303a2c90044acdc15ef06c7b50720e47a43dabf79c50e8a94989R1) and [git-issue-1514c.dfy](https://github.com/dafny-lang/dafny/pull/1677/files#diff-d17a8bce9c0a174caeb2a8784e335727bc60ede3171d1d222fa0178ff7bb6386R1) triggered unexpected resolution errors


Test/git-issues/git-issue-1514.dfy
```diff
@@ -0,0 +1,31 @@
...
+    requires None?
+  {
+    None
+  }
+  function method Extract(): T
+    requires Some?
+  {
+    value
+  }
+}
```
fabiomadge, line 18:
```
include "../libraries/src/Wrappers.dfy"
import opened Wrappers
```
This should do the trick, right?

MikaelMayer: That's a good idea. I never saw that the library was imported within the Test folder itself.

Source/Dafny/Resolver.cs
```diff
@@ -13216,7 +13217,10 @@ private class RBranchExpr : RBranch {
         call = (asx.Lhs.Resolved as MemberSelectExpr)?.Member as Method;
         if (call != null) {
           // We're looking at a method call
-          if (call.Outs.Count != 0) {
+          var outTypes = (asx.Lhs.Resolved as MemberSelectExpr)?.ResolvedOutparameterTypes;
+          if (outTypes != null && outTypes.Count != 0) {
+            firstType = outTypes[0];
+          } else if (call.Outs.Count != 0) {
```
RustanLeino, line 13223: The information you're storing in the new `ResolvedOutparameterTypes` is already available in the `MemberSelectExpr` node. Instead of the new lines 13220-13223, just set `firstType` as follows:

    var typeMap = ((MemberSelectExpr)asx.Lhs.Resolved).TypeArgumentSubstitutionsWithParents();
    firstType = SubstType(call.Outs[0].Type, typeMap);

You'll also need to fix another problem in `Translator.cs`, namely to move

    vdecl.Locals.Iter(RemoveDefiniteAssignmentTracker);

out from the `else` branch (so that it's done whether or not `vdecl.Update is AssignOrReturnStmt` holds) in method `RemoveDefiniteAssignmentTrackers`.

RustanLeino (changes requested):
I think two of these fixes are better done a different way.

* The information you store in `ResolvedOutparameterTypes` is already available in `MemberSelectExpr`. Bug #1514 is that this information was not applied. I show a fix in the code comments.
* Bug #1679 is a bigger bug that manifests itself for any `var x: T;` statement where `T` is not an auto-init type. To get the more general fix, see my comments in the code.

Source/Dafny/Resolver.cs
```diff
@@ -13388,11 +13392,12 @@ private class RBranchExpr : RBranch {
...
-          new UpdateStmt(s.Tok, s.Tok,
-            new List<Expression>() { lhsExtract },
-            new List<AssignmentRhs>() { new ExprRhs(VarDotMethod(s.Tok, temp, "Extract")) }
-            ));
+        var assignStatement = new UpdateStmt(s.Tok, s.EndTok,
+          new List<Expression>() { lhsExtract },
+          new List<AssignmentRhs>() { new ExprRhs(VarDotMethod(s.Tok, temp, "Extract")) }
+        );
+
+        s.ResolvedStatements.Add(assignStatement);
```
fabiomadge, line 13400: No change here, right?

Source/Dafny/Compilers/Compiler.cs
```diff
@@ -3068,14 +3068,15 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
...
+          bool willBeAssignedAnyway = s.Update is AssignOrReturnStmt;
           if (!hasRhs && s.Update is UpdateStmt u) {
             if (i < u.Rhss.Count && u.Rhss[i] is HavocRhs) {
               // there's no specific initial value
             } else {
               hasRhs = true;
             }
           }
-          TrLocalVar(local, !hasRhs, wr);
+          TrLocalVar(local, !hasRhs && !willBeAssignedAnyway, wr);
```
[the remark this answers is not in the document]
MikaelMayer: Nice. Right, it makes sense to not initialize it if the type itself is not known to have a value, that's even better.

However, even if the type is known to have a value, we still don't need to initialize it if the statement is an `:-` with a right hand side.
Hence I suggest the strongest statement:

          TrLocalVar(local, !hasRhs && local.Type.KnownToHaveToAValue(false) && !willBeAssignedAnyway, wr);

Are you ok with it, or do you still prefer to do the initialization although it will be either overriden or discarded for the case of the AssignAndUpdate statement?

Source/Dafny/Resolver.cs
MikaelMayer: The doc of `TypeArgumentSubstitutionsWithParents` says `NOTE: This method should be called only when all types have been fully and successfully resolved.`
We are not in the phase when all types have been successfully resolved, because we are still resolving symbols inside the AssignAndReturnStmt. Should I update the comment?

I did implement your suggested change.

Test/git-issues/git-issue-970b.dfy.expect
```diff
@@ -1,5 +1,5 @@
 git-issue-970b.dfy(7,7): Error: unresolved identifier: Pind
-git-issue-970b.dfy(7,4): Error: The type of the first expression is not a failure type in :- statement
+git-issue-970b.dfy(7,4): Error: The type of the first expression to the right of ':-' could not be determined to be a failure type (got '?')
```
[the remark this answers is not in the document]
MikaelMayer: Yes, because the symbol is not defined so the type inference assigns a placeholder. I thought of rewriting the message specifically for this use case but I don't think it will be more useful than knowing the symbol does not exists.

Source/Dafny/Compilers/Compiler.cs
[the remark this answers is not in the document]
MikaelMayer: You gave me the idea of simplifying even further:

`bool hasRhs = s.Update is AssignSuchThatStmt || s.Update is AssignOrReturnStmt;`

so that I can remove `willBeAssignedAnyway`, since it's redundant with `hasRhs`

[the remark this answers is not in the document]
MikaelMayer: I completely overlooked the meaning of `hasRHS`. I can remove `willBeAssignedAnyway` by changing the definition of `hasRHS`:

`bool hasRhs = s.Update is AssignSuchThatStmt || s.Update is AssignOrReturnStmt`

Source/Dafny/Resolver.cs
MikaelMayer: Right. Reverted it so that it's not distracting.

RustanLeino (approved):

PR #1540, opened 2021-10-25
In-memory implementation of %diff, enable integration tests on windows

robin-aws:
Fixes #1313
Fixes #1538

PR #1685, opened 2022-01-05
fix: `lit` has no `%sed`

fabiomadge:
Fixes a regression in the raw `lit` tests introduced in #1540. Not tested on Windows.


RustanLeino (approved):
Thanks!

PR #1686, opened 2022-01-05
docs: add material on interacting with java code

fabiomadge:
We add two examples about passing strings between Dafny and Java code. In one example, the extern method obtains a Java string and returns it as a Dafny string. In the other, the extern method obtains a Java string and returns it as such. The Dafny code converts that Java string to a Dafny string.


keyboardDrummer:
Could you specify in the description of the PR what the intention of the code changes is? I couldn't get that from the title.

RustanLeino (changes requested):
This will be very useful to have in the test suite as an example for others, and also to make sure we keep this recipe up-to-date. I have two suggestions.

One suggestion is to move the tests from `Test/dafny0` to `Test/comp`.

The other suggestion is to add a comment at the top of two files. In `ExternDafnyString.dfy`, I suggest

> In this example, the extern method obtains a Java string and returns it as a Dafny string.

and in `ExternJavaString.dfy`, I suggest

> In this example, the extern method obtains a Java string and returns it as such. The Dafny code converts that Java string to a Dafny string.

fabiomadge:
Should we move `Test/dafny0/Extern.dfy` and `Test/dafny0/ExternCopyFromTrait.dfy` as well?

RustanLeino:
> Should we move `Test/dafny0/Extern.dfy` and `Test/dafny0/ExternCopyFromTrait.dfy` as well?

Yes, that would be great. Thanks!

RustanLeino (approved):

PR #1687, opened 2022-01-06
Add option to mimic the verification behavior of Dafny 3.3

keyboardDrummer:
### Changes

Add option to mimic the verification behavior of Dafny 3.3

### Testing

Manually checked that using this option on the CLI brings the verification behavior closer to Dafny 3.3


cpitclaudel:
Can we generalize this a bit, maybe `/compatibilityMode:3.3` and explain that the option is a best effort to mitigate changes between versions?  Right now it reads as if it's very specific to unstable proofs.

keyboardDrummer:
> Can we generalize this a bit, maybe `/compatibilityMode:3.3` and explain that the option is a best effort to mitigate changes between versions? Right now it reads as if it's very specific to unstable proofs.

I wasn't expecting this option to ever effect anything other than verification, hence the name. Verification in particular is something that is difficult to be fully backwards compatible for because of the nature of unstable proofs, but in other aspects I think we should either be backwards compatible (99% of the time) or otherwise I think we would ask customers to migrate.

cpitclaudel:
OK, your call :)

Source/Dafny/DafnyOptions.cs
```diff
@@ -966,6 +978,14 @@ not sound
     experimental and subject to change!
 {TestGenOptions.Help}
 
+/mimicVerificationOf:<Dafny version>
+    Let Dafny attempt to mimic the verification as it was in a previous version of Dafny. 
+    Useful during migration to a newer version of Dafny when a Dafny program has proofs, such as methods or lemmas, 
+    that are unstable in the sense that their verification may become slower or fail altogether 
+    after logically irrelevant changes are made in the verification input.
+
+    Accepted versions are: 3.3 (note that this turns of features that prevent classes of verification instability)
```
RustanLeino, line 987:
```suggestion
    Accepted versions are: 3.3 (note that this turns off features that prevent classes of verification instability)
```
```suggestion
    Accepted versions are: 3.3 (note that this turns of features that prevent classes of verification instability)
```

RustanLeino (changes requested):

RustanLeino (approved):
This seems like a fine option for now, but I don't expect we'll support the `3.3` option indefinitely.

Issue #697, opened 2020-07-02
Set comprehension should verify element-type subset constraints


RustanLeino:
A subset type

    type T = x: Base | P(x)

denotes the subset of values of type `Base` that satisfy `P`. The constraint `P(x)` is allowed to be a ghost expression and does need to be compilable. Now, here's the problem:

```
datatype Cell = Cell(x: int)
type EvenCell = c: Cell | c.x % 2 == 0 witness Cell(0)

method P(b: bool, s: set<Cell>) returns (t: set<EvenCell>)
{
  if b {
    // in the following, every element selected by the set comprehension is an EvenCell value
    t := set c: EvenCell | c in s && c.x % 2 == 0;
  } else {
    t := set c: EvenCell | c in s;  // error: set elements are not constraint to satisfy EvenCell
  }
}
```

Both set comprehensions have type `set<EvenCell>`. However, in order for the latter to construct a set containing only `EvenCell` values, the evenness check needs to be performed somewhere. It cannot be generated by the compiler, because the subset constraint is a ghost expression. Therefore, (either set comprehensions with an element type that is a subset type must be banned altogether, or) the constraint must be checked by the verifier.

I think there's a way to change the type checker so that the verifier will (with no change) generate the appropriate checks. It is to insist that the type of a set comprehension is always a type `set<A>` where `A` is a base type (that is, not a subset type). If that were done for the program above, the assignment to `t` will be an assignment from `set<Cell>` to `set<EvenCell>`, and that will cause the verifier to generate the appropriate verification condition.

RustanLeino:
A better suggestion (which is even easier to implement) is to insist that no bound variable of the set comprehension be of a subset type.

robin-aws:
Copying over additional examples from #698 (which I'm closing as a duplicate of this one).

Verifies but crashes at runtime, because of the same issue in the context of runtime qualification:

```dafny
type Small = x: int | 0 <= x < 100 && x != 3

function method F(x: int): int
  requires x != 3
{
  if x == 3 then 1/0 else 100
}

method Main() {
  var b := forall n: Small :: F(n) == 100;
  print b, "\n";
}
```

Causes the Java and Go compilers to produce malformed code:

```dafny
newtype byte = x | -128 <= x < 128

method ByteIndexedArray(s: array<int>) returns (rev: array<int>)
  requires s.Length < 128
{
  rev := new int[s.Length];
  forall i: byte | 0 <= i < s.Length as byte {
[...]

PR #1691, opened 2022-01-06
Fixed specifications misclassified as compiled expressions

MikaelMayer:
This fixes a regression introduced by #697 where it should be legit to use subset types in specifications expressions


Source/Dafny/Reporting.cs
```diff
@@ -199,7 +199,14 @@ public class ConsoleErrorReporter : BatchErrorReporter {
 
         ConsoleColor previousColor = Console.ForegroundColor;
         Console.ForegroundColor = ColorForLevel(level);
-        Console.WriteLine(ErrorToString(level, tok, msg));
+        var errorLine = ErrorToString(level, tok, msg);
+        while (tok is NestedToken nt) {
+          tok = nt.Inner;
+          msg = nt.Message ?? "[Related location]";
+          errorLine += " " + msg + " (" + (TokenToString(tok) + ").");
```
RustanLeino, line 206: There's an unnecessary pair of parentheses (or you may choose to use the `$" ... "` form to format the appended string).

Source/Dafny/Resolver.cs
```diff
@@ -18566,10 +18566,21 @@ public class SubsetConstraintGhostChecker : ProgramTraverser {
       return skip;
     }
 
+    private bool isFieldSpecification(string field, object parent) {
```
RustanLeino, line 18569: The style in C# is to capitalize all methods, including private ones.

Test/git-issues/git-issue-697h.dfy
```diff
@@ -0,0 +1,35 @@
...
+
+type substring = x: string | |x| < maxi
+
+// It's ok to have not compiled subset types in all specification expressions
+method test(s: seq<substring>) returns (r: int)
+  requires forall x: substring :: x in s ==> |x| < maxi
+  ensures forall x: substring :: x in s ==> |x| < maxi
+{
+  for j := 0 to 5
+    invariant forall x: substring :: x in s ==> |x| < maxi + 5- j
```
RustanLeino, line 14:
```suggestion
    invariant forall x: substring :: x in s ==> |x| < maxi + 5 - j
```

Test/git-issues/git-issue-697h.dfy
```diff
@@ -0,0 +1,35 @@
...
+  {
+    r := j;
+  }
+  assert forall x: substring :: x in s ==> |x| < maxi;
+  while (true)
```
RustanLeino, line 19:
```suggestion
  while true
```

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -93,12 +93,19 @@ public class DiagnosticErrorReporter : ErrorReporter {
       if (ErrorsOnly && level != ErrorLevel.Error) {
         return false;
       }
-
+      var relatedInformation = new List<DiagnosticRelatedInformation>();
+      if (tok is NestedToken nTok) {
```
keyboardDrummer, line 97: Please use full names such as `nestedToken`

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -6927,8 +6927,35 @@ public Statement(IToken tok, IToken endTok)
...
+        }
+
+        foreach (var e in NonSpecificationSubExpressions) {
+          yield return e;
+        }
+      }
+    }
+
+    /// <summary>
+    /// Returns the non-null expressions of this statement proper (that is, do not include the expressions of substatements).
```
keyboardDrummer, line 6945: What is a null Expression? I think we don't need to specify that this shouldn't returns nulls.

What are specifications in the context of a statement? Specification makes me think of requires and ensures clauses. Sorry for the noob question :-)

MikaelMayer: I just took this comment from the previous comment of "SubExpressions", and I understand it that "proper subexpressions" (that is, referenced by no other member of this expression) could sometimes be null, meaning they were not provided. We don't return them in this case, so that's why I think this comment makes sense although it should be straightforward.

Regarding the specifications in the context of a statement, let me show you the 7 cases in which it matters, for which I override `SpecificationSubExpressions` in `DafnyAST.cs`. all the following expressions E and their variants are specifications because they are not compiled.

First the four cases that can be compiled, which were regression problems, because they were previously assuming that their specifications were compiled for this PR (there was no way to distinguish Guard from E using only SubExpressions for the while loop for example)
```
for j := 0 to 5 invariant E(j) {...}
while(Guard) invariant E(j) {...}
forall j | 0 <= j < 5 ensures E {...}
modify E {...}
```

Second, the next three cases were modified accordingly to be consistent, but the sentence is marked as ghost anyway.
```
assert E1;
assume E2;
calc { E1; E2; ....}
```

Note that `expect E` is not in this category because in this case E would be executed at run-time (i.e. for tests)

RustanLeino (dismissed):

MikaelMayer:
I just splitted the content of this PR so that the regression bug is fixed here, while the reporting enhancement is fixed in #1695

RustanLeino (dismissed):

MikaelMayer:
@RustanLeino  I had to remove code that was actually used in the other PR, but basically it's the same code. Can you approve it again after a reviewing it?

RustanLeino (approved):

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -6927,8 +6927,35 @@ public Statement(IToken tok, IToken endTok)
 
     /// <summary>
     /// Returns the non-null expressions of this statement proper (that is, do not include the expressions of substatements).
+    /// Filters all sub expressions that are not part of specifications
     /// </summary>
-    public virtual IEnumerable<Expression> SubExpressions {
+    public IEnumerable<Expression> SubExpressions {
```
keyboardDrummer, line 6932: I would expect us to have all error messages of the form "...is not compilable..." to occur after a de-ghosting phase in the compiler pipeline, after which all Expressions should be non-specification/ghost expressions, so I'm a bit surprised we need the change this PR makes, but I guess without the current architecture we need it.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -6927,8 +6927,35 @@ public Statement(IToken tok, IToken endTok)
...
+          yield return e;
+        }
+      }
+    }
+
+    /// <summary>
+    /// Returns the non-null expressions of this statement proper (that is, do not include the expressions of substatements).
+    /// Filters only expressions that are always part of specifications
+    /// </summary>
+    public virtual IEnumerable<Expression> SpecificationSubExpressions {
```
keyboardDrummer, line 6948: Did you consider using the term `Ghost` instead of `Specification` here? What are the pros/cons?

keyboardDrummer (approved):

Source/Dafny/Resolver.cs
```diff
@@ -18571,10 +18571,21 @@ public class SubsetConstraintGhostChecker : ProgramTraverser {
...
+        (parent is Function && (field is "Req.E" or "Reads.E" or "Ens.E" or "Decreases.Expressions")) ||
+        (parent is Method && (field is "Req.E" or "Mod.E" or "Ens.E" or "Decreases.Expressions"))
+      );
+    }
+
     public override bool Traverse(Expression expr, [CanBeNull] string field, [CanBeNull] object parent) {
       if (expr == null) {
         return false;
       }
+      if (IsFieldSpecification(field, parent)) {
```
keyboardDrummer, line 18586: I would expect us to have all error messages of the form "...is not compilable..." to be created after a de-ghosting phase in the compiler pipeline, after which all Expressions should be non-ghost, so I'm a bit surprised we need the change this PR makes but I guess we need it with the current architecture.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: Indeed, I could have detailed my thinking to double-proof it with the rest of the team.

Pros of using the wording "ghost"
* Seems consistent with the rest, "ghost" meaning "non-compiled"
* Shorter than "specification"

Cons of using the wording "ghost"
* I wanted a clear disjunction between the two types of expressions, so I had the choice between "Ghost/Compiled" or "Specification/NonSpecification". The problem with "Compiled" (or "NonGhost"?) is that such expressions can still be ghost if they are in a ghost context (like a lemma or a ghost method). Ghost seems like a final property, where the code appears in the final compiler or not. Here I want the reasoning about the expressions to be local.
* We currently mark things as "Ghost" in the IDE only when they are ambiguously ghost (I would have preferred to visualize everything ghost the same way, but that's the current consensus). Here, we are dealing with expressions that are unambiguously "ghost", so they are specifications, like pre and post conditions.

keyboardDrummer: Thanks, for explaining. Makes sense :-)

Source/Dafny/Resolver.cs
MikaelMayer: That seems like a reasonable expectation. I would go one step further: Everything has to happen before compilation, and there should be no such "is not compilable" error thrown in practice by the compiler.
Indeed, ideally subset types should be verified and then removed at compile time, and I'm currently working on that.

Issue #1678, opened 2022-01-03
IDE shows "Verified", but input contains a resolution error


cpitclaudel:
The following example confuses the IDE:

```dafny
function f(x: nat) : nat

method test() {
  assert forall x: nat {:trigger false} :: f(x) == f(x);
}
```

It says verification succeeded, but the CLI says this instead:

```
"c:/Apps/dafny-3.3/Dafny.exe" "/compile:0" "c:/Users/cpitcla/wip/difftests/inconsistent.dfy"
c:/Users/cpitcla/wip/difftests/inconsistent.dfy(4,9): Error: trigger must mention all quantified variables, but does not mention: x#1
c:/Users/cpitcla/wip/difftests/inconsistent.dfy(4,9): Error: trigger must mention all quantified variables, but does not mention: x#1
2 name resolution errors detected in C:\Users\cpitcla\AppData\Local\Temp\inconsistent__module.bpl

*** Encountered internal translation error - re-running Boogie to get better debug information

C:\Users\cpitcla\AppData\Local\Temp\inconsistent__module.bpl(2871,6): Error: trigger must mention all quantified variables, but does not mention: x#1
C:\Users\cpitcla\AppData\Local\Temp\inconsistent__module.bpl(2875,6): Error: trigger must mention all quantified variables, but does not mention: x#1
2 name resolution errors detected in C:\Users\cpitcla\AppData\Local\Temp\inconsistent__module.bpl
```

PR #1683, opened 2022-01-04
NullPointerException fixed in Parser

MikaelMayer:
This PR fixes one problem of #1682
* [No more null pointer exception](https://github.com/dafny-lang/dafny/pull/1683/files#diff-2a1129310e5714b2628a1412c696feac09752d71a6498154491ec02b4ca53b4fR2628) as found by the issue

I added a tests to ensure does not happen again for this particular case

#1694 fixes the two other problems

Issue #1682, opened 2022-01-03
Unhandled NullReferenceException in ExprRhs..ctor


dijkstracula:
Dear Dafny team,

While typing the following silly and incomplete Dafny program:

```
method IsUnique(str: array<char>) returns (ret: bool)
    ensures forall i, j :: 0 <= i < j < str.Length ==> ret ==> str[i] != str[j]
{
    ret := multiset
}
```

 immediately upon typing the `multiset` token the VSCode language plugin crashes with the following stacktrace:

```
[Error - 4:07:07 PM] Request textDocument/completion failed.
  Message: Internal Error - System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Dafny.ExprRhs..ctor(Expression expr, Attributes attrs) in /Users/ntaylor/code/dafny/Source/Dafny/AST/DafnyAst.cs:line 7262
   at Microsoft.Dafny.Parser.Rhs(AssignmentRhs& r) in /Users/ntaylor/code/dafny/Source/Dafny/Parser.cs:line 4049
   at Microsoft.Dafny.Parser.UpdateStmt(Statement& s) in /Users/ntaylor/code/dafny/Source/Dafny/Parser.cs:line 3232
   at Microsoft.Dafny.Parser.OneStmt(Statement& s) in /Users/ntaylor/code/dafny/Source/Dafny/Parser.cs:line 3121
   at Microsoft.Dafny.Parser.Stmt(List`1 ss) in /Users/ntaylor/code/dafny/Source/Dafny/Parser.cs:line 3103
   at Microsoft.Dafny.Parser.BlockStmt(BlockStmt& block, IToken& bodyStart, IToken& bodyEnd) in /Users/ntaylor/code/dafny/Source/Dafny/Parser.cs:line 2682
   at Microsoft.Dafny.Parser.MethodDecl(DeclModifierData dmod, Boolean allowConstructor, Method& m) in /Users/ntaylor/code/dafny/Source/Dafny/Parser.cs:line 2148
   at Microsoft.Dafny.Parser.ClassMemberDecl(DeclModifierData dmod, List`1 mm, Boolean allowConstructors, Boolean isValueType, Boolean moduleLevelDecl) in /Users/ntaylor/code/dafny/Source/Dafny/Parser.cs:line 1415
[...]

PR #1694, opened 2022-01-07
Parser exception recovery for Language Server

MikaelMayer:
This PR fixes two problems of #1682

- A parser throwing an exception no longer crashes the Dafny Language Server.
- It no longer displays "Verification succeeded" because the Dafny Language Server crashed.

I added a test to simulate a successful parsing exception recovery.

#1683 fixes the parsing error in itself.


cpitclaudel:
I suspect that this may fix https://github.com/dafny-lang/dafny/issues/1678 as well

MikaelMayer:
How so?

Source/DafnyLanguageServer.Test/Various/ParserExceptionTest.cs
```diff
@@ -0,0 +1,93 @@
...
+
+      public override int Count(ErrorLevel level) {
+        throw new NotImplementedException();
+      }
+    }
+
+    [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+    public void DocumentWithParserExceptionDisplaysIt() {
+      var source = "function t() { / }";
+      var documentItem = CreateTestDocument(source, TestFilePath);
```
camrein, line 86: I've got the feeling that the `CreateTestDocument` method is the only reason you're basing your test on `DafnyLanguageServerTestBase`, adding lots of unnecessary complexity and boilerplate to this unit test. Therefore, I suggest either creating a utility class for this method or simply creating an inline implementation for this particular test.
I've omitted to review the parts I feel unnecessary because of this assumption, if I'm wrong I'll review them accordingly.

Source/DafnyLanguageServer/Language/DafnyLangParser.cs
```diff
@@ -72,9 +73,22 @@ public sealed class DafnyLangParser : IDafnyParser, IDisposable {
...
+
+        return program;
+      } catch (Exception e) {
+        logger.LogDebug(e, "encountered an exception while parsing {DocumentUri}", document.Uri);
+        var internalErrorDummyToken = new Token {
+          filename = document.GetFilePath(),
+          line = 1,
+          col = 1,
+          pos = 0,
+          val = " "
```
camrein, line 89: What's the reason for this space? Wouldn't an empty string, i.e. `string.Empty`, work?

Source/DafnyLanguageServer.Test/Various/ParserExceptionTest.cs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public CustomLogger customLogger = new();
+
+    public Task SetUp(IDictionary<string, string> newConfiguration) {
+      this.configuration = newConfiguration;
+      return Task.Run(() => parser = DafnyLangParser.Create(customLogger));
+    }
+
+    class ParserExceptionSimulatingErrorReporter : ConsoleErrorReporter {
```
camrein, line 61: Wouldn't `ErrorReporter` be sufficient as a base class here?

camrein (changes requested):
I've reviewed your changes. I've got some inputs to the test class. Moreover, I suggest moving the test class to the unit test namespace since it doesn't interact with the language server through the LSP API.

Source/DafnyLanguageServer.Test/Various/ParserExceptionTest.cs
MikaelMayer: Your assumption is correct. I'm now removing the dependency to `DafnyLanguageServerTestBase` so that I can remove the boilerplate with the setup.

Source/DafnyLanguageServer/Language/DafnyLangParser.cs
MikaelMayer: I previously thought that an empty val would prevent the message from being displayed, but that happened only if the line was 0. With the line 1 and the string empty, it works (and it still underlines the first character of the file in red), so I'm pushing this that @atomb also asked for.

Source/DafnyLanguageServer.Test/Unit/ParserExceptionTest.cs
```diff
@@ -0,0 +1,99 @@
...
+      lastDebugLogger = new LastDebugLogger();
+      parser = DafnyLangParser.Create(lastDebugLogger);
+    }
+
+    [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+    public void DocumentWithParserExceptionDisplaysIt() {
+      var source = "function t() { / }";
+      var documentItem = CreateTestDocument(source, TestFilePath);
+      var errorReporter = new ParserExceptionSimulatingErrorReporter();
+      parser.Parse(documentItem, errorReporter, new CancellationToken());
```
camrein, line 33: You can use `default` in place of `new CancellationToken()` since it's a `struct` type.

Source/DafnyLanguageServer.Test/Unit/ParserExceptionTest.cs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// <summary>
+    /// An error reporter that throws an exception on the first reported error, to simulate a parser exception.
+    /// </summary>
+    class ParserExceptionSimulatingErrorReporter : ErrorReporter {
+      private int numberOfErrors;
+      public string LastMessage = "";
+      public override bool Message(MessageSource source, ErrorLevel level, IToken tok, string msg) {
+        numberOfErrors++;
```
camrein, line 45: This statement should probably be inside the if-statement below.
Just out of curiosity: Is it necessary to limit throwing the exception only upon the first error? Wouldn't it be easier to ensure that the input Dafny code only yields one error? Or is this some sort of fail-safety to not break this unit test in the future if the language changes?

camrein (changes requested):
Looks way better now. I've just got some minor input now.

Source/DafnyLanguageServer.Test/Unit/ParserExceptionTest.cs
MikaelMayer: Yes it is necessary to limit to throwing the exception only upon the first error (the simulated parsing exception), because there is now a try/catch that will emit another error afterwards. (`errorReporter.Error`)

Source/DafnyLanguageServer.Test/Unit/ParserExceptionTest.cs
```diff
@@ -0,0 +1,99 @@
...
+        throw new NotImplementedException();
+      }
+    }
+
+    // Helpers and definitions
+
+    /// <summary>
+    /// Retains the last debug logged message
+    /// </summary>
+    public class LastDebugLogger : ILogger<DafnyLangParser> {
```
camrein, line 68: `private class`

Source/DafnyLanguageServer.Test/Unit/ParserExceptionTest.cs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// <summary>
+    /// An error reporter that throws an exception on the first reported error, to simulate a parser exception.
+    /// </summary>
+    class ParserExceptionSimulatingErrorReporter : ErrorReporter {
+      private int numberOfErrors;
+      public string LastMessage = "";
+      public override bool Message(MessageSource source, ErrorLevel level, IToken tok, string msg) {
+        if (level == ErrorLevel.Error) {
```
camrein, line 45: You may want to invert this condition to reduce the nesting.

keyboardDrummer (approved):

camrein (approved):

PR #1695, opened 2022-01-07
Related locations available in Resolver and prints to Console and LSP Diagnostics

MikaelMayer:
This PR fixes the error reporting system so that resolution can provide related errors.

VSCode IDE: before displayed these informations as not related to the original error:
image

VSCode IDE after now correctly nests this information under the original error.
image

[Command line before/after](https://github.com/dafny-lang/dafny/pull/1695/files#diff-f84e2f59e62ce60701eaca1c9c94144f4fa9b7dd2b8d4b3e367d7281de288fc5R1) has a similar behavior that will be less disturbing.


RustanLeino (approved):

Source/Dafny/Resolver.cs
```diff
@@ -18599,15 +18599,19 @@ public class SubsetConstraintGhostChecker : ProgramTraverser {
             }
 
             if (!constraintIsCompilable) {
```
keyboardDrummer, line 18601: Not for this PR, but it would be nice if we can split `Resolver.cs` into multiple smaller segments, around the 3k line range, if only to let the syntax highlighting in GH work. Doing this using partial classes would be fine.

Test/allocated1/dafny0/DiscoverBounds.dfy.expect
```diff
@@ -1,8 +1,8 @@
-DiscoverBounds.dfy(32,7): Warning: /!\ No terms found to trigger on.
-DiscoverBounds.dfy(33,7): Warning: /!\ No terms found to trigger on.
-DiscoverBounds.dfy(34,7): Warning: /!\ No terms found to trigger on.
-DiscoverBounds.dfy(35,7): Warning: /!\ No terms found to trigger on.
-DiscoverBounds.dfy(36,7): Warning: /!\ No terms found to trigger on.
+DiscoverBounds.dfy(32,7): Warning: /!\ No terms found to trigger on. [Related location] (DiscoverBounds.dfy(32,33)).
```
keyboardDrummer, line 1: Is the new information in the error message useful? Is it nice to get a related location without having a message explaining how the location is related?

I look at DiscoverBounds.dfy and it wasn't obvious to me how the related location is related. Could you tell me how it is?

Source/Dafny/Reporting.cs
```diff
@@ -235,4 +247,19 @@ public class ErrorReporterWrapper : BatchErrorReporter {
       return WrappedReporter.Message(source, level, tok, msgPrefix + msg);
     }
   }
+
+  public class SimpleErrorCorrector : ErrorReporter {
+    public List<string> CollectedMessages = new();
```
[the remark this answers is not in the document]
MikaelMayer: I had considered something like `List<Tuple<string,IToken>>` instead of two lists, but not dictionaries, since they don't represent key/values. For this class, I only need to recover the last message and token, so I could simplify it further if you wish.

Source/Dafny/Reporting.cs
```diff
@@ -235,4 +247,19 @@ public class ErrorReporterWrapper : BatchErrorReporter {
       return WrappedReporter.Message(source, level, tok, msgPrefix + msg);
     }
   }
+
+  public class SimpleErrorCorrector : ErrorReporter {
```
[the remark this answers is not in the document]
MikaelMayer: I think I'll move it closer to where it is used. Thanks for the suggestion.

Source/Dafny/Resolver.cs
MikaelMayer: Agreed. We could split it by phases.

Test/allocated1/dafny0/DiscoverBounds.dfy.expect
MikaelMayer: In general, It is always better to have a related location without having a message explaining how the location is related than having no related location at all, if the related location is important.
The information was already there, stored in nested tokens, this PR only reveals it. But I agree with your comment, it would be better if it was a nicer message here.

In `DiscoverBounds.dfy` the related location is actually the position of the trigger that was tried.

I went ahead and added information that completely replaces `[Related location]` in context, explaining how the related location is actually related. For example in the test `suppressing-warnings-behaves-properly.dfy`, before, it was saying

```
suppressing-warnings-behaves-properly.dfy(10,9): Warning: /!\ No terms found to trigger on.  [Related location] (suppressing-warnings-behaves-properly.dfy(10,33)).
suppressing-warnings-behaves-properly.dfy(11,9): Info: (Suppressed warning) No terms found to trigger on. [Related location] (suppressing-warnings-behaves-properly.dfy(11,43)).
suppressing-warnings-behaves-properly.dfy(12,9): Info: Not generating triggers for "n >= 0 || n < 0". Note that {:autotriggers false} can cause instabilities. Consider using {:nowarn}, {:matchingloop} (not great either), or a manual trigger instead. [Related location] (suppressing-warnings-behaves-properly.dfy(12,55)).
suppressing-warnings-behaves-properly.dfy(14,9): Info: Selected triggers: {f(n)} [Related location] (suppressing-warnings-behaves-properly.dfy(14,46)).
```

Now it says:


```
suppressing-warnings-behaves-properly.dfy(10,9): Warning: /!\ No terms found to trigger on. Potential trigger not suitable: (suppressing-warnings-behaves-properly.dfy(10,33)).
suppressing-warnings-behaves-properly.dfy(11,9): Info: (Suppressed warning) No terms found to trigger on. Potential trigger not suitable: (suppressing-warnings-behaves-properly.dfy(11,43)).
suppressing-warnings-behaves-properly.dfy(12,9): Info: Not generating triggers for "n >= 0 || n < 0". Note that {:autotriggers false} can cause instabilities. Consider using {:nowarn}, {:matchingloop} (not great either), or a manual trigger instead. Trigger position: (suppressing-warnings-behaves-properly.dfy(12,55)).
suppressing-warnings-behaves-properly.dfy(14,9): Info: Selected triggers: {f(n)} Trigger position: (suppressing-warnings-behaves-properly.dfy(14,46)).
```

I think it's better too! Thanks for suggesting!

cpitclaudel:
Normally related positions are reported as a separate Info-level entry; why not do that here as well?

MikaelMayer:
> Normally related positions are reported as a separate Info-level entry; why not do that here as well?

Where are "related positions [normally] reported" ? As far as I know, they were simply ignored until now, except when they came from Boogie.
Are you talking about the command line or the LSP?

cpitclaudel:
> Where are "related positions [normally] reported" ? 

When you have a verification error

> Are you talking about the command line or the LSP?

Both I think?

MikaelMayer:
> > Where are "related positions [normally] reported" ?
> 
> When you have a verification error
> 
> > Are you talking about the command line or the LSP?
> 
> Both I think?

Oh so you meant "usually reported". I see, this seems like an inconsistency. The printing of these related errors are defined in Boogie's `ConsolePrinter`, itself used in `DafnyDriver`. They don't have a message except for "Related location" This message along helps to distinguish between an stand-alone error and a dependent error.

The difference with Dafny is that these nested token happen not only for errors (like Boogie does), but also for warnings and information logs. I have the feeling that they do not need an extra line, it's clearer to have everything in one line, since it has an error message and is part of the error message (there in practice at most 1 other related error, contrary to Boogie). For example, the selected quantifiers don't need an extra separate information message.

Source/Dafny/Resolver.cs
RustanLeino: Yes, a split of `Resolver.cs` into several files, especially along the lines of phases (and sub-parts thereof), would be great!

RustanLeino (dismissed):
I'm okay with this, if the other questions/remarks have been resolved.

Test/allocated1/dafny0/DiscoverBounds.dfy.expect
```diff
@@ -1,8 +1,8 @@
-DiscoverBounds.dfy(32,7): Warning: /!\ No terms found to trigger on. [Related location] (DiscoverBounds.dfy(32,33)).
-DiscoverBounds.dfy(33,7): Warning: /!\ No terms found to trigger on. [Related location] (DiscoverBounds.dfy(33,33)).
-DiscoverBounds.dfy(34,7): Warning: /!\ No terms found to trigger on. [Related location] (DiscoverBounds.dfy(34,36)).
-DiscoverBounds.dfy(35,7): Warning: /!\ No terms found to trigger on. [Related location] (DiscoverBounds.dfy(35,39)).
-DiscoverBounds.dfy(36,7): Warning: /!\ No terms found to trigger on. [Related location] (DiscoverBounds.dfy(36,42)).
+DiscoverBounds.dfy(32,7): Warning: /!\ No terms found to trigger on. Potential trigger not suitable: (DiscoverBounds.dfy(32,33)).
```
keyboardDrummer, line 1: Can we drop the parenthesis around the related location?

Source/Dafny/Triggers/QuantifiersCollection.cs
```diff
@@ -285,10 +285,18 @@ internal class QuantifierGroup {
       var msg = new StringBuilder();
       var indent = addHeader ? "  " : "";
       bool suppressWarnings = Attributes.Contains(q.quantifier.Attributes, "nowarn");
+      var reportingToken = q.quantifier.tok;
+
+      void MaybeUpdateReportingToken(string msg) {
```
keyboardDrummer, line 290: Consider renaming to `SetRelatedLocationMessage` or `SetNestedTokenMessage`

keyboardDrummer:
> > Are you talking about the command line or the LSP?
> 
> Both I think?

Since recently, the Dafny IDE will report related locations for verification errors using VSCode's "Relation location" UI, which means it's not a separate info rule.

keyboardDrummer (dismissed):

keyboardDrummer (approved):

cpitclaudel:
> The difference with Dafny is that these nested token happen not only for errors (like Boogie does), but also for warnings and information logs. I have the feeling that they do not need an extra line, it's clearer to have everything in one line, since it has an error message and is part of the error message (there in practice at most 1 other related error, contrary to Boogie).

But this breaks existing code that parses Dafny's output to display related locations, like Flycheck.

> For example, the selected quantifiers don't need an extra separate information message.

But we do have a separate message for them at the moment.

MikaelMayer:
> But this breaks existing code that parses Dafny's output to display related locations, like Flycheck.
> But we do have a separate message for them at the moment.

Thanks for mentioning it. I wasn't aware of Flycheck. Where is it located?
Also, Boogie's nested tokens are still displayed one separate lines. This is only modifying the ConsoleErrorReporter that is used by the resolver. So it shouldn't break anything.

cpitclaudel:
It's fine, I'm not sure we need to bother.  Flycheck is used as part of the Emacs mode for Dafny.

It parses Dafny's command line output.  So does the differential checker.  With this PR we now have two ways to report related locations: as separate messages and as inline messages within another message

Source/Dafny/Triggers/QuantifiersCollection.cs
```diff
@@ -285,10 +285,18 @@ internal class QuantifierGroup {
...
+      void SetRelatedLocationMessage(string msg) {
+        if (reportingToken is NestedToken nestedToken) {
+          reportingToken = new NestedToken(nestedToken.Outer, nestedToken.Inner, msg);
+        }
+      }
 
       if (!TriggerUtils.NeedsAutoTriggers(q.quantifier)) { // NOTE: split and autotriggers attributes are passed down to Boogie
         var extraMsg = TriggerUtils.WantsAutoTriggers(q.quantifier) ? "" : " Note that {:autotriggers false} can cause instabilities. Consider using {:nowarn}, {:matchingloop} (not great either), or a manual trigger instead.";
         msg.AppendFormat("Not generating triggers for \"{0}\".{1}", Printer.ExprToString(q.quantifier.Term), extraMsg).AppendLine();
+        SetRelatedLocationMessage("Trigger position:");
```
cpitclaudel, line 299: Could we have a newline before this?  Otherwise currently the message looks like this:

```
constructors-cause-matching-loops.dfy(10,9): Info: Selected triggers: {Succ(s)}
 Rejected triggers:
   {f(s)} (may loop with "f(Succ(s))")
   {f(Succ(s))} (more specific than {Succ(s)}) Trigger position: constructors-cause-matching-loops.dfy(10,40)

```

This is confusing: it's not clear *which* trigger the location points to.

Source/Dafny/Triggers/QuantifiersCollection.cs
```diff
@@ -299,32 +307,42 @@ internal class QuantifierGroup {
...
 #if QUANTIFIER_WARNINGS
-        var WARN_TAG = DafnyOptions.O.UnicodeOutput ? "⚠ " : "/!\\ ";
+        var WARN_TAG = DafnyOptions.O.UnicodeOutput ? "⚠ " : @"/!\ ";
         var WARN_TAG_OVERRIDE = suppressWarnings ? "(Suppressed warning) " : WARN_TAG;
         var WARN_LEVEL = suppressWarnings ? ErrorLevel.Info : ErrorLevel.Warning;
         var WARN = indent + WARN_TAG_OVERRIDE;
         if (!q.CandidateTerms.Any()) {
           errorLevel = WARN_LEVEL;
           msg.Append(WARN).AppendLine("No terms found to trigger on.");
+          SetRelatedLocationMessage("Potential trigger not suitable:");
```
cpitclaudel, line 326: I don't understand what this message means and what it points to

Source/Dafny/Triggers/QuantifiersCollection.cs
```diff
@@ -299,32 +307,42 @@ internal class QuantifierGroup {
...
         } else if (!q.Candidates.Any()) {
           errorLevel = WARN_LEVEL;
           msg.Append(WARN).AppendLine("No trigger covering all quantified variables found.");
+          SetRelatedLocationMessage("Potential trigger not suitable:");
```
cpitclaudel, line 330: Same here: what does this mean?  There are many trigger candidates within a term, so why highlight one?

Source/Dafny/Triggers/QuantifiersCollection.cs
```diff
@@ -299,32 +307,42 @@ internal class QuantifierGroup {
...
         } else if (!q.CouldSuppressLoops && !q.AllowsLoops) {
           errorLevel = WARN_LEVEL;
           msg.Append(WARN).AppendLine("Suppressing loops would leave this expression without triggers.");
+          SetRelatedLocationMessage("Trigger position:");
         } else if (suppressWarnings) {
           errorLevel = ErrorLevel.Warning;
           msg.Append(indent).Append(WARN_TAG).AppendLine("There is no warning here to suppress.");
+          SetRelatedLocationMessage("Trigger position:");
```
cpitclaudel, line 338: Why do you add a location message to a warning about `nowarn`?

Source/Dafny/Triggers/QuantifiersCollection.cs
```diff
@@ -299,32 +307,42 @@ internal class QuantifierGroup {
         }
 
         AddTriggersToMessage("Selected triggers:", q.Candidates, msg, indent);
+        if (q.Candidates.Count > 0) {
+          SetRelatedLocationMessage("Trigger position:");
```
cpitclaudel, line 311: If there are multiple trigger it doesn't make sense to point to one of them in particular.

Source/Dafny/Triggers/QuantifiersCollection.cs
```diff
@@ -299,32 +307,42 @@ internal class QuantifierGroup {
...
+        }
         AddTriggersToMessage("Rejected triggers:", q.RejectedCandidates, msg, indent, true);
+        if (q.RejectedCandidates.Count > 0 && q.Candidates.Count == 0) {
+          SetRelatedLocationMessage("Rejected trigger position:");
```
cpitclaudel, line 315: There's more than one rejected trigger typically, so adding a link to just one of them is a bit weird

cpitclaudel:
I hadn't seen the part about quantifier trigger messages; I think they are confusing.

Source/Dafny/Triggers/QuantifiersCollection.cs
MikaelMayer: I changed:

` Trigger position`

to newline  +
` Selected trigger position`

Source/Dafny/Triggers/QuantifiersCollection.cs
MikaelMayer: So in the token, there was a nested token that was recorded to contain a position of a token, but it turned out that it was not chosen as a selected token. Since nested tokens are always displayed now, I wanted to clarify the message and instead of "Related location", I chose the message above. Do you have any other suggestion?

Source/Dafny/Triggers/QuantifiersCollection.cs
MikaelMayer: Because this trigger candidate is the only one that is being recorder in `q.tok` as a nested token. Otherwise, there would be no good reason to not include all of them, except for the legibility of the message. The alternative would be to get rid of this message. What do you think?

Source/Dafny/Triggers/QuantifiersCollection.cs
MikaelMayer: I am not adding it, it was already there, I'm just replacing the "Related location" message with something more explicit. The alternative would be to just ignore it. What do you think?

Source/Dafny/Triggers/QuantifiersCollection.cs
cpitclaudel: Yes, I'd suggest removing the nested token :)

Source/Dafny/Triggers/QuantifiersCollection.cs
cpitclaudel: Let's get rid of the message and of the nested token

Source/Dafny/Triggers/QuantifiersCollection.cs
cpitclaudel: Sounds good.

cpitclaudel: Except wait, this line is for the case where no triggers are being generated because the user turned that feature off, so that data looks wrong too.

Source/Dafny/Triggers/QuantifiersCollection.cs
MikaelMayer: After reading carefully your comment, I investigated what are exactly these nested tokens, because although they coincide with triggers, they are not covering all triggers since there is only one.
It turns out, the nested token does not describe triggers, it describes only the considered sub-expression after splitting it.

For example, if we had `forall a in A :: (R(a) ==> P(a)) && !(R(a) ==> Q(a))`, then it will first split this expression into three comprehensions `forall a in A :: R(a) ==> P(a))`, `forall a in A :: R(a)` and `forall a in A :: !Q(a)` (and this "not" is reusing the same token as its argument), and then it will look for trigger terms in each sub-expressions. However, triggers _candidates_ are a set of trigger terms, not just terms, so the position this "Related position" gives is actually the positions of the sub-expression that cannot be split further (e.g.  in our case `R(a) ==> P(a)`, `R(a)` and `Q(a)`

Hence, instead of calling them "trigger position", it would be more appropriate to call them "While considering sub-expression at position" if this sub-expression is not the body of the original entire forall expression.

Source/Dafny/Triggers/QuantifiersCollection.cs
cpitclaudel: This should probably be in an else branch:

```
else {
  if (suppressWarnings) { … }
  SetRelatedLocation…
}

File: Source/Dafny/Triggers/QuantifiersCollection.cs (hunk 11 above)
Comment (cpitclaudel) [reply Source/Dafny/Triggers/QuantifiersCollection.cs:311]: Ah, that makes a lot more sense!
Don't we already report that information?  For example this:

```
predicate P(x: int)
predicate Q(x: int)

function f() : int
  requires forall x: int :: P(x) && (Q(x) ==> P(x+1))
```

…gives me the following message:

```
For expression "P(x)":
  Selected triggers: {P(x)}
For expression "Q(x) ==> P(_t#0)":
   Selected triggers: {P(_t#0), Q(x)}
```

Comment (cpitclaudel) [reply Source/Dafny/Triggers/QuantifiersCollection.cs:311]: Annoyingly that message seems to have regressed, since originally it said `P(x)` instead of `P(_t#0)`

Comment (MikaelMayer) [reply Source/Dafny/Triggers/QuantifiersCollection.cs:311]: One thing to consider is that, once the message is built, it is passed to "reporter.Message", and we don't know which implementation it is (console or IDE). In the IDE, and later also for Boogie, we do want this information positionally if it is useful (i.e. here is a header, which means there are more than one splitted sub-expression)
Indeed, the NestedToken is used by boogie to report "Related location" which point to the sub-expression that is being considered, using a single click, so we definitely want to keep the NestedToken with its message.

Concerning the command-line, I think it's acceptable to have more information about the position of sub-expressions, especially since those sub-expressions might not be exactly the same as the original expressions (i.e. the "not" example")

```
For expression "P(x)":
  Selected triggers: {P(x)} in subexpression at document.dfy(5, 29)
For expression "Q(x) ==> P(_t#0)":
   Selected triggers: {P(_t#0), Q(x)} in subexpression at document.dfy(5, 42)
```

What do you think of the above?

I can add your test and verify that it still works (and let me see if I can fix this hideous generated variable, which does make sense since it's not x but x+1)

MikaelMayer: I also hesitate between

* `in sub-expression at`
* `at considered expression position:`
* `inside the expression at`
* `after considering the expression at`

Which one would you think is better?

PR #1696, opened 2022-01-07
Update use-local-boogie.sh to .NET 6.0

atomb:
Also make it use Release configuration builds. Given that it copies binaries already, it seems that it makes sense to get optimized binaries. If you want to debug Boogie along with Dafny, the customBoogie.patch approach seems more appropriate, as it makes the source more readily available in the IDE.


keyboardDrummer (approved):

Scripts/use-local-boogie.sh
```diff
@@ -2,5 +2,5 @@ BOOGIE_PATH="Source/boogie/"
 allDlls=("Boogie.AbsInt.dll" "Boogie.BaseTypes.dll" "Boogie.CodeContractsExtender.dll" "Boogie.Concurrency.dll" "Boogie.Core.dll" "Boogie.ExecutionEngine.dll" "Boogie.Graph.dll" "Boogie.Houdini.dll" "Boogie.Model.dll" "Boogie.Predication.dll" "Boogie.Provers.SMTLib.dll" "Boogie.VCExpr.dll" "Boogie.VCGeneration.dll" "BoogieDriver.dll" "System.Configuration.ConfigurationManager.dll" "System.Runtime.Caching.dll" "System.Security.Cryptography.ProtectedData.dll" "System.Security.Permissions.dll")
```
keyboardDrummer, line 2: Is there documentation that refers to the `use-local-boogie.sh` file? Is the use-case to use a version of Boogie that hasn't been released to NuGet yet?

keyboardDrummer: > which can be useful if, for example, you want to build Dafny in Debug configuration but Boogie in Release configuration.

Why would you want that?

atomb: Yep, I agree. It's occasionally useful but usually not necessary. The wiki section linked above now describes how to do it when needed, though.

Issue #1263, opened 2021-06-30
Type of - inferred to be nat


RustanLeino:
The type of an integer-subtraction expression should never be inferred to be a subset type, but should be the base type thereof. This is not done properly, which causes the verifier to give an error in the following program:

``` dafny
method M(N: nat) {
  var i := N;
  while i != 0
    invariant i - 1 == i - 1 // error reported: result of operation may not be a 'nat'
  {
    i := i - 1;
  }
}
```

While it is true that the result of `i - 1` above may not be a `nat`, this property should not have been checked in the first place.

Workaround: change `i - 1` to `i as int - 1`

DavePearce:
> The type of an integer-subtraction expression should never be inferred to be a subset type

But, if you don't do that then the subsequent statement `i := i - 1` won't type check?  In Whiley I resolve this by forward propagation type information.  Then we know that `i - 1` in `i := i - 1` is expected to be a `nat` whilst `i-1` in `i - 1 == i - 1` doesn't have to be.

RustanLeino:
Here's another example, using the more innocently known operator `+`:

``` dafny
type EvenUpTo100 = x | x % 2 == 0 && x <= 100

method P(N: EvenUpTo100) {
  var i := N;
  while i != 100
    invariant i + 1 == i + 1 // error reported: result of operation may not be an 'EvenUpTo100'
  {
    i := i + 2;
  }
}
```

It seems we'd like to specify the type of the numeric `+` and `-` (and similar operators) as

    A * B --> C

where `A`, `B`, and `C` are any numeric types where `C` is a supertype of both `A` and `B` and `C` is (this is the new constraint, which the type inference engine doesn't currently know about) a base type.

RustanLeino:
@DavePearce If `i` has type `nat` and `i - 1` is inferred to have type `int`, then Dafny would still allow the assignment `i := i - 1`, provided the verifier can prove that `i - 1` is a `nat` at the time of assignment.

DavePearce:
Hey @RustanLeino,

So basically you're doing a form of type inference where constraints are checked by the verifier?  That sounds similar to Whiley then!

RustanLeino:
Technically, the desired types are inferred by a more standard type-inference engine. Then, whenever there are conditions to be validated, they are part of what gets passed to the verifier. So, in Dafny and unlike Whiley (I think), the output of the verifier does not affect what types are inferred.

PR #1704, opened 2022-01-12
fix: Fill in missing subset-type checks

### LINKED ISSUE #1676: Soundness issue in concatenation of sequences of subset types
The following verifies in Dafny 3.3 and on today's HEAD (0dcbfac6a4e5e9bd2ca9d4848955169e84b74cbd):

```dafny
lemma nat_pos(n: nat) ensures n >= 0 {}

method Main() {
  nat_pos(([0 as nat] + [-1 as int])[1]);
  assert false;
}
```

Type inference infers that `([0 as nat] + [-1 as int])` should be a `seq<nat>` (not a `seq<int>`), but there does not seem to be a verification condition generated to confirm that the RHS can indeed be cast to a `seq<nat>`. 

I suspect there's an additional translation issue, because `assert v >= 0` fails in the following (it succeeds with an additional call to `nat_pos(v)`, but I don't see why it should be needed):

```dafny
lemma nat_pos(n: nat) ensures n >= 0 {}

method Main() {
  var v: nat := ([0 as nat] + [-1 as int])[1];
  assert v == -1;
  assert v >= 0;
}
```

Here are some other examples in the same theme to help pinpoint the source(s?) of unsoundness:

```dafny
function id<T>(t: T) : T { t }

method Main() {
  var v := id<seq<nat>>(id<seq<nat>>([0]) + id<seq<int>>([-1]))[1];
  assert false;
}
```

```dafny
lemma nonz(x: nat) ensures x >= 0 {}

method Main() {
  var s1 : seq<nat> := [0];
  var s : seq<int> := [-1];
  var s1or0: seq<nat> := s1 + s;
  nonz(s1or0[0]); nonz(s1or0[1]);
  assert -1 >= 0;
  assert false;
}
```

In this one Dafny rejects the program when `s1 + s` is replaces with `s + s1` (`value does not satisfy the subset constraints of 'seq<nat>'`).

```dafny
type int1 = x : int | x == 1 witness 1

method needs1(x: int1) {
  print(1 / x);
}

method Main() {
  var s : seq<int> := [0];
  var s1 : seq<int1> := [1];
  var s1or0: seq<int1> := s1 + s;
  needs1(s1or0[1]);
}
```

```dafny
type int1 = x : int | x == 1 witness 1
type int0 = x : int | x == 0 witness 0

method needs1(x: int1) {
  print(1 / x);
}

method Main() {
  var s1 : seq<int1> := [1];
  var s0 : seq<int0> := [0];
  var s1or0: seq<int1> := s1 + s0;
  needs1(s1or0[1]);
}
```

This issue was discovered jointly with @MikaelMayer .

RustanLeino:
This PR contains several improvements to type inference, and in particular it fixes the soundness issue reported in #1676.

Main type fixes:
* The commit "fix: Look for subset types before basic types" fixes the soundness, but revealed other issues
* fix: Obtain all `ParentTypes()` (previously, a call to this method had been invoked on a non-normalized type, which produced an empty list of parents)
* Improve in type inference: For bound variables (like those in let expressions and the bound variables in `case` patterns), eagerly set the type of the bound variable to the type of the RHS. After all, this is the most general type one can be sure the variable can have. (This improves the situation reported in #1263, but is not a full solution for that problem.)

Other improvements:
* fix: The desugaring of `match` into `VarDeclPattern` had incorrectly set the `isGhost` bit. (I don't think the previous error caused any problems, except for producing confusing output with `/rprint`.)
* fix: Add missing space in error message
* Remove an unnecessary set of `{` `}` in the desugaring of `match` `case`s
* chore: Use the `out var` and `out _` syntax when calling `CheckEnds`
* chore: Use the `.Count` getter instead of the `.Count()` method for `List`s

The PR is probably best reviewed through the individual commits.

Fixes #1676


Source/Dafny/Resolver.cs
```diff
@@ -4546,13 +4546,7 @@ public class ModuleBindings {
       Contract.Ensures(!Contract.ValueAtReturn(out isRoot) || Contract.ValueAtReturn(out headIsRoot)); // isRoot ==> headIsRoot
```
MikaelMayer, line 4546: Could you add a description of what this method does or in which context is is being used? I don't know what "CheckEnds" means, which "End" it is refered to.
Also, what is an artificial type? I did not find it in the documentation

Source/Dafny/Resolver.cs
```diff
@@ -4423,7 +4423,7 @@ public class ModuleBindings {
         Contract.Assert(cl.TypeArgs.Count == B.Count);
         moreXConstraints = false;
         for (int i = 0; i < B.Count; i++) {
-          var msgFormat = "variance for type parameter" + (B.Count == 1 ? "" : "" + i) + " expects {0} {1} {2}";
+          var msgFormat = "variance for type parameter" + (B.Count == 1 ? "" : " " + i) + " expects {0} {1} {2}";
```
MikaelMayer, line 4426: Since the integer is an index and not an ordinal, would it be possible to say "at index 0" instead of just "0" ? It will help us be consistent with the error messages [after this recent change](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Resolver.cs#L16993).
That way, it'll be less confusing when encountering two type parameters, the second one having an error,, and the error would currently say `variance for type parameter 1 expects ....`

I would even prefer to have ordinals here ( i + 1 ) so we wouldn't need the extra "at index".

MikaelMayer (changes requested):
I reviewed all the files through the individual commits, thanks for suggesting that.
Let's have all the tests pass, and a few minor changes and explanations asked.

Source/Dafny/Resolver.cs
RustanLeino: I'll change it to "at index 0" to be consistent with the other recent change.

(Btw, ordinals contain the natural numbers, and both start with 0. And I find it confusing to add arbitrary offsets, like 1. :))

Source/Dafny/Resolver.cs
RustanLeino: The method is used during type inference. I renamed the method from `CheckEnds` to `DetermineRootLeaf`, which is more direct. The comment above the method describes what "root" and "leaf" mean in this context.

An "artificial type" refers to `ArtificialType`, as declared in `DafnyAst.cs`. I clarified this in the comment for `DetermineRootLeaf`, and I also expanded on the description for the class in `DafnyAst.cs`.

Source/Dafny/Resolver.cs
MikaelMayer: Interesting, I never thought of the ambiguity between "Ordinal" as a mathematical meaning (the natural numbers including zero + infinity + (infinity + 1) + ... and so on), and the English meaning of "ordinal" which is always 1-based (the first, the second, the third...) as opposed to "cardinal" which is the number itself (and can contain zero).
Prehaps we should simply have a converter from number to plain text English ordinals at some point.

MikaelMayer (dismissed):
Congratulations on solving this deep problems of type inference. Let's have the tests pass now.

MikaelMayer (approved):
Well done !

PR #1707, opened 2022-01-12
Add env var for extra Dafny arguments during tests

atomb:
The new DAFNY_EXTRA_TEST_ARGUMENTS environment variable will be appended
to the command line arguments for `dafny` when running the integration
tests. This can be useful for extracting intermediate output from the
test suite, such as SMT-Lib files, Boogie source, etc.


nchong-at-aws (approved):

keyboardDrummer (approved):
Alternatively, you could edit the test file right? Is this easier than that?

atomb:
> Alternatively, you could edit the test file right? Is this easier than that?

This will let you supply additional arguments to _every_ test, and seems like it's much easier in that case. My use case, for context, is gathering (lots and lots of) SMT-Lib output.

PR #1708, opened 2022-01-13
Refactor compiler to allow loading backends from separate DLLs

cpitclaudel:
This is the first step towards making it easier to plug in compiler backends.

- To minimize code churn I didn't move each existing backend to a separate DLL.  Should I?
- I'm not a huge fan of the two-layered structure (factory + actual prover), but IIUC that's needed to load a DLL (since types are not first-class values in C#)
- There are surprising dependencies between DafnyAST.cs and the compiler (the AST depended on some values from the compiler); I tagged the corresponding commit with `NEEDS REVIEW`.

With this PR the `/compileTarget` flag can take a path to a DLL.


MikaelMayer:
> This is the first step towards making it easier to plug in compiler backends.
> 
> * To minimize code churn I didn't move each existing backend to a separate DLL.  Should I?
> * I'm not a huge fan of the two-layered structure (factory + actual prover), but IIUC that's needed to load a DLL (since types are not first-class values in C#)
> * There are surprising dependencies between DafnyAST.cs and the compiler (the AST depended on some values from the compiler); I tagged the corresponding commit with `NEEDS REVIEW`.
> 
> With this PR the `/compileTarget` flag can take a path to a DLL.

For the moment, I would advise to not move each existing backend to a separate DLL because it makes development easier. I'm working on refactoring the compiler itself so that it's more modular (e.g. no call to a single CreateForEachLoop that is responsible of subtype checking). Hence I not only need to modify the base compiler, but also the different instances, and I run a test for all of them at the same time.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -21,13 +21,31 @@
...
+
+    public override bool SupportsInMemoryCompilation => true;
+    public override bool TextualTargetIsExecutable => false;
+
+    public override ICompiler CreateInstance(ErrorReporter reporter, ReadOnlyCollection<string> otherFileNames) {
+      return new Compiler(this, reporter);
+    }
+  }
+
+  public class Compiler : SinglePassCompiler {
```
keyboardDrummer, line 45: Might be easier for code navigation if this is still called `CsharpCompiler`

Source/Dafny/Compilers/Compiler-go.cs
```diff
@@ -1886,8 +1902,8 @@ private class SpecialNativeType : UserDefinedType {
     protected override ConcreteSyntaxTree CreateForeachLoop(string tmpVarName, Type collectionElementType, string boundVarName, Type boundVarType, bool introduceBoundVar,
       Bpl.IToken tok, out ConcreteSyntaxTree collectionWriter, ConcreteSyntaxTree wr) {
 
-      var okVar = FreshId("_ok");
-      var iterVar = FreshId("_iter");
+      var okVar = idGenerator.FreshId("_ok");
+      var iterVar = idGenerator.FreshId("_iter");
```
keyboardDrummer, line 1906: Where does this rename come from?

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -2381,7 +2399,7 @@ protected class ClassWriter : IClassWriter {
 
       var w = new ConcreteSyntaxTree();
       if (to.IsRefType) {
-        wr.Format($"({TypeName(to, wr, tok)})({w})");
+        wr.Format($"(({TypeName(to, wr, tok)})({w}))");
```
keyboardDrummer, line 2402: Is this a bugfix?

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -58,19 +50,24 @@ public abstract class Compiler {
...
+    protected static void ReportError(ErrorReporter reporter, Bpl.IToken tok, string msg, ConcreteSyntaxTree/*?*/ wr, params object[] args) {
       Contract.Requires(msg != null);
       Contract.Requires(args != null);
 
-      Reporter.Error(MessageSource.Compiler, tok, msg, args);
-      if (wr != null) {
-        wr.WriteLine("/* {0} */", string.Format("Compilation error: " + msg, args));
-      }
+      reporter.Error(MessageSource.Compiler, tok, msg, args);
+      // FIXME this is not valid syntax in all supported languages (e.g. Python)
```
keyboardDrummer, line 65: FIXME. Please replace this either with a fix or a GH issue.

Source/Dafny/DafnyOptions.cs
```diff
@@ -99,7 +88,7 @@ public enum CompilationTarget {
     public bool AllowGlobals = false;
     public bool CountVerificationErrors = true;
     public bool Optimize = false;
-    public bool AutoTriggers = true;
+    public bool AutoTriggers = true; // FIXME: unused?
```
keyboardDrummer, line 91: FIXME

Source/Dafny/Resolver.cs
```diff
@@ -4314,8 +4314,8 @@ public class ModuleBindings {
 
       // Finally, of the big-enough native types, pick the first one that is
       // supported by the selected target compiler.
-      foreach (var nativeT in bigEnoughNativeTypes) {
-        if ((nativeT.CompilationTargets & DafnyOptions.O.CompileTarget) != 0) {
+      foreach (var nativeT in bigEnoughNativeTypes) { // FIXME shouldn't the resolver be agnostic to the compilation target?
```
keyboardDrummer, line 4317: FIXME. I suggest you cut an issue.

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -149,72 +149,17 @@ public enum CommandLineArgumentsResult {
...
-              extension == null ? "" : extension);
-            return CommandLineArgumentsResult.PREPROCESSING_ERROR;
-          }
+        var supportedExtensions = DafnyOptions.O.CompilerFactoryInstance.SupportedExtensions;
+        if (supportedExtensions.Contains(extension)) {
+          otherFiles.Add(file);
+        } else if (!isDafnyFile) {
+          ExecutionEngine.printer.ErrorWriteLine(Console.Out, 
+            "*** Error: '{0}': Filename extension '{1}' is not supported. Input files must be Dafny programs (.dfy) or supported auxiliary files ({2})",
+            file, extension ?? "", string.Join(", ", supportedExtensions.Contains(extension)));
```
keyboardDrummer, line 158: Did you mean `string.Join(", ", supportedExtensions)` ?

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -665,33 +572,13 @@ private record TargetPaths(string Directory, string Filename) {
...
-          break;
-        case DafnyOptions.CompilationTarget.Python:
-          compiler = new Dafny.PythonCompiler(dafnyProgram.reporter);
-          break;
-      }
+      Dafny.ICompiler compiler =
+        DafnyOptions.O.CompilerFactoryInstance.CreateInstance(dafnyProgram.reporter, otherFileNames);
 
-      var hasMain = compiler.HasMain(dafnyProgram, out var mainMethod);
+      var hasMain = SinglePassCompiler.HasMain(dafnyProgram, out var mainMethod); // CPC FIXME: Where should this method go?
```
keyboardDrummer, line 581: FIXME

Test/SelfHosting/Main.cs
```diff
@@ -0,0 +1,58 @@
+using Antlr4.Runtime;
```
keyboardDrummer, line 1: I'm confused. What is the purpose of the SelfHosting folder?

cpitclaudel:
> For the moment, I would advise to not move each existing backend to a separate DLL because it makes development easier. 

:+1:, although it might be nice to split off the Java compiler, because compiling the runtime every time is a pain.

MikaelMayer:
> compiling

I investigated more why I believed we should hold off splitting the compilers into different projects. It turns out I had some instability with some versions of Newtonsoft.Json.dll (DafnyDriver requires 9.0.0.0, DafnyLanguageServer requires 11.0.0.0, so if I compile DafnyDriver, the language server does not work).
So I don't have any more strong argument of not splitting the compilers.

Source/Dafny/AST/ASTExport/Program.cs
```diff
@@ -0,0 +1,249 @@
+﻿using System;
```
keyboardDrummer, line 1: Should these changes be in this PR? This doesn't seem to relate to the PR title, and the PR was already quite large.

keyboardDrummer:
>> For the moment, I would advise to not move each existing backend to a separate DLL because it makes development easier.

> 👍, although it might be nice to split off the Java compiler, because compiling the runtime every time is a pain.

What do we mean by splitting off the compilers? We'd move the compiler file to a new .NET project, sure, but where do we leave that project? 

If we keep that project in the Dafny solution then it wouldn't impact development, right? It just means that Rider can build DafnyPipeline without building the Java compiler.

cpitclaudel:
>  It just means that Rider can build DafnyPipeline without building the Java compiler.

That was my hope, yes.

Source/Dafny/Compilers/Compiler-go.cs
cpitclaudel: I don't think there's a rename here?

Source/Dafny/Compilers/Compiler-Csharp.cs
cpitclaudel: It was, I removed it from the branch (I shouldn't have pushed it)

Source/Dafny/Compilers/SinglePassCompiler.cs
cpitclaudel: https://github.com/dafny-lang/dafny/issues/1729

Test/SelfHosting/Main.cs
cpitclaudel: This shouldn't have been checked in, sorry

Source/Dafny/AST/ASTExport/Program.cs
cpitclaudel: No, I mistakenly continued working on the PR branch.  I've reset it now.

Source/Dafny/Compilers/ICompiler.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.ObjectModel;
+using System.IO;
+
+namespace Microsoft.Dafny;
+
+public interface ICompiler {
```
robin-aws, line 6: This interface is so much nicer. :)

Here's an idea to consider for the very next PR (since it's not necessary here and the PR is already large): could we add a method to be invoked earlier in the pipeline, before verification, that can produce compiler-specific errors? Something like `CheckCompatibility`.

The use case would be a codebase using a specific compiler, especially one of the more incomplete ones like C++, where it's currently pretty frustrating to spend a ton of time getting your code to verify successfully only to then find out "oh wait, my compiler doesn't support unbounded `int`s" and have to start over. This is particularly easy for newer Dafny programmers to hit, since the documentation on which features are supported by which compilers is something you have to think to look up, and isn't that obvious for newcomers.

cpitclaudel: > Here's an idea to consider for the very next PR (since it's not necessary here and the PR is already large): could we add a method to be invoked earlier in the pipeline, before verification, that can produce compiler-specific errors? Something like CheckCompatibility.

The compiler is fast and verification errors lead to runtime (not compile-time) errors, so we could just run the compiler first or in parallel with verification instead of running verification first and only then running the compiler.  So instead of `resolve → verify → translate → compile` we could just run 

```
resolve → verify
        → translate → compile
```

Or even `resolve → translate → verify → compile` if we don't want concurrency.

cpitclaudel: Probably a good entry for your compiler planning doc!

robin-aws: Yup, you're absolutely right that there's other ways to get the same result (and it would be better to not have another method to implement if possible). I'm not sure I believe "verification errors lead to runtime (not compile-time) errors" is always true, or that we want to ensure it is always true in the future, but we can work that into the design. :)

Source/Dafny/Resolver.cs
cpitclaudel: https://github.com/dafny-lang/dafny/issues/1732

keyboardDrummer:
> > It just means that Rider can build DafnyPipeline without building the Java compiler.
> 
> That was my hope, yes.

Sounds great, but let's do it in a separate PR

keyboardDrummer (dismissed):

keyboardDrummer (approved):

Source/Dafny/DafnyOptions.cs
```diff
@@ -71,18 +71,7 @@ public enum PrintModes {
...
-      JavaScript = 2,
-      Go = 4,
-      Java = 8,
-      Cpp = 16,
-      Php = 32,
-      Python = 64
-    }
-
-    public CompilationTarget CompileTarget = CompilationTarget.Csharp;
+    public CompilerFactory CompilerFactoryInstance;
```
MikaelMayer, line 74: With flags it was obvious we could run multiple compilers at the same time.
How are you going to do it with this? I suggest you use a comma-separated list of compilers

MikaelMayer (commented):
Also, how will it work to provide new compiler assemblies? Are you reusing the same compileTarget option?

Source/Dafny/DafnyOptions.cs
cpitclaudel: > With flags it was obvious we could run multiple compilers at the same time.

I don't think we ever could actually run multiple compilers at the same time, could we?

cpitclaudel:
> Also, how will it work to provide new compiler assemblies? Are you reusing the same compileTarget option?

That's right, you just pass `/compileTarget:path/to/a/library.dll`

cpitclaudel:
@MikaelMayer and I disucssed this today.  We think it's better if instead of reusing the compileTarget flag to give a DLL we instead make a new flag `plugins` that takes a list of DLLs, each of which exposes an `ICompiler`.  Then we can resolve the value of `compileTarget` by looking for an ICompiler with the appropriate `SupportedLanguage`.  So you'd write `/plugin:XYZCompiler.dll /compileTarget:xyz`.

This same plugin flag can be used for other kinds of plugins, and @MikaelMayer will implement it as part of https://github.com/dafny-lang/dafny/pull/1739 .  Then I will rebase this PR on top of that.

Finally we decided it would be simpler to get rid of the factory and instantiate an `ICompiler` instance  (renamed to `ICodeGenerator`) with no arguments right from the beginning of the process, (and later on pass the `reporter`).  I will make these changes after @MikaelMayer completes his.

keyboardDrummer:
> This same plugin flag can be used for other kinds of plugins, and @MikaelMayer will implement it as part of #1739 . Then I will rebase this PR on top of that.

Why don't we merge this PR first, since this is already approved?

MikaelMayer:
> Why don't we merge this PR first, since this is already approved?

Maybe because 1) the tests are not working yet, and 2) offering an API to implement a factory and then immediately switching to a plugin strategy would be not backward compatible with the factory (for example, /compilationTarget won't accept a DLL anymore once we switch to the plugin system). and 3) it makes sense to accept DLLs as a separate argument rather than changing the CompilationTarget.

MikaelMayer:
I'm voluntarily not writing "changes requested" because I don't want to block this PR, but we are still in discussion about the plugins system which would make this PR easier to implement.

cpitclaudel:
> Why don't we merge this PR first, since this is already approved?

The only client of this PR is my compiler work, which can happen in a branch.  The plugin API is the result of discussing with @MikaelMayer how to unify his resolver plugins and my compiler plugins, so I don't think we should merge this until the plugin work is done :)

cpitclaudel:
Closing this, since it was superseded by my later PR on top of Mikael's one

PR #1711, opened 2022-01-13
First refactoring to discharge a bit the Resolver.

MikaelMayer:
Moved the class `TypeConstraint` into its own file in a subfolder `Resolver/`


MikaelMayer:
I moved one class, and I found the following problem. It won't let me create the _namespace_ `Microsoft.Dafny.Resolver` because the _class_ already exists.

I have the choice between declaring Resolver a partial class (and thus keep everything the same), or actually make the class `TypeConstraint` top-level, except that it will be in a subfolder. I could also create a new namespace (like "Resolution") but that's not very useful I guess.
What would you have in mind concerning the best way to refactor this @keyboardDrummer ?

Source/Dafny/Resolver/TypeConstraint.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Linq;
+using System.Numerics;
+using System.Diagnostics.Contracts;
+using JetBrains.Annotations;
+using Microsoft.BaseTypes;
+using Microsoft.Boogie;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.Dafny {
+  public class TypeConstraint {
```
keyboardDrummer, line 20: Consider changing this class into a record with a primary constructor `record TypeConstraint(Type super, Type sub, ErrorMsg errMsg, bool keepConstraints)`, that's an easy to do refactoring which makes it easy to see from the header of the record what it does.

keyboardDrummer (approved):

cpitclaudel:
Maybe we can do this without losing history?  https://stackoverflow.com/a/46484848

The idea is that you 1. move the file on a separate branch; 2. merge with both copies kept; and 3. make changes to the copy.  With that git keeps the file's history, so tools like `git blame` work.

(This is one case where squash-merging is bad: we need a proper merge for this to work).

keyboardDrummer:
> Maybe we can do this without losing history? https://stackoverflow.com/a/46484848

The history isn't actually lost right? It's just harder to traverse. I'm OK with that.

> The idea is that you 1. move the file on a separate branch; 2. merge with both copies kept; and 3. make changes to the copy. With that git keeps the file's history, so tools like `git blame` work.

How does that work? Is the crux to do a copy in one commit and remove code from both the copy and the original file in a second commit? Why do the instructions mention a branch?

MikaelMayer:
For such a small file, I'm fine with making the history a bit harder to traverse. As @keyboardDrummer said, it's possible to see where it comes from anyway, so it might not be worth the trouble.

cpitclaudel:
> The history isn't actually lost right? It's just harder to traverse. 
> I'm fine with making the history a bit harder to traverse.

Not harder, fully unsupported, AFAIK.
But I could be wrong: do you have a way to blame your new file, @MikaelMayer ?

> How does that work? Is the crux to do a copy in one commit and remove code from both the copy and the original file in a second commit? Why do the instructions mention a branch?

Git has crappy copy detection and good move detection.  So the trick is to implement a copy as two moves, using a branch to force it to keep both the original file and the moved version.

MikaelMayer:
> Not harder, fully unsupported, AFAIK.

When you git blame TypeConstraint, you'll eventually find the commit in which it was refactored.
Then, you can look at the version of Resolver.cs at that commit, take the commit before, and perform a git blame on the line you want. So it's just harder, but you'll eventually find it since the history of Resolver.cs is not erased.

cpitclaudel:
> Then, you can look at the version of Resolver.cs at that commit, take the commit before, and perform a git blame on the line you want. So it's just harder, but you'll eventually find it since the history of Resolver.cs is not erased.

Yes, that's pretty much what I meant by breaking ;)  You have to do it by hand.  It's especially painful when you're looking at the history of a range of lines with `git log -L`.

MikaelMayer:
We got another problem now.
https://github.com/dafny-lang/dafny/runs/4857055436?check_suite_focus=true

Basically the dotnet formatter complains that the type TypeConstraint.ErrorMsg is not defined because it's internal to TypeConstraint (and it's a struct record, not a class anymore), but I cannot move it outside of it without making `private static readonly List<ErrorMsg> ErrorsToBeReported = new List<ErrorMsg>();` public because the class `ErrorMsg` depends on it.
@keyboardDrummer  what is your take-away on that.

keyboardDrummer:
> We got another problem now. https://github.com/dafny-lang/dafny/runs/4857055436?check_suite_focus=true
> 
> Basically the dotnet formatter complains that the type TypeConstraint.ErrorMsg is not defined because it's internal to TypeConstraint (and it's a struct record, not a class anymore), but I cannot move it outside of it without making `private static readonly List<ErrorMsg> ErrorsToBeReported = new List<ErrorMsg>();` public because the class `ErrorMsg` depends on it. @keyboardDrummer what is your take-away on that.

Seems like a bug in the formatter, as if it's using a too old version of Roslyn. Might it's using this version: https://github.com/dotnet/format/blob/main/eng/Version.Details.xml#L4

Changing `readonly record struct` to `record` fixes the problem for some reason, and also resolves another warning. I pushed this change to your branch.

keyboardDrummer (approved):

PR #1712, opened 2022-01-14
Add CSV logger for test results and timing

atomb:
Adds the ability to log proof goal results and timing in CSV format.

Cherry-picked from robin-aws and slightly tweaked.

### DIFF
--- a/Source/DafnyDriver/BoogieXmlConvertor.cs
+++ b/Source/DafnyDriver/BoogieXmlConvertor.cs
@@ -115,7 +115,7 @@ namespace Microsoft.Dafny {
         EndTime = DateTimeOffset.Parse(endTime)
       };
 
-      if (outcome is "correct" or "valid") {
+      if (outcome == "correct") {
         testResult.Outcome = TestOutcome.Passed;
       } else {
         testResult.Outcome = TestOutcome.Failed;
--- a/Source/DafnyDriver/BoogieXmlConvertor.cs
+++ b/Source/DafnyDriver/BoogieXmlConvertor.cs
@@ -24,8 +24,6 @@ namespace Microsoft.Dafny {
   /// </summary>
   public static class BoogieXmlConvertor {
 
-    public static TestProperty ResourceCountProperty = TestProperty.Register("TestResult.ResourceCount", "TestResult.ResourceCount", typeof(int), typeof(TestResult));
-
     public static void RaiseTestLoggerEvents(string fileName, List<string> loggerConfigs) {
       // Provide just enough configuration for the loggers to work
       var parameters = new Dictionary<string, string> {
@@ -108,7 +106,6 @@ namespace Microsoft.Dafny {
                                        .Single(n => n.Name.LocalName == "conclusion");
       var endTime = conclusionNode.Attribute("endTime")!.Value;
       var duration = float.Parse(conclusionNode.Attribute("duration")!.Value);
-      var resourceCount = conclusionNode.Attribute("resourceCount")?.Value;
       var outcome = conclusionNode.Attribute("outcome")!.Value;
 
       var testCase = TestCaseForEntry(currentFileFragment, name);
@@ -117,10 +114,6 @@ namespace Microsoft.Dafny {
         Duration = TimeSpan.FromMilliseconds((long)(duration * 1000)),
         EndTime = DateTimeOffset.Parse(endTime)
       };
-
-      if (resourceCount != null) {
-        testResult.SetPropertyValue(ResourceCountProperty, int.Parse(resourceCount));
-      }
 
       if (outcome is "correct" or "valid") {
         testResult.Outcome = TestOutcome.Passed;
--- a/Source/DafnyDriver/BoogieXmlConvertor.cs
+++ b/Source/DafnyDriver/BoogieXmlConvertor.cs
@@ -24,33 +24,25 @@ namespace Microsoft.Dafny {
   /// </summary>
   public static class BoogieXmlConvertor {
 
-    public static void RaiseTestLoggerEvents(string fileName, string loggerConfig) {
-      string loggerName;
-      Dictionary<string, string> parameters;
-      int semiColonIndex = loggerConfig.IndexOf(";");
-      if (semiColonIndex >= 0) {
-        loggerName = loggerConfig[..semiColonIndex];
-        var parametersList = loggerConfig[(semiColonIndex + 1)..];
-        parameters = parametersList.Split(",").Select(s => {
-          var equalsIndex = s.IndexOf("=");
-          return (s[..equalsIndex], s[(equalsIndex + 1)..]);
-        }).ToDictionary(p => p.Item1, p => p.Item2);
-      } else {
-        loggerName = loggerConfig;
-        parameters = new();
-      }
+    public static TestProperty ResourceCountProperty = TestProperty.Register("TestResult.ResourceCount", "TestResult.ResourceCount", typeof(int), typeof(TestResult));
 
-      // The only supported value for now
-      if (loggerName != "trx") {
-        throw new ArgumentException($"Unsupported verification logger name: {loggerName}");
-      }
-
-      // Provide just enough configuration for the TRX logger to work
-      parameters["TestRunDirectory"] = Constants.DefaultResultsDirectory;
-
+    public static void RaiseTestLoggerEvents(string fileName, List<string> loggerConfigs) {
       var events = new LocalTestLoggerEvents();
-      var logger = new TrxLogger();
-      logger.Initialize(events, parameters);
+      // Provide just enough configuration for the loggers to work
+      var parameters = new Dictionary<string, string> {
+        ["TestRunDirectory"] = Constants.DefaultResultsDirectory
+      };
+      foreach (var loggerConfig in loggerConfigs) {
+        if (loggerConfig == "trx") {
+          var logger = new TrxLogger();
+          logger.Initialize(events, parameters);
+        } else if (loggerConfig == "csv") {
+          var csvLogger = new CSVTestLogger();
+          csvLogger.Initialize(events, parameters);
+        } else {
+          throw new ArgumentException("Unsupported verification logger config: {loggerConfig}");
+        }
+      }
       events.EnableEvents();
 
       // Sort failures to the top, and then slower procedures first.
@@ -102,6 +94,7 @@ namespace Microsoft.Dafny {
                                        .Single(n => n.Name.LocalName == "conclusion");
       var endTime = conclusionNode.Attribute("endTime")!.Value;
       var duration = float.Parse(conclusionNode.Attribute("duration")!.Value);
+      var resourceCount = conclusionNode.Attribute("resourceCount")?.Value;
       var outcome = conclusionNode.Attribute("outcome")!.Value;
 
       var testCase = TestCaseForEntry(currentFileFragment, name);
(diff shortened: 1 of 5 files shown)


robin-aws:
Thanks! I'll add a few thoughts as comments for how I was thinking of trimming this down as a first step. Mainly, let's punt the "that includes resource count" for now and just add the CSV logger with the three obvious fields, while we work on adding the resource count to the Boogie XML.

Source/DafnyDriver/BoogieXmlConvertor.cs
```diff
@@ -24,33 +24,25 @@ namespace Microsoft.Dafny {
...
-      }
-
-      // The only supported value for now
-      if (loggerName != "trx") {
-        throw new ArgumentException($"Unsupported verification logger name: {loggerName}");
-      }
-
-      // Provide just enough configuration for the TRX logger to work
-      parameters["TestRunDirectory"] = Constants.DefaultResultsDirectory;
+    public static TestProperty ResourceCountProperty = TestProperty.Register("TestResult.ResourceCount", "TestResult.ResourceCount", typeof(int), typeof(TestResult));
```
robin-aws, line 27: We can just remove this and the references for now. When we do put it back, I'd prefer it wasn't inside the `BoogieXMLConvertor` namespace since we want to change the implementation of how the test events are raised later.

Source/DafnyDriver/CSVTestLogger.cs
```diff
@@ -0,0 +1,35 @@
...
+    private void TestResultHandler(object sender, TestResultEventArgs e) {
+      results.Add(e.Result);
+    }
+
+    private int? GetResourceCount(TestResult result) {
+      return result.GetPropertyValue<int?>(BoogieXmlConvertor.ResourceCountProperty, null);
+    }
+
+    private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e) {
+      foreach (var result in results.OrderByDescending(GetResourceCount)) {
```
robin-aws, line 30: We can just make this sort by duration instead for now. I'm amused that I made it sort by resource count without actually adding the column :)

Source/DafnyDriver/BoogieXmlConvertor.cs
```diff
@@ -24,33 +24,25 @@ namespace Microsoft.Dafny {
...
-        var parametersList = loggerConfig[(semiColonIndex + 1)..];
-        parameters = parametersList.Split(",").Select(s => {
-          var equalsIndex = s.IndexOf("=");
-          return (s[..equalsIndex], s[(equalsIndex + 1)..]);
-        }).ToDictionary(p => p.Item1, p => p.Item2);
-      } else {
-        loggerName = loggerConfig;
-        parameters = new();
-      }
-
```
robin-aws, line 42: It makes sense to parse all logger configs like this as the syntax is not logger-specific.

Source/DafnyDriver/BoogieXmlConvertor.cs
atomb: Oh, right. I started to remove references to resource counts but didn't remove all of them. Thanks!

Source/DafnyDriver/CSVTestLogger.cs
atomb: Ah, you weren't actually that strange. :) I removed that column, just not the code to track it.

Source/DafnyDriver/BoogieXmlConvertor.cs
atomb: Yep, that's back in now, and makes logging to a file possible again.

Source/DafnyDriver/BoogieXmlConvertor.cs
```diff
@@ -111,7 +115,7 @@ public static class BoogieXmlConvertor {
         EndTime = DateTimeOffset.Parse(endTime)
       };
 
-      if (outcome == "correct") {
+      if (outcome is "correct" or "valid") {
```
robin-aws, line 118: This isn't necessary any more, as an earlier change made `TestResultForSplit` a separate method.

Source/DafnyDriver/CSVTestLogger.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+
+namespace Microsoft.Dafny {
+  public class CSVTestLogger : ITestLoggerWithParameters {
+
+    private readonly List<TestResult> results = new();
```
robin-aws, line 12: This probably should be a thread-safe implementation of some kind, to be a robust implementation of `ITestLogerWithParameters`.

The one and only use right now is from `BoogieXmlConvertor` so not technically necessary, but especially if we move away from the `/xml`-based implementation we could easily want to handle concurrent `TestResultHandler` invocations.

Source/DafnyDriver/CSVTestLogger.cs
```diff
@@ -0,0 +1,38 @@
...
+        writer = new StreamWriter(filename);
+      }
+    }
+
+    private void TestResultHandler(object sender, TestResultEventArgs e) {
+      results.Add(e.Result);
+    }
+
+    private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e) {
+      var realWriter = writer ?? Console.Out;
```
robin-aws, line 31: I wonder if imitating the TRX logger and defaulting to a generated file name would be better? We could support `csv:LogFileName=-` if we wanted to dump to stdout, and I'm not even sure that's worth it.

Source/DafnyDriver/CSVTestLogger.cs
```diff
@@ -0,0 +1,38 @@
...
+      foreach (var result in results.OrderByDescending(r => r.Duration)) {
+        realWriter.WriteLine($"{result.TestCase.DisplayName}, {result.Outcome}, {result.Duration}");
```
robin-aws, line 33: We should include a header row to label the columns too - that will probably make for smoother tooling support. I'd suggest using the "property" identifiers (https://github.com/dafny-lang/dafny/pull/1708) for maximum compatibility:

`TestResult.DisplayName, TestResult.Outcome, TestResult.Duration`

robin-aws (commented):
LGTM, just a couple more tweaks

Source/DafnyDriver/CSVTestLogger.cs
atomb: That does overall seem like the right way to go, though the specific logic the TRX logger uses to choose file names is quite complex. I'll look into approaches that could give us the core benefits of that without hundreds of lines of extra code. :)

atomb: And now there's some code to auto-generate log file names by default. It uses similar default naming conventions to the TRX logger, but is less configurable.

Source/DafnyDriver/CSVTestLogger.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public void Initialize(TestLoggerEvents events, Dictionary<string, string> parameters) {
+      events.TestResult += TestResultHandler;
+      events.TestRunComplete += TestRunCompleteHandler;
+      if (parameters.TryGetValue("LogFileName", out string filename)) {
+        writer = filename == "-" ? Console.Out : new StreamWriter(filename);
+      } else {
+        // Auto-generate a file name if none is specified. This uses a
+        // similar approach to the TRX logger, but with simpler logic.
```
robin-aws, line 26: Appreciate the effort to trim the scope, but unfortunately I think we actually need at least some of the logic to avoid collisions. Otherwise if you invoke `dafny` in parallel (which lots of Makefiles will do) you could actually collide on the same second relatively easily.

robin-aws (changes requested):
Just one issue left around concurrent file name collisions

Source/DafnyDriver/CSVTestLogger.cs
atomb: Fortunately, that logic isn't the most complex part of the TRX implementation, and is pretty easy to replicate (or approximate, with what I think is perhaps a slightly better approach).

Source/DafnyDriver/CSVTestLogger.cs
```diff
@@ -0,0 +1,63 @@
...
+    private TextWriter writer;
+
+    public void Initialize(TestLoggerEvents events, string testRunDirectory) {
+    }
+
+    public void Initialize(TestLoggerEvents events, Dictionary<string, string> parameters) {
+      events.TestResult += TestResultHandler;
+      events.TestRunComplete += TestRunCompleteHandler;
+      if (parameters.TryGetValue("LogFileName", out string filename)) {
+        writer = filename == "-" ? Console.Out : new StreamWriter(filename);
```
keyboardDrummer, line 23: Is there a way to avoid the `Console.Out` reference? Seems like both it and the filename have some potential concurrency issues.

How does this `ITestLoggerWithParameters` interface work? Is that something we need to use? It seems a little restrictive.

atomb: Ah, it looks like retrying if an `IOException` is thrown when opening with `CreateNew` will do the trick.

keyboardDrummer: Could you say something more about how `ITestLoggerWithParameters` is called? How is the `parameters` parameter filled in?

atomb: The BoogieXmlConverter class creates either a CSV or TRX logger and initializes them with parameters based on command-line options, using the code [here](https://github.com/atomb/dafny/blob/csv-logger/Source/DafnyDriver/BoogieXmlConvertor.cs#L34), which is called from [here](https://github.com/atomb/dafny/blob/csv-logger/Source/DafnyDriver/DafnyDriver.cs#L70).

robin-aws: There an example of providing those parameters at the command line here FYI: https://github.com/dafny-lang/dafny/blob/master/Test/logger/VerificationLogger.dfy#L1

The `BoogieXmlConverter` class is really translating the data in the Boogie xml file into events using the VSTest `TestResult` class, and then the TRX and/or CSV loggers have event handlers to process those events. In the near future we want to move away from relying on the `/xml` Boogie feature and instead have Boogie raise verification events in a more incremental way for clients. At that point we can translate the Boogie events directly to `TestResult` events, but the loggers can work exactly as they do now (although hence caring about concurrent access, even those that's not possible yet).

We're really borrowing the whole .NET VSTest core model here, which I assert makes sense given that verification tasks are not that different from tests in a development workflow. The core model supports adding extra properties nicely: they won't show up in the TRX file format (which is less extensible), but can easily be added to the CSV output.

robin-aws (changes requested):
Looks great! Only thing missing is a test. :) Sorry I didn't think of that earlier. You should be able to just copy and slightly modify https://github.com/dafny-lang/dafny/blob/master/Test/logger/VerificationLogger.dfy though

Source/DafnyDriver/CSVTestLogger.cs
keyboardDrummer: Alright, thanks :-) PR LGTM

keyboardDrummer (approved):

robin-aws (approved):

PR #1716, opened 2022-01-14
Changes in Python compiler to successfully pass the test comp/Hello.dfy.

prvshah51:
Needs cleaning, keeping draft till then.


Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -228,7 +251,9 @@ protected class ClassWriter : IClassWriter {
 
     protected override void DeclareLocalVar(string name, Type type, IToken tok, bool leaveRoomForRhs, string rhs,
       ConcreteSyntaxTree wr) {
-      throw new NotImplementedException();
+      wr.Write("{0}{1}", name, type != null ? "=" + TypeName(type, wr, tok) + "()" : "");
```
prvshah51, line 254: The else part can be used to define object of a class.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -1,6 +1,6 @@
 class _dafny:
     def print(toPrintVal):
         if type(toPrintVal)== bool:
-            print("true" if toPrintVal else "false")
+            print("true" if toPrintVal else "false",end="")
```
prvshah51, line 4: "end" to keep all printed statements in the same line as newLine operators are printed by compiler.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -200,9 +205,27 @@ protected class ClassWriter : IClassWriter {
...
+      var xType = type.NormalizeExpand();
+
+      if (xType is BoolType) {
+        return "bool";
+      } else if (xType is CharType) {
+        return "char";
+      } else if (xType is IntType || xType is BigOrdinalType) {
+        return "int";
+      } else {
+        Contract.Assert(false); throw new cce.UnreachableException();  // unexpected type
```
MikaelMayer, line 227: Prefer `NotImplementedException()`

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -332,11 +357,21 @@ protected class ClassWriter : IClassWriter {
...
+        var i = (BigInteger)e.Value;
+        wr.Write($"{i}");
       }
     }
 
     protected override void EmitStringLiteral(string str, bool isVerbatim, ConcreteSyntaxTree wr) {
-      throw new NotImplementedException();
+
+      if (!isVerbatim) {
+        wr.Write("\"{0}\"", str);
```
MikaelMayer, line 373: Please escape characters accordingly (double quotes, perhaps newlines)

Test/comp/Hello.dfy
```diff
@@ -2,12 +2,13 @@
...
+// RUN: %dafny /compile:3 /spillTargetCode:2 /compileTarget:py "%s" >> "%t"
 // RUN: %diff "%s.expect" "%t"
 
 method Main() {
-  print "Hello, JavaScript! Best, Dafny\n";
-  var x := 14;
-  print "x is ", x, "\n";
-  var y := false;
-  print "y is ", y, "\n";
+    print "Hello, JavaScript! Best, Dafny\n";
```
MikaelMayer, line 9: Hello, Python!

MikaelMayer (changes requested):
Three things and your are good to go ! It's great to see the python compiler come to life.

Test/comp/Hello.dfy
prvshah51: This is a test file, I think we should not make any changes in there.

robin-aws: On the one hand I THINK Mikael was just being cheerful and silly :)

On the other hand, it would be better for a generic test run against multiple compilers not to specifically greet JavaScript and make the other languages feel somehow left out. Seems like a good change to make this "Hello, my favourite language!" while we're here! We can certainly change test cases to improve understanding or coverage.

On the third hand I'm probably overthinking the whole thing. :)

prvshah51: I also think it would be better to have some string not mentioning any language.I will do it here.

prvshah51: How does "Hello, World! Best, Dafny\n" sound?

Source/Dafny/Compilers/Compiler-python.cs
prvshah51: I understand your concern, I just want to keep it related only to the test file  so nothing is added which is not used and thus less chance of errors.

MikaelMayer: I understand your strategy and it does makes sense, especially in early stage.
However, consider this other point of view: this is intentionally implementing a function in a way that is obviously wrong (not incomplete, it returns a result even if this result is wrong) just to make that test pass. Although this might be fine in this particular case (which still bothers me) is in general a recipe for late errors that one will discover later and that will be more painful. On the other side, it's ok to implement something correctly that will tested later.

You might at least want to write the following:

`if(str.Contains("\"") || str.Contains("\\")) { throw new NotImplementedException(); }`

so that you are not caught by surprise later.

prvshah51: just found out the flag isVerbatim is to check if the string has any escape characters, thanks for the comment I will add the `NotImplementedException()` in the else.

MikaelMayer: I think that isVerbatim is a Dafny language that just checks if the flag "@" was in front of the original string, it does not assume anything about the content. Please double check.

prvshah51: sure.

Source/Dafny/Compilers/Compiler-python.cs
MikaelMayer: Can you write `$"={TypeName(type,wr,tok)}()"`? It's more readable (less double quotes and plus symbols.

Source/Dafny/Compilers/Compiler-python.cs
MikaelMayer: Ok, thanks for double checking. Actually, it's my bad. Dafny records string escales in the string value directly, so there isn't anything you need to do, since the subset that Dafny supports is also supported by Python. You can consider this resolved. Thanks again and sorry for the trouble.

prvshah51: It's always better to recheck, thanks for your input.

Source/Dafny/Compilers/Compiler-python.cs
prvshah51: edited.

MikaelMayer (approved):
Looks good to me ! Great job @prvshah51

PR #1733, opened 2022-01-19
updates to RELEASE.txt and  a comment in version.cs

prvshah51:


RELEASE.txt
```diff
@@ -5,6 +5,7 @@ This is the procedure for generating a new Dafny release.
 
 1) In an up-to-date repository, with the master branch checked out
    and up-to-date (e.g., git pull upstream)
+   checkout a nre branch from master (e.g., git checkout -b branchname master)
```
fabiomadge, line 8:
```suggestion
   checkout a new branch from master (e.g., git checkout -b branchname master)
```

Source/version.cs
```diff
@@ -1,4 +1,4 @@
 using System.Reflection;
-// Version 3.4.0, year 2018+4 month 1 day 12
+// Version 3.4.0, year 2018+4, month 1 day 12 as for date January 12th
```
fabiomadge, line 2: Why do we need to change this file? The way without the comma seems to be how we did it before and the date is reflected in the commit date.

RELEASE.txt
```diff
@@ -28,6 +29,8 @@ This is the procedure for generating a new Dafny release.
 5) In the Github releases, edit the new release to add in the release notes.
 
 6) Push the Publish button.
+   if doing a pre-release stop here, as steps further are to define the release as
+   latest version to be installed which is not needed.
```
fabiomadge, line 33:
```suggestion
   If preparing a pre-release, stop here, as the following steps declare the release
   as the latest version, which is not the intention.
```

Source/version.cs
prvshah51: This can help future releases as there is no mention about how to choose version number, so I thought I can add an example. Do you think instead of this the RELEASE.txt file should hold this information?

fabiomadge: I agree that some additional explanation can't hurt and think the `RELEASE.txt` is the right place for that. That way we don't have to think about keeping it up to date with every release.

Source/version.cs
```diff
@@ -1,4 +1,4 @@
 using System.Reflection;
-// Version 3.4.0, year 2018+4 month 1 day 12
+// Version 3.4.0, year 2018+4, month 1 day 12
```
fabiomadge, line 2:
```suggestion
// Version 3.4.0, year 2018+4 month 1 day 12
```

RELEASE.txt
```diff
@@ -5,10 +5,13 @@ This is the procedure for generating a new Dafny release.
 
 1) In an up-to-date repository, with the master branch checked out
    and up-to-date (e.g., git pull upstream)
+   checkout a new branch from master (e.g., git checkout -b branchname master)
    cd to the repo root (i.e., dafny, which contains Source, Binaries, etc.)
 
 2) Select a version number (descriptor) $VER (e.g., "3.0.0" or "3.0.0-alpha")
    and select an internal version number (e.g., "3.0.0.30203")
+   guideline for internal version number:Use the no. of years since 2018 and release date
+   e.g.,Release on January 12th 2022 = 2018 + 4 years and date in "mmdd" - x.y.z.40112
```
fabiomadge, line 14:
```suggestion
   and select an internal version number (e.g., "3.0.0.30203").
   Obtain it by concatenating the digits in the release date,
   e.g.,Release on January 12th 2022 = 2018 + 4 years and date in "mmdd" - x.y.z.40112
```

fabiomadge: Sorry, I changed my mind after. Feel free to adopt what makes sense to you.

RELEASE.txt
```diff
@@ -5,10 +5,14 @@ This is the procedure for generating a new Dafny release.
...
 1) In an up-to-date repository, with the master branch checked out
    and up-to-date (e.g., git pull upstream)
+   checkout a new branch from master (e.g., git checkout -b branchname master)
    cd to the repo root (i.e., dafny, which contains Source, Binaries, etc.)
 
 2) Select a version number (descriptor) $VER (e.g., "3.0.0" or "3.0.0-alpha")
    and select an internal version number (e.g., "3.0.0.30203")
+   The last section is in format "ymmdd" format
+   where y is years since 2018 and then release date in "mmdd"
+   (e.g., Release on January 12th 2022  2018 + 4 years and date in "mmdd" - x.y.z.40112)
```
fabiomadge, line 15:
```suggestion
   and select an internal version number (e.g., "3.0.0.30203").
   The last section is in "ymmdd" format, where "y" is the number of years
   since 2018 and "mmdd" the month and day portions of the release date (e.g.,
   release on January 12th, 2022 -> x.y.z.40112).
```

fabiomadge (approved):
Can you use a more meaningful title when merging this?

PR #1739, opened 2022-01-20
Resolving plugin support

MikaelMayer:
This PR opens up the Dafny Rewriting steps to any external assembly. Namely:

*Dafny pipeline*
* I added in DafnyOptions the option `CompilerBackends`, which takes a list of strings
* In the `Resolver`, when it gather all internal `IRewrite` interfaces, it now loads all the assemblies from `DafnyOptions.O.CompilerBackends` and extract every IRewriter from them, and add them to the list of `rewriters` I fires errors appropriately depending on whether it cannot find the file, the file fails to load, or the file does not have the right version. I'm not sure how to ensure compatibility with version though.
* I added a test in DafnyPipeline.Test that verifies it can load a dummy `IRewriter` (credits to @keyboardDrummer  for the code to do so with the CSharpCompiler which I heavily copied)  and it is correctly during resolution.

*Dafny driver*:
* I added the command option `/compiler:` which takes a comma-separated list of paths to assemblies (I could chose another symbol than comma to separate if preferred), and fills in the option `CompilerBackends` with it.

*Dafny language server*
* To ensure any error in loading such assembly is visible in the VSCode extension, I created a routine that fetches the first non-trivial token, which is used to report the error, rather than `Token.NoToken`
* I added an option to the `DafnyLanguageServer`, namely `--compiler:backends=` which serves the same purpose as the `/compiler:` option of Dafny above. This option can be launched in VSCode by adding `"dafny.languageServerLaunchArgs": ["--compiler:backends=absolute_path_to_resolver.dll"]` to the `settings.json`
* Rather than modifying the settings.json that way, I also updated the VSCode extension so that one can provide this argument in the settings rather the settings.sjon. Will update soon with a link.

### DIFF
--- a/Source/Dafny/DafnyOptions.cs
+++ b/Source/Dafny/DafnyOptions.cs
@@ -137,6 +137,9 @@ namespace Microsoft.Dafny {
     // Working around the fact that xmlFilename is private
     public string BoogieXmlFilename = null;
 
+    // invariant: All strings of this list are non-empty
+    public readonly List<string> CompilerBackends = new();
+
     public virtual TestGenerationOptions TestGenOptions =>
       testGenOptions ??= new TestGenerationOptions();
 
@@ -227,6 +230,16 @@ namespace Microsoft.Dafny {
             int verbosity = 0;
             if (ps.GetNumericArgument(ref verbosity, 2)) {
               CompileVerbose = verbosity == 1;
+            }
+
+            return true;
+          }
+
+        case "compiler": {
+            if (ps.ConfirmArgumentCount(1)) {
+              if (args[ps.i].Length > 0) {
+                CompilerBackends.AddRange(args[ps.i].Split(',').Where(s => s.Length > 0));
+              }
             }
 
             return true;
@@ -804,6 +817,9 @@ namespace Microsoft.Dafny {
     Note that the C++ backend has various limitations (see Docs/Compilation/Cpp.md).
     This includes lack of support for BigIntegers (aka int), most higher order
     functions, and advanced features like traits or co-inductive types.
+/compiler:<assemblies>
+    (experimental) List of comma-separated DLLs or assemblies that contain at least one
+    instantiatable class extending Microsoft.Dafny.IRewriter
 /Main:<name>
     The (fully-qualified) name of the method to use as the executable entry point.
     Default is the method with the {{:main}} atrribute, or else the method named 'Main'.
--- a/Source/Dafny/DafnyOptions.cs
+++ b/Source/Dafny/DafnyOptions.cs
@@ -818,7 +818,7 @@ namespace Microsoft.Dafny {
     This includes lack of support for BigIntegers (aka int), most higher order
     functions, and advanced features like traits or co-inductive types.
 /compiler:<assemblies>
-    (experimental) List of comma-separated DLLs or assemblies that contain at least one
+    (experimental) List of comma-separated paths to assemblies that contain at least one
     instantiatable class extending Microsoft.Dafny.IRewriter
 /Main:<name>
     The (fully-qualified) name of the method to use as the executable entry point.
--- a/Source/Dafny/DafnyOptions.cs
+++ b/Source/Dafny/DafnyOptions.cs
@@ -138,7 +138,7 @@ namespace Microsoft.Dafny {
     public string BoogieXmlFilename = null;
 
     // invariant: All strings of this list are non-empty
-    public readonly List<string> CompilerBackends = new();
+    public List<string> CompilerBackends = new();
 
     public virtual TestGenerationOptions TestGenOptions =>
       testGenOptions ??= new TestGenerationOptions();
--- a/Source/Dafny/Reporting.cs
+++ b/Source/Dafny/Reporting.cs
@@ -153,7 +153,7 @@ namespace Microsoft.Dafny {
   }
 
   public abstract class BatchErrorReporter : ErrorReporter {
-    private readonly Dictionary<ErrorLevel, List<ErrorMessage>> allMessages;
+    protected readonly Dictionary<ErrorLevel, List<ErrorMessage>> allMessages;
 
     protected BatchErrorReporter() {
       ErrorsOnly = false;
--- a/Source/Dafny/Resolver.cs
+++ b/Source/Dafny/Resolver.cs
@@ -442,7 +442,7 @@ namespace Microsoft.Dafny {
         foreach (var assemblyPath in DafnyOptions.O.CompilerBackends) {
           Assembly compilerBackend;
           try {
-            compilerBackend = Assembly.Load(assemblyPath);
+            compilerBackend = Assembly.LoadFrom(assemblyPath);
           } catch (ArgumentException) {
             reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend paths cannot be empty.");
             continue;
--- a/Source/Dafny/Resolver.cs
+++ b/Source/Dafny/Resolver.cs
@@ -439,21 +439,29 @@ namespace Microsoft.Dafny {
       rewriters.Add(new InductionRewriter(reporter));
 
       if (DafnyOptions.O.CompilerBackends.Count() >= 1) {
+        IToken GetFirstTopLevelToken() {
+          return prog.DefaultModuleDef.TopLevelDecls.Select(x => x as ClassDecl)
+            .Where(x => x != null)
+            .SelectMany(classDecl => classDecl.Members)
+            .Where(member => member.tok.line > 0)
+            .Select(member => member.tok).FirstOrDefault(Token.NoToken);
+        }
+
         foreach (var assemblyPath in DafnyOptions.O.CompilerBackends) {
           Assembly compilerBackend;
           try {
             compilerBackend = Assembly.LoadFrom(assemblyPath);
           } catch (ArgumentException) {
-            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend paths cannot be empty.");
+            reporter.Error(MessageSource.Resolver, GetFirstTopLevelToken(), $"Compiler backend paths cannot be empty.");
             continue;
           } catch (FileNotFoundException) {
-            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend '{assemblyPath}' could not be found ");
+            reporter.Error(MessageSource.Resolver, GetFirstTopLevelToken(), $"Compiler backend '{assemblyPath}' could not be found ");
             continue;
           } catch (FileLoadException e) {
-            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend '{assemblyPath}' could not be loaded: {e.Message}");
+            reporter.Error(MessageSource.Resolver, GetFirstTopLevelToken(), $"Compiler backend '{assemblyPath}' could not be loaded: {e.Message}");
             continue;
           } catch (BadImageFormatException e) {
-            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend version of '{assemblyPath}' is incompatible: {e.Message}");
+            reporter.Error(MessageSource.Resolver, GetFirstTopLevelToken(), $"Compiler backend version of '{assemblyPath}' is incompatible: {e.Message}");
             continue;
           }
 
@@ -461,7 +469,7 @@ namespace Microsoft.Dafny {
           foreach (System.Type type in compilerBackend.GetTypes().Where(t => t.IsAssignableTo(typeof(IRewriter)))) {
             var pluginRewriter = Activator.CreateInstance(type, reporter) as IRewriter;
             if (pluginRewriter == null) {
-              reporter.Error(MessageSource.Resolver, Token.NoToken, $"[Internal error] '{type.FullName}' could not be created as an IRewriter");
+              reporter.Error(MessageSource.Resolver, GetFirstTopLevelToken(), $"[Internal error] '{type.FullName}' could not be created as an IRewriter");
             } else {
               rewriters.Add(pluginRewriter);
               oneRewriterInstantiated = true;
--- a/Source/Dafny/Resolver.cs
+++ b/Source/Dafny/Resolver.cs
@@ -11,6 +11,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
 using System.Diagnostics.Contracts;
+using System.IO;
+using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.BaseTypes;
 using Microsoft.Boogie;
@@ -435,6 +437,42 @@ namespace Microsoft.Dafny {
       }
 
       rewriters.Add(new InductionRewriter(reporter));
+
+      if (DafnyOptions.O.CompilerBackends.Count() >= 1) {
+        foreach (var assemblyPath in DafnyOptions.O.CompilerBackends) {
+          Assembly compilerBackend;
+          try {
+            compilerBackend = Assembly.Load(assemblyPath);
+          } catch (ArgumentException) {
+            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend paths cannot be empty.");
+            continue;
+          } catch (FileNotFoundException) {
+            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend '{assemblyPath}' could not be found ");
+            continue;
+          } catch (FileLoadException e) {
+            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend '{assemblyPath}' could not be loaded: {e.Message}");
+            continue;
+          } catch (BadImageFormatException e) {
+            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Compiler backend version of '{assemblyPath}' is incompatible: {e.Message}");
+            continue;
+          }
+
+          var oneRewriterInstantiated = false;
+          foreach (System.Type type in compilerBackend.GetTypes().Where(t => t.IsAssignableTo(typeof(IRewriter)))) {
+            var pluginRewriter = Activator.CreateInstance(type, reporter) as IRewriter;
+            if (pluginRewriter == null) {
+              reporter.Error(MessageSource.Resolver, Token.NoToken, $"[Internal error] '{type.FullName}' could not be created as an IRewriter");
+            } else {
+              rewriters.Add(pluginRewriter);
+              oneRewriterInstantiated = true;
+            }
+          }
+
+          if (!oneRewriterInstantiated) {
+            reporter.Error(MessageSource.Resolver, Token.NoToken, $"Could not find an Microsoft.Dafny.IRewriter in '{assemblyPath}'");
+          }
+        }
+      }
 
       systemNameInfo = RegisterTopLevelDecls(prog.BuiltIns.SystemModule, false);
       prog.CompileModules.Add(prog.BuiltIns.SystemModule);
--- a/Source/Dafny/Rewriter.cs
+++ b/Source/Dafny/Rewriter.cs
@@ -16,11 +16,11 @@ namespace Microsoft.Dafny {
       this.reporter = reporter;
     }
 
-    internal virtual void PreResolve(ModuleDefinition m) {
+    virtual void PreResolve(ModuleDefinition m) {
       Contract.Requires(m != null);
     }
 
-    internal virtual void PostResolve(ModuleDefinition m) {
+    virtual void PostResolve(ModuleDefinition m) {
       Contract.Requires(m != null);
     }
 
--- a/Source/Dafny/Rewriter.cs
+++ b/Source/Dafny/Rewriter.cs
@@ -16,11 +16,11 @@ namespace Microsoft.Dafny {
       this.reporter = reporter;
     }
 
-    virtual void PreResolve(ModuleDefinition m) {
+    public virtual void PreResolve(ModuleDefinition m) {
       Contract.Requires(m != null);
     }
 
-    virtual void PostResolve(ModuleDefinition m) {
+    public virtual void PostResolve(ModuleDefinition m) {
       Contract.Requires(m != null);
     }
 
--- a/Source/Dafny/Rewriter.cs
+++ b/Source/Dafny/Rewriter.cs
@@ -67,7 +67,7 @@ namespace Microsoft.Dafny {
       Contract.Requires(reporter != null);
     }
 
-    internal override void PostResolve(ModuleDefinition m) {
+    public override void PostResolve(ModuleDefinition m) {
       var splitter = new Triggers.QuantifierSplitter();
       foreach (var decl in ModuleDefinition.AllCallables(m.TopLevelDecls)) {
         splitter.Visit(decl);
@@ -82,7 +82,7 @@ namespace Microsoft.Dafny {
       Contract.Requires(reporter != null);
     }
 
-    internal override void PostResolve(ModuleDefinition m) {
+    public override void PostResolve(ModuleDefinition m) {
       var forallvisiter = new ForAllStmtVisitor(reporter);
       foreach (var decl in ModuleDefinition.AllCallables(m.TopLevelDecls)) {
         forallvisiter.Visit(decl, true);
@@ -456,7 +456,7 @@ namespace Microsoft.Dafny {
       this.builtIns = builtIns;
     }
 
-    internal override void PreResolve(ModuleDefinition m) {
+    public override void PreResolve(ModuleDefinition m) {
       foreach (var d in m.TopLevelDecls) {
         bool sayYes = true;
         if (d is ClassDecl && Attributes.ContainsBool(d.Attributes, "autocontracts", ref sayYes) && sayYes) {
@@ -543,7 +543,7 @@ namespace Microsoft.Dafny {
       }
     }
 
-    internal override void PostResolve(ModuleDefinition m) {
+    public override void PostResolve(ModuleDefinition m) {
       foreach (var d in m.TopLevelDecls) {
         bool sayYes = true;
         if (d is ClassDecl && Attributes.ContainsBool(d.Attributes, "autocontracts", ref sayYes) && sayYes) {
@@ -910,7 +910,7 @@ namespace Microsoft.Dafny {
       revealOriginal = new Dictionary<Method, Function>();
     }
 
-    internal override void PreResolve(ModuleDefinition m) {
+    public override void PreResolve(ModuleDefinition m) {
       foreach (var d in m.TopLevelDecls) {
         if (d is TopLevelDeclWithMembers) {
           ProcessOpaqueClassFunctions((TopLevelDeclWithMembers)d);
@@ -918,7 +918,7 @@ namespace Microsoft.Dafny {
       }
     }
 
-    internal override void PostResolve(ModuleDefinition m) {
+    public override void PostResolve(ModuleDefinition m) {
       foreach (var decl in ModuleDefinition.AllCallables(m.TopLevelDecls)) {
         if (decl is Lemma || decl is TwoStateLemma) {
           var lem = (Method)decl;
@@ -1053,7 +1053,7 @@ namespace Microsoft.Dafny {
       Contract.Requires(reporter != null);
     }
 
-    internal override void PostResolve(ModuleDefinition m) {
+    public override void PostResolve(ModuleDefinition m) {
       var components = m.CallGraph.TopologicallySortedComponents();
 
       foreach (var scComponent in components) {  // Visit the call graph bottom up, so anything we call already has its prequisites calculated
@@ -1416,7 +1416,7 @@ namespace Microsoft.Dafny {
       Contract.Requires(reporter != null);
     }
 
-    internal override void PreResolve(ModuleDefinition m) {
+    public override void PreResolve(ModuleDefinition m) {
       var declarations = m.TopLevelDecls;
 
       foreach (var d in declarations) {
@@ -1481,7 +1481,7 @@ namespace Microsoft.Dafny {
       Contract.Requires(reporter != null);
     }
 
-    internal override void PreResolve(ModuleDefinition m) {
+    public override void PreResolve(ModuleDefinition m) {
       foreach (var d in m.TopLevelDecls) {
         if (d is TopLevelDeclWithMembers tld) {
           foreach (MemberDecl member in tld.Members) {
--- a/Source/DafnyLanguageServer/Language/CompilerOptions.cs
+++ b/Source/DafnyLanguageServer/Language/CompilerOptions.cs
@@ -1,0 +1,16 @@
+﻿namespace Microsoft.Dafny.LanguageServer.Language {
+  /// <summary>
+  /// Options for the resolution and compilation pipeline
+  /// </summary>
+  public class CompilerOptions {
+    /// <summary>
+    /// The IConfiguration section of the compiler options.
+    /// </summary>
+    public const string Section = "Compiler";
+
+    /// <summary>
+    /// Gets or sets which backends will be targeted so that their resolver can run before verification.
+    /// </summary>
+    public string Backends { get; set; } = "";
+  }
+}
--- a/Source/DafnyLanguageServer/README.md
+++ b/Source/DafnyLanguageServer/README.md
@@ -75,3 +75,12 @@ Options provided through the command line have higher priority than the options 
 # Default: true (mark the statements)
 --ghost:markStatements=true
 ```
+
+### Compilation Backends
+
+```sh
+# Provides a comma-separated list of paths to assemblies
+# that contain an instantiatable IRewrite that is used during resolution
+# Default: "" (nothing extra is loaded)
+--compiler:backends=
+```
--- a/Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
+++ b/Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
@@ -10,6 +10,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Dafny.LanguageServer.Workspace {
   /// <summary>
@@ -32,6 +33,7 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
     private readonly ICompilationStatusNotificationPublisher notificationPublisher;
     private readonly ILoggerFactory loggerFactory;
     private readonly BlockingCollection<Request> requestQueue = new();
+    private readonly IOptions<CompilerOptions> compilerOptions;
 
     private TextDocumentLoader(
       ILoggerFactory loggerFactory,
@@ -40,7 +42,8 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
       IProgramVerifier verifier,
       ISymbolTableFactory symbolTableFactory,
       IGhostStateDiagnosticCollector ghostStateDiagnosticCollector,
-      ICompilationStatusNotificationPublisher notificationPublisher) {
+      ICompilationStatusNotificationPublisher notificationPublisher,
+      IOptions<CompilerOptions> compilerOptions) {
       this.parser = parser;
       this.symbolResolver = symbolResolver;
       this.verifier = verifier;
@@ -48,6 +51,7 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
       this.ghostStateDiagnosticCollector = ghostStateDiagnosticCollector;
       this.notificationPublisher = notificationPublisher;
       this.loggerFactory = loggerFactory;
+      this.compilerOptions = compilerOptions;
     }
 
     public static TextDocumentLoader Create(
@@ -57,9 +61,10 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
       ISymbolTableFactory symbolTableFactory,
       IGhostStateDiagnosticCollector ghostStateDiagnosticCollector,
       ICompilationStatusNotificationPublisher notificationPublisher,
-      ILoggerFactory loggerFactory
-    ) {
-      var loader = new TextDocumentLoader(loggerFactory, parser, symbolResolver, verifier, symbolTableFactory, ghostStateDiagnosticCollector, notificationPublisher);
+      ILoggerFactory loggerFactory,
+      IOptions<CompilerOptions> compilerOptions
+      ) {
+      var loader = new TextDocumentLoader(loggerFactory, parser, symbolResolver, verifier, symbolTableFactory, ghostStateDiagnosticCollector, notificationPublisher, compilerOptions);
       var loadThread = new Thread(loader.Run, MaxStackSize) { IsBackground = true };
       loadThread.Start();
       return loader;
@@ -111,6 +116,11 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
         notificationPublisher.SendStatusNotification(textDocument, CompilationStatus.ParsingFailed);
         return CreateDocumentWithEmptySymbolTable(loggerFactory.CreateLogger<SymbolTable>(), textDocument, errorReporter, program, loadCanceled: false);
       }
+
+      if (compilerOptions.Value.Backends.Length > 0) {
+        DafnyOptions.O.CompilerBackends = compilerOptions.Value.Backends.Split(",").Where(x => x.Length > 0).ToList();
+      }
+
       var compilationUnit = symbolResolver.ResolveSymbols(textDocument, program, cancellationToken);
       var symbolTable = symbolTableFactory.CreateFrom(program, compilationUnit, cancellationToken);
       if (errorReporter.HasErrors) {
--- a/Source/DafnyPipeline.Test/ExternalResolverTest.cs
+++ b/Source/DafnyPipeline.Test/ExternalResolverTest.cs
@@ -1,0 +1,94 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Xunit;
+using Microsoft.Dafny;
+
+namespace DafnyPipeline.Test;
+
+[Collection("External Resolver plug-in tests")]
+public class ExternalResolverTest {
+  /// <summary>
+  /// This method creates a library and returns the path to that library.
+  /// The library extends an IRewriter so that we can verify that Dafny invokes it if provided in argument.
+  /// </summary>
+  public async Task<string> GetLibrary(string code) {
+    var temp = Path.GetTempFileName();
+    var compilation = CSharpCompilation.Create("tempAssembly");
+    var standardLibraries = new List<string>()
+    {
+      "DafnyPipeline",
+      "System.Runtime",
+      "Boogie.Core"
+    };
+    compilation = compilation.AddReferences(standardLibraries.Select(fileName =>
+        MetadataReference.CreateFromFile(Assembly.Load(fileName).Location)))
+      .AddReferences(
+        MetadataReference.CreateFromFile(typeof(object).GetTypeInfo().Assembly.Location))
+      .WithOptions(
+        new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary)
+      );
+    var syntaxTree = CSharpSyntaxTree.ParseText(code);
+    compilation = compilation.AddSyntaxTrees(syntaxTree);
+    var assemblyPath = $"{temp}.dll";
+    var result = compilation.Emit(assemblyPath);
+
+    var configuration = JsonSerializer.Serialize(
+      new {
+        runtimeOptions = new {
+          tfm = "net6.0",
+          framework = new {
+            name = "Microsoft.NETCore.App",
+            version = "6.0.0",
+            rollForward = "LatestMinor"
+          }
+        }
+      }, new JsonSerializerOptions() { WriteIndented = true });
+    await File.WriteAllTextAsync(temp + ".runtimeconfig.json", configuration + Environment.NewLine);
+
+    Assert.True(result.Success, string.Join("\n", result.Diagnostics.Select(d => d.ToString())));
+    return assemblyPath;
+  }
+
+  class CollectionErrorReporter : BatchErrorReporter {
+    public string GetLastErrorMessage() {
+      return allMessages[ErrorLevel.Error][0].message;
+    }
+  }
+
+  [Fact]
+  public async void EnsureItIsPossibleToPluginIRewriter() {
+    var library = await GetLibrary(@"
+      using Microsoft.Dafny;
+      public class ErrorRewriter: IRewriter {
+        public ErrorRewriter(ErrorReporter reporter) : base(reporter)
+        {}
+
+        public override void PostResolve(ModuleDefinition m) {
+          reporter.Error(MessageSource.Compiler, m.tok, ""Impossible to continue"");
+        }
+      }");
+
+    var reporter = new CollectionErrorReporter();
+    var options = new DafnyOptions(reporter);
+    options.CompilerBackends.Add(library);
+    DafnyOptions.Install(options);
+
+    var programString = "function test(): int { 1 }";
+    ModuleDecl module = new LiteralModuleDecl(new DefaultModuleDecl(), null);
+    Microsoft.Dafny.Type.ResetScopes();
+    BuiltIns builtIns = new BuiltIns();
+    Parser.Parse(programString, "virtual", "virtual", module, builtIns, reporter);
+    var dafnyProgram = new Program("programName", module, builtIns, reporter);
+    Main.Resolve(dafnyProgram, reporter);
+
+    Assert.Equal(1, reporter.Count(ErrorLevel.Error));
+    Assert.Equal("Impossible to continue", reporter.GetLastErrorMessage());
+  }
+}
--- a/Source/Dafny/RefinementTransformer.cs
+++ b/Source/Dafny/RefinementTransformer.cs
@@ -164,7 +164,7 @@ namespace Microsoft.Dafny {
     public ModuleSignature RefinedSig;  // the intention is to use this field only after a successful PreResolve
     private ModuleSignature refinedSigOpened;
 
-    internal override void PreResolve(ModuleDefinition m) {
+    public override void PreResolve(ModuleDefinition m) {
       if (m.RefinementQId?.Decl != null) { // There is a refinement parent and it resolved OK
         RefinedSig = m.RefinementQId.Sig;
 
@@ -485,7 +485,7 @@ namespace Microsoft.Dafny {
       }
     }
 
-    internal override void PostResolve(ModuleDefinition m) {
+    public override void PostResolve(ModuleDefinition m) {
       if (m == moduleUnderConstruction) {
         while (this.postTasks.Count != 0) {
           var a = postTasks.Dequeue();
--- a/Source/DafnyLanguageServer.Test/Unit/TextDocumentLoaderTest.cs
+++ b/Source/DafnyLanguageServer.Test/Unit/TextDocumentLoaderTest.cs
@@ -8,6 +8,7 @@ using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Unit {
   [TestClass]
@@ -20,6 +21,7 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Unit {
     private Mock<ICompilationStatusNotificationPublisher> notificationPublisher;
     private TextDocumentLoader textDocumentLoader;
     private Mock<ILoggerFactory> logger;
+    private Mock<IOptions<CompilerOptions>> compilerOptions;
 
     [TestInitialize]
     public void SetUp() {
@@ -30,6 +32,7 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Unit {
       ghostStateDiagnosticCollector = new();
       notificationPublisher = new();
       logger = new Mock<ILoggerFactory>();
+      compilerOptions = new Mock<IOptions<CompilerOptions>>();
       textDocumentLoader = TextDocumentLoader.Create(
         parser.Object,
         symbolResolver.Object,
@@ -37,7 +40,8 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Unit {
         symbolTableFactory.Object,
         ghostStateDiagnosticCollector.Object,
         notificationPublisher.Object,
-        logger.Object
+        logger.Object,
+        compilerOptions.Object
       );
     }
 
--- a/Source/DafnyLanguageServer/Language/LanguageServerExtensions.cs
+++ b/Source/DafnyLanguageServer/Language/LanguageServerExtensions.cs
@@ -25,6 +25,7 @@ namespace Microsoft.Dafny.LanguageServer.Language {
       return services
         .Configure<VerifierOptions>(configuration.GetSection(VerifierOptions.Section))
         .Configure<GhostOptions>(configuration.GetSection(GhostOptions.Section))
+        .Configure<CompilerOptions>(configuration.GetSection(CompilerOptions.Section))
         .AddSingleton<IDafnyParser>(serviceProvider => DafnyLangParser.Create(serviceProvider.GetRequiredService<ILogger<DafnyLangParser>>()))
         .AddSingleton<ISymbolResolver, DafnyLangSymbolResolver>()
         .AddSingleton<IProgramVerifier>(CreateVerifier)
--- a/Source/DafnyLanguageServer/Workspace/LanguageServerExtensions.cs
+++ b/Source/DafnyLanguageServer/Workspace/LanguageServerExtensions.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using OmniSharp.Extensions.LanguageServer.Server;
 using System;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Dafny.LanguageServer.Workspace {
   /// <summary>
@@ -44,7 +45,8 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
         services.GetRequiredService<ISymbolTableFactory>(),
         services.GetRequiredService<IGhostStateDiagnosticCollector>(),
         services.GetRequiredService<ICompilationStatusNotificationPublisher>(),
-        services.GetRequiredService<ILoggerFactory>()
+        services.GetRequiredService<ILoggerFactory>(),
+        services.GetRequiredService<IOptions<CompilerOptions>>()
       );
     }
   }


cpitclaudel:
Can you say how this PR interacts with https://github.com/dafny-lang/dafny/pull/1708 ?  Does is replace it?

Source/Dafny/DafnyOptions.cs
```diff
@@ -232,6 +235,16 @@ public enum IncludesModes {
             return true;
           }
 
+        case "compiler": {
+            if (ps.ConfirmArgumentCount(1)) {
+              if (args[ps.i].Length > 0) {
+                CompilerBackends.AddRange(args[ps.i].Split(',').Where(s => s.Length > 0));
```
camrein, line 241: Since `args[ps.i]` is used multiple times within this method, I suggest creating a local variable here.

Source/DafnyLanguageServer/Language/CompilerOptions.cs
```diff
@@ -0,0 +1,16 @@
...
+  public class CompilerOptions {
+    /// <summary>
+    /// The IConfiguration section of the compiler options.
+    /// </summary>
+    public const string Section = "Compiler";
+
+    /// <summary>
+    /// Gets or sets which backends will be targeted so that their resolver can run before verification.
+    /// </summary>
+    public string Backends { get; set; } = "";
```
camrein, line 14: .NET Core's configuration utilities support the usage of arrays. However, they're not as comfortable to use from the command line as your implementation. If I remember right, it'd look something like this:
```sh
--compiler:backends:0=<assembly1> --compiler:backends:1=<assembly2> 
```
How often do you expect this feature to be relevant for end-users? As an alternative, users may also provide (any) language server options through `LanguageServer.appsettings.json`.

Source/DafnyLanguageServer/README.md
```diff
@@ -75,3 +75,12 @@ Options provided through the command line have higher priority than the options
...
 --ghost:markStatements=true
 ```
+
+### Compilation Backends
+
+```sh
+# Provides a comma-separated list of paths to assemblies
+# that contain an instantiatable IRewrite that is used during resolution
+# Default: "" (nothing extra is loaded)
+--compiler:backends=
```
camrein, line 85: I didn't come across this case yet. Does the argument work as expected when passed an empty string?

Source/DafnyPipeline.Test/ExternalResolverTest.cs
```diff
@@ -0,0 +1,94 @@
...
+        runtimeOptions = new {
+          tfm = "net6.0",
+          framework = new {
+            name = "Microsoft.NETCore.App",
+            version = "6.0.0",
+            rollForward = "LatestMinor"
+          }
+        }
+      }, new JsonSerializerOptions() { WriteIndented = true });
+    await File.WriteAllTextAsync(temp + ".runtimeconfig.json", configuration + Environment.NewLine);
```
camrein, line 53: Is the `*.runtimeconfig.json` file necessary for dynamically loaded assemblies?

camrein (changes requested):
Looks good so far. Is there a reason you didn't create a test for the language server?

keyboardDrummer:
Does this PR also support letting the language server show errors generated by existing backends, such as the error here? https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Compilers/Compiler-Csharp.cs#L2469

I think we should build one solution that enables error reporting both for existing and custom back-ends. What are your thoughts?

Source/DafnyLanguageServer/Language/CompilerOptions.cs
MikaelMayer: I don't expect most users to actually interact with the DafnyLanguageServer on the command line.
LanguageServer.appsettings.json. is not used to pass options to the Dafny process itself, so I had to add this option to the language server itself so that it initializes the options of Dafny when resolving a document.

Since by default I don't expect to have more than one assembly, it's better to keep one single parameter. It makes it easy as well for the VSCode extension to offer a single field to provide the argument.

Source/DafnyLanguageServer/README.md
MikaelMayer: Yes it does work as expected when passed an empty string.

Source/DafnyPipeline.Test/ExternalResolverTest.cs
MikaelMayer: I don't know, I just copied-pasted the code of @keyboardDrummer . I removed it and it works, thanks for reporting.

camrein: Perfect. I guess you've copied this from @keyboardDrummer's code that test's the shutdown of the language server if the parent process quits. The difference to his implementation, is that you're creating a DLL that's loaded by a running process while his implementation is used to spawn a new process. However, I'm not sure if the `*.runtimeconfig.json` file is required if the application is launched with the `dotnet` command; it sure is when calling the executable directly.

keyboardDrummer (changes requested):
Does this PR also support letting the language server show errors generated by existing backends, such as the error here? https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Compilers/Compiler-Csharp.cs#L2469

I think we should build one solution that enables error reporting both for existing and custom back-ends. What are your thoughts?

robin-aws (changes requested):
Just flagging that I'm especially interested in making sure we're crystal clear on the API here - consider this a ad-hoc mechanism for saying "I'm making myself a required reviewer" :)

MikaelMayer:
> Looks good so far. Is there a reason you didn't create a test for the language server?

No good reason and plenty of good reasons to actually do this. I'm going to work on that. One of the big catch of a plugin is that it can theoretically write to standard output, which messes up the disagnostics for the LSP. For now, I'm thinking of putting a comment on the class that plugins need to extend, because error reporters directly print to the console and we don't want to change that I guess.

> I think we should build one solution that enables error reporting both for existing and custom back-ends. What are your thoughts?

I agree. Since the compilers are going to be refactored in #1708, it's better to wait to delegate that mechanism there, and @cpitclaudel and I thought about using generic "plugins" arguments, from which both the Resolver phase and the Compiler phase are going to pull their extra phases; thus, it will be easy to define for every compiler an IRewriter (used in Resolver) that will take care of doing exactly what you want, and nothing will need to be changed on this PR.

> Just flagging that I'm especially interested in making sure we're crystal clear on the API here - consider this a ad-hoc mechanism for saying "I'm making myself a required reviewer" :)

Copy that, required reviewer ! Here are the pros and the cons of the current approach:

+ One single DLL for rewriters and compilers in the future.
+ One simple class to extend, an error reporter provided, and later the post-resolved tree in a method to override, and where to report errors.
+ Plugins can manipulate the actual Dafny tree directly.
+ Works for both DafnyDriver and DafnyLanguageServer
- It will depend on the version of Dafny itself, especially if we change the AST implementation (having a tool that takes care of parsing itself would have less of this problem). I'll focus on this next so that the plugin and Dafny are aware of version compatibility.
- We offer an API for pre-resolve, which might not be desirable. (we could split the IRewriter)
- We are not offering a way to "prioritize" IRewriters, but this could be added later.
- We rely on the IRewriters to not modify the AST too much so that it would break further verification and compilation. However, in most places, we ensure that expressions are resolved only once, and typed only once.

keyboardDrummer:
> I agree. Since the compilers are going to be refactored in #1708, it's better to wait to delegate that mechanism there

I'd rather do it immediately, which would mean in this PR. We can wait for #1708 to be merged right?

MikaelMayer:
> I feel less plugin surface is better.

Resolvers plugins and compilers plugins have different needs. In the world of plugins, it is very common to offer several ways to extend a program, rather than offering a monolithic interface that needs to be extended for something to be called "a plugin".

MikaelMayer:
> I think we should build one solution that enables error reporting both for existing and custom back-ends. What are your thoughts?

I think this is already the right solution to report both errors for existing and custom backends.
To adapt the existing back-ends to error reporting, we only need to do a few things:
- Instantiate the compiler directly in the options
- Ensure that every compiler defines an IRewriter interface (or implement it), by for example relying on a subclass of ConcreteSyntaxTree that does not create an actual tree, only report errors.
- Recover the IRewriter a compiler defines (either the compiler itself, or one of its fields) and put them into the resolving pipeline.

I'll be happy to work on that, but I won't advise to put it in the same PR. A different PR is better because else, review will become a mess and the second PR clearly depends on the first PR.

amaurremi:
I was able to successfully use this with a custom Dafny compiler, and it works great. Given that different Dafny compiler backends support different features, it's super useful to be able to get immediate IDE feedback while programming rather than having different results in the IDE and in the command line, and having to switch back and forth.
Thank you so much for implementing this! One (non-urgent) request for my particular use case would be to add the ability of passing options to the `dll` file in `"--dafny:plugins=absolute_path_to_plugin.dll"`.

MikaelMayer:
> > Why don't we only use the `ICompiler`/`ICodeGenerator` interface for plugins? I feel less plugin surface is better.

> Resolvers plugins and compilers plugins have different needs. In the world of plugins, it is very common to offer several ways to extend a program, rather than offering a monolithic interface that needs to be extended for something to be called "a plugin

On a second thought, if we want to provide arguments to the plugin on the command line, it would make sense to provide them in only one place, which would be the first time we instantiate "their class" that extends a Plugin class we define.
That way, if parsing of these arguments needs to occur, a plugin wouldn't need to pass them twice.

On a third thought, we could always provide an API for that, meaning a class IDafnyOnInit that, if the plugin extends, will be loaded and called with plugin's arguments the first time the plugin is loaded.

MikaelMayer:
On a fourth thought, I found that it was better to centralize the plugin's interface in a single `Configuration` so that it has control over what it provides to Dafny when requested, and can accept command-line arguments as described in the tests.

MikaelMayer:
Description updated:

* the flag `/plugin:` now accepts both a path and an optional list of space-separated arguments after a comma.
* Plugins can now define an optional subclass of `Configuration` that will be in charge of receiving command-line arguments and explicitly providing `Rewriter`s.

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,56 @@
...
+    assembly = Assembly.LoadFrom(pluginPath);
+
+    System.Type[] rewriterTypes = assembly.GetTypes().Where(t =>
+      t.IsAssignableTo(typeof(Rewriter))).ToArray();
+    // Checks about the plugin to be well-behaved.
+    if (!rewriterTypes.Any()) {
+      throw new Exception($"Plugin {pluginPath} does not contain any Microsoft.Dafny.Plugins.Rewriter");
+    }
+
+    foreach (var configurationType in assembly.GetTypes()
```
camrein, line 38: I suggest to extract the load of the configuration into a (possibly static) method.

Source/Dafny/Resolver.cs
```diff
@@ -436,6 +439,14 @@ enum ValuetypeVariety {
 
       rewriters.Add(new InductionRewriter(reporter));
 
+      if (DafnyOptions.O.Plugins.Count() >= 1) {
```
camrein, line 442: Use `Length` over `Count()`. The latter enumerates all the items to compute the count, unless there are some optimizations underneath (there are some in LINQ that check for certain types).
On the other hand, since it's only a guard for the `foreach` loop on the next line, you can skip this altogether. If there are no items, no iteration will happen.

Source/DafnyLanguageServer.Test/Various/PluginsTest.cs
```diff
@@ -0,0 +1,109 @@
...
+  public override void PostResolve(ModuleDefinition moduleDefinition) {
+    Reporter.Error(MessageSource.Compiler, moduleDefinition.tok, ""Impossible to continue ""+configuration.Argument);
+  }
+}");
+    client = await InitializeClient(options => options.OnPublishDiagnostics(diagnosticReceiver.NotificationReceived));
+  }
+
+  protected override IConfiguration CreateConfiguration() {
+    return new ConfigurationBuilder().AddCommandLine(
+      new[] { "--dafny:plugins:0=" + libraryPath + ",\"because, whatever\"" }).Build();
```
camrein, line 66: String-interpolation might be handy here

Source/DafnyLanguageServer/README.md
```diff
@@ -75,3 +75,13 @@ Options provided through the command line have higher priority than the options
...
+
+### Plugins
+
+```sh
+# Provides a path to assemblies
+# that contain an instantiatable Microsoft.Dafny.Plugins.Configuration to use during resolution
+# Optionally, after a comma, provide an extra list of space-separated arguments.
+# Default: "" (nothing extra is loaded)
+--dafny:plugins:0=
+```
```
camrein, line 87: I have the feeling that this documentation needs to be updated now.

camrein (changes requested):
Looks fantastic. A few minor suggestions.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -4223,6 +4227,19 @@ public class ModuleDefinition : INamedRegion, IAttributeBearingDeclaration {
...
       return true;
     }
+
+    public IToken GetFirstTopLevelToken() {
+      return TopLevelDecls.Select(x => x as ClassDecl)
+        .Where(x => x != null)
+        .SelectMany(classDecl => classDecl.Members)
+        .Where(member => member.tok.line > 0)
+        .Select(member => member.tok)
+        .Concat(TopLevelDecls.Select(x => x as LiteralModuleDecl)
```
camrein, line 4237: `Select(...).Where(...)` can be replaced with `OfType<LiteralModuleDecl>()`.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -4223,6 +4227,19 @@ public class ModuleDefinition : INamedRegion, IAttributeBearingDeclaration {
...
+          .Where(x => x != null)
+          .Select((moduleDecl => moduleDecl.ModuleDef.GetFirstTopLevelToken()))
```
camrein, line 4239: Unnecessary parentheses

Source/Dafny/Reporting.cs
```diff
@@ -153,7 +153,7 @@ public abstract class ErrorReporter {
   }
 
   public abstract class BatchErrorReporter : ErrorReporter {
-    private readonly Dictionary<ErrorLevel, List<ErrorMessage>> allMessages;
+    protected readonly Dictionary<ErrorLevel, List<ErrorMessage>> allMessages;
```
camrein, line 156: This might be something for the style-guide (i.e., `.editorconfig` settings). This `protected readonly` field starts with a lower-case letter, whereas `Rewriter.Reporter` starts with an upper-case one.

Source/DafnyLanguageServer.Test/Various/PluginsAdvancedTest.cs
```diff
@@ -0,0 +1,168 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using System;
+using Microsoft.Dafny.LanguageServer.Language;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
camrein, line 10: Lots of build warnings here.

Source/DafnyLanguageServer.Test/Various/PluginsAdvancedTest.cs
```diff
@@ -0,0 +1,168 @@
...
+  protected override IConfiguration CreateConfiguration() {
+    return new ConfigurationBuilder().AddCommandLine(
+      new[] { $@"--dafny:plugins:0=""{libraryPath},force you""" }).Build();
+  }
+
+  /// <summary>
+  /// This method creates a library and returns the path to that library.
+  /// The library extends a Rewriter so that we can verify that Dafny invokes it if provided in argument.
+  /// </summary>
+  public async Task<string> GetLibrary(string code) {
```
camrein, line 119: The compiler already states it. Either change this method to be truly async by awaiting some statements (I think some of the used Roslyn APIs do have async counterparts), or change the signature to be purely synchronous. I'd opt for the former, if there are suitable APIs.

Source/DafnyLanguageServer.Test/Various/PluginsTest.cs
```diff
@@ -0,0 +1,109 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using System;
+using Microsoft.Dafny.LanguageServer.Language;
```
camrein, line 9: Many compiler warnings/hints here

Source/DafnyLanguageServer.Test/Various/PluginsTest.cs
```diff
@@ -0,0 +1,109 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
```
camrein, line 3: This test-class has lots of similarities to `PluginsAdvancedTest.cs`. It might be handy to share the common parts between these files.

Source/DafnyLanguageServer/Language/DafnyPluginsOptions.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Microsoft.Dafny.LanguageServer.Language {
+  /// <summary>
+  /// Options for the resolution and compilation pipeline
+  /// </summary>
+  public class DafnyPluginsOptions {
+    /// <summary>
+    /// The IConfiguration section of the compiler options.
+    /// </summary>
+    public const string Section = "Dafny";
```
camrein, line 9: Maybe to stick to the naming scheme of the other configuration sections. You may consider changing this section's name to `plugins` and the property/option to `list`.

Source/DafnyPipeline.Test/PluginsTest.cs
```diff
@@ -0,0 +1,127 @@
...
+
+namespace DafnyPipeline.Test;
+
+[Collection("Dafny plug-ins tests")]
+public class PluginsTest {
+  /// <summary>
+  /// This method creates a library and returns the path to that library.
+  /// The library extends a Rewriter so that we can verify that Dafny invokes it if provided in argument.
+  /// </summary>
+  public async Task<string> GetLibrary(string code) {
```
camrein, line 21: Since this is a different test project we may want to double think that. However, I'd suggest to share these common test parts between the compiler and the language server.

Source/DafnyPipeline.Test/PluginsTest.cs
```diff
@@ -0,0 +1,127 @@
...
+}");
+
+    var reporter = new CollectionErrorReporter();
+    var options = new DafnyOptions(reporter);
+    options.Plugins.Add(new Microsoft.Dafny.Plugin(library, new string[] { "ignored arguments" }, reporter));
+    DafnyOptions.Install(options);
+
+    var programString = "function test(): int { 1 }";
+    ModuleDecl module = new LiteralModuleDecl(new DefaultModuleDecl(), null);
+    Microsoft.Dafny.Type.ResetScopes();
```
camrein, line 118: These initialization methods might be suitable for an initialization method marked with the `TestInitialize` attribute. This would make the tests conciser and easier to follow.

camrein (changes requested):

Source/Dafny/Reporting.cs
MikaelMayer: It was already a warning, but fixed, thanks !

Source/DafnyLanguageServer.Test/Various/PluginsAdvancedTest.cs
MikaelMayer: I did not find suitable APIs to do this asynchronously. I think it's fine to do it synchronously, so I changed the code to reflect that

Source/DafnyLanguageServer/Language/DafnyPluginsOptions.cs
MikaelMayer: Other configuration sections do not have options that are lists. I don't think it makes sense to name a field based on its type. Moreover, these are plugins for Dafny core itself, not the DafnyLanguageServer, so the option" --dafny:plugins:0=..." makes more sense than "--plugins:list:0=...". Can you please elaborate?

Source/Dafny/Plugins/Configuration.cs
```diff
@@ -0,0 +1,29 @@
...
+/// A class that plugins should extend, in order to receive plugin-specific command-line arguments
+///
+/// If no class extending Microsoft.Dafny.Plugins.Configuration is defined in the assembly plugin,
+/// Dafny will implicitly load all classes extending Rewriter
+/// providing them with an ErrorReporter as the first and only argument
+/// </summary>
+public abstract class Configuration {
+  /// <summary>
+  /// A Microsoft.Dafny.Plugins.Configuration will be automatically instantiated an arguments
+  /// will be provided to the plugin by the method `ParseArguments``;
```
robin-aws, line 16: This is a bit under-specified: is a plugin intended to document which additional arguments it handles?

Would it make more sense to include arguments in the same way that `/verificationLogger:trx;LogFileName=...` works? That way you don't have to worry about conflicts across plugins. You could share the parsing logic here: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyDriver/BoogieXmlConvertor.cs#L35-L46

Source/Dafny/Plugins/Rewriter.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// </summary>
+    /// <param name="moduleDefinition">A module definition after it
+    /// is resolved, type-checked and SCC/Cyclicity/Recursivity have been performed</param>
+    public virtual void PostCyclicityResolve(ModuleDefinition moduleDefinition) {
+      Contract.Requires(moduleDefinition != null);
+    }
+
+    /// <summary>
+    /// Phase 4/5
+    /// Override this method to obtain the module definition after resolving decreasesResolve
```
robin-aws, line 71: Is this as specific method somewhere? Or a typo? :)

Source/Dafny/Plugins/Rewriter.cs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// <summary>
+    /// Phase 3/5
+    /// Override this method to obtain the module definition after resolution and
+    /// SCC/Cyclicity/Recursivity analysis.
+    /// You can then report errors using reporter.Error (see above)
+    /// </summary>
+    /// <param name="moduleDefinition">A module definition after it
+    /// is resolved, type-checked and SCC/Cyclicity/Recursivity have been performed</param>
```
robin-aws, line 64: Is there somewhere you could link to, to explain "SCC/Cyclicity/Recursivity" a bit?

Source/Dafny/Plugins/Configuration.cs
robin-aws: Now I realize that you ARE accepting lists of plugin-scoped arguments instead already. :) But my format suggestion still stands for more consistency, and being able to pass "key=value" arguments is probably smoother.

Source/Dafny/Plugins/Rewriter.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System.Diagnostics.Contracts;
+
+namespace Microsoft.Dafny.Plugins {
+  /// <summary>
+  /// A class that plugins should extend, in order to provide an extra Rewriter to the pipeline.
+  ///
+  /// If the plugin defines no Configuration, then Dafny will instantiate every sub-class
+  /// of Rewriter from the plugin, providing them with an ErrorReporter as the first and only argument.
+  /// </summary>
+  public abstract class Rewriter {
```
robin-aws, line 10: Thanks for the effort in making the contract as clear as possible here. I think this meets the bar for the feature as long as its marked as experimental. I do think we'll need more of a mechanism to break the build if we change the various AST types in a way that break plugins before we fully commit, though, so good to start thinking about that now.

robin-aws (commented):
Excellent work on the testing. :)

I've expressed any lingering worries but no intention of blocking this awesome (if experimental) feature!

Source/DafnyLanguageServer/Language/DafnyPluginsOptions.cs
camrein: I agree that it's not good to name fields/properties by their type. This one's indeed a hard one. My suggestion was based on the naming scheme that targets a certain feature of Dafny, e.g., `verifier` or `ghost`. That's why I came up with the suggestion of naming the section `plugins` to combine settings related to the plugin feature. Unfortunately, I couldn't come up with a better name than `--plugins:list` in the meantime. But as I noted, I don't like this name either; thus, I'm fine with `--dafny:plugins`, but you may have another idea with my motivation.

Source/Dafny/Plugins/Rewriter.cs
keyboardDrummer, line 10: Is there a need for plugins that aim to support custom compiler back-ends to override anything other than the final `PostResolve(Program program)` ?

I feel the use-case of this PR is underspecified. The title reads "This PR creates an extensible plugin structure, which is primarily used to provide Dafny with extra Rewriters", which I think is more of a how than a why.

My impression from our conversations is that this PR is meant to allow custom back-ends to add additional error reporting. I would think that this only requires adding errors after resolution has completed entirely. Is that right?

Source/DafnyLanguageServer/README.md
```diff
@@ -75,3 +75,18 @@ Options provided through the command line have higher priority than the options
 # Default: true (mark the statements)
 --ghost:markStatements=true
 ```
+
+### Plugins
```
keyboardDrummer, line 79: Could you say something about what you can do with plugins here? I think it might be good for users to know that they can use this to implement custom diagnostics, but not custom refactorings, for example. Maybe something like what is described here: https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/compiler-api-model#diagnostic-apis

Source/DafnyPipeline.Test/PluginsTest.cs
MikaelMayer: I'd suggest to put these test parts in a common file too. However, I don't think it's relevant to put it in DafnyPipeline, and I don't think we should have DafnyLanguageServer.Test depend on DafnyPipeline.Test nor the other way around.
So the only solution would be to create another project, but if it just for the sake of putting this single method, it might be overkill.

Source/DafnyPipeline.Test/PluginsTest.cs
MikaelMayer: I refactored them in a separate method for readability, However, I don't think they should be "TestInitialize" because they have to be called before every single test that use the parsing method.

Source/Dafny/Plugins/Configuration.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Dafny.Plugins;
+
+/// <summary>
+/// A class that plugins should extend, in order to receive plugin-specific command-line arguments
+///
+/// If no class extending Microsoft.Dafny.Plugins.Configuration is defined in the assembly plugin,
+/// Dafny will implicitly load all classes extending Rewriter
+/// providing them with an ErrorReporter as the first and only argument
```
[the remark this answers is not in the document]
MikaelMayer: I synchronized the message with the one of Rewriter because it was more explicit, and used the correct wording.

Source/Dafny/Plugins/Configuration.cs
robin-aws: Okay, fair point that it's more flexible to let each plugin decide how to parse arguments.

Source/Dafny/Plugins/Rewriter.cs
MikaelMayer: There is no stable link for that, since it's not documented anywhere, it's in the Resolver.
Would you prefer me to hide these methods for plugin developers? I think I would prefer to but I wasn't sure.

Here is the plan if I want to do so. I would move the class Rewriter to outside of Plugin, create a class PluginRewriter that extends it, and that would contain a Plugin.Rewriter that developers would extend, which would contain strictly less methods. No change for developers except the removal of these opaque methods.

Source/Dafny/Plugins/Rewriter.cs
MikaelMayer: For that, we need access relevant plugins that we intend to support, indeed. This will be the case for compilers soon.

MikaelMayer: > Is there a need for plugins that aim to support custom compiler back-ends to override anything other than the final PostResolve(Program program) ?

Plugin support is not about just compiler back-ends (see last answer).

> I think is more of a how than a why.

I added a statement after the how to explain why.

> My impression from our conversations is that this PR is meant to allow custom back-ends to add additional error reporting. I would think that this only requires adding errors after resolution has completed entirely. Is that right?

This PR goes beyond our conversation indeed. It opens the door to plugins actually rewriting the AST before it is resolved (see PreResolve()), and with little effort to later support custom back-ends later.
Hence it's not just "custom back-ends", since we could definitely have plugins that do other checks that are not part of any back-end. For example, a plugin that verifies some naming conventions (think of a linter), another that extern methods are tested, another plugin performs some tests on external methods,, another plugin that replaces some macros before resolution... support custom back-ends.

Source/DafnyLanguageServer/README.md
```diff
@@ -79,14 +79,32 @@ Options provided through the command line have higher priority than the options
...
-# Default: "" (nothing extra is loaded)
+# Repeat with --dafny:plugins:0=... --dafny:plugins:1=... for multiple plugins.
 --dafny:plugins:0=example.dll
---dafny:plugins:1=example2.dll,oneArgument
-"--dafny:plugins:1=\"example2.dll,\\\"firstArgument with space\\\" secondArgument\""
+--dafny:plugins:0=example2.dll,oneArgument
+"--dafny:plugins:0=\"example2.dll,\\\"firstArgument with space\\\" secondArgument\""
 ```
+
+#### About plugins
```
keyboardDrummer, line 90: I think it would be good to state somewhere that all of the APIs are likely to change. Through the types of the API such as `Program`, we're exposing a lot of Dafny code that is likely to change.

Source/DafnyLanguageServer/README.md
keyboardDrummer: > custom refactorings or tree-to-tree transformations

With a refactoring I meant enabling extra "Code actions" to be shown in the IDE that trigger a rewrite of source code.

Is there a use-case for doing a tree-to-tree transformation before resolution finishes, if it doesn't enable adding new syntax? If we're not sure then I'd prefer hiding those methods. We can make them as `internal` instead of `public` to hide them.

Source/DafnyLanguageServer/README.md
MikaelMayer: I just added an italicized paragraph at the beginning of this section that explains exactly this.
https://github.com/dafny-lang/dafny/blob/resolving-plugin-support/Source/DafnyLanguageServer/README.md#about-plugins
It should also address one concern of @robin-aws

Source/DafnyLanguageServer/README.md
MikaelMayer: I am not sure about the use case. I prefer to remove the extra API for now until it is requested to us. I splitted Plugin rewriters from internal rewriters now, so that we can be cherry-picking about the API to expose.

keyboardDrummer (approved):
Nice work

Source/Dafny/Plugins/Rewriter.cs
robin-aws: No, that's not the right model. We need to explicitly decide on exactly what parts of the implementation plugins have access to, and at the very least document it clearly. Even better, we should use mechanisms to make sure we don't expose things we don't intend to keep the same in the future (i.e. `public` vs `internal`), and don't accidentally change things we said we wouldn't (i.e. API compatibility tests, which there may be tooling out there for on .NET).

Even with all that, there will still be unfortunate cases where users start relying on something you didn't intend them to (I refer you to the infamous XKCD on that subject: https://xkcd.com/1172/ :) ) and changing things without breaking clients will sometimes be tricky. If you make your API "whatever the users we're aware of are using" you're making it that much harder to scale your resources and far more likely you're going to break people and/or tying your hands for refactoring internals for the better in the future.

robin-aws: </rant> :)

Note I'm not asking for any deeper changes on this PR right now - it will be great to have folks experiment as long as they know the API isn't finalized yet!

Source/DafnyLanguageServer/README.md
robin-aws: It does, and thanks for that!

Shouldn't this be be located somewhere more central though, since you can provide plugins on the command line as well? You can still have a section here about how to configure the language server to use plugins, but link to wherever we have the content below for the details about the plugin interface.

robin-aws (dismissed):

robin-aws:
I left a mild rant for later and a suggestion about documentation, but nothing blocking. Really excited about having this!

Source/DafnyLanguageServer/README.md
MikaelMayer: The primary use case of plugins being to report errors for the language server for now, it currently makes sense to put it in its README.md.
There is a link from DafnyOptions.cs to this README.md
I think that, after implementing compilers on the plugins API, it will be time to migrate this documentation elsewhere, what do you think, @cpitclaudel ?

Source/Dafny/Plugins/Rewriter.cs
MikaelMayer: I totally agree on that !

Source/DafnyLanguageServer.Test/Various/PluginsTest.cs
```diff
@@ -0,0 +1,75 @@
...
+    return "PluginsTest";
+  }
+
+  protected override string[] GetCommandLineArgument() {
+    return new[] { $@"--dafny:plugins:0=""{LibraryPath},\""because\\ whatever\""""" };
+  }
+
+  [TestMethod]
+  public async Task EnsureItIsPossibleToLoadAPluginWithArguments() {
+    // TODO: Need to clean up the plugins from the options after they are set.
```
camrein, line 58: Is this TODO still relevant?

Source/DafnyLanguageServer.Test/Various/PluginsTestBase.cs
```diff
@@ -0,0 +1,75 @@
...
+  protected abstract string GetLibraryCode();
+
+  protected abstract string GetLibraryName();
+
+  protected abstract string[] GetCommandLineArgument();
+
+  public async Task SetUpPlugin() {
+    DiagnosticReceiver = new();
+    LibraryPath = GetLibrary(GetLibraryCode(), GetLibraryName());
+    Monitor.Enter(PluginSync);
```
camrein, line 63: What's the reason for this monitor synchronization? Tests (should) run synchronously by default (other tests would fail miserably if they didn't). Moreover, I believe a `SemaphoreSlim` would be more suitable here.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -71,7 +71,7 @@ public class ShutdownTest {
 
       // Wait for the language server to kill itself by waiting until it closes the output stream.
       await process.StandardOutput.ReadToEndAsync();
-      Thread.Sleep(300); // Give the process some time to die
+      Thread.Sleep(400); // Give the process some time to die
```
camrein, line 74: Use `await Task.Delay(400)` inside asynchronous methods.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -111,6 +116,18 @@ ILoggerFactory loggerFactory
         notificationPublisher.SendStatusNotification(textDocument, CompilationStatus.ParsingFailed);
         return CreateDocumentWithEmptySymbolTable(loggerFactory.CreateLogger<SymbolTable>(), textDocument, errorReporter, program, loadCanceled: false);
       }
+
+      // We load plugins at most once.
```
camrein, line 120: I must have missed this before. Plugin loading should happen somewhere during the initialization of the server; thus, I strongly believe it's the wrong place to do that inside the document loading procedure.
By moving it somewhere to the initialization you no longer have to check if the plugins have been loaded before. While it's not relevant yet, you also won't have to deal with concurrency.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -111,6 +116,18 @@ ILoggerFactory loggerFactory
...
+      if (dafnyPluginsOptions.Value.Plugins.Length > 0 && !DafnyOptions.O.Plugins.Any()) {
+        try {
+          foreach (var pluginPathArgument in dafnyPluginsOptions.Value.Plugins) {
+            DafnyOptions.O.Parse(new[] { "-plugin:" + pluginPathArgument });
+          }
+        } catch (Exception e) {
+          errorReporter.Error(MessageSource.Parser, program.GetFirstTopLevelToken(), "Error while instantiating the plugins:" + e.ToString());
```
camrein, line 127: Maybe log this error?
Calling `ToString()` explicitly should be unnecessary in string-concatenations.

Source/DafnyLanguageServer.Test/Various/PluginsTestBase.cs
```diff
@@ -0,0 +1,75 @@
...
+    var syntaxTree = CSharpSyntaxTree.ParseText(code);
+    compilation = compilation.AddSyntaxTrees(syntaxTree);
+    var assemblyPath = $"{temp}.dll";
+    var result = compilation.Emit(assemblyPath);
+
+    Assert.IsTrue(result.Success, string.Join("\n", result.Diagnostics.Select(d => d.ToString())));
+    return assemblyPath;
+  }
+
+  protected abstract string GetLibraryCode();
```
camrein, line 54: Up to your decision: But these three methods might be suitable as properties.

camrein (changes requested):

Source/DafnyLanguageServer.Test/Various/PluginsTestBase.cs
MikaelMayer: What makes you believe tests should run synchronously by default? I routinely run 8 tests in parallel and the CLI does this too I believe.
The Monitor solves a problem that happens both in the CLI and locally: PluginAdvancedTest and PluginTest both dynamically load a plugin, and DafnyLanguageServer loads plugins only the first time they are provided, so running the tests independently worked, but running them at the same time would prevent the second test to install its plugin.

I don't think `SemaphoreSlim` would be appropriate here because really we only want at most one plugin test to be run at a time during tests, not many. I found this doc here:
https://www.onlinebuff.com/article_understand-monitor-vs-mutex-vs-semaphore-vs-semaphoreslim-onlinebuff_60.html

MikaelMayer: I actually removed the synchronization code since I'm now removing the plugin at the end, and it works now.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
MikaelMayer: Thanks a lot for the tip. I was able to also remove the check that plugins were not loaded yet.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
MikaelMayer: One unrelated problem is that, by moving this code to the initialization, I don't have the reporter anymore to report to the IDE that plugin loading failed though. In the newly moved code, I am using a logger so I can just pass the exception as the first parameter.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -35,6 +36,15 @@ public static class DafnyLanguageServer {
         Action killLanguageServer) {
       var logger = server.GetRequiredService<ILogger<Program>>();
       logger.LogTrace("initializing service");
+      var dafnyPluginsOptions = server.GetRequiredService<IOptions<DafnyPluginsOptions>>();
```
camrein, line 39: I'd prefer having the plugin load code being extracted into a dedicated method.
One usual pattern I apply: If I start grouping statements with empty lines it's a good sign to extract the code into a new method.

Source/DafnyLanguageServer.Test/Various/PluginsTestBase.cs
camrein: > What makes you believe tests should run synchronously by default? I routinely run 8 tests in parallel and the CLI does this too I believe.

I thought it was opt-in. The reason I think that tests would fail, if they'd run indeed in parallel, is that the access to Dafny's compiler core is not mutually exclusive. However there's shared memory because of the static fields; thus we have low level data races and race conditions.

Even though it's now obsolete, I still think it's valuable to answer your question:
The reason why I suggested to not use a monitor lock is that it's rather uncommon to split the acquisition and the release of the lock across multiple methods; that's why the `lock` statement is handy here and should be preferred.
You are right that `SemaphoreSlim` allows multiple acquisitions of permits, and depending on the implementation, effectively allowing multiple threads to enter a critical section. However, this is not a problem if there's only a single permit, making the semaphore a binary semaphore that can be used the same as a dedicated mutex type. Moreover, `SemaphoreSlim` provides asynchronous APIs.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
camrein: That's a pity. I think it'd be useful to somehow display such errors within the IDE. But I suggest leaving this to a separate PR.

camrein (dismissed):

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
MikaelMayer: I did not follow your suggestion and implemented this mechanism right into this PR. Basically, for every load request, after parsing the program, it will fail if there were any errors during loading. I do so because I need a valid token of the program to be able to report the error in the diagnostics window. It works like a charm. See my last commit.

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Dafny.Plugins;
+
+namespace Microsoft.Dafny;
+
+/// <summary>
+/// This class is internal to Dafny. It wraps an assembly and an extracted configuration from this assembly,
```
keyboardDrummer, line 10: > This class is internal to Dafny. 

What do you mean by that? Why not mark the visibility as internal then?

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,82 @@
...
+        funcErrorReporterRewriter(errorReporter)).ToArray();
+    }
+  }
+
+  public static IEnumerable<System.Type> GetConfigurationsTypes(Assembly assembly) {
+    return assembly.GetTypes()
+      .Where(t => t.IsAssignableTo(typeof(Configuration)));
+  }
+
+  public Plugin(string pluginPath, string[] args, ErrorReporter reporter) {
```
keyboardDrummer, line 37: Can you put the constructor immediately below the fields?

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,82 @@
...
+/// The configuration provides the methods to parse command-line arguments and obtain Rewriters 
+/// </summary>
+public class Plugin {
+  public Configuration Configuration;
+
+  private readonly string path;
+  private readonly Assembly assembly;
+  private System.Type[] rewriterTypes;
+
+  class AutomaticConfiguration : Configuration {
```
keyboardDrummer, line 20: Consider using a `record` instead of a class. Why is this a nested class?

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,82 @@
...
+namespace Microsoft.Dafny;
+
+/// <summary>
+/// This class is internal to Dafny. It wraps an assembly and an extracted configuration from this assembly,
+/// The configuration provides the methods to parse command-line arguments and obtain Rewriters 
+/// </summary>
+public class Plugin {
+  public Configuration Configuration;
+
+  private readonly string path;
```
keyboardDrummer, line 16: path to what?

Source/Dafny/Plugin.cs
keyboardDrummer, line 37: Given that the constructor does IO, consider making the constructor private and introducing a `static Plugin Load(...)` method instead to make it clear to the caller that IO work is done.

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,82 @@
...
+    rewriterTypes = CheckPluginForRewriters(assembly, path);
+    Configuration = LoadConfiguration(assembly, args, path, rewriterTypes);
+  }
+
+  private static System.Type[] CheckPluginForRewriters(Assembly assembly, string path) {
+    System.Type[] rewriterTpes = assembly.GetTypes().Where(t =>
+      t.IsAssignableTo(typeof(Rewriter))).ToArray();
+    // Checks about the plugin to be well-behaved.
+    if (!rewriterTpes.Any()) {
+      throw new Exception($"Plugin {path} does not contain any Microsoft.Dafny.Plugins.Rewriter");
```
keyboardDrummer, line 49: Consider using `assembly.Location` and dropping the `path` argument.

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    return rewriterTpes;
+  }
+
+  private static Configuration LoadConfiguration(Assembly assembly, string[] args, string path, System.Type[] rewriterTypes) {
+    Configuration configuration = null;
+    foreach (var configurationType in GetConfigurationsTypes(assembly)) {
+      if (configuration != null) {
+        throw new Exception(
+          $"Only one class should extend Microsoft.Dafny.Plugins.Configuration from {path}. Please remove {configurationType}.");
```
keyboardDrummer, line 60: Consider using assembly.Location and dropping the path argument.

Source/Dafny/Plugins/Configuration.cs
```diff
@@ -0,0 +1,31 @@
...
+/// <summary>
+/// Plugins should define a class that extends Configuration,
+/// in order to receive plugin-specific command-line arguments by overwriting the method `ParseArguments`
+/// It is also used to provide to Dafny a list of Rewriter using the method `GetRewriters`
+///
+/// If the plugin defines no Configuration, then Dafny will instantiate every sub-class
+/// of Rewriter from the plugin, providing them with an ErrorReporter in the constructor
+/// as the first and only argument.
+/// </summary>
+public abstract class Configuration {
```
keyboardDrummer, line 15: Consider naming this `PluginConfiguration`, even though it's already in the `Plugins` namespace

Source/Dafny/Resolver.cs
```diff
@@ -687,6 +694,16 @@ enum ValuetypeVariety {
 
       Type.DisableScopes();
       CheckDupModuleNames(prog);
+
+      foreach (var module in prog.Modules()) {
+        foreach (var r in rewriters) {
```
keyboardDrummer, line 699: I'd prefer avoiding letter single variables. Why not write `rewriter` ?

Source/Dafny/Plugin.cs
MikaelMayer: I removed "is internal to Dafny. It wraps" since most classes in Dafny are never used outside of Dafny but could be in the future.

Source/Dafny/Plugin.cs
MikaelMayer: Renamed to `pluginPath`

Source/Dafny/Resolver.cs
MikaelMayer: I prefer avoiding single letter variables. That was part of the code I just copy-pasted. With the upcoming commit, in every loop of this file that is ` (var r in rewriters)`, I renamed the variable to "rewriter", that sounds better.

Source/Dafny/Plugin.cs
keyboardDrummer: > Nothing else than the class Plugin needs this class which is very short, so it makes sense to keep it as a nested class.

Okay. In my experience working on C# codebases, nested classes are only used when access to private members of the containing class is needed, but that's not the case here. For a class that's used by another class through public members, I'm used to using top-level classes, which compared to nested classes may separate concerns more and reduce indentation. I don't have any hard arguments for this way of using nested classes, so feel free to keep it like it is, but consider this an extra data point ^,^

> Based on your suggestion, I considered using a record instead of a class, but as it is extending an abstract class, as far as I know, this is not possible.

You'd have to change the `abstract class` into an `abstract record`

Source/Dafny/Plugin.cs
```diff
@@ -7,19 +7,30 @@
...
 
-  private readonly string path;
+  private readonly string pluginPath;
   private readonly Assembly assembly;
   private System.Type[] rewriterTypes;
 
-  class AutomaticConfiguration : Configuration {
+  private Plugin(string path, string[] args) {
+    pluginPath = path;
+    assembly = Assembly.LoadFrom(pluginPath);
```
keyboardDrummer, line 22: Nitpick: I would even do this work in the `Load` method and then pass the assembly to the constructor.

Source/Dafny/Rewriter.cs
```diff
@@ -2007,6 +2080,22 @@ class Induction_Visitor : BottomUpVisitor {
       }
     }
   }
+
+  class PluginRewriter : IRewriter {
```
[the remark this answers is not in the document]
MikaelMayer: A Dafny rewriter (IRewriter) can support more phases than a plugin rewriter (which only supports PostResolve)
Therefore, we need to encapsulate any plugin-defined Rewriter into the (more internal) Dafny IRewriter.

Source/DafnyLanguageServer.Test/Various/PluginsAdvancedTest.cs
```diff
@@ -0,0 +1,131 @@
...
+namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various;
+
+[TestClass]
+public class PluginsAdvancedTest : PluginsTestBase {
+  [TestInitialize]
+  public async Task SetUp() {
+    await SetUpPlugin();
+  }
+
+  protected override string LibraryCode =>
```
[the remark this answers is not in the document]
MikaelMayer: Implemented.

Source/Dafny/Rewriter.cs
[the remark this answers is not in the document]
keyboardDrummer: I would have thought a single abstract class with some virtual methods marked as internal and some as public (the ones we want to expose) would have been sufficient. How come it's not?

Source/Dafny/Plugin.cs
MikaelMayer: Great alternative. I like to put arguments to the class too. Note that it's `Assembly.LoadFrom`, not `Assembly.Load`.

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  public static IEnumerable<System.Type> GetConfigurationsTypes(Assembly assembly) {
+    return assembly.GetTypes()
+      .Where(t => t.IsAssignableTo(typeof(PluginConfiguration)));
+  }
+
+  private static System.Type[] CheckPluginForRewriters(Assembly assembly) {
+    System.Type[] rewriterTpes = assembly.GetTypes().Where(t =>
```
camrein, line 38: For the sake of consistency within the same class (single letter identifiers), I'd use `type` instead of `t` as the identifier name.

camrein, line 38: Is it necessary to specify `System.Type[]` explicitly here?

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,76 @@
...
+        throw new Exception($"Could not instantiate a {configurationType} from {assembly.Location}");
+      }
+
+      pluginConfiguration.ParseArguments(args);
+    }
+
+    pluginConfiguration ??= new AutomaticPluginConfiguration(
+      rewriterTypes.Select<System.Type, Func<ErrorReporter, Rewriter>>((System.Type rewriterType) =>
+        (ErrorReporter errorReporter) =>
+          Activator.CreateInstance(rewriterType, new object[] { errorReporter }) as Rewriter).ToArray());
```
camrein, line 69: Use `(Rewriter)Activator.CreateInstance(rewriterType, new object[] { errorReporter })` instead. Alternatively, the LINQ statement `.Cast<Rewriter>()` might be handy too.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -11,9 +12,11 @@
 using OmniSharp.Extensions.LanguageServer.Server;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Dafny.LanguageServer {
   public static class DafnyLanguageServer {
+    public static List<string> LoadErrors = new();
```
camrein, line 19: Expose this list as a property of type `IReadOnlyList<string>`.

camrein (commented):
A bit more nitpicking from my side.

keyboardDrummer (dismissed):

Source/Dafny/Rewriter.cs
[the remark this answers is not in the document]
MikaelMayer: It would have been sufficient. However, it may not be a good practice for a plugin system because all the "internal" methods would still be in the Plugin namespace, exposed to anyone looking at the source code - which I expect from plugins writers since they need to have their plugin in sync with the Dafny source code.
I prefer to really split what we offer between two things, one for the plugin, and one for internal. That way, if we want to change how the internal work, we are not stuck by a public API that we can always re-route to the new constructs.

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.Dafny.Plugins;
+
+namespace Microsoft.Dafny;
+
+/// <summary>
+/// This class wraps an assembly and an extracted configuration from this assembly,
+/// The configuration provides the methods to parse command-line arguments and obtain Rewriters 
+/// </summary>
+public record Plugin(Assembly Assembly, string[] Args) {
+  public PluginConfiguration PluginConfiguration { get; init; } = LoadConfiguration(Assembly, Args);
```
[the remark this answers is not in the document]
MikaelMayer: Thanks for the tip ! I did not know what this init was doing, so your explanation was very useful.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
MikaelMayer: Then I have to assign the whole property at once in `LoadPlugins`, which mean I can't make `LoadErrors` readonly. Is that ok?

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,76 @@
...
+
+      if (pluginConfiguration == null) {
+        throw new Exception($"Could not instantiate a {configurationType} from {assembly.Location}");
+      }
+
+      pluginConfiguration.ParseArguments(args);
+    }
+
+    pluginConfiguration ??= new AutomaticPluginConfiguration(
+      rewriterTypes.Select<System.Type, Func<ErrorReporter, Rewriter>>((System.Type rewriterType) =>
```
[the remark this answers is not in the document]
MikaelMayer: I tried to remove it several times (because the IDE suggests it to me), but every time I do, dotnet compiler says it needs that type.

Source/Dafny/Plugin.cs
MikaelMayer: I'm going to revert the use of "abstract record" for PluginConfiguration back to an "abstract class", after testing its usability.
Since the goal is for developers to extend the class PluginConfiguration, if I mark it as an abstract record, they will be forced to implement internal methods such as `PluginConfiguration.<Clone>$()`, and the other goodies that record routinely offer, that abstract record actually require. It would be fine if they developed their plugins in c#, but the dotnet ecosystem has many other languages which do not know about records (e.g. F#).

Source/DafnyLanguageServer/DafnyLanguageServer.cs
camrein: You can implement the property so it uses a backing field. Making the external API readonly, but the internal mutable. For example you can do this:
```cs
private static List<string> loadErrors = new();
public static IReadOnlyList<string> LoadErrors => loadErrors;
```

keyboardDrummer (dismissed):

Source/Dafny/Plugin.cs
keyboardDrummer: Makes sense, nice UX research :)

Source/Dafny/Plugin.cs
```diff
@@ -0,0 +1,74 @@
...
+      if (pluginConfiguration == null) {
+        throw new Exception($"Could not instantiate a {configurationType} from {assembly.Location}");
+      }
+
+      pluginConfiguration.ParseArguments(args);
+    }
+
+    pluginConfiguration ??= new AutomaticPluginConfiguration {
+      Rewriters =
+      rewriterTypes.Select<System.Type, Func<ErrorReporter, Rewriter>>((System.Type rewriterType) =>
```
camrein, line 64: The explicitly typed lambda expressions should be redundant due to the type arguments to `Select`.
However, the requirement for the explicit type arguments might be a good sign that it's best to extract the projection into a separate method in terms of readability:
```cs
    pluginConfiguration ??= new AutomaticPluginConfiguration {
      Rewriters = rewriterTypes.Select(CreateRewriterFactory).ToArray()
    };
    return pluginConfiguration;
  }

  private static Func<ErrorReporter, Rewriter> CreateRewriterFactory(System.Type rewriterType) {
    return errorReporter => (Rewriter)Activator.CreateInstance(rewriterType, new object[] { errorReporter });
  }
```

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -11,9 +12,11 @@
 using OmniSharp.Extensions.LanguageServer.Server;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Dafny.LanguageServer {
   public static class DafnyLanguageServer {
+    public static IReadOnlyList<string> LoadErrors = new List<string>();
```
camrein, line 19: Just to make sure that you didn't miss a small detail in my suggestion. The implementation I proposed was an expression-bodied getter-only property:
```cs
private static List<string> loadErrors = new();
public static IReadOnlyList<string> LoadErrors => loadErrors; // Mind the arrow here
```
This will effectively compile into something like this:
```cs
private static List<string> loadErrors = new();
public static IReadOnlyList<string> LoadErrors  {
  get {
    return loadErrors;
  }
}
```

The current implementation allows changing the value of `LoadErrors` from outside the `DafnyLanguageServer` class which I'd like to avoid.

camrein, line 19: Nitpicking as of now, but I'd like to avoid (mutable) static fields within the language server as much as possible.
However, we may want to consider introducing a global state object for the language server in the future for such cases.

camrein (changes requested):
Some nitpicking from my side

MikaelMayer:
I just enabled auto-squash-and-merge. @camrein  let me know if there is anything else I should change, else you can approve this PR.

camrein (approved):
Fantastic work!

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -71,7 +71,7 @@ public class ShutdownTest {
 
       // Wait for the language server to kill itself by waiting until it closes the output stream.
       await process.StandardOutput.ReadToEndAsync();
-      Thread.Sleep(300); // Give the process some time to die
+      await Task.Delay(400); // Give the process some time to die
```
cpitclaudel, line 74: Why not `WaitForExitAsync`?

Source/DafnyLanguageServer.Test/Various/PluginsTestBase.cs
```diff
@@ -0,0 +1,75 @@
...
+  protected abstract string[] CommandLineArgument { get; }
+
+  public async Task SetUpPlugin() {
+    DiagnosticReceiver = new();
+    LibraryPath = GetLibrary(LibraryName);
+    Client = await InitializeClient(options => options.OnPublishDiagnostics(DiagnosticReceiver.NotificationReceived));
+  }
+
+  protected void CleanupPlugin() {
+    DafnyOptions.O.Plugins.RemoveAt(0);
```
cpitclaudel, line 69: What does this part do?

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
MikaelMayer: I don't know what this does but sometimes a test fail because the process did not die.
@keyboardDrummer ?

Source/DafnyLanguageServer.Test/Various/PluginsTestBase.cs
MikaelMayer: When testing a plugin, it adds it to the DafnyOptions.
I remove it from the DafnyOptions after the a test class is finished.

PR #1751, opened 2022-01-21
fix: Move concrete value substitution from test generation to counterexample extraction

Dargones:
Until now, the counterexample extraction tool would return a placeholder symbol (`?#0`, `?#1`, etc.) if the solver did not specify a concrete value for some variable of basic type. The test generation module contains code that substitutes concrete values for these placeholder symbols. This PR proposes to move this code to counterexample extraction stage, which would make counterexamples more readable. 

Here is a simple example. Consider this failing assertion:
```Dafny
method test(c:char) {
    assert c == 'c';
}
```

Previously, the counterexample extraction tool would return `c:char = ?#0`. Now it will return `c:char = '!'`.


Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
```diff
@@ -948,10 +948,10 @@ method T_map0(m:map<int,int>, key:int, val:int)
     [TestMethod]
     public async Task MapsValues() {
       var source = @"
-      method test(m:map<int,char>) {
-       var values := m.Values;
-       assert ('c' !in values);
-      }".TrimStart();
+method test(m:map<int,char>) {
```
rakamaric, line 951: The indentation looks weird here. Can you please double-check?

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -458,34 +513,65 @@ public class DafnyModel {
...
+          reservedValuesMap[element] = "true";
+        } else {
+          reservedValuesMap[element] = "false";
+        }
+        return reservedValuesMap[element];
+      }
+
+      var i = 0;
+      if (primitiveType == "char") {
+        i = 33; // 33 is the first non special character (excluding space)
```
rakamaric, line 556: I wonder if we should maybe be a bit more user friendly, and first try to inject some kind of more reasonable characters, such as letters. Not a big deal, just something to maybe think about.

Test/server/counterexample_commandline.dfy.expect
```diff
@@ -4,21 +4,21 @@ counterexample_commandline.dfy(17,22): Related location: This is the postconditi
 Dafny program verifier finished with 1 verified, 1 error
 Counterexample for first failing assertion: 
 counterexample_commandline.dfy(20,8): initial state:
-	s : seq<char> = [?#0]
+	s : seq<char> = ['!']
```
rakamaric, line 7: Do you have regressions that test other types, such as boolean, integer, etc.?
Or maybe that is not important for some reason I am maybe missing?

rakamaric (changes requested):

Test/server/counterexample_commandline.dfy.expect
Dargones: There are XUnit tests for other types (`ArbitraryBitVector`, `ArbitraryBool`, etc. in [CounterExampleTest.cs](https://github.com/Dargones/dafny/blob/282b02ec488177875d2f20772cca90ba96f9d3d0/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs)).

Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
Dargones: Christoph asked me to keep the indentation this way for continuity reasons (here is a [link to a comment about it](https://github.com/dafny-lang/dafny/pull/1336#discussion_r684984964)) earlier. That said, the majority of the tests in this file are now written for the new versions of the tool, so perhaps it makes sense to fix the indentation. Let me know and I will make the fix!

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Now starting from character `'A'` - will fix this in the paper along with the badges once we get the instructions on how to add them.

keyboardDrummer (dismissed):

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -458,34 +513,65 @@ public class DafnyModel {
           utfCode = ((Model.Integer)fCharToInt.OptEval(elt)).AsInt();
           return "'" + char.ConvertFromUtf32(utfCode) + "'";
         }
-        return "?#" + GetShortElementId(elt);
+        return GetUnspecifiedValue(elt, "char");
```
rakamaric, line 516: Sorry, I just noticed this. Does Dafny have some kind of an enum for (primitive) types or something like that?
It is kind of not ideal to be passing strings here. Just curious.

rakamaric (changes requested):
I just have a quick question that could lead to an improvement.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Done - I am now using instances of Dafny.Type for this! Is a bit tricky with bit vectors, since they have custom bases, but I believe I made it work.

rakamaric (approved):
Looks good to me.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -458,34 +527,64 @@ public class DafnyModel {
...
 
-    private int GetShortElementId(Model.Element element) {
-      if (!shortElementIds.ContainsKey(element.Id)) {
-        shortElementIds[element.Id] = shortElementIds.Count;
+    private string GetUnspecifiedValue(Model.Element element, Type primitiveType) {
+      if (reservedValuesMap.ContainsKey(element)) {
+        return reservedValuesMap[element];
+      }
+      if (primitiveType == Type.Bool) {
+        if (!reservedValues[Type.Bool].Contains("true")) {
```
camrein, line 558: You may want to create constants for these magic numbers.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -154,6 +160,64 @@ public class DafnyModel {
...
+      foreach (var app in fCharToInt.Apps) {
+        var UTFCode = int.Parse(((Model.Integer)app.Result).Numeral);
+        reservedValues[Type.Char].Add(Convert.ToChar(UTFCode).ToString());
+      }
+
+      var fU2Int = Model.MkFunc("U_2_int", 1);
+      reservedValues[Type.Int] = new();
+      foreach (var app in fU2Int.Apps) {
+        // this skips negative values
+        if (app.Result is Model.Integer) {
```
camrein, line 179: C#'s pattern matching allows declaring a variable inline, making the cast on the following redundant, for example:
```cs
if (app.Result is Model.Integer integerResult) {
  reservedValues[Type.Int].Add(integerResult.Numeral);
}
```

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -154,6 +160,64 @@ public class DafnyModel {
       return result;
     }
 
+    /// <summary>
+    /// Registered all values of basic types specified by the model in
+    /// the reservedValues map;
+    /// </summary>
+    private void RegisterReservedValues() {
```
camrein, line 167: My personal preference would be extracting the steps of this method into distinct methods, for example:

- RegisterCharType
- RegisterIntType
- RegisterBoolType
- etc.

But current style is already common within the compiler.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -458,34 +527,64 @@ public class DafnyModel {
...
+        if (!reservedValues[Type.Bool].Contains("true")) {
+          reservedValues[Type.Bool].Add("true");
+          reservedValuesMap[element] = "true";
+        } else {
+          reservedValuesMap[element] = "false";
+        }
+        return reservedValuesMap[element];
+      }
+
+      var i = 0;
```
camrein, line 567: Personal preference, but rather common: Declare primitive types such as `int` explicitly to not be surprised by type inferences.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -458,34 +527,64 @@ public class DafnyModel {
...
+      if (primitiveType == Type.Char) {
```
camrein, line 568: For me it's hard to follow what these sections are intended to do without studying the code line by line. I suggest extracting these parts into separate methods to make their intent clear. That helps anyone else getting easier a big picture as well as checking if the small method really does what it's meant to do.

camrein (commented):
The code looks good. I've added a few notes about my personal preference that you're free to include or not.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Thank you for catching this! I have made the change.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Made this change for all variable of primitive types in the CounterExample Extraction code

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: I have split this method in three and added a docstring to each, so the code should be more readable

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Done!

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Removed all magic numbers for the method in my most recent commit.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -458,34 +540,84 @@ public class DafnyModel {
...
+      return reservedValuesMap[element];
+    }
+
+    /// <summary>
+    /// Find a bool value that is different from any other value
+    /// of that type in the entire model (if possible).
+    /// Reserve that value for given element
+    /// </summary>
+    private string GetUnreservedBoolValue(Model.Element element) {
+      if (reservedValuesMap.ContainsKey(element)) {
```
camrein, line 589: `reservedValuesMap.TryGetValue(element, out var reservedValue)` might be handy here. Should work for all other `GetUnreserved*` methods too.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -154,6 +171,66 @@ public class DafnyModel {
...
+        if ((app.Result is Model.Real) && int.TryParse(valueAsString, out int value)) {
+          reservedNumerals[Type.Real].Add(value);
+        }
+      }
+    }
+
+    /// <summary> Registered all bv values specified by the model </summary>
+    private void RegisterReservedBitVectors() {
+      foreach (var func in Model.Functions) {
+        if (!Regex.IsMatch(func.Name, "^U_2_bv[0-9]+$")) {
```
camrein, line 213: You may want to [pre-compile](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.-ctor?view=net-6.0#system-text-regularexpressions-regex-ctor(system-string)) this regular expression outside the loop, or even as a static field.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -458,34 +540,84 @@ public class DafnyModel {
...
+      return reservedValuesMap[element];
+    }
+
+    /// <summary>
+    /// Find a value of the given numericType that is different from
+    /// any other value of that type in the entire model.
+    /// Reserve that value for given element
+    /// </summary>
+    private string GetUnreservedNumericValue(Model.Element element, Type numericType) {
+      if (reservedValuesMap.ContainsKey(element)) {
```
camrein, line 607: I see the differences so it's probably unjustified, but thought I'd leave a note regardless. The methods `GetUnreservedNumericValue` and `GetUnreservedCharValue` look very similar at first look.

camrein (changes requested):
Your changes are phenomenal. It helped me to spot one or two more things. 👍

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: This does look much nicer! Thank you, changed this in the last commit

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: I now pre-compile regular expressions here and in other places throughout the counterexample generation code.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: The methods are, indeed, similar but merging them together would undermine readability, since there are a few subtle differences, so I would keep the two separate.

camrein (approved):
Looks good!
Just out of curiosity: What's the reason that some of the regex patterns are declared as fields while others are declared as locals?

Dargones:
Thank you, @camrein! I tried to put the regex patterns closer to the point where they are used, so I declare them as locals when they are only used in one method and as fields otherwise.

Dargones:
@keyboardDrummer, would it be possible to merge this PR? We are presenting Dafny's counterexamples at TACAS next week and it would be great to have this fix implemented!

PR #1645, opened 2021-12-14
feat: `DowncastCopy` for data types in C#

fabiomadge:
This is a follow-up to #1578, which omitted to generate `DowncastCopy` for data types. We add it here, provided there is no contravariant type parameter. To enable testing, we increase the type checker’s eagerness, allowing it to accept contravariant downcasts of data types.

PR #1755, opened 2022-01-22
feat: `DowncastClone` for functions and data types with contravariant type parameters

fabiomadge:
This is a follow-up to #1645, which omitted to generate `DowncastCopy` for data types with contravariant type parameters.  To make this useful we also add it for functions. We still don't generate `DowncastCopy`, when a `datatype` has a constructor expecting an argument of a type that contains a generic reference type reusing the `datatype`'s type parameters.
```Dafny
datatype GenDifficult<G> = MakeGenDifficult(GenClass?<G>)
class GenClass<H> { }
```
This shouldn't be a real limitation in practice, as the nonvariant nature of the reference type's parameters prevents any casts.


Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -124,6 +124,46 @@ public CsharpCompiler(ErrorReporter reporter)
           w.WriteLine("return a;");
         }
       }
+
+      //Generates casts for functions of those arities present in the program
+      var funcExtension = wr.NewNamedBlock("public static class FuncExtension");
```
keyboardDrummer, line 129: I think the common name for classes containing extension function is `<X>Extensions`, with the `s` at the end.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -124,6 +124,46 @@ public CsharpCompiler(ErrorReporter reporter)
           w.WriteLine("return a;");
         }
       }
+
+      //Generates casts for functions of those arities present in the program
```
keyboardDrummer, line 128: Could you put the block that the comment pertains to in a separate method?

Could you also show put an example of a generated extension function in a comment?

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -490,30 +530,66 @@ public CsharpCompiler(ErrorReporter reporter)
...
+        wr.Write($"public {modifiers}");
       }
 
-      wr.Write($"{dtArgs} DowncastClone{typeArgs}({nonGhostTypeArgs.Comma(PrintConverter)})");
+      if (!(toInterface && customReceiver)) {
+        var receiver = customReceiver ? $"{"_I" + datatype.CompileName + typeArgs} _this, " : "";
+        wr.Write($"{dtArgs} DowncastClone{uTypeArgs}({receiver}{nonGhostTypeArgs.Comma(PrintConverter)})");
+      }
 
       if (ctor == null && !lazy) {
```
keyboardDrummer, line 559: It seems that there's two big blocks, one for `ctor == null && !lazy` and the other for not that. Can we name those blocks and put them in separate methods?

keyboardDrummer: What overhead are you referring to?

I'm suggesting something like:

```
if (ctor == null && !lazy) {
    ComeUpWithAName(datatype, wr, toInterface, customReceiver, uTypeArgs, typeArgs, converters, lazyClass);
    return;
  }

  ComeUpWithAnotherName(datatype, wr, nonGhostTypeArgs, lazy, ctor, customReceiver, uTypeArgs, typeArgs, dtArgs, converters, lazyClass);
 ```

File: Source/Dafny/Compilers/Compiler-Csharp.cs [hunk 4]
```diff
@@ -92,6 +92,62 @@ public CsharpCompiler(ErrorReporter reporter)
...
+          return l;
+        }
+
+        string TPs(IEnumerable<string> l) => $"<{l.Comma()}>";
+
+        var us = TyList("U");
+        var ts = TyList("T");
+        var tsus = ts.Concat(us).ToList();
+
+        string ArgConv((string u, string t) tp) => $"Func<{tp.u}, {tp.t}> ArgConv";
```
keyboardDrummer, line 129: It doesn't really matter since this is not likely to be read by someone, but C# function parameters generally start with a lowerCase.

keyboardDrummer: Also, I prefer having two separate function arguments over a tuple. I'd rather deconstruct the tuple at the call-site.

fabiomadge: I couldn't figure out a nice way to do that in the lambda and the autogenerated names `First` and `Second` are cumbersome. Re the capitalisation: You are right, that appears to be the standard, but it feels wrong, as regular functions a capitalized.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -92,6 +92,62 @@ public CsharpCompiler(ErrorReporter reporter)
...
+        List<string> TyList(string prefix) {
+          var l = arity switch {
+            1 => new List<string> { prefix },
+            _ => Enumerable.Range(1, arity).Select(i => $"{prefix}{i}").ToList()
+          };
+          l.Add($"{prefix}Result");
+          return l;
+        }
+
+        string TPs(IEnumerable<string> l) => $"<{l.Comma()}>";
```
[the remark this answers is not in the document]
fabiomadge: It's about reusing, not explaining in this instance.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -92,6 +92,62 @@ public CsharpCompiler(ErrorReporter reporter)
...
+
+        var us = TyList("U");
```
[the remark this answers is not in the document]
fabiomadge: It's a plural s, as in many Us.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -92,6 +92,62 @@ public CsharpCompiler(ErrorReporter reporter)
...
+        var ts = TyList("T");
+        var tsus = ts.Concat(us).ToList();
+
+        string ArgConv((string u, string t) tp) => $"Func<{tp.u}, {tp.t}> ArgConv";
+        var argConvs = arity switch {
```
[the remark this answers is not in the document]
fabiomadge: I honestly feel like that's reasonably clear.

Source/Dafny/Compilers/Compiler-Csharp.cs
fabiomadge: The one hidden in the enormous number of arguments. When you want to understand what any of those mean, you need to jump back to the caller.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -92,6 +92,62 @@ public CsharpCompiler(ErrorReporter reporter)
...
+    //     ...
+    //   }
+    // They aren't in any namespace to make them universally available.
+    private static void EmitFuncExtensions(BuiltIns builtIns, ConcreteSyntaxTree wr) {
+      var funcExtensions = wr.NewNamedBlock("public static class FuncExtensions");
+      foreach (var kv in builtIns.ArrowTypeDecls) {
+        int arity = kv.Key;
+
+        List<string> TyList(string prefix) {
+          var l = arity switch {
```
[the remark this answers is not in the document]
fabiomadge: I did go with `list` though, as one of the elements contains the string _Result_.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -92,55 +92,34 @@ public CsharpCompiler(ErrorReporter reporter)
...
+    //   }
+    // They aren't in any namespace to make them universally accessible.
+    private static void EmitFuncExtensions(BuiltIns builtIns, ConcreteSyntaxTree wr) {
+      var funcExtensions = wr.NewNamedBlock("public static class FuncExtensions");
       foreach (var kv in builtIns.ArrowTypeDecls) {
         int arity = kv.Key;
 
         List<string> TyList(string prefix) {
-          var l = arity switch {
+          var list = arity switch {
```
keyboardDrummer, line 115: `list` to me is similar to `l` since it doesn't tell me what this variable contains except for what type is it. `result` is I believe a commonly used variable name for the result of a method so I'm not too worried about the confusion with the `{prefix}Result` elements it might contain. Alternatively, you could also give it the name of the method `tyList`.

Source/Dafny/Compilers/Compiler-Csharp.cs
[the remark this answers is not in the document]
keyboardDrummer: I'm ok with keeping it like this, and thanks for the name changes so far, but personally I wouldn't use variable names like this. I'm able to read this code because of the comment at the top of the method, so I know what the result should be and from that I can determine that every `u` relates to the output function and everything `t` relates to the input function, but I think ideally you'd be able to read the code without that comment.

I'd name this something like `inputFunctionTypes` and then indeed the interpolated string would become bigger: `public static Func<{inputFunctionTypes.Comma()}>`, but I'm OK with that.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -92,55 +92,34 @@ public CsharpCompiler(ErrorReporter reporter)
...
+    //     }
+    //     ...
+    //   }
+    // They aren't in any namespace to make them universally accessible.
+    private static void EmitFuncExtensions(BuiltIns builtIns, ConcreteSyntaxTree wr) {
+      var funcExtensions = wr.NewNamedBlock("public static class FuncExtensions");
       foreach (var kv in builtIns.ArrowTypeDecls) {
         int arity = kv.Key;
 
         List<string> TyList(string prefix) {
```
keyboardDrummer, line 114: Consider renaming this to `FunctionTypes` or `FunctionTypeArguments`

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -92,55 +92,34 @@ public CsharpCompiler(ErrorReporter reporter)
...
-          return l;
+          list.Add($"{prefix}Result");
+          return list;
         }
 
-        string TPs(IEnumerable<string> l) => $"<{l.Comma()}>";
-
         var us = TyList("U");
         var ts = TyList("T");
         var tsus = ts.Concat(us).ToList();
```
keyboardDrummer, line 125: Can we leave out the `ToList` call?  Also I would inline `tsus` since it's only used once and the variable name is harder to read than the contents.

keyboardDrummer (approved):

Source/Dafny/Compilers/Compiler-Csharp.cs
keyboardDrummer: > When you want to understand what any of those mean, you need to jump back to the caller.

Why? Can't you name the arguments and the method so you don't need to jump back? I generally don't need to inspect the caller(s) of a method to understand what values the parameters of a method might contain.

Also, could you explain to me in words what these two cases are. Suppose you would create those two methods, what would the names be?

Also, as a guideline for method size, I like to keep it small enough so it fits on one screen.

Source/Dafny/Compilers/Compiler-Csharp.cs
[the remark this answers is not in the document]
keyboardDrummer: I see. Is the purpose of such reuse then to improve maintainability, so that if say C#'s `<>` syntax is changed, we only need to update `TPs` and not all the locations where the `<>` string occurs? 

I think in the general case, nameless variables/methods that exist only for the purpose of reuse, do not improve maintainability. Suppose I have a widely used method called X and at some point I discover that one of the callers of X needs X to do something slightly different, so I change `X`. However, now I'm not sure whether the other callers should still call `X` or a copy of the old `X`. Without having name that explains what a thing does/is, it's difficult to know what the contents should be at any point in time.

Source/Dafny/Compilers/Compiler-Csharp.cs
fabiomadge: I guess that's where we differ. For me the truth is in the definition.

In one case we generate the actual `DowncastClone`, in the other only a signature. In some special circumstance the signature becomes a jump table.

Source/Dafny/Compilers/Compiler-Csharp.cs
fabiomadge: I do know that it returns a list and that there only is one.

Source/Dafny/Compilers/Compiler-Csharp.cs
fabiomadge: What are you referring to?

Source/Dafny/Compilers/Compiler-Csharp.cs
fabiomadge: If I inline it, I can skip it. Just another reason to do it.

Source/Dafny/Compilers/Compiler-Csharp.cs
keyboardDrummer: > I guess that's where we differ. For me the truth is in the definition.

Isn't the main point of naming things so you don't need to read their definition? For example, if I see a method body like this:

```
var program = Parse(input);
var resolvedProgram = Resolve(program);
var output = CodeGen(resolvedProgram);
```

Then I have some assumption on what the three called methods do, even if I don't dive into them, which will likely make it easier for me to navigate to whichever code I'm interested in reading more about. If the three methods had been inlined, then I would have many more lines to inspect before knowing where to go.

Source/Dafny/Compilers/Compiler-Csharp.cs
keyboardDrummer: to `TyList`

Source/Dafny/Compilers/Compiler-Csharp.cs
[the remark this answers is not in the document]
keyboardDrummer: Well there's different places where argument conversion occurs right? There are argument conversion parameters and there are calls to those parameters. I wasn't sure which of those two the variable `argConvs` was for.

Source/Dafny/Compilers/Compiler-Csharp.cs
fabiomadge: Yes, that would be ideal, but generating such a name is not trivial in general.

RustanLeino (approved):
In trying this out, I found a different bug (a missing cast) related to type conversions, but I'll file that as a separate issue that can be fixed in a separate PR.

fabiomadge:
This is probably tracked in #1757.

Source/Dafny/Compilers/Compiler-Csharp.cs
robin-aws: I fully agree with @keyboardDrummer's points here. Breaking things up into smaller methods has a lot of advantages: the cognitive load is lower because you have to hold less information in your head at once to understand it, but also compilers/runtimes tend to bias towards making many small methods fast as well. Having a correct implementation is great, but it's far better to have a correct implementation that can be efficiently maintained and expanded by future humans. The Dafny code base isn't necessarily a shining example everywhere as lots of older code didn't go through the review process we're now imposing, so we have some catching up to do.

However, in this exact instance I wouldn't insist on the refactoring in this PR. I think it's introducing scope creep, and since the poor organization was already there and isn't being introduced, it's going to be a judgement call about whether it needs doing NOW. And in some cases biasing for understanding the PR diff is more important.

I'd suggest a comment or two to document @keyboardDrummer's observation about the flow, possibly with a TODO.

Can I leave it to the two of you to follow this rough guidance for any other bones of contention in this PR? :)

PR #1756, opened 2022-01-22
chore: revert `libraries` to master

fabiomadge:
#1695 included a workaround which set the commit of the `libraries` submodule to some experimental branch. This one gets us back to master. The problem was a circular dependency between the submodule and this repository. The change in the PR broke the tests of the submodule, but its CI is set to an old release version, making a fix impossible.

In the meantime we released a new version, which runs `libraries`'s master, making the changes possible.


keyboardDrummer:
Consider running the `libraries` integration tests in a separate GH action that doesn't block merges

robin-aws (approved):

PR #1763, opened 2022-01-25
fix: two one-line fixes in the C# compiler

### LINKED ISSUE #1761: Extern constructor has bad signature in C#
```Dafny
class {:extern "ABC"} XYZ {
  var y: bool
  constructor {:extern} (x: bool) {
    y := x;
  }
}

method Main() {
  var xyz := new XYZ(true);
  print xyz.y, "\n";
}
```
`csc` can only fail as the generated partial class erroneously contains some attempt at emitting the constructor.

### LINKED ISSUE #1762: Extern multi-argument constructor invoked incorrectly
```Dafny
class {:extern "ABC"} XYZ {
  var y: bool
  constructor {:extern} Init(x: bool) {
    y := x;
  }
  constructor {:extern} Create(x: bool, z: bool) {
    y := x && z;
  }
}

method Main() {
  var xyz := new XYZ.Init(true);
  print xyz.y, "\n";
  xyz := new XYZ.Create(true, false);
  print xyz.y, "\n";
}
```
The separator is written to an incorrect writer.

fabiomadge:
Fixes #1761 and also fixes #1762.

### DIFF
--- a/Source/Dafny/Compilers/Compiler-Csharp.cs
+++ b/Source/Dafny/Compilers/Compiler-Csharp.cs
@@ -1039,6 +1039,9 @@ namespace Microsoft.Dafny {
       AddTestCheckerIfNeeded(m.Name, m, wr);
       var typeParameters = TypeParameters(TypeArgumentInstantiation.ToFormals(ForTypeParameters(typeArgs, m, lookasideBody)));
       var parameters = GetMethodParameters(m, typeArgs, lookasideBody, customReceiver, returnType);
+
+      if (!createBody && m is Constructor) { return null; }
+
       wr.Format($"{keywords}{returnType} {IdName(m)}{typeParameters}({parameters})");
 
       if (!createBody || hasDllImportAttribute) {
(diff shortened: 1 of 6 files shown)


Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1039,6 +1039,9 @@ protected class ClassWriter : IClassWriter {
       AddTestCheckerIfNeeded(m.Name, m, wr);
       var typeParameters = TypeParameters(TypeArgumentInstantiation.ToFormals(ForTypeParameters(typeArgs, m, lookasideBody)));
       var parameters = GetMethodParameters(m, typeArgs, lookasideBody, customReceiver, returnType);
+
+      if (!createBody && m is Constructor) { return null; }
```
fabiomadge, line 1043: This could also be done one level up, but even though it doesn't break any tests, it's unclear to me if that's valid for all compilers.

keyboardDrummer (approved):

PR #1765, opened 2022-01-25
fix: Meaningful nested tokens for quantifier splitting.

MikaelMayer:
This fixes the problem unveiled in #1695: The related locations were not about the trigger positions, but about sub-expressions.

* fix: If there was no splitting (i.e. the body of the comprehension is the same as the considered sub-expression), this PR also simply REMOVES the extra message about the considered sub-expression which did not bring any extra information
* feat: If there is splitting (info message starting with "For expression"), this PR ensures that it indicates the position of the sub-expression being considered, with a properly capitalized nominal phrase.
* chore: One fix for a Windows test (InferMethodVerificationStability.cs) (unrelated to this PR, but since I had to run all the tests I saw that one was not working locally)

### DIFF
--- a/Test/triggers/loop-detection-messages--unit-tests.dfy.expect
+++ b/Test/triggers/loop-detection-messages--unit-tests.dfy.expect
@@ -1,28 +1,28 @@
 loop-detection-messages--unit-tests.dfy(11,9): Info: Selected triggers: {f(f(i))}
- Rejected triggers: {f(i)} (may loop with "f(f(i))") Trigger position: loop-detection-messages--unit-tests.dfy(11,36)
-loop-detection-messages--unit-tests.dfy(12,9): Info: Selected triggers: {f(_t#0), f(i)} Trigger position: loop-detection-messages--unit-tests.dfy(12,36)
-loop-detection-messages--unit-tests.dfy(13,9): Info: Selected triggers: {f(_t#0), f(i)} Trigger position: loop-detection-messages--unit-tests.dfy(13,52)
+ Rejected triggers: {f(i)} (may loop with "f(f(i))")
+loop-detection-messages--unit-tests.dfy(12,9): Info: Selected triggers: {f(_t#0), f(i)}
+loop-detection-messages--unit-tests.dfy(13,9): Info: Selected triggers: {f(_t#0), f(i)}
 loop-detection-messages--unit-tests.dfy(15,9): Info: For expression "false ==> f(i) == f(_t#0)":
-   Selected triggers: {f(_t#0), f(i)} Trigger position: loop-detection-messages--unit-tests.dfy(15,36)
+   Selected triggers: {f(_t#0), f(i)} in sub-expression at loop-detection-messages--unit-tests.dfy(15,36)
 loop-detection-messages--unit-tests.dfy(15,9): Info: For expression "false ==> f(i) == g(i)":
    Selected triggers:
-     {g(i)}, {f(i)} Trigger position: loop-detection-messages--unit-tests.dfy(15,54)
+     {g(i)}, {f(i)} in sub-expression at loop-detection-messages--unit-tests.dfy(15,54)
 loop-detection-messages--unit-tests.dfy(16,9): Info: For expression "false ==> f(i) == f(_t#0)":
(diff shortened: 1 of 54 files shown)


Test/triggers/loop-detection-messages--unit-tests.dfy.expect
```diff
@@ -1,28 +1,28 @@
 loop-detection-messages--unit-tests.dfy(11,9): Info: Selected triggers: {f(f(i))}
- Rejected triggers: {f(i)} (may loop with "f(f(i))") Trigger position: loop-detection-messages--unit-tests.dfy(11,36)
-loop-detection-messages--unit-tests.dfy(12,9): Info: Selected triggers: {f(_t#0), f(i)} Trigger position: loop-detection-messages--unit-tests.dfy(12,36)
-loop-detection-messages--unit-tests.dfy(13,9): Info: Selected triggers: {f(_t#0), f(i)} Trigger position: loop-detection-messages--unit-tests.dfy(13,52)
+ Rejected triggers: {f(i)} (may loop with "f(f(i))")
+loop-detection-messages--unit-tests.dfy(12,9): Info: Selected triggers: {f(_t#0), f(i)}
+loop-detection-messages--unit-tests.dfy(13,9): Info: Selected triggers: {f(_t#0), f(i)}
 loop-detection-messages--unit-tests.dfy(15,9): Info: For expression "false ==> f(i) == f(_t#0)":
-   Selected triggers: {f(_t#0), f(i)} Trigger position: loop-detection-messages--unit-tests.dfy(15,36)
+   Selected triggers: {f(_t#0), f(i)} in sub-expression at loop-detection-messages--unit-tests.dfy(15,36)
```
cpitclaudel, line 6: I think there's still a problem here: the trigger is made up of two subexpressions, `f(_t#0)` and `f(i)`; which position are we reporting?

Test/triggers/loop-detection-messages--unit-tests.dfy.expect
```diff
@@ -1,28 +1,28 @@
...
 loop-detection-messages--unit-tests.dfy(15,9): Info: For expression "false ==> f(i) == g(i)":
    Selected triggers:
-     {g(i)}, {f(i)} Trigger position: loop-detection-messages--unit-tests.dfy(15,54)
+     {g(i)}, {f(i)} in sub-expression at loop-detection-messages--unit-tests.dfy(15,54)
```
cpitclaudel, line 9: Same here, two triggers and one position

cpitclaudel (changes requested):

Test/triggers/loop-detection-messages--unit-tests.dfy.expect
MikaelMayer: We are reporting the position of the token "==", which is the token held by the binaryExpression "f(i) == g(i)" that cannot be further divided.

Test/triggers/loop-detection-messages--unit-tests.dfy.expect
MikaelMayer: We are reporting the position of the token "==", which is the token held by the binaryExpression "f(i) == f(_t#0)" that cannot be further divided.

MikaelMayer: Note that when expressions are going to report ranges instead of tokens, we might be able to display the range of the expression instead of the token.

cpitclaudel: I *think* I understand what this position means, but I expect it will confuse users. Do you think we need to display it?

MikaelMayer: Glad that we both think we understand what these positions mean.
The alternative would be to remove it, but 1) we won't think of putting it back when we will have ranges and 2) it makes sense for someone who see that there are several "For expression" headers to understand where is that expression that it considered.

We don't display these trigger information by default anyway (it requires a special command line, and in the IDE it's a popup on the comprehensions that we probably want to simplify anyway), only those who care about selected triggers will want to learn about them.
These positions are also used in Boogie to report which condition in a for loop is not satisfied, so that's the single good reason why we want to keep them around. Sounds fair?

cpitclaudel: OK, sounds fair

PR #1766, opened 2022-01-25
Include Z3 resource counts in CSV logs

### LINKED ISSUE #1605: Measure and communicate Z3 rlimit value
This can be used as a more stable cost metric than wall clock time, which depends on the local CPU speed and can be artificially inflated when running multiple verification processes in parallel.

atomb:
Adds Z3 resource counts to CSV log output. Depends on an updated version of Boogie.

Fixes #1605


Source/Dafny/DafnyOptions.cs
```diff
@@ -980,6 +980,10 @@ not sound
         /verificationLogger:trx;LogFileName=<...>.
     The exact mapping of verification concepts to the TRX and CSV formats is
     experimental and subject to change!
+
+    The CSV format also accepts a `Metric` parameter to indicate which metric
```
robin-aws, line 984: Is there a good reason we couldn't just emit both as separate columns all the time, for simplicity?

Test/logger/TickLogger.dfy
```diff
@@ -0,0 +1,42 @@
+// RUN: %dafny /verificationLogger:csv;LogFileName="%t-res.csv",Metric=ticks "%s"
+// RUN: %OutputCheck --file-to-check "%t-res.csv" "%s"
+
+// CHECK: TestResult\.DisplayName,TestResult\.Outcome,TestResult\.ResourceCount
+// CHECK-NEXT: Impl\$\$_module.__default\.ExampleWithSplits\$\$1,Passed,.*
```
robin-aws, line 5: I'd prefer to bump the version of Boogie we depend on first so that we actually get the resource count metric in the `/xml` output, and then actually assert that we get a value here, rather than just asserting the header is there.

robin-aws (changes requested):
SO excited about this. :)

How high is your confidence in using the term "ticks" for this? The only firm label I see at the Z3 level is `rlimit` so I wasn't sure what the most popular term was for what that parameter was limit. I just knew "resources" was not great, and certainly not just "Rs" :)

Source/Dafny/DafnyOptions.cs
atomb: The key reason for this is to prepare for including multiple samples for each proof obligation, along with statistics about those (variance, and so on). With a single metric, those statistics will always be in the same fields and can be processed uniformly.

Test/logger/TickLogger.dfy
atomb: Oh, yes, we definitely need to release Boogie and depend on that before merging this. And I agree about strengthening the check.

atomb:
On the topic of the "ticks" terminology, it emerged [here](https://github.com/boogie-org/boogie/pull/482). I'm not attached to it, but it seems like it's worth having a term.

Test/logger/TickLogger.dfy
atomb: There's now an additional test that fails. Once the new Boogie release is out, I'll incorporate it into this branch and that test should start passing.

Source/Dafny/DafnyOptions.cs
atomb: The latest commit includes all metrics, with the idea that there could be multiple rows in the output for a given proof obligation and that external tools could compute any statistics of interest.

Test/logger/ResCountXML.dfy
```diff
@@ -0,0 +1,41 @@
+// RUN: %dafny /xml:%t.xml %s"
```
robin-aws, line 1: Isn't this just redundantly testing the Boogie `/xml` feature?

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -412,6 +412,10 @@ public enum CommandLineArgumentsResult {
         if (stats.OutOfResourceCount != 0) {
           Console.Write(", {0} out of resource", stats.OutOfResourceCount);
         }
+        if (stats.SolverExceptionCount != 0) {
+          Console.Write(", {0} solver exceptions", stats.SolverExceptionCount);
```
robin-aws, line 416: I see you're also handling the side effects of https://github.com/boogie-org/boogie/pull/488 - for the record I assume you weren't forced to by any build errors when you upgraded the version of Boogie, but just knew it was a good idea to sneak this in as well? :)

robin-aws (approved):
Just a couple of non-blocking questions, LGTM and I don't want to make you run the CI again :)

Test/logger/ResCountXML.dfy
atomb: It's checking that it specifically includes the resource count (which it didn't before) to ensure that tests fail with too old a Boogie version.

Source/DafnyDriver/DafnyDriver.cs
atomb: Yeah, I made the change to Boogie to include that metric, and added a commit to this branch to account for it when I did.

PR #982, opened 2020-12-17
Trace through match expressions for more precise errors

RustanLeino:
Fixes #977

PR #1768, opened 2022-01-27
Trace through match expressions for more precise errors

### LINKED ISSUE #977: more detailed assertion failures for match predicates
Dafny usually gives really specific line numbers in its assertion failures - which is great - but this currently fails when predicates are built out of `match` expressions.

To illustrate, here are two equivalent code snippets:

**match-style.dfy**
```
datatype Option<V> = Some(value: V) | None

predicate is_good_int(x: int)
{
  && x >= 0
  && x != 5
}

predicate is_good_opt(opt: Option<int>)
{
  match opt {
    case Some(x) => is_good_int(x)
    case None => true
  }
}

method Main() {
  var x := Some(5);
  assert is_good_opt(x);
}
```

**impl-style.dfy**
```
datatype Option<V> = Some(value: V) | None

predicate is_good_int(x: int)
{
  && x >= 0
  && x != 5
}

predicate is_good_opt(opt: Option<int>)
{
  && (opt.Some? ==> is_good_int(opt.value))
  && (opt.None? ==> true)
}

method Main() {
  var x := Some(5);
  assert is_good_opt(x);
}
```

The second style gives a much more specific error message than the first, pointing me all the way to the `x != 5` line.
```
$ dafny match-style.dfy 
Dafny 3.0.0.20817
match-style.dfy(19,20): Error: assertion violation
Execution trace:
    (0,0): anon0

$ dafny impl-style.dfy 
Dafny 3.0.0.20817
impl-style.dfy(17,9): Error: assertion violation
impl-style.dfy(11,20): Related location
impl-style.dfy(6,7): Related location
Execution trace:
    (0,0): anon0

```

As such, I've taken to generally writing code in the second style, because the specific error messages are so valuable. That said, I'd much rather be writing in the first style, which seems more idiomatic, so it would be cool if the first style could give the more specific error messages as well.

RustanLeino:
This PR lets "related location" messages trace through the `case`s of `match` expressions. The implementation change itself is small, but it affects the generation of Boogie code--in particular, it affects `TrSplitExpr`, which is called for every proof obligation generated from the user-provided expressions. Thus, the improvements in this PR may be destabilizing for some verifications.

The PR also contains a couple of bug fixes, which were discovered during the testing of the new feature and which were needed in order for the new feature to work. Finally, the PR also contains a few code beautifications. It looks to me that these somewhat orthogonal changes are easy to recognize in the diffs, but if they cause problems for reviewers, then I suggest reviewing each commit separately--I've ordered them in such a way that each of the "chore"s passes the test suite.

I had started a PR #982 for this a year ago, but never returned to it to investigate all the places where the change broke the test suite. Since so many changes happened since then, I will close that PR and am replacing it by this one.

Fixes #977
Supersedes https://github.com/dafny-lang/dafny/pull/982


Test/dafny0/Fuel.dfy
```diff
@@ -332,8 +332,8 @@ module TestModule8 {
             requires ValInGrammar(val, CRequest_grammar());
         {
             if val.c == 0 then
-                var ep := parse_EndPoint(val.val.t[0]); // With default fuel, error: function precondition, destructor, index
-                CRequest(ep, val.val.t[1].u, parse_AppMessage(val.val.t[2]))    // error: index out of range, destructor
+                var ep := parse_EndPoint(val.val.t[0]); // With default fuel, error: function precondition (3x), destructor, index
+                CRequest(ep, val.val.t[1].u, parse_AppMessage(val.val.t[2]))    // (there are errors here, too, but only 5 are reported per method)
```
cpitclaudel, line 336: Or pass `/errorLimit:10` for this test :)

robin-aws:
Applied `do-not-merge` to ensure this doesn't go in until after 3.4, since it will likely cause verification instability in some cases.

atomb (commented):
I'm not deeply familiar with the code involved, but from what I can tell the changes make sense. And it seems like a nice usability improvement!

It's perhaps possible to argue that `git-issue-977.dfy` isn't purely testing issue #977, but the changes are intertwined enough that it probably still makes sense to include all the tests in a single file.

Given my relative unfamiliarity with the code, I'm leaving this as a comment instead of an approval for now.

keyboardDrummer:
@RustanLeino could you complete the documentation of this section? https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#234-well-founded-functions-and-extreme-predicates

Source/Dafny/Verifier/Translator.cs
```diff
@@ -11142,6 +11142,12 @@ public enum K { Free, Checked, Both }
         }
 
         return true;
+
+      } else if (expr is MatchExpr) {
+        var e = (MatchExpr)expr;
+        var ite = etran.DesugarMatchExpr(e);
```
keyboardDrummer, line 11148: Shouldn't `DesugareMatchExpr` be done in an earlier phase?

Source/Dafny/Verifier/Translator.cs
```diff
@@ -11142,6 +11142,12 @@ public enum K { Free, Checked, Both }
         }
 
         return true;
+
+      } else if (expr is MatchExpr) {
+        var e = (MatchExpr)expr;
```
keyboardDrummer, line 11147: Please put the name `e` in the if condition. Also, I suggest you rename it to `matchExpr` since the difference between `e` and `expr` is unclear from the names.

Source/Dafny/Verifier/Translator.cs
```diff
@@ -11183,7 +11183,7 @@ public enum K { Free, Checked, Both }
         var module = f.EnclosingClass.EnclosingModuleDefinition;
         var functionHeight = module.CallGraph.GetSCCRepresentativePredecessorCount(f);
 
-        if (functionHeight < heightLimit && f.Body != null && RevealedInScope(f) && !(f.Body.Resolved is MatchExpr)) {
+        if (functionHeight < heightLimit && f.Body != null && RevealedInScope(f)) {
```
keyboardDrummer, line 11186: It would be nice to have some of the larger cases in `TrSplitExpr`, such as this `expr is FunctionCallExpr && position` one, to be in a separate method. I know such extract method refactorings aren't immediately rewarding but making small refactorings in the code we touch should be the way forward.

If such as refactoring is in the first or last commit of a PR then I think it should have minimal impact on the reviewability.

Test/git-issues/git-issue-977.dfy
```diff
@@ -177,3 +177,42 @@ lemma Test7(k: nat, n: int)
...
+  {
+    BBB(r)
+  }
+
+  greatest predicate BBB(s: nat)
+
+  lemma P(k: ORDINAL, n: int)
+    requires forall t :: AAA#[k](t)
+  {
+    // We DO NOT want the following call to be inlined--a trigger would then contain a "<".
```
keyboardDrummer, line 194: Where can I find more information on this type of inlining? I don't understand this test.

RustanLeino:
@keyboardDrummer Yes, I can write documentation for extreme predicates, but not part of this PR (since this PR doesn't touch them).

Source/Dafny/Verifier/Translator.cs
```diff
@@ -11422,7 +11428,13 @@ public enum K { Free, Checked, Both }
 
     private bool CanSafelyInline(FunctionCallExpr fexp, Function f) {
       var visitor = new TriggersExplorer();
-      visitor.Visit(f);
+      if (f.Body != null) {
+        var body = f.Body;
+        if (f is PrefixPredicate pp) {
```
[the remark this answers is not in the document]
RustanLeino: I probably didn't give it much though. If I did, I'd say I used the short name `pp` since the name is short-lived and thus seemed more readable.

I would like it if C# changed the type of `f` inside the branch, but there's no way to do that and keep the same name, right?

Source/Dafny/Verifier/Translator.cs
RustanLeino: Thanks for the suggestion. I looked into it, but my conclusion is "no". Although TrExpr, TrSplitExpr, and CanCallAssumptions all do the desugaring, CheckWellformedness starts from the sugared version (because it wants to check the well-formedness of the match-source expression once, not just per `case`).

Test/git-issues/git-issue-977.dfy
RustanLeino: I added an explanatory comment in the test file.

Test/git-issues/git-issue-977.dfy
```diff
@@ -181,6 +181,31 @@ lemma Test7(k: nat, n: int)
...
+   * to inline a predicate call P(e) if the body of P uses P's argument in a way that makes it
+   * considered to be a trigger (aka matching pattern). For example, consider predicate
+   * GreatestManualOrd(k, num) defined above, whose body contains
+   *
+   *         forall m :: m < k ==> GreatestManualOrd(m, num)
+   *
+   * The (automatically generated) trigger for this quantifier is GreatestManualOrd(m, num).
+   * Note that this trigger includes GreatestManualOrd's second argument, so if that
+   * argument is something that is not allowed in a trigger, then by inlining the call,
+   * TrSplitExpr would generate malformed Boogie code.
```
cpitclaudel, line 195: I don't understand this paragraph.  If the second argument is something that is not allowed in a trigger, then `GreatestManualOrd(m, num)` is already malformed, isn't it?

Source/Dafny/Verifier/Translator.cs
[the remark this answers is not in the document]
keyboardDrummer: > I would like it if C# changed the type of f inside the branch, but there's no way to do that and keep the same name, right?

No I don't think so. I imagine such a design would create problems for mutable variables since inside the `if`, when read before a write `f` is a `PrefixPredicate`, but when writing to it `f` is still a `Function`.

keyboardDrummer (approved):

keyboardDrummer (approved):

PR #1771, opened 2022-01-28
fix: "Stuck on Verifying..."

MikaelMayer:
This PR adds no new operational code because the fix is in Boogie:
https://github.com/boogie-org/boogie/pull/491

This fix has been merged, and this PR bumps the Boogie version, so the test included should now pass.

This tests asks the verifier to do something hard, and cancels verification by performing a change on the text.
Until this PR, the verifier is stuck verifying, but with the fix above, it will cancel verification, kill zombie processes, and continue smoothly


keyboardDrummer (approved):

Source/DafnyLanguageServer.Test/Various/CancelVerification.cs
```diff
@@ -0,0 +1,72 @@
...
+  else
+    Ack(m - 1, Ack(m, n - 1))
+}
+
+method test() {
+  assert Ack(5, 5) == 0;
+}".TrimStart();
+      var documentItem = CreateTestDocument(source);
+      client.OpenDocument(documentItem);
+      Thread.Sleep(5000);
```
camrein, line 40: Prefer using `await Task.Delay(5_000);` here

camrein (changes requested):
Only some small inputs, and it appears something's wrong with the tests.
I'm quite impressed that this feature didn't require changes to the language server itself. 😄
This will sure have a huge impact on UX.

camrein:
Just in case you're waiting for the approval: I think you didn't push your latest changes. 😄

MikaelMayer:
Indeed, I did not, because there is still a problem in Boogie where counter-examples are not reported correctly anymore. I'll be waiting on Boogie to be compatible with Dafny again to solve this issue.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -98,12 +98,8 @@ public class DafnyProgramVerifier : IProgramVerifier {
       ExecutionEngine.CollectModSets(program);
       ExecutionEngine.CoalesceBlocks(program);
       ExecutionEngine.Inline(program);
-      // TODO The requestId is used to cancel a verification.
-      //      However, the cancelling a verification is currently not possible since it blocks a text document
-      //      synchronization event which are serialized. Thus, no event is processed until the pending
-      //      synchronization is completed.
       var uniqueRequestId = Guid.NewGuid().ToString();
-      using (cancellationToken.Register(() => CancelVerification(uniqueRequestId))) {
+      using (cancellationToken.Register(() => CancelVerification(requestId))) {
```
MikaelMayer, line 102:
```suggestion
      using (cancellationToken.Register(() => CancelVerification(uniqueRequestId))) {
```

camrein (approved):

PR #1773, opened 2022-01-31
docs: Improve documentation of `if-case` statements

cpitclaudel:


docs/DafnyRef/Statements.md
```diff
@@ -748,6 +748,8 @@ expression. For example,
   }
 ```
 
+Unlike `match` statements and `if` expressions, `if` statements do not have to be exhaustive: omitting the `else` block is the same as including an empty `else` block.  To ensure that an `if` statement is exhaustive, use the `if-case` statement documented below.
```
MikaelMayer, line 751: I don't think it makes sense to say "if-expressions" are exhaustive, since they are exhaustive by syntax. I think you can remove `and if expressions`.

MikaelMayer (changes requested):
It's great to have the doc updated, thanks

docs/DafnyRef/Statements.md
cpitclaudel: Thanks, done!

keyboardDrummer (approved):
Nice

robin-aws:
Good update, thanks!

Related note: the reference manual lists its authors as "K. Rustan M. Leino, Richard L. Ford, David R. Cok". If we are going to name authors on this, we should update it! There have almost certainly been contributions from other folks, and any new language change PRs should be bundled with documentation updates as well.

MikaelMayer (approved):
Great explainer. I think this will add legibility.

PR #1780, opened 2022-02-03
fix: thread-safety issue in the runtime

### LINKED ISSUE #1454: Lazy sequence concatenation in C# and Java runtimes is not thread-safe
Both [C#](https://github.com/dafny-lang/dafny/blob/90eb1084d9862384accd03fb5a1c8af500212a05/Source/DafnyRuntime/DafnyRuntime.cs#L1111-L1120) and [Java](https://github.com/dafny-lang/dafny/blob/90eb1084d9862384accd03fb5a1c8af500212a05/Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/DafnySequence.java#L756-L761) optimize sequence concatenation by creating a lazy representation with internal mutable state. I've suspected this isn't thread safe and have now confirmed it isn't in C# at least, using Coyote. Namely, two threads can both determine that the lazy representation needs to be resolved, one thread will complete this operation and clear the `left` and `right` fields so they can be GC'd, and then the second thread throws a NullReferenceException when trying to do the same.

One possible answer to this is "not an issue, the runtimes and Dafny itself aren't guaranteed to be thread-safe", but I'd be personally disappointed by this given this seems to be the only such issue with the runtimes AFAICT so far. The solution in C# is likely to use the fairly sophisticated and thread-safe `ImmutableList` class instead of our hand-rolled implementation of this immutable datatype, and the Java runtime likely has something similar we can wrap.

See also https://github.com/dafny-lang/rfcs/pull/6.

fabiomadge:
Fixes #1454


Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -1189,15 +1189,23 @@ internal class ConcatSequence<T> : Sequence<T> {
...
       while (toVisit.Count != 0) {
         var seq = toVisit.Pop();
-        var cs = seq as ConcatSequence<T>;
-        if (cs != null && cs.elmts.IsDefault) {
-          toVisit.Push(cs.right);
-          toVisit.Push(cs.left);
+        if (seq is ConcatSequence<T> { elmts: { IsDefault: true } } cs) {
+          (leftBuffer, rightBuffer) = (cs.left, cs.right);
+          if (cs.left == null || cs.right == null) {
+            toVisit.Push(cs);
```
cpitclaudel, line 1204: Doesn't this cause an infinite loop?

fabiomadge: Not unless someone tries really hard. `left` and `right` are non-null when the sequence is fresh, otherwise the `Count` call would fail. The only place they are altered thereafter is `ImmutableElements`, which, assuming `ComputeElements` works, will always undefault `elmts`. In summary, there should be at most one retry.

robin-aws: I don't know .NET's memory model as well as Java's - can we definitively say that if one thread sets `elmts` and then unsets `left` and `right`, then any other thread will see some prefix of those state changes in that order, even when there is no synchronization between them?

robin-aws (changes requested):

Source/DafnyRuntime/DafnyRuntime.cs
cpitclaudel: I think I'm missing something easy :/ Assume a stack with a single element `cs` such that `IsDefault` is `true` and `cs.left` is `null`.  Won't the code pop `cs`, take the `seq is ConcatSequence<T>` branch, take the `cs.left == null` branch, push `cs` back into `toVisit`, and on the next iteration `pop` the same `cs`, thus getting into a loop?

Looking more at the code it looks like it's not really possible to have `IsDefault` set to `true` while `left` is `null`?  But in that case I don't understand when this branch will be taken (is it related to the concurrency problem?)

fabiomadge: > A write of a volatile field is called a volatile write. A volatile write has “release semantics”; that is, it is guaranteed to happen after any memory references prior to the write instruction in the instruction sequence.
The write is happening in ImmutableElements, so we should be good.

[ECMA-334](https://www.ecma-international.org/wp-content/uploads/ECMA-334_5th_edition_december_2017.pdf) _15.5.4 Volatile fields_

That's exactly what we need for `ImmutableElements`.

fabiomadge: > Looking more at the code it looks like it's not really possible to have IsDefault set to true while left is null? But in that case I don't understand when this branch will be taken (is it related to the concurrency problem?)

Yes that is to avoid concurrency issues. The idea is that the condition in L1201 becomes `false` by the time we hit L1203. We know that this will happen up to once, so we retry expecting to end up in L1210.

fabiomadge: No, because we want to keep going. There might be more on the stack.

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -1154,7 +1154,7 @@ internal class ArraySequence<T> : Sequence<T> {
   internal class ConcatSequence<T> : Sequence<T> {
     // INVARIANT: Either left != null, right != null, and elmts's underlying array == null or
     // left == null, right == null, and elmts's underlying array != null
-    private ISequence<T> left, right;
+    private volatile ISequence<T> left, right;
     private ImmutableArray<T> elmts;
```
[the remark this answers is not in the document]
robin-aws: Ah shoot. In that case I'm not totally sure we're safe here, but I'm okay deferring to @cpitclaudel to approve the PR

Source/DafnyRuntime/DafnyRuntime.cs
cpitclaudel: Alright, i understand now (@fabiomadge explained ^^).  Let's add a comment in the code that points out the mutation.

Source/DafnyRuntime/DafnyRuntime.cs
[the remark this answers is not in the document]
cpitclaudel: This code is cursed, but I think @fabiomadge 's implementation is right, because of the release semantics of volatile writes (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#1454-volatile-fields has an example that does basically the same as this code).

cpitclaudel (approved):
🧵🧵🧵

PR #1786, opened 2022-02-03
feat: New language constructs for the python backend

prvshah51:
[added functionalities like declaring function method , Binary operations.


Test/comp/firstSteps/Calls-F.dfy
```diff
@@ -0,0 +1,12 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:py "%s" >> "%t"
```
MikaelMayer, line 2: As a side note, since this is a simple test for a compiler, would you be able to make all the other compilers pass this test as well? e.g. Duplicate this line and add js, java, cpp, cs and go.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -233,7 +260,66 @@ protected class ClassWriter : IClassWriter {
...
+      }
+
+    }
+
+    static string FormatDefaultTypeParameterValue(TopLevelDecl tp) {
+      Contract.Requires(tp is TypeParameter || tp is OpaqueTypeDecl);
+      return $"_default_{tp.CompileName}";
+    }
+    string IntegerLiteral(BigInteger i) {
+      if (i.IsZero || i.IsOne) {
```
MikaelMayer, line 317: Why `i.isOne` would result in `int()` ? I thought `int()` always returned zero.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -233,7 +260,66 @@ protected class ClassWriter : IClassWriter {
...
+            // lay down some bits to please the C#'s compiler's different definite-assignment rules.
+            return "null";
+          }
+        } else {
+          return TypeInitializationValue(td.RhsWithArgument(udt.TypeArgs), wr, tok, usePlaceboValue, constructTypeParameterDefaultsFromTypeDescriptors);
+        }
+      } else {
+        Contract.Assert(false); throw new cce.UnreachableException();  // unexpected type
+      }
+
```
MikaelMayer, line 309: Remove empty trailing line

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -189,8 +192,28 @@ protected class ClassWriter : IClassWriter {
...
       List<Formal> formals, Type resultType, IToken tok, bool isStatic, bool createBody, MemberDecl member,
-      ConcreteSyntaxTree methodWriter, bool forBodyInheritance, bool lookasideBody) {
-      throw new NotImplementedException();
+      ConcreteSyntaxTree wr, bool forBodyInheritance, bool lookasideBody) {
+      if (!createBody) {
+        return null;
+      }
+
+      var customReceiver = !forBodyInheritance && NeedsCustomReceiver(member);
+      wr.Write("{0} {1}(", isStatic || customReceiver ? "def" : "", name);
```
MikaelMayer, line 201: did you consider writing`"def "` and remove the space next to `{0}` ?

prvshah51: I did, the else part here will be used for creating object of a class which will also need the space and. And once added in actual practice it will never print an empty string.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -548,6 +640,75 @@ protected override void
...
+      staticCallString = null;
+      reverseArguments = false;
+      truncateResult = false;
+      convertE1_to_int = false;
+
+      switch (op) {
+        case BinaryExpr.ResolvedOpcode.Add:
+          truncateResult = true;
+          if (resultType.IsCharType) {
+            preOpString = "(char) (";
```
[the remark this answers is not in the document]
prvshah51: It's not used, the else part is used for now. I will remove it.

Source/Dafny/Compilers/Compiler-python.cs
prvshah51: made a mistake there, thanks. Removing the `i.IsOne` from if condition so that it will go through the else condition.

Test/comp/firstSteps/Calls-F.dfy
prvshah51: so for context, This is a test file is made up from already present test file Calls.dfy where this file is only testing for 1 function from it because that is what is implemented in the python compiler, I was thinking adding all the languages here will just add repetitive tests.

MikaelMayer: So the plan on the long run is that you will remove this file and add python to the tests of `//calls.dfy` when it will be able to run it?
If so, please add a comment to this file indicating the purpose, such as:
```suggestion
// Once the python compiler will be able to run the file call.dfy, we can remove this file
// RUN: %dafny /noVerify /compile:4 /compileTarget:py "%s" >> "%t"
```

prvshah51: I think the plan is to keep these files for more compilers development. I will add a comment to it.

robin-aws: Yup, the intention is for `firstSteps` to contain an ordered sequence of very simple test files for every new compiler to handle. It will provide a more incremental breakdown of features than the current set of compilation test cases. In the next PR, we should also move `IKnowThisMuchIs.dfy` into this folder and label it as the very first file to attempt. We'll probably want to name the files with prefixes like `0-IKnowThisMuchIs.dfy` to make that even clearer.

MikaelMayer (approved):

Issue #1791, opened 2022-02-07
Assert for the elephant let-or-fail operator in expressions


MikaelMayer:
It's possible to write the following:

```
function SubTest(a: int): (r: Result<string, int>)

function Test(): (r: Result<string, int>) {
    var s :- SubTest(3);
    Success(s)
}
```
but let's say I would like to assert that SubTest won't fail (because of some conditions in the inputs of SubTest).
In methods, it's possible to prefix the call to `SubTest` with `assert`, and it will emit the assertion. However, it's not possible for functions yet.

It would be however obvious that
```
function SubTest(a: int): (r: Result<string, int>)

function Test(): (r: Result<string, int>) {
    var s :- assert SubTest(3);
    Success(s)
}
```
should be rewritten to:
```
function SubTest(a: int): (r: Result<bool, int>)

function Test(): (r: Result<string, int>) {
    var tmp := SubTest(3);
    assert !tmp.IsFailure();
    var s := tmp.Extract();
    Success(s);
}
```

The same goes for the `assume` and `expect` keywords, since many users use `function methods`, might use them for testing, and also to assume something before coming back to it and revisiting the keyword later.

cpitclaudel:
This makes sense for `expect`, but for `assert`, why not just write `var s := SubTest(3).value;`?

MikaelMayer:
I get your point in this particular case; however, there is no guarantee in the general case that `.Extract()` is equivalent to `.value`. The precondition of `.Extract()` is that the object is not `.IsFailure()`, which might be stronger than just being of the success sub-constructor.

cpitclaudel:
Then why not just call `Extract`?

cpitclaudel:
I was not aware of that syntax, and it's mightily confusing!

```
function method SubTest(a: int): (r: Result<string, int>)

method Test() returns (r: Result<string, int>) {
    var s :- assert X; Success("A");
    return Success(s);
}
```

does not mean the same as

```
function method SubTest(a: int): (r: Result<string, int>)

method Test() returns (r: Result<string, int>) {
    var s :- (assert X; Success("A"));
    return Success(s);
}
```

Normally `assert …` is unambiguous in expression context despite using `;` like `var` does because `assert …` must be followed by a `;` and another expression… except in expressions passed to the elephant operator?!

MikaelMayer:
This ambiguity is actually documented in the Dafny reference already.
[...]

PR #1792, opened 2022-02-07
Update Statements.md

MikaelMayer:
I thought it would be good to add a link to the update statement inside the variable declaration statement, in the context of the let-or-fail operator.
I also removed the TODO on the let-or-fail expressions and converted it into an issue instead:
https://github.com/dafny-lang/dafny/issues/1791

### DIFF
--- a/docs/DafnyRef/Statements.md
+++ b/docs/DafnyRef/Statements.md
@@ -649,6 +649,8 @@ function usesTuple() : int
 }
 ```
 
+The special assignment operator `:-` enables to quickly return if the value evaluates to an error-compatible type, see [Section 19.7](sec-update-failure).
+
 ## 19.9. Guards
 ````grammar
 Guard = ( "*"
--- a/docs/DafnyRef/Statements.md
+++ b/docs/DafnyRef/Statements.md
@@ -649,7 +649,7 @@ function usesTuple() : int
 }
 ```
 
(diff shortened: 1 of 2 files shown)


docs/DafnyRef/Statements.md
```diff
@@ -649,6 +649,8 @@ function usesTuple() : int
 }
 ```
 
+The special assignment operator `:-` enables to quickly return if the value evaluates to an error-compatible type, see [Section 19.7](sec-update-failure).
```
keyboardDrummer, line 652: Is 'special assignment operator' the right name? What about 'assignment with failure operator' ?

I think there's a grammatical error in the short summary: "enables to quickly return" -> "enables quickly returning"

Consider rephrasing to "The assignment with failure operator `:-` returns [from the method] if the value evaluates to a failure value of a failure-compatible type, see [Section 19.7](sec-update-failure)".

keyboardDrummer (approved):

docs/DafnyRef/Statements.md
```diff
@@ -649,6 +649,8 @@ function usesTuple() : int
 }
 ```
 
+The assignment with failure operator `:-` returns (from the method) if the value evaluates to a failure value of a failure-compatible type, see [Section 19.7](sec-update-failure).
```
keyboardDrummer, line 652: I think without the parenthesis around 'from the method' would be better. AFAIK parenthesis are used to switch topics, but here the part 'from the method' doesn't do that.

PR #1794, opened 2022-02-07
doc: Fill in section 24.10.5

cpitclaudel:


docs/DafnyRef/UserGuide.md
```diff
@@ -308,8 +308,13 @@ TO BE WRITTEN
...
 
-* `-verifyAllModules` - verify modules that come from include directives (default is to not verify included modules)
-TO BE WRITTEN
+* `-verifyAllModules` - verify modules that come from include directives
+
+By default, Dafny only verifies files explicitly listed on the command line: if `a.dfy` includes `b.dfy`, a call to `Dafny a.dfy` will detect and report verification errors from `a.dfy` but not from `b.dfy`'s.
+
+With this flag, Dafny will instead verify everything: all input modules and all their transitive dependencies.  This way `Dafny a.dfy` will verify `a.dfy` and all files that it includes (here `b.dfy`), as well all files that these files include, etc.
+
+Running Dafny with `/verifyAllModules` on the file containing your main result is a good way to ensure that all its dependencies verify.
```
MikaelMayer, line 317: Can you add a comment that it's the default option in the Dafny IDE when it comes to compiling a file? That way, it ensures there are no hidden compilation problems.

cpitclaudel: (I don't think this is the right place to document the IDE (and I'm not sure it's a good idea that the IDE does this!)

keyboardDrummer: AFAIK the IDE only verifies the current file, as if there is no `verifyAllModules`

keyboardDrummer (approved):

docs/DafnyRef/UserGuide.md
cpitclaudel: It adds /verifyAllModules when you invoke the "compile" command

MikaelMayer: Indeed. Let's skip this comment then.

PR #1795, opened 2022-02-07
Why do we disallow "decreases *" on function (methods)?

cpitclaudel:
I get that it's unsound, and arguably it's worth on methods than on functions, but we do allow `{:verify false}` on functions and function methods.

The problem with `{:verify false}` is that it disables too much; it would be nice to be able to turn off just termination checks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 DIR=$(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 
-default: parser runtime boogie exe
+default: parser boogie exe
 
-all: runtime boogie exe refman
+all: boogie exe refman
 
 exe:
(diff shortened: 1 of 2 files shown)


MikaelMayer:
> I get that it's unsound

It is documented it is not possible, but not really why
https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#513-decreases-clause

Until this PR, the verifier would always verify that functions do terminate, so that specifications themselves make sense.
It makes sense to not verify termination of a method if you want it to possibly run indefinitely, and every method calling a non-terminating method or loop should also be declared "decreases *".

Hence, to be consistent with methods, you might have to consider that shouldn't be allowed to call a `function method` with a `decrease *` from a pure specification context. I cannot judge of this right now.

PR #1800, opened 2022-02-08
Updates to reflect the 3.4.0 release

atomb:


docs/DafnyRef/Types.md
```diff
@@ -1274,7 +1274,7 @@ type.  For such assignments, an explicit conversion must be used, see
 
 Dafny builds in three families of subset types, as described next.
 
-### 11.3.1 Type`nat`
+### 11.3.1.1 Type`nat`
```
cpitclaudel, line 1277: ?

atomb: I think this was done by one of the packaging scripts... Looking at it now, I don't think it should be included.

cpitclaudel: There's some fancy regex magic going on somewhere :)

cpitclaudel:
Do I understand correctly that this test always runs on the release?  Does it only run only when we make a new release, or with every PR?

atomb:
> Do I understand correctly that this test always runs on the release? Does it only run only when we make a new release, or with every PR?

It looks like it runs only when we make a new release or when it's manually triggered.

atomb:
Oh, but this PR is obsolete now that @fabiomadge has released 3.4.1. I'm going to close it.

PR #1801, opened 2022-02-08
feat: Trivia stored in the AST

### LINKED ISSUE #1528: Add trivia (whitespace, comments, and other non-semantic data) to the AST
This will be necessary to augment the existing pretty-printer into a true, opinionated auto-formatter in the style of `gofmt` or `rustfmt`, which has been a very common ask from Dafny users and the most promising solution to #1311.

The issue is that the current parser discards things like whitespace and comments, and the AST does not represent them in its model. It may be difficult to customize the Coco/R-generated parser to not throw away this information. Even if that is true, however, it should be possible to do an additional pass on the initial AST to rescue the skipped-over text, by retrieving the ranges in between AST nodes based on source position data on the nodes (and cheers to @mschlaipfer for passing on this idea :).

MikaelMayer:
This PR brings the following changes to Dafny
* Replace the use of `Boogie.IToken` by a derivative `Dafny.IToken`, and same for `Token` (no need to change Boogie like in a previous attempt)
* Adds and populate two new fields for every token `leadingTrivia` and `trailingTrivia`
* Makes the first token of every declaration available through `FirstDeclarationToken`, so that it's possible to recover the trivia before it
* Makes the token associated to docstring (new !) available through TokenBeforeDocstring, so that it's possible to recover the trivia after it.

This fixes #1528 for all trivia technically accessible a token stored in the AST.

**Implementation details:**
I ignored the following very relevant advice here:
https://github.com/terrajobst/minsk/blob/master/docs/episode-24.md
because we will visibly put comments after definitions, except when using /**  */ before a top-level definition and separated by at least an empty newline

**Work in Progress**

This PR is the first step towards having a full Dafny formatter.

# What to review

This PR seems huge, but there isn't as much to review. Here is what to look for, in the same order as when you click on the "files" tab. In bold, files that are important reviewing. Other files are just automated refactorings.

* **RELEASE_NOTES.md**
* **Source/Dafny/AST/DafnyAst.cs** Added classes `IToken` and `Token`, and field `FirstDeclarationToken` to declarations
* *Source/Dafny/AST/Printer.cs* Just updated `IToken` and `Token`
* *Source/Dafny/Compilers/Compiler-Csharp.cs* `Bpl.Token` => `Token` and same for `IToken`
* *Source/Dafny/Compilers/Compiler-cpp.cs* Same boring stuff
* *Source/Dafny/Compilers/Compiler-go.cs*	 Same
* *Source/Dafny/Compilers/Compiler-java.cs* Same
* *Source/Dafny/Compilers/Compiler-js.cs* Same
* *Source/Dafny/Compilers/Compiler-python.cs* Same	
* *Source/Dafny/Compilers/SinglePassCompiler.cs* Same
* **Source/Dafny/Dafny.atg** I used the declaration modifiers to set a FirstToken once that is then used by various fields FirstDeclarationToken 	
* *Source/Dafny/DafnyOptions.cs* `Bpl.Token` => `Token`
* *Source/Dafny/RefinementTransformer.cs* Same	
* *Source/Dafny/Rewriter.cs* Same	
* *Source/Dafny/Verifier/BoogieStmtListBuilder.cs* Explicited the use of Boogie token	
* *Source/Dafny/Verifier/CaptureStateExtensions.cs*  `Bpl.Token` => `Token`	
* **Source/Dafny/Verifier/Translator.BoogieFactory.cs** Definition of the method `Translator.ToDafnyToken` used when doing roundtrips with translating to Boogie
* *Source/Dafny/Verifier/Translator.ClassMembers.cs* Just three uses of `splitExpr.Tok` (defined later)
* *Source/Dafny/Verifier/Translator.ExpressionTranslator.cs* Refactoring to use `splitExpr.Tok`, `ToDafnyToken` and use of Dafny Token.
* *Source/Dafny/Verifier/Translator.TrStatement.cs* Same
* *Source/Dafny/Verifier/Translator.cs*	 Same
* *Source/DafnyLanguageServer/Language/AssertionBatchCompletedObserver.cs*	One IToken migrated
* *Source/DafnyLanguageServer/Language/BoogieExtensions.cs* IToken migrated	
* *Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs* Same	
* *Source/DafnyLanguageServer/Language/GhostStateDiagnosticCollector.cs* Same	
* *Source/DafnyLanguageServer/Language/Symbols/DafnyLangSymbolResolver.cs* Migrated a token.	
* *Source/DafnyLanguageServer/Language/SyntaxTreeVisitor.cs* One token migration	
* *Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs* A few migrations	
* *Source/DafnyServer/SymbolTable.cs* Easy migration	
* *Source/DafnyTestGeneration/ProgramModifier.cs* Token disambiguation	
* **third_party/Coco/src/Scanner.frame** Change to the scanner so that it can attach trivia to tokens. Note that the first token takes all the previous trivia.


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -13,14 +13,92 @@
...
-using Microsoft.Boogie;
 using System.Diagnostics;
 using System.Threading;
+using Microsoft.Boogie;
 
 namespace Microsoft.Dafny {
   [System.AttributeUsage(System.AttributeTargets.Field)]
   public class FilledInDuringResolutionAttribute : System.Attribute { }
 
+  public interface IToken : Microsoft.Boogie.IToken {
```
keyboardDrummer, line 24: What's a token?

A usage of the word token that I'm familiar with is to refer to individual words in a program, so the program "method Foo(x: int)" would have the tokens "method, Foo, (, x, :, int, )" (https://en.wikipedia.org/wiki/Lexical_analysis)

Is that also what you mean by token here? Or do you use token to say, an object that exists for every AST node, that contains the syntax related information of that AST node, such as start and end position, and leading and trailing trivia?

If the answer is the former, where there's a token for every word in the program, then I wonder why you'd want to store that information.

If the answer is the latter, then I wonder if we should rename token to "NodeSyntax"

keyboardDrummer, line 24: I would prefer not to inherit from Boogie tokens.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -13,14 +13,92 @@
...
+
+    public Token() => this._val = "anything so that it is nonnull";
+
+    public Token(int linenum, int colnum) {
+      this._line = linenum;
+      this._col = colnum;
+      this._val = "anything so that it is nonnull";
+    }
+
+    public int kind {
```
keyboardDrummer, line 58: Why not use auto-properties?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -13,14 +13,92 @@
...
+  public record Token : IToken {
+    private int _kind;
+    private string _filename;
+    private int _pos;
+    private int _col;
+    private int _line;
+    private string _val;
+    private string _leadingTrivia;
+    private string _trailingTrivia;
+    public Token next;
```
keyboardDrummer, line 47: What is `next` ?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -13,14 +13,92 @@
...
+    public static readonly IToken NoToken = (IToken)new Token();
```
keyboardDrummer, line 48: What's this? Shouldn't the optionality be one level higher, in the AST Nodes, to avoid having bogus values for source location related fields.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -3393,6 +3471,7 @@ public abstract class Declaration : INamedRegion, IAttributeBearingDeclaration {
     public IToken tok;
     public IToken BodyStartTok = Token.NoToken;
     public IToken BodyEndTok = Token.NoToken;
+    public IToken FirstDeclarationToken = Token.NoToken;
```
keyboardDrummer, line 3474: What's this? I would expect any AST node to be able to tell me at what position it starts and ends, and where it's trivia start and end. Why is there something specific to declarations?

Source/Dafny/Verifier/Translator.BoogieFactory.cs
```diff
@@ -867,5 +867,25 @@ enum BuiltinFunction {
       fvars.Add(BplFormalVar(name, ty, incoming, out var e));
       return e;
     }
+
+    public static IToken ToDafnyToken(Bpl.IToken exprTok) {
```
keyboardDrummer, line 871: When do we need to go from Boogie to Dafny tokens? For translating I would expect us to only go from Dafny to Boogie tokens.

MikaelMayer: Because we have Dafny expressions that wrap around Boogie expressions in multiple places. And they use to copy the token of the wrapped expressions.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: The something specific to declarations is that the leading trivia might contain some useful information.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: It's the former, but let me widen your definition of what a token _is_ and _has_.

First, a token _is_ not just a "word", it's a word only for toy grammars and projects. In practice, beyond having a `val` (the word itself) and a potential initial `kind`, a token _has_ a filename from which it was extracted, _has_ a position, _has_ a line, _has_ a column. In our case, tokens even _have_ `next`, which is a pointer to the next token, making it possible to traverse the token list if the first one is available.
Essentially, the last three contain concrete formatting information that is not strictly needed in the scanning phase to build the AST. So, there is a precedent into attaching metadata to tokens.

Second, usual definition define tokens and extract them by stripping out whitespace. This is why we also call the resulting trees after parsing "abstract syntax tree", because they don't contain any formatting information.
I think we identified that it is not desirable to remove formatting information from the syntax tree for many reasons (docstring, translation with comments, auto-formatter, warnings if indentation makes something ambiguous, clever refactoring...).
Hence I will gradually make our syntax trees "concrete", so that they still retain (after parsing phase, at least), concrete information about parsing.

I know two ways of doing this efficiently:
* [Sketch-n-Sketch](https://github.com/ravichugh/sketch-n-sketch) builds the parse tree without going through the lexing phase, so there is no need for the notion of "token", and all the trivia is stored directly as fields of the AST. However, they don't make the distinction between "leading trivia" and "trailing trivia", which is sometimes cumbersome when refactoring.
* In our case, the scanning phase is separate from the parser phase. Moreover, there are 1 more trivia than tokens in any program, and each trivia can be split between what could visibly be attached to the previous token and what could be long to the following token. Thus, it is natural to me to make them "owned" by tokens themselves immediately. This scales better than going back to the file and take substrings according to positions when building the AST, potentially missing out trivia.

With that said, if the AST starts to be aware of tokens it was parsed on, we could simply verify that all tokens are owned by an AST node to ensure we are not missing trivia. This means more robust development down the road.

For now, I am attaching only the first token of a declaration to the declaration itself, because it's the token that typically would have the comment preceding the declaration. This might change as we are discussing where to attach docstring.

MikaelMayer: The reason I agree down the road to not inherit Boogie tokens will be when we will package the verification through Boogie as a different dependency, so we will want to break the dependency.
However, this is not in the near future, Boogie is still be a dependency for now as far as I can tell. Moreover, when we translate the Dafny AST to Boogie, we need to provide Boogie tokens. Inheriting from Boogie tokens ensures we don't have the need to convert Dafny tokens to Boogie tokens, which is redundant, costly in time and resources (there are a lot of tokens) and thus not necessary for now.

I made the refactoring so that, at least, when we will want to break up 💔  with Boogie, imports of Boogie's tokens are already gone, and we will just need a few conversions there and there, which is a boring refactoring.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: Good point. I was just copying Boogie's code. Updated.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: > What is next ?

The token following the current token.

Source/Dafny/Verifier/Translator.BoogieFactory.cs
keyboardDrummer: > Because we have Dafny expressions that wrap around Boogie expressions in multiple places.

Could you give me a pointer for where to find this?

MikaelMayer: * `Bpl.PredicateCmd AssertNS` in translator.cs uses two tokens, one which is the expression token and the other one the refinement token. The check uses `RefinementToken.IsInherited`  which is defined in the resolver pipeline and requires a Dafny token.
https://github.com/dafny-lang/dafny/blob/07d614f193502b07b4cea671c2eeb2018ced726d/Source/Dafny/Verifier/Translator.cs#L8413-L8419
* We first convert an expression to boogie induction cases (it returns a sequence of `Boogie.Expr`), and then use their tokens to build a `NestedToken`, which is requires Dafny tokens as inputs.
https://github.com/dafny-lang/dafny/blob/07d614f193502b07b4cea671c2eeb2018ced726d/Source/Dafny/Verifier/Translator.cs#L11416-L11419
* When splitting an expression, we create multiple `Boogie.Expr`, and then we recreate other NestedTokens based on their tokens.
https://github.com/dafny-lang/dafny/blob/07d614f193502b07b4cea671c2eeb2018ced726d/Source/Dafny/Verifier/Translator.TrStatement.cs#L99-L104
* When substituting expressions, we use a BoogieWrapper that takes a Boogie expression and converts it into a Dafny expression, and there are 27 such calls to `new BoogieWrapper`:
https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.cs#L10633-L10638
* Also, when recovering counter-examples and Boogie expressions after verification, we get Boogie tokens, but now the LSP uses Dafny tokens.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: I renamed it to `peekedToken` because it's an implementation detail in the scanning strategy. It's not exposed in the `IToken` interface.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: `FirstDeclarationToken` doesn't seem to be used in this PR except for in a test, and it doesn't make sense to me that you would store that here. If you want to know the leading trivia of the first member of a class, then you could just select the first member and get its leading trivia, right?

keyboardDrummer:
> This PR is the first step towards having a full Dafny formatter, because it makes theoretically the entire source code available within the AST already. More effort will be needed later for the printer though.

I imagine we need to know per AST Node which trivia/token it owns. In the parser we could have something like:

```c#
AstNodeSyntax GetSyntaxFromStartToken(IToken startToken) {
  var endToken = t;
  var tokensForNode = new List<IToken>();
  var loopToken = startToken;
  while(loopToken != endToken) {
    if (consumedTokens.Add(loopToken) {
      tokensForNode.Add(loopToken);
    }
    loopToken = loopToken.next;
  }
  if (consumedTokens.Add(loopToken) {
    tokensForNode.Add(loopToken);
  }
  return AstNodeSyntax(tokensForNode);
}
```

The `consumedTokens` is reused during the whole parse, and the above method `GetSyntaxFromStartToken` ensures that for an expression `/*triv*/3 /*triv2*/+ /*triv3*/2`, `triv` and `triv3` are attached to the literals while `triv2` is attached to the `+` operator.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: Maybe I should rename it "FirstToken"? The token of a declaration is usually its symbol name. There can be a lot of tokens before such as "method", "function", "twostate"... "ghost". `FirstDeclarationToken` stores the first one of such tokens, for every declaration.
It's the only token that has the trivia that exists before a declaration, which might contain comments (I'm doing this in a follow-up PR)

MikaelMayer:
> The `consumedTokens` is reused during the whole parse, and the above method `GetSyntaxFromStartToken` ensures that for an expression `/*triv*/3 /*triv2*/+ /*triv3*/2`, `triv` and `triv3` are attached to the literals while `triv2` is attached to the `+` operator. The assumption is that `GetSyntaxFromStartToken` is called in order for the AST nodes `3`,`2`,`+`.

I have some code that does something similar to what you suggest in a follow-up PR.
Basically, every node will store a pointer to the start token and its end token, and tokens will always know their successor. That way, we can just iterate over tokens as needed and don't need to create a list of token for each AST node.

Moreover, it will make it very easy to change the indentation of an existing file: After parsing, we traverse every node, mark every token we encounter with an explicit indentation amount, and then traverse the entire linked list of tokens by correcting individual's token indentation, concatenating adjusted trivia and token values.

But that's again for another PR. For now, I just want to start with the trivia stored in certain tokens, and then iterate.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: I see. I think all AST nodes in general should know their range, which would include "method", "function", etc. Declarations should also know the range of their name which we could bind to a separate field, say "NameRange". How I've also seen this is that the name is stored in a separate AST Node, so then the NameRange is simply the regular range of that child node.

keyboardDrummer:
> Basically, every node will store a pointer to the start token and its end token, and tokens will always know their successor. That way, we can just iterate over tokens as needed and don't need to create a list of token for each AST node.

But if you iterate like that, you will also iterate over tokens that belong to subnodes.

For the printing with trivia, I imagine code like:

```c#
WritePlus(writer, plusNode) {
  WriteExpr(writer, plusNode.Left);
  WriteTrivia(plusNode.Trivia[0]);
  Write("+");
  WriteExpr(writer, plusNode.Right);
}
```

If for the program `/*triv*/3 /*triv2*/+ /*triv3*/2` you set `/*triv*/3` and `/*triv3*/2` as start and ending tokens for the AST node of `+` that determine the trivia, then `plusNode.trivia[0]` returns `/*triv*/` instead of `/*triv2*/`.

So I think `+` should look like:
```c#
{
  startToken: /*triv*/3
  ownedTokens: [/*triv2*/+]
  endToken: /*triv3*/2
}
```

Although I'd prefer not using tokens altogether so it would be `{ fileRange: FileRange, trivia: string[] }`

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: If we wanted, we could definitely add a stronger API in declarations like this:
```
class Declaration { ...
  public IToken NameRange => tok
}
```
that would consolidate the meaning of "tok", which otherwise does not carry this meaning. That would make it easier for compilers to know what the token is about. For now, I don't thing there is an urging case to do it, but I might be biased as usual.

In a follow-up PR, all AST nodes will be truly aware of their entire range. For that, they will need to have ownership of tokens (and their trivia optionally), but being able to access their children's token in the case of prefix, infix  postfix operations or parentheses calls.

MikaelMayer:
> But if you iterate like that, you will also iterate over tokens that belong to subnodes.

> For the printing with trivia, I imagine code like:
> 
> ```cs
> WritePlus(writer, plusNode) {
>   WriteExpr(writer, plusNode.Left);
>   WriteTrivia(plusNode.Trivia[0]);
>   Write("+");
>   WriteExpr(writer, plusNode.Right);
> }
> ```

That's what I have in mind too, and I already implemented such printing of binary expressions in another separate branch.

> So I think `+` should look like:
> 
> ```cs
> {
>   startToken: /*triv*/3
>   ownedTokens: [/*triv2*/+]
>   endToken: /*triv3*/2
> }
> ```

Yes that's what I have in mind too !

> Although I'd prefer not storing tokens altogether

How would not storing tokens be better?

> I guess you could determine which tokens are owned by which node during printing, and maybe that makes no difference, but my first inclination would be to already do that during parsing.

Ok, I probably should have detailed in more, but I still plan to assign token ownerships to trees. The linked list structure of tokens is there only for two purposes:

First, @cpitclaudel remarked that, after printing, we could theoretically verify that we have gone through every token. That means, we can guarantee that no token was left behind.
But for me I have another value proposition of this linked list of token. Since scanning guarantees that the following loop prints the file as it was parsed:

```
IToken t = firstToken;
string reconstructedFile = "";
while(t != null) {
  reconstructedFile += t.leadingTrivia + t.val + t.trailingTrivia;
  t = t.next;
}
return reconstructedFile;
```

I was saying that, before we dive into a printer that is costly to maintain as the language evolves (although tricks like verifying tokens can help), we could simply visit the Dafny tree, create a map from some tokens to their expected indentation, and then use the loop above and fix the indentation. That way, we can obtain a free _guarantee_ that the printed program would be parsed the same (except for trivia):

```
string Print(IToken firstToken, Dictionary<IToken, int> indentationLevel) {
  IToken t = firstToken;
  string reconstructedFile = "";
  var indentationLevel = 0;
  while(t != null) {
    indentationLevel.TryGetValue(t, out indentationLevel);
    reconstructedFile += fixIndentation(t.leadingTrivia, indentationLevel) + t.val + fixIndentation(t.trailingTrivia, indentationLevel);
    t = t.next;
  }
  return reconstructedFile;
}
```

Of course, in the final setting, `fixIndentation` might be more complex, since it could decide to insert newlines, and thus change the current indentation level.

But in any case, for the above to work, we need token ownership so that every node can tell which tokens should be indented differently or not.
That's an idea to elaborate on.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: > In a follow-up PR, all AST nodes will be truly aware of their entire range.

Would that make `FirstDeclarationToken` obsolete then?

Source/Dafny/Coco/Scanner.frame
```diff
@@ -495,10 +545,10 @@ public class Scanner {
   // get the next token (possibly a token already seen during peeking)
   public Token/*!*/ Scan () {
    Contract.Ensures(Contract.Result<Token>() != null);
-    if (tokens.next == null) {
+    if (tokens.peekedTokens == null) {
```
keyboardDrummer, line 548: Actually `next` is fine for me, but shouldn't this be capital case?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -13,14 +13,60 @@
...
+    // Inherited for now.
+    /*
+    int kind { get; set; }
+    string filename { get; set; }
+    int pos { get; set; }
+    int col { get; set; }
+    int line { get; set; }
+    string val { get; set; }
+    bool IsValid { get; }*/
+    string leadingTrivia { get; set; }
```
keyboardDrummer, line 34: These should start with a capital right?

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: I imagine we want to end with something like:

```c#
record TokenRange(IToken: start, IToken end)
record ASTNode(range: TokenRange?)
```

So there's no `NoToken`, generated AST nodes just use `null` instead of an instance of `TokenRange`.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -13,14 +13,60 @@
...
+    string trailingTrivia { get; set; }
+  }
+
+  public record Token : IToken {
+    public Token peekedTokens; // Used only internally when the scanner "peeks" tokens. Normallly null at the end of parsing
+    public static readonly IToken NoToken = (IToken)new Token();
+
+    public Token() => this.val = "anything so that it is nonnull";
+
+    public Token(int linenum, int colnum) {
```
keyboardDrummer, line 44: Shouldn't this also take a value for `val` ?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -13,14 +13,60 @@
...
+    bool IsValid { get; }*/
+    string leadingTrivia { get; set; }
+    string trailingTrivia { get; set; }
+  }
+
+  public record Token : IToken {
+    public Token peekedTokens; // Used only internally when the scanner "peeks" tokens. Normallly null at the end of parsing
+    public static readonly IToken NoToken = (IToken)new Token();
+
+    public Token() => this.val = "anything so that it is nonnull";
```
keyboardDrummer, line 42: Why do we need this constructor?

keyboardDrummer:
> Yes that's what I have in mind too !

Great :)

> How would not storing tokens be better?

It's just less duplicated information, but let's stick with tokens for now. Once we have all the right information in the AST, it should be simpler to see if replacing tokens with the bar location/trivia information is nicer or not. I can have a look then.

> First, @cpitclaudel remarked that, after printing, we could theoretically verify that we have gone through every token. That means, we can guarantee that no token was left behind. But for me I have another value proposition of this linked list of token. Since scanning guarantees that the following loop prints the file as it was parsed:
> 
> ```
> IToken t = firstToken;
> string reconstructedFile = "";
> while(t != null) {
>   reconstructedFile += t.leadingTrivia + t.val + t.trailingTrivia;
>   t = t.next;
> }
> return reconstructedFile;
> ```
> 
> I was saying that, before we dive into a printer that is costly to maintain as the language evolves (although tricks like verifying tokens can help), we could simply visit the Dafny tree, create a map from some tokens to their expected indentation, and then use the loop above and fix the indentation. That way, we can obtain a free _guarantee_ that the printed program would be parsed the same (except for trivia):
> 
> ```
> string Print(IToken firstToken, Dictionary<IToken, int> indentationLevel) {
>   IToken t = firstToken;
>   string reconstructedFile = "";
>   var indentationLevel = 0;
>   while(t != null) {
>     indentationLevel.TryGetValue(t, out indentationLevel);
>     reconstructedFile += fixIndentation(t.leadingTrivia, indentationLevel) + t.val + fixIndentation(t.trailingTrivia, indentationLevel);
>     t = t.next;
>   }
>   return reconstructedFile;
> }
> ```
> 
> Of course, in the final setting, `fixIndentation` might be more complex, since it could decide to insert newlines, and thus change the current indentation level.
> 
> But in any case, for the above to work, we need token ownership so that every node can tell which tokens should be indented differently or not. That's an idea to elaborate on.

Very cool! Maintaining the manual printing of trivia in the printer seems like a big problem to me. However, would the token iterating technique stop working if you do any kind of refactoring? I think a different way of preventing you from getting the trivia printing for free would be to use an `ASTWriter` instead of a regular `TextWriter`, and have the `ASTWriter` have some understanding of tokens.

Your printer would look like:

```c#
WritePlus(ASTWriter writer, ASTNode plusNode) {
  writer.EnterNode(plusNode, () => {
    WriteExpr(writer, plusNode.Left);
    writer.WriteToken("+");
    WriteExpr(writer, plusNode.Right);
  });
}

class ASTWriter {
  Stack<IEnumerator<IToken>> remainingOwnedTokensPerNode = new();
  TextWriter writer;

  public void EnterNode(ASTNode node, Action action) {
    remainingNodeTokenStack.Push(node.OwnedTokens.GetEnumerator());
    action();
    if (remainingNodeTokenStack.Peek().MoveNext()) {
      throw new Exception("Exiting node before all owned tokens were consumed");
    }
    remainingNodeTokenStack.Pop();
  }

  public void WriteToken(string value) {
    var ownedTokens = remainingOwnedTokensPerNode.Peek();
    ownedTokens.MoveNext();
    writer.Write(ownedTokens.Current.LeadingTrivia);
    // note that ownedTokens.Current.value is ignored and instead the parameter value is used.
    writer.Write(value);
    writer.Write(ownedTokens.Current.TrailingTrivia);
  }
}
```

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -3113,6 +3159,8 @@ public abstract class Declaration : INamedRegion, IAttributeBearingDeclaration {
     public IToken tok;
     public IToken BodyStartTok = Token.NoToken;
     public IToken BodyEndTok = Token.NoToken;
+    public IToken FirstDeclarationToken = Token.NoToken;
+    public IToken TokenBeforeDocstring = Token.NoToken;
```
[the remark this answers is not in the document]
MikaelMayer: Oh actually, since the trivia is attached a bit differently now, the token before the docstring actually contains the docstring in the trailing trivia of the token (see TriviaTest.cs)
The splitting occurs on the first two consecutive newlines without comments, if any.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: Yes they should like in most interfaces, if we wrote this interface from scratch.
However, since I'm inheriting from Boogie tokens, I keep their style.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: For now, I just copied the token's constructors from Boogie. We could add more constructors based on the need later, but I did not want to overweight this PR.

MikaelMayer:
> ```cs
> remainingOwnedTokensPerNode
> ```

This is a fascinating implementation, something along the lines what I was dreaming about. Yes, with a linked list of tokens, we can achieve something like this.
But I was also thinking, the while loop trick does not need to be applied only at the top-level. it can be applied at every node down the road.
For example, consider the following code:

```dafny
class A {
  var a: int, b: int;
}
```
when it is being parsed, the parser actually automatically expands the one line containing `var` into two different fields:
```dafny
class A {
  var a: int;
  var b: int;
}
```

Remembering that it was written like a single field is going to be tricky (but not impossible)


However, with the while loop of tokens, we can detect that the two declarations overlap:
```dafny
class A {
  var a: int, b: int;
  ^^^^^^^^^^ range of first field
  ^^^^^^^^^^^^^^^^^^^ range of second field
}
```
so after printing the first field, all we are left to do is to print the tokens from the second field that don't overlap.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: There are 220 occurrences of `Token.NoToken` in the DafnyPipeline, 39 in the parser, 21 in the resolver, 50 in the verifier, 67 in the compilers, and 36 in the AST. 
Although you might be right, I implore you to consider that removing NoToken from the source code is an implementation detail and out of scope for this PR.
Besides, there are places that actually check if a token is null or not (which is different from being a NoToken), to enable or disable some characteristics (like in the parser, if there is a token for "method" after "function", then it's a function method). I know the implementation should be better, but this PR is already touching a lot of code just to remove the dependency on Boogie's tokens until verification,.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: No, it won't be obsolete, I will just add a new field `FinalDeclarationToken`.
I could also rename them StartToken and EndToken if you preferred.

MikaelMayer:
> For the formatter I think it would be good to move the rewrites that the parser does into rewrite passes, so we can skip them when formatting. That means a little extra work up-front but also means we don't need temporary tricks during printing.

Mmmmh in our case that would me we would need a new concrete syntax tree node called MultipleFieldDecl.
I would prefer to have the AST be aware of how it was built so that we don't need an extra rewrite step.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: I think you can make all of them capital case by using an explicit interface implementation for the Boogie members, like this:

```c#
int Boogie.IToken.pos {
  get => Pos;
  set => Pos = value;
}

public abstract int Pos { get; set; }
```

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: I looked into how easy it would be to have a Dafny token that doesn't inherit form the Boogie one, as a separate PR, and found that because C# implicit conversions are only between classes, so not between `Dafny.Token` and `Boogie.IToken`, it takes lots of tiny changes in the translator code to do those casts explicitly.

So that refactoring is a little expensive for what it brings and I think it's OK to leave it for now.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: > No, it won't be obsolete, I will just add a new field FinalDeclarationToken.
I could also rename them StartToken and EndToken if you preferred.

Do we agree on what the end-state should look like? I imagine this:

```
record TokenRange(start: Token, end: Token);
record ASTNode(range: TokenRange);
record Declaration(nameRange: TokenRange) : ASTNode
```

Given that, I don't see how introducing `FirstDeclarationToken` moves toward that, since then there won't be a need for the `FirstDeclarationToken ` and `FinalDeclarationToken` fields in `Declaration` right?

Source/Dafny/Coco/Scanner.frame
```diff
@@ -453,13 +463,53 @@ public class Scanner {
...
+        lastWasCr = (ch == '\r');
+      }
+    }
     while (ch == ' ' ||
 -->scan1
-    ) NextCh();
+    ) {
+      if(triviaTrailing && (ch == '\r' || ch == '\n')) {
+        if(ch == '\r' || (ch == '\n' && !lastWasCr)) { // Don't count \r\n as two newlines.
+          triviaNewlines++;
```
keyboardDrummer, line 485: What's the definition of trailing trivia? Shouldn't any newlines interrupt set `triviaTrailing = false;` ?

Source/Dafny/Verifier/Translator.BoogieFactory.cs
```diff
@@ -867,5 +867,24 @@ enum BuiltinFunction {
...
+      if (exprTok == null) {
```
keyboardDrummer, line 872: Can't we just cast to `Dafny.Token` since those are the only tokens that we are passing to Boogie? Or does Boogie generate fresh `Boogie.IToken` instances?

keyboardDrummer:
> Mmmmh in our case that would me we would need a new concrete syntax tree node called MultipleFieldDecl. I would prefer to have the AST be aware of how it was built so that we don't need an extra rewrite step.

Is there a downside to having that extra rewrite step?

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: That can be done in another PR, right?

MikaelMayer: In another PR, we could do this change AND also make sure we don't use the Boogie's lowercase interface, but only Dafny's one.

Source/Dafny/Verifier/Translator.BoogieFactory.cs
```diff
@@ -867,5 +867,24 @@ enum BuiltinFunction {
...
+        return null;
+      } else if (exprTok is IToken t) {
+        return t;
+      } else {
```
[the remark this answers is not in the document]
MikaelMayer: We have no guarantee that all the Boogie tokens are Dafny tokens. There exists at least Boogie.Token.NoToken which is not a Dafny token, and we don't have a formal contract indicating it will never be passed here.

Note that the code here won't have any performance penalty.

On the other side, I would block a PR that would have the same performance but casts the token without checking:
```
      if (exprTok == null) {
        return null;
      } else {
        return (IToken)t; // I don't see a reason why it shouldn't be a Dafny token
      }
```
We are not writing in Dafny, so robustness implies we take into account potential crashes.
If you prefer, I could write a unit test so that this branch is covered and this case occur.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: For a single node, we want and will want to access tokens for different reasons:
- To access the token that is just before the official whitespace where we expect documentation.
- To access certain tokens that delimits an inner region, not the entire node (e.g. the body of a method)
- To access certain tokens that indicate whether some keyword was added or not (existing behavior, see extends, refines, etc.)
- To access the first token of a declaration
- To access the final token of a declaration, to know where the printer should stop
- To access the range, with the start token and the end token
- To access individual owned tokens to be able to print them later

We could define a single field "Range" as you state, and tell our users that if they want the first token, they have to use `RangeToken.Start` instead of `FirstDeclarationToken`. But I developed this PR based on needs: Our users probably don't care about the final token of a declaration. It's mostly an internal thing for the pretty-printer down the road, and possibly for refactoring. So why should we create a TokenRange so early? Compiler implementers, will care more about the first token than the range because the first token that is likely to contain docstring.
In this PR, we don't need the final token at all.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: I don't like approving new lines of code that don't follow our coding standards. 

How about you introduce the properties like this:

```c#
interface IToken : Boogie.IToken {
  int Boogie.IToken.pos {
    get => pos;
    set => pos = value;
  }

  int pos { get; set; }
}
```

and then rename `pos` to `Pos` using the IDE, so it's almost no work on the editing side?

Removing the inheritance on `Boogie.IToken` doesn't seem so valuable right now. I think we can shelf that refactoring for the time being.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: You could use the constructor that takes parameters and pass default values, or at least make that constructor private.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: > I did not want to overweight this PR

One option could have been for the PR introduce Dafny.Token but not do anything with trivia, and make `Dafny.Token` a class that follows our coding standards. I wouldn't expect `Dafny.Token` to be mutable for example, and while the setters are not often called it would take some extra changes to remove them.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: Ok. I did try the refactoring. + manually on the frames files (Scanner/Parser). But it's just impossible: Coco/R uses the given class Token and believes that the fields should be lowercase (and in the generated code they are treated as such). There is no way I can modify the tool to change that.
https://ssw.jku.at/Research/Projects/Coco/

Again, the code we are merging was already part of a dependency, so not doing this refactoring is the statu quo, and, based on the experience above, makes more sense as of today.

keyboardDrummer: Okay, let's leave it like this then. Do you know which fields Coco/R is targeting? Could you put comments on them to explain why they're lowercase?

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: My bad, it is also used 11 times in generated code (5 Parser.cs and 6 Parser.cs), 1 time in the translator.
Do you agree it's not appropriate to do this change then?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -24,32 +24,56 @@ public class FilledInDuringResolutionAttribute : System.Attribute { }
...
+    string Filename { get; set; }
+
+    /// <summary>
+    /// TrailingTrivia contains everything after the token,
+    /// until and including two newlines between which there is no commment
+    /// All the remaining trivia is for the LeadingTrivia of the next token
+    ///
+    /// ```
+    /// const /*for const*/ x /*for x*/ := /* for := */ 1/* for 1 */
+    /// // for 1 again
```
keyboardDrummer, line 48: How can `// for 1 again` become a docstring for the const, if the trivia is attached to a childnode of the body of the const?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -24,32 +24,56 @@ public class FilledInDuringResolutionAttribute : System.Attribute { }
   public interface IToken : Microsoft.Boogie.IToken {
     // Inherited for now.
```
keyboardDrummer, line 25: I think we can drop this comment.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -24,32 +24,56 @@ public class FilledInDuringResolutionAttribute : System.Attribute { }
...
     /*
-    int kind { get; set; }
-    string filename { get; set; }
-    int pos { get; set; }
-    int col { get; set; }
-    int line { get; set; }
-    string val { get; set; }
+    int kind { get; set; } // Used by coco, so we can't rename it to Kind
```
keyboardDrummer, line 27: Could you move these comments so they occur on the lowercase properties?

Source/Dafny/AST/DafnyAst.cs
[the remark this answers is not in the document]
keyboardDrummer: The name `TokenBeforeDocstring` is a little confusing for me, since our `Token` class also contains the trivia, so the trivia are not separate entities between `Token`s, leading me to read this as `TokenBeforeDocStringToken`, while it's really `TokenWithTrailingDocString`.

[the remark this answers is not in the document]
MikaelMayer: Very good comment and suggestion. Renamed.

keyboardDrummer (approved):

PR #1803, opened 2022-02-09
Update release preparation instructions

atomb:
* Convert to Markdown
* Add a note about creating a GitHub Discussion for each release
* Update Homebrew instructions
* General cleanup


cpitclaudel (approved):
LGTM.  In the future I'd recommend doing two commits (one to rename one, then one with the edits), since it leads to more useful diffs

atomb:
> LGTM. In the future I'd recommend doing two commits (one to rename one, then one with the edits), since it leads to more useful diffs

Oh, you're right, that is it a pretty useless diff. The funny thing is that I originally created a commit with a nice diff (a single one, but using `git mv`) but then undid it and decided to commit to a different branch. I was overly optimistic that Git would recognize it as a rename automatically. I should have double-checked!

PR #1526, opened 2021-10-21
fix: mocking objects in Dafny tests

Dargones:
This PR makes a few modifications to the way the `{:test}`-annotated methods are compiled to C#, which should make it possible to use this attribute with tests generated by Dafny's Test Generation module as previously suggested by Robin here (https://github.com/dafny-lang/dafny/pull/1385#discussion_r699787246). If these modifications seem adequate, I am planning  to make `{:test}` compilable to Java as well. The changes are:

- If the tested method takes parameters, mock them using C#'s `FormatterServices.GetUninitializedObject` method. This is useful in situations in which you know the state of the object you would like to test and can set its fields accordingly but don't know what constructor to use to allocate the object without running into a runtime exception. In other target languages, the mocking could be done by exploiting the zero constructor.

- Allow tested methods to return non failure-compatible values. Only call `IsFailure` on the first returned value if it is failure-compatible. This is useful as it allows using methods annotated with `{:test}` for other purposes, e.g. for printing the returned values.

- Not sure if this is important but the previous version of the code would only call `IsFailure` on the first returned value of a method **if** there were more than one returned value. Is there any reason for requiring two or more returned values? I removed this condition and the tests pass but do let me know if it should still be there.

Thank you!

Issue #1767, opened 2022-01-27
Allow detection of inconsistent assumptions


atomb:
It's currently possible to state inconsistent assumptions (equivalent to false, though perhaps only when combined) and have no notification that you've done so. The easiest way to do this is through an inconsistent postcondition on an extern function, but it can also occur due to explicit branches in a program that are (unintentionally?) unreachable. Assertions within such branches will always succeed, which may lead to false confidence.

This could potentially be computationally expensive, so we'd probably want to make it optional.

atomb:
Related to #620.

atomb:
This is now supported with `--warn-contradictory-assumptions`.

PR #1809, opened 2022-02-10
Support for Mocking in C# Compiler

Dargones:
### UPDATE (Mar 28 2022):

As per the ensuing discussion, the functionality of the `{:mock}` and `{:fresh}` attributes has been merged into a single `{:synthesize}` attribute. Methods annotated with `{:synthesize}`  no longer need to be annotated with `{:extern}`. The compilation behavior described below no longer requires a special CLI option to be triggered. Finally, the compiler will now throw an error if a stubbed method includes any post-conditions, which should guard against a use of the `{:synthesize}` attribute that might introduce unsoundness.

### ORIGINAL PR DESCRIPTION:

This PR adds mocking support to the C# compiler. The idea is to have a special annotation for external methods that makes the compiler translate a postcondition to a series of API calls in C#'s Moq library (provided the `/compileMocks` command-line argument is also supplied). As an example, consider the following Dafny code, which tests the static method `getOrDefault` of class `StringMap` by mocking a `StringMap` object and stubbing its two instance methods. Using this approach, one is able to test the implementation of `getOrDefault` without specifying how one would add elements to the map.

```Dafny
class StringMap {

	var m:map<string, string>;

	function method Contains(key:string):bool reads this {
		key in m
	}

	function method Get(key:string):string reads this
		requires Contains(key)
	{
		m[key]
	}

	static function method GetOrDefault(m:StringMap, key:string, default:string):string reads m
	{
		if m.Contains(key) then m.Get(key) else default
	}
}

method {:extern} {:mock} MockStringMap(k:string, v:string) 
	returns (m:StringMap)
	ensures m.Contains(k) == true && m.Get(k) == v
	ensures fresh(m)

method {:test} PassingTestGetOrDefault() {
	var stringMap := MockStringMap("a", "b");
	assert StringMap.GetOrDefault(stringMap, "a", "c") == "b";
	expect StringMap.GetOrDefault(stringMap, "a", "c") == "b";
}
```

The `mockStringMap` method is automatically translated to the following in C#, allowing a runtime test that would otherwise be impossible without fully implementing the StringMap class:

```cs
public static StringMap mockStringMap(Dafny.ISequence<char> k, Dafny.ISequence<char> v) {
    var mTmp = new Mock<StringMap>();
    mTmp.Setup(x => x.contains(k)).Returns(true);
    mTmp.Setup(x => x.get(k)).Returns(v);
    return mTmp.Object;
}
```

Only functions can be mocked - methods can have side effects and are, therefore, not allowed in postconditions. Not all Dafny postconditions can be translated to calls in Moq either - compiling an arbitrary postcondition would be a full-blown synthesis problem and, while intersting from a research perspective, it would probably not be fast enough for a compiler. Here is the grammar for postconditions that are supported (`S` is the start symbol, `EXP` stands for an arbitrary Dafny expression, and `ID` stands for variable/method/type identifiers):

```
S       = FORALL 
        | EQUALS 
        | S && S
EQUALS  = ID.ID (ARGLIST) == EXP // stubs a method call
        | ID.ID            == EXP // stubs field access
        | EQUALS && EQUALS
FORALL  = forall SKOLEMS :: EXP ==> EQUALS
ARGLIST = ID // this can be one of the skolem variables
        | EXP // this exp may not reference any of the skolems 
        | ARGLIST, ARGLIST
SKOLEMS = ID : ID 
        | SKOLEMS, SKOLEMS
```

The compiler throws a NotImplementedException if postconditions on the mocked method do not follow this grammar. Support for mocking could be easily extended to Java and other target languages (and I am happy to work on this). Here is a [design document](https://github.com/Dargones/TestingFrameworkPaper/blob/master/main.pdf) that outlines the overall idea of a testing framework and of how one would do this in Java.

An additional feature that this PR introduces is the `{:fresh}` annotation that prompts the compiler to translate an external method to one that returns a fresh instance of an object. This could be useful in situations where inferring the right parameters to a constructor is difficult. So, for instance, an automatic test generation framework could use this to get a fresh instance of an object to then assign its fields as needed. This is the feature that was previously the topic of [this PR](https://github.com/dafny-lang/dafny/pull/1526).

Finally, this PR also fixes a small issue with the `{:test}` annotation - currently, the C# compiler only calls `IsFailure` on the first returned value of a testing method if there are more than one (should be more than zero) returned value ([this](https://github.com/dafny-lang/dafny/blob/c6e84137226510094fd9079534e967d35f07a3f3/Source/Dafny/Compilers/Compiler-Csharp.cs#L3185) is the line with the bug)


Dargones:
Added a small fix that allows mocked objects to cross-reference each other (the problem was that C#'s out parameters cannot appear in lambda expressions)

robin-aws:
Very exciting! As per our chat I think the general approach seems good and at least viable for other target languages, although I'm sure I'll have suggestions when I have time to review it properly.

FYI here's a good issue to track the mostly orthogonal concern we discussed, about catching when a mock's specification is actually unsatisfiable: https://github.com/dafny-lang/dafny/issues/1767

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1636,13 +1701,21 @@ protected class ClassWriter : IClassWriter {
         }
         cw.StaticMemberWriter.WriteLine(";");
       } else if (cw.CtorBodyWriter == null) {
-        cw.InstanceMemberWriter.Write($"{publik}{konst} {typeName} {name}");
+        if (isPublic && !isConst) {
```
keyboardDrummer, line 1704: Why is the `!isConst` check here?

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1636,13 +1701,21 @@ protected class ClassWriter : IClassWriter {
...
+          cw.InstanceMemberWriter.Write($"{publik}{konst} {typeName} {name}");
+        }
         if (rhs != null) {
           cw.InstanceMemberWriter.Write($" = {rhs}");
         }
         cw.InstanceMemberWriter.WriteLine(";");
       } else {
-        cw.InstanceMemberWriter.WriteLine($"{publik} {typeName} {name};");
+        if (isPublic) {
+          cw.InstanceMemberWriter.Write($"{publik} virtual {typeName} {name} {{get; set;}}");
```
keyboardDrummer, line 1715: I think a read-only property should not have a setter.

Source/Dafny/DafnyPipeline.csproj
```diff
@@ -23,6 +23,7 @@
   <ItemGroup>
       <PackageReference Include="JetBrains.Annotations" Version="2021.1.0" />
       <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.7.0" />
+      <PackageReference Include="Moq" Version="4.16.1" />
```
keyboardDrummer, line 26: What is this used for? I wouldn't expect a `Moq` reference is something other than a testing package.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -72,12 +74,63 @@ public CsharpCompiler(ErrorReporter reporter)
...
+          this.predicate = predicate;
+          this.nArgs = nArgs;
+          currArgs = new object[nArgs];
+          for (int i = 0; i < currArgs.Length; i++) {
+            currArgs[i] = null;
+          }
+          nextArgId = 0;
+        }
+
+        public bool Match(object arg) {
```
keyboardDrummer, line 122: I would rename `arg` to `argument`

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1154,8 +1219,8 @@ protected class ClassWriter : IClassWriter {
       return block;
     }
 
-    static string Keywords(bool isPublic = false, bool isStatic = false, bool isExtern = false) {
-      return (isPublic ? "public " : "") + (isStatic ? "static " : "") + (isExtern ? "extern " : "");
+    static string Keywords(bool isPublic = false, bool isStatic = false, bool isExtern = false, bool isVirtual = true) {
```
keyboardDrummer, line 1222: Where is this new parameter used?

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1154,8 +1219,8 @@ protected class ClassWriter : IClassWriter {
...
+      return (isPublic ? "public " : "") + (isStatic ? "static " : "") + (isExtern ? "extern " : "") + (isVirtual && !isStatic && isPublic ? "virtual " : "");
```
keyboardDrummer, line 1223: Does `isVirtual == true && isStatic == true` ever occur? Consider making this method less smart if these inconsistencies never occur.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -3207,5 +3291,259 @@ private class CSharpCompilationResult {
...
+      private List<Tuple<IVariable, string>> bounds = new();
+
+      public MockWriter(CsharpCompiler compiler) {
+        this.compiler = compiler;
+      }
+
+      /// <summary>
+      /// Create a body for a method returning a fresh instance of an object 
+      /// </summary>
+      public ConcreteSyntaxTree CreateFreshMethod(Method m,
```
keyboardDrummer, line 3334: Please use fully written out variable names in the general case, so `method` here.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -146,6 +199,7 @@ public CsharpCompiler(ErrorReporter reporter)
 
     private void EmitInitNewArrays(BuiltIns builtIns, ConcreteSyntaxTree wr) {
       var dafnyNamespace = CreateModule("Dafny", false, false, null, wr);
+      AddMultiMatcher(dafnyNamespace);
```
keyboardDrummer, line 202: The name `EmitInitNewArrays` doesn't match this content. Consider moving the two statements

```csharp
var dafnyNamespace = CreateModule("Dafny", false, false, null, wr);
AddMultiMatcher(dafnyNamespace);
```

out of `EmitInitNewArrays` and passing `dafnyNamespace` to `EmitInitNewArrays`

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -3207,5 +3291,259 @@ private class CSharpCompilationResult {
...
+    /// 
+    /// SKOLEMS = ID : ID 
+    ///         | SKOLEMS, SKOLEMS
+    /// 
+    /// </summary>
+    private class MockWriter {
+
+      private readonly CsharpCompiler compiler;
+      // maps identifier names to the names of corresponding mocks:
+      private Dictionary<string, string> mockNames = new();
```
keyboardDrummer, line 3323: Consider renaming this to `methodNameToMockName`

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -3207,5 +3291,259 @@ private class CSharpCompilationResult {
...
+        foreach (var o in m.Outs) {
+          retNames[o.CompileName] = FreshId(o.CompileName + "Return");
+          parameterString = Regex.Replace(parameterString,
+            $"(^|[^a-zA-Z0-9_]){o.CompileName}([^a-zA-Z0-9_]|$)",
+            "$1" + retNames[o.CompileName] + "$2");
+        }
+        wr.FormatLine($"{keywords}{returnType} {compiler.IdName(m)}{typeParameters}({parameterString}) {{");
+
+        // Initialize the mocks:
+        mockNames = new Dictionary<string, string>();
```
keyboardDrummer, line 3378: Consider writing 

```csharp
mockNames = m.Outs.ToDictionary(o => o.CompileName, o => compiler.idGenerator.FreshId(o.CompileName + "Mock"));
foreach(var (methodName, mockName) in mockNames) {
  wr.FormatLine($"var {mockName} = new Mock<{compiler.TypeName(o.Type, wr, o.tok)}>();");
  wr.FormatLine($"var {methodName} = {mockName}.Object;");
}
```

to separate the two concerns.

Consider doing something similar for the initialization of `retNames`.

Dargones: Done!

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: Renamed to objectToMockName

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: Done! I also moved `AddMultiMatcher`, now renamed `EmitMultiMatcher` to the `CsharpMockWriter` class.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -72,12 +74,63 @@ public CsharpCompiler(ErrorReporter reporter)
...
+        private readonly Func<object[], bool> predicate;
+        private readonly int nArgs;
+        private readonly object[] currArgs;
+        private int nextArgId;
+
+        public MultiMatcher(int nArgs, Func<object[], bool> predicate) {
+          this.predicate = predicate;
+          this.nArgs = nArgs;
+          currArgs = new object[nArgs];
+          for (int i = 0; i < currArgs.Length; i++) {
```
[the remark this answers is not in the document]
Dargones: Thank you! Have fixed this in the following commit.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -72,12 +74,63 @@ public CsharpCompiler(ErrorReporter reporter)
...
+    private static void AddMultiMatcher(ConcreteSyntaxTree wr) {
+      const string multiMatcher = @"
+      class MultiMatcher {
+
+        private readonly Func<object[], bool> predicate;
+        private readonly int nArgs;
+        private readonly object[] currArgs;
+        private int nextArgId;
+
+        public MultiMatcher(int nArgs, Func<object[], bool> predicate) {
```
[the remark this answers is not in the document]
Dargones: Done!

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -72,12 +74,63 @@ public CsharpCompiler(ErrorReporter reporter)
...
+    /// is converted to:
+    /// 
+    ///   var matcher = new Dafny.MultiMatcher(2, args => {
+    ///     return args[0] > args[1];
+    ///   });
+    ///   o.Setup(x => x.m(It.Is<int>(a => matcher.Match(a)),
+    ///                    It.Is<int>(b => matcher.Match(b)))).Returns(4);
+    /// 
+    /// </summary>
+    private static void AddMultiMatcher(ConcreteSyntaxTree wr) {
```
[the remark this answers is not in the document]
Dargones: Done!

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -72,12 +74,63 @@ public CsharpCompiler(ErrorReporter reporter)
...
+      const string multiMatcher = @"
+      class MultiMatcher {
+
+        private readonly Func<object[], bool> predicate;
+        private readonly int nArgs;
+        private readonly object[] currArgs;
```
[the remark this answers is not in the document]
Dargones: Done.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1636,13 +1701,21 @@ protected class ClassWriter : IClassWriter {
         }
         cw.StaticMemberWriter.WriteLine(";");
       } else if (cw.CtorBodyWriter == null) {
-        cw.InstanceMemberWriter.Write($"{publik}{konst} {typeName} {name}");
+        if (isPublic && !isConst) {
+          cw.InstanceMemberWriter.Write($"{publik}{konst} virtual {typeName} {name} {{get; set;}}");
```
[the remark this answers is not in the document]
Dargones: I now only add virtual if `DafnyOptions.O.CompileMocks` is set to `true`.

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: Made this change for my code. `m` seems to be used instead of `method` throughout the compilers - should I change the other instances as well?

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -3207,5 +3291,259 @@ private class CSharpCompilationResult {
...
+    ///         | SKOLEMS, SKOLEMS
+    /// 
+    /// </summary>
+    private class MockWriter {
+
+      private readonly CsharpCompiler compiler;
+      // maps identifier names to the names of corresponding mocks:
+      private Dictionary<string, string> mockNames = new();
+      // associates a skolem variable with the lambda passed to argument matcher
+      private List<Tuple<IVariable, string>> bounds = new();
```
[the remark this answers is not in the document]
Dargones: I was somewhat worried that `IVariable` does not have a custom `Equals` or `GetHashCode` methods defined for it, so using dictionaries might be risky if different `IVariable` objects are created for different occurrences of a variable in the AST. But, upon testing, dictionaries seem to be working, so this must be consistent with how the AST is constructed.

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: I only ever used the default value, so I removed the parameter.

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: I rewrote this to refer directly to `DafnyOptions.O.CompileMocks`. The check for `isStatic` seems to be necessary for the generated C# code to compile.

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: Thank you for catching this! I don't think this check should be made here.

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: Done

Source/Dafny/DafnyPipeline.csproj
Dargones: Thank you, you are absolutely right - I think Rider added this line automatically when I experimented with `Moq`. I have removed the reference.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -72,12 +74,63 @@ public CsharpCompiler(ErrorReporter reporter)
...
+        public bool Match(object arg) {
+          currArgs[nextArgId++] = arg;
+          if (nextArgId != currArgs.Length) {
+            return true;
+          }
+          nextArgId = 0;
+          return predicate(currArgs);
+        }
+      }";
+      wr.WriteLine(multiMatcher);
```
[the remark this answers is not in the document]
Dargones: It seems that the indentation is still preserved but I wrapped the code around `SyntaxFactory.ParseMemberDeclaration` to be certain. Let me know if there is anything else I should do in this respect!

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1636,13 +1701,21 @@ protected class ClassWriter : IClassWriter {
         }
         cw.StaticMemberWriter.WriteLine(";");
       } else if (cw.CtorBodyWriter == null) {
```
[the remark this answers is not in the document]
Dargones: It seems that the two cases are necessary to distinguish between the situation in which the initialization directly follows the declaration and the situation in which the initialization happens in the constructor. I tried merging these two together but it amounts to an even larger line count because of the differences in how semicolons should be treated, etc, so I would keep the two cases separate, if possible.

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: I tried removing the setter but it causes C# compilation errors. I don't think `virtual` implies read-only, if I understand your concern correctly.

Dargones:
Thank you, @keyboardDrummer for all the feedback! I made the changes you suggested in my latest commits. I also moved the mocking code in a separate file, as you proposed, but am not sure if the file name / location is appropriate. Let me know if there are any other changes I should make!

Source/Dafny/Compilers/Compiler-Csharp.cs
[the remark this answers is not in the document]
keyboardDrummer: I was thinking something like:

```csharp
nding;
if (isPublic) {
  cw.InstanceMemberWriter.Write($"{publik} {virtuall} {typeName} {name} {{get; set;}}");
  ending = "";
} else {
  cw.InstanceMemberWriter.WriteLine($"{publik} {typeName} {name}");
  ending = ";";
}
if (cw.CtorBodyWriter == null) {
  if (rhs != null) {
    cw.InstanceMemberWriter.Write($" = {rhs}");
  }
} else {
  if (rhs != null) {
    cw.CtorBodyWriter.WriteLine($"this.{name} = {rhs};");
  }
}
cw.InstanceMemberWriter.WriteLine(ending);
```

Source/Dafny/Compilers/Compiler-Csharp.cs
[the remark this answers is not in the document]
keyboardDrummer: Sorry, I misspoke. What I meant was that the indentation of the `dafnyNamespace` writer is lost except for on the first line `class MultiMatcher {`. So you could get something like:

```csharp
  class MultiMatcher {
  private readonly Func<object[], bool> predicate;
  private readonly int argumentCount;
  private readonly List<object> collectedArguments;
  public MultiMatcher(int argumentCount, Func<object[], bool> predicate) {
    this.predicate = predicate;
    this.argumentCount = argumentCount;
    collectedArguments = new();
  }
  public bool Match(object argument) {
    collectedArguments.Add(argument);
    if (collectedArguments.Count != argumentCount) {
      return true;
    }
    bool result = predicate(collectedArguments.ToArray());
    collectedArguments.Clear();
    return result;
  }
}
```

keyboardDrummer (dismissed):
Looks great, thanks!

robin-aws (changes requested):
I mainly focussed on the API of this feature from an end-user POV, and made some suggestions that hopefully won't change the actual implementation very much.

One other top-level suggestion: as you mention, this style of mocking really is a form of synthesis, and it's entirely possible other target languages will want to implement this with more code generation and not rely on any mocking libraries. The term "mock" has a fairly specific meaning in testing libraries and given the current limitations of the implementation this feels more like "stubs" instead. I would consider naming the attribute something more generic, and allowing the Dafny code that uses it to label it based on usage instead.

Test/DafnyTests/TestAttribute.dfy
```diff
@@ -67,3 +67,171 @@ method {:test} PassingTestUsingNoLHSAssignOrHalt() {
...
+method {:test} PassingTestMockSumAsMultiplication() {
+    var e:Even := MockSumAsMultiplication();
+    expect(e.Sum(3, 2) == 6);
+    expect(e.Sum(3, 0) == 0);
+}
+
+method {:extern} {:mock} MockSumWithArgumentMatcher() returns (e:Even) 
+    ensures fresh(e) 
+    ensures forall a:int, b:int :: (b < a) ==> (e.Sum(a, b) == a * b)
+    ensures forall a:int, b:int :: (b >= a) ==> (e.Sum(a, b) == -a * b)
```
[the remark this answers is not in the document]
ericmercer: I'm not sure forcing the mock to obey the original post-conditions is consistent with what we want the mocks to be able to do. 

Mocks are _not_ intended to be implementations of the original methods; rather, they provide some limited behavior, usually inconsistent with the original method behavior, to achieve some testing goal.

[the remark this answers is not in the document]
ericmercer: Might we include a switch to enable or disable the addition of the original contract for the mocked method?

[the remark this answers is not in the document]
robin-aws: Thanks for the helpful discussion on this yesterday! I still believe that using a mock to drive code to behave in a way that is impossible otherwise is questionable (such as reaching a statement that is otherwise unreachable), but I can appreciate that there is at least some use for this for debugging.

Therefore I am adequately mollified if we add a switch for including the original contract, AND we make it on by default, with a warning in the documentation that turning it off can lead to unsound results.

Test/DafnyTests/TestAttribute.dfy
```diff
@@ -67,3 +67,171 @@ method {:test} PassingTestUsingNoLHSAssignOrHalt() {
...
+    expect(e.IsValid());
+}
+
+method {:test} FailingTestUsingFresh() {
+    var e:Even := freshEven();
+    e.value := 5;
+    expect(e.IsValid());
+}
+
+method {:extern} {:mock} MockValidEven() returns (e:Even)
```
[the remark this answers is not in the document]
Dargones: Done! You no longer have to add the `:extern` attribute on methods annotated with `:mock` (now called `:synthesize`).

Source/Dafny/DafnyOptions.cs
```diff
@@ -1056,6 +1060,9 @@ not sound
         /verificationLogger:trx;LogFileName=<...>.
     The exact mapping of verification concepts to the TRX and CSV formats is
     experimental and subject to change!
+/compileMocks 
+    If true, automatically compile external methods annotated with {{:mock}}
+    into the target language (currently only supported for C#)
```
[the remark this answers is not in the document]
Dargones: I have updated Attributes.md in the Reference Manual with a brief description of the `:synthesize` attribute (I renamed `:mock` to this).

My reasoning behind adding a command line option (now also renamed to `:synthesize`) is that synthesizing mocks is only possible with the Moq library, which we probably shouldn't require by default since it is not part of C#'s standard library. This command-line option would tell the compiler that it is OK to include Moq. 

Alternatively, one can traverse the program's AST and only try to import Moq if there is a method annotated with `:mock`. The problem with this approach is that most of the compiler code is agnostic to the program currently being compiled. This would mean that I would have to compile all fields as virtual (necessary for using Moq) even if no mocking is actually performed because I don't know which program I am compiling at the moment. @keyboardDrummer  previously mentioned that doing so might slow done the resuling binary, so I went for the command-line-argument approach. But I do have both approaches implemented locally, so I can easily make the change. Let me know if I should redo this one way or the other!

Test/DafnyTests/TestAttribute.dfy
```diff
@@ -67,3 +67,171 @@ method {:test} PassingTestUsingNoLHSAssignOrHalt() {
...
+    function method Next():Even? {
+        null // not implemented yet but can be mocked
+    }
+    
+    function method Sum(a:int, b:int):int {
+        a + b
+    }
+}
+
+method {:extern} {:fresh} freshEven() returns (e:Even) ensures fresh(e)
```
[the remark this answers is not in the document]
Dargones: I didn't realize one can do this before, but I was able to merge these two cases! From now on, a mocked object behaves exactly as the real object in all cases except for when a function call or field is explicitly stubbed.

Test/DafnyTests/TestAttribute.dfy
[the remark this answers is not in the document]
Dargones: As per the discussion on Slack, I have made it so that Dafny spits an error message if a stubbed method includes any post-postconditions (or if the code tries to mock an object's field) - see the updated test files for an example.

Dargones:
Thank you, @robin-aws, for all your feedback! I changed "mock" to "synthesize" to make the attribute name more generic and have made other changes you suggest above. Let me know if there is anything else you would like me to change!

docs/DafnyRef/Attributes.md
```diff
@@ -360,6 +360,39 @@ They are translated into Boogie triggers.
 The `{:typeQuantifier}` attribute must be used on a quantifier if it
 quantifies over types.
 
+### 22.1.21. mock
```
robin-aws, line 363:
```suggestion
### 22.1.21. synthesize
```

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1634,23 +1648,32 @@ protected class ClassWriter : IClassWriter {
...
-          cw.InstanceMemberWriter.Write($" = {rhs}");
-        }
-        cw.InstanceMemberWriter.WriteLine(";");
       } else {
-        cw.InstanceMemberWriter.WriteLine($"{publik} {typeName} {name};");
-        if (rhs != null) {
-          cw.CtorBodyWriter.WriteLine($"this.{name} = {rhs};");
+        string ending = "";
+        if (isPublic) {
+          cw.InstanceMemberWriter.Write($"{publik} {virtuall} {typeName} {name} {{get; set;}}");
```
robin-aws, line 1661: It looks like you're ignoring `konst` here and emitting a setter even if `isConst` is true. Is that intentional?

Source/Dafny/Compilers/Compiler.cs
```diff
@@ -1273,6 +1275,15 @@ protected class NullClassWriter : IClassWriter {
       public ConcreteSyntaxTree/*?*/ CreateMethod(Method m, List<TypeArgumentInstantiation> typeArgs, bool createBody, bool forBodyInheritance, bool lookasideBody) {
         return createBody ? block : null;
       }
+
+      public ConcreteSyntaxTree CreateFreshMethod(Method m) {
```
robin-aws, line 1279: I think this is leftover from earlier revisions and can be deleted.

Source/Dafny/Compilers/Compiler.cs
```diff
@@ -1784,8 +1795,18 @@ protected class NullClassWriter : IClassWriter {
...
         } else if (member is Method m) {
-          if (m.Body == null && !(c is TraitDecl && !m.IsStatic) &&
-              !(!DafnyOptions.O.DisallowExterns && (Attributes.Contains(m.Attributes, "dllimport") || (IncludeExternMembers && Attributes.Contains(m.Attributes, "extern"))))) {
+          if (Attributes.Contains(m.Attributes, "synthesize") &&
+              DafnyOptions.O.Synthesize) {
+            if (m.IsStatic && m.Outs.Count > 0 && m.Body == null) {
+              classWriter.SynthesizeMethod(m, CombineAllTypeArguments(m), true, true, false);
+            } else {
+              Error(m.tok, "Method {0} is annotated with :synthesize but " +
+                           "does not have the correct specification or has a " +
```
robin-aws, line 1804: Can you be more specific about what "the correct specification" is (i.e. static and returning at least one value) in the error message?

Source/Dafny/Compilers/Synthesizer-Csharp.cs
```diff
@@ -0,0 +1,337 @@
...
+  ///     e1Return = e1;
+  ///     e2Return = e2;
+  /// }
+  /// </summary>
+  public ConcreteSyntaxTree SynthesizeMethod(Method method,
+    List<Compiler.TypeArgumentInstantiation> typeArgs, bool createBody,
+    ConcreteSyntaxTree wr, bool forBodyInheritance, bool lookasideBody) {
+
+    lastSynthesizedMethod = method;
+    // The following few lines are identical to those in CreateMethod above:
```
robin-aws, line 75:
```suggestion
    // The following few lines are identical to those in Compiler.CreateMethod:
```

Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
```diff
@@ -1942,6 +1942,8 @@ public ExpressionTranslator(ExpressionTranslator etran, string modifiesFrame)
             name = "timeLimit";
           } else if (name == "_rlimit") {
             name = "rlimit";
+          } else if (name == "synthesize") {
+            name = "extern";
```
robin-aws, line 1946: Hey that was even easier than I expected. :)

Test/DafnyTests/TestAttribute.dfy
```diff
@@ -67,3 +67,205 @@ method {:test} PassingTestUsingNoLHSAssignOrHalt() {
...
+    var e1:Even, e2:Even := CrossReferentialMock();
+    expect(e1.Next() == e2);
+    expect(e1.Next() != e1);
+    expect(e2.Next() == e1);
+    expect(e2.Next() != e2);
+}
+
+class StringMap {
+
+	var m:map<string, string>;
```
robin-aws, line 244: nit: the indentation on this file is all over the place. :) We've generally settled on a two-space indent.

Test/DafnyTests/TestAttribute.dfy
```diff
@@ -1,4 +1,4 @@
-// RUN: %dafny /compileVerbose:1 /compile:0 /spillTargetCode:3 /noVerify "%s" > "%t"
+// RUN: %dafny /compileVerbose:1 /compile:0 /spillTargetCode:3 /noVerify /synthesize "%s" > "%t"
 // RUN: dotnet test -v:q -noLogo %S 2> %t.testresults.raw || true
 // Remove the timestamp prefixes on the expected errors
 // RUN: sed 's/[^]]*\]//' "%t".testresults.raw >> "%t"
```
robin-aws, line 4: Off-topic, but I just realized this post-processing can be replaced with OutputCheck-style `// CHECK:` directives now that I've added that support to the xUnit lit test runner :)

Source/Dafny/DafnyOptions.cs
```diff
@@ -1116,6 +1120,9 @@ not sound
         /verificationLogger:trx;LogFileName=<...>.
     The exact mapping of verification concepts to the TRX and CSV formats is
     experimental and subject to change!
+/synthesize 
+    If true, synthesizes bodies of methods annotated with  {{:synthesize}}
+    during compilation (currently only supported for C#)
```
robin-aws, line 1125: > My reasoning behind adding a command line option (now also renamed to :synthesize) is that synthesizing mocks is only possible with the Moq library, which we probably shouldn't require by default since it is not part of C#'s standard library. This command-line option would tell the compiler that it is OK to include Moq.

> Alternatively, one can traverse the program's AST and only try to import Moq if there is a method annotated with :mock. The problem with this approach is that most of the compiler code is agnostic to the program currently being compiled. This would mean that I would have to compile all fields as virtual (necessary for using Moq) even if no mocking is actually performed because I don't know which program I am compiling at the moment. @keyboardDrummer previously mentioned that doing so might slow done the resuling binary, so I went for the command-line-argument approach. But I do have both approaches implemented locally, so I can easily make the change. Let me know if I should redo this one way or the other!

Okay - I would still prefer not to have the extra option, but the point about virtual fields being a potential performance hit is the most valid reason to keep it. If it's feasible to instead scan ahead for any `{:synthesize}` attributes and only then make fields virtual and add the `using` declaration, that would be great, but it's not a hard requirement.

robin-aws (changes requested):
Great changes, I'm super happy we were able to simplify the surface area of the feature! Just some relatively minor points to address.

Can you add a bullet point to `RELEASE_NOTES.md` for this feature too?

Source/Dafny/Compilers/Compiler.cs
Dargones: Done!

Source/Dafny/Compilers/Compiler.cs
Dargones: Changed to "...is not static, has a body, or does not return anything"

Test/DafnyTests/TestAttribute.dfy
Dargones: Have changed everything to two-space indent.

Source/Dafny/Compilers/Compiler-Csharp.cs
Dargones: I borrowed this from @keyboardDrummer's proposed [solution](https://github.com/dafny-lang/dafny/pull/1809#discussion_r810911776), which does seem to emit the setter regardless. In my latest commit I have changed the code so that it only emits it if `isConst` is false, as below. Would this be the correct way of doing it?

```csharp
if (isPublic) {
  if (isConst) {
    cw.InstanceMemberWriter.Write($"{publik}{konst} {virtuall} {typeName} {name} {{get;}}");
  } else {
    cw.InstanceMemberWriter.Write($"{publik} {virtuall} {typeName} {name} {{get; set;}}");
   }
} else {
  cw.InstanceMemberWriter.WriteLine($"{publik}{konst} {typeName} {name}");
  ending = ";";
}
```

Test/DafnyTests/TestAttribute.dfy
Dargones: I am now using the `//CHECK:` directive

Source/Dafny/DafnyOptions.cs
Dargones: I now scan the AST for `{:synthesize}` attribute instead of relying on the command line argument as a result. The result of the scan is stored in an instance variable that the compiler checks whenever it makes a decision about making fields virtual. A possible downside of this approach is potential issues when trying to parallelize compilation of multiple programs using a single compiler object (is such a situation likely?). In any case, this is my final commit at the moment so I can easily revert it back to the command-line-argument approach if needed!

Dargones:
Thank you, @robin-aws! I have added the feature to `RELAEASE_NOTES.md` and made all the other changes you have suggested.

Source/Dafny/DafnyOptions.cs
robin-aws: You're already ahead of the game by not using a global variable, so I'm totally fine with a compiler instance variable. :)

robin-aws (approved):
Looks great! Thanks for all the improvements on this.

One last ask for @Dargones: can you update the PR description to reflect the fact that we ended up dropping the CLI option and the `{:fresh}` attribute? We'll use that content as the commit description when we merge and squash, and I wasn't 100% sure how you'd like to update it.

Dargones:
Thank you, @robin-aws! I updated the PR description by adding a paragraph at the top that summarizes all the key differences with the original version.

PR #1810, opened 2022-02-10
Make resolved match cases know if they originate from variable or constructor/literal patterns

amaurremi:
The Dafny compiler desugars pattern matching in a way that has
an explicitly generated case for each constructor. I.e.,
```
var x: Option<bool>;
match x
  case Some(true) => s1;
  case _ => s2;
```
gets resolved as follows:
```
match x
  case Some(b) => if b { s11; } else { s2; }
  case None => s2;
```
This commit allows us to know that the first case arose from
an explicit constructor invocation whereas the latter came from
a bound variable.


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -8784,20 +8784,31 @@ public MatchStmt(IToken tok, IToken endTok, Expression source, [Captured] List<M
...
+    // Has the case for this constructor been generated by the resolver because the pattern was
+    // a bound variable, or was it an explicit constructor case in the source code?
+    // E.g. 
+    // var x: Option<bool>;
+    // match x
+    //   case Some(true) => ... // FromBoundVar == false
+    //   case Some(_)    => ... // FromBoundVar == false
+    //   case v          => ... // FromBoundVar == true
+    //   case _ =>       => ... // FromBoundVar == true (this case would be unreachable; added for illustration purposes)
+    public readonly bool FromBoundVar;
```
cpitclaudel, line 8796: Is this variable true for all auto-generated cases, or just a subset of them?

Source/Dafny/Resolver.cs
```diff
@@ -12223,6 +12223,8 @@ private class RBranchExpr : RBranch {
         mti.UpdateBranchID(PB.Item2.BranchID, ctors.Count - 1);
       }
 
+      var ctorToFromBoundVar = new Dictionary<string, bool>();
```
cpitclaudel, line 12226: You can use a set here instead of a dictionary

Source/Dafny/Resolver.cs
```diff
@@ -12257,6 +12259,9 @@ private class RBranchExpr : RBranch {
               currBranch.Patterns.InsertRange(0, currPattern.Arguments);
               currBranches.Add(currBranch);
               ctorCounter++;
+              if (!ctorToFromBoundVar.ContainsKey(ctor.Key)) {
+                ctorToFromBoundVar.Add(ctor.Key, false);
+              }
```
cpitclaudel, line 12264: … then you would remove this code

Source/Dafny/Resolver.cs
```diff
@@ -12279,6 +12284,9 @@ private class RBranchExpr : RBranch {
               currBranch.Patterns.InsertRange(0, freshArgs);
               LetBindNonWildCard(currBranch, currPattern, rhsExpr);
               currBranches.Add(currBranch);
+              if (!ctorToFromBoundVar.ContainsKey(ctor.Key)) {
+                ctorToFromBoundVar.Add(ctor.Key, true);
+              }
```
cpitclaudel, line 12289: And here you would add to the set

cpitclaudel (approved):
👍 works for me, I left a few nits on the variable's name and on the implementation

Source/Dafny/Resolver.cs
amaurremi: Good point.

Source/Dafny/AST/DafnyAst.cs
amaurremi: If you have
```
datatype X = A(x: X) | B(x: X) | C
match A(A(C))
  case A(A(x))) => 1
  case A(y) => 2
  case _ => 3 
```
then I believe Dafny will generate
```
datatype X = A(a: X) | B(b: X) | C
match A(A(C))
  case A(a) =>               // 1
    match a
      case A(a') =>          // 2
        match a'
          case A(a'') => 2   // 3
          case B(b'') => 2   // 4
          case C => 1        // 5
      case B(b) => 2         // 6
      case C => 2            // 7
  case B(b) => 3             // 8
  case C =>  3               // 9
```
I want to capture all the cases that capture `x` or `y`. So `FromBoundVar` should be true for (3-5) because those are captured by `x`, and (6-9) because those are captured by `y`.

amaurremi: Essentially I'd like to capture all match cases that originate from a bound variable in the user's match case. Are you thinking of some specific example?

cpitclaudel: Wonderful, thanks!

cpitclaudel (dismissed):
LGTM, will leave open until next week to give others a chance to chime in if they want to, but then we can merge!

keyboardDrummer (changes requested):
I see you've documented the meaning of the field `FromBoundVar`. Can you also document the purpose? The field `FromBoundVar` is never read so after merging this PR there's nothing preventing someone from removing it again.

amaurremi:
> I love the documentation you added on the meaning of the field `FromBoundVar`. Can you also document that field's purpose? The field `FromBoundVar` is never read so after merging this PR there's nothing preventing someone from removing it again.

Absolutely, done.

> I'm assuming you have some transformation that uses `FromBoundVar`. Is it not possible to run this transformation before the `_` case desugaring?

Some transformations need to be done on the resolved AST which stores a lot of information not present in the parsed AST. It's very difficult to establish a correspondence between the two when it comes to pattern matching, especially when it's nested (i.e. when there's matching on nested constructors which gets resolved into nested pattern matching).

> I think there's a token attached to the generated `MatchCase` objects. Could you use that to determine whether they originated from a `_` case ?

This is a good question, and I did try that at first. But from what I remember the resolved AST creates new nodes that don't store precise token information. The tokens might correspond simply to the start of the pattern match, for example.

amaurremi:
Technically I could try to change the way that tokens are generated, and try to make that information more precise. But I'm hesitant about trying to depend on tokens for establishing relationships between AST nodes--e.g., if someone changes the way tokens are generated, they won't know that we used tokens for inferring information about the program in this way, and we won't know that this change happened, either.

keyboardDrummer:
> Technically I could try to change the way that tokens are generated, and try to make that information more precise. But I'm hesitant about trying to depend on tokens for establishing relationships between AST nodes--e.g., if someone changes the way tokens are generated, they won't know that we used tokens for inferring information about the program in this way, and we won't know that this change happened, either.

I think that's a healthy hesitations. 

One more question, have you looked into turning off the desugaring of match statements?

keyboardDrummer (approved):

keyboardDrummer:
> Resolving pattern matching includes a number of type/sanity checks (like exhaustiveness) that rely on the pattern matching desugaring/compilation algorithm. 

Is that not strange? Shouldn't exhaustiveness checking be easy in the face of `case _` statements? Sorry I'm afraid I don't know the resolver well. I can try to answer this question myself if that helps.

Let me try to explain why I'm hesitant about merging. To me this PR is doing more than adding one variable, it's also setting a precedent for keeping old information present when making simplifying transformations, which partly defeats the purpose of a simplification.

fabiomadge (changes requested):
Should we add something to `RELEASE_NOTES.md`, or is this too deep?

fabiomadge (approved):

PR #1813, opened 2022-02-10
fix: Fix CanCall assumptions for loop invariants

### LINKED ISSUE #1812: CanCall for loop invariant applied incorrectly
Every expression in Dafny is checked for well-formedness before it is used. As a kind of "certificate" that the well-formedness check has been done, Dafny emits so-called CanCall assumptions. These CanCall assumptions are not generated correctly at the end of loop bodies.

## Details

In more detail, consider a loop with loop invariants

``` dafny
  invariant A
  invariant B
```

The well-formedness checks for these essentially look like

```
CheckWellformedness[[ A ]];
if A {
  CheckWellformedness[[ B ]];
}
```

This is also what is done for the well-formedness checks for a single `invariant` declaration with a conjunction (that is, `invariant A && B`). Note that the well-formedness of `B` is checked under the assumptions that `A` is well-formed and evaluates to `true`. This is important, because `A` often includes conditions that are needed in order for `B` to be well-formed. CanCall assumptions must follow the same pattern.

Alas, Dafny currently generates

```
EmitCanCallAssumptions[[ A ]];
EmitCanCallAssumptions[[ B ]];
```

where it should be generating

```
EmitCanCallAssumptions[[ A ]];
if A {
  EmitCanCallAssumptions[[ B ]];
}
```

This happens when there are multiple `invariant` declarations (`invariant A && B` is already treated correctly).

## Repro

It is surprisingly difficult to find an example that exhibits this unsoundness. The program must be one that does not maintain the invariant `A`, and furthermore the situation of having the CanCall assumptions for `B` in a state where `!A` holds must be exploitable to derive `false`. I've only been able to do this with a program that uses sets where the loop invariant uses a let-such-that expression (see below), and even small variations (in either Dafny or Boogie) of that program sometimes can fail to find the exploit (in which case an error is reported, so the unsoundness does not manifest).

The following program shows the unsoundness:

``` dafny
function Max(s: set<int>): (m: int)
  requires s != {}
{
  var x :| x in s;
  if s == {x} then
    x
  else
    var s' := s - {x};
    var y := Max(s');
    y
}

method IncorrectLoop(m: int)
{
  var r := {m};
  while r != {}
    invariant r != {}
    invariant m == Max(r)
  {
    r := r - {m};
  }
}
```

The verifier should report that loop invariant `r != {}` is not maintained, but it currently reports no errors at all.

RustanLeino:
The CanCall assumptions generated at the end of a loop body (to support the expressions in the loop invariant to be proved) were not generated correctly. In particular, when there are multiple `invariant` declarations, the CanCall assumptions were not given the right antecedents. (See issue #1812 for more details.)

Fixes #1812


Source/Dafny/Verifier/Translator.TrStatement.cs
```diff
@@ -1910,10 +1910,14 @@ public partial class Translator {
...
       // Finally, assume the well-formedness of the invariant (which has been checked once and for all above), so that the check
-      // of invariant-maintenance can use the appropriate canCall predicates.
+      // of invariant-maintenance can use the appropriate canCall predicates. Note, it is important (see Test/git-issues/git-issue-1812.dfy)
+      // that each CanCall assumption uses the preceding invariants as antecedents--this is achieved by treating all "invariant"
+      // declarations as one big conjunction, because then CanCallAssumption will add the needed antecedents.
+      Expression allInvariants = Expression.CreateBoolLiteral(s.Tok, true);
       foreach (AttributedExpression loopInv in s.Invariants) {
-        loopBodyBuilder.Add(TrAssumeCmd(loopInv.E.tok, CanCallAssumption(loopInv.E, etran)));
+        allInvariants = Expression.CreateAnd(allInvariants, loopInv.E);
       }
```
cpitclaudel, line 1919: This might be marginally more readable as (untested)

```
s.Invariants.Select(inv => inv.E).Aggregate(Expression.CreateAnd, Expression.CreateBoolLiteral(s.Tok, true))
```

Although I'd further simplify it by changing it to:

```
if (invariants.Any()) {
  var allInvariants = s.Invariants.Select(inv => inv.E).Aggregate(Expression.CreateAnd);
  loopBodyBuilder.Add(…);
}
```

to void the extra `true` at the end.

cpitclaudel (dismissed):
Approving since the code looks good; left a comment.

Source/Dafny/Verifier/Translator.TrStatement.cs
RustanLeino: Thanks. I followed your suggestion. (Annoyingly, `CreateAnd` has a third parameter, with a default value. So, I had to eta convert the argument to `Aggregate`. Still, the result is nicer than the previous loop.)

cpitclaudel (dismissed):

cpitclaudel (approved):

PR #1816, opened 2022-02-11
fix(lsp): Allow range of text document modification to be null

cpitclaudel:
Without this the LSP server crashes every time a client sends a null range, which happens every few minutes in Emacs.


MikaelMayer (dismissed):
Simple change, well documented, tested, up to the point, very nice catch !

camrein (changes requested):
It makes sense to relax this pre-condition. However, I've got the feeling that the symbol and diagnostic relocators will be affected by this change as well.

Diagnostics migration:
https://github.com/dafny-lang/dafny/blob/ec8a2feea8f817b86ecbf65f5ee85ea474eb070e/Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs#L53-L59

Symbol table migration:
https://github.com/dafny-lang/dafny/blob/ec8a2feea8f817b86ecbf65f5ee85ea474eb070e/Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs#L89-L111

They should fail if you either cancel out the change with a newer one or (easier to reproduce) provide a document that has syntactic errors.

cpitclaudel:
Thanks @camrein, I pushed further changes

Source/DafnyLanguageServer.Test/Synchronization/DeclarationLocationMigrationTest.cs
```diff
@@ -279,5 +279,22 @@ var a
       Assert.AreEqual(new Range((6, 6), (6, 7)), yLocation.Name);
       Assert.AreEqual(new Range((6, 6), (6, 7)), yLocation.Declaration);
     }
+
+    [TestMethod]
+    public async Task NullRangeClearsSymbolsTable() {
```
camrein, line 284: Make the test method name describe what the test does and what's the expected outcome.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -52,29 +52,30 @@ CancellationToken cancellationToken
 
       private IReadOnlyList<Diagnostic> MigrateDiagnostics(IReadOnlyList<Diagnostic> originalDiagnostics, TextDocumentContentChangeEvent change) {
         if (change.Range == null) {
-          throw new System.InvalidOperationException("the range of the change must not be null");
+          return new List<Diagnostic>();
```
camrein, line 55: Many build warnings within this file.

camrein (changes requested):
Great tests 👍 
I suggest changing the test-method names so they depict the action and the expected outcome. Then it's easier to judge if the test is behaving correctly/according to its intent when changing some of the logic in the future .

camrein (changes requested):
Thanks a lot for the improvements, they look fantastic.
I've been giving the tests a further thought regarding your changes. If I followed your changes correctly, it leaves the symbols located in a foreign document unchanged, i.e., it only drops the symbols of the current document. This absolutely makes sense and I think we should add a test to ensure it stays like this.

camrein:
The test looks good. However, I'd prefer having two separate tests since there are two unique goals: (1) Clear the symbol table from all entries belonging to the updated document (2) leave symbols in the table that are from foreign documents
I'm fine with the current solution, so feel free to include my suggestion or not. I'm primary afraid of that the test cases grow exceptionally large and become harder to debug when they test for multiple things at once.

camrein (dismissed):

cpitclaudel:
Done, thanks.

cpitclaudel:
I wonder what caused ChangeDocumentCancelsPreviousOpenAndChangeVerification to fail.  @camrein, do you think it could be related to this?  Or is it a concurrency issue?

camrein:
That's a strange error. Sounds like a race condition. What I'm confused about is the following: The test checks with each call to `AwaitNextDiagnosticsAsync` that document has the expected version. That works for the first and second diagnostics, having the document version `1` and `2`. The third invocation correctly expects `3`, but the language server sends diagnostics for version `2`.
This situation is probably reasoned to #1666: Resolver and parser diagnostics are now sent before the verification completed; verifier diagnostics are sent once they're available.
I believe the race condition we have here is that the resolution finishes before it's canceled. According to [this code](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs#L80), verifier diagnostics are sent even if the verification was canceled. Therefore, it may happen that two diagnostics are sent, if the resolution finishes before the cancellation was handled.

cpitclaudel:
> I believe the race condition we have here is that the resolution finishes before it's canceled. 

Hmm, I'm not sure we cancel resolution in this test — in fact we seem to have some code specifically to avoid this issue.

```
      // Wait for resolution diagnostics now, so they don't get cancelled.
      // After this we still have slow verification diagnostics in the queue.
      var parseErrorFixedDiagnostics = await diagnosticReceiver.AwaitNextDiagnosticsAsync(CancellationTokenWithHighTimeout, documentItem);
      Assert.AreEqual(0, parseErrorFixedDiagnostics.Length);
```

I'm having trouble reproducing this locally :/

camrein:
Ahh I see, I have some talent to read over comments I guess. 😬 
https://github.com/dafny-lang/dafny/blob/cf0427b87715fd358cbbffed21d51defbeb58b17/Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs#L152-L153
This code intends to cancel the verification. But if I follow the the test implementation correctly, I think the verification isn't supposed to complete before the diagnostics of the change arrive. Maybe that's what happens here?

cpitclaudel:
> I think the verification isn't supposed to complete before the diagnostics of the change arrive.

I don't understand what you mean :/

(Thanks for the debugging help btw!)

camrein:
Sorry, I hope I can clarify myself a bit.
The diagnostics are sent with the version of the document they belong (versioned according to the changes). So diagnostics will come in the order 1, 2, 3.
However, since the PR I've mentioned now sends diagnostics directly after completing resolution and not awaiting the verification to complete, it'll look something like this: 1, 1, 2, 2, 3, 3
Or more precisely:
```
1 w/o verification
1 w/ verification
2 w/o verification
2 w/ verification
3 w/o verification
3 w/ verification
```

If I follow the code of the test correctly, I have the feeling there are some race conditions that rely on the fact that the cancellation does not work properly and is unreliable. For example, this code:
https://github.com/dafny-lang/dafny/blob/cf0427b87715fd358cbbffed21d51defbeb58b17/Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs#L145-L150
It seems to expect that the diagnostics for the change come in before the resolution. You can't be sure of that if cancelling a verification would be quick. It's not exactly the line the test case failed, but I think that's the main issue.

cpitclaudel:
I have a repro for that bug (it's not due to this branch), so I propose we merge this and debug that other bug later.

MikaelMayer (approved):
Straightforward to accept, well done covering this with tests.

PR #1817, opened 2022-02-11
fix: Fixed the escaper to simplify the plugin syntax

MikaelMayer:
This PR makes the plugin syntax with arguments and escape simpler by only unescaping plugin arguments.
It leaves to the caller the responsibility of providing the right arguments.

### DIFF
--- a/Source/Dafny/DafnyOptions.cs
+++ b/Source/Dafny/DafnyOptions.cs
@@ -238,15 +238,6 @@ namespace Microsoft.Dafny {
         case "plugin": {
             if (ps.ConfirmArgumentCount(1)) {
               var pluginAndArgument = args[ps.i];
-              if (pluginAndArgument.Length > 0 &&
-                  pluginAndArgument[0] == '"' &&
-                  pluginAndArgument[^1] == '"'
-                  ) {
-                var unescapeRegex = new Regex(@"\\""|\\\\");
-                pluginAndArgument = unescapeRegex.Replace(pluginAndArgument.Substring(1, pluginAndArgument.Length - 2),
-                  match => match.Groups[0].Value == @"\""" ? "\"" : @"\"
-                );
-              }
               if (pluginAndArgument.Length > 0) {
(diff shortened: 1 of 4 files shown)


Source/Dafny/DafnyOptions.cs
```diff
@@ -255,11 +246,13 @@ public enum IncludesModes {
                   // There are no commas in paths, but there can be in arguments
                   var argumentsString = string.Join(',', pluginArray.Skip(1));
                   // Parse arguments, accepting and remove double quotes that isolate long arguments
-                  var splitter = new Regex(@"""((?:[^""]|\\"")*)""|([^ ]+)");
+                  var splitter = new Regex(@"""((?:[^""\\]|\\\\|\\"")*)""|([^ ]+)");
```
camrein, line 249: Regular expressions with these many escaping always make me feel uncomfortable. 😬 
I'd suggest to extract this whole unescaping procedure into a separate method. Moreover, as I'm always in favor of expressive names I suggest using named capturing groups rather than numeric. Then it's easier to follow what you're actually trying to capture.

Source/DafnyLanguageServer.Test/Various/PluginsAdvancedTest.cs
```diff
@@ -19,7 +19,7 @@ public class PluginsAdvancedTest : PluginsTestBase {
     "PluginsAdvancedTest";
 
   protected override string[] CommandLineArgument =>
-    new[] { $@"--dafny:plugins:0=""{LibraryPath},force you""" };
+    new[] { $@"--dafny:plugins:0={LibraryPath},force you" };
```
camrein, line 22: I guess `@` is no longer necessary here.

Source/Dafny/DafnyOptions.cs
MikaelMayer: I can't imagine what this regex would look like if we did not have the raw symbol "@" :-)

PR #1818, opened 2022-02-12
fix: Change datatype deconstruction in match statements

### LINKED ISSUE #1815: Bad cast involving variant `datatype`s
Running the compiled version of this program results in a failed cast. The cast is emitted to support the match statement.
```Dafny
datatype X<+U> = X(x: U)

trait Tr {}
class Cl extends Tr {
    constructor () {}
}

method Main() {
    var cl := new Cl();
    var e: X<Tr> := X(cl);
    match e
    case X(tr) => return;
}
```

fabiomadge:
Fixes #1815 by replacing a manual datatype deconstruction by the one we already export though the interface. This made exporting unnamed destructors necessary.


Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -642,14 +642,17 @@ bool ContainsTyVar(TypeParameter tp, Type ty)
...
+        ConcreteSyntaxTree interfaceTree, string DtT_TypeArgs) {
       foreach (var ctor in dt.Ctors) {
-        foreach (var dtor in ctor.Destructors) {
+        foreach (var (dtor, index) in ctor.Destructors.Select((v, i) => (v, i))) {
           if (dtor.EnclosingCtors[0] == ctor) {
             var arg = dtor.CorrespondingFormals[0];
-            if (!arg.IsGhost && arg.HasName) {
+            if (!arg.IsGhost) {
+              var name = arg.HasName ? arg.CompileName : FormalName(arg, index);
+              var dtor_name = $"dtor_{(arg.HasName ? "" : ctor.CompileName)}{name}"
```
keyboardDrummer, line 652: Consider renaming to `dtorName` or `destructorName`

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -642,14 +642,17 @@ bool ContainsTyVar(TypeParameter tp, Type ty)
     }
 
     private void CompileDatatypeDestructorsAndAddToInterface(DatatypeDecl dt, ConcreteSyntaxTree wr,
-      ConcreteSyntaxTree interfaceTree, string DtT_TypeArgs) {
+        ConcreteSyntaxTree interfaceTree, string DtT_TypeArgs) {
       foreach (var ctor in dt.Ctors) {
-        foreach (var dtor in ctor.Destructors) {
+        foreach (var (dtor, index) in ctor.Destructors.Select((v, i) => (v, i))) {
```
keyboardDrummer, line 647: Funny looking Select expression ^^

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -2614,8 +2617,7 @@ protected class ClassWriter : IClassWriter {
     }
 
     protected override void EmitDestructor(string source, Formal dtor, int formalNonGhostIndex, DatatypeCtor ctor, List<Type> typeArgs, Type bvType, ConcreteSyntaxTree wr) {
-      var dtorName = FormalName(dtor, formalNonGhostIndex);
-      wr.Write("(({0}){1}{2}).{3}", DtCtorName(ctor, typeArgs, wr), source, ctor.EnclosingDatatype is CoDatatypeDecl ? "._Get()" : "", dtorName);
+      wr.Write($"{source}.dtor_{(dtor.HasName ? dtor.CompileName : ctor.CompileName + FormalName(dtor, formalNonGhostIndex))}");
```
keyboardDrummer, line 2620: Could you extract the {..} expression into a variable to make the interpolated string easier to read?

Source/Dafny/Compilers/Compiler-Csharp.cs
keyboardDrummer: Consider replacing 
`$"dtor_{(arg.HasName ? "" : ctor.CompileName)}{name}"` 
with
`$"dtor_{ctor.CompileName}{(arg.HasName ? "" : FormalName(arg, index))}"`

I found it hard to read the former since mentally I inline `name`, given that variable name doesn't tell me much, and then it becomes 
`$"dtor_{(arg.HasName ? "" : ctor.CompileName)}{arg.HasName ? arg.CompileName : FormalName(arg, index)}"`

fabiomadge: Those aren't equivalent.

keyboardDrummer: `$"dtor_{(arg.HasName ? "" : ctor.CompileName)}{arg.HasName ? arg.CompileName : FormalName(arg, index)}"`

if `arg.HasName` is true that becomes:
`$"dtor_{""}{arg.CompileName}"` => `$"dtor_{arg.CompileName}"` 

and otherwise:
`$"dtor_{ctor.CompileName}{FormalName(arg, index)}"`

So.. you could replace that with
`$"dtor_{ctor.CompileName}{arg.HasName ? "" : FormalName(arg, index)}"`

right?

fabiomadge: Otherwise, one uses `arg` the other `ctor`.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -642,14 +642,16 @@ bool ContainsTyVar(TypeParameter tp, Type ty)
...
-      ConcreteSyntaxTree interfaceTree, string DtT_TypeArgs) {
+        ConcreteSyntaxTree interfaceTree, string DtT_TypeArgs) {
       foreach (var ctor in dt.Ctors) {
-        foreach (var dtor in ctor.Destructors) {
+        foreach (var (dtor, index) in ctor.Destructors.Select((v, i) => (v, i))) {
           if (dtor.EnclosingCtors[0] == ctor) {
             var arg = dtor.CorrespondingFormals[0];
-            if (!arg.IsGhost && arg.HasName) {
+            if (!arg.IsGhost) {
+              var name = arg.HasName ? arg.CompileName : FormalName(arg, index);
```
keyboardDrummer, line 651: How come you can't write `var name = FormalName(arg, index);` here? The `HasName` check is already in FormalName. Is the extra `IdProtect` call a problem here?

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -2614,8 +2616,8 @@ protected class ClassWriter : IClassWriter {
     }
 
     protected override void EmitDestructor(string source, Formal dtor, int formalNonGhostIndex, DatatypeCtor ctor, List<Type> typeArgs, Type bvType, ConcreteSyntaxTree wr) {
-      var dtorName = FormalName(dtor, formalNonGhostIndex);
-      wr.Write("(({0}){1}{2}).{3}", DtCtorName(ctor, typeArgs, wr), source, ctor.EnclosingDatatype is CoDatatypeDecl ? "._Get()" : "", dtorName);
+      var dtorName = dtor.HasName ? dtor.CompileName : ctor.CompileName + FormalName(dtor, formalNonGhostIndex);
```
keyboardDrummer, line 2619: The expression `dtor.HasName ? dtor.CompileName : ctor.CompileName + FormalName(dtor, formalNonGhostIndex)` seems the same as what's on line 652, and it also has to be the same for correctness. Could you extract that into a method and use it in both locations?

Could you add a comment explaining why destructors for named formals don't need the `ctor.CompileName` qualification? Maybe refer to the formal name collision detection in https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Resolver.cs#L2164

Source/Dafny/Compilers/Compiler-Csharp.cs
keyboardDrummer: I see, thanks.

Source/Dafny/Compilers/Compiler-Csharp.cs
fabiomadge: I believe that's exactly right.

fabiomadge: We also don't want an index when there is a name.

fabiomadge: You can't have something like `dtor_@this`.

RELEASE_NOTES.md
```diff
@@ -1,3 +1,10 @@
+# Upcoming
+
+- feat: Plugin support in the compiler pipeline (https://github.com/dafny-lang/dafny/pull/1739)
```
MikaelMayer, line 3:
```suggestion
- feat: Plugin support in the resolution pipeline (https://github.com/dafny-lang/dafny/pull/1739)
```

MikaelMayer (commented):
Thanks for incorporating my changes  in the RELEASE_NOTES.md

Source/Dafny/Compilers/Compiler-Csharp.cs
keyboardDrummer: I think I would prefer defining `dtorName` without using `name` anyways, like so:

```
dtorName = arg.HasName ? arg.CompileName : ctor.CompileName + FormalName(arg, index)
```

So then sharing the expression with `EmitDestructor` becomes more attractive.

keyboardDrummer (approved):

RELEASE_NOTES.md
```diff
@@ -1,3 +1,10 @@
+# Upcoming
+
+- feat: Plugin support in the resolution pipeline (https://github.com/dafny-lang/dafny/pull/1739)
+- fix: NullPointerException in the AST (https://github.com/dafny-lang/dafny/pull/1805)
+- fix: Change datatype deconstruction in match statements (https://github.com/dafny-lang/dafny/issues/1815)
```
fabiomadge, line 5:
```suggestion
- fix: Change datatype deconstruction in match statements for C# (https://github.com/dafny-lang/dafny/issues/1815)
```

keyboardDrummer (approved):

Issue #1631, opened 2021-12-05
On failure to prove tautologies


tjhance:
Occasionally, Dafny will fail to prove a post-condition which is already so obviously true that there is no possible way to insert intermediate steps. Here is an example I ran into recently:

```dafny
function {:opaque} Choose<A>(s: multiset<A>) : (result: A)
  requires 0 < |s|
  ensures result in s
{
  var a :| a in s;
  a
}

function {:opaque} Apply<A, B>(fn: A ~> B, s: multiset<A>) : (result: multiset<B>)
  requires forall x | x in s :: fn.requires(x)
  ensures |result| == |s|
  reads set x, o | x in s && o in fn.reads(x) :: o
{
  if |s| == 0 then
    multiset{}
  else
    var x := Choose(s);
    multiset{fn(x)} + Apply(fn, s - multiset{x})
}

lemma ApplyAdditive<A,B>(fn: A ~> B, s1: multiset<A>, s2: multiset<A>)
  requires forall x | x in s1 :: fn.requires(x)
  requires forall x | x in s2 :: fn.requires(x)
  ensures Apply(fn, s1+s2) == Apply(fn, s1) + Apply(fn, s2)
  // ...

lemma MultisetSimplificationForall<P,Q>()
  // This is the postcondition that might not hold.
  ensures forall fn: P -> Q, s, t :: Apply(fn, s + t) == Apply(fn, s) + Apply(fn, t)
{
  forall fn: P -> Q, s, t
  ensures Apply(fn, s + t) == Apply(fn, s) + Apply(fn, t)
  {
    ApplyAdditive(fn, s, t);
  }

  // Assertion violation
  assert forall fn: P -> Q, s, t :: Apply(fn, s + t) == Apply(fn, s) + Apply(fn, t);

  // Assertion violation
  assert forall fn: P -> Q, s, t :: Apply(fn, s + t) == Apply(fn, s) + Apply(fn, t);
}
```

As you can see, the code proves the post-condition within the lemma body, but Dafny fails to recognize it.
I added the assertions for extra emphasis: Dafny can't prove the last one, even though it's syntactically identical to the preceding one.

In this particular instance, it's probably a bug relating to the triggers involving function values, multisets, generics, or the intersection thereof. I also know that I can work around it by munging around with triggers (this often involves creating some kind of dummy predicate to act as a trigger).

[...]

PR #1821, opened 2022-02-14
chore: Fixing opaque documentation

### LINKED ISSUE #1295: Documentation of `{:opaque}` doesn't match the implementation
[The documentation of `{:opaque}`](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-opaque) describes the attribute `opaque_full` even though that doesn't occur in the Dafny codebase. There are also two unused dictionaries in the codebase:

```
    protected Dictionary<Function, Function> fullVersion; // Given an opaque function, retrieve the full
    protected Dictionary<Function, Function> original;    // Given a full version of an opaque function, find the original opaque version
```

Was the `opaque` feature finished? Is using it currently any different than annotating a method with `{:fuel 0,0}` ?

MikaelMayer:
This PR is a draft to fix the documentation about the  `{:opaque}` attribute.
Specifically:
* I made it clear what developers need to know and what is internal. Developers only need to know about `{:opaque}` and `reveal f()` and what it does.
* I reverse-engineered the previously undocumented internal implementation of `{:opaque}` and `reveal_`, how it interacts with the fuel, and exposed its details.
* I removed the section about `{:opaque_full}` because it's no longer needed

This PR fix #1295


docs/DafnyRef/Attributes.md
```diff
@@ -300,38 +300,32 @@ values that satisfy the constraint.
 
 ### 22.1.13. opaque {#sec-opaque}
 Ordinarily the body of a function is transparent to its users but
-sometimes it is useful to hide it. If a function `f` is given the
-`{:opaque}` attribute then Dafny hides the body of the function,
+sometimes it is useful to hide it. If a function `foo` or `bar` is given the
+`{:opaque}` attribute, then Dafny hides the body of the function,
 so that it can only be seen within its recursive clique (if any),
-or if the programmer specifically asks to see it via the `reveal_f()` lemma.
+or if the programmer specifically asks to see it via the statement `reveal foo(), bar();` or the expression `reveal foo(), bar(); ...`.
```
RustanLeino, line 306: Thanks for "un-documenting" the old `reveal_f()` lemmas. The official support for this feature is with the `reveal` statement, as you have described.

I think it's okay to say just "statement ..." (and omit "or the expression ..."). Technically, there is no special `reveal` _expression_. Instead, the `reveal` _statement_ is one of about 5 statements that are allowed to be part of a `StmtExpr`.

RustanLeino (changes requested):

docs/Compilation/Boogie.md
```diff
@@ -0,0 +1,23 @@
+# Dafny compilation to Boogie
+
+## `{:opaque} attribute`
+
+Internally, and for verification in Boogie only, all functions are given an extra first parameter, which is the _fuel_ (represented in Boogie by unary numbers, e.g. `$LS($LS($LZ)`) is a fuel of two).
+To unroll a function, axioms ensures that a recursive call is provided with the previous fuel minus 1, so that we don't unroll undefinitely.
```
keyboardDrummer, line 6: ensures => ensure
Consider replaying `previous` with `current`

docs/Compilation/Boogie.md
```diff
@@ -0,0 +1,23 @@
+# Dafny compilation to Boogie
+
+## `{:opaque} attribute`
+
+Internally, and for verification in Boogie only, all functions are given an extra first parameter, which is the _fuel_ (represented in Boogie by unary numbers, e.g. `$LS($LS($LZ)`) is a fuel of two).
```
keyboardDrummer, line 5: I wouldn't use internally because of it was truly internal you wouldn't need to put it in the documentation. I'd say "During verification, all functions are given..... "

Also I'd leave out references to Boogie since that's an implementation detail

Also, do users need to know about the peano/unary numbers? Seems like an implementation detail.

docs/Compilation/Boogie.md
```diff
@@ -0,0 +1,23 @@
...
+To unroll a function, axioms ensures that a recursive call is provided with the previous fuel minus 1, so that we don't unroll undefinitely.
+
+**Normal behavior:** When verifying each method or function, Dafny will look at every `assert`. If it finds one, every function inside it will be given a fuel of at least 2 (`$LS($LS($LZ))`), meaning the verifier will be able to unroll the function's body twice for this assertion. When this assertion is proven, Dafny provides an `assume` about the result, but in this case the functions are provided a fuel of only 1 (`$LS($LZ)`), meaning it unrolls the body only once.
```
keyboardDrummer, line 8: Consider changing `each` to `a`.

keyboardDrummer, line 8: `every function inside it will be given ...` => `every function call inside it will use a fuel argument of at least 2`

keyboardDrummer, line 8: As a novice, the assert/assume getting 2/1 fuel behavior doesn't make too much sense to me and so this paragraph leaves me with more questions than answers.

docs/Compilation/Boogie.md
```diff
@@ -0,0 +1,23 @@
+# Dafny compilation to Boogie
+
+## `{:opaque} attribute`
```
keyboardDrummer, line 3: I would first explain the effect of the opaque attribute without going into fuel, that the body becomes completely invisible for opaque functions until reveal is called.

I suggest only explaining fuel if you want to explain the `{:fuel ...}` attribute.

docs/DafnyRef/Attributes.md
```diff
@@ -147,7 +147,7 @@ It means that the post-condition may be assumed to be true
 without proof. In that case also the body of the function or
 method may be omitted.
 
-The `{:axiom}` attribute is also used for generated `reveal_*`
+The `{:axiom}` attribute is internally equivalent to `{:verify false}` and is also used for the internal `reveal_*`
```
keyboardDrummer, line 150: What's the difference between equivalent and internally equivalent?

I think any use of the word `internal` in documentation to refer to something not so important to the user, is a smell. I'd say either the user needs to know and then it's not internal, or they don't and then it shouldn't be mentioned in the documentation.

docs/Compilation/Boogie.md
MikaelMayer: Again, this document explains how we actually compile to Boogie, so I'm not sure why remove references to Boogie would be helpful.
If I had had this documentation before, I would have been able to solve https://github.com/dafny-lang/dafny/issues/1631 immediately instead of waiting for months.
I do want to describe implementation details, in case someone wants to understand the generated Boogie code.

docs/Compilation/Boogie.md
MikaelMayer: I leaves more questions to me too. I am awaiting some feedback from @RustanLeino to justify this.

docs/DafnyRef/Attributes.md
MikaelMayer: There is a user-facing difference between something being an `{:axiom}`, meaning it's assumed to be true and will be axiom forever (and usually functions or methods annotated with `{:axiom}` do not have body) , vs. something annotated with `{:verify false}`, which looks like a temporary modifier to instruct Dafny to not proofread an existing body. Typically, in an IDE, we would emit warnings for `{:verify false}` but not for `{:axiom}`.
However, at the end, they have exactly the same semantics, and I was actually surprised to see they were the first time I saw that, which I think is brilliant. That's what I mean by "internally equivalent".
If this helps, you can think of an analogy with comments: Double slash is the same as triple slash for the code (it's code documentation), but triple slash means documentation that is typically API documentation, whereas double slash means it's for developers maintaining the code. In a concrete AST, I wouldn't be surprised if the two comment nodes are represented with the same structure, but with a flag.

So, in short, I believe that the first "internally equivalent" is appropriate ("Oh by the way, user, don't expect any difference from these two attributes from Dafny point of view, but of course use them as you think appropriate").

I'll remove the second one however because since we now hide the implementation of reveal_* as a lemma we shouldn't mention it in the documentation.

docs/Compilation/Boogie.md
keyboardDrummer: > Since this is a file that explains how we compile to Boogie

Shouldn't documentation on that be inside the code?

docs/Compilation/Boogie.md
keyboardDrummer: > I would have been able to solve #1631 immediately instead of waiting for months.

I think documentation needed to resolve that ticket should be part of the main Dafny documentation, not hidden in a document about Dafny internals.

RustanLeino (changes requested):

docs/DafnyRef/Attributes.md
```diff
@@ -147,8 +147,7 @@ It means that the post-condition may be assumed to be true
 without proof. In that case also the body of the function or
 method may be omitted.
 
-The `{:axiom}` attribute is internally equivalent to `{:verify false}` and is also used for the internal `reveal_*`
-lemmas as shown in Section [#sec-opaque].
+The `{:axiom}` attribute is internally equivalent to `{:verify false}`.
```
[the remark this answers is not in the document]
cpitclaudel: By the way, why is the compiler complaining about body-less functions?  Given that functions are ghost, I wouldn't expect the compiler to look at them at all.

docs/Compilation/Boogie.md
MikaelMayer: I'm not sure if I need to do anything after your comment. This file is named "Boogie.md", so I'm not sure I understand your previous comments.

docs/DafnyRef/Attributes.md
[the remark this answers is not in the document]
MikaelMayer: I think this is to enable quick prototyping while not being bothered by "missing body" errors. This check has to be done somewhere.
But if we run the compiler's diagnostics as part of the resolving diagnostics, we might get a "missing body" error which would prevent us from triggering verification. I'm just thinking.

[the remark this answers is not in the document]
RustanLeino: Declaring a function or lemma without giving it a body is like adding a body with an `assume false` -- it lets users of the function or lemma assume the specification, but there's nothing to ensure that the specification holds. If a user accidentally left off the body of the function or lemma, then it could be that the program is not correct even if everything else in the program verifies. So, _something_ in the Dafny pipeline should give an error about this.

One choice would be to let the verifier complain about it. A different choice would be to make the compiler complain about it. A third choice would be to have another pass in the Dafny pipeline, perhaps between the verifier and compiler, that complains about it. It's nice _not_ to give the error in the verifier, because then users don't have to get verification errors until they start writing the body of the function or lemma. And it _is_ convenient to let the compiler give the error, because the compiler already has to check that all functions and methods that it's supposed to turn into code have bodies. So, the choice made in the Dafny pipeline is for the compiler for look for and complain about body-less entities, including body-less _ghost_ entities.

docs/Compilation/Boogie.md
RustanLeino: Code comments are nice, but I think it's perfectly fine to complement them with out-of-band documentation. Indeed, we do this already for some compiler issues (see `docs/Compilation/*.md`).

RustanLeino (approved):

PR #1822, opened 2022-02-14
docs: Document :termination attribute

RustanLeino:
This PR extends the `/attrHelp` output to document the `{:termination}` attribute.


MikaelMayer:
Can  `{:termination false}` be used anywhere to say "don't verify termination" or is it only for trait methods? How is it different from `decreases *`

RustanLeino:
Good points, @MikaelMayer . I updated the PR with answers to your questions.

MikaelMayer (approved):
Great job documenting the difference with `decreases *`

PR #1823, opened 2022-02-14
Clarify what types can extend a trait

### LINKED ISSUE #1730: Clarify in the reference manual that datatypes cannot implement traits
This is taken as implicit but is a common misunderstanding by new Dafny users. https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#14-trait-types talks about "reference types" but we should explicitly clarify that excludes datatypes.

Related to the actual feature request to support it after all (#1226)

atomb:
Fixes #1730


docs/DafnyRef/Types.md
```diff
@@ -2315,10 +2315,11 @@ TraitDecl =
 ````
 
 A _trait_ is an abstract superclass, similar to an "interface" or
-"mixin".[^fn-traits]
+"mixin".[^fn-traits] A trait can be _extended_ only by another trait or
+by a class (and in the latter case we say that the class _implements_
+the trait). More specifically, algebraic datatypes cannot extend traits.
```
MikaelMayer, line 2320:
```suggestion
the trait). More specifically, algebraic datatypes cannot extend traits.[^fn-traits]
```

docs/DafnyRef/Types.md
```diff
@@ -2315,10 +2315,11 @@ TraitDecl =
 ````
 
 A _trait_ is an abstract superclass, similar to an "interface" or
-"mixin".[^fn-traits]
+"mixin".[^fn-traits] A trait can be _extended_ only by another trait or
```
MikaelMayer, line 2318:
```suggestion
"mixin". A trait can be _extended_ only by another trait or
```

MikaelMayer (changes requested):
Just move the footnote next to the biggest pain point

MikaelMayer (approved):
Looks good, thanks a lot for updating the doc !

PR #1824, opened 2022-02-14
Fix: Output in JavaScript not consistent

MikaelMayer:
By switching to the asynchronous API of processing output data of process, this PR makes it possible to run the option `/compiler:3` and have an output when the compiler is `js`.

Fix https://github.com/dafny-lang/ide-vscode/issues/131

I don't know how to reproduce this in a test.  Reading `nodeProcess.StandardOutput.ReadtoEnd()` works as well, and without it or without this PR nothing prints to screen.

Note for reviewing:
* The [event specification](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.outputdatareceived?view=net-6.0) says that the callbacks are called after each newline, and do not include the newline (I tested)
* However, if the process exited, the callbacks are called with the remaining data.
* Therefore, I test if the process exited, and if so, I add a newline after the data, else I just write the data. This seems to works for all of our tests and for my use case.


Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2386,12 +2386,32 @@ protected class ClassWriter : IClassWriter {
...
-        RedirectStandardOutput = false,
-        RedirectStandardError = false,
+        RedirectStandardOutput = true,
+        RedirectStandardError = true,
       };
 
       try {
         using var nodeProcess = Process.Start(psi);
+        nodeProcess.BeginErrorReadLine();
+        nodeProcess.BeginOutputReadLine();
```
cpitclaudel, line 2396: Don't you want to put these after the ProcessOutputData setup?

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2386,12 +2386,32 @@ protected class ClassWriter : IClassWriter {
...
+          }
+        }
+
+        void ProcessOutputData(object sender, DataReceivedEventArgs e) {
+          if (!nodeProcess.HasExited) {
+            Console.Out.WriteLine(e.Data);
+          } else {
+            Console.Out.Write(e.Data);
+          }
+        }
```
cpitclaudel, line 2412: Remove the duplication by extracting a method

cpitclaudel:
Should this be equivalent to just not redirecting output at all?  I'm confused why it works.

That being said, nodejs does something really weird with IO on Windows. https://github.com/nodejs/node/issues/4697

Source/Dafny/Compilers/Compiler-js.cs
camrein: Just to add to @cpitclaudel's suggestion: I wouldn't declare it as an inner method since you don't require any closures.

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2386,12 +2386,32 @@ protected class ClassWriter : IClassWriter {
...
+        nodeProcess.ErrorDataReceived += ProcessErrorData;
```
camrein, line 2413: Move the registration of the event-handlers before the asynchronous setup `BeginErrorReadLine` to not miss any process output due to the race condition.

Source/Dafny/Compilers/Compiler-js.cs
MikaelMayer: I _did_ close on on `nodeProcess`. However, after reading the documentation, it looks like the closure's first argument is the process itself, it just needed to be casted. I removed the duplication by creating a private method,

MikaelMayer:
> That being said, nodejs does something really weird with IO on Windows. [nodejs/node#4697](https://github.com/nodejs/node/issues/4697)

I think you nailed it ! Yes, it should be equivalent to do no redirection at all, but it is not on Windows for node.js and I don't know why.

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2378,6 +2378,16 @@ protected class ClassWriter : IClassWriter {
       return SendToNewNodeProcess(dafnyProgramName, targetProgramText, callToMain, targetFilename, otherFileNames, outputWriter);
     }
 
+    DataReceivedEventHandler ProcessData(TextWriter writer) {
+      return (sendingProcess, e) => {
+        if (!((Process)sendingProcess).HasExited) {
```
camrein, line 2383: I don't have a strong opinion about this, but I tend to avoid casts where possible. Since you're already providing `writer` as a closure, you may as well do so for the `Process` instance. But I leave it up to you.

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2378,6 +2378,16 @@ protected class ClassWriter : IClassWriter {
       return SendToNewNodeProcess(dafnyProgramName, targetProgramText, callToMain, targetFilename, otherFileNames, outputWriter);
     }
 
+    DataReceivedEventHandler ProcessData(TextWriter writer) {
```
camrein, line 2381: I think you could provide a more descriptive name here. On the one hand it's actually a factory for event handlers, on the other hand the created event handler prints the data to the specified `writer`.

camrein (dismissed):
Looks good, just two notes where it's up to your preference.

Source/Dafny/Compilers/Compiler-js.cs
MikaelMayer: Thanks for your comment.
I will keep the cast as since the closure is already provided with the process, I don't think that adding it to the closure builder would be beneficial, especially just for having the type.

MikaelMayer:
Whatever I do, I can't properly transfer STDOUT from the child process to the parent. There are unreproducible race conditions between the process having exited and the need to guess that a line ending triggered `e.Data` was sent. The last newline might or might not be taken into account.

Hence, I'm completely removing this solution and replace it with a simple one where two while loops take care of transferring the process's output to current process's output, character by character.

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2401,6 +2402,14 @@ protected class ClassWriter : IClassWriter {
...
         nodeProcess.StandardInput.Flush();
         nodeProcess.StandardInput.Close();
+        int current;
+        // Fixes a problem of Node on Windows, where Node does not prints to the parent console its standard outputs.
+        while ((current = nodeProcess.StandardOutput.Read()) != -1) {
+          Console.Out.Write((char)current);
+        }
+        while ((current = nodeProcess.StandardError.Read()) != -1) {
+          Console.Error.Write((char)current);
+        }
```
cpitclaudel, line 2412: I don't think this is safe, the process may fill the stderr buffer and block on it while we wait for stdout.

cpitclaudel: From https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=net-6.0:

> Synchronous read operations introduce a dependency between the caller reading from the [StandardOutput](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=net-6.0) stream and the child process writing to that stream. These dependencies can result in deadlock conditions. When the caller reads from the redirected stream of a child process, it is dependent on the child. The caller waits on the read operation until the child writes to the stream or closes the stream. When the child process writes enough data to fill its redirected stream, it is dependent on the parent. The child process waits on the next write operation until the parent reads from the full stream or closes the stream. The deadlock condition results when the caller and child process wait on each other to complete an operation, and neither can proceed. You can avoid deadlocks by evaluating dependencies between the caller and child process.

MikaelMayer: I pushed something so that the error writing is done in a different thread. I think it solves the deadlock problem. Let me know if you agree with this.

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2401,6 +2404,17 @@ protected class ClassWriter : IClassWriter {
...
+        Task.Run(() => {
+          int current;
+          while ((current = nodeProcess.StandardError.Read()) != -1) {
+            Console.Error.Write((char)current);
+          }
+        });
+        int current;
+        while ((current = nodeProcess.StandardOutput.Read()) != -1) {
+          Console.Out.Write((char)current);
+        }
```
cpitclaudel, line 2417: Looks good now, though reading character-by-character is a bit scary.  I'm not sure it's that important since it's just a debugging facility, but if it is I would spawn two instances of a class that does something roughly like 

```
while (var count = await ReadAsync(buf, 0, bufsize)) {
  outputStream.Write(buf)
}
```

IIRC `ReadAsync` is only blocks until some output is received.

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2401,6 +2404,17 @@ protected class ClassWriter : IClassWriter {
         }
         nodeProcess.StandardInput.Flush();
         nodeProcess.StandardInput.Close();
+        // Fixes a problem of Node on Windows, where Node does not prints to the parent console its standard outputs.
+        Task.Run(() => {
```
camrein, line 2408: Try to avoid fire-and-forget tasks. Errors happening inside the task will be lost.
Either add an exception handler that prints to the desired output, add a continuation responsible to handle errors, or best: [Wait](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.wait?view=net-6.0) for the task's completion so potential errors are reported back to the main thread. Mind that exceptions that occured inside a task are wrapped by an [AggregateException](https://docs.microsoft.com/en-us/dotnet/api/system.aggregateexception?view=net-6.0) that you may want to unwrap first.

Source/Dafny/Compilers/Compiler-js.cs
camrein: I second that it might be beneficial to not read char by char. However, since the Dafny Compiler consists purely of synchronous methods, I suggest using the non-async version instead.
Either use `Read` with a buffer or `ReadLine`. Although, the latter might be not the best choice since we cannot clearly say that the output is always separated by lines.

camrein (changes requested):
I preferred the previous implementation. However, since it appears that there's an unsolvable problematic when catching the final output of a process it's OK to change it according to your suggestion.

Source/Dafny/Compilers/Compiler-js.cs
MikaelMayer: About character by character vs. buffering.
Buffering is a best practice when the goal is to recover the entire output into some kind of file, or for real-time application processing. Not when it's user-facing.
Here, we are talking about `dafny.exe -compile:3 -compileTarget:js`, which is a tool mostly used while developing a JavaScript application in Dafny. Its output is mean to be visible by the user, for example in the IDE, else they would just use `node` on a generated file.
So now, if the output has I/O operations, buffering the output would result in not displaying in real-time the output, which would result in a disastrous user experience.

MikaelMayer: I refactored the code so that it's now inside a function

Source/Dafny/Compilers/Compiler-js.cs
MikaelMayer: I wait for the result of this task now just after waiting for the process to exit. I will not handle errors at this point because I don't see a use case for that yet.

camrein (approved):
Looks good. Aren't there any test cases for similar cases?

MikaelMayer:
> Aren't there any test cases for similar cases?

Yes there are but none of them are able to catch this error and that's obscure to my why.
There is a test for JavaScript, which works because it calls "node" on the shell.
https://github.com/dafny-lang/dafny/blob/master/.github/workflows/release-downloads.yml#L109
However, looking at recent output of the dafny command that is also supposed to run the file, it also correctly outputs the result
https://github.com/dafny-lang/dafny/runs/4797180341?check_suite_focus=true
There is nothing we can test at this stage if this is not a consistent error.

PR #1828, opened 2022-02-16
feat(lsp): Show verification progress by boogie program

camrein:
This a small extension to the status reporting during verification. It intends to improve the UX by letting users know at which boogie program is currently being verified out of the total number of boogie programs (that is, the number of Dafny modules) to verify.

I drafted this PR for quick review if the intended UX improvement is actually valuable hence why I didn't fix the failing tests yet.

image

PS: If someone has a larger sample project and screenshot to see if it's actually displaying more than 1/1, that be great.
PPS: We could further improve upon this if we'd know the number of methods to verify in advance.

### DIFF
--- a/Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
+++ b/Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
@@ -4,6 +4,7 @@ using Microsoft.Extensions.Options;
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
@@ -67,8 +68,9 @@ namespace Microsoft.Dafny.LanguageServer.Language {
       mutex.Wait(cancellationToken);
       try {
         // The printer is responsible for two things: It logs boogie errors and captures the counter example model.
+        int totalPrograms = Translator.VerifiableModules(program).Count();
         var errorReporter = (DiagnosticErrorReporter)program.reporter;
-        var printer = new ModelCapturingOutputPrinter(logger, errorReporter, progressReporter);
+        var printer = new ModelCapturingOutputPrinter(logger, errorReporter, progressReporter, totalPrograms);
         ExecutionEngine.printer = printer;
         // Do not set these settings within the object's construction. It will break some tests within
         // VerificationNotificationTest and DiagnosticsTest that rely on updating these settings.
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Won't it be confusing that the number doesn't always change when the string before it does?  For example, what will it show for this program?

```dafny
module A0 {
  method f0() returns (r: nat) { return 0; }
  method g0() returns (r: nat) { return 0; }
  method h0() returns (r: nat) { return 0; }
}

module A1 {
  method f1() returns (r: nat) { return 1; }
  method g1() returns (r: nat) { return 1; }
  method h1() returns (r: nat) { return 1; }
}

module A2 {
  method f2() returns (r: nat) { return 2; }
  method g2() returns (r: nat) { return 2; }
  method h2() returns (r: nat) { return 2; }
}
```

camrein:
That was one of my concerns as well. That's why I was wondering if we could actually retrieve the number of boogie functions being verified. The example code you provided will be presented like this:
image

If there's any benefit for this information, we'd definitely have to make it clear that these numbers represent the modules being verifeid.

MikaelMayer:
> If there's any benefit for this information, we'd definitely have to make it clear that these numbers represent the modules being verifeid.

To my experience, the Dafny extension is primarily used to display per-file feedback, and I don't expect usually more than 1 module per file (except when a user is trying to prototype a large-scale system). Thus, having a ratio per module will be confusing, especially because the number does not change when the name changes.

I would love if you could actually retrieve a number that will increase by 1 not per boogie program, but per method name change. That would be immediately useful.
I think you could do it the following way:

* Replace `CurrentProgram` by `MethodsVerified`
* Increment `MethodsVerified` not in the loop per boogie program, but where `ReportProgress` is called
* Replace `totalProgram` by `totalMethods`
* Compute `totalMethods` by going through the Dafny AST and manually computing the methods that Boogie is going to verify

cpitclaudel:
> I don't expect usually more than 1 module per file

!! I find that very surprising: given that the IDE currently only gives per-file feedback, it's hard to edit things with tight dependencies that are spread across files.

camrein:
Thanks for your suggestions @MikaelMayer. I was hoping there'd be a quicker way to retrieve the necessary information.

> !! I find that very surprising: given that the IDE currently only gives per-file feedback, it's hard to edit things with tight dependencies that are spread across files.

Making the language server verify the whole tree isn't a big deal. If I remember correctly, you'd have to set `DafnyOptions.VerifyAllModules` to `true`. You could do this here:
https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs#L47-L48

I have the feeling that this will have a huge impact on responsiveness. However, it depends on how well incomplete verifications can be canceled. Therefore, it might be worth giving it a try with a larger code base. We could also think about providing `/verifyAllModules` as a new command within VSCode.

keyboardDrummer:
I'm working on changes to expose more information about individual proofs from Boogie to Dafny, which'll allow us to show the total number of proofs being verified. Consider pausing this PR until that information is available.

cpitclaudel:
I think this feature is now available in the latest Dafny LSP :)

camrein:
Nice! Closing this PR

PR #1830, opened 2022-02-16
fix(lsp): Do not mark the bodies of lemmas as ghost.

camrein:
Resolves https://github.com/dafny-lang/ide-vscode/issues/126 by explicitly excluding lemmas when traversing the AST to collect ghost statements.

Taking this a step further: Since @RustanLeino couldn't always reproduce the issue, I suspect there's more than just lemmas being highlighted.

I have two things in mind that could lead to the situation that the bug is not easily reproducible:
[GhostStateDiagnosticCollector](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Language/GhostStateDiagnosticCollector.cs) runs even when the resolution failed. See [TextDocumentLoader](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs#L128-L129). The reported ghost statement in the AST might be incorrect due to this. The fix would be quite easy here: Only report ghost states for documents that resolved successfully. However, if that was the case, the error should surface not only within lemmas.

The other thing that might be worth a thought is the situation when the document moves into *migrated* state. That happens when the currently processed change is being canceled by a newer change. The language server then relocates the symbol table as well as the diagnostics according to this change (see [DocumentDatabase](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs#L139-L144)). We'd either need to relocate the ghost diagnostics or discard them. Otherwise, I believe they might show invalid locations.

### DIFF
--- a/Source/DafnyLanguageServer.Test/Synchronization/GhostDiagnosticsTest.cs
+++ b/Source/DafnyLanguageServer.Test/Synchronization/GhostDiagnosticsTest.cs
@@ -69,6 +69,25 @@ class C {
     }
 
     [TestMethod]
+    public async Task OpeningFlawlessDocumentOnlyContainingLemmasWithGhostMarkStatementsDoesNotMarkAnything() {
+      var source = @"
+class C {
+  lemma Test(x: int) {
+    var y := 0;
+    y := x;
+  }
+}".TrimStart();
+      await SetUp(new Dictionary<string, string>() {
+        { $"{GhostOptions.Section}:{nameof(GhostOptions.MarkStatements)}", "true" }
+      });
+      var documentItem = CreateTestDocument(source);
+      await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+      var report = await diagnosticReceiver.AwaitNextNotificationAsync(CancellationToken);
+      var diagnostics = report.Diagnostics.ToArray();
+      Assert.AreEqual(0, diagnostics.Length);
+    }
+
+    [TestMethod]
     public async Task OpeningFlawlessDocumentWithGhostMarkStatementsMarksGhostVariableDeclarations() {
       var source = @"
 class C {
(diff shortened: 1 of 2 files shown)


cpitclaudel:
But aren't lemmas in fact ghost?  I think it's not a bad thing to gray them out.
https://github.com/dafny-lang/ide-vscode/issues/126 complained about incorrect highlighting, but didn't advocate for not highlighting lemmas, did it?

camrein:
The motivation to not highlight the bodies comes from this discussion: https://github.com/dafny-lang/ide-vscode/issues/53#issuecomment-947816592
We opted to not highlight ghost statements in contexts where the statements are always ghost. That's why I've created this PR to no longer highlight lemmas.

Of course, it doesn't resolve highlighting statements falsely. To resolve this issue, I'll need more information. I'm pretty positive that the issue relates to one of the reasons I've outlined in the PR, or even both. I hope someone who's a more frequent Dafny user may confirm my assumptions before I (or someone else) start implementing potential fixes.

cpitclaudel (approved):

MikaelMayer:
With the new italicized and slightly transparent highlighting of ghost statements, I think it makes sense that we don't mark the body of lemmas as ghost in our current mode where we already don't display something as ghost if it's in a ghost context anyway. While writing a lot of lemmas in VSCode recently, I would have really appreciated to have natural highlighting rather than italicized highlighting.
However, down the road, I really think it should be an option. Newcomers in Dafny should see everything that is ghost highlighted in ghost even if it's not ambiguous, that way, they can quickly learn the difference. Intermediate and expert users should be able to configure VSCode to have the current disambiguation-only highlighting.

MikaelMayer:
I'm willing to help finish to merge this PR @camrein let me know how I can help !

PR #1833, opened 2022-02-16
Move banner back to the top of the page

robin-aws:


cpitclaudel (approved):

MikaelMayer (changes requested):
Moving the banner it to the bottom was not an error. A video is worth 1000 images, and the GIF aimed at giving a better view of what the Dafny experience looks like.

More than 50% of the "banner" contains a screenshot of Dafny in emacs mode, with mathematical operators that we know now frighten some developers ("how do I input this!?").
My take on that is that I would still like to keep the emacs screenshot, but it would be better at the bottom (not in the very first few moments someone new views when searching for Dafny), and with a proper documentation (This is Dafny in emacs mode, with mathematical operators).

I would also accept to just put the Dafny logo at the top, but not the screenshot, as it would start to give away some information that is not relevant on the very first few moments people want to spend with Dafny.

PR #1839, opened 2022-02-18
feat: continue statements

### LINKED ISSUE #1738: Feature request: continue statement
Dafny supports `break` statement and `labeling` but it doesn't have a support for `continue` statement.
Currently we uses labeling to mimic continue.
```
for i := 0 to |values1|
{
  label BodyOfLoop1:
  {
    if (p) {
      break BodyOfLoop1;
    }
    blah ...
  }  // breaking out of BodyOfLoop1 will get you here
}
```
It would be nice for Dafny to also support `continue`. 
```
for i := 0 to |values1|
{
  if (p) {
    continue;
  }
  blah ...
}
```

RustanLeino:
This PR adds `continue` statements, in the two forms analogous to the two forms of `break` statements:

```
ContinueStmt ::=
  | "continue" LabelName ";"
  | { "break" } "continue" ";"
```

The statement `continue L;` causes control flow to jump to the very end of the current iteration of the enclosing loop labeled `L`. The statement `break ... break continue;` with _n_ occurrences of `break` (where _n_ is 0 or more) breaks out of _n_ loops and then ends the current iteration.

For more details, see `docs/DafnyRef/Statements.md`.

Fixes #1738


Source/Dafny/Compilers/Compiler.cs
```diff
@@ -2872,10 +2879,10 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
           ConcreteSyntaxTree guardWriter;
           var wBody = CreateWhileLoop(out guardWriter, wr);
           guardWriter.Write("true");
-          EmitBreak(s.Labels?.Data.AssignUniqueId(idGenerator), wBody);
+          EmitBreak(null, wBody);
```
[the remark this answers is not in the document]
RustanLeino: The Dafny statement `while * { Body }` says to iterate `Body` an arbitrary (and possibly infinite) number of times. One correct way to compile that statement is to decide always to iterate exactly 0 times. That is what the compiler does.

This means the compiler can ignore `Body` and in fact just emit nothing. But (for reasons I don't entirely remember), the compiler does introduce a loop structure in the target code. (The reason may have to do with Dafny `/coverage` functionality. I don't remember.) Such a loop could look like `while (false) { }`. Alas, this causes `javac` to emit an "unreachable code" error. So, instead the compiler generates `while (true) { break; }`.

With all that said, the answer to your question is that, _if_ the Dafny loop had a label or was using a `break` or `continue` inside `Body`, then the target code would be generated as

```
L: while (true) {
  break L;
}
```

The code change I made on line 2875 simplifies the target code to just

```
L: while (true) {
  break;
}
```

docs/DafnyRef/Statements.md
```diff
@@ -20,46 +21,58 @@ They are described in subsequent sections.
...
-must have been encountered during the control flow in route to the `old`
-expression. That is, again, the label must dominate the use of the label.
-
-## 19.2. Break Statement
+referenced in a `break` or `continue` statement within the labeled statement
+(see [Section 19.2](#sec-break-continue)). That is, the break or continue that
+mentions the label must be _enclosed_ in the labeled statement.
+The label may also be used in an `old` expression ([Section 20.24](#sec-old-expression)). In this case, the label
+must have been encountered during the control flow en route to the `old`
+expression, which is to say the label must _dominate_ the use of the label.
```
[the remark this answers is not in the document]
RustanLeino: I improved the wording along your suggestion, and also tried to be more clear about "dominates".

The word "dominates" is used in the Dragon Book ("Compilers" by Aho et al.). A program point (or basic block) `A` _dominates_ a program point (or basic block) `B` when the only way to get to `B` is via `A`.

docs/DafnyRef/Statements.md
```diff
@@ -20,46 +21,58 @@ They are described in subsequent sections.
...
+
+## 19.2. Break and Continue Statements {#sec-break-continue}
 ````grammar
-BreakStmt = "break" ( LabelName | { "break" } ) ";"
+BreakStmt =
+  ( "break" LabelName ";"
+  | "continue" LabelName ";"
+  | { "break" } "break" ";"
+  | { "break" } "continue" ";"
+  )
```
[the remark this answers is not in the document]
RustanLeino: Good suggestion.

Source/Dafny/Resolver.cs
```diff
@@ -11075,9 +11075,14 @@ public class ResolveTypeReturn {
             s.TargetStmt = target;
           }
         } else {
-          if (loopStack.Count < s.BreakCount) {
-            var jump = s.IsContinue && s.BreakCount != 1 ? "break/continue" : s.Kind;
-            reporter.Error(MessageSource.Resolver, s, $"trying to {jump} out of more loop levels than there are enclosing loops");
+          var jumpStmt = s.BreakCount == 1 ?
+            $"a non-labeled '{s.Kind}' statement" :
+            $"a '{Util.Repeat(s.BreakCount - 1, "break ")}{s.Kind}' statement";
```
MikaelMayer, line 11080: s.BreakCount might be zero at this stage, which could trigger an infinite loop. Consider either testing it or changing Repeat so that 0 or -1 are considered the same.

MikaelMayer (changes requested):
Just a possible infinite loop to fix, otherwise we good to go !

Source/Dafny/Resolver.cs
RustanLeino: I disagree. There's an invariant in class `BreakStmt` that says

```
Contract.Invariant(TargetLabel != null || 1 <= BreakCount);
```

and the `if` statement surrounding the assignment to `jumpStmt` ensures that this code is used only when `s.TargetLabel == null`.

The count `s.BreakCount` include the final `continue`. So, the name of this field is not the best. Would `BreakContinueCount` or `Levels` be more readily understood?

MikaelMayer: Good argument. I can't see the surrounding block in GitHub because the file is too large, so I couldn't have had this assumption, I'm just watching over your shoulder :-)
However, while debugging Dafny, I have seen surprisingly many times invariants which did not hold but the code did not crash. It looks like the invariants are not checked (how odd), and I wouldn't be surprised if some refactoring down the road silently breaks such an invariant.
My not-so-strong opinion is that, until we migrate our codebase to Dafny, it is better not to rely on contracts. Changing a `== 0` to a `<= 0` seems like a very minor change that would it even more easily to reason locally.
If you prefer not to do that change, at least put a comment recalling the BreakStmt invariant on the else branch.

MikaelMayer: About the variable name, I think it would be better to have it meaning what it does, since It's a public API: breakCount looks instinctively like the number of times the word "break" appears. "level" might thus be more appropriate, but you could as well name it explicitly "SurroundingLoopLevelTarget".

RustanLeino: Unfortunately, `Contract.Invariant`, `Contract.Requires`, and `Contract.Ensures` contracts are not checked at run time, so it happens they are out of date or that bugs in the code are missed. Still, they often express the intent, and in this case the intent can be checked manually at least back to the constructor. For the particular code in this PR, I added a `Contract.Assert` (which is checked at run time in our `DEBUG` build) at this source location. I also added a comment that describes why I expect the asserted condition to hold.

RustanLeino: I renamed `BreakCount` to `BreakAndContinueCount`.

MikaelMayer (approved):
I think we can remove the comment now since the variable is self-explaining.
Otherwise, looks good !

MikaelMayer (approved):

PR #1840, opened 2022-02-18
fix: File contains no code even when there is code in prefixed modules

### LINKED ISSUE #1838: Inconsistent submodule behavior
Consider a Dafny file `A.dfy` with the following contents:

```
module A.B {
  type u64 = x | 0 <= x <= 0xffff_ffff_ffff_ffff
}
```

When verifying this file with the `/warningsAsErrors` option, we get the following error, even though the file is not empty:

```
$ dafny /compile:0 /warningsAsErrors A.dfy 
(0,-1): Error: File contains no code: A.dfy
1 resolution/type errors detected in A.dfy
```

But if we rewrite the above module declaration as follows, the warning / error goes away:

```
module A { 
  module B {
    type u64 = x | 0 <= x <= 0xffff_ffff_ffff_ffff
  }
}
```

```
% dafny /compile:0 /warningsAsErrors A.dfy

```

Can we make the dotted module declaration behave like the nested version?  

Thanks!

MikaelMayer:
Fix #1838 

This PR changes `Dafny.atg` so that it now reads (added part in bold)

>if (membersDefaultClass.Count == 0 && defaultModule.Includes.Count == 0 && defaultModule.TopLevelDecls.Count == 0
		   **&& (defaultModule.PrefixNamedModules == null || defaultModule.PrefixNamedModules.Count == 0)**) {

That way, it does not trigger the warning "File contains no code" if there are some prefixed modules inside the main module.
The original files that were testing this warning were actually testing [when we accidentally put a comment that comments the entire file](https://github.com/dafny-lang/dafny/blob/master/Test/git-issues/git-issue-845.dfy), and another file that [does not actually contain code](https://github.com/dafny-lang/dafny/blob/master/Test/wishlist/we-should-always-print-tooltips.dfy), so this PR is consistent with them.


RELEASE_NOTES.md
```diff
@@ -1,7 +1,7 @@
 # Upcoming
 
 - fix: No output when compiling to JavaScript on Windows (https://github.com/dafny-lang/dafny/pull/1824)
-
+- fix: No warning "File contains no code" if module contains a submodule (https://github.com/dafny-lang/dafny/pull/1840)
```
cpitclaudel, line 4: Should this be in the release notes?

MikaelMayer: Isn't this the file RELEASE_NOTES.md ? Note how in 1 commit, I was able to predict the pull request number :-)

Test/git-issues/git-issue-1838.dfy
```diff
@@ -0,0 +1,6 @@
+// RUN: %dafny /useBaseNameForFileName /compile:0 %S/git-issue-1838.dfy > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+module A.B {
+  type u64 = x | 0 <= x <= 0xffff_ffff_ffff_ffff
+}
```
cpitclaudel, line 6: Github complains that there's no newline at the end of your file

RELEASE_NOTES.md
cpitclaudel: I meant that maybe this should not be mentioned in the release notes?

MikaelMayer: We can always decide to remove it later, right @fabiomadge ?

fabiomadge (approved):

PR #1842, opened 2022-02-19
docs: Completed (and corrected) IteratorSpec section

robin-aws:


docs/DafnyRef/Specifications.md
```diff
@@ -520,16 +520,44 @@ IteratorSpec =
...
+A `yield ensures` clause applies to the `MoveNext` method only
+when another element is yielded, while an `ensures` clause similarly
+applies to the `MoveNext` method only when another element is NOT yielded.
+This means `yield ensures` clauses must hold for each `yield` statement
+in an iterator's body, while `ensures` clauses must hold when the
+body completes.
+
+Note also that all iterators define an implicit `Valid()` predicate
+that is used as a pre-condition for `MoveNext`. An iterator is only
+guaranteed to be `Valid()` after invoking `MoveNext` if that method
```
RustanLeino, line 535: I suggest moving the "only" to just before the "if"

RustanLeino (changes requested):
Would you mind also describing `_new` (including the fact that an iterator body can _remove_ objects from `_new`) and the `x`/`xs` pun in specifications? Maybe showing one small example that uses these would do the trick.

PR #1855, opened 2022-02-23
fix: Don’t entirely delete by-method part in export scope checks

### LINKED ISSUE #1852: `function-by-method` becomes `function method` with export set
The following raises an error on line 7: `Raised while checking export set A: function calls are allowed only in specification contexts (consider declaring the function a 'function method')`

```
module A {
  export
    reveals F, G

  function G(): int { 5 }
  function F(): int {
    G()
  } by method {
    return 5;
  }
}
```

No errors will be raised if export set is removed. It appears that when export set is declared on a function-by-method, it sees `F` as a `function method` and ignores the `by-method` part.

RustanLeino:
Previously, the by-method part of a function-by-method was entirely erased in the module clone used to perform export-set consistency checks. That was incorrect. The by-method part should be erased only if the function-by-method is export-provided. If the function-by-method is export-revealed, then the appropriate thing to do for the clone is to remove the contents of the by-method body, but still keep the empty body around so that the function still looks like a function-by-method.

Fixes #1852


RELEASE_NOTES.md
```diff
@@ -2,7 +2,7 @@
 
 - feat: `continue` statements. Like Dafny's `break` statements, these come in two forms: one that uses a label to name the continue target and one that specifies the continue target by nesting level. See section [19.2](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-break-continue) of the Reference Manual. (https://github.com/dafny-lang/dafny/pull/1839)
 - fix: No warning "File contains no code" if a file only contains a submodule (https://github.com/dafny-lang/dafny/pull/1840)
-
+- fix: export-reveals of function-by-method now allows the function body to be ghost
```
fabiomadge, line 5:
```suggestion
- fix: export-reveals of function-by-method now allows the function body to be ghost (https://github.com/dafny-lang/dafny/pull/1855)

```
I believe that I links to the full PRs are useful, but we can also agree to skip them to avoid the noise.

MikaelMayer: I like it to see links to PRs in the release notes.

cpitclaudel (approved):
LGTM

PR #1857, opened 2022-02-24
fix(lsp): Catch didChange errors as event/telemetry notifications

cpitclaudel:
Previously errors were caught in a try/catch block in `HandleUpdateAndPublishDiagnostics`, but this stopped being the case after https://github.com/dafny-lang/dafny/pull/1666.

The event/telemetry notification is useful for debugging, and maybe that's in fact what we should use instead of the current custom notifications?

FWIW, I think that the current code is as safe as the previous one, but I'm not sure that it's actually safe; see the FIXME that I added.

### DIFF
--- a/Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
+++ b/Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
@@ -1,5 +1,4 @@
 ﻿using MediatR;
-using Microsoft.Dafny.LanguageServer.Language;
 using Microsoft.Dafny.LanguageServer.Workspace;
 using Microsoft.Extensions.Logging;
 using OmniSharp.Extensions.LanguageServer.Protocol;
@@ -11,7 +10,7 @@ using System;
 using System.Threading;
 using System.Threading.Tasks;
 
-// Justification: The handler must not await document loads. Errors are handled within RunAndPublishDiagnosticsAsync.
+// Justification: The handler must not await document loads. Errors are handled within the observer set up by ForwardDiagnostics.
 #pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
 #pragma warning disable VSTHRD110 // Observe result of async calls
 
@@ -21,7 +20,7 @@ using System.Threading.Tasks;
 namespace Microsoft.Dafny.LanguageServer.Handlers {
   /// <summary>
   /// LSP Synchronization handler for document based events, such as change, open, close and save.
-  /// The documents are managed using an implementaiton of <see cref="IDocumentDatabase"/>.
+  /// The documents are managed using an implementation of <see cref="IDocumentDatabase"/>.
   /// </summary>
   /// <remarks>
   /// The <see cref="CancellationToken"/> of all requests is not used here. The reason for this is because document changes are applied in the
@@ -35,13 +34,16 @@ namespace Microsoft.Dafny.LanguageServer.Handlers {
 
     private readonly ILogger logger;
     private readonly IDocumentDatabase documents;
+    private readonly ITelemetryPublisher telemetryPublisher;
     private readonly IDiagnosticPublisher diagnosticPublisher;
 
     public DafnyTextDocumentHandler(
-      ILogger<DafnyTextDocumentHandler> logger, IDocumentDatabase documents, IDiagnosticPublisher diagnosticPublisher
+      ILogger<DafnyTextDocumentHandler> logger, IDocumentDatabase documents,
+      ITelemetryPublisher telemetryPublisher, IDiagnosticPublisher diagnosticPublisher
     ) {
       this.logger = logger;
       this.documents = documents;
+      this.telemetryPublisher = telemetryPublisher;
       this.diagnosticPublisher = diagnosticPublisher;
     }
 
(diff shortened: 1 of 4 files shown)


Source/DafnyLanguageServer/Workspace/ITelemetryPublisher.cs
```diff
@@ -0,0 +1,37 @@
...
+      UpdateComplete,
+      UnhandledException
+    }
+
+    /// <summary>
+    /// Publish a telemetry event.
+    /// </summary>
+    /// <param name="kind">The kind of telemetry event.</param>
+    /// <param name="evt">The telemetry event.</param>
+    protected void PublishTelemetry(TelemetryEventKind kind, object? evt);
```
camrein, line 21: It feels super strange to me to have `protected` members in an interface. Although that's a feature introduced with C# 8. I've just noticed this because it's rather unusual to declare accessibility in interfaces since all members are public by default.
However, I'm not a fan of abbreviations. Why not use `payload` or something like that? The event kind is determined by the `kind` parameter, so it's not actually a full event object from my perspective.

Source/DafnyLanguageServer/Workspace/TelemetryPublisher.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.Dafny.LanguageServer.Workspace {
+  public class TelemetryPublisher : ITelemetryPublisher {
+    private readonly ILanguageServerFacade languageServer;
+
+    public TelemetryPublisher(ILanguageServerFacade languageServer) {
+      this.languageServer = languageServer;
+    }
+
+    void ITelemetryPublisher.PublishTelemetry(ITelemetryPublisher.TelemetryEventKind kind, object? evt) {
+      languageServer.Window.SendTelemetryEvent(new TelemetryEventParams() {
```
camrein, line 19: The parentheses `()` are redundant when defining the object initializer expression.

camrein (changes requested):
I see the benefit for this and it probably makes sense disclosing errors as telemetry.
However, if we'd like to display this information in the client implementations  think a strongly typed alternative with a clear API would be more suitable (not a strong opinion).

Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
```diff
@@ -70,21 +72,47 @@ public class DafnyTextDocumentHandler : TextDocumentSyncHandlerBase {
...
+
+      public void OnCompleted() {
+        telemetryPublisher.PublishUpdateComplete();
+      }
+
+      public void OnError(Exception e) {
+        if (e is TaskCanceledException) {
+          OnCompleted();
+        } else {
+          // FIXME: Since we return from the caller without `await`ing, don't we risk closing `logger` too early?
```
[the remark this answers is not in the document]
cpitclaudel: I was worried that the `logger` object might be closed before this code runs, but it looks like it gets closed only when it gets disposed, so we're probably all good.

Source/DafnyLanguageServer/Workspace/ITelemetryPublisher.cs
cpitclaudel: Good points, fixed!

cpitclaudel:
> However, if we'd like to display this information in the client implementations think a strongly typed alternative with a clear API would be more suitable (not a strong opinion).

I agree, but I think we can cross that bridge when we get there.  The reason I introduced the `kind` parameter was precisely to give us some room later to adjust the payload shapes.

Source/DafnyLanguageServer/Workspace/ITelemetryPublisher.cs
cpitclaudel: Actually, I fixed the parameter name, but not the interface part.  Do you think I should make it an abstract class instead of an interface?  I'm using this interface + implementation pattern because that's the style of the codebase, but I didn't give it that much thought.

I made the method protected because it shouldn't be exposed to clients — the only thing that they should call are the higher-level methods.

keyboardDrummer (approved):

Source/DafnyLanguageServer/Workspace/TelemetryPublisher.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public TelemetryPublisher(ILanguageServerFacade languageServer) {
+      this.languageServer = languageServer;
+    }
+
+    void ITelemetryPublisher.PublishTelemetry(ITelemetryPublisher.TelemetryEventKind kind, object? payload) {
+      languageServer.Window.SendTelemetryEvent(new TelemetryEventParams {
+        ExtensionData = new Dictionary<string, object> {
+          {"kind", kind.ToString()},
+          {"data", payload ?? new Dictionary<string, object>()}
```
camrein, line 22: Maybe rename `data` to `payload` to reflect the argument name.
And personal preference: I'd leave the `data` property of the serialized object if it's empty rather than falling back to an empty dictionary.

camrein (approved):

Source/DafnyLanguageServer/Workspace/TelemetryPublisher.cs
cpitclaudel: Rename done.  I don't know how to leave the payload empty: annoyingly `TelemetryEventParams` expects a `Dictionary<string, object>`, not a `Dictionary<string, object?>`, so I can't put `null` there.

keyboardDrummer (approved):

camrein (approved):

PR #1864, opened 2022-02-26
fix(lsp): Memory leak in call to InferAndVerify (closes #1863) 

### LINKED ISSUE #1863: Incorrect `programId` passed to Boogie causes memory leak in LSP server
Boogie accepts a `programId` along with each program in `ExecutionEngine.InferAndVerify`.  This ID is used as a key to partition a cache of recent parses:

https://github.com/boogie-org/boogie/blob/ce03249fa06a7442df96352bb55e1bfba3e30365/Source/ExecutionEngine/ExecutionEngine.cs#L253-L257

A reasonable choice of key, for example, is the path of the current program.  Alternatively, if there is only one program being verified, using `null` is also reasonable; in that case Boogie uses a default string (https://github.com/boogie-org/boogie/blob/ce03249fa06a7442df96352bb55e1bfba3e30365/Source/ExecutionEngine/ExecutionEngine.cs#L318-L321).  This is what the Dafny CLI does:

https://github.com/dafny-lang/dafny/blob/38be5325dca74bd1d6af82fd2a8e764e4ede910f/Source/DafnyDriver/DafnyDriver.cs#L263

https://github.com/dafny-lang/dafny/blob/b6d7ae8c9376dbb253690823c65e1a85e8c977c7/Source/Dafny/DafnyMain.cs#L236-L239

The legacy `DafnyServer` used by Emacs uses just a module name, because it can't operate on more than one file at a time, so the file name part isn't needed:

https://github.com/dafny-lang/dafny/blob/495e7d2667a8768590feafc3c99562112e751152/Source/DafnyServer/DafnyHelper.cs#L85

The LSP server does something different:

https://github.com/dafny-lang/dafny/blob/5becffac988e1cb5890c3278a6b196da5a7ffd56/Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs#L103-L107

The test generator does the same thing, which is equally wrong, I think:

https://github.com/dafny-lang/dafny/blob/b6d7ae8c9376dbb253690823c65e1a85e8c977c7/Source/DafnyTestGeneration/ProgramModification.cs#L85-L88

In both cases one new program entry is created in the cache every time Boogie is queries. The cache's default expiration policy is a 10-minutes window, so for 10 minutes memory usage just grows.

The fix is to pass the program name instead of a freshly created unique string as the `programId`.  I will open a PR soon.

cpitclaudel:


camrein:
Would mind explaining me why the original implementation caused a memory-leak?
If I understood your change correctly, your fix is ensuring that the same program id is used for the same code. Extending upon this, I think the memory-leak is not fully resolved but only mitigated since boogie retains information about previous builds. And essentially, these previous builds are yet to be freed (for example when closing the document).

Source/Dafny/DafnyMain.cs
```diff
@@ -233,10 +233,7 @@ private class IncludeComparer : IComparer<Include> {
 
     public static bool BoogieOnce(string baseFile, string moduleName, Microsoft.Boogie.Program boogieProgram, string programId,
       out PipelineStatistics stats, out PipelineOutcome oc) {
-      if (programId == null) {
-        programId = "main_program_id";
-      }
-      programId += "_" + moduleName;
+      programId = (programId ?? "main_program_id") + "_" + moduleName;
```
camrein, line 236: I suggest creating a shared method.

keyboardDrummer:
@cpitclaudel how does the cache know when an item is still because a file has been modified?

cpitclaudel:
> Would you mind explaining me why the original implementation caused a memory-leak?

Boogie keeps a cache with one entry per file being edited.  Entries are automatically discarded after 10 minutes of inactivity.
That cache isn't expected to grow past a few entries, so the 10 minutes expiration deadline is reasonable.

Our code in the LSP server was telling boogie that each new request was a new file.  So (with caching enabled) Boogie ended up with hundreds of open "files".  They were being discarded after ten minutes, but since we send changes on every edits that could still amount to hundreds of files.

cpitclaudel:
> @cpitclaudel how does the cache know when an item is stale because a file has been modified?

The cache serves as a starting point for processing new files: Boogie gets a file with a cache key, loads the corresponding cache entry, updates that cache entry based on the contents of the new file, and then stores it back.  So there's no notion of an entry being stale really (or equivalently entries are always assumed to be stale)

keyboardDrummer (approved):

PR #1870, opened 2022-02-28
Improve parallelism

keyboardDrummer:
### Changes

- Make `Microsoft.Dafny.Type.PushScope` and related APIs thread-safe.
- Verify Dafny modules in parallel in the Dafny IDE
- Verify Dafny modules in parallel in the Dafny CLI
- Replace `Utils.CaptureConsoleOutput` with new APIs that take a TextWriter instead of defaulting to `Console.Out`

### Tests
- Resolve instability in unit tests that relied on slow proofs, by adding a `{:slow}` attribute only available in tests that can be used to tell Dafny to never finish verifying this proof.
- Enable `/vcsCores:2` in all Dafny integration tests


Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -25,6 +26,7 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
     }
 
     private void PublishDocumentDiagnostics(DafnyDocument document) {
+      Console.WriteLine($"Publishing document version {document.Version} with {GetDiagnostics(document).Count()} diagnostics.");
```
MikaelMayer, line 29: Can you please put this in a logger rather than in Console.WriteLine?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -84,34 +79,13 @@ IOptions<DafnyPluginsOptions> compilerOptions
     }
 
     public async Task<DafnyDocument> LoadAsync(TextDocumentItem textDocument, CancellationToken cancellationToken) {
-      var request = new LoadRequest(textDocument, cancellationToken);
-      requestQueue.Add(request, cancellationToken);
-      return await request.Document.Task;
+#pragma warning disable CS1998
+      return await await Task.Factory.StartNew(async () => LoadInternal(textDocument, cancellationToken), cancellationToken,
```
MikaelMayer, line 83: Can you get rid of an `await` ? I am not sure but I don't see a reason why to use an `async` closure here.
```suggestion
      return await Task.Factory.StartNew(() => LoadInternal(textDocument, cancellationToken), cancellationToken,
```

Test/logger/CSVLogger.dfy
```diff
@@ -6,7 +6,7 @@
 // wasn't populating the resource count test result property, and the CSV logger was
 // defaulting to 0 for that column.
 // CHECK-NOT: Impl\$\$_module.__default\.ExampleWithSplits\$\$.*,Passed,.*,0
-// CHECK-NEXT: Impl\$\$_module.__default\.ExampleWithSplits\$\$1,Passed,.*,.*
+// CHECK: Impl\$\$_module.__default\.ExampleWithSplits\$\$1,Passed,.*,.*
```
MikaelMayer, line 9: Can you please explain this?

MikaelMayer (changes requested):
Great job ! A few comments to get things sorted out.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -64,52 +66,52 @@ public class DafnyProgramVerifier : IProgramVerifier {
...
       try {
         // The printer is responsible for two things: It logs boogie errors and captures the counter example model.
         var errorReporter = (DiagnosticErrorReporter)program.reporter;
         var printer = new ModelCapturingOutputPrinter(logger, errorReporter, progressReporter);
-        DafnyOptions.O.Printer = printer;
         // Do not set these settings within the object's construction. It will break some tests within
         // VerificationNotificationTest and DiagnosticsTest that rely on updating these settings.
         DafnyOptions.O.TimeLimit = options.TimeLimit;
         DafnyOptions.O.VcsCores = GetConfiguredCoreCount(options);
+        DafnyOptions.O.Printer = printer;
```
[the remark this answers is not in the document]
keyboardDrummer: In Boogie there are no longer any static references to Options, but in Dafny there still are. I will remove them in another PR.

Test/logger/CSVLogger.dfy
keyboardDrummer: The order of elements in the XML has become non-deterministic due to the concurrency, and CHECK-NEXT is very specific about the order.

MikaelMayer (dismissed):
Every looks good to me! I wasn't sure of the `Unwrap()` but I don't see how to make without.

Source/Dafny/DafnyMain.cs
```diff
@@ -231,9 +232,12 @@ private class IncludeComparer : IComparer<Include> {
...
     }
 
-    public static bool BoogieOnce(ExecutionEngine engine, string baseFile, string moduleName, Microsoft.Boogie.Program boogieProgram, string programId,
-      out PipelineStatistics stats, out PipelineOutcome oc) {
-      programId = (programId ?? "main_program_id") + "_" + moduleName;
+    public static async Task<(bool IsVerified, PipelineOutcome Outcome, PipelineStatistics Statistics)> BoogieOnce(
+      TextWriter output,
+      ExecutionEngine engine, string baseFile,
+      string moduleName,
+      Microsoft.Boogie.Program boogieProgram, string programId) {
```
cpitclaudel, line 239: Curly on its own line

Source/Dafny/DafnyMain.cs
```diff
@@ -231,9 +232,12 @@ private class IncludeComparer : IComparer<Include> {
...
+      programId ??= "main_program_id" + "_" + moduleName;
```
cpitclaudel, line 240: ?

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -304,79 +318,95 @@ public enum CommandLineArgumentsResult {
...
+    public async Task<(bool IsVerified, PipelineOutcome Outcome, IDictionary<string, PipelineStatistics> ModuleStats)>
+      Boogie(string baseName, IEnumerable<Tuple<string, Bpl.Program>> boogiePrograms, string programId) {
+
+      var concurrentModuleStats = new ConcurrentDictionary<string, PipelineStatistics>();
+      var writerManager = new ConcurrentToSequentialWriteManager(Console.Out);
+
+      var moduleTasks = boogiePrograms.Select(async program => {
+        await using var moduleWriter = writerManager.AppendWriter();
+        // ReSharper disable once AccessToDisposedClosure
+        return await Task.Run(() => BoogieOnceWithTimer(moduleWriter, baseName, programId, concurrentModuleStats, program.Item1,
```
cpitclaudel, line 330: Can we do a destructuring bind on `program` to avoid `.Item1` and `.Item2`?

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -304,79 +318,95 @@ public enum CommandLineArgumentsResult {
...
+          program.Item2));
+      }).ToList();
 
-      bool isVerified = true;
-      outcome = PipelineOutcome.VerificationCompleted;
-      statss = new Dictionary<string, PipelineStatistics>();
+      await Task.WhenAll(moduleTasks);
+      var outcome = moduleTasks.Select(t => t.Result.Outcome)
+        .Aggregate(PipelineOutcome.VerificationCompleted, MergeOutcomes);
```
cpitclaudel, line 336: Is that the right logic?  It looks like we'll wait for the slowest of all modules, rather than stopping after the first failure?

Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
```diff
@@ -15,11 +15,17 @@
...
+using MediatR;
+using Microsoft.Dafny.LanguageServer.Language;
+using OmniSharp.Extensions.LanguageServer.Client;
 
 namespace Microsoft.Dafny.LanguageServer.IntegrationTest {
   public class DafnyLanguageServerTestBase : LanguageServerTestBase {
-    protected const string SlowToVerify = @"
-lemma {:timeLimit 10} SquareRoot2NotRational(p: nat, q: nat)
+    protected readonly string SlowToVerify = @"
+lemma {:timeLimit 3} SquareRoot2NotRational(p: nat, q: nat)
```
cpitclaudel, line 28: Why `const` to `readonly`?

Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
```diff
@@ -43,16 +55,60 @@ public class DafnyLanguageServerTestBase : LanguageServerTestBase {
...
     public DafnyLanguageServerTestBase() : base(new JsonRpcTestOptions(LoggerFactory.Create(builder => builder.AddConsole().SetMinimumLevel(LogLevel.Warning)))) { }
 
-    protected override (Stream clientOutput, Stream serverInput) SetupServer() {
+
+
+    protected virtual async Task<ILanguageClient> InitializeClient(
+      Action<LanguageClientOptions> clientOptionsAction = null,
+      [CanBeNull] Action<LanguageServerOptions> serverOptionsAction = null) {
+      var client = CreateClient(clientOptionsAction, serverOptionsAction);
+      await client.Initialize(CancellationToken).ConfigureAwait(false);
```
cpitclaudel, line 64: Why do you need the configureawait(false) here?

Source/DafnyLanguageServer.Test/Various/CompilationStatusNotificationTest.cs
```diff
@@ -185,21 +170,7 @@ method Abs(x: int) returns (y: int)
...
-    calc == {
-      (2 * q - p) * (2 * q - p);
-      4 * q * q + p * p - 4 * p * q;
-      {assert 2 * q * q == p * p;}
-      2 * q * q + 2 * p * p - 4 * p * q;
-      2 * (p - q) * (p - q);
-    }
-  }
-}".TrimStart();
+      var source = SlowToVerify.TrimStart();
```
cpitclaudel, line 173: No need to trimstart

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -122,35 +129,25 @@ decreases y
...
-      (2 * q - p) * (2 * q - p);
-      4 * q * q + p * p - 4 * p * q;
-      {assert 2 * q * q == p * p;}
-      2 * q * q + 2 * p * p - 4 * p * q;
-      2 * (p - q) * (p - q);
-    }
-  }".TrimStart();
+lemma {:neverVerify} HasNeverVerifyAttribute(p: nat, q: nat)
+  ensures true
+{".TrimStart();
```
cpitclaudel, line 134: Why not use `NeverVerifies` from above?

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -29,7 +31,7 @@ public class DafnyProgramVerifier : IProgramVerifier {
 
     DafnyOptions Options => DafnyOptions.O;
 
-    private DafnyProgramVerifier(ILogger<DafnyProgramVerifier> logger, VerifierOptions options) {
+    private DafnyProgramVerifier(ILogger<IProgramVerifier> logger, VerifierOptions options) {
```
cpitclaudel, line 34: ?

cpitclaudel (changes requested):
Thanks for this, looks like a great platform that we'll be able to build on very soon.

Do you have a sense of "what", concretely, becomes more concurrent after this?  In a regular invocation of Dafny like we do today, on the CLI, how will the control flow change?

Also, is there a way to turn concurrency off?  With Dafny's single-threaded mode I'm already running out of RAM on my laptop (Z3 often taking ~1GB, times the number of server instances already open?)

Source/DafnyTestGeneration.Test/Setup.cs
```diff
@@ -5,7 +5,7 @@ namespace DafnyTestGeneration.Test {
   public class Setup {
 
     public static void SetupDafnyOptions(string[] extraArgs = null) {
-      var options = new DafnyOptions();
+      var options = DafnyOptions.Create(extraArgs);
       options.Parse(extraArgs ?? System.Array.Empty<string>());
```
atomb, line 9: This looks sort of like it's processing `extraArgs` twice. Is that true?

atomb (dismissed):
I focused mostly on the CSV logger change.

Source/Dafny/DafnyMain.cs
keyboardDrummer: I'll change it, but why?

Source/Dafny/DafnyMain.cs
keyboardDrummer: This code was already there in the form `programId = (programId ?? "main_program_id") + "_" + moduleName;`

Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
keyboardDrummer: `const` was not possible because of the `TrimStart()` call

Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
keyboardDrummer: I don't know. I copied this from `OmniSharp.Extensions.LanguageProtocol.Testing.LanguageServerTestBase`

Source/Dafny/DafnyMain.cs
keyboardDrummer: The formatter puts it on the same line

MikaelMayer: Let's stick with dotnet-format then.

MikaelMayer (dismissed):

Source/Dafny/DafnyMain.cs
cpitclaudel: For readability: `var moduleId` below ends up on the same column as `Microsot.Boogie`.  Let's stick with dotnet-format though, sorry for the noise.

cpitclaudel (approved):

MikaelMayer (approved):
Looks good, thanks !

Issue #1390, opened 2021-08-25
Contract.Assert does not cause failures in published Dafny builds


robin-aws:
Follow-up from #1347 - we've addressed the actual assertion failure, but not the fact that the assertion was only hit when building Dafny from source.

AFAICT, it's likely an issue with how we are calling `dotnet publish` in `Scripts/package.py` - perhaps the `-c Checked` argument is not being picked up by dependent packages?

robin-aws:
It looks like this broke back when we moved to .NET Core and simplified the `*.csproj` files: https://github.com/dafny-lang/dafny/pull/794

The `Contracts` methods are conditionally compiled only if there are matching symbols defined. In particular, `Assert` will be active if `DEBUG` is defined. The `Checked` configuration (and others) used to define `TRACE;DEBUG;NO_ENABLE_IRONDAFNY`, but now only define `TRACE`.

@sorawee To address your case, it looks like `Requires` is only active if `CONTRACTS_FULL` is defined, not just `DEBUG`, so AFAICT those calls are never evaluated (which is a related but separate problem - I've not yet tracked down at what point we lost or intentionally stopped exercising that mode).

keyboardDrummer:
I believe contracts don't work in .NET Core: https://github.com/dotnet/docs/issues/6361

Only `Contracts.Assert` still has an effect and it's the same as `Debug.Assert`, meaning it will throw an error if the assertion fails when the `DEBUG` constants is defined, which is when running a debug build.

sorawee:
I separated my issue to #1438.

PR #1871, opened 2022-02-28
feat: Modules in the python backend

fabiomadge:
Adds support for modules, and increases the stinginess w.r.t. whitespace in the generated code.


Source/Dafny/Compilers/Compiler.cs
```diff
@@ -1194,6 +1196,8 @@ public enum ResolvedUnaryOp { BoolNot, BitwiseNot, Cardinality }
             var w = DeclareDatatype(dt, wr);
             if (w != null) {
               CompileClassMembers(program, dt, w);
+            } else if (this is not CppCompiler) {
```
fabiomadge, line 1199: The alternative is living with a `\n` when there's a tuple. I personally don't like either.

cpitclaudel: Could this be an attribute of each of the compilers that the main code checks to decide whether to add the `\n`?

Source/Dafny/Compilers/Compiler.cs
```diff
@@ -1238,8 +1242,10 @@ public enum ResolvedUnaryOp { BoolNot, BitwiseNot, Cardinality }
             }
           } else if (d is ValuetypeDecl) {
             // nop
+            wr.DeleteLast();
```
cpitclaudel, line 1245: That's a bit scary: are we sure that we know which character this will be deleting?

fabiomadge: Reasonably so, the test go trough after all. At least in this case it should be.

Source/Dafny/Compilers/Compiler.cs
fabiomadge: This code is a little strange, we always add a `\n` in L1161 for every TopLevelDecl that's compiled. Sometimes we only find out later that we don't have to do anything. This leaves unsightly whitespace in the generated sources. Looking it is no fun.

The concrete problem here is a hack in the cpp compiler. They include the ClassMembers on the first pass, so they can't have the second one. To avoid that, they act like they didn't do anything.

Source/Dafny/ConcreteSyntax/ConcreteSyntaxTree.cs
```diff
@@ -7,10 +7,11 @@
 
 namespace Microsoft.Dafny {
   public enum BraceStyle {
```
robin-aws, line 9: Since you're already touching most of the references to this type, I think the time is right to fix the name and call it `BlockStyle` instead. That will resolve the weird redundancy in `BraceStyle.Brace` and just plain contradiction in `BraceStyle.NewlineNoBrace` (which would just become `BlockStyle.Newline`).

I think then `Space` would also be renamed to `SpaceBrace`?

Source/Dafny/Compilers/Compiler.cs
robin-aws: Sounds like we can still define an attribute, even if it's a weird name like `CompilesClassMembersOnTheFirstPass` or whatever. :)

Referring to concrete subclasses here is just a recipe for weird hard-to-debug edge cases as we add more compilers.

Source/Dafny/Compilers/Compiler.cs
robin-aws: I agree this is very risky and makes the code an order of magnitude harder to reason about confidently. I'd define whatever additional `Compiler` attributes you need to specialize the behavior without doing this.

I hesitate to even mention it, but IF this was absolutely necessary, I would at least provide an argument to `DeleteLast` to specify WHAT you expect to be deleted, so at least you have a chance of catching bugs if the state is ever not what you thought it would be.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -161,30 +165,15 @@ protected class ClassWriter : IClassWriter {
...
     }
 
     private ConcreteSyntaxTree CreateMethod(Method m, List<TypeArgumentInstantiation> typeArgs, bool createBody,
       ConcreteSyntaxTree wr, bool forBodyInheritance, bool lookasideBody) {
-      if (!createBody) {
-        return null;
-      }
+      wr.Write($"def {IdName(m)}(");
 
-      var customReceiver = !forBodyInheritance && NeedsCustomReceiver(m);
```
robin-aws, line 174: Did all of this disappear because `customReceiver` is never true in our test cases so far?

robin-aws (changes requested):
Thanks for starting to tackle the confusing labelling on the `BraceStyle` enum!

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: Exactly.

Source/Dafny/Compilers/Compiler.cs
fabiomadge: Now, it's this one field we happen to overwrite in this one compiler.

Source/Dafny/Compilers/Compiler.cs
fabiomadge: I think it's reasonably nice now.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -499,15 +484,26 @@ protected class ClassWriter : IClassWriter {
...
+        out string compiledName, out string preString, out string postString) {
+      compiledName = "";
+      preString = "";
+      postString = "";
+      switch (id) {
+        case SpecialField.ID.UseIdParam:
+          compiledName = IdProtect((string)idParam);
+          break;
+        default:
+          Contract.Assert(false); // unexpected ID
```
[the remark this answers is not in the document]
fabiomadge: This is a very common pattern in our code base and all the other implementations of `GetSpecialFieldInfo` do it this way, but I can change it if you feel strongly about it. I'd argue that a user will notice it at the very latest when their generated code doesn't compile.

Source/Dafny/Compilers/Compiler-python.cs
robin-aws: Great, I'm definitely in favour of removing code with no testing coverage (assuming it's not easy to add a test instead).

Source/Dafny/Compilers/Compiler-python.cs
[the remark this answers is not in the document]
robin-aws: Okay, I'll let it go for now. I'll consider this something we should clean up when we address https://github.com/dafny-lang/dafny/issues/1390 more generally.

Source/Dafny/Compilers/Compiler.cs
```diff
@@ -132,13 +132,14 @@ protected interface IClassWriter {
...
     protected abstract IClassWriter CreateTrait(string name, bool isExtern, List<TypeParameter>/*?*/ typeParameters, List<Type>/*?*/ superClasses, Bpl.IToken tok, ConcreteSyntaxTree wr);
-    protected virtual bool SupportsProperties { get => true; }
+    protected virtual bool SupportsProperties => true;
     protected abstract ConcreteSyntaxTree CreateIterator(IteratorDecl iter, ConcreteSyntaxTree wr);
     /// <summary>
     /// Returns an IClassWriter that can be used to write additional members. If "dt" is already written
     /// in the DafnyRuntime.targetlanguage file, then returns "null".
     /// </summary>
     protected abstract IClassWriter/*?*/ DeclareDatatype(DatatypeDecl dt, ConcreteSyntaxTree wr);
+    protected virtual bool DatatypeDeclarationAndMemberCompilationAreSeparate => true;
```
robin-aws, line 142: Much better name than my suggestion :)

Source/Dafny/ConcreteSyntax/ConcreteSyntaxTree.cs
```diff
@@ -114,6 +115,12 @@ public T Append<T>(T node)
       return this;
     }
 
+    public void DeleteLast() {
```
robin-aws, line 118: Very glad we don't need this any more. Now let's remove the method so no one is tempted by the dark side in the future. :)

Source/Dafny/ConcreteSyntax/ConcreteSyntaxTree.cs
```diff
@@ -6,11 +6,12 @@
...
-  public enum BraceStyle {
+  public enum BlockStyle {
     Nothing,
-    Space,
     Newline,
-    NewlineNoBrace
+    Brace,
+    SpaceBrace,
+    NewlineBrace
   }
```
robin-aws, line 15: SO much better, thank you!

robin-aws (changes requested):
Much better, thanks for the improvements! Just the one small tweak and then LGTM.

robin-aws (approved):

PR #1872, opened 2022-02-28
docs: Document the -out compiler parameter

cpitclaudel:


docs/DafnyRef/UserGuide.md
```diff
@@ -357,7 +357,7 @@ TO BE WRITTEN
    * 2 - write the compiled program if it passes verification, regardless of the `-compile` setting
    * 3 - write the compiled program regardless of verification success and the `-compile` setting
 
-* `-out:<file>` - TODO
+* `-out:<file>` - sets the name to use for compiled code files (by default Dafny reuses the name of the file being compiled).  Used as-is by compilers that generate a single file (e.g. the C# backend will generate `<file>.dll` and optionally `<file>.cs` with `-spillTargetCode`).  Used as a directory name by compilers that generate multiple files (e.g. the Java backend will generate files in directory `<file>-java/`).  Any file extension is ignored, so `-out:<file>` is the same as `-out:<file>.<ext>` if `<file>` contains no periods.
```
MikaelMayer, line 360:
```suggestion
* `-out:<file>` - sets the name to use for compiled code files.

By default, Dafny reuses the name of the Dafny file being compiled.  Compilers that generate a single file use the file name as-is (e.g. the C# backend will generate `<file>.dll` and optionally `<file>.cs` with `-spillTargetCode`).  Compilers that generate multiple files use the file name as a directory name (e.g. the Java backend will generate files in directory `<file>-java/`).  Any file extension is ignored, so `-out:<file>` is the same as `-out:<file>.<ext>` if `<file>` contains no periods.
```

MikaelMayer: to match other documentation, and to use less passive voice.

MikaelMayer (changes requested):
Great job adding this documentation ! I learned something.

docs/DafnyRef/UserGuide.md
cpitclaudel: Committed, though I think the active voice makes things less clear here (the important part is really the command line argument, not the compilers, so putting the compilers first doesn't help).

MikaelMayer: I understand your point, which is totally coherent. My point was to make merely complete sentences, which was more important to me than the passive/active voice.
What about making it like this:
```
By default, Dafny reuses the name of the Dafny file being compiled. The name provided after `-out:` is:
* Used as a file name by compilers that generate a single file, e.g. the C# backend will generate `<file>.dll` and optionally `<file>.cs` with `-spillTargetCode`.
* Used as a directory name by compilers that generate multiple files, e.g. the Java backend will generate files in directory `<file>-java/
Any file extension is ignored, so `-out:<file>` is the same as `-out:<file>.<ext>` if `<file>` contains no periods.
```

cpitclaudel: I think your previous version is fine :)

MikaelMayer (approved):

PR #1874, opened 2022-03-01
Fill in some sections on refinement in the manual

atomb:
Add some documentation to the reference manual about how refinement works with methods, lemmas, functions, and predicates.


docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
 * one or both of the following holds:
    * the child has an initializing expression
-   * the child is declared `ghost` and the parent is not `ghost`, or vice versa
+   * the child is declared `ghost` and the parent is not `ghost`, or vice versa.
 
-To indicate it is a refining declaration, a refining `const` declaration
-contains an ellipsis in this syntax:
-```grammar
-"const" { Attribute } CIdentType "..." [ ":=" Expression ]
-```
```
cpitclaudel, line 108: Will it be clear what syntax is used for refining?

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
+    var y :| y > x;
+    return y;
+  }
+}
+
+module M1 refines M0 {
+  method m(x: int) returns (r: int)
+  {
+    ...;
+    return y + 1;
```
cpitclaudel, line 187: Can you show the resulting code?

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
+  method m(x: int) returns (r: int)
+    ensures r > x;
+  {
+    var y :| y > x;
+    return y;
+  }
+}
+
+module M1 refines M0 {
+  method m(x: int) returns (r: int)
```
cpitclaudel, line 184: Why not use `m...` here?

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
+definition to
+
+* provide a body missing in the parent,
+* replace the body of the parent with a new, semantically-compatible
+  body,
+* strengthen the postcondition of the parent method by adding one or more
+  `ensures` clauses,
+* provide a more deterministic version of a non-deterministic parent
+  body, or
+* amend the body of the parent method with additional statements.
```
cpitclaudel, line 119: Do we say somewhere what the correctness criterion is for refinement? (If child can return `x` then parent can return `x`).  IIUC these are all variants of that general pattern.

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
+paper](https://dl.acm.org/doi/10.1007/s00165-012-0254-3).
+
+A refining module can also introduce new `method` declarations or
+definitions that do not exist in the refinement parent.
 
 ## 21.6. Lemma declarations
 
-TODO
+As lemmas are (ghost) methods, the description of method refinement from
+the previous section also applies to lemma refinement.
```
cpitclaudel, line 203: One question though: what happens to lemmas that are invalidated by the refinement? (Are they impossible?).

For example for your earlier example, a lemma (which may not be expressible in Dafny?) is "method `m(0)` can return `3`".  This lemma is wrong after refinement, since `m(0) == 1`.

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
 
-TODO
+A child `function` (or `predicate`) definition can refine a parent
+`function` (or `predicate`) declaration or definition to
+
+* provide a body missing in the parent,
+* replace the body from the parent with a new, semantically-compatible
+  body, or
+* strengthen the postcondition of the parent function by adding one or more
+  `ensures` clauses.
```
cpitclaudel, line 214: But also weaken the preconditions presumably?
I wonder if a general description of the relation between the parent and the child might not be simpler than exhaustive enumeration

cpitclaudel (dismissed):
LGTM as is but I added suggestions.

docs/DafnyRef/Refinement.md
RustanLeino: The rules for superimposition in a refining module are such that no proofs are invalidated. The refining module must prove that any statement replacements are valid, that added statements don't fail, and that any new `return` correctly establishes the enclosing method's postcondition.

MikaelMayer (changes requested):
I learned a lot, thanks for filling these sections ! Suggesting some improvements.

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
 
 ## 21.5. Method declarations
 
-TODO
+A child `method` definition may refine a parent `method` declaration or
+definition to
+
+* provide a body missing in the parent,
+* replace the body of the parent with a new, semantically-compatible
+  body,
```
RustanLeino, line 114: Is this true? Going from memory, I think a body in the parent can only be changed by doing a combination of:

* making the child more deterministic
* changing `assume` statements into `assert` statements
* superimposing new statements onto the parent (you use the word "amend", whereas the literature tends to say "superposition refinement", and I might say "superimpose")

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
+* strengthen the postcondition of the parent method by adding one or more
+  `ensures` clauses,
```
RustanLeino, line 116: This list reads as if you have to pick one of the bullets. But the changes of the specification are independent of any change to the body.

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
+The type signature of a refining method must be the same as that of the
+parent method it refines. This can be ensured by providing an explicit
+type signature equivalent to that of the parent (with renaming of
+parameters allowed) or by using an ellipsis (`...`) to indicate copying
+of the parent type signature. The following example illustrates that
+latter case.
+
+```
+module M0 {
+  method m(x: int) returns (r: int)
```
RustanLeino, line 130: Capitalize names of methods

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
+
+To introduce additional statements, the child method can include
+ellipses within the body to stand in for portions of code from the
+parent body. Dafny then attempts to merge the body of the child with the
+body of the parent by filling in the ellipses. For example:
+
+```
+module M0 {
+  method m(x: int) returns (r: int)
+    ensures r > x;
```
RustanLeino, line 176: No semi-colon needed

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,133 @@ TODO
...
+    ...;
+    return y + 1;
+  }
+}
+```
+
+Refinement of method bodies can go far beyond adding statements to the
+end, however. Full details of the algorithm used to perform the merge
+operation are available in [this
+paper](https://dl.acm.org/doi/10.1007/s00165-012-0254-3).
```
RustanLeino, line 195: For anyone interested in these details, the long comment before class `RefinementTransformer` in `Source/Dafny/RefinementTransformer.cs` may also be useful.

Also, it would be good to point out that hover text in the IDE will tell you what statements were "inherited" from the parent. To try this, hover over a `...;` statement, over a `}` (before which there is an implicit `...;`), or some other statement that has been changed.

docs/DafnyRef/Refinement.md
RustanLeino: Oh, and new proof obligations (like newly introduced `ensures` clauses) are checked in the child.

docs/DafnyRef/Refinement.md
RustanLeino: This is sound, but there is no notation for it. (This is done differently for `trait`s and for refinement modules. In a type that extends a trait, the entire specification as to be given again. This is verbose, but makes it clear what the specification is and gives the freedom to, for example, weaken preconditions. For refinement modules, there are so many things that are not shown explicitly in the child, and the basic idea is to just "superimpose" things onto the parent. Thus, you can *add* `ensures` clauses, but can't add `modifies` or `requires` clauses, and there is no way to *delete* a previously given specification clause.

RustanLeino (changes requested):

docs/DafnyRef/Refinement.md
atomb: I partly removed this because it seemed wrong, since I'd never seen an example of `...` used with `const`. I experimented, though, and it does work. So I'll replace this with an example instead of a grammar.

docs/DafnyRef/Refinement.md
atomb: Good idea.

docs/DafnyRef/Refinement.md
atomb: Yeah, I think you're right. The previous one was explicit to demonstrate that possibility, but `...` probably makes more sense for this example.

docs/DafnyRef/Refinement.md
atomb: My understanding is that the correctness criteria for refinement are less strict than that, and are derived from that list. I may be wrong about that, though. @RustanLeino?

docs/DafnyRef/Refinement.md
cpitclaudel: !! That's a great example and if it's intended it should be in the manual: my expectation re. refinement is that behaviors of the child are found in the parent, but this isn't the case here.

docs/DafnyRef/Refinement.md
cpitclaudel: Solution: use `||` as the bullet :P

docs/DafnyRef/Refinement.md
cpitclaudel: > To try this, hover over a ...; statement, over a } (before which there is an implicit ...;), or some other statement that has been changed.

Sweet

docs/DafnyRef/Refinement.md
atomb: Good point!

docs/DafnyRef/Refinement.md
atomb: Oh, this is probably just due to the implicit `...` before the closing `}` combined with the rule that additional `return` statements are allowed!

cpitclaudel: Ah, so the method becomes `return x + 1; return x + 2;` after this?

atomb: My current understanding is that the implicit `...` at the end would match `return x + 1`, so it'd actually be `return x + 2; return x + 1;`

cpitclaudel: Indeed, it prints 2, thanks!

cpitclaudel: I guess the trick is that the visible part of the refinement is a refinement of the spec, not of the body, since method bodies are always opaque.  Hence refining x+1 into x+2 is valid as long as the spec of the `x+1` part is vague enough to also allow x+2.  And this doesn't work for functions (whose body is transparent) because you're not allowed to change function bodies.  Interesting.

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,205 @@ TODO
...
+type signature equivalent to that of the parent (with renaming of
+parameters allowed) or by using an ellipsis (`...`) to indicate copying
+of the parent type signature. The following example illustrates that
+latter case.
+
+To introduce additional statements, the child method can include
+ellipses within the body to stand in for portions of code from the
+parent body. Dafny then attempts to merge the body of the child with the
+body of the parent by filling in the ellipses. In the `ToSuperimpose`
+example, the explicit `...` at the beginning will expand to the variable
```
MikaelMayer, line 206: Where is the explicit "..." at the beginning?

MikaelMayer (changes requested):

docs/DafnyRef/Refinement.md
atomb: Fixed! I had a local file I was testing that wasn't fully reflected in the text.

MikaelMayer (approved):

docs/DafnyRef/Refinement.md
RustanLeino: Correctly deduced. Because of the implicit `...` before the `}`, the method body in `M1` becomes `return x + 1; return x + 2;`. It would be sound to let a method replace its entire body with something different, but there is no special syntax for doing so.

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,207 @@ TODO
...
 
-A parent `const` declaration may be refined by a refining `const` declaration
-if
+Const declarations can be refined as in the following example.
+
+```
+module A {
+  const ToDefine : int
+  const ToDefineWithoutType : int
+  const ToGhost : int := 1
```
RustanLeino, line 101: Style: Use no space before colon (here and below).

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,207 @@ TODO
...
+}
+
+module B refines A {
+  const ToDefine : int := 2
+  const ToDefineWithoutType ... := 3
+  ghost const ToGhost : int
+  const NewConst : int
+}
+
+
```
RustanLeino, line 111: Delete these empty lines

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,207 @@ TODO
...
+  method ToDeterminize(x: int) returns (r: int)
+    ensures r >= x
+  {
+    var y :| y >= x;
+    return y;
+  }
+
+  method ToSuperimpose(x: int) returns (r: int)
+  {
+    var y : int := x;
```
RustanLeino, line 146: No space before `:`

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,207 @@ TODO
...
+    if (y < 0) {
```
RustanLeino, line 147: Parentheses not needed (this also occurs a couple of more times below)

docs/DafnyRef/Refinement.md
```diff
@@ -92,32 +92,207 @@ TODO
...
 
 ## 21.7. Function and predicate declarations
 
-TODO
+Function (and equivalently predicate) declarations can be refined as in
+the following example.
+
+```
+module A {
+  function F(x: int) : (r: int)
```
RustanLeino, line 266: No space before `:` (also on L269)

RustanLeino (changes requested):
Looks great! Thanks for taking the time to describe all these cases. I only have some whitespace-like comments on the examples.

MikaelMayer (approved):

RustanLeino (approved):

PR #1894, opened 2022-03-09
Allow users to plug in custom compilers

cpitclaudel:
This is a rebase of #1708 on top of #1739.

@MikaelMayer I don't understand the plugins test; why are we generating a library on the fly in [Source/DafnyLanguageServer.Test/Various/PluginsTestBase.cs](https://github.com/dafny-lang/dafny/pull/1739/files#diff-d979f9e6f68f77b5f473b6d96e8d661044a077adcdf9c0de5003b086d0b82935) instead of just having checking in the source code?

I've marked two commits as NEEDS REVIEW, since I'm not sure the original code was doing the right thing in the first place.


Source/Dafny/Compilers/Compiler-java.cs
```diff
@@ -19,15 +19,33 @@
...
+    public override void LateInitialize(ErrorReporter reporter, ReadOnlyCollection<string> otherFileNames) {
+      base.LateInitialize(reporter, otherFileNames);
       IntSelect = ",java.math.BigInteger";
       LambdaExecute = ".apply";
     }
 
-    public override String TargetLanguage => "Java";
+    public override IReadOnlySet<string> SupportedExtensions => new HashSet<string> { ".java" };
+
+    public override string TargetLanguage => "Java";
```
MikaelMayer, line 32: Should we include the version of java as well? Just a question, no need to change anything.

Source/Dafny/Compilers/Compiler-java.cs
```diff
@@ -19,15 +19,33 @@
...
+    public override string TargetExtension => "java";
+    public override string Basename(string dafnyProgramName) =>
+      TransformToClassName(base.Basename(dafnyProgramName));
+    public override string TargetBaseDir(string baseName) => baseName + "-java";
+    public string TransformToClassName(string baseName) =>
+      Regex.Replace(baseName, "[^_A-Za-z0-9\\$]", "_");
```
MikaelMayer, line 38:
```suggestion
      Regex.Replace(baseName, "[^_A-Za-z0-9$]", "_");
```
We don't need the backslash because '$' is interpreted literally in a character range, let's fix this while you are refactoring this code.

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -17,13 +17,18 @@
...
+    public override IReadOnlySet<string> SupportedExtensions => new HashSet<string> { ".js" };
 
     public override string TargetLanguage => "JavaScript";
+    public override string TargetExtension => "js";
+
+    public override bool SupportsInMemoryCompilation => true;
+    public override bool TextualTargetIsExecutable => true;
+
+    public override IReadOnlySet<string> SupportedNativeTypes =>
+      new HashSet<string>(base.SupportedNativeTypes.Union(new List<string> { "number" }));
```
MikaelMayer, line 31: In the previous huge switch statement, the JavaScript supported native types was only "number". Why are you unioning it with the base.SupportedNativeTypes? Can you confirm it is safe?

Source/Dafny/Plugin.cs
```diff
@@ -5,44 +5,86 @@
...
+    public AutomaticPluginConfiguration(Assembly assembly) {
+      var types = assembly.GetTypes();
+
+      Rewriters = FindPluginComponents<Rewriter, Func<ErrorReporter, Rewriter>>(assembly, CreateRewriterFactory);
+      Compilers = FindPluginComponents<Compiler, Func<Compiler>>(assembly, CreateCompilerFactory);
+
+      // Checks about the plugin to be well-behaved.
+      if (Rewriters.Length == 0 && Compilers.Length == 0) {
+        throw new Exception($"Plugin {assembly.Location} does not contain any supported plugins.  " +
+                            "Expecting on of the following:\n" +
```
MikaelMayer, line 53:
```suggestion
                            "Expecting one of the following:\n" +
```

Source/Dafny/Plugin.cs
```diff
@@ -5,44 +5,86 @@
...
-      return Rewriters.Select(funcErrorReporterRewriter =>
-        funcErrorReporterRewriter(errorReporter)).ToArray();
+    public AutomaticPluginConfiguration(Assembly assembly) {
+      var types = assembly.GetTypes();
+
+      Rewriters = FindPluginComponents<Rewriter, Func<ErrorReporter, Rewriter>>(assembly, CreateRewriterFactory);
+      Compilers = FindPluginComponents<Compiler, Func<Compiler>>(assembly, CreateCompilerFactory);
+
+      // Checks about the plugin to be well-behaved.
+      if (Rewriters.Length == 0 && Compilers.Length == 0) {
```
MikaelMayer, line 51: It could be possible that the assembly itself loads other assemblies, in which case there might not be Rewriter or Compiler directly in this assembly.
Please also check that there is no `PluginConfiguration` to match your error message.

Source/Dafny/Plugins/PluginConfiguration.cs
```diff
@@ -22,10 +24,18 @@ public abstract class PluginConfiguration {
...
-  /// Override this method to provide specific rewriters to Dafny
+  /// Override this method to provide rewriters
   /// </summary>
   /// <returns>a list of Rewriter that are going to be used in the resolution pipeline</returns>
   public virtual Rewriter[] GetRewriters(ErrorReporter errorReporter) {
     return Array.Empty<Rewriter>();
   }
+
+  /// <summary>
+  /// Override this method to provide specific compilers to Dafny
```
MikaelMayer, line 35:
```suggestion
  /// Override this method to provide compilers
```
To match the change you made on `GetRewriters`

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -734,12 +621,13 @@ private record TargetPaths(string Directory, string Filename) {
...
 
-      compiler.Coverage.WriteLegendFile();
+      compiler.WriteCoverageLegendFile();
 
       // blurt out the code to a file, if requested, or if other target-language files were specified on the command line.
-      var paths = GenerateTargetPaths(compiler, dafnyProgramName);
+      var paths = GenerateTargetPaths(dafnyProgramName);
       if (DafnyOptions.O.SpillTargetCode > 0 || otherFileNames.Count > 0 || (invokeCompiler && !compiler.SupportsInMemoryCompilation) ||
           (invokeCompiler && compiler.TextualTargetIsExecutable && !DafnyOptions.O.RunAfterCompile)) {
+        compiler.CleanSourceDirectory(paths.SourceDirectory);
```
MikaelMayer, line 630: Do we always want to do that? I'm personally relying on the generated files to debug any problem.

MikaelMayer (changes requested):
Great job of refactoring compilers to plugins !
A few comments to think about to enhance the code and readability, no major change requested.
I think one of the next steps will be, like @keyboardDrummer  suggested, to enable compilers to act as resolvers without emiting code so that DafnyLanguageServer can optionally use their error messages. What would the next PR looks like for that? I guess it would be too much to do in a single PR.

MikaelMayer:
> I don't understand the plugins test; why are we generating a library on the fly in

Because I wanted to ensure the loading of an external DLL works correctly, that's what this test is about. So, it tests two things, the use of plugins, and the use of external plugins.

Source/Dafny/Plugin.cs
cpitclaudel: Hmm, no, I think this code path is only reached if there is no plugin configuration in the first place?

Source/Dafny/Compilers/Compiler-go.cs
```diff
@@ -17,16 +17,23 @@
 using Bpl = Microsoft.Boogie;
 using static Microsoft.Dafny.ConcreteSyntaxTreeUtils;
 
-namespace Microsoft.Dafny {
-  public class GoCompiler : Compiler {
-    public GoCompiler(ErrorReporter reporter)
-    : base(reporter) {
+namespace Microsoft.Dafny.Compilers.Go {
```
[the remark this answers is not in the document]
cpitclaudel: Good point, will clean up

Source/Dafny/Compilers/Compiler-java.cs
cpitclaudel: No, I think not; in fact previously these strings were not used at all

Source/Dafny/Plugins/Compiler.cs
```diff
@@ -0,0 +1,105 @@
...
+  public abstract bool SupportsInMemoryCompilation { get; }
+
+  protected ErrorReporter? Reporter = null;
+  protected ReadOnlyCollection<string>? OtherFileNames = null;
+
+  // This method exists because compilers are constructed very early in the pipeline (to consult their
+  // `SupportedExtensions`, `TargetLanguage`, etc.).  C# doesn't support static fields in abstract classes, so we have
+  // to create an instance to access these parameters.  The alternative is to have a factory class, but we deemed the
+  // added complexity unnecessary.
+  public virtual void LateInitialize(ErrorReporter reporter, ReadOnlyCollection<string> otherFileNames) {
```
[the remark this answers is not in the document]
cpitclaudel: Renamed; I'll keep he comment because I think it still helps explain the slightly surprising design.

Source/Dafny/Plugin.cs
MikaelMayer: oh ok that makes sense, because it wouldn't even try to list those.

Source/DafnyDriver/DafnyDriver.cs
MikaelMayer: oh great, I thought it was cleaning at the end.

Source/Dafny/Plugins/Compiler.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Microsoft.Dafny.Plugins;
+
+/// <summary>
+/// A class that plugins should extend in order to provide an extra Compiler to the pipeline.
+///
+/// If the plugin defines no PluginConfiguration, then Dafny will instantiate every sub-class
+/// of Compiler from the plugin.
+/// </summary>
+public abstract class Compiler {
+  public abstract IReadOnlySet<string> SupportedExtensions { get; }
```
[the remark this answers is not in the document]
cpitclaudel: I added documentation for each method

Source/Dafny/Plugins/Compiler.cs
```diff
@@ -0,0 +1,105 @@
...
+
+  public abstract string TargetLanguage { get; }
+  public abstract string TargetExtension { get; }
+  public virtual string TargetId => TargetExtension;
```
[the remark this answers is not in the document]
cpitclaudel: Hopefully the docs make this clear now

MikaelMayer (dismissed):
Great job it everything looks consistent.
So now the compilers are part of the "default plugins", good job restoring them in the tests

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -9,12 +9,18 @@
...
+  public class PythonCompiler : SinglePassCompiler {
+    public override IReadOnlySet<string> SupportedExtensions => new HashSet<string> { ".py" };
 
     public override string TargetLanguage => "Python";
+    public override string TargetExtension => "py";
+
+    public override bool SupportsInMemoryCompilation => true;
+    public override bool TextualTargetIsExecutable => true;
+
+    public override IReadOnlySet<string> SupportedNativeTypes => new HashSet<string> { }; // FIXME
```
atomb, line 22: Is the Python compiler just not finished enough for this to be worth filling in yet?

Source/Dafny/Plugins/Compiler.cs
```diff
@@ -0,0 +1,142 @@
...
+  /// Supported file extensions for additional compilation units (e.g. <c>.cs</c> for C#).
+  /// </summary>
+  public abstract IReadOnlySet<string> SupportedExtensions { get; }
+
+  /// <summary>
+  /// Human-readable string describing the language targeted by this compiler.
+  /// </summary>
+  public abstract string TargetLanguage { get; }
+  /// <summary>
+  /// Extension given to generated code files (e.g. `cs` for C#)
```
atomb, line 29: Are backticks and `<c>`/`</c>` the same in C# comments?

Source/Dafny/Plugins/Compiler.cs
```diff
@@ -0,0 +1,142 @@
...
+  /// </summary>
+  /// <param name="sourceDirectory">Name of the directory to delete.</param>
+  public virtual void CleanSourceDirectory(string sourceDirectory) { }
+
+  public abstract void Compile(Program dafnyProgram, ConcreteSyntaxTree output);
+
+  /// <summary>
+  /// Emits a call to "mainMethod" as the program's entry point, if such an explicit call is
+  /// required in the target language.
+  /// </summary>
```
atomb, line 110: This (and the following two methods) could use `<c>` instead of `"`.

atomb (dismissed):
This looks very nice. I added a few comments about things that could potentially be slightly cleaned up, but none of them are critical.

Source/Dafny/Compilers/Compiler-python.cs
cpitclaudel: Correct

keyboardDrummer (dismissed):

atomb (approved):

PR #1900, opened 2022-03-12
fix: For attributes, add missing resolution and fix pointless parsing

RustanLeino:
* fix: Resolve attributes on `requires` and `ensures` clauses for functions (previously, these were never resolved)
 * fix: Resolve attributes on all specification components of iterators (previously, these were never resolved)
 * fix: Previously, attributes on `reads` clauses were parsed and then ignored. Now, they are not parsed at all.
 * fix: Previously, attributes on `requires` clauses of lambda expressions were parsed and then ignored. Now, they are not parsed at all.
 * fix: Previously, attributes on `decreases` clauses of functions were parsed and then ignored. Now, they are parsed and stored in the AST.
 * Add test cases that make sure attributes on `modify` statements and `module` declarations are resolved


cpitclaudel (commented):
Are there any call left to ResolveAttributes that are not of the form `ResolveAttributes(X, X.attributes)`?  If not then maybe it's worth changing the signature of that function to take one fewer parameter

RustanLeino:
> Are there any call left to ResolveAttributes that are not of the form `ResolveAttributes(X, X.attributes)`? If not then maybe it's worth changing the signature of that function to take one fewer parameter

@cpitclaudel Good call! This simplified all 39 call sites.

cpitclaudel (approved):

PR #1908, opened 2022-03-15
fix(compiler): Add parentheses around C# casts to fix #1714

### LINKED ISSUE #1714: Miscompilation due to incorrect parenthetisation in C# output for casts
Dafny miscompiles `(x as Y).z` into `x.z as Y`.  Here's a concrete example:

```dafny
trait Base {}
class Derived extends Base { var n: int; constructor() { n := 0; } }

method f(b: Base) {
  if (b is Derived) {
    print((b as Derived).n);
  }
}

method Main() {
  var d := new Derived();
  f(d);
}
```

```console
$ Dafny /compile:3 casts.dfy
Errors compiling program:
(1846,42): error CS1061: 'Base' does not contain a definition for 'n' and no accessible extension method 'n' accepting a first argument of type 'Base' could be found (are you missing a using directive or an assembly reference?)
```

This is the (broken) generated code:

```
    public static void f(Base b)
    {
      if (Dafny.Helpers.Let<Base, bool>(b, _is_0 => _is_0 is Derived)) {
        Dafny.Helpers.Print((Derived)(b).n);
      }
    }
```

`(Derived)(b).n` associates as `(Derived)((b).n)`, whereas Dafny expects it to associate as `((Derived)(b)).n`, hence the problem.

cpitclaudel:
Fixes #1714

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,7 @@
 - fix: Regain C# 7.3 compatibility of the compiled code. (https://github.com/dafny-lang/dafny/pull/1877)
 - fix: The error "assertion violation" is replaced by the better wording "assertion might not hold". This indicates better that the verifier is unable to prove the assertion. (https://github.com/dafny-lang/dafny/pull/1862)
 
+
 # 3.4.2
 
 - fix: No output when compiling to JavaScript on Windows (https://github.com/dafny-lang/dafny/pull/1824)
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 # Upcoming
 
+- fix: Miscompilation due to incorrect parenthesization in C# output for casts. (#1908)
+
 # 3.5.0
 
 - feat: `continue` statements. Like Dafny's `break` statements, these come in two forms: one that uses a label to name the continue target and one that specifies the continue target by nesting level. See section [19.2](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-break-continue) of the Reference Manual. (https://github.com/dafny-lang/dafny/pull/1839)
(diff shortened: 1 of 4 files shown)


Test/git-issues/git-issue-1714.dfy.expect
```diff
@@ -0,0 +1,5 @@
+
+Dafny program verifier finished with 2 verified, 0 errors
+Running...
+
+(b as Derived).n == 0
```
fabiomadge, line 5:
```suggestion
Dafny program verifier finished with 2 verified, 0 errors
(b as Derived).n == 0
```
This should fix the test.

RELEASE_NOTES.md
```diff
@@ -11,7 +13,6 @@
 - fix: Regain C# 7.3 compatibility of the compiled code. (https://github.com/dafny-lang/dafny/pull/1877)
 - fix: The error "assertion violation" is replaced by the better wording "assertion might not hold". This indicates better that the verifier is unable to prove the assertion. (https://github.com/dafny-lang/dafny/pull/1862)
 
-
```
fabiomadge, line 14: Let's keep this for consistency?

fabiomadge: Or, remove the others as well.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1479,7 +1479,7 @@ protected class ClassWriter : IClassWriter {
         if (Attributes.ContainsBool(cl.Attributes, "handle", ref isHandle) && isHandle) {
           return "0";
         } else {
-          return $"({TypeName(xType, wr, udt.tok)})null";
+          return $"(({TypeName(xType, wr, udt.tok)})null)";
```
fabiomadge, line 1482: Do you expect this will ever part of a bigger expression?

RELEASE_NOTES.md
```diff
@@ -11,7 +13,6 @@
 - fix: Regain C# 7.3 compatibility of the compiled code. (https://github.com/dafny-lang/dafny/pull/1877)
 - fix: The error "assertion violation" is replaced by the better wording "assertion might not hold". This indicates better that the verifier is unable to prove the assertion. (https://github.com/dafny-lang/dafny/pull/1862)
```
cpitclaudel, line 14:
```suggestion
- fix: The error "assertion violation" is replaced by the better wording "assertion might not hold". This indicates better that the verifier is unable to prove the assertion. (https://github.com/dafny-lang/dafny/pull/1862)

```

Source/Dafny/Compilers/Compiler-Csharp.cs
cpitclaudel: I changed it for consistency with `return $"(({TypeName(xType, wr, udt.tok)})null)";` above

fabiomadge (approved):

PR #1911, opened 2022-03-16
fix: Populate TestResult.ResourceCount for assertion batches

robin-aws:
Also add support for CHECK-NOT[-L] OutputCheck directives, and use them to add a regression test.

Also address the fact that CHECK-NEXT[-L] directives were never implemented, so their use was silently doing nothing. :(


atomb (dismissed):
LGTM. Those new directives are nice, and cover things I've thought it might be good to test in the past.

atomb (approved):

Issue #1754, opened 2022-01-22
Dafny feature to diagnose individual cost of individual Dafny assertions


robin-aws:
It's possible to use the new Boogie `/vcsSplitOnEveryAssert` feature to approximate the verification cost of individual assertions, which can be very helpful in identifying why verification of a single method is timing out, or in getting the cost of a method down to improve its stability. However, identifying the contents of each split requires scrolling through a ton of `/traceverify` output, and then attempting to map a Boogie assert statement back to the corresponding element of the Dafny program.

Depends on #1735 and #1657: we probably want a switch that turns on `/vcsSplitOnEveryAssert` and then labels each split with the Dafny description of the assertion that induced it.

atomb:
PR #1766 provides part of this functionality more conveniently than `/trace`. The key remaining step is to describe each assertion in terms of Dafny constructs instead of Boogie constructs.

atomb:
PR #1947 now provides a human-readable description of each assertion along with time and solver resource count used to prove it, enabled with the combination of `/verificationLogger:text` and `/vcsSplitOnEveryAssert`.

keyboardDrummer:
Shouldn't this be integrated with the IDE as well?

atomb:
Yeah, I think it should (and will be once #1942 and related PRs are merged). I considered having one way to achieve the goal to be sufficient for closing this, but one could argue that it's not fully done until those are merged.

Issue #1735, opened 2022-01-20
Investigate providing understandable identifiers/descriptions for individual Dafny assertions


robin-aws:
Currently these are only available as verification error strings instead. E.g. “possible violation of function precondition”. This is necessary to describe what a given VC contains and hence understand how to fix it, especially when using /vcsSplitOnEveryAssert.

It also includes being better about the distinction between “assertion definitely not always true” and “assertion not necessarily always true, I’m not sure” - the error messages are currently inconsistent, but at the moment the actual answer is always the latter.

atomb:
I believe that #1915 fixed this. There is certainly room to improve individual assertion descriptions and error messages, but the mechanism is in place to describe what each assertion is attempting to establish.

Issue #1734, opened 2022-01-20
Improve display name of "test results" when using `/verificationLogger:...`


robin-aws:
Splitting this off from #1597

What is currently something like Impl$$MyModule.__default.MyMethod should be something more like MyModule.MyMethod (implementation)

This MIGHT not be worth doing if we want to model the metadata of test results separately somehow, and use more of a unique identifier in the test results themselves, so this depends on some decisions as part of #1657

atomb:
I've closed #1657, and plan to use this issue to track the naming facet of the design.

PR #1915, opened 2022-03-16
Add structured proof obligation descriptions

atomb:
This PR adapts Dafny to use the structured form of proof obligation description introduced in https://github.com/boogie-org/boogie/pull/505.

In itself, this PR makes very few functional changes. It should pass existing tests with several caveats:
* A few assertion failure messages have changed slightly to allow shared descriptions in several places. Expected test results have been updated to reflect this, and those updates are perhaps the most important thing to review. They include:
    * Sometimes multiple assertions from a single location are listed in a different order than before.
    * Some messages related to decreases clauses have changed slightly.
    * Some checks for definite assignment have a slightly different message.
    * Some failures messages are now in terms of what might not be true rather than what must be true.

It's a somewhat long PR, but the changes are very mechanical and include very little logic.

Makes progress toward #1734, #1735, and #1754.


Test/allocated1/dafny0/Basics.dfy.expect
```diff
@@ -9,14 +9,14 @@ Basics.dfy(148,15): Error: assertion might not hold
 Basics.dfy(167,9): Error: when left-hand sides a[b[18]] and a[4] refer to the same location, they must be assigned the same value
 Basics.dfy(183,9): Error: when left-hand sides x.f and y.f refer to the same location, they must be assigned the same value
 Basics.dfy(195,18): Error: assertion might not hold
-Basics.dfy(197,9): Error: assignment may update an object not in the enclosing context's modifies clause
 Basics.dfy(197,9): Error: target object may be null
+Basics.dfy(197,9): Error: assignment may update an object not in the enclosing context's modifies clause
```
MikaelMayer, line 13: Do you know why these positions changed?

Test/allocated1/dafny0/ControlStructures.dfy.expect
```diff
@@ -8,8 +8,8 @@ ControlStructures.dfy(218,17): Error: assertion might not hold
 ControlStructures.dfy(235,20): Error: assertion might not hold
 ControlStructures.dfy(238,29): Error: assertion might not hold
 ControlStructures.dfy(241,16): Error: assertion might not hold
-ControlStructures.dfy(350,2): Error: cannot prove termination; try supplying a decreases clause for the loop
-ControlStructures.dfy(381,2): Error: cannot prove termination; try supplying a decreases clause for the loop
+ControlStructures.dfy(350,2): Error: cannot prove termination; try supplying a decreases clause
```
MikaelMayer, line 11: Can you put back the message "for the loop"? I'm not sure but I feel it was better with it.

Test/allocated1/dafny0/Maps.dfy.expect
```diff
@@ -31,7 +31,7 @@ Maps.dfy(447,17): Error: assertion might not hold
 Maps.dfy(448,13): Error: assertion might not hold
 Maps.dfy(520,18): Error: assertion might not hold
 Maps.dfy(525,20): Error: assertion might not hold
-Maps.dfy(562,2): Error: decreases expression might not decrease
-Maps.dfy(576,2): Error: decreases expression might not decrease
+Maps.dfy(562,2): Error: decreases clause might not decrease
```
MikaelMayer, line 34: Here I have a doubt, but I think this message is better, since it's not just one expression, it can be this special kind of tuple.

Test/allocated1/dafny0/Twostate-Functions.dfy.expect
```diff
@@ -5,18 +5,18 @@ Twostate-Functions.dfy(70,17): Error: assertion might not hold
 Twostate-Functions.dfy(57,14): Related location
 Twostate-Functions.dfy(72,15): Error: assertion might not hold
 Twostate-Functions.dfy(57,14): Related location
-Twostate-Functions.dfy(96,24): Error: argument ('u') must be allocated in the two-state function's previous state
-Twostate-Functions.dfy(101,40): Error: argument 1 ('x') must be allocated in the two-state function's previous state
-Twostate-Functions.dfy(107,40): Error: argument 1 ('x') must be allocated in the two-state function's previous state
+Twostate-Functions.dfy(96,24): Error: argument ('u') may not be allocated in the two-state function's previous state
```
MikaelMayer, line 8: "may not" implicitly means "does not have the authorization to", which is the opposite of the previous error message. Can you explain this or fix it? You might consider `might not`, which we use in other places.

MikaelMayer (changes requested):
Fantastic ! I took the time to carefully review each assertion and added comments for the most relevant ones in #1930.
Just a few things we can optimize before merging this PR.

Source/Dafny/Verifier/DafnyAssertionDescription.cs
```diff
@@ -0,0 +1,890 @@
+using JetBrains.Annotations;
+using Microsoft.Boogie;
+
+namespace Microsoft.Dafny;
```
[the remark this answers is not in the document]
atomb: Good idea.

Test/allocated1/dafny0/Twostate-Functions.dfy.expect
atomb: Oh, yeah, "might not" is much better. :)

Source/Dafny/Verifier/DafnyAssertionDescription.cs
```diff
@@ -0,0 +1,890 @@
...
+  public override string SuccessDescription =>
+    "postcondition of forall statement always holds";
+
+  public override string FailureDescription =>
+    "possible violation of postcondition of forall statement";
+
+  public override string ShortDescription => "forall ensures";
+}
+
+public class DafnyYieldEnsuresDescription : DafnyAssertionDescription {
```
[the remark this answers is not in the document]
atomb: Huh. I guess Dafny never generates assertions to check `yield requires` annotations?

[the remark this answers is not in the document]
MikaelMayer: Nope, they are just translated to regular `requires`. We should explain in the documentation how this happens.

Source/Dafny/Verifier/Translator.TrStatement.cs
```diff
@@ -1381,12 +1383,12 @@ public partial class Translator {
...
         if (!splitHappened) {
           var wInv = Bpl.Expr.Imp(w, etran.TrExpr(loopInv.E));
-          invariants.Add(Assert(loopInv.E.tok, wInv, errorMessage ?? "loop invariant violation"));
+          invariants.Add(Assert(loopInv.E.tok, wInv, new DafnyLoopInvariantDescription(errorMessage)));
         } else {
           foreach (var split in ss) {
             var wInv = Bpl.Expr.Binary(split.E.tok, BinaryOperator.Opcode.Imp, w, split.E);
             if (split.IsChecked) {
-              invariants.Add(Assert(split.E.tok, wInv, errorMessage ?? "loop invariant violation"));  // TODO: it would be fine to have this use {:subsumption 0}
+              invariants.Add(Assert(split.E.tok, wInv, new DafnyLoopInvariantDescription(errorMessage)));  // TODO: it would be fine to have this use {:subsumption 0}
```
cpitclaudel, line 1391: Will it make a difference that we pass a non-null value in all cases now?

Source/Dafny/Verifier/Translator.TrStatement.cs
```diff
@@ -1495,13 +1497,7 @@ public partial class Translator {
...
             AddComment(loopBodyBuilder, s, "loop termination check");
             Bpl.Expr decrCheck = DecreasesCheck(toks, types, types, decrs, oldBfs, loopBodyBuilder, " at end of loop iteration", false, false);
-            string msg;
-            if (s.InferredDecreases) {
-              msg = "cannot prove termination; try supplying a decreases clause for the loop";
-            } else {
-              msg = "decreases expression might not decrease";
-            }
-            loopBodyBuilder.Add(Assert(s.Tok, decrCheck, msg));
+            loopBodyBuilder.Add(Assert(s.Tok, decrCheck, new DafnyTerminationDescription(s.InferredDecreases)));
```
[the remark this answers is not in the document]
cpitclaudel: But shouldn't these two assertions be two different subclasses?

Source/Dafny/Verifier/Translator.cs
```diff
@@ -6552,7 +6571,7 @@ private class WFOptions {
             var toBound = BaseTypes.BigNum.FromBigInt(BigInteger.One << toWidth); // 1 << toWidth
             var bound = BplBvLiteralExpr(tok, toBound, expr.Type.AsBitVectorType);
             var boundsCheck = FunctionCall(expr.tok, "lt_bv" + fromWidth, Bpl.Type.Bool, o, bound);
-            builder.Add(Assert(tok, boundsCheck, $"{errorMsgPrefix}bit-vector value to be converted might not fit in {toType}"));
+            builder.Add(Assert(tok, boundsCheck, new DafnyConversionFitDescription("bit-vector value", toType.ToString(), errorMsgPrefix)));
```
cpitclaudel, line 6574: Can we delay the conversion to a string?

Source/Dafny/Verifier/Translator.TrStatement.cs
```diff
@@ -1495,13 +1497,7 @@ public partial class Translator {
           if (includeTerminationCheck) {
             AddComment(loopBodyBuilder, s, "loop termination check");
             Bpl.Expr decrCheck = DecreasesCheck(toks, types, types, decrs, oldBfs, loopBodyBuilder, " at end of loop iteration", false, false);
-            string msg;
-            if (s.InferredDecreases) {
-              msg = "cannot prove termination; try supplying a decreases clause for the loop";
-            } else {
-              msg = "decreases expression might not decrease";
```
cpitclaudel, line 1502: This used to say "expression", but the new code says "clause"

Test/allocated1/dafny0/Basics.dfy.expect
```diff
@@ -9,14 +9,14 @@ Basics.dfy(148,15): Error: assertion might not hold
...
 Basics.dfy(202,11): Error: left-hand sides m.x and m.x may refer to the same location
 Basics.dfy(213,14): Error: assertion might not hold
 Basics.dfy(273,9): Error: when left-hand sides d.x and c.x refer to the same location, they must be assigned the same value
 Basics.dfy(463,11): Error: assertion might not hold
 Basics.dfy(474,18): Error: assertion might not hold
 Basics.dfy(476,11): Error: assertion might not hold
-Basics.dfy(13,13): Error: variable 'g', which is subject to definite-assignment rules, might be used before it has been assigned
+Basics.dfy(13,13): Error: variable 'g', which is subject to definite-assignment rules, might not have been defined before it is used
```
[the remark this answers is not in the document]
cpitclaudel: For me they are different: 

```
var x; // Declaration
trait T { function f() } : int // Declaration
var x := 1 // Declaration with definition
class C extends T { function f() : int { 0 } } // Declaration
x := 3; // Assignment
```

So assignment is the right word to use here, because it's about mutating a variable.
I also think `initialized` would be a viable alternative.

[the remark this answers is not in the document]
cpitclaudel: Or cut out a lot fo the text: `variable 'g', which is subject to definite-assignment rules, might be uninitialized here`

cpitclaudel (dismissed):
Looks like a nice refactoring.  While we're at it, though, I wonder if it might be worth to go a touch further.

All of these descriptions are attached to assertions, right?  Could we maybe move the code that builds the assertion into the description object as well?  We don't necessarily have to display these back to users just yet, of course.

It doesn't have to happen with this PR, of course.

Source/Dafny/Verifier/Translator.TrStatement.cs
atomb: The check for null has moved inside the `DafnyLoopInvariantDescription` class, and the non-null object being created here is stored in a different property than the string was before. It's a little fiddly, but I'm hoping the logic for null vs non-null strings goes away entirely before long.

Source/Dafny/Verifier/Translator.TrStatement.cs
atomb: Yeah, true. Which do you prefer? The previous code sometimes used one and sometimes the other.

Source/Dafny/Verifier/Translator.cs
atomb: Yeah, we could. I'll do that. I think there are also other cases where it would make more sense to pass a structured object instead of a string into these descriptions, and I'd love to move toward doing more of that.

atomb:
> Looks like a nice refactoring. While we're at it, though, I wonder if it might be worth to go a touch further.
> 
> All of these descriptions are attached to assertions, right? Could we maybe move the code that builds the assertion into the description object as well? We don't necessarily have to display these back to users just yet, of course.
> 
> It doesn't have to happen with this PR, of course.

Yeah, I think something like that could be nice. Probably a future PR, as you say, though.

Test/dafny0/Modules1.dfy.expect
```diff
@@ -7,8 +7,8 @@ Modules1.dfy(213,22): Error: assertion might not hold
 Modules1.dfy(219,16): Error: assertion might not hold
 Modules1.dfy(220,16): Error: assertion might not hold
 Modules1.dfy(221,16): Error: assertion might not hold
-Modules1.dfy(56,8): Error: decreases expression must be bounded below by 0
+Modules1.dfy(56,8): Error: decreases clause must be bounded below by 0
```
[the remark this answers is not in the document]
atomb: That seems about right to me. I wonder whether @RustanLeino has thoughts.

Test/allocated1/dafny0/Basics.dfy.expect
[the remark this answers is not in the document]
atomb: I like @cpitclaudel's last suggestion. What do you think, @MikaelMayer? @RustanLeino?

Source/Dafny/Verifier/Translator.cs
```diff
@@ -8260,12 +8269,12 @@ public ForceCheckToken(IToken tok)
...
         var args = new List<object>();
         args.Add(Bpl.Expr.Literal(0));
-        Bpl.AssertCmd cmd = TrAssertCmd(tok, condition, new Bpl.QKeyValue(tok, "subsumption", args, kv));
-        cmd.ErrorData = "Error: " + errorMessage;
+        Bpl.AssertCmd cmd = TrAssertCmdDesc(tok, condition, desc, new Bpl.QKeyValue(tok, "subsumption", args, kv));
         return cmd;
       }
     }
 
+    // TODO: update to include structured description
```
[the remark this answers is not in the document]
atomb: Ah, no, it's just a further progression of this refactoring that remains to be done. I'm thinking of doing it in a future PR, since this one is functional, to break things down into more bite-sized pieces.

Source/Dafny/Verifier/Translator.TrStatement.cs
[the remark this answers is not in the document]
atomb: Arguably? They're both describing failure to prove termination, just for different reasons.

Source/Dafny/Verifier/Translator.ClassMembers.cs
```diff
@@ -1382,9 +1383,12 @@ public partial class Translator {
         foreach (var formal in m.Ins) {
           if (formal.IsOld) {
             var dafnyFormalIdExpr = new IdentifierExpr(formal.tok, formal);
-            req.Add(Requires(formal.tok, false, MkIsAlloc(etran.TrExpr(dafnyFormalIdExpr), formal.Type, prevHeap),
-              string.Format("parameter{0} ('{1}') must be allocated in the two-state lemma's previous state",
-                m.Ins.Count == 1 ? "" : " " + index, formal.Name), null));
+            var pIdx = m.Ins.Count == 1 ? "" : " " + index;
```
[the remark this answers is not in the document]
MikaelMayer: Let me know if this change cause problem

Source/Dafny/Verifier/Translator.ClassMembers.cs
```diff
@@ -660,7 +661,7 @@ public partial class Translator {
           if (formal.IsOld) {
             Boogie.Expr wh = GetWhereClause(e.tok, etran.TrExpr(e), e.Type, etran.Old, ISALLOC, true);
             if (wh != null) {
-              var desc = new DafnyAllocatedDescription("default value", "in the two-state lemma's previous state");
+              var desc = new IsAllocated("default value", "in the two-state lemma's previous state");
```
MikaelMayer, line 664: Could you perhaps write `PODesc.IsAllocated` like in the other files?

Source/Dafny/Verifier/Translator.cs
```diff
@@ -8781,7 +8782,7 @@ public ForceCheckToken(IToken tok)
     }
 
     /// <summary>
-    /// Clone Dafny variable "v" into a new Dafny local variable "l".
+    /// Clone Dafny variable "v" into a new PODesc. local variable "l".
```
MikaelMayer, line 8785:
```suggestion
    /// Clone Dafny variable "v" into a new Dafny local variable "l".
```

MikaelMayer (changes requested):
Looks much better !

Source/DafnyLanguageServer/Workspace/ThreadTaskScheduler.cs
```diff
@@ -0,0 +1,42 @@
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Dafny.LanguageServer.Workspace;
```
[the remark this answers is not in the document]
atomb: I had merged with another PR that hadn't yet been merged to `master` to allow builds to succeed. This doesn't show up in the diff anymore now that that PR has been merged.

Source/Dafny/Verifier/Translator.ClassMembers.cs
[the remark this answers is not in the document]
atomb: I'm undecided about this. I tend to use the phrase "parameter 1" in practice more than the phrase "parameter at index 1".

Test/allocated1/dafny0/Basics.dfy.expect
atomb: My guess, though I haven't confirmed this, is that these are sorted by position at some point, but that process wouldn't be stable with respect to multiple errors on the same line. By changing the structure of the objects in that collection, their order could change, as well.

atomb:
Since I first requested a review, I've added one bit of functionality: a command-line flag `/printAllVerificationResults`. This prints details about the status, meaning (using these new descriptions), time, and solver resource count for every assertion batch. This allows for a test of a couple of `SuccessDescription` properties, and makes the new descriptions user-visible. I'm not entirely sure of the format of the output, though. I'd love comments on that.

A also thought it might be worth consolidating the open questions I think we have about message contents, to make it easier to reference what work is left. Here's my take on the unanswered comments from above:
* Should we use @cpitclaudel's suggestion of `variable 'g', which is subject to definite-assignment rules, might be uninitialized here`? I like it, myself.
* Should we say `parameter at index n` or `parameter n` when talking about two-state allocation?
* Should we use `clause` or `expression` to talk about what a function or loop must decrease to be proved terminating? Or both, in different contexts?
* Should the `Termination` description class be split in two?
* Should we use @MikaelMayer's suggestion of `every integer element of the decreases clause must be bounded below by 0` instead of `decreases clause must be bounded below by 0`?

Source/Dafny/Verifier/Translator.ClassMembers.cs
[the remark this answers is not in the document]
MikaelMayer: I would say that "parameter 2" is the equivalent of "parameter at index 1", not "parameter 1", simply because indexing in everyday life consists of assigning the number of the elements already seen to the new element, that new element included.
i would wholeheartedly prefer "parameter 2" instead of "parameter at index 1", but for consistency with other error messages, until we decide to change "at index" to a more human-readable indexing numbering system, I would really prefer to warn users that "parameter 1" really means the second parameter.

Test/allocated1/dafny0/Basics.dfy.expect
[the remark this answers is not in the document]
MikaelMayer: Indeed I prefer uninitialized !

MikaelMayer:
> * Should we use `clause` or `expression` to talk about what a function or loop must decrease to be proved terminating? Or both, in different contexts?

I think we could say "expression" when there is only one element in the clause, and "clause" otherwise.

> * Should the `Termination` description class be split in two?

I don't think it's necessary. A class is made to be reused as much as possible if it's not too costly.

> * Should we use @MikaelMayer's suggestion of `every integer element of the decreases clause must be bounded below by 0` instead of `decreases clause must be bounded below by 0`?

I am biased on it, but I would even say be more precise and say which integer of the decreases clause is not bounded by zero.

Source/Dafny/DafnyOptions.cs
```diff
@@ -538,6 +539,10 @@ public enum IncludesModes {
           ExtractCounterexample = true;
           return true;
 
+        case "printAllVerificationResults":
```
atomb, line 542: I'm leaning toward making this a part of `/verificationLogger`, perhaps `/verificationLogger:text` or something like that.

robin-aws (changes requested):
Doing my annoying thing of blocking until I get a proper review in as I'm really interested in this one. :)

Source/Dafny/DafnyConsolePrinter.cs
```diff
@@ -57,4 +58,32 @@ class DafnyConsolePrinter : ConsolePrinter {
       ReportBplError(nt.Inner, "Related location", false, tw);
     }
   }
+
+  public override void ReportEndVerifyImplementation(Implementation implementation, Boogie.VerificationResult result) {
+    verificationResults.Add((implementation, result));
+  }
```
robin-aws, line 64: We can use this hook to raise test logger events now instead of reading the `/xml` file now, right?

(in a separate PR of course :)

Source/Dafny/DafnyOptions.cs
```diff
@@ -1124,6 +1129,14 @@ not sound
         /verificationLogger:trx;LogFileName=<...>.
     The exact mapping of verification concepts to the TRX and CSV formats is
     experimental and subject to change!
+/printAllVerificationResults
+    Prints a summary of all verification results after verification has
+    completed. For every assertion batch of every method, prints the
+    list of assertions that occur within that batch, the status of the
+    batch, and the resources (in terms of time and solver resources)
+    needed analyze the batch. For fine-grained analysis of the result
+    and cost of verifying every assertion, combine this with the Boogie
```
robin-aws, line 1138: Nit (and probably moot given my other suggestions), but we should be careful about implying that we're measuring the true cost of each assertion that accurately. Probably want to sneak "approximate" into any description like this, or go into more detail about how batching (and hence un-batching) can affect the cost unpredictably.

Source/Dafny/DafnyOptions.cs
robin-aws: Yup, I'd suggest attaching the list of assertions for each batch to the corresponding `TestCase` using another custom property, just as we did for the `TestResult.ResourceCount`. If you squint you could imagine this concept applying to other kinds of test results too (i.e. the location and message of each test assertion such as `Assert.AreEqual` in an xUnit test method).

Then we can have the `csv` logger emit this data as a secondary file, or we can add a `/verificationLogger:console` option that will print out something like what you've got here. Or both!

robin-aws (commented):
Just commenting on the `/printAllVerificationResults` functionality for now: I made some suggestions for reshaping it, but I'd also suggest adding that as a separate PR (or two) instead of tacking it on here.

Source/Dafny/DafnyConsolePrinter.cs
atomb: Yep, I've been starting to think that.

atomb:
I think I agree with your comments on making the logger a separate PR, @robin-aws . I was torn, originally, and even have a draft branch to do it separately. I think I'll pull it out of here and move it to a PR created there.

Test/allocated1/dafny0/Basics.dfy.expect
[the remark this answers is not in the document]
atomb: I just pushed a change adopting @cpitclaudel's suggestion.

Test/allocated1/dafny0/Reads.dfy.expect
```diff
@@ -4,8 +4,8 @@ Reads.dfy(18,45): Error: insufficient reads clause to read field
 Reads.dfy(28,69): Error: insufficient reads clause to read field
 Reads.dfy(37,45): Error: insufficient reads clause to read field
 Reads.dfy(51,45): Error: insufficient reads clause to read field
-Reads.dfy(117,35): Error: insufficient reads clause to invoke function
 Reads.dfy(117,35): Error: possible violation of function precondition
+Reads.dfy(117,35): Error: insufficient reads clause to invoke function
```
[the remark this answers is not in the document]
atomb: Agreed.

atomb:
When it comes to "clause" versus "expression" for termination metrics, I decided to punt and just leave the messages close to what they were before. We can perhaps improve the messages in a future PR.

RustanLeino (approved):

Source/Dafny/Verifier/ProofObligationDescription.cs
```diff
@@ -244,7 +244,7 @@ public class PreconditionSatisfied : ProofObligationDescription {
       : $"error is impossible: {customErrMsg}";
 
   public override string FailureDescription =>
-    customErrMsg ?? "possible violation of function precondition";
+    customErrMsg ?? "function precondition might not hold";
```
MikaelMayer, line 247: Wonderful ! That has resulted in a lot of test fixes (good job!) but this message is much better.

Source/Dafny/Verifier/ProofObligationDescription.cs
```diff
@@ -416,31 +416,34 @@ public class Terminates : ProofObligationDescription {
...
 
   public override string FailureDescription =>
-    $"decreases clause{component} must be bounded below by {zeroStr}{suffix}";
+    $"decreases {component} must be bounded below by {zeroStr}{suffix}";
 
-  public override string ShortDescription => "bounded decreases clause";
+  public override string ShortDescription => "bounded decreases expression";
 
-  private string component => N == 1 ? "" : $" (component {k})";
+  private string component => N == 1 ? "expression" : $"clause (component at index {k})";
```
MikaelMayer, line 446:
```suggestion
  private string component => N == 1 ? "expression" : $"expression at index {k}";
```
As Rustan suggested, since we know the components index, we might as well use the shorter version of this error message.
I added another suggestion so that if you accept both, it should also fix the test.

Test/allocated1/dafny0/Twostate-Functions.dfy.expect
```diff
@@ -6,17 +6,17 @@ Twostate-Functions.dfy(57,14): Related location
 Twostate-Functions.dfy(72,15): Error: assertion might not hold
 Twostate-Functions.dfy(57,14): Related location
 Twostate-Functions.dfy(96,24): Error: argument ('u') might not be allocated in the two-state function's previous state
-Twostate-Functions.dfy(101,40): Error: argument 1 ('x') might not be allocated in the two-state function's previous state
-Twostate-Functions.dfy(107,40): Error: argument 1 ('x') might not be allocated in the two-state function's previous state
+Twostate-Functions.dfy(101,40): Error: argument at index 1 ('x') might not be allocated in the two-state function's previous state
```
MikaelMayer, line 9: Yay thanks in the name of all old-fashioned thinkers like me who still believe that argument 1 is the first argument and not the second :-)

Test/dafny3/Inc.dfy.expect
```diff
@@ -18,7 +18,7 @@ Inc.dfy(238,0): Error: A postcondition might not hold on this return path.
 Inc.dfy(237,12): Related location: This is the postcondition that might not hold.
 Inc.dfy(247,2): Error: A postcondition might not hold on this return path.
 Inc.dfy(244,12): Related location: This is the postcondition that might not hold.
-Inc.dfy(318,17): Error: decreases clause (component 1) must be bounded below by 0
+Inc.dfy(318,17): Error: decreases clause (component at index 1) must be bounded below by 0
```
MikaelMayer, line 21:
```suggestion
Inc.dfy(318,17): Error: decreases expression at index 1 must be bounded below by 0
```

MikaelMayer (changes requested):
One minor change I believe, and perhaps optionally consider Rustan's comment about the fact that the message for decreases clause might be misleading. Otherwise, I'm ready to approve it !

Source/Dafny/Verifier/ProofObligationDescription.cs
atomb: Ah, yes, I like this better, too.

MikaelMayer (approved):
Great job! Can't wait to use these!

Issue #1918, opened 2022-03-18
Crash of the language server


MikaelMayer:
The following code crashes the language server

```
method doFocus(x: bool) returns (y: int) {
  while * {
    assert y >= 0;
  }
}
```

The trouble is that the while's `Guard` is null but is still being visited. I'm working on a fix in my branch.

cpitclaudel:
Is there documentation somewhere as to why the LSP server has its own visitors?

MikaelMayer:
I didn't find any. It shouldn't be the case. We should have only one visitor.

cpitclaudel:
https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Language/SyntaxTreeVisitor.cs

MikaelMayer:
Oh so it's because of the auto-generated nodes.

MikaelMayer:
I mean, the two visitors exist because the language server node does not traverse auto-generated nodes.
You did find the problem of this issue, and I'm fixing it in my branch, which is going to take a while to merge though.

cpitclaudel:
Oh, I get it, thanks!

Issue #1913, opened 2022-03-16
Language Server crashing under heavy load


MikaelMayer:
I created the following test that crashes the language server without recovery.
I basically verifies an assertion that would take an almost infinite amount of time to verify, add 100 spaces in front of  it (so that every edit triggers verification), at increasing time intervals, and at the end replaces the impossible assertion with "true".
I expected to see it verify quickly because the last edit makes it trivial.
Instead, it never finishes.

```dafny
public async Task CancelingVerificationStopsTheProver() {
      var source = @"
function method {:unroll 100} Ack(m: nat, n: nat): nat
  decreases m, n
{
  if m == 0 then
    n + 1
  else if n == 0 then
    Ack(m - 1, 1)
  else
    Ack(m - 1, Ack(m, n - 1))
}

method test() {
  assert Ack(5, 5) == 0;
}".TrimStart();
      var documentItem = CreateTestDocument(source);
      var version = documentItem.Version;
      DidChangeTextDocumentParams ChangeRequest(int line, int character, int deleted, string added) {
        version += 1;
        return new DidChangeTextDocumentParams {
          TextDocument = new OptionalVersionedTextDocumentIdentifier {
            Uri = documentItem.Uri,
            Version = version
          },
          ContentChanges = new[] {new TextDocumentContentChangeEvent {
            Range = new Range((line - 1, character - 1), (line - 1, character - 1 + deleted)),
            Text = added
          }}
        };
      }

      client.OpenDocument(documentItem);
      await Task.Delay(1_000);
      var numberOfSpacesToAdd = 100;

      // Just add hundred spaces in 5 seconds
      for (var i = 1; i <= numberOfSpacesToAdd; i++) {
        client.DidChangeTextDocument(ChangeRequest(13, 10, 0, " "));
        await Task.Delay(i);
      }
      // This cancels the previous request.
      client.DidChangeTextDocument(ChangeRequest(13, 10, numberOfSpacesToAdd + 14, "true"));
      await client.WaitForNotificationCompletionAsync(documentItem.Uri, CancellationToken);
      var document = await Documents.GetDocumentAsync(documentItem.Uri);
      Assert.IsNotNull(document);
      Assert.IsTrue(!document.Errors.HasErrors);
    }
```

MikaelMayer:
I don't think this is a problem anymore

PR #1923, opened 2022-03-21
feat: Verification diagnostics

MikaelMayer:
Real-time verification diagnostics for the language server.
Do not merge yet.


Source/Dafny/Dafny.atg
```diff
@@ -1431,6 +1435,7 @@ SynonymTypeDecl<DeclModifierData dmod, ModuleDefinition module, out TopLevelDecl
        }
        td = new OpaqueTypeDecl(id, id.val, module, characteristics, typeArgs, members, attrs, isRefining);
      }
+     td.BodyEndTok = t;
```
MikaelMayer, line 1438: Same here, the subset type or newtype needs a witness check and I need to know the end of it, until we have proper ranges in nodes.

Source/DafnyLanguageServer.Test/DafnyLanguageServer.Test.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="DiffPlex" Version="1.7.0" />
```
MikaelMayer, line 10: I use a developer-friendly diff engine to test the gutter highlighting

Source/DafnyLanguageServer.Test/Util/AssertWithDiff.cs
```diff
@@ -0,0 +1,30 @@
...
+      var prefix = line.Type switch {
+        ChangeType.Inserted => '+',
+        ChangeType.Deleted => '-',
+        _ => ' '
+      };
+      message.Append(prefix);
+      message.AppendLine(line.Text);
+    }
+
+    Assert.AreEqual(expected, actual, message.ToString());
```
MikaelMayer, line 28: This file is the same as in XUnitExtensions's AssertWithDiff excepted for this line which uses the framework that the DafnyLanguageServer relies on. I can't factor them both in a single file because we don't have yet a project on which all test projects depend, and I don't think this PR should take care of that.

Source/DafnyLanguageServer.Test/Various/CompilationStatusNotificationTest.cs
```diff
@@ -96,23 +99,20 @@ method Abs(x: int) returns (y: int)
...
-      Assert.AreEqual(CompilationStatus.VerificationStarted, inprogress.Status);
-      Assert.AreEqual("Abs", inprogress.Message);
-      var completed = await notificationReceiver.AwaitNextNotificationAsync(CancellationToken);
-      Assert.AreEqual(documentItem.Uri, completed.Uri);
-      Assert.AreEqual(documentItem.Version, completed.Version);
-      Assert.AreEqual(CompilationStatus.VerificationSucceeded, completed.Status);
+      await AssertProgress(documentItem, CompilationStatus.CompilationSucceeded);
+      await AssertProgress(documentItem, CompilationStatus.VerificationStarted);
+      await AssertProgress(documentItem, CompilationStatus.VerificationStarted, "0/1 Abs");
+      await AssertProgress(documentItem, CompilationStatus.VerificationStarted, "1/1 (Abs finished)");
```
MikaelMayer, line 105: Note the refactoring to make it easier to understand what the test does. Furthermore, now there are more messages because of real-time indications on how many methods need to be verified, the total number of methods, and what was the last verified method.

Source/DafnyLanguageServer.Test/Various/StressVerification.cs
```diff
@@ -0,0 +1,71 @@
...
+using OmniSharp.Extensions.LanguageServer.Protocol.Client;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using OmniSharp.Extensions.LanguageServer.Protocol.Document;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
+  [TestClass]
+  public class StressVerificationTest : DafnyLanguageServerTestBase {
```
MikaelMayer, line 12: This stress test is still failing (looping forever). I am considering removing it from the test suite for now or deactivate it until we can properly fix it. I already filed an issue on it (#1913)

Source/DafnyLanguageServer/DafnyRequestNames.cs
```diff
@@ -3,5 +3,6 @@ public static class DafnyRequestNames {
     public const string CounterExample = "dafny/counterExample";
     public const string CompilationStatus = "dafny/compilation/status";
     public const string GhostDiagnostics = "dafny/ghost/diagnostics";
+    public const string VerificationDiagnostics = "dafny/verification/diagnostics";
```
MikaelMayer, line 6: This is Public API so, it should be our first and only change to decide what it looks like. Please comment on that.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +30,145 @@ public class DafnyHoverHandler : HoverHandlerBase {
       };
     }
 
-    public async override Task<Hover?> Handle(HoverParams request, CancellationToken cancellationToken) {
+    public override async Task<Hover?> Handle(HoverParams request, CancellationToken cancellationToken) {
```
MikaelMayer, line 33: I modified the hover handler so that it can prepend assertion diagnostics if it finds any.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +30,145 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
         return null;
       }
-      return CreateHover(symbol, cancellationToken);
+
+      var hoverContent = diagnosticHoverContent == null ? "" : diagnosticHoverContent;
+      hoverContent += symbolHoverContent == null ? (hoverContent != "" ? "  \n" : "") : symbolHoverContent;
+      return CreateMarkdownHover(hoverContent);
+    }
+
+    private string? GetDiagnosticsHover(DafnyDocument document, Position position) {
```
MikaelMayer, line 55: If you review this PR linearly, you might not yet be familiar of the new NodeDiagnostic data structure that is both being sent to the client and used in the server to report verification diagnostics on hover.

Roughly, everything ending with `NodeDiagnostic` inherits from this base class, and although they have `Children` of type `NodeDiagnostic `they are really"nested" in a tree like the following
```
DocumentNodeDiagnostic {
  Children: List<TopLevelDeclMemberNodeDiagnostic>
}
TopLevelDeclMemberNodeDiagnostic {
  Children: List<ImplementationNodeDiagnostic>
  AssertionBatches: List<AssertionBatchNodeDiagnostic>
}

ImplementationNodeDiagnostic {
  Children: List<AssertionNodeDiagnostic>
}
AssertionBatchNodeDiagnostic {
  Children: List<AssertionNodeDiagnostic>
}
```

The idea is that `AssertionBatchNodeDiagnostic ` are automatically computed from all the `ImplementationNodeDiagnostic`, and they are used for all the reporting, because users don't care about implementations.
However, we do care about implementations to be able to relocate previous diagnostics. Hence the tree structure.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +30,145 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
+      foreach (var node in document.VerificationNodeDiagnostic.Children.OfType<TopLevelDeclMemberNodeDiagnostic>()) {
+        if (node.Range.Contains(position)) {
+          var assertionBatchCount = node.AssertionBatchCount;
+          var assertionBatchIndex = -1;
+          var information = "";
```
MikaelMayer, line 60: There can be multiple assertions on the same token, from different assertion batches. This field records them all.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +30,145 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
+            assertionBatchIndex += 1;
+            if (!assertionBatch.Range.Contains(position)) {
+              continue;
+            }
+
+            var assertionIndex = 0;
+            var assertions = assertionBatch.Children.OfType<AssertionNodeDiagnostic>().ToList();
+            foreach (var assertionNode in assertions) {
+              if (assertionNode.Range.Contains(position) ||
+                  assertionNode.ImmediatelyRelatedRanges.Any(range => range.Contains(position))) {
```
MikaelMayer, line 71: For any immediately related ranges on the assertion, we also return the same node diagnostics. That way, it's possible to view the assertion diagnostics on postconditions directly.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +30,145 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
+                  }
+
+                  statusMessage = assertionNode?.StatusVerification switch {
+                    VerificationStatus.Verified => description.SuccessDescription,
+                    VerificationStatus.Error => description.FailureDescription,
+                    _ => statusMessage
+                  };
+                }
+
+                if (counterexample is ReturnCounterexample returnCounterexample) {
```
MikaelMayer, line 111: This is where I hope I can properly use @atomb 's work to get more precise error messages

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +30,145 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
+              assertionIndex++;
+            }
+          }
+
+          if (information != "") {
+            return information;
+          }
+          // Ok no assertion here. Maybe a method?
+          if (node.Position.Line == position.Line &&
+              node.Filename == document.Uri.GetFileSystemPath()) {
```
MikaelMayer, line 135: For the line containing the method, hovering over the first line of a method provides statistics of this method

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -158,14 +174,36 @@ private class ModelCapturingOutputPrinter : OutputPrinter {
 
       public void Inform(string s, TextWriter tw) {
         logger.LogInformation(s);
-        var match = Regex.Match(s, "^Verifying .+[.](?<name>[^.]+) [.][.][.]$");
-        if (match.Success) {
-          progressReporter.ReportProgress(match.Groups["name"].Value);
-        }
```
MikaelMayer, line 164: This code is no longer useful because verification diagnostics take care of reporting progress. Unless we deactivate it...?

Source/DafnyLanguageServer/Language/SyntaxTreeVisitor.cs
```diff
@@ -489,7 +489,10 @@ public abstract class SyntaxTreeVisitor {
           Visit(letExpression);
           break;
         default:
-          VisitUnknown(expression, expression.tok);
+          if (expression != null) {
+            VisitUnknown(expression, expression.tok);
+          }
+
```
MikaelMayer, line 495: This is a fix for #1918

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -32,6 +34,35 @@ public record DafnyDocument(
     /// </summary>
     public string? SerializedCounterExamples { get; init; }
 
+
+    /// <summary>
+    /// True is the resolution succeeded, false if resolution failed
+    /// <c>null</c> If the verification did not start (e.g. because of resolution errors)
+    /// </summary>
+    public bool? ResolutionSucceeded { get; set; } = null;
```
MikaelMayer, line 42: I need this on the VSCode level to know if the gutter is displayed in gray (resolution not succeeded) or regular.

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -32,6 +34,35 @@ public record DafnyDocument(
...
+
+    /// <summary>
+    /// True is the verification pass went through, false if it failed because of verification errors
+    /// <c>null</c> If the verification did not start (e.g. because of resolution errors or verification deactivated)
+    /// </summary>
+    public bool? VerificationPass { get; set; } = null;
```
MikaelMayer, line 48: In the case of on-save verification, I know that verification is deactivated, and I don't trigger the gutter highlight on the client.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+using Range = OmniSharp.Extensions.LanguageServer.Protocol.Models.Range;
+
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+
+namespace Microsoft.Dafny.LanguageServer.Workspace.Notifications {
+  /// <summary>
+  /// DTO used to communicate the current compilation status to the LSP client.
+  /// </summary>
+  [Method(DafnyRequestNames.VerificationDiagnostics, Direction.ServerToClient)]
+  public class VerificationDiagnosticsParams : IRequest, IRequest<Unit> {
```
MikaelMayer, line 22: This class is Public API, and after we agree on it, it should not be modified anymore, only completed because else future VSCode extensions would break.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+
+    /// <summary>
+    /// Gets the version of the document.
+    /// </summary>
+    public int? Version { get; init; }
+
+    /// <summary>
+    /// Gets the same diagnostics as displayed in the diagnostics window
+    /// </summary>
+    public Container<Diagnostic> Diagnostics { get; init; }
```
MikaelMayer, line 36: I do include the diagnostics so that if there are resolution or parse errors, the VSCode extension also highlights them in the IDE (while the gutter icons are greyed out)

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+
+    /// <summary>
+    /// Returns a tree of diagnostics that can be used
+    /// First-level nodes are methods, or witness subset type checks
+    /// Second-level nodes are preconditions, postconditions, body verification status
+    /// Third level nodes are assertions inside functions 
+    /// </summary>
+    public NodeDiagnostic[] PerNodeDiagnostic { get; init; }
```
MikaelMayer, line 44: I do not return the DocumentNodeDiagnostic, only its children.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+
+    /// <summary>
+    /// Returns per-line real-time diagnostic
+    /// </summary>
+    public LineVerificationStatus[] PerLineDiagnostic { get; set; } = new LineVerificationStatus[] { };
+
+    public void RecomputePerLineDiagnostics() {
+      PerLineDiagnostic = RenderPerLineDiagnostics(this, PerNodeDiagnostic, LinesCount,
+        NumberOfResolutionErrors, Diagnostics);
+    }
```
MikaelMayer, line 61: This is where the magic occurs. It works like a depth buffer: Every node is rendered on the same array of diagnostics for the gutter, but the priority determines what to really display, e.g. errors have a stronger priority than verified assertions, which have a stronger priority than error ranges.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+      }
+
+      // Render node content into lines.
+      foreach (var nodeDiagnostic in perNodeDiagnostic) {
+        if (nodeDiagnostic.Filename == verificationDiagnosticsParams.Uri.GetFileSystemPath() ||
+            "untitled:" + nodeDiagnostic.Filename == verificationDiagnosticsParams.Uri) {
+          nodeDiagnostic.RenderInto(result);
+        }
+      }
+
```
MikaelMayer, line 87: In order to not distract users, we fill the gap between functions and methods with "verified" status if either side is verified. This can be implemented by a forward and backward pass.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+    }
+  }
+
+
+  public enum VerificationStatus {
+    Unknown = 0,
+    Verified = 200,
+    Inconclusive = 270,
+    Error = 400
+  }
```
MikaelMayer, line 141: This is public API and should carefully be agreed on. It is the verification status of every node.
Note that, when rendering linearly, Inconclusive nodes are not displayed (so that we don't distract the user with assertions that were not checked)

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+
+  public enum CurrentStatus {
+    Current = 0,
+    Obsolete = 1,
+    Verifying = 2
+  }
```
MikaelMayer, line 147: This is public API and should carefully be reviewed. Together with VerificationStatus, they define the matrix of all possible node statuses.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+    ErrorRangeAssertionVerifiedObsolete = 351,
+    ErrorRangeAssertionVerifiedVerifying = 352,
+    ErrorRangeAssertionVerified = 350,
+    // For specific lines which have errors on it. They take over verified assertions
+    ErrorObsolete = 401,
+    ErrorVerifying = 402,
+    Error = 400,
+    // For lines containing resolution or parse errors
+    ResolutionError = 16
+  }
```
MikaelMayer, line 175: This is public API, and it roughly corresponds to all the possible icons in the gutter.
Moreover, higher number means higher priority to be displayed in the `perLineDiagnostic` array.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+
+  public record NodeDiagnostic(
```
MikaelMayer, line 177: This is public API. All the properties are important, not just the one in the record constructor.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+    public int ResourceCount { get; set; } = 0;
+
+
+
+    // If this node is an error, all the trace positions
+    public ImmutableList<Range> RelatedRanges { get; set; } = ImmutableList<Range>.Empty;
+
+    // Sub-diagnostics if any
+    public List<NodeDiagnostic> Children { get; set; } = new();
+    private List<NodeDiagnostic> NewChildren { get; set; } = new();
```
MikaelMayer, line 211: To work with existing diagnostics and ensure that at the end everything is coherent,, I use several times a mechanism of "double buffering". I keep the previous (migrated) children, potentially updating them with the new verification results, but I also add new verification results in `NewChildren`, so that when the verification of this node finishes, I just replace `Children` with `NewChildren`.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+          child.Stop();
+        }
+        Finished = true;
+        return true;
+      }
+
+      return false;
+    }
+
+    public void PropagateChildrenErrorsUp() {
```
MikaelMayer, line 270: Before sending diagnostics, I do this bottom-up propagation of errors to ensure that, if there is a descendant of that node with one error, this node is marked with error.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+        }
+      }
+
+      if (childrenHaveErrors) {
+        StatusVerification = VerificationStatus.Error;
+      }
+    }
+
+    // Requires PropagateChildrenErrorsUp to have been called before.
+    public virtual void RenderInto(LineVerificationStatus[] perLineDiagnostics, bool contextHasErrors = false, bool contextIsPending = false, Range? otherRange = null, Range? contextRange = null) {
```
MikaelMayer, line 285: This is a quite complex yet straightforward way to convert the verification status and current status of a node into a "linear diagnostic", using double nested switch statements.
It is context-sensitive, so that if a node has errors, we display verifying assertions with special icons and not green lines.
Moreover, for ranges that exceed one line (i.e. everything not an assertion), we use special range icons to render them if they are errors.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+        foreach (var child in Children) {
+          child.SetVerifiedIfPending();
+        }
+        return true;
+      }
+
+      return false;
+    }
+
+    public virtual NodeDiagnostic GetCopyForNotification() {
```
MikaelMayer, line 380: I solved all my concurrency problems by cloning the node diagnostics before sending them to the client. Indeed, that way the language server protocol can enumerate them while I keep updating the old copy.

Source/Directory.Build.props
```diff
@@ -7,7 +7,7 @@
 
   <!-- Boogie dependency -->
   <ItemGroup>
-    <PackageReference Include="Boogie.ExecutionEngine" Version="2.13.2" />
+    <ProjectReference Include="..\..\boogie\Source\ExecutionEngine\ExecutionEngine.csproj" />
```
MikaelMayer, line 10: Ignore this, it's going to be removed anyway.

MikaelMayer (commented):
Here are my preliminary remarks for my dear reviewers

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+    Error = 400
+  }
+
+  public enum CurrentStatus {
+    Current = 0,
+    Obsolete = 1,
+    Verifying = 2
+  }
+
+  public enum LineVerificationStatus {
```
keyboardDrummer, line 149: I have trouble understanding the semantics of all these values.

Does `Obsolete` mean `Stale` ?

My understanding is that one implementation can have multiple splits, and in the case of `:focus` some assertions may occur in multiple splits. Each split can have multiple counterexamples but in each counterexample only a single assertion is failing. I believe a single split only has at most one counterexample per assertion.

Given the above, I would expect each piece of code to have the following verification states: 
- Verified. All splits containing this code have verified.
- Error. There is at least one counterexample in which this code occurs.
- Unknown. There's at least one not run split that contains this code. Has these fields:
  - Reason: CouldNotVerify | NotQueued | Verifying. Could not verify means there was a pre-verification error. NotQueued means verification still needs to be done and is not queued. Verifying means it is either currently being verified or queued to be verified.
  - Previous: WasVerified | WasError | WasUnknown.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
keyboardDrummer, line 177: What do you need something hierarchical for? Generally LSP sends information that corresponds to the UI in the IDE, without sending any semantically interpretable information. One example of where LSP sends hierarchical data is with the `textDocument/documentSymbol` request, which is uses to populate the symbol tree. Is there a tree UI that you wish to populate using these NodeDiagnostics?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+    /// Third level nodes are assertions inside functions 
+    /// </summary>
+    public NodeDiagnostic[] PerNodeDiagnostic { get; init; }
+
+    /// <summary>
+    /// The number of lines in the document
+    /// </summary>
+    public int LinesCount { get; init; }
+
+    public int NumberOfResolutionErrors { get; init; }
```
keyboardDrummer, line 51: What's this relevant for?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,664 @@
...
+
+    /// <summary>
+    /// Returns per-line real-time diagnostic
+    /// </summary>
+    public LineVerificationStatus[] PerLineDiagnostic { get; set; } = new LineVerificationStatus[] { };
+
+    public void RecomputePerLineDiagnostics() {
```
keyboardDrummer, line 58: I'm surprised to have methods in a DTO. I would expect this to be an immutable type that gets all the information that it's supposed to send over the wire, and only that information, passed to it upon construction.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
keyboardDrummer: Seems like you can do this on the IDE side, by having any code that's not covered by verification diagnostics be marked as verified.

keyboardDrummer (commented):
At 2000+ lines this PR is fairly large. Is there a way to reduce the scope?

One of the features mentioned is "Automatic setting priority for verifying the last edited methods". Could we move that to a separate PR?

The same for "Offer verification diagnostics on hover of every assertion when the method check is finished".

keyboardDrummer:
Could you add instructions to the PR for how to try out these features in the VSCode IDE?

MikaelMayer:
> Could you add instructions to the PR for how to try out these features in the VSCode IDE?

Just added. Thanks for asking

atomb (commented):
I've just left a few comments for the moment. I'm continuing to review. I agree with @keyboardDrummer that anything we could do to review it in pieces could be helpful.

MikaelMayer (commented):
Thanks for all these comments ! They are very helpful and I hope I explained most of what was still obscure, that should give enough meat to further refine this PR.

>One of the features mentioned is "Automatic setting priority for verifying the last edited methods". Could we move that to a separate PR?

It might be harder than it seems. To do this, I would need to record a list of the last edited ranges, migrate it on every edit, and reimplement the traversal of `RecomputeVerificationNodeDiagnostics` and, without creating node diagnostics, I would just have to extract the range.
Moreover, this would not be a meaningful achievement until we publish in real-time diagnostics, because the ordering of implementations does not matter in the IDE, since all the verification diagnostics are sent at the very end.
I could however remove this logic from this PR, and then re-add it in another PR once this one is accepted and merged. Is that what you would expect?

I can explain however in more details the implementation of this feature, so that it can be reviewed more easily:
* In `TextDocumentLoader`, in the `VerificationProgressReporter`; I override the method `ReportImplementationsBeforeVerification` that boogie calls on any `OutputPrinter`, with the list of all the implementations to be verified and, in the second foreach loop, I call `GetVerificationPriority` on each implementation.  If a priority is set, I will assign it with a last `{:priority}` attribute of the implementation.
* I added the field `LastChange` in DafnyDocument. This field is set in `DocumentDatabase` with a loop iterating over all the changes sent in the same request:
```cs
        foreach (var change in documentChange.ContentChanges) {
          newDocument.LastChange = change.Range;
        }
```
`LastChange` obviously does not need migration and is read in `GetVerificationPriority`.
* I added the field `LastTouchedMethodPositions` in `DafnyDocument`. This contains a list of the last 5 ranges that were edited in the document. This field requires migration and is migrated in `DocumentDatabase.cs` line 150
* `GetVerificationPriority` uses the new field `Lastchange` and `LastTouchedMethodPositions` to determine an implementation's priority. To do so, it first recovers the top-level verification node that contains the implementation's position, and
  * If that node's range contains the last change position, it records this method's starting position in `RememberLastTouchedMethod` as the last element of `LastTouchedMethodPositions` (removing any duplicate). and returns a priority of 10
  * Else, it looks for whether this node's starting position is in the list `LastTouchedMethodPositions`, and returns 1 + its index in the list as a priority.

 > "Offer verification diagnostics on hover of every assertion when the method check is finished"
Of course I can definitely offer these diagnostics as a separate PR that would depend on this one. Note that however this is one of the most interesting achievements of this PR and that's why I call it "Verification Diagnostics" and not "dynamic gutter highlight, diagnostics", because this is something that has been long-time awaited.

Before I take the costly decision to pull it in a separate PR, let me give you some walkthrough of this feature so that you can review it more easily. If after that it is still too much, I'll happily move it to another PR that won't be mergeable until this PR is merged.

Everything in this feature is in `DafnyHoverHandler`, the method `Handle`
* First, instead of immediately returning an hover when a symbol is found, I assign the markdown string in a variable, which enables me to concatenate it with any assertion diagnostics with the following lines:
```cs
hoverContent = diagnosticHoverContent == null ? "" : diagnosticHoverContent;
hoverContent += symbolHoverContent == null ? (hoverContent != "" ? "  \n" : "") : symbolHoverContent;
```

* `diagnosticHoverContent` is assigned using the new method `GetDiagnosticsHover`
  * In this method, I am traversing the hierarchy of TopLevelDeclMemberNodeDiagnostic, then I skip their children (which are implementations that users don't care about), and traverse instead the aggregated `node.AssertionBatches` which are of type `AssertionBatchNodeDiagnostic`. I'm keeping track of the assertion batch index with `assertionBatchIndex` to be able to display it to users.
  * Since every node contains a `Range`, I can only traverse node that contain the hovering position, which is efficient.
  * I either give two kinds of hovering statistics: Individual assertions or per-method/subset type/field. If an individual assertion is hovered, the other statistics are not displayed.
    * For an individual assertions, I tolerate it's "hovered" also if the hover position is touching an `ImmediatelyRelatedRange` (i.e. the postcondition for return AssertCmd). `assertionNode.ImmediatelyRelatedRanges.Any(range => range.Contains(position))`
    * I then add all the statistics in the string `information`: The assertion number and total number of assertions in this batch (only if it's more than 1), the batch number over the total number of batches and its timing and resource count.
    * If the assertion is for a return statement and we are hovering the postcondition (which has normally no error associated with it), I display the position of the corresponding return (which contains more information about the error)
    * If there is a counter example for the current assertion, I use its `Description` to generate an even better message about its status. (i.e. `SetDescription(returnCounterexample.FailingReturn.Description);`)
    * At the end, I display "recently" if the status is obsolete, or "recently and verifying..." if the status of the current assertion is verifying (which is not implemented yet but soon)
  * If there is some information about a single assertion, I return it, else I check if the hover position is on the first line of a method, and if so:
    * I display "_Verification not started yet_" if its status is "Obsolete" (not started yet)
    * I display "_Still verifying " and the time elapsed if it "Verifying" (started to verify)
    * I display the longest assertion batch time and the line number of its first assertion, and the resource count for that batch. This is the best measures of instability, that users care about. We can definitely indicate different informations if someone has cleverer ideas. One of my concerns is that it does not tell if this longest time is the same for other assertions or if it's an outlier. I guess we can handle this later.

PR #1924, opened 2022-03-21
Properly check if types may contain references

### LINKED ISSUE #1419: function value can depend on set of allocated objects
The following verifies:
```
type Obs = set<object>
predicate lotsOfObjects() { 
    exists o:Obs :: |o| > 10
}
```
(It doesn't verify if you replace ```o:Obs``` with ```o:set<object>```.)

RustanLeino:
Previously, types were not properly checked for whether or not they may contain references. This PR fixes that.

Fixes #1419


cpitclaudel (approved):
For datatypes, why is it enough to check just the type parameters?  The following looks off to me:

```
class C {}
datatype D = DC(c: C)
type X(!new) = D // No error

datatype Obs' = Obs'(s: set<object>)
predicate lotsOfObjects'() {
  exists o: Obs' :: |o.s| > 10 // No error
}
```

And also this:

```
class C {}

datatype Box<T> = Box(t: T)
type Box'(!new) = Box<C> // Rejected

datatype BoxC = BoxC(c: C)
type BoxC'(!new) = BoxC // No error
```

Am I missing something?

RustanLeino:
@cpitclaudel Thank you. These are (still) bugs.

RustanLeino:
@cpitclaudel I addressed the further issues in this PR with the commit I just pushed. However, the fix causes errors that the old Schorr-Waite test programs have a function that depends on the allocation state. To sort this out, Dafny needs a more flexible mechanism to indicate that quantifiers only quantify over allocated state. I will add this mechanism in a separate PR and then return to this one.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -1406,10 +1406,31 @@ public enum AutoInitInfo { MaybeEmpty, Nonempty, CompilableValue }
...
+    ///
+    /// The type parameters to recursive uses of datatypes may be passed in in different ways. In fact,
+    /// there is no bound on the set of different instantiations one can encounter with the recursive uses
+    /// of the datatypes. Rather than keeping track of type instantiations in (some variation of)
+    /// visitedDatatypes, the type arguments passed to a datatype are checked separately. If the datatype
+    /// uses all the type parameters it declares, then this will have the same effect. During the second
+    /// phase, formal type parameters (which necessarily are ones declared in datatypes) are ignored.
+    /// </summary>
+    public virtual bool ComputeMayInvolveReferences(ISet<DatatypeDecl>/*?*/ visitedDatatypes) {
+      return false;
```
cpitclaudel, line 1433: Is false the right default here?  It risks blowing up silently if we add more AST nodes later.

cpitclaudel (changes requested):

cpitclaudel (approved):
Very nice, thanks!

cpitclaudel (approved):

PR #1933, opened 2022-03-24
fix: DafnyOptions.O.Compiler is null, preventing instantiation of ModuleExportDecl

### LINKED ISSUE #1932: Bug: Using Dafny.dll to build an AST fails if no compiler specified.
Since #1894 , `ModuleExportDecl.FullCompileName` used in the constructor of `ModuleExportDecl` can throw a null pointer exception because `TopLevelDecl.FullCompileName` now [depends on](https://github.com/dafny-lang/dafny/commit/d9aa57676e1ceca6bf151e0fe30d7b0b83e63cc5#diff-fa257574acd41f12645bbc3b82e23fc93ac386d13a9fb58c0f5f54944a8359e3R4313) `DafnyOptions.O.Compiler`, which does not have a default value.

MikaelMayer:
Fixes #1932


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -4312,9 +4312,12 @@ public TopLevelDecl(IToken tok, string name, ModuleDefinition enclosingModule, L
...
           }
         }
-
-        return DafnyOptions.O.Compiler.GetCompileName(EnclosingModuleDefinition.IsDefaultModule,
+        if (DafnyOptions.O.Compiler != null) {
+          return DafnyOptions.O.Compiler.GetCompileName(EnclosingModuleDefinition.IsDefaultModule,
           EnclosingModuleDefinition.CompileName, CompileName);
+        } else {
+          return Declaration.IdProtect(EnclosingModuleDefinition.CompileName) + "." + Declaration.IdProtect(CompileName);
+        }
```
cpitclaudel, line 4320: Why do we not instantiate the compiler?

RELEASE_NOTES.md
```diff
@@ -1,7 +1,8 @@
 # Upcoming
 
-- fix: Miscompilation due to incorrect parenthesization in C# output for casts. (#1908)
+- fix: Miscompilation due to incorrect parenthesization in C# output for casts. (https://github.com/dafny-lang/dafny/pull/1908)
```
cpitclaudel, line 3: Why do you add the full link?

cpitclaudel (changes requested):
I don't think this is the right fix: we should make sure that the compiler is properly instantiated.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: I don't think the AST is the right place to instantiate a compiler anyway, it should be instantiated in the options if really instantiated manually. I we don't want the compiler to be null, I think we should always instantiate a default C# compiler. Would you prefer that?

RELEASE_NOTES.md
MikaelMayer: So that it's browsable when reading the README.md from a text editor, and in Github markdown it still displays nicely "#1908"

Source/Dafny/DafnyOptions.cs
```diff
@@ -27,6 +27,7 @@ public DafnyOptions()
       Prune = true;
       NormalizeNames = true;
       EmitDebugInformation = false;
+      Compiler = new CsharpCompiler();
```
cpitclaudel, line 30: Won't that break the .compileTarget flag?

cpitclaudel:
Could you have a look?  I don't think defaulting to the C# compiler is the right fix.  https://github.com/dafny-lang/dafny/pull/1945

cpitclaudel (approved):

cpitclaudel:
Can you fix the conflicts in RELEASE_NOTES?

PR #1702, opened 2022-01-11
Subset types fully accepted in comprehensions

MikaelMayer:
This fixes #1604 

More details to come later.

TODO: Need to test nested subset types + nested reference type and subset types.

PR #1936, opened 2022-03-26
feat: Introduce :older attribute

RustanLeino:
This PR introduces a new parameter modifier, `older`, which can be used with the parameters of a predicate to say that, if the predicate returns `true`, then the parameters declared as `older` are necessarily equally old as or older than the other parameters.

The updated reference manual contains a section that motivates and describes `older`. That's the text aimed at users. Here is a smaller--but not terribly motivated--example, along with an implementation-centric explanation:

For a predicate

``` dafny
predicate P(older x: X, y: Y)
```

the `older` modifier claims

``` dafny
forall h: Heap :: IsAlloc(y, h) && P(x, y) ==> IsAlloc(x, h)
```

That is, if the predicate returns `true`, then in any heap where `y` is allocated, then so is `x` --- in short, `x` is _older_ than `y`.

The reason for wanting to declare a predicate with `older` parameters is so that the predicate can be used as a way to ensure a quantified expression does not depend on the allocation state. For example,

``` dafny
function F(y: Y): int {
  if forall x: X :: P(x, y) ==> G(x, y) == 3 then
    100
  else
    0
}
```

Here is another example, using an `iset` comprehension:

``` dafny
function Collection(y: Y): iset<X> {
  iset x: X | P(x, y)
}
```

Dafny does not allow a function to depend on the allocation state; that is, the act of allocating more objects should not change the value of a function. But if `X` is a type that contains references, then, at first glance, it looks as if this function does depend on the allocation state. For instance, if `X` is a class type, then allocating more `X` objects gives more candidate values for the `iset` comprehension to consider, and if those `X` objects satisfy the `P` constraint, then the `iset` would become larger. By using the `older` modifier on `P`'s parameter `x`, it becomes known that any `x` that satisfies `P(x, y)` in this comprehension is older than `y`. Thus, the `iset` comprehension, in fact, does not depend on the allocation state.


keyboardDrummer:
Do I understand correctly that using `{:older}`, a Dafny user can write 

```
predicate {:older x} P(x: X, y: Y) {
 .. body ..
}
```

instead of 

```
predicate P(x: X, y: Y) 
  ensures forall h: Heap :: IsAlloc(y, h) && P(x, y) ==> IsAlloc(x, h)
{
 .. body ..
}
```

Could there be library functions that do something similar? Is that not possible because Heap and IsAlloc are not accessible in user space? Would it be an option to make those accessible in certain circumstances?

---

How often do we expect `{:older}` to be used? Can you give some indication of how the cost vs benefit of having this attribute? I'm sure I'm missing all the context for this PR so sorry for that.

Also would it be possible to implement this change are purely an extension to the Dafny compiler, with all the code being in a separate file? Given that it's adding an attribute that seems like it should not be a problem. I think that having the code not be intertwined with the existing code would lower the cost of having it by a lot.

robin-aws:
+1 to @keyboardDrummer's questions - I get the motivation of being able to ensure a pure expression does not depend on future allocations, but I'm definitely not convinced this feature is the best way to deliver that functionality.

keyboardDrummer:
> The reason for wanting to declare a predicate with :older is so that the predicate can be used as a way to ensure a quantified expression does not depend on the allocation state.

Would it also be possible to use reads clauses to indicate this?

Some like a reads block:

```
function F(y: Y): int
 reads {} {
  if forall x: X :: G(x, y) == 3 then
    100
  else
    0
}
```

RustanLeino:
Here are two cryptic answers to questions brought up above:
* `IsAlloc` and `Heap` refer to things in the internal encoding of Dafny into Boogie. They are not available directly to Dafny programmers.
* It's not possible to encode the situation using `reads` clauses.

To understand the situation better, let me write down the motivation. (please stand by)

RustanLeino:
# Motivation

Let me lead up to the motivation, starting with some background information about memory safety, references, quantifiers, and functions.

## Quantifications implicitly quantify over allocated state

Dafny is a memory-safe programming language. Consequently, every object reference is either `null` or points to an allocated object. In other words, every reference (pointer) is _valid_. There is no way for a program to get hold of a pointer to un-allocated memory.

This property should carry over to comprehensions and quantifications as well. (Everything I'll say about quantifications applies to comprehensions in general; for example, set comprehensions. But I'll just say "quantifications" to stay more focused.) So, in Dafny, if `C` is an object type, then

``` dafny
forall c: C :: ...
```

quantifies over all references to _allocated_ `C` objects.

What I said applies not just to object types, but to any type that _contains_ a reference. For example, given

``` dafny
datatype List<T> = Nil | Cons(T, List<T>)
```

the type `List<int>` does not contain references, but the type `List<C>` does (where `C` is an object type). So, a quantification

``` dafny
forall cs: List<C> :: ...
```

quantifies over all lists whose every `C` reference is allocated.

## Functions are not allowed to depend on the allocation state

Consider the following attempted program:

``` dafny
class Cell {
  const data: int
  constructor (x: int) {
    this.data := x;
  }
}

function P() {
  forall cell: Cell :: c.data == 10 // Dafny does not allow this quantifier in a function
}

method M() returns (c: Cell)
  requires P()
  ensures P()
{
  c := new Cell(8);
}
```

It would be bad if this program would verify, because `P` does not hold at the end of `M`. (You might also reasonably ask how any program can live up to the precondition of `M`. I don't know of such a way. But `P()` does hold in some possible program states, so we do want to make sure `M` ensures its postcondition given its precondition.)

If a function were allowed to depend on the allocation state (like the attempted `P` does above), then any object allocation could potentially change the value returned by the function. This would make it difficult to specify and verify programs, especially considering that any method is allowed to allocate new objects.

Instead, Dafny bans quantification over any reference-containing type in function bodies. Thus, Dafny rejects the program above. The check is enforced syntactically by the resolver--even before the program gets to the verifier. 

## Allowing some quantifications over reference-containing types

There are cases where Dafny's ban of quantifications over reference-containing types (henceforth called _The Ban_) is too strict. For example, suppose you want to define a predicate that says a given relation is transitive. You might write:

``` dafny
predicate Transitive<X>(R: (X, X) -> bool) {
  forall x, y, z :: R(x, y) && R(y, z) ==> R(x, z) // Dafny does not allow this quantifier in a function
}
```

Alas, The Ban forbids the predicate, since it may be that type `X` contains reference. You are, however, allowed to write this predicate if you restrict the type parameter `X` to types that do not contain references. This is expressed by the somewhat cryptic-looking `(!new)` type characteristic:

``` dafny
predicate Transitive<X(!new)>(R: (X, X) -> bool) {
  forall x, y, z :: R(x, y) && R(y, z) ==> R(x, z) // since X is declared with (!new), the quantification is allowed
}
```

## Allowing more quantifications

Dafny's syntactic check that enforces The Ban is not quite as strict as I've described it. It allows quantifications if the bound variables are limited to things that are allocated already. For example, you are allowed to write

``` dafny
predicate Transitive<X>(R: (X, X) -> bool, universe: set<X>) {
  forall x, y, z ::
    x in universe && y in universe && z in universe ==>
    R(x, y) && R(y, z) ==> R(x, z)
}
```

(note the absence of `(!new)` in this definition). The justification for allowing this quantifier in a function is, roughly speaking, that `x`, `y`, and `z` range over values that are necessarily allocated. More precisely, the fact that `x`, `y`, and `z` are allocated follows from the fact that the parameter `universe` is allocated. This means that the following program snippet always verifies:

``` dafny
var b0 := Transitive(R, U);
var c := new Cell(8);
var b1 := Transitive(R, U);
assert b0 == b1;
```

Stated differently, `Transitive(R, U)` does not depend on any object allocated _after_ the allocation state of `U`. In other words, `x`, `y`, and `z` range over values that are allocated _before_ `U`. That is, `x`, `y`, and `z` are _older_ (in the allocation-state order) than `U`.

## User-declared "older" relations

Here is a simple, typical quantification that is allowed in a function:

``` dafny
predicate AllZeros(S: set<Cell>) {
  forall c :: c in S ==> c.data == 0
}
```

As I mentioned above, the syntactic scan performed by the enforcement of The Ban detects the expression `c in S` in the antecedent of the quantifier, and this shows that `c` is "older" than parameter `S`, and therefore the quantification is allowed.

But what if you wanted to write the expression `c in S` in another function? For example, what if you wrote the following?

``` dafny
predicate AllZeros(S: set<Cell>) {
  forall c :: InMySet(c, S) ==> c.data == 0
}

predicate InMySet(c: Cell, S: set<Cell>) {
  c in S
}
```

This is not allowed, because the syntactic check performed by The Ban doesn't trace through the call to `InMySet`.

This PR introduces an attribute that lets The Ban understand user-defined predicates as ones that ensure something about "older". The idea is that you declare `InMySet` as

``` dafny
predicate {:older c} InMySet(c: Cell, S: set<Cell>) {
  c in S
}
```

which expresses that

> `InMySet(c, S)` implies that `c` is older than `S`.                (*)

This PR introduces the `{:older}` attribute and adds the semantic check that (*) holds. It also changes The Ban to look for `{:older}` attributes. So, with this PR, the updated `InMySet` above allows the `AllZeros` predicate.

Note that this design lets The Ban remain a syntactic check (it looks for the `{:older}` attribute), whereas the _meaning_ of `{:older}` is enforced by the verifier.

## Why bother?

Predicate `InMySet` above is too simple to motivate the effort of supporting `{:older}`. Here is an example (originally from `Test/dafny1/SchorrWaite.dfy`, and also included in this PR as `Test/dafny0/OlderVerification.dfy`) that could not be done without `{:older}`:

``` dafny
class Node {
  var children: seq<Node>
}

datatype Path<T> = Empty | Extend(Path, T)

predicate Reachable(source: Node, sink: Node, S: set<Node>)
  reads S
{
  exists via: Path<Node> :: ReachableVia(source, via, sink, S) // allowed because of {:older p} on ReachableVia
}

predicate {:older p} ReachableVia(source: Node, p: Path<Node>, sink: Node, S: set<Node>)
  reads S
  decreases p
{
  match p
  case Empty => source == sink
  case Extend(prefix, n) => n in S && sink in n.children && ReachableVia(source, prefix, n, S)
}
```

Since `Node` is a class type, `Path<Node>` is a reference-containing type. Therefore, The Ban will scrutinize the quantifier in `Reachable`. Bound variable `via` is passed in as parameter `p` to `ReachableVia`. Since `ReachableVia` is declared with `{:older p}`, The Ban is satisfied that the quantification in `Reachable` does not depend on the allocation state.

A final note: If this was a problem in `Test/dafny1/SchorrWaite.dfy`, how come the Schorr-Waite program has been in the test suite for 10 years without needing `{:older}`? Because of bug #1419, which is being fixed in PR #1924. The fix depends on this PR in order for the test suite to go through.

keyboardDrummer:
Thanks for the great explanations :D this clarifies things. I'm afraid I'm still not confident about any thoughts I have on this topic, but I'll share them anyways. Please bear with me.

> Quantifications implicitly quantify over allocated state

Is that to make the proofs more tractable? 

What if quantifications also implicitly quantify over readable state? In your example, `Reachable` already specifies `reads S`. Can't that mean that the instances of via in `exists via: Path<Node>` are limited to those containing references from `S` ?

Or differently, shouldn't quantifier variables always be older than the parameters of the function containing the quantifier? Why do I still need to specify an older annotation?

---

In your example, the `n in S` in 

```case Extend(prefix, n) => n in S && sink in n.children && ReachableVia(source, prefix, n, S)```

 already causes references in `p` to be older than `S`. Is it not possible to let the verifier propagate that information up to the quantifier without needing the developer to use the older attribute?


---

Would it be possible to use an expression instead of an attribute for this, to make it more flexible? Like

```
predicate Reachable(source: Node, sink: Node, S: set<Node>)
  reads S
{
  exists via: Path<Node> :: older(via, S) && ReachableVia(source, via, sink, S)
}

predicate ReachableVia(source: Node, p: Path<Node>, sink: Node, S: set<Node>)
  reads S
  decreases p
{
  match p
  case Empty => source == sink
  case Extend(prefix, n) => n in S && sink in n.children && ReachableVia(source, prefix, n, S)
}
```

Or 

```
predicate Reachable(source: Node, sink: Node, S: set<Node>)
  reads S
{
  exists via: Path<Node> :: ReachableVia(source, via, sink, S)
}

predicate ReachableVia(source: Node, p: Path<Node>, sink: Node, S: set<Node>)
  reads S
  decreases p
  ensures ReachableVia(source, p, sink, S) ==> older(via, S)
{
  match p
  case Empty => source == sink
  case Extend(prefix, n) => n in S && sink in n.children && ReachableVia(source, prefix, n, S)
}
```

**Comment by MikaelMayer** [comment]
This is a fantastic detailed explanation on why we need `{:older}`.

Something isn't quite right though on your last example:
It's impossible for `p` to be "older" than the parameters `source` and `sink`, because the path has to be allocated after these two nodes are allocated.
Can you please clarify on that?

**Comment by RustanLeino** [comment]
@MikaelMayer 

> Something isn't quite right though on your last example: It's impossible for `p` to be "older" than the parameters `source` and `sink`, because the path has to be allocated after these two nodes are allocated. Can you please clarify on that?

Yikes! Let me think about this one.

**Comment by RustanLeino** [comment]
@MikaelMayer Thanks for your observation. I believe the semantic property is encoded correctly and the `/help` message is stated correctly, but the error message was not right. I rephrased the error message, hoping the new phrasing will make it clear.

In a nutshell, previously, the error message implied that (when the predicate returns `true`, then) each `:older` parameter must be (equally old or) older than _each of the other parameters_. More accurately, the actual condition is that each `:older` parameter must be older than _one of the other parameters_. Even that isn't right, because there may be no other parameters and the `:older` parameter may be, say, an integer, which is always allocated. So, the accurate condition is

* If the predicate returns `true` in a state where all the other parameters are allocated, then the `:older` parameters are also allocated in that state.

I changed the error message to say you must prove that

* `:older` parameters are not newer than any other parameter when the predicate returns `true`

**Comment by keyboardDrummer** [comment]
Have you considered making `older` a keyword that can be placed in front of the parameter of a predicate? The annotation syntax and the description of this PR made me think this could be implemented as some sort of macro, but looking at the actual implementation that doesn't seem so.

File: Source/Dafny/Resolver.cs [hunk 1]
```diff
@@ -17338,6 +17401,30 @@ class DefaultValueSubstituter : Substituter {
           }
           continue;
         }
+        if (conjunct is FunctionCallExpr fce) {
```
keyboardDrummer, line 17404: I don't like to ask this, but could you follow-up with moving all bounds-related code from Resolver.cs to a separate file?

Also, could you put this new block of code in a separate method?

keyboardDrummer:
Wouldn't it be possible to always pass a modified version of the heap to functions, where the modified heap consists of only objects reachable from the function's explicit parameters? That way if the heap changes in ways unrelated to the function's explicit parameters, such as new objects being allocated, then modified heap passed to the function does not change either.

Source/Dafny/Resolver/BoundsDiscovery.cs
```diff
@@ -0,0 +1,592 @@
...
+      Contract.Requires(fce != null);
+      Contract.Requires(bv != null);
+      Contract.Requires(bounds != null);
+
+      var older = GetOlderParameters(fce.Function);
+      if (older == null) {
+        return;
+      }
+
+      // local worker function:
```
keyboardDrummer, line 265: Isn't any nested function a local worker function?

Source/Dafny/Resolver/BoundsDiscovery.cs
```diff
@@ -0,0 +1,592 @@
...
+            }
+            break;
+          default:
+            break;
+        }
+      }
+      return bounds;
+    }
+
+    private static void DiscoverBoundsFunctionCallExpr<VT>(FunctionCallExpr fce, VT bv, List<ComprehensionExpr.BoundedPool> bounds)
```
keyboardDrummer, line 254: I would prefer `boundVariable` instead of `bv`

Source/Dafny/Resolver/BoundsDiscovery.cs
```diff
@@ -0,0 +1,592 @@
...
+        return;
+      }
+
+      // local worker function:
+      bool DoesOlderApply(Expression e, string formalName) {
+        return e.Resolved is IdentifierExpr ide && ide.Var == (IVariable)bv && older.Contains(formalName);
+      }
+
+      var isOlderArgument = false;
+      if (!fce.Function.IsStatic && DoesOlderApply(fce.Receiver, "this")) {
```
keyboardDrummer, line 271: Consider changing this into 

```
if (!fce.Function.IsStatic && DoesOlderApply(fce.Receiver, "this")) {
  isOlderArgument = true;
}

if (!isOlderArgument) {
  ..
}
```

To make it more clear that the else branch code being in a branch is only an optimisation.

keyboardDrummer:
> So, I still favor `{:older}` as an attribute. What do you think?

I don't have a strong opinion, but a keyword in this case doesn't require you to repeat the parameter variable name, and has no boilerplate syntax `{: }`.

```
predicate {:older trimOfLastWheelOnExtraLargeTrucksDrivingThroughEurope} X(trimOfLastWheelOnExtraLargeTrucksDrivingThroughEurope: Trim, cell: Cell) {
}
```

Source/Dafny/Resolver/BoundsDiscovery.cs
RustanLeino: Yes, that's what the comment (redundantly) says.

Source/Dafny/Resolver/BoundsDiscovery.cs
```diff
@@ -0,0 +1,592 @@
...
+      Contract.Requires(bounds != null);
+
+      var older = GetOlderParameters(fce.Function);
+      if (older == null) {
+        return;
+      }
+
+      // local worker function:
+      bool DoesOlderApply(Expression e, string formalName) {
+        return e.Resolved is IdentifierExpr ide && ide.Var == (IVariable)bv && older.Contains(formalName);
```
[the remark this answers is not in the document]
RustanLeino: Without it, C# complains that you cannot `==` compare an `IVariable` and a `VT`.

RustanLeino:
I agree it would be nice to mark the parameter with `{:older}` rather than marking the enclosing function with an attribute that has to repeat the name of the parameter. That is, instead of writing

``` dafny
function {:older x} F(x: X, y: Y): Z
```

(as in the current PR), it would be nice if one could write

``` dafny
function F({:older} x: X, y: Y): Z
```

(This design point also applies if we were using a keyword instead of an attribute.)

In fact, this is what I originally thought that "older" mark-up might look like. However, there are two complications.

The smaller complication is that Dafny currently doesn't allow parameters to have attributes. I think it would be nice to allow such attributes. But such a change to the language seems better considered as a separate issue/PR. If such a feature does go into the language in the future, then we can revisit the current `:older` attributes to instead place them directly on each parameter.

The other complication is that there needs to be a way to mark the receiver parameter (`this`) with `{:older}`. This complication is not specific to `{:older}`, but applies to any attribute that we'd want to put on the receiver parameter. One way to deal with this complication is to allow `{:older}` on the enclosing function, with the meaning that it applies to the implicit `this` parameter of the function. This is less clear, since it may look like `{:older}` applies to the _function_, not just the receiver parameter. I seem to recall that both Java/JML and Spec# used this convention when the `this` parameter needed to be annotated. Another way to deal with this complication is to extend Dafny's syntax to allow you to declare the receiver parameter explicitly. For example, equivalently to the function declaration above, you could then write following (where I have also added the `{:older}` attribute on the receiver parameter):

``` dafny
function F({:older} this, x: X, y: Y): Z  // here, the type of 'this' is implicit
```

or

``` dafny
function F({:older} this: C, x: X, y: Y): Z  // here, the type of 'this' is explicit
```

In the latter, `C` would usually be the name of the enclosing type (in which `F` is a member). But it's also possible to allow `C` to be a subset type of the enclosing type, with the meaning that `F` could be invoked only on that subset type. As another example, the explicit type of the receiver could be given as `C?`, if `C` is a reference type. This would mean that you could also invoke the function on `null`, which for classes like linked lists and trees would be very handy. (Such a function could not be virtual.)

So, one possible addition to the language is to allow explicit `this` parameters. Another one is to allow attributes on parameters. But I think both of those possible designs are better done outside the current PR.

RustanLeino:
> @RustanLeino Is the following a correct understanding of the origin of the problem?
> 
> A function with a reads clause takes an implicit heap parameter. We want to restrict references that originate from quantifiers to those that are allocated in the implicitly passed in heap, to provide memory safety. However, we want the result of the function to only depend on its explicit parameters, so we have to restrict references originating from quantifiers further to a subset that's entirely determined by the explicit parameters.

Yes, this is correct.

> Given the above, wouldn't it be possible to always pass a modified version of the heap to functions, where the modified heap consists of only objects reachable from the function's explicit parameters? That way if the heap changes in ways unrelated to the function's explicit parameters, such as new objects being allocated, then the modified heap passed to the function does not change.

Yes, in principle, but this would be complicated for verification. It would mean that
* Every call to a function needs to be done with a restriction of the heap.
* The definition of this restriction involves reachability, which is tricky to encode.
* To (use the frame axiom to) reason that two calls to a function yield the same result, one needs to prove that two restricted heaps are the same. This would involve proving that what was modified in the heap does not overlap to what remains after the restriction, and this involves inductive proofs over the reachability. Most likely, such proofs would have to be done by the Dafny user, since many of them could not be done automatically.
* Such proofs would be especially difficult to carry out (by the Dafny user) if the types of the function's parameters are opaque to the caller.

So, I think such an encoding would be a bad idea.

> One I don't understand is why functions that don't have a reads clause, so they don't take an implicit heap, still have restrictions on their quantifiers.

Without the restriction, the

    $Is(c#1, Tclass._module.Cell())

in the Boogie encoding in your example would have to be

    $IsAlloc(c#1, Tclass._module.Cell(), $h)

in order to encode the semantics that `c#1` quantifies only over allocated objects. The `$h` I'm showing here is supposed to denote the heap in which the function is invoked, which means it needs to be a parameter to the function after all. In other words, a function that quantifies over references would then need to be encoded with a heap parameter, even if the function has an empty `reads` clause in Dafny.

> Also, in Boogie I don't see where the references from quantifiers are restricted to being allocated. Is that always done implicitly, in this case by the `c in S` that The Ban enforces?

Correct. The condition `Lit(S#0)[$Box(c#1)]` implies that `c#1` is older than `S#0`, that is, that if `S#0` is allocated, then so is `c#1`. This is precisely what allows the `$IsAlloc` above to be replaced by just `$Is`.

keyboardDrummer:
Thanks for explanation. I regret that we're asking Dafny users to understand yet another concept, that of `{:older}`, but I don't see an alternative. I feel this problem is related to https://github.com/dafny-lang/dafny/pull/1702, since both are about how to get the values for a quantifier, although that one has problems during compilation and here we have problems during verification.

Is it really needed to use quantifiers like you do in the motivating example? Can your motivating example not be replaced by:
```
predicate Reachable(source: Node, sink: Node, S: set<Node>)
  reads source
  reads S
  decreases S
{
  if source == sink then
    true
  else
    exists step: Node | step in S && step in source.children :: Reachable(step, sink, S - {step})
}
```

?

---

Regarding the syntax. You mention
> The other complication is that there needs to be a way to mark the receiver parameter (this) with {:older}

Can the user not workaround that by making the callable static and the receiver explicit? If yes, then I'm not sure we need to solve this problem.

`static {:older _this} function F(_this: C, x: X, y: Y): Z`

> The smaller complication is that Dafny currently doesn't allow parameters to have attributes.

Why would you want it to be an attribute, instead of a modifier keyword on parameters?

`function F(older x: X, y: Y): Z`

---

One thing I'm missing in error messages on quantifiers is the suggestion to `{:older}`. For example,
```
function Collection<X>(S: set<X>): iset<List<X>> {
  iset xs: List<X> | ElementsContainedIn(xs, S) // error: needs {:older xs} on ElementsContainedIn
}
```

Gives the error message:

```
Error: a set comprehension involved in a function definition is not allowed to depend on the set of allocated references, but values of 'xs' may contain references
```

How does the user find out they should use `{:older}` ?

---

Regarding the documentation, for me the motivating example you provided:
```
class Node {
  var children: seq<Node>
}

datatype Path<T> = Empty | Extend(Path, T)

predicate Reachable(source: Node, sink: Node, S: set<Node>)
  reads S
{
  exists via: Path<Node> :: ReachableVia(source, via, sink, S) // allowed because of {:older p} on ReachableVia
}

predicate {:older p} ReachableVia(source: Node, p: Path<Node>, sink: Node, S: set<Node>)
  reads S
  decreases p
{
  match p
  case Empty => source == sink
  case Extend(prefix, n) => n in S && sink in n.children && ReachableVia(source, prefix, n, S)
}
```

to provide more clarity than just

```
predicate {:older} P(x: X, y Y)

function F(y: Y): int {
  if forall x: X :: P(x, y) ==> G(x, y) == 3 then
    100
  else
    0
}
```

Because the motivating example makes it clear why we can't add `x in <Z>` in the quantifier, and how the `{:older}` annotation can be proven, but the latter example does not.

docs/DafnyRef/Types.md
```diff
@@ -2294,9 +2295,230 @@ When `{:opaque}` is specified for function `g`, `g` is opaque,
...
+  allocated state. Ater all, Dafny is a type-safe language where every
+  object reference is _valid_ (that is, a pointer to allocated storage
+  of the right type)---it should not be possible, not even through a
+  bound variable in a quantifier or comprehension, for a program to
+  obtain an object reference that isn't valid.
+
+* This property is ensured by disallowing _open-ended_ quantifiers.
+  More precisely, the object references that a quantifier may range
+  over must be shown to be confined to object references that are
+  already allocated. Quantifiers that are not open-ended are called
```
keyboardDrummer, line 2384: "already", what is the current time? I think it's not the time at which the predicate is called, because if I do two predicate calls with equivalent arguments, then the quantifier references must be allocated before the first call, even when evaluating the second call.

Shouldn't this say "More precisely, the object references that a quantifier may range
  over must be shown to be confined to object references that were allocated before one of the parameters passed to the predicate." ?

keyboardDrummer (approved):
Added some minor comments. I'm OOTO from the 26th PST time to the end of the week. If you want to make more changes, feel free to have someone else review just those. I'm approving the current ones.

Source/Dafny/Resolver.cs
```diff
@@ -6997,67 +6998,21 @@ public CheckTypeInference_Visitor(Resolver resolver, ICodeContext codeContext)
...
-        } else {
-          reporter.Error(MessageSource.Resolver, arg.tok, "each argument to :older is expected to be a parameter of the enclosing predicate");
-          continue;
-        }
-        if (!givenFormals.Contains(name)) {
-          givenFormals.Add(name);
-        } else {
-          reporter.Error(MessageSource.Resolver, arg.tok, ":older attribute has duplicated argument: '{0}'", name);
+        // parameters are not allowed to be marked 'older'
+        foreach (var formal in f.Formals) {
```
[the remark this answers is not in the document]
RustanLeino: The code produces one error for each `older` formal. This could be done with an `.Iter(formal => { ... })`, but for that, I think the current `foreach` is more readable.

Test/dafny0/OlderVerification.dfy.expect
```diff
@@ -0,0 +1,15 @@
+OlderVerification.dfy(89,12): Error: the 'older' parameter might be newer than all non-'older' parameters when the predicate returns 'true'
```
cpitclaudel, line 1: Maybe mention the name of the parameter?  `Parameter… is marked older but may be newer than all non-'older' parameters when predicate … returns 'true' `?

cpitclaudel (approved):

cpitclaudel (approved):

PR #1942, opened 2022-03-29
feat: Verification diagnostics (core version)

### LINKED ISSUE #1918: Crash of the language server

MikaelMayer:
Real-time verification diagnostics for the language server. Illustration on an example put in the documentation:
image

Also fixes #1918.

Features implemented:
- Record the real-time verification diagnostics provided by Boogie into the `DafnyDocument`
- Send verification diagnostics along with per-line status to the client for display
- Migrates verification diagnostics on document change
- Activate/deactivate verification diagnostics

## Try this in VSCode IDE yourself:
1. Checkout the branch [verification-diagnostics](https://github.com/dafny-lang/dafny/tree/verification-diagnostics) of Dafny
1. in `dafny/`, clone boogie and checkout the commit 6c703b1070fe3c96965289b7cb43c8fc79c5b2e3 (currently two above master)
1. Compile Dafny. It automatically depends on your installed nested boogie.
1. Clone [verification-diagnostics-gutter](https://github.com/dafny-lang/ide-vscode/tree/verification-diagnostics-gutter) from the VSCode extension, run `code .` in a terminal to open VSCode on it (or just run vSCode and open the `ide-vscode` folder), and press F5 to run another instance of VSCode.
1. Ensure in the settings that the paths to DafnyLanguageServer.dll points to the `dafny/Binaries/DafnyLanguageServer.dll` compiled in step 3.


Source/DafnyLanguageServer.Test/Diagnostics/LinearVerificationDiagnosticTester.cs
```diff
@@ -0,0 +1,264 @@
...
+
+  // Finds all the "?" at the beginning in expected and replace the characters at the same position in traceObtained
+  // by a question mark, so that we don't care what is verified first.
+  // Do this only if lengths are the same
+  public string AcceptQuestionMarks(string traceObtained, string expected) {
+    if (traceObtained.Length != expected.Length) {
+      return traceObtained;
+    }
+
+    var startOfLine = true;
```
atomb, line 248: There's a CI warning that this is never used.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -50,14 +61,35 @@ CancellationToken cancellationToken
...
       }
 
+      public List<Position> MigratePositions(List<Position> originalRanges) {
+        return contentChanges.Aggregate(originalRanges, MigratePositions);
+      }
+
       private IReadOnlyList<Diagnostic> MigrateDiagnostics(IReadOnlyList<Diagnostic> originalDiagnostics, TextDocumentContentChangeEvent change) {
         if (change.Range == null) {
-          throw new System.InvalidOperationException("the range of the change must not be null");
+          return new List<Diagnostic>();
```
atomb, line 70: What makes `null` allowed now?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,671 @@
...
+    /// </summary>
+    public Container<Diagnostic> Diagnostics { get; init; }
+
+    /// <summary>
+    /// Returns a tree of diagnostics that can be used
+    /// First-level nodes are methods, or witness subset type checks
+    /// Second-level nodes are preconditions, postconditions, body verification status
+    /// Third level nodes are assertions inside functions 
+    /// </summary>
+    public NodeDiagnostic[] PerNodeDiagnostic { get; init; }
```
atomb, line 61: If this is an external API, I worry that "node" is too generic a term to be widely understood here.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,671 @@
...
+      return result;
+    }
+  }
+
+
+  public enum VerificationStatus {
+    Unknown = 0,
+    Verified = 200,
+    Inconclusive = 270,
+    Error = 400
```
atomb, line 145: Is there any reasoning behind the specific numbers chosen here?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,671 @@
...
+    ErrorRangeVerifying = 302,
+    ErrorRange = 300,
+    // For individual assertions in error ranges
+    ErrorRangeAssertionVerifiedObsolete = 351,
+    ErrorRangeAssertionVerifiedVerifying = 352,
+    ErrorRangeAssertionVerified = 350,
+    // For specific lines which have errors on it. They take over verified assertions
+    ErrorObsolete = 401,
+    ErrorVerifying = 402,
+    Error = 400,
```
atomb, line 177: The names of the `ErrorRange...`, `ErrorRangeAssertion...`, and `Error...` constants don't seem super clear to me. I don't have a concrete suggestion for what they should be, though. :)

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,671 @@
...
+      LineVerificationStatus targetStatus = StatusVerification switch {
+        VerificationStatus.Unknown => StatusCurrent switch {
+          CurrentStatus.Current => LineVerificationStatus.Unknown,
+          CurrentStatus.Obsolete => LineVerificationStatus.Scheduled,
+          CurrentStatus.Verifying => LineVerificationStatus.Verifying,
+          _ => throw new ArgumentOutOfRangeException()
+        },
+        // let's be careful to no display "Verified" for a range if the context does not have errors and is pending
+        // because there might be other errors on the same range.
+        VerificationStatus.Verified => StatusCurrent switch {
```
atomb, line 304: I have a feeling this logic could be simplified, perhaps by defining some relationships between status constants.

atomb (changes requested):
This is an early review, with perhaps more to come! My main comment at this point is that I find some of the terms you use slightly confusing. It's probably less important in implementation details, but some of those terms show up in the API to communicate with the LSP  client, so are perhaps worth discussing more. In particular, the term "node" is very generic in the field, and could perhaps be replaced with something more descriptive?

See also my comment on the various types of error status. Slightly different names might be clearer?

Source/DafnyLanguageServer.Test/Various/SlowVerifier.cs
```diff
@@ -17,7 +18,8 @@ class SlowVerifier : IProgramVerifier {
 
   private readonly DafnyProgramVerifier verifier;
 
-  public Task<VerificationResult> VerifyAsync(Dafny.Program program, IVerificationProgressReporter progressReporter, CancellationToken cancellationToken) {
+  public Task<VerificationResult> VerifyAsync(DafnyDocument document, IVerificationProgressReporter progressReporter, CancellationToken cancellationToken) {
+    var program = document.Program;
```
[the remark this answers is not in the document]
MikaelMayer: Not in the slow verifier, but in the `DafnyProgramVerifier`, I need the entire document to publish `progressReporter.ReportRealtimeDiagnostics(document);`

Source/DafnyLanguageServer/Workspace/ChangeProcessors/IRelocator.cs
```diff
@@ -30,5 +31,7 @@ public interface IRelocator {
     /// <exception cref="System.OperationCanceledException">Thrown when the cancellation was requested before completion.</exception>
     /// <exception cref="System.ObjectDisposedException">Thrown if the cancellation token was disposed before the completion.</exception>
     IReadOnlyList<Diagnostic> RelocateDiagnostics(IReadOnlyList<Diagnostic> originalDiagnostics, DidChangeTextDocumentParams changes, CancellationToken cancellationToken);
+
+    NodeDiagnostic RelocateNodeDiagnostic(NodeDiagnostic oldVerificationNodeDiagnostic, DidChangeTextDocumentParams documentChange, CancellationToken none);
```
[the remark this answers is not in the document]
MikaelMayer: I'll add a comment.
I could rename it "TreeDiagnostic" instead, or "VerificationTree", what do you prefer?

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
MikaelMayer: This PR:;
https://github.com/dafny-lang/dafny/pull/1816

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
MikaelMayer: Yes although these numbers are totally arbitrary, there are several causes to them.
1) HTTP codes 200 means OK and HTTP code 400 means error.
2) I wanted that the linear status code be roughly the sum of this verification code and the status code.
3) One of these two codes has to be a higher number than the other, and it made sense that verified + pending is lower than error + current, although pending is greater (more up-to-date) than current.
3) Code 300 is currently used to report various error ranges (see LinearVerificationStatus)
4) I am not using Inconclusive right now, and I wanted something between verified and error range, but closer to error range.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
MikaelMayer: Alternatives, please comment on them:
* I could add underscores to split names? For example `Error_Obsolete`.
* I could explicit what `Error` means: `Error_Current` or even `ErrorAssertionFailedCurrent`, because every error is an assertion failure. Or just `AssertionFailedCurruent`
* I could replace `ErrorRange` with `RangeWithError`. They indicate we are in the context of a method that has errors.
* I could replace `ErrorRange` with `ErrorContext` as well
* I could replace `ErrorRangeAssertionverified` with `AssertionFailedContextButAssertionsVerified`

As I developed the naming myself, I'm pretty biased and would love to hear your external feedback.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/IRelocator.cs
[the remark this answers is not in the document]
atomb: Yeah, `VerificationTree` sounds nice to me.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
atomb: Seems reasonable!

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
atomb: Given those description, I now think I understand what they mean! :)

I think I like `AssertionFailedCurrent` (or even just `AssertionFailed`) and `ErrorContext`. Given those, maybe the last could be `AssertionVerifiedInErrorContext`?

atomb (changes requested):
This is a really nice new feature! And I appreciate how much work you've put into testing it. 

I think the only remaining thing is to choose the names for the public API concepts and constants. I think you suggested pretty good ones, and you can see my comments for one additional suggestion.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
MikaelMayer: Nice ! I like all three.

atomb (dismissed):
I like the new names. Nice work!

MikaelMayer:
> Looking at what the IDE extension does with this PR.
> 
> image
> 
> In the above image, what should I learn from the gutter icons? Are these all lines that relate to an assertion failure? If so, then why isn't line 10 shown red in the gutter as well, since that's the line with a bug?
> 
> When I click on one of the lines with a gutter error, I see this:
> 
> image
> 
> Here I would have expected the line `result := "goodbye"` to be highlighted in some way, to indicate that this line is part of the counterexample for the assertion failure, but it's not. Why are only the ensures clause and the return statement being highlighted?

You just highlighted the default behavior of Dafny, it does not report anything else in the verification path.
However, If you now turn on "verification trace" (a setting in VSCode), it should be able to highlight the line that is failing. The problem is, verification trace highlights so many things I don't think it's relevant to have it by default.

keyboardDrummer:
I like the verification trace option, although I have a few questions:

image

- Shouldn't the condition `x == "hi"` be highlighted instead of `if (`?
- Why are the braces `{` highlighted?
- Why is the return value `r` in `(r: string)` highlighted?
- Why is the parameter `x` in `(x: string)` highlighted?

keyboardDrummer:
image

What's the meaning of the green icons? I was under the assumption they were telling me every assertion that is reached and passes in all counterexamples, but it seems they can also be green for assertions that are not reached in all counterexamples.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,640 @@
...
+    Error = 400
+  }
+
+  public enum CurrentStatus {
+    Current = 0,
+    Obsolete = 1,
+    Verifying = 2
+  }
+
+  public enum LineVerificationStatus {
```
keyboardDrummer, line 156: I still have trouble understanding the meaning of these entries. The comments indicate that some values apply to assertions and other values apply to methods. Some values don't have a comment on them assigning them to other methods or assertions, so which do those belong to?

Does `Obsolete` mean, value is out of date and won't automatically become up to date, like when a user has 'verify on save' configured and has made edits?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,640 @@
...
+    // Overriden by checking children if there are some
+    public CurrentStatus StatusCurrent { get; set; } = CurrentStatus.Obsolete;
+
+    // Used to relocate a verification tree and to determine which function is currently verifying
+    public Position Position => Range.Start;
+
+    /// Time and Resource diagnostics
+    public bool Started { get; set; } = false;
+    public bool Finished { get; set; } = false;
+    public DateTime StartTime { get; protected set; }
```
keyboardDrummer, line 205: This information and the other performance related information is currently not used by the client right? I suggest taking it out to reduce the scope of this PR

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,640 @@
...
+        existsError = false;
+      }
+
+      return result;
+    }
+  }
+
+
+  public enum VerificationStatus {
+    Unknown = 0,
```
keyboardDrummer, line 144: To the user, I would think both Unknown and Verified+Obsolete mean "Don't worry about it yet". Do you need both?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,640 @@
...
+    AssertionVerifiedInErrorContext = 350,
+    // For specific lines which have errors on it. They take over verified assertions
+    AssertionFailedObsolete = 401,
+    AssertionFailedVerifying = 402,
+    AssertionFailed = 400,
+    // For lines containing resolution or parse errors
+    ResolutionError = 500
+  }
+
+  public record VerificationTree(
```
keyboardDrummer, line 184: Could you add a comment explaining what this tree represents?

MikaelMayer:
> Shouldn't the condition x == "hi" be highlighted instead of if (?
> Why are the braces { highlighted?
> Why is the return value r in (r: string) highlighted?
> Why is the parameter x in (x: string) highlighted?

The code to compute all these ranges in the verification trace is [here](https://github.com/dafny-lang/dafny/blob/bb05e64e80a598fe300eaff46b01c80d5b79405c/Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs#L573):
```csharp
                counterexample.Trace.SelectMany(block => {
                  var tokens = new List<IToken>();
                  if (block.TransferCmd != null) {
                    tokens.Add(block.TransferCmd.tok);
                  }
                  tokens.AddRange(block.cmds.Select(cmd => cmd.tok));
                  return tokens;
                }).ToList() : new List<IToken>();
```
There is more processing to remove duplicates and ensure they are ordered before returning to the client, but the code above is the main logic. It answers the question on "why is X highlighted".

Concerning the questions "shouldn't X be highlighted", I wholeheartedly agree.
I think that, in future work, we should clean up this verification trace and actually display what matters most, because all these ranges came from the translation to Boogie.
I prefer to leave this feature as experimental, unless you see something obvious I could do to enhance the display now.
If you prefer, I could also remove this feature for now and put it in a separate PR since it's hard to specify. But, I already found it to be sometimes useful for now.

MikaelMayer:
> What's the meaning of the green icons? Is any assertion that is not marked red automatically marked green?
> 
> The above example gives me the false impression that the `assert true;` marked green is live, but I'm guessing that's an edge case so I think that' okay.

Green icons are displayed when, in an assertion batch, it can no longer extract failing assertions when assuming that those which failed are true. So if you are asserting `false`, there is no failing assertions afterwards since you assumed `false`. In the future, we might want to have "conditional verification icons" that show an assertion holds but one of the assertions it depends on has an error.
Since we also extract at most 5 failing assertions consecutively, after 5 note that we don't highlight remaining assertions. So, in the case below, we don't highlight `assert true` for that reason
image

but if you remove one assertion, it's able to prove that no other assertion is failed after the first four and it can highlight `assert true` in green.
image

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,640 @@
...
+
+    /// <summary>
+    /// Gets the version of the document.
+    /// </summary>
+    public int? Version { get; init; }
+
+    /// <summary>
+    /// Gets the same diagnostics as displayed in the diagnostics window
+    /// </summary>
+    public Container<Diagnostic> Diagnostics { get; init; }
```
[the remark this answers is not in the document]
MikaelMayer: Yes, but I am unable to recover them anyway from the IDE, as it's sent directly to built-in methods that do underline highlighting and diagnostics window.
I need these to consistently highlight parsing errors the same as I do for verification errors, for consistency. It is useful to actually display only resolution errors if they exist and not highlight verification errors in that case, which differs from the current diagnostics mechanism that highlights everything.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,640 @@
...
+
+    /// <summary>
+    /// Returns a tree of diagnostics that can be used
+    /// First-level nodes are methods, or witness subset type checks
+    /// Second-level nodes are preconditions, postconditions, body verification status
+    /// Third level nodes are assertions inside functions 
+    /// </summary>
+    public VerificationTree[] VerificationTrees { get; init; }
```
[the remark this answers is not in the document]
MikaelMayer: As indicated in the PR description: _"Send verification diagnostics along with per-line status to the client for display"_, this PR is not just scoped for showing icons in the gutter, but to also display dynamic relationships between errors (e.g. return paths and postconditions) and also the verification trace, for which the verification tree is useful because it contains all this information.

I opens the way to displaying the verification tree later during extension development.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
MikaelMayer: I see what you mean. That's always the problem of null versus undefined.

Let me rename "Unknown" to "Nothing", because it's the initialization value for every line.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
MikaelMayer: I use `Started` in the `VerificationTree.Start()` method to ensure I set all the children's current status as `Verifying` only once.
I use `Finished` in the `Verification.Stop()` method to ensure I stop only once, because this action replaces Children by the "NewChildren" array.

Other than that, yes, I'll move `ResourceCount`, `StartTime`, `EndTime` and `TimeSpent` and other associated diagnostics metrics to the branch -verification-cover-hover.

keyboardDrummer:
> > Shouldn't the condition x == "hi" be highlighted instead of if (?
> > Why are the braces { highlighted?
> > Why is the return value r in (r: string) highlighted?
> > Why is the parameter x in (x: string) highlighted?
> 
> The code to compute all these ranges in the verification trace is [here](https://github.com/dafny-lang/dafny/blob/bb05e64e80a598fe300eaff46b01c80d5b79405c/Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs#L573):
> 
> ```cs
>                 counterexample.Trace.SelectMany(block => {
>                   var tokens = new List<IToken>();
>                   if (block.TransferCmd != null) {
>                     tokens.Add(block.TransferCmd.tok);
>                   }
>                   tokens.AddRange(block.cmds.Select(cmd => cmd.tok));
>                   return tokens;
>                 }).ToList() : new List<IToken>();
> ```
> 
> There is more processing to remove duplicates and ensure they are ordered before returning to the client, but the code above is the main logic. It answers the question on "why is X highlighted".
> 
> Concerning the questions "shouldn't X be highlighted", I wholeheartedly agree. I think that, in future work, we should clean up this verification trace and actually display what matters most, because all these ranges came from the translation to Boogie. I prefer to leave this feature as experimental, unless you see something obvious I could do to enhance the display now. If you prefer, I could also remove this feature for now and put it in a separate PR since it's hard to specify. But, I already found it to be sometimes useful for now.

I really like the feature, but I think API wise the information should be provided through a separate request, instead of through a notification on every textual change, since the information is only shown after a user action (selecting an error). I think it's better to have it in a separate PR.

keyboardDrummer:
> > What's the meaning of the green icons? Is any assertion that is not marked red automatically marked green?
> > The above example gives me the false impression that the `assert true;` marked green is live, but I'm guessing that's an edge case so I think that' okay.
> 
> Green icons are displayed when, in an assertion batch, it can no longer extract failing assertions when assuming that those which failed are true. So if you are asserting `false`, there is no failing assertions afterwards since you assumed `false`. In the future, we might want to have "conditional verification icons" that show an assertion holds but one of the assertions it depends on has an error. Since we also extract at most 5 failing assertions consecutively, after 5 note that we don't highlight remaining assertions. So, in the case below, we don't highlight `assert true` for that reason image
> 
> but if you remove one assertion, it's able to prove that no other assertion is failed after the first four and it can highlight `assert true` in green. image

And suppose we wouldn't have the 5 limit, would then any assertion not marked as red be marked green?

keyboardDrummer:
I think it would be easier for me to understand the computation of `RenderPerLineDiagnostics` by first understanding the goal. I get the impression there are several goals of the gutter icons, namely:
- Say whether the information on this line is up-to-date, and if it's not, whether it will become up to date if the user waits or not. Is this information always passed down from the method to the individual lines?
- Say whether this line contains a failed assertion or a return location part of a failed assertion, and if so mark it red.
- In case the line was not marked red, mark it green if we did not hit the error reporting upper bound of 5.

Is the above correct?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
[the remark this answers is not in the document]
keyboardDrummer: Can't you use `languages.onDidChangeDiagnostics` to determine the diagnostics that get sent to the IDE ?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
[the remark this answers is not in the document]
keyboardDrummer: > to also display dynamic relationships between errors (e.g. return paths and postconditions)

What part of the tree is used for that?

> It opens the way to displaying the verification tree later during extension development.

I'd prefer leaving that out of scope for this PR. We should think about what an efficient way to update the tree is. Maybe we should have a request from the client that sends over the entire tree, and then notifications that only update subtrees within the trees.

MikaelMayer:
> I think it would be easier for me to understand the computation of `RenderPerLineDiagnostics` by first understanding the goal. I get the impression there are several goals of the gutter icons, namely:
> 
> * Say whether the information on this line is up-to-date, and if it's not, whether it will become up to date if the user waits or not. Is this information always passed down from the method to the individual lines?
> * Say whether this line contains a failed assertion or a return location part of a failed assertion, and if so mark it red.
> * In case the line was not marked red, mark it green if we did not hit the error reporting upper bound of 5.
> 
> Is the above correct? Am I missing something?

The goal of the gutter icon is to portray a useful view of the verification status of every code line, in the current editor. The same icons are even abstracted in the scroll bar by simple colors on the right to give an overview of the entire file.

> * Say whether the information on this line is up-to-date, and if it's not, whether it will become up to date if the user waits or not.

Correct

> Is this information always passed down from the method to the individual lines?

I'm not sure about the meaning of this question, but if the question is "will the information about the verification status of a method will be always visible on the lines that this range has", the answer is yes. 

Moreover, [I even added a mechanism here](https://github.com/dafny-lang/dafny/blob/a77d2e11defe262afe744a7d8a23ce4db915a2de/Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs#L347): so that in the case when all the lines contain verified assertions in a context that did not have any status before (i.e. the first time), it will still override the first line of the context range with the "verifying status" so that we don't claim the range is fully verified until it's really entirely verified. I had a case with a subset type like this:

```dafny
type IssueId = i : int | isIssueIdValid(i)
     witness 99
```
Since it does two implementations, one for well-formedness and one to prove the witness, it happened that the well-formedness implementation had two assertions that covered the entire range. It was thus displaying two "green" vertical bars, giving the false impression that the entire range was verified, even momentarily. [The mechanism](https://github.com/dafny-lang/dafny/blob/a77d2e11defe262afe744a7d8a23ce4db915a2de/Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs#L347) ensures that at least the first line of the range is displayed as "verifying" if there are still other implementations to verify.

> * Say whether this line contains a failed assertion or a return location part of a failed assertion, and if so mark it red.

Correct

> > * In case the line was not marked red, mark it green if we did not hit the error reporting upper bound of 5.

Almost: mark a line as green (either Verified if it's the first time or the context was previously verified, or AssertionVerifiedInErrorContext if the context was previously an error) only if there was at least an assertion on this line that was verified. Lines without assertions are using only context icons, either ErrorContext if at least one child has an error, or Verified if no error was reported.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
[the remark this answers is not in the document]
MikaelMayer: That works ! Thanks, I had been hoping these existed before. I did not know that they were in "languages", that's odd, but they work as expected. I'm going to remove the inclusion of diagnostics from this API.

keyboardDrummer:
> Since this verification trace consumes bandwidth only in the case of assertion errors, it has a very low bandwidth footprint. 

Instead of talking about why it's not a problem to do X, let me take a step back and ask what is the benefit of doing X. Is there a reason to include the verification traces as part of an API that also contains other information? To me verification traces seems like a feature that stands on its own and can be implemented separately, so why mix it with other features?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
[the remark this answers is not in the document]
MikaelMayer: > What part of the tree is used for that?

relatedRanges (for the verification trace), immediatelyRelatedRanges (for the errors in the same context as the current error range, such as the return path), and dynamicallyRelatedRanges (for ranges that would otherwise be displayed as "Related location" in the diagnostics but happen outside of the context, i.e. for failing call preconditions)

> I'd prefer leaving that out of scope for this PR. We should think about what an efficient way to update the tree is. Maybe we should have a request from the client that sends over the entire tree, and then notifications that only update subtrees within the trees.

What you are suggesting is indeed so complex it is clearly out of scope for this PR.
I like your idea of sending only partial updates to the client rather than the entire tree every time, but that does not need to be the first time we will expose the verification tree. It's merely an low-level optimization that we need to carefully consider. This optimization would first consist in creating and exposing a new API to send partial verification diagnostics update. If the client does not define this API, then the server will just send the whole tree every time instead of updates.

Hence, for now and in order to display related ranges, we can keep it simple and just focus on the definition of the tree itself, and keep sending the verification tree to the client in the scope of this PR.
There is always an option to deactivate this altogether if one suffers from performance, but so far I haven't seen performance issue even on large files. The biggest problem currently is merely from the language server not canceling requests fast enough when it's stuck, which we should address separatedly.

MikaelMayer:
> Instead of talking about why it's not a problem to do X, let me take a step back and ask what is the benefit of doing X. Is there a reason to include the verification traces as part of an API that also contains other information? To me verification traces seems like a feature that stands on its own and can be implemented separately, so why mix it with other features?

Why mix it with other features:
- You said it, it's useful.
- It's not something to be advertised in a RELEASE_NOTES.txt for now.
- In general, a PR does not need to implement "one feature at a time", especially if these features are strongly related and/or they happen in a new context. We regularly ship multiple features together in one PR. For example:
  - https://github.com/dafny-lang/dafny/commit/deba09ef5bc7af039bb19f2fcf39284d942887cd
  - https://github.com/dafny-lang/dafny/commit/87b41dfde315712972125695e83ae99bcff076e9
  - https://github.com/dafny-lang/dafny/commit/6c9e62165e48b3545044a40f55152aa979124932 (continue with label could be a stand-alone feature compared to the introduction of continue)
- And really, the feature is called "verification diagnostics", the verification trace is only a field of it. I wouldn't call a field a "feature".

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,599 @@
...
+
+    /// <summary>
+    /// Returns per-line real-time diagnostic
+    /// </summary>
+    public LineVerificationStatus[] PerLineDiagnostic { get; init; }
+
+    static LineVerificationStatus[] RenderPerLineDiagnostics(
```
keyboardDrummer, line 62: I haven't been able to precisely understand what this method does, nor what the meaning of the output, the entries of the `LineVerificationStatus` enum, is. Would it be possible to include various example programs and their `RenderPerLineDiagnostics` output somewhere, either in test files or in diagnostics?

MikaelMayer: Could you perhaps comment on the tests first?
https://github.com/dafny-lang/dafny/pull/1942/files#diff-245d3daf9d22b82772587d8de3b8a6c8097a5dbeeb6f5b58d77e0f0f3a3d2a8dR25

I created a format that should be straightforward: it display various `LineVerificationStatus[]` vertically, from left to right, to simulate what would happen in the IDE, with 3-char icons.
The same string is used both to define the program (after extracting all expected `LineVerificationStatus[]`), to perform some edits (using the ``//Next:` directive), and to test the successive `LineVerificationStatus[]` (after removing duplicates).

Let me know how if this helps.

Source/DafnyLanguageServer.Test/Diagnostics/SimpleLinearVerificationDiagnosticTester.cs
```diff
@@ -0,0 +1,110 @@
...
+  private const int MaxTestExecutionTimeMs = 10000;
+
+  // To add a new test, just call VerifyTrace on a given program,
+  // the test will fail and give the correct output that can be use for the test
+  // Add '//Next<n>:' to edit a line multiple times
+
+  [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+  public async Task EnsureVerificationDiagnosticsAreWorking() {
+    await VerifyTrace(@"
+ .  |  |  |  I  I  |  | :predicate Ok() {
```
keyboardDrummer, line 26: predicate `Ok` doesn't have any associated verification condition right? How come it gets any icons, and also why do its icons change?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
keyboardDrummer: > Let me know how if this helps.

It's great! Really cool testing format. This is exactly what I was looking for. Sorry I must not have looked well before.

Source/DafnyLanguageServer.Test/Diagnostics/SimpleLinearVerificationDiagnosticTester.cs
MikaelMayer: Here is the story. I build the verification tree by first going through the program, modules, and extract methods, subset types, and field declarations. For each of them either I take a previous verification tree if it was there, or I create a new verification tree. In either case, I set their "StatusCurrent" to "Obsolete", (either ` . ` if it's the first time or ` I ` if it was verified before)

On receiving all the implementations that are going to be verified, I observed that any of these trees without an implementation can be marked as "verified" immediately, that's why it does not go through the "CurrentStatus.Verifying". Now, let me answer your questions

> predicate Ok doesn't have any associated verification condition right? 

Yes that's right, so it's why it does not get the "verifying" state.

> How come it gets any icons ?

The Dafny pipeline verified that there is nothing to do and everything works well, without the need to go through the Verifer. It's different from "Nothing" that says the verifier doesn't know yet the state of that line.

> Why do its icons change?

Until we get the verifier's information that there is no implementation, we cannot know whether the function is verified or not in general. Moreover, when the functions are edited, they have 

Now, here is a possible optimization I want to do in the future but not this PR, which is to compute the call graph and if the method was not changed and is not part of another method's call graph that was modified, we can immediately ensure its state is currently verified and, besides not even send it to Boogie, have no intermediate state in the gutter.

Source/DafnyLanguageServer.Test/Diagnostics/SimpleLinearVerificationDiagnosticTester.cs
```diff
@@ -0,0 +1,110 @@
...
+ .  |  |  |  I  I  |  | :  true
+ .  |  |  |  I  I  |  | :}
+    |  |  |  I  I  |  | :
+ .  S [S][ ][I][I][S] | :method Test(x: bool) returns (i: int)
+ .  S [=][=][-][-][~] | :   ensures i == 2
+ .  S [S][ ][I][I][S] | :{
+ .  S [S][ ][I][I][S] | :  if x {
+ .  S [S][ ][I][I][S] | :    i := 2;
+ .  S [=][=][-][-][~] | :  } else {
```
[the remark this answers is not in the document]
MikaelMayer: Yes it is useful, and here is the subtle difference.
- Column 6 marks that verification is obsolete. meaning Dafny acknowledge its verification status is outdated. Not only it's useful in "verify on save", to indicate that the user will need to save to see more up-to-date verification information, but it's also useful in "verify on change", because it indicates that the verifier did not start handling this range yet.
- Column 7 marks that there is a process currently working on verifying the range, so the "verification time" can be grasped by the time between column 7 and column 8, which is what users care the most.
If we had only one column, for very long files it would feel like some methods are extremely slow to verify, while others verify very fast, and it does not matches both the mental model (users know that methods are verified one by one) nor the underlying verification model (we split into implementations and verify them one by one, depending on the number of available cores)

In practice, once [this](https://github.com/dafny-lang/dafny/pull/1944) will be merged, the column 6 will not be visible because verification will start immediately.

Source/DafnyLanguageServer.Test/Diagnostics/SimpleLinearVerificationDiagnosticTester.cs
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+  public async Task EnsureItWorksForPostconditionsRelatedOutside() {
+    await VerifyTrace(@"
+ .  |  |  | :predicate F(i: int) {
+ .  |  |  | :  false // Should not be highlighted in gutter.
+ .  |  |  | :}
+    |  |  | :
+ .  S [S][ ]:method H()
```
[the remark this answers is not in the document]
MikaelMayer: ` S ` means that it's checking the method for the first time.
`[S]` means that it's still checking the method but now there is an error in the context of the method.  This is consecutive to a `ReportSplitResult` notification from Boogie.
This is displayed before the method knows there is nothing more to verify, because there could be multiple splits for the same implementation, and multiple implementations for the same method. Hence, in practice, I detect if the method is done verifying only in the notification `ReportEndVerifyImplementation`.
`[ ]` means that verification is over and there is no error on this line but somewhere else in that context.

Source/DafnyLanguageServer.Test/Diagnostics/SimpleLinearVerificationDiagnosticTester.cs
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+
+  [TestMethod/*, Timeout(MaxTestExecutionTimeMs)*/]
+  public async Task TopLevelConstantsHaveToBeVerifiedAlso() {
+    await VerifyTrace(@"
+    |  |  | :// The following should trigger only one error
+ .  |  |  | :ghost const a := [1, 2];
+    |  |  | :
+ .  S [~][=]:ghost const b := a[-1];");
```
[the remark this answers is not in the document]
MikaelMayer: For the same reason as https://github.com/dafny-lang/dafny/pull/1942#discussion_r848477354, 
First, the error is found, so the line displays "error" (in error context). `[=]`, but since the implementation is not done verifying, it show `[~]`, although arguably there cannot be any more feedback there (except if there were another assertion verifying)
In practice, the `[~]` is barely visible.

keyboardDrummer:
Personally I find that the API makes more choices about the visuals than I'd like.

It seems that he gutter icons combine information from different dimensions: status of verification pipeline, errorness, reachability of assertions. Personally I have trouble understanding all the combinations, and I think I'd prefer to see these dimensions separately. For example, I wouldn't mind having a single icon show the status of the verification pipeline, possibly on the method header. For the errorness, I don't understand why we need more information than what the diagnostics currently provide. Is that because for postcondition violations, the violated postcondition is in a related location and that doesn't get a red underline by default? For the reachability dimensions I don't understand the use-case well yet, but I guess that's on me.

Would it be possible to keep the different dimensions of information separate in the API and leave the precise choice of how to visualise things to the IDE ?

Something like:

Status of verification pipeline:
```c#
  public record FileVerificationStatus(
    DocumentUri Uri,
    int? Version,
    IReadOnlyList<MethodVerificationStatus> Methods) : IRequest;

  public record MethodVerificationStatus(Location Name, Location Body, VerificationStatus Status);
```

Reachability of assertions:
```c#
  public record FileAssertionInformation(
    DocumentUri Uri,
    int? Version,
    IReadOnlyList<AssertionInformation> Assertions) : IRequest;

  public record AssertionInformation(Location Body, bool WasVerified);
```

For the errorness, I guess that the information contained in `ImmediatelyRelatedRanges` is already available in the IDE through diagnostics.

For the counterexample trace, I would prefer a pull based API where the IDE has to request the information before it gets it.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,599 @@
...
+    AssertionFailed = 400,
+    // For lines containing resolution or parse errors
+    ResolutionError = 500
+  }
+
+  /// <summary>
+  /// A verification tree is an abstraction over the code to represent the verification
+  /// status of a region of the document, useful for IDE verification inspection.
+  /// A verification tree can contain other child trees.
+  /// It can currently be rendered linearly, e.g. for gutter display, or used as a tree in a test-like display.
```
keyboardDrummer, line 182: I'm unable to understand what this tree is doing. I feel like it's supposed to provide information for many different features. I'd like to have separate APIs for different features and to be able to verify that each API sends the minimal amount of required information, but with this huge record that also contains a lot of logic I find it impossible to verify that.

MikaelMayer:
> Personally I find that the API makes more choices about the visuals than I'd like.

Thanks for expressing your personal concern. I'll do my best to address it.

> It seems that he gutter icons combine information from different dimensions: status of verification pipeline, errorness, reachability of assertions.

Errorness and reachability of assertions are the same dimension: Error, Verified, or not run.

> Personally I have trouble understanding all the combinations, and I think I'd prefer to see these dimensions separately. For example, I wouldn't mind having a single icon show the status of the verification pipeline, possibly on the method header.

The icons on the gutter make it possible to see the verification status of the entire method only with the peripheral vision, which is less constraining than the focus that would be required by a single icon on the method header, icon which is almost always more far away than the gutter (and not viewing horizontally).

> For the errorness, I don't understand why we need more information than what the diagnostics currently provide. Is that because for postcondition violations, the violated postcondition is in a related location and that doesn't get a red underline by default?

Yes, that's one point, violated postconditions don't get a red underline by default. And we have no interest in duplicating diagnostics (that would be required to get underlines on two different things).
But moreover, and this was even more important, since diagnostics were published at the very end, all these errors were not available until verification was finished, which was extremely frustrating. This might have changed with your recent incremental diagnostic publication PR, but I haven't checked yet.

> For the reachability dimensions I don't understand the use-case well yet, but I guess that's on me.

The way the verifier works is that it will first figure out if an assertion fails, and it will normally be the first one. Then it will assume it, and figure out if other assertions fail, up to 5 times. Hence, if there are less than 5 errors, we can put a green check on all remaining assertions, because they are verified (conditionally to all assertions before them, but I assume users know that). If not, I don't display anything (to represent that the assertion wasn't even checked). Hence, this is the same dimension as the errorness.
 
> Would it be possible to keep the different dimensions of information separate in the API and leave the precise choice of how to visualise things to the IDE ?
> 
> Something like:
> 
> Status of verification pipeline:
> 
> ```cs
>   public record FileVerificationStatus(
>     DocumentUri Uri,
>     int? Version,
>     IReadOnlyList<MethodVerificationStatus> Methods) : IRequest;
> 
>   public record MethodVerificationStatus(Location Name, Location Body, VerificationStatus Status);
> ```
> 
> Reachability of assertions:
> 
> ```cs
>   public record FileAssertionInformation(
>     DocumentUri Uri,
>     int? Version,
>     IReadOnlyList<AssertionInformation> Assertions) : IRequest;
> 
>   public record AssertionInformation(Location Body, bool WasVerified);
> ```

There is also the notion of assertion batch that I want to expose, so the format above seems limited. Did you have in mind two send two records?
Also, since down the road we want the tests to be triggerables, we also need the relationship between methods and assertions, so I don't think we should put them in two different buckets from an API's perspective.

> For the errorness, I guess that the information contained in `ImmediatelyRelatedRanges` is already available in the IDE through diagnostics.

Yes, indirectly. Making it available directly makes it easier for IDE extensions not to reimplement the wheel.

> For the counterexample trace, I would prefer a pull based API where the IDE has to request the information before it gets it.

Counterexample trace is another topic to be discussed in this other PR:
https://github.com/dafny-lang/dafny/pull/1977

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -0,0 +1,590 @@
...
+  }
+
+  /// <summary>
+  /// A verification tree is an abstraction over the code to represent the verification
+  /// status of a region of the document, useful for IDE verification inspection.
+  /// A verification tree can contain other child trees.
+  /// It can currently be rendered linearly, e.g. for gutter display, or used as a tree in a test-like display. 
+  /// The verification status consists of two orthogonal concepts:
+  /// - StatusVerification: Nothing (initial), Error, Verified, or Inconclusive
+  /// - StatusCurrent: Current (Up-to-date), Obsolete (outdated), and Verifying (as notified by the verifier)
```
robin-aws, line 176: I don't think these are orthogonal, especially if you agree to drop Obsolete. You just end up with `Nothing (initial), Verifying, Error, Verified, or Inconclusive`

robin-aws:
Working on catching up and just offering high-level thoughts for now.

I agree with @keyboardDrummer's difficulty in keeping track of all the different verification statuses, but I don't think it's beneficial to separate dimensions more because they are causally linked. You can't simultaneously have a "currently verifying" pipeline state and a "failed" verification status on the same line.

Instead to simplify things, I'd suggest not keeping around obsolete data. The important thing to communicate is whether verification obligations exist (i.e. the difference between `Nothing` and `Scheduled` or any other state), and where they are currently being determined or there is an answer. I don't think there's value in knowing that a line WAS correct/invalid previously if it may or may not change once verification catches up (either automatically or once you save).

I think having a more generic tree of data works better here to a point. Methods are not THAT significant a type here, given that each gets two assertion batches anyway, and there are lots of assertions that don't live in methods. Splitting up per file/document certainly makes sense though.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
keyboardDrummer: With comments from our call:

```
public enum LineVerificationStatus {
    // Default value for every line, before the renderer figures it out.
    // Used to separate lines that are not correctly verified.
    Nothing = 0,
    // For first-time computation not actively computing but soon. Synonym of "obsolete"
    // (scheduledComputation)
    // Either stale or queued. No previous result known.
    Scheduled = 1,
    // For first-time computations, actively computing
    Verifying = 2,

    // Context previously correctly verified, now stale or queued
    VerifiedObsolete = 201,
    // The context is previously correctly verified, and is not running.
    VerifiedVerifying = 202,

    // Either the context is verified, or there's an assertion in this line which verified and the context didn't error.
    // Or to context blank lines between contexts to make the green flow
    Verified = 200,

    // 300-302: For trees containing children with errors (e.g. functions, methods, fields, subset types)

    // The context previously had errors, but is now stale or queued.
    ErrorContextObsolete = 301,
    // The context previously had errors, but is now running.
    ErrorContextVerifying = 302,
    // The context has an error and no more work needs to be done.
    ErrorContextCompleted = 300,

    // 350-352: For individual assertions in error contexts

    // Note: take into account two assertions on one line.
    // Same as ErrorContextObsolete, but also the line contains no error diagnostics AND at least one assertion that passed in at least one batch.
    AssertionVerifiedInErrorContextObsolete = 351,
    // Same as ErrorContextVerifying, but also the line contains no error diagnostics AND at least one assertion that passed in at least one batch.
    AssertionVerifiedInErrorContextVerifying = 352,
    // Same as ErrorContextCompleted, but also the line contains no error diagnostics AND at least one assertion that passed in at least one batch.
    AssertionVerifiedInErrorContextCompleted = 350,
    // For specific lines which have errors on it. They take over verified assertions

    // Line previously contained at least one failed assertion, context is stale or queued.
    AssertionFailedObsolete = 401,
    // Line previously contained at least one failed assertion, context is running.
    AssertionFailedVerifying = 402,

    // Line contains at least one failed assertion, context can be in any state.
    AssertionFailed = 400,

    // For lines containing resolution or parse errors
    ResolutionError = 500
  }
```

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -147,37 +161,468 @@ bool loadCanceled
         TaskCreationOptions.None, LargeStackScheduler).Unwrap();
     }
 
+    // Called only in the case there is a parsing or resolution error on the document
+    public void PublishVerificationDiagnostics(DafnyDocument document) {
```
keyboardDrummer, line 165: Why not inline this method?

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -26,11 +29,39 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
...
+      var errors = currentDiagnostics
+        .Concat(document.OldVerificationDiagnostics)
+        .Where(x => x.Severity == DiagnosticSeverity.Error)
+        .ToArray();
+      var linesCount = Regex.Matches(document.Text.Text, "\r?\n").Count + 1;
+      var verificationStatusGutter = new VerificationStatusGutter(
+        document.Uri,
+        document.Version,
+        document.VerificationTree.Children.Select(child => child.GetCopyForNotification()).ToArray(),
+        errors,
```
keyboardDrummer, line 59: Why do you include errors here? Isn't publishing LSP diagnostics what `PublishDiagnosticsParams` is for?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -166,11 +171,17 @@ IRelocator relocator
...
           SerializedCounterExamples = null,
           LoadCanceled = true,
-          OldVerificationDiagnostics = migratedVerificationDiagnotics
+          OldVerificationDiagnostics = migratedVerificationDiagnotics,
+          VerificationTree = migratedVerificationTree
         };
       }
     }
 
+    private DafnyDocument WithRealtimeDiagnosticsPublished(DafnyDocument document) {
```
keyboardDrummer, line 180: I initially thought this method performs a mutating operation on `DafnyDocument`, since most `With` methods return a value that's different from the input value. I'd prefer inlining this method.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -147,37 +161,468 @@ bool loadCanceled
...
       public void ReportProgress(string message) {
-        publisher.SendStatusNotification(document, CompilationStatus.VerificationStarted, message);
+        publisher.SendStatusNotification(document.Text, CompilationStatus.VerificationStarted, message);
+      }
+
+      /// <summary>
+      /// Fills up the document with empty verification diagnostics, one for each top-level declarations
+      /// Possibly migrates previous diagnostics
+      /// </summary>
+      public void RecomputeVerificationTree() {
```
keyboardDrummer, line 235: Could you put the verification tree computation in a separate file?

keyboardDrummer:
> Working on catching up and just offering high-level thoughts for now.
> 
> I agree with @keyboardDrummer's difficulty in keeping track of all the different verification statuses, but I don't think it's beneficial to separate dimensions more because they are causally linked. You can't simultaneously have a "currently verifying" pipeline state and a "failed" verification status on the same line.
> 
> Instead to simplify things, I'd suggest not keeping around obsolete data. The important thing to communicate is whether verification obligations exist (i.e. the difference between `Nothing` and `Scheduled` or any other state), and where they are currently being determined or there is an answer. I don't think there's value in knowing that a line WAS correct/invalid previously if it may or may not change once verification catches up (either automatically or once you save).
> 
> I think having a more generic tree of data works better here to a point. Methods are not THAT significant a type here, given that each gets two assertion batches anyway, and there are lots of assertions that don't live in methods. Splitting up per file/document certainly makes sense though.

I've discussed this PR with @MikaelMayer and @cpitclaudel and we decided to keep the gutter icon enumeration as is, so that we can get customer feedback on this feature. 

We're accepting the risk that we might have to change the API in the future in a backwards incompatible way. Because this is a push API, we can update the API by adding a version number `dafny/verification/status/gutterv2`, and that means existing IDEs will simply not pick it up anymore while newer IDEs that only listen to `dafny/verification/status/gutterv2` will simply not pick up the `dafny/verification/status/gutter` from older language servers.

@cpitclaudel expressed performance concerns with the added LSP API in the context of Emacs, so we might also change the API later to address that.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
MikaelMayer: > I don't think there's value in knowing that a line WAS correct/invalid previously if it may or may not change once verification catches up (either automatically or once you save).

The problem of removing the previous verification result is that any given line will transition is that, during the time it verifies (which are a few seconds), what you suggests really means hiding the previous state during the time Dafny verifies. Why would you hide the previous state way before you have some more up-to-date information? Users trying to fix an error might be thinking in parallel about the code and the previous verification result in other places which they believe would not have been impacted. Removing them this information would result in a frustrating experience.

There are precedents about that: We decided to still display previous verification diagnostics (migrate them) although there might be parse/resolution errors. Moreover, in general, diagnostics are "updated" rather than "removed" and then "re-added" a few seconds later.
So, in short, it's something that users would expect and it's desirable.

To illustrate my point, let's consider a CALC statemen with four statements, where Q2 is failing but Q3 and Q4 are succeeding. A user makes an edit on Q2, which makes it verify, but now Q3 does not verify anymore. The trace under this PR would be the following:
```dafny
[ ][I][S][ ]: calc {
[ ][I][S][ ]:   Q1;
[=][-][~][O]:   Q2;
[O][o][Q][=]:   Q3;
[O][o][Q][O]:   Q4;
[ ][I][S][ ]: }
```

With your approach of not making statuses orthogonal, I don't think how it would look better than this:
```dafny
[ ] I  S [ ]: calc {
[ ] I  S [ ]:   Q1;
[=] I  S [O]:   Q2;
[O] I  S [=]:   Q3;
[O] I  S [O]:   Q4;
[ ] I  S [ ]: }
```

Here are notable differences in the two approaches:
- In the first version, the switching state from "error" to "verified" is instant on the fourth update (letting the user know the error was first obsolete, and then recomputing), whereas in the second version, there is an intermediate state that completely hides all the previous information.
- If I did not pay attention about the verification result between Q3 and Q4 (which can be farther away) because I was focused on Q1 to Q2, during the whole time the verifier runs (a few seconds), I would be blind about having reasons to believe that Q3 is still equals to Q4.
- A few seconds of completely removing the previous information would have the effect to remove alignment information for the eye, so instead of seeing if an error has been fixed and the user can move on, the eye will have to start its course again from the line being edited to the left where the gutter is.
- During the UX development with a UX designer, she agreed with me that it's better to minimize the frequency of full visual changes made to the gutter, because it should be supportive for the eye, not distract it.
- Ideally, the eye should be able to stay on the code and only use peripheral vision to know the verification status. With this PR, the eye will be able to see an error "going down" with the peripheral vision, and since the user previously saw that this error was on the line being edited, they won't even need to look from right to left to see if the line has an error or not. With your alternative, after few seconds during which that information disappears, the eye will not be able to use peripheral vision to determine if a line verifies or not until it does the actual movement from right to left.

I can go on and on, and also explain you why displaying the error context is important as well, rather than just repeating the verification status on every line. Just ask me.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -147,37 +161,468 @@ bool loadCanceled
...
+          }
+          result.Add(verificationTree);
+        }
+
+        var documentFilePath = document.GetFilePath();
+        foreach (var module in document.Program.Modules()) {
+          foreach (var topLevelDecl in module.TopLevelDecls) {
+            if (topLevelDecl is TopLevelDeclWithMembers topLevelDeclWithMembers) {
+              foreach (var member in topLevelDeclWithMembers.Members) {
+                if (member.tok.filename != documentFilePath) {
```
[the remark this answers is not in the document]
MikaelMayer: I think, when we include a module from another file.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -147,37 +161,468 @@ bool loadCanceled
...
+                  continue;
+                }
+                if (member is Field) {
+                  if (member.BodyEndTok.line == 0) {
```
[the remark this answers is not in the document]
MikaelMayer: Looking deeper, it currently means that there is no body, so the constant does not have a value. See here:
https://github.com/dafny-lang/dafny/blob/verification-diagnostics-core/Source/Dafny/Dafny.atg#L1374
However, the goal of this if was to prevent the `GetLspRange` next line to raise an exception

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -147,37 +161,468 @@ bool loadCanceled
...
+                    continue; // Nothing to verify
+                  }
+                  var diagnosticRange = member.tok.GetLspRange(member.BodyEndTok);
+                  var diagnostic = new TopLevelDeclMemberVerificationTree(
```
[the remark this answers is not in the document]
MikaelMayer: Good catch. It was based on the old name of VerificationTree, which was VerificationNodeDiagnostics. I renamed all "diagnostics" to "verificationTree"

Source/DafnyLanguageServer.Test/Util/AssertWithDiff.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Text;
+using DiffPlex.DiffBuilder;
+using DiffPlex.DiffBuilder.Model;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Util;
+
+public class AssertWithDiff {
```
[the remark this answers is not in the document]
MikaelMayer: These two projects do not have the same testing framework (one throws an XUnitException, the other one is ` Microsoft.VisualStudio.TestTools.UnitTesting`).
Alternatively, I could refactor both codes, but I'm worried that it could cause trouble to have two different testing frameworks in the same project.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -143,20 +142,26 @@ IRelocator relocator
       var migratedVerificationDiagnotics =
         relocator.RelocateDiagnostics(oldVerificationDiagnostics, documentChange, CancellationToken.None);
       logger.LogDebug($"Migrated {oldVerificationDiagnostics.Count} diagnostics into {migratedVerificationDiagnotics.Count} diagnostics.");
+      var oldVerificationTree =
```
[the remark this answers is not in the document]
MikaelMayer: > Why set the old verification tree to obsolete if you're filling the documentation with the migrated tree? I would expect the obsolete setting to happen inside `RelocateVerificationTree`

Good point.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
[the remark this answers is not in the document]
keyboardDrummer: Would it be possible to replace `if (member is Field) { if (member.BodyEndTok.line == 0).. ` with `if (member is ConstantField)` ?

Alternatively, you could extract this into a variable to explain the condition, like `var constantHasNoBody = member.BodyEndTok.line == 0`.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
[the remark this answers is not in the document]
keyboardDrummer: Could you document that by extracting the condition into a variable: `var memberWasIncluded = member.tok.filename != documentFilePath` ?

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -26,11 +29,39 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
...
+        linesCount,
```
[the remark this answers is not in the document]
keyboardDrummer: I see. I would prefer it if `VerificationStatusGutter` was a record with a primary constructor, to make it more clear how much data it contains. However, I don't have good arguments for this so feel to leave it if you prefer that ^,^

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -58,6 +73,22 @@ CancellationToken cancellationToken
...
+      private List<Position> MigratePositions(List<Position> originalRanges, TextDocumentContentChangeEvent change) {
+        if (change.Range == null) {
+          return new List<Position> { };
+        }
+
+        return originalRanges.SelectMany(position => MigratePosition(change.Range, change.Text, position)).ToList();
+      }
+
+      private IEnumerable<Position> MigratePosition(Range changeRange, string changeText, Position position) {
+        var afterChangeEndOffset = GetPositionAtEndOfAppliedChange(changeRange, changeText);
```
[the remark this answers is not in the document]
MikaelMayer: > GetPositionAtEndOfAppliedChange is potentially expensive, since it scales with the size of the change and a customer could paste in a large file, so I think we should ensure it only gets called once per change. MigratePositions calls it for each position. Maybe we can compute GetPositionAtEndOfAppliedChange in the constructor of ChangeProcessor ?

Done.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -143,6 +174,11 @@ Position afterChangeEndOffset
         return new Position(changeStart.Line + changeEof.Line, characterOffset);
       }
 
+      private static IEnumerable<Range> MigratePositions(Range[] rangesToMigrate, Range changeRange, Position afterChangeEndOffset) {
```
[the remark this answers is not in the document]
MikaelMayer: > Is the method name intentional?

The method was unused, looks like a leftover. removed.

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -26,11 +29,39 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
...
+      }
+      // Document.GetDiagnostics() returns not only resolution errors, but previous verification errors
+      var currentDiagnostics =
+        document.Errors.GetDiagnostics(document.GetFilePath())
+          .Where(x => x.Severity == DiagnosticSeverity.Error).ToList();
+      var errors = currentDiagnostics
+        .Concat(document.OldVerificationDiagnostics)
+        .Where(x => x.Severity == DiagnosticSeverity.Error)
+        .ToArray();
+      var linesCount = Regex.Matches(document.Text.Text, "\r?\n").Count + 1;
```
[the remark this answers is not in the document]
MikaelMayer: >I think that we want to support files with thousands of lines at some point, so we should refrain from doing operations that scale with the size of the file such as Regex.Matches(document.Text.Text, "\r?\n"). I think eventually we'll have an incremental parser that doesn't reparse the whole file on each change.

I mitigated the problem above by adding a property `LinesCount` to `DafnyDocument` which retrieves the number of lines from the range of the top verification tree, so that it is not computed on every notification, only when building the tree for the first time.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
MikaelMayer: Because it is used in two places in TextDocumentLoader.

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
MikaelMayer: These errors are part of the constructor, not of the information sent to the client. I use these errors to render resolution errors on the gutter, if any.

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -26,11 +29,39 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
...
+      var verificationStatusGutter = new VerificationStatusGutter(
+        document.Uri,
+        document.Version,
+        document.VerificationTree.Children.Select(child => child.GetCopyForNotification()).ToArray(),
+        errors,
+        linesCount,
+        document.VerificationPass != null,
```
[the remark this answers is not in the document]
MikaelMayer: We can arguably find better names, but:

VerificationPass is
* null if verification hasn't started
* false if verification won't start
* true if verification starts

It is used in the constructor of VerificationStatusGutter to determine if verification started. That way, we don't display some green "verified" gutter on the entire file if there is no resolution error, nothing to verify, but verification isn't enabled.

ResolutionSucceeded is
* null if resolution didn't finish
* true if there are no errors after resolution
* false if there are errors after resolution

It is used to know if the current diagnostics start with current resolution errors, and if so, use them to display them in the gutter.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: But it only contains a single call right? `diagnosticPublisher.PublishVerificationDiagnostics(document);` What's the use of wrapping that in another method?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -32,6 +35,29 @@ public record DafnyDocument(
...
+    /// True is the resolution succeeded, false if resolution failed
+    /// <c>null</c> If the verification did not start (e.g. because of resolution errors)
+    /// </summary>
+    public bool? ResolutionSucceeded { get; set; } = null;
+
+    /// <summary>
+    /// True is the verification pass went through, false if it failed because of verification errors
+    /// <c>null</c> If the verification did not start (e.g. because of resolution errors or verification deactivated)
+    /// </summary>
+    public bool? VerificationPass { get; set; } = null;
```
keyboardDrummer, line 49: Consider renaming to `VerificationPassed` or `Verified`

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -26,11 +29,39 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
...
+      var diagnosticParameters = new PublishDiagnosticsParams {
         Uri = document.Uri,
         Version = document.Version,
         Diagnostics = GetDiagnostics(document).ToArray(),
-      });
+      };
+      languageServer.TextDocument.PublishDiagnostics(diagnosticParameters);
+    }
+
+    public void PublishVerificationDiagnostics(DafnyDocument document) {
```
keyboardDrummer, line 40: Nitpick: consider moving this to `VerificationProgressReporter` since it is called there.

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -32,6 +35,29 @@ public record DafnyDocument(
     /// </summary>
     public string? SerializedCounterExamples { get; init; }
 
+
+    /// <summary>
+    /// True is the resolution succeeded, false if resolution failed
+    /// <c>null</c> If the verification did not start (e.g. because of resolution errors)
+    /// </summary>
+    public bool? ResolutionSucceeded { get; set; } = null;
```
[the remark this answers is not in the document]
MikaelMayer: `ResolutionSuceeded` was never using "set", only initialization, so I replaced `set` by `init` to make it immutable.
I removed `VerificationPass` because I could make it an argument to the methods that were using it.
However, for `VerificationTree`, I don't think it's appropriate. I tried to do my best so that VerificationTree scales with the file, so mutability is the best option I had. Moreover, this [follow-up PR](https://github.com/dafny-lang/dafny/pull/1946/files) relies on accessing the verification tree in real-time, and it only has access to the current `DafnyDocument`.
Mutability can be a burden, but in this case, I think it's appropriate to keep it.

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
MikaelMayer: I renamed it to `verificationStarted` and removed it from `DafnyDocument`

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
MikaelMayer: I considered it, but since this file is the endpoint of all publications (notifications and diagnostics), I think it's better there than in `VerificationProgressReporter`.

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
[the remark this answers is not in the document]
keyboardDrummer: We can keep it like this and see for now, as long as we try to keep DafnyDocument immutable :-)

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
MikaelMayer: Ok, I inlined this method.

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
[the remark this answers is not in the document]
MikaelMayer: If it was a record, indeed, every initialization variable would be public, so that's why I did not use a record.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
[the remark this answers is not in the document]
MikaelMayer: Since that ConstantFields can have bodies, I'll go the second route.

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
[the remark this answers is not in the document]
keyboardDrummer: Note that you can also add non-primary constructors to a record with a primary contructor:

```
record Computed(int TotalLength) {
  Computed(int bodyHeight, int shoeHeight) : this(bodyHeight + shoeHeight) { }
}
```

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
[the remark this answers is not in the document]
keyboardDrummer: > Since that ConstantFields can have bodies, I'll go the second route.

My impression was that only ConstantFields have bodies and they always have them, so if you check for `ConstantField1` you no longer need another check. Did I misread?

keyboardDrummer (approved):

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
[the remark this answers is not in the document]
MikaelMayer: You can declare Constant fields without a body. it's just that the compiler won't accept it, but it works for verification purposes, I think.

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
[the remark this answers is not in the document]
MikaelMayer: Or, as a viable alternative, add a static method to `GutterIcons` that will do this computation and use the primary constructor, right?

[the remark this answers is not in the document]
keyboardDrummer: That works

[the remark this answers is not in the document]
MikaelMayer: > - Or, as a viable alternative, add a static method to GutterIcons that will do this computation and use the primary constructor, right?
> - That works

Implemented here
https://github.com/dafny-lang/dafny/pull/2143

PR #1943, opened 2022-03-29
Added failing stressing verification test for language server

### LINKED ISSUE #1913: Language Server crashing under heavy load

MikaelMayer:
I am adding a stressing verification test for the language server, the test passes, however, it always remains a dangling z3 process that I have to manually kill.
@keyboardDrummer  Would you know what is wrong here?

This PR should fix #1913 when it is ready.


MikaelMayer:
Closed, as not maintained.

PR #1945, opened 2022-03-29
feat: Improve error message when compiler is used without initialization

cpitclaudel:
Closes GH-1932.

If I understand what's going on correctly, #1932 isn't a bug: the API isn't (wasn't?) being used correctly, and it's surfacing that incorrect usage.  The right fix is for the client code to call `ApplyDefaultOptions()`.

I've changed from `null` to a placeholder so that it prints a more helpful message.  WDYT @MikaelMayer ?

### DIFF
--- a/Source/Dafny/DafnyOptions.cs
+++ b/Source/Dafny/DafnyOptions.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Dafny {
       Prune = true;
       NormalizeNames = true;
       EmitDebugInformation = false;
+      Compiler = new PlaceholderCompiler();
     }
 
     public override string VersionNumber {
@@ -582,7 +583,9 @@ namespace Microsoft.Dafny {
         XmlSink = new Bpl.XmlSink(this, BoogieXmlFilename);
       }
 
-      Compiler ??= new CsharpCompiler();
+      if (Compiler == null || Compiler is PlaceholderCompiler) {
+        Compiler = new CsharpCompiler();
+      }
 
       // expand macros in filenames, now that LogPrefix is fully determined
       ExpandFilename(DafnyPrelude, x => DafnyPrelude = x, LogPrefix, FileTimestamp);
--- a/Source/Dafny/Plugins/Compiler.cs
+++ b/Source/Dafny/Plugins/Compiler.cs
@@ -140,3 +140,42 @@ public abstract class Compiler {
   public abstract bool RunTargetProgram(string dafnyProgramName, string targetProgramText, string callToMain, string pathsFilename,
     ReadOnlyCollection<string> otherFileNames, object compilationResult, TextWriter outputWriter);
 }
+
+/// <summary>
+/// A placeholder class used to initialize <c>Dafny.Options.O.Compiler</c>.
+/// </summary>
+class PlaceholderCompiler : Compiler {
+  public override IReadOnlySet<string> SupportedExtensions => new HashSet<string>();
+
+  public override string TargetLanguage => "unset";
+  public override string TargetExtension => "unset";
+
+  private Exception UnimplementedExeception() {
+    return new NotImplementedException("DafnyOptions.O.Compiler unset.\n" +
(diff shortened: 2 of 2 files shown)


Source/Dafny/DafnyOptions.cs
```diff
@@ -582,7 +583,9 @@ public enum IncludesModes {
         XmlSink = new Bpl.XmlSink(this, BoogieXmlFilename);
       }
 
-      Compiler ??= new CsharpCompiler();
+      if (Compiler == null || Compiler is PlaceholderCompiler) {
```
MikaelMayer, line 586: Compiler will never be null, right? Since you initialize it in the constructor. However, I would keep it that way in case someone else sets this field to null manually just to get the default.

Source/Dafny/Plugins/Compiler.cs
```diff
@@ -140,3 +140,42 @@ public abstract class Compiler {
...
+/// A placeholder class used to initialize <c>Dafny.Options.O.Compiler</c>.
+/// </summary>
+class PlaceholderCompiler : Compiler {
+  public override IReadOnlySet<string> SupportedExtensions => new HashSet<string>();
+
+  public override string TargetLanguage => "unset";
+  public override string TargetExtension => "unset";
+
+  private Exception UnimplementedExeception() {
+    return new NotImplementedException("DafnyOptions.O.Compiler unset.\n" +
```
MikaelMayer, line 154: I see now how an error is thrown now that is much more meaningful, but it's not backward compatible.

What about _extending_ the `CSharpCompiler`  in this placeholder class, and printing a warning message while calling the base classes? At least, things couldn't break, and people would be notified dynamically.
This would be similar to having "deprecated" in public API.

MikaelMayer (commented):
It's a good move forward ! it inspires me to modify it slightly so that it is still backwards compatible.

Source/Dafny/Plugins/Compiler.cs
cpitclaudel: OK, that could work.  But why does backward compatibility matter here?  This is a broken usage of the API that trivial (and good!) to fix, no?

MikaelMayer: Sorry I did not answer your question before:

> the API isn't (wasn't?) being used correctly, and it's surfacing that incorrect usage. The right fix is for the client code to call ApplyDefaultOptions().

Previously, there was no need of `ApplyDefaultOptions()` to set up the compiler target to `c#`. It _was_ the default, see here:
https://github.com/dafny-lang/dafny/blob/dc410d680e90429dd3852340021a3a2a37ea974d/Source/Dafny/DafnyOptions.cs#L88

So it became an incorrect usage when the field `CompilationTarget` was replaced with `Compiler`

cpitclaudel: I understand this, but I don't think it's worth fixing beyond this patch (which I already think is a bit too much :/).  If you think it's worth it, I will do it.

MikaelMayer: I think it's worth it. Thanks: -)

Source/Dafny/DafnyOptions.cs
```diff
@@ -27,6 +27,7 @@ public DafnyOptions()
       Prune = true;
       NormalizeNames = true;
       EmitDebugInformation = false;
+      Compiler = new PlaceholderCompiler();
```
keyboardDrummer, line 30: Do we really need this? It seems like we need something like a `DafnyOptionsBuilder` that forces you to create properly initialised DafnyOptions, meaning `ApplyDefaultOptions`, which maybe should be called `ReadyOptionsForUse`, is called at the end.

cpitclaudel: I think we don't.  Let's just revert to the default of having C# out of the box and move on ^^

cpitclaudel:
Abandoning this in favor of #1933

PR #1946, opened 2022-03-29
feat: Verification diagnostics hover

MikaelMayer:
This PR adds assertion information when hovering them (without needing to change the VSCode plugin).

Hovering over f (empty body):
image

Hovering over f (one failing assertion), "assertion batch" points to [this link](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-assertion-batches).
image

Hovering over false, "Error" points to [this link](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-verification-debugging).
image

Hovering over a failing postcondition shows where the return path is failing
image

Hovering over `x == 2`
image

Hovering over `true`, I tried to put it in green but it does not work.
image

Hovering over `f` but this time there are 3 assertions. It indicates the longest batch and where it started:
image

- Hovering over a lemma very expensive to verify before it starts verifying
image

- Hovering over a lemma very expensive to verify before it finishes verifying
image

- Hovering over a lemma very expensive to verify, the warning icon points to [this link](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-verification-debugging-slow).
image

- Hovering over a lemma very expensive to verify with multiple assertion batches:
image

### Formal tests

One way to discover the effect of this PR is by visiting the first [hover test](https://github.com/dafny-lang/dafny/pull/1946/files#diff-ae2f566c523a0087efdc131112cb0100c8df46e673479c36a66ec5065a9e4d46R49)

### Implementation

Everything in this feature is in `DafnyHoverHandler.cs`, the method `Handle`
* First, instead of immediately returning an hover when a symbol is found, I first compute `symbolHoverContent` (for type information) and `diagnosticGoverContent` (for verification performance and insights). If the diagnostics are only about statistics of an entire method (not individual assertions), it should not be displayed if there is a symbol under the caret, to prevent distractions, and `areMethodStatistics` is set to true in that case.
```cs
      var hoverContent = areMethodStatistics && symbolHoverContent != null ? "" : (diagnosticHoverContent ?? "");
      hoverContent = symbolHoverContent != null ? hoverContent + (hoverContent != "" ? "  \n" : "") + symbolHoverContent : hoverContent;
```

* `diagnosticHoverContent` is assigned using the new method `GetDiagnosticsHover`
  * In this method, I am traversing the hierarchy of TopLevelDeclMemberVerificationTree, then I skip their children (which are implementations that users don't care about), and traverse instead the aggregated `node.AssertionBatches` which are of type `AssertionBatchVerificationTree`. I'm keeping track of the assertion batch index with `assertionBatchIndex` to be able to display it to users.
  * Since every node contains a `Range`, I can only traverse node that contain the hovering position, which is efficient. I can't do a binary search however because I don't require verification trees to be ordered by range.
  * I either give two kinds of hovering statistics: Individual assertions or per-method/subset type/field. If an individual assertion is hovered, the per-method statistics are not displayed.
    * For an individual assertions, I tolerate it's "hovered" also if the hover position is touching an `ImmediatelyRelatedRange` (i.e. the postcondition for return AssertCmd). `assertionNode.ImmediatelyRelatedRanges.Any(range => range.Contains(position))`
    * I then add all the statistics in the string `information`: The assertion number and total number of assertions in this batch (only if it's more than 1), the batch number over the total number of batches and its timing and resource count.
    * If the assertion is for a return statement and we are hovering the postcondition (which has normally no error associated with it), I display the position of the corresponding return (which contains more information about the error)
    * If there is a counter example for the current assertion, I use its `Description` to generate an even better message about its status. (i.e. `SetDescription(returnCounterexample.FailingReturn.Description);`)
    * At the end, I display "recently" if the status is obsolete, or "recently and verifying..." if the status of the current assertion is verifying (which is not implemented yet but soon)
  * If there is no information about a single assertion this PR checks if the hover position is on the first line of a method (after the method name, because that's the first token of a method 😕 ), and if so:
    * I display "_Verification not started yet_" if its status is "Obsolete" (not started yet)
    * I display "_Still verifying_" and the time elapsed if the status is "Verifying" (started to verify)
    * I display the top 3 most expensive assertion batch with their resource usage.


Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -0,0 +1,156 @@
...
+", CompilationStatus.VerificationFailed);
+      // When hovering the postcondition, it should display the position of the failing path
+      await AssertHoverMatches(documentItem, (2, 15),
+        @"assertion #1/2 of [batch](???) #1/1 checked in ???ms with ??? resource count:  
+`testFile0.dfy(6, 5): `*A postcondition might not hold on this return path.*"
+      );
+      // When hovering the failing path, it does not display the position of the failing postcondition
+      // because the IDE extension already does it.
+      await AssertHoverMatches(documentItem, (5, 4),
+        @"assertion #1/2 of [batch](???) #1/1 checked in ???ms with ??? resource count:
```
MikaelMayer, line 66: Question to reviewers: should I display `batch #1/1` when there is only one batch?
It's useful to have the word batch so that the link is available, but I could as well put the link on assertion and remove `of batch #1/1`. I currently favored consistency with other messages.

MikaelMayer (commented):
This PR is ready for review.

Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
```diff
@@ -123,7 +123,7 @@ IDiagnosticPublisher diagnosticPublisher
   /// </summary>
   /// <param name="implementations">The implementations to be verified</param>
   public void ReportImplementationsBeforeVerification(Implementation[] implementations) {
-    if (document.LoadCanceled || implementations.Length == 0) {
+    if (document.LoadCanceled) {
```
MikaelMayer, line 126: About this: It happenes that in the past somewhere, for some time, this function was called twice, once without any implementations, and the other one with all the implementations. This started to happen after I had merged once with master, so I had this guard to prevent this. However, it would not enable to consider files without really any implementation to verify.
This "bug" does not happen anymore (probably thanks to recent's refactoring with Boogie) so I remove this check.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +31,142 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
+      // If diagnostics are method diagnostics, we prioritize displaying the symbol information.
+      // This makes testing easier and less surprise for the user.
+      var hoverContent = areMethodStatistics && symbolHoverContent != null ? "" : (diagnosticHoverContent ?? "");
+      hoverContent = symbolHoverContent != null ? hoverContent + (hoverContent != "" ? "  \n" : "") + symbolHoverContent : hoverContent;
+      return CreateMarkdownHover(hoverContent);
+    }
+
+    private string? GetDiagnosticsHover(DafnyDocument document, Position position, out bool areMethodStatistics) {
+      areMethodStatistics = false;
+      foreach (var node in document.VerificationTree.Children.OfType<TopLevelDeclMemberVerificationTree>()) {
```
atomb, line 60: Eventually it might be possible to use a more efficient data structure for this, but that's probably only worth it if we find that this repeated iteration is too expensive.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +31,142 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
+        if (node.Range.Contains(position)) {
+          var assertionBatchCount = node.AssertionBatchCount;
+          var assertionBatchIndex = -1;
+          var information = "";
+          foreach (var assertionBatch in node.AssertionBatches) {
+            assertionBatchIndex += 1;
```
atomb, line 66: It might not always be the case that the assertion batches are listed in order. This should probably use `VCResult.vcNum`, piped through appropriately. The assertions in each batch should always be in the same order, though.

atomb (changes requested):
The only one of my comments that I think it's important to address before merging is the one about assertion batch numbering/ordering.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -405,14 +419,28 @@ Range Range
...
           }
         }
       }
 
       AssertionBatches = result;
     }
+
+    public AssertionBatchVerificationTree? GetLongestAssertionBatch() =>
+      !AssertionBatches.Any() ? null :
+      AssertionBatches.MaxBy(assertionBatch => assertionBatch.TimeSpent);
```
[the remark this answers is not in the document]
MikaelMayer: Only if I only wanted the time, but I want the entire assertion batch.

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -0,0 +1,172 @@
...
+    private async Task<CompilationStatus> WaitUntilDafnyFinishes(TextDocumentItem documentItem) {
+      CompilationStatusParams lastResult;
+      do {
+        lastResult = await notificationReceiver.AwaitNextNotificationAsync(CancellationToken);
+        Assert.AreEqual(documentItem.Uri, lastResult.Uri);
+        Assert.AreEqual(documentItem.Version, lastResult.Version);
+      } while (lastResult.Status != CompilationStatus.VerificationFailed &&
+               lastResult.Status != CompilationStatus.VerificationSucceeded &&
+               lastResult.Status != CompilationStatus.ParsingFailed &&
+               lastResult.Status != CompilationStatus.ResolutionFailed);
```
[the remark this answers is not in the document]
MikaelMayer: > Would it be worth having a separate definition for "done" that encapsulates these? It would make the code a little less messy, and would make it clear what the meaning of that big conjunction is.

Refactored, thanks for pointing this out.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: I did not know about  `vcNum`, that's a great insight ! I refactored the code to make use of it instead of a counter that is incrementing. It will even make the migration of assertion batch diagnostics easier down the road since this counter is probably more deterministic than mine.

keyboardDrummer:
For the UX, how would diagnostic hints (they show up as ... under the first bit of the range they cover) compare to hover? I'm wondering if hints will provide greater discoverability of this feature.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +36,220 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
-      return CreateHover(symbol, cancellationToken);
+
+      // If diagnostics are method diagnostics, we prioritize displaying the symbol information.
+      // This makes testing easier and less surprise for the user.
+      var hoverContent = areMethodStatistics && symbolHoverContent != null ? "" : (diagnosticHoverContent ?? "");
+      hoverContent = symbolHoverContent != null ? hoverContent + (hoverContent != "" ? "  \n" : "") + symbolHoverContent : hoverContent;
+      return CreateMarkdownHover(hoverContent);
+    }
+
+    class TupleComparer : IComparer<(int, int)> {
```
keyboardDrummer, line 63: Isn't this the default way of comparing tuples? https://stackoverflow.com/a/27781538/93197

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
keyboardDrummer: In the past I've done this type of thing by first finding a particular AST element, in this case it would be finding the assertion statement, which you do by checks of the `node.Range.Contains(position)` starting at the root node of the tree. That's similar to what's being done here although it would include modules and classes as filters. I think for finding elements in lists, like finding a particular statement in a list of statements, you can also do a binary search instead of a linear traversal.

Then given a particular AST node, compilation results relating to that AST node, such as resolved declarations for nodes that are references, or in this case verification performance data, should be attached to it using some form of a dictionary.

@MikaelMayer have you considered storing the verification results on the AST instead of on a separate tree that mimics parts of the AST?

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -24,30 +36,220 @@ public class DafnyHoverHandler : HoverHandlerBase {
...
+      public int Compare((int, int) key1, (int, int) key2) {
+        return key1.Item1 < key2.Item1 ? -1 :
+          key1.Item1 == key2.Item1 ? key1.Item2 - key2.Item1 : 1;
+      }
+    }
+
+    private string? GetDiagnosticsHover(DafnyDocument document, Position position, out bool areMethodStatistics) {
```
keyboardDrummer, line 70: This method seems fairly long. Is there are way to cut it up into smaller parts?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -64,5 +64,9 @@ public record DafnyDocument(
     }
 
     public int LinesCount => VerificationTree.Range.End.Line;
+
+    public static int NumberOfLines(string text) {
+      return text.Count(c => c == '\n') + 1;
```
keyboardDrummer, line 69: I don't think we should compute a number of lines from the full text anywhere besides in `TextChangeProcessor`. Instead of working with `TextDocumentItem` we can work with our own datatype (like `DocumentTextBuffer`) that besides storing the text as a `string` also stores the number of lines.

In the future we'll change `TextChangeProcessor` so it'll update `DocumentTextBuffer` with a cost linear to the amount of changes (instead of the the current cost which is linear in the document size). I'm not sure what the text buffer data-structure will look like, but maybe this can inspire us: https://code.visualstudio.com/blogs/2018/03/23/text-buffer-reimplementation

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -352,54 +366,82 @@ Range Range
...
     // Used to re-trigger the verification of some diagnostics.
     string Identifier,
     string Filename,
     // The range of this node.
     Range Range
-  ) : VerificationTree(DisplayName, Identifier, Filename, Range) {
+  ) : VerificationTree(Kind, DisplayName, Identifier, Filename, Range) {
     // Recomputed from the children which are ImplementationVerificationTree
-    public List<AssertionBatchVerificationTree> AssertionBatches { get; set; } = new();
+    public ImmutableDictionary<(int, int), AssertionBatchVerificationTree> AssertionBatches { get; private set; } =
```
keyboardDrummer, line 388: Could you replace the anonymous tuple in this field with a record to make it easier to understand what the field contains?

MikaelMayer:
> For the UX, how would diagnostic hints (they show up as ... under the first bit of the range they cover) compare to hover? I'm wondering if hints will provide greater discoverability of this feature.

Here my quick and probably incomplete thinking about hints vs hovering:

- Hints required the language server pushing them. I'm not sure I want to push information about 1000 asserts in a single file.

- I don't worry about the discoverability if it is implemented using the hovering mechanism, because 
1) For single failing assertions, user already are used to hover them to see the error message.
2) For methods since the hover is for the entire line, users will very likely always accidentally discover this feature very quickly.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
MikaelMayer: I replaced it with the record `ASsertionBatchIndex` for now

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: After I changes tuples to a record, I kept this class since the tuple comparison wouldn't work anymore, but thanks for this interesting finding !

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: I haven't considered changing the AST to store verification results there, not just because I think it is better if we modified the AST as little as possible,
but especially because the AST does not have the notion of implementations and assertion batches like what we get from verification.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: Thanks for pointing it out ! I refactored it in 3 smaller parts.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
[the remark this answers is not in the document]
atomb: Oh, of course! 🤦

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
MikaelMayer: I implemented your suggestion, I like it ! Makes more sense than in the DafnyDocument.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -1,18 +1,30 @@
...
+using Microsoft.Dafny.LanguageServer.Workspace.Notifications;
+using VerificationStatus = Microsoft.Dafny.LanguageServer.Workspace.Notifications.VerificationStatus;
 
 namespace Microsoft.Dafny.LanguageServer.Handlers {
   public class DafnyHoverHandler : HoverHandlerBase {
     // TODO add the range of the name to the hover.
     private readonly ILogger logger;
     private readonly IDocumentDatabase documents;
 
+    private const int RuLimitToBeOverCostly = 10000000;
```
atomb, line 24: I'm not entirely sure this is the right value, but I think we'll only find out by using it in practice.

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
atomb: Yeah, I think it's reasonable to always say what batch it's in, partly because it clarifies that the time and resource counts listed are for batches, not individual assertions (and only apply to individual assertions if they're 1/1 in the current batch).

atomb (dismissed):
This is a great feature! And I think all of the previous comments from @keyboardDrummer and myself have been addressed.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/TextChangeProcessor.cs
```diff
@@ -5,29 +5,38 @@
...
       if (change.Range == null) {
+        numberOfLines = DocumentTextBuffer.ComputeNumberOfLines(change.Text);
         // https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocumentContentChangeEvent
         return change.Text;
       }
       int absoluteStart = change.Range.Start.ToAbsolutePosition(text, cancellationToken);
       int absoluteEnd = change.Range.End.ToAbsolutePosition(text, cancellationToken);
+      numberOfLines -=
+        DocumentTextBuffer.ComputeNumberOfNewlines(text.Substring(absoluteStart, absoluteEnd - absoluteStart));
+      numberOfLines += DocumentTextBuffer.ComputeNumberOfNewlines(change.Text);
```
keyboardDrummer, line 39: Seems like you can combine line 39 and 31 by putting line 39 before line 30.

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -65,4 +65,25 @@ public record DafnyDocument(
...
+        Uri = textDocumentItem.Uri,
+        NumberOfLines = ComputeNumberOfLines(textDocumentItem.Text),
+        Version = textDocumentItem.Version,
+        LanguageId = textDocumentItem.LanguageId
+      };
+    }
+
+    public int NumberOfLines { get; init; } //
+
+    public static int ComputeNumberOfLines(string text) {
```
keyboardDrummer, line 82: I would put this function and the next in TextChangeProcessor and make `NumberOfLines` a parameter of `DocumentTextBuffer`

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -65,4 +65,25 @@ public record DafnyDocument(
...
+      return new DocumentTextBuffer {
+        Text = textDocumentItem.Text,
+        Uri = textDocumentItem.Uri,
+        NumberOfLines = ComputeNumberOfLines(textDocumentItem.Text),
+        Version = textDocumentItem.Version,
+        LanguageId = textDocumentItem.LanguageId
+      };
+    }
+
+    public int NumberOfLines { get; init; } //
```
keyboardDrummer, line 80: Is `init` used? `//` can be removed.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/TextChangeProcessor.cs
```diff
@@ -5,29 +5,38 @@
...
       return originalTextDocument with {
         Version = documentChange.TextDocument.Version,
-        Text = ApplyTextChanges(originalTextDocument.Text, documentChange.ContentChanges, cancellationToken)
+        Text = newText,
+        NumberOfLines = newNumberOfLines
       };
     }
 
-    private static string ApplyTextChanges(string originalText, Container<TextDocumentContentChangeEvent> changes, CancellationToken cancellationToken) {
+    private static string ApplyTextChanges(string originalText, int originalLines, Container<TextDocumentContentChangeEvent> changes, out int numberOfLines, CancellationToken cancellationToken) {
```
keyboardDrummer, line 17: I think we can clean up these APIs by letting DafnyDocument not contain a `TextDocumentItem` but instead a `VersionedTextDocumentIdentifier` (containing the Version and Uri) and a `DocumentTextBuffer` (containing just the text and the # of lines), and then these APIs could take and return a `DocumentTextBuffer`.

But we might want to do this in a follow-up PR (or even better, an earlier one), since these changes are all a bit out of scope of hover.

At some point we should also move `PositionsExtensions.ToAbsolutePosition` to `DocumentTextBuffer` and implement it in constant time.

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -45,7 +45,7 @@ public record DafnyDocument(
     /// </summary>
     public VerificationTree VerificationTree { get; init; } = new DocumentVerificationTree(
       TextDocumentItem.Uri.ToString(),
-      TextDocumentItem.Text.Count(c => c == '\n') + 1
+      TextDocumentItem.NumberOfLines
```
keyboardDrummer, line 48: Nice :)

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
keyboardDrummer: > especially because the AST does not have the notion of implementations and assertion batches like what we get from verification.

Good point. Food for thought :-)

keyboardDrummer (approved):

keyboardDrummer:
> * Hints required the language server pushing them. I'm not sure I want to push information about 1000 asserts in a single file.

That's a good callout. Let's start with hover and see how that works.

keyboardDrummer (approved):

PR #1947, opened 2022-03-29
Add human-readable verification logger

atomb:
Adds a new `/verificationLogger:text` option which logs all verification results in a human-readable form, including the positive description of all assertions verified. The result is printed to the console unless the `LogFileName` parameter is provided to `/verificationLogger` (as with the `csv` and `trx` versions).


MikaelMayer:
Well done !
Could you add a test that at least looks whether the output has the shape you describe in VerificationLogger? That way, people won't accidentally remove resource count, etc.

atomb:
> Well done ! Could you add a test that at least looks whether the output has the shape you describe in VerificationLogger? That way, people won't accidentally remove resource count, etc.

Just to be clear, you mean having additional lines in the test to cover everything that's output, rather than only the four lines currently checked, right?

MikaelMayer:
> Just to be clear, you mean having additional lines in the test to cover everything that's output, rather than only the four lines currently checked, right?

Yes. By the way, how did you check only the four lines? I haven't seen the command `OutputCheck` before

atomb:
> Well done ! Could you add a test that at least looks whether the output has the shape you describe in VerificationLogger? That way, people won't accidentally remove resource count, etc.

The test now includes more of the output, though not every line. The `OutputCheck` command is nice because it allows regular expressions, which is good for cases where, for example, the time might change, but it also biases a little toward checking only a subset of the output: the part you really care about. It does now enforce that duration and resource counts are included, though.

MikaelMayer:
Great, thanks! I'll approve it once it's rebased.

Test/logger/TextLogger.dfy
```diff
@@ -0,0 +1,20 @@
+// RUN: %dafny /compile:0 /verificationLogger:text /vcsSplitOnEveryAssert "%s" > "%t"
```
robin-aws, line 1: Great test, and I continue to appreciate having support for the OutputCheck DSL. :)

We should also have a test that DOESN'T use `/vcsSplitOnEveryAssert` too, so we know we handle the case of more than one assertion in a batch.

Source/DafnyDriver/BoogieXmlConvertor.cs
```diff
@@ -54,6 +55,13 @@ public static class BoogieXmlConvertor {
         } else if (loggerName == "csv") {
           var csvLogger = new CSVTestLogger();
           csvLogger.Initialize(events, parameters);
+        } else if (loggerName == "text") {
+          // This doesn't actually use the XML converter. It instead uses a collection of VerificationResult
+          // objects. Ultimately, the other loggers should be converted to use those objects, as well,
+          // and then it would make sense to rename this class.
```
robin-aws, line 61: As in making it `TextTestLogger` instead?

Source/DafnyDriver/TextLogger.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.Dafny;
+
+public class TextLogger {
+  private TextWriter tw;
+
+  public void Initialize(Dictionary<string, string> parameters) {
+    tw = parameters.TryGetValue("LogFileName", out string filename) ? new StreamWriter(filename) : Console.Out;
+  }
+
+  public void LogResults(List<(Implementation, VerificationResult)> verificationResults) {
```
robin-aws, line 15: Would you mind attaching some sample output, just to double check that the formatting makes sense etc.? The CHECK testing is perfect but omits some of that since it has to regex-away details.

Source/Dafny/DafnyOptions.cs
```diff
@@ -1118,13 +1118,25 @@ not sound
...
+    The ""trx"" and ""csv"" loggers automatically choose an output file
+    name by default, and print the name of this file to the console. The
+    ""text"" logger prints its output to the console by default, but can
+    send output to a file given the `LogFileName` option.
+
+    The ""text"" logger also includes a more detailed breakdown of what
+    assertions appear in each assertion batch. When combined with the
+    `/vcsSplitOnEveryAssert` option, it will provide approximate time
+    and resource use costs for each assertion, allowing identification
+    of especially expensive assertions.
```
robin-aws, line 1139: Much better, thanks! :)

robin-aws (changes requested):

Source/DafnyDriver/BoogieXmlConvertor.cs
atomb: Ah, I meant the class containing this code: `BoogieXmlConverter` will probably no longer be appropriate. ;)

Test/logger/TextLogger.dfy
atomb: Oh, yes!

Source/DafnyDriver/TextLogger.cs
atomb: Here's without `/vcsSplitOnEveryAssert`:

````
Results for Impl$$_module.__default.M
  Overall outcome: Errors
  Overall time: 00:00:00.4675430
  Overall resource count: 10947

  Assertion batch 0:
    Outcome: Invalid
    Duration: 00:00:00.1252410
    Resource count: 10947

    Assertions:
      /Users/aarotomb/Repositories/dafny/Test/logger/TextLogger.dfy(18,14): divisor is always non-zero.
      /Users/aarotomb/Repositories/dafny/Test/logger/TextLogger.dfy(19,9): assertion always holds
````

And with:
````
Results for Impl$$_module.__default.M
  Overall outcome: Errors
  Overall time: 00:00:00.5055420
  Overall resource count: 22035

  Assertion batch 1:
    Outcome: Valid
    Duration: 00:00:00.0892450
    Resource count: 3267

    Assertions:

  Assertion batch 2:
    Outcome: Valid
    Duration: 00:00:00.0343600
    Resource count: 8473

    Assertions:
      /Users/aarotomb/Repositories/dafny/Test/logger/TextLogger.dfy(18,14): divisor is always non-zero.

  Assertion batch 3:
    Outcome: Invalid
    Duration: 00:00:00.0191820
    Resource count: 10295

    Assertions:
      /Users/aarotomb/Repositories/dafny/Test/logger/TextLogger.dfy(19,9): assertion always holds
````

atomb: The names are still in Boogie format. Improving that is something I'd still like to do soon, but will require a bit of work.

robin-aws: Thanks, LGTM - I mostly wanted to double check the formatting on the durations was reasonable.

Interesting how a batch with zero assertions still has a non-zero resource count isn't it! I assume that has to be assertions in the `DafnyPrelude.bpl` or some other kind of fixed cost.

Source/DafnyDriver/BoogieXmlConvertor.cs
robin-aws: Oh yeah that class can be fully deleted once we switch over, and all related code.

Source/DafnyDriver/TextLogger.cs
atomb: That's something I've been meaning to investigate for a while. I looked at one instance and it was just `true` (with some background assumptions, which is probably what makes it take non-zero resources). I wonder if it's an artifact of splitting. Do you know, @keyboardDrummer?

robin-aws (approved):

PR #1949, opened 2022-03-30
fix: Allow datatype update expressions for constructors with nameonly parameters

### LINKED ISSUE #1909: Datatype update syntax doesn't work with nameonly fields
```dafny
datatype ABC = ABC(nameonly a: int, nameonly b: int, nameonly c: int)

function JustChangeB(abc: ABC): ABC {
  abc.(b := 42) // Error: nameonly parameter 'a' must be passed using a name binding; it cannot be passed positionally
                // (same error repeated for every nameonly field)
}
```

RustanLeino:
Fixes #1909


Source/Dafny/Resolver.cs
```diff
@@ -15601,16 +15601,23 @@ public class ResolveOpts {
...
           } else {
-            ctor_args.Add(new ExprDotName(tok, d, f.Name, null));
+            ctorArg = new ExprDotName(tok, d, f.Name, null);
           }
+          ctorArguments.Add(ctorArg);
+          var bindingName = new Token(tok.line, tok.col) {
+            filename = tok.filename,
+            val = f.Name
+          };
+          actualBindings.Add(new ActualBinding(bindingName, ctorArg));
```
MikaelMayer, line 15618: Everything else is just refactoring, replacing `null` by `bindingName` is the actual fix, right?

MikaelMayer (approved):
Thanks for also cleaning up the code, along with fixing that issue.

Source/Dafny/Resolver.cs
RustanLeino: Correct.

MikaelMayer (approved):
Great job !

Issue #1799, opened 2022-02-08
Function methods only ever have one return path, which makes debugging harder


cpitclaudel:
Consider these two equivalent programs:

```dafny
function method fIncorrectInOneCase(x: int) : (y: int)
  ensures y > 0
{
  if (x < -3) then 5
  else if (x < 5) then -2
  else 4
}

method mIncorrectInOneCase(x: int) returns (y: int)
  ensures y > 0
{
  if (x < -3) { return 5; }
  else if (x < 5) { return -2; }
  else { return 4; }
}
```

Here's Dafny's output:

![image](https://user-images.githubusercontent.com/8181630/153058927-031b93a2-c1ab-44e7-9952-175269e4ac5e.png)

The error is correctly positioned in the method, but not in the function.  This makes debugging function methods much less pleasant than methods.

Here's the textual output:

```
test.dfy(1,16): Error: A postcondition might not hold on this return path.
test.dfy(2,12): Related location: This is the postcondition that might not hold.
Execution trace:
    (0,0): anon0
    (0,0): anon9_Else
    (0,0): anon10_Else
    (0,0): anon11_Then
    (0,0): anon8
test.dfy(13,20): Error: A postcondition might not hold on this return path.
test.dfy(10,12): Related location: This is the postcondition that might not hold.
Execution trace:
    (0,0): anon0
    (0,0): anon5_Else
    (0,0): anon6_Then
```

Btw, what's up with line numbers in execution traces?

robin-aws:
+1, I've seen this frequently irritate Dafny users (including myself!)

It also makes Boogie-level techniques for measuring verification cost like `{:vcs_split_on_every_assert}` less effective, since you can only measure the cost of the one post-condition assertion in the unified return block.

MikaelMayer:
I thought about this problem over the week-end, and here is maybe how we should tackle it.

When encountering a "return" statement in boogie, instead of translating literally the return expression, we should translate "if expressions", "let-var" and "match expressions" to their respective _statements_.
That way, it would work only for `function`s, but also for `methods` who return an expression.

MikaelMayer:
Actually, it looks like Boogie knows exactly the path in which a function's postcondition does not hold. It's in the trace. For example:

```dafny
function test_function(x: bool): (r: int)
  ensures r == 2
 {
  if x then
    1
  else
    2
}
```

Boogie trace
```
function-implementation.bpl(2947,5): Error: A postcondition might not hold on th
is return path.
[...]

PR #1954, opened 2022-03-31
feat: Some reads clauses don't need to be verified

### LINKED ISSUE #1953: Reads is checked even if nothing needs to be checked.
Consider the following example:

```dafny
predicate method P()
predicate TestMap(s0: set<int>, s1: set<int>) {
  var m0 := map key | key in (s0 + s1) && P() :: true;
  true
}
```

Dafny will emit a well-formedness check for the call of P() for its reads clause... which will result in the following Boogie formula:

```dafny
assert (forall<alpha> $o: ref, $f: Field alpha :: false ==> $_Frame[$o, $f]);
```
There is no reason to include this assert as we know its precondition is false.
Removing it removes the implementation alltogether, which can takes up to 70ms just to verify if that's the only check.

MikaelMayer:
This changes removes the emission of an assert in the context of frames when it's obvious that it is true.
Note that we don't want to do this for every assert because users might want to keep them for keeping annotations like `{:focus}`

**Discussion:**
If we remove an assert, although it's nicer for the verification time, it might be less nice for users who might wrongly think that the verifier does not perform reads verification at this call site.
1) @atomb do you care?
2) Is there an alternative where we keep this assert for positive report, but Boogie does not verify it?

I'm in favor of just removing the assert

Fixes #1953 

## Interesting side-effect
While updating the tests, I came across the following one
```
function SumBad(a: List): int
  ensures 0 <= Sum(a);  // this is still okay, because this is calling the good Sum
  ensures 0 <= SumBad(a);  // error: cannot prove postcondition
{
  match a
  case Nil => 0
  case Cons(x, tail) => if x < 0 then 0 else FibWithoutPost(x)
}
```
Previously, the second postcondition was failing, and the return path was indicating the "function" keyword.
Now, the return path is indicating the "if" keyword in the second case.
I'm wondering why this partially solves #1799, although it's not working for the main example.

## Performance measurement.
Dafny returns the number of implementations with errors, and the number of verified implementations.
That total number decreased from:

### 2680 implementations verified or with errors

to

### 2301 implementations verified or with errors

meaning that 379 implementations were only doing one thing: testing a trivial assert.
If every overhead implementation was taking 50ms, that's a whole 19 seconds of verification that we remove from the test suite. It's not that much but always appreciated for little projects :-)


Source/Dafny/Verifier/Translator.cs
```diff
@@ -3737,8 +3737,13 @@ public enum StmtType { NONE, ASSERT, ASSUME };
       Bpl.Expr ante = Bpl.Expr.And(Bpl.Expr.Neq(o, predef.Null), etran.IsAlloced(tok, o));
       Bpl.Expr oInCallee = InRWClause(tok, o, f, calleeFrame, etran, receiverReplacement, substMap);
       Bpl.Expr inEnclosingFrame = Bpl.Expr.Select(etran.TheFrame(tok), o, f);
+
+      var preCondition = Bpl.Expr.And(ante, oInCallee);
       Bpl.Expr q = new Bpl.ForallExpr(tok, new List<TypeVariable> { alpha }, new List<Variable> { oVar, fVar },
-                                      Bpl.Expr.Imp(Bpl.Expr.And(ante, oInCallee), inEnclosingFrame));
+                                      Bpl.Expr.Imp(preCondition, inEnclosingFrame));
+      if (preCondition is Bpl.LiteralExpr lit && lit.isBool && lit.asBool == false) {
```
cpitclaudel, line 3744: I was surprised to not be able to find similar tests elsewhere in the code.  Still, maybe it's worth extracting this logic to a separate function?

cpitclaudel (commented):
Nice, but… how does this code work?  Are the Bpl constructors already "smart" (ie perform simplifications on their arguments)?  You have `var preCondition = Bpl.Expr.And` — how does that become a `LiteralExpr`?  If this is done by a later simplification pass, then maybe that pass should be changed to recognize `false ==> b`; and if it's done using smart constructors, then maybe that's what we should do for implifications and foralls as well, so that this simplification happens for free?

Source/Dafny/Verifier/Translator.cs
MikaelMayer: Oh, do you mean, taking a Boogie expression, and determining if it's obviously always true?
A kind of `IsAlwaysTrue(Boogie.Expr): Boolean` ?

cpitclaudel: Yep

robin-aws:
> It's not that much but always appreciated for little projects

Who are you calling "little" :)

MikaelMayer:
I just added units tests to this branch, because otherwise the function `ÌsAlwaysExpr` is fragile.

cpitclaudel (commented):
This is good, but I'd like to discuss strategy a bit.  Shouldn't this be done in smart constructors, so that all expressions that get constructed get this simplification, "for free"?

(This would have wider implications, since the following wouldn't verify anymore, for example:

```
predicate PPP(i: int)
predicate QQQ(i: int)

lemma QQQi() ensures forall i :: QQQ(i);
lemma PPPQQQi() ensures forall i :: QQQ(i) ==> PPP(i + 1);

method main() {
  QQQi();
  PPPQQQi();
  assert (forall x: nat :: true) || QQQ(0);
  assert PPP(1);
}
```
)

MikaelMayer:
> This is good, but I'd like to discuss strategy a bit. Shouldn't this be done in smart constructors, so that all expressions that get constructed get this simplification, "for free"?

It's good that you find an example where we shouldn't do this simplification: when the expression of the assert is statically used to instantiate quantifiers that can make further assertions to succeed.

So let's be conservative. What I changed concerns only the "reads" clause, which i don't think adds information that would otherwise be accessible, is it @RustanLeino ?

cpitclaudel:
> It's good that you find an example where we shouldn't do this simplification: when the expression of the assert is statically used to instantiate quantifiers that can make further assertions to succeed.

Unless I'm misreading the code you're not doing that right now, are you?  So I think your code as-is is safe?

atomb:
I'm somewhat of the mind that the set of assertions that arise from a program should be very predictable, even if they're obviously true, and that if someone is looking through a list of things that have been proven they'll see everything they expect. If there is a way to keep the assertion, so that we can report it to the user, I'd slightly prefer that. But I think it's also reasonable to avoid sending obviously true assertions to the solver (and perhaps even avoid generating them in Boogie). 

Now that there's a data type for `VerificationResult` and `VCResult`, maybe we could synthesize a list of "free" VC results before translating to Boogie and then include those along with the results we get back from Boogie+Z3. There's probably enough information to construct `VCResult` instances at the point of discovering that a check is trivial. What do you think, @MikaelMayer?

MikaelMayer:
> Now that there's a data type for `VerificationResult` and `VCResult`, maybe we could synthesize a list of "free" VC results before translating to Boogie and then include those along with the results we get back from Boogie+Z3. There's probably enough information to construct `VCResult` instances at the point of discovering that a check is trivial. What do you think, @MikaelMayer?

I agree in general it would be good if we could keep a predictable set of assertions, so let's keep this optimization only in very specific cases. The more assertions, the more guarantees.

Now, in this particular case of the reads clause, it seems to me that, beyond the obvious performance improvement that could be actually be more generally implemented in Boogie (which I agree we should do, if possible), here there is however nothing to "check" since the method or function from which the call is performed does not have a reads clause, neither does the callee method.

Let's look at a more obvious example. Let's pretend we told the user "The decreases conditions are checked" on a method that does not contain any recursive call. For sure, we checked that if there is a decrease clause and if the method is recursive, the decreases clause is satisfied. But there is none, so the verification message isn't relevant. I see empty reads clauses the same way. Let me know what you think of that.

cpitclaudel:
> Let's look at a more obvious example. Let's pretend we told the user "The decreases conditions are checked" on a method that does not contain any recursive call. For sure, we checked that if there is a decrease clause and if the method is recursive, the decreases clause is satisfied. But there is none, so the verification message isn't relevant. I see empty reads clauses the same way. Let me know what you think of that.

I agree with this part ^^

cpitclaudel (approved):

PR #1955, opened 2022-04-01
fix: autocontracts tooltip for Valid() doesn't use expr printer

jamesbornholt:
The autocontracts tooltip for `Valid` is using a direct string interpolation instead of the Expression printer, which means it doesn't actually include the expression.

Before:
```
git-issue-1955.dfy(5,14): Info: autocontracts:
   reads Microsoft.Dafny.ThisExpr, Microsoft.Dafny.MemberSelectExpr
   ensures Microsoft.Dafny.BinaryExpr
```

After:
```
git-issue-1955.dfy(5,14): Info: autocontracts:
   reads this, Repr
   ensures Valid() ==> this in Repr
```


Test/git-issues/git-issue-1955.dfy
```diff
@@ -0,0 +1,8 @@
+// RUN: %dafny /compile:0 /print:"%t.print" /dprint:"%t.dprint" /printTooltips "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+class {:autocontracts} Thing {
+    predicate Valid() {
+        true
+    }
+}
```
fabiomadge, line 8:
```suggestion
}

```

fabiomadge:
Looks good, thanks! Would you mind opening a separate issue?

MikaelMayer (approved):
Thanks for this reported issue and this fix at the same time ! That's amazing !

cpitclaudel:
Looks good, but alternatively: Could we change the ToString of expressions to call `Printer.ExprToString`?

MikaelMayer:
> Looks good, but alternatively: Could we change the ToString of expressions to call `Printer.ExprToString`?

We could do that as well in general, but here this mechanism was already implemented in this function.
Note that changing ToString might suffer performance penalty when debugging, because the debugger will try to invoke this expression

cpitclaudel:
> Note that changing ToString might suffer performance penalty when debugging, because the debugger will try to invoke this expression

Only when stepping through, which is nice because performance isn't really a concern at that step, right?

PR #1960, opened 2022-04-01
chore: Attach continuous-integration environment to most expensive CI jobs

robin-aws:
This attaches a "continuous-integration" [environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) to all but the cheapest CI jobs that are triggered on all PRs. This environment requires a manual approval from maintainers, and hence lets us decide explicitly when to kick off CI on a PR. This should help manage the concurrency limit better and avoid wasting cycles before a PR is refined enough to actually seek a green run before getting merged.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -9,7 +9,6 @@ on:
 jobs:
   build:
 
-    environment: continuous-integration
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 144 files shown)


robin-aws:
Okaaaaaaay this is a little noisier than I'd hoped. :)

cpitclaudel:
Re noise: indeed, I got about 50 emails about this one ^^

PR #1965, opened 2022-04-04
 fix #1903 change to substitute's behavior for collection data types.

### LINKED ISSUE #1903: Compiling issue when using generic types and default values
While using methods with generic types and default initialisation I encountered a compiling issue.

### The issue
The following program
```[dafny]
method g<S>(x : seq<S> := []) 
{
   print "worked!\n";  
}


method {:verify true} Main() { 
    g<nat>([]);   // works 
    // g<nat>();   //  compile error
}
```

fails to compile with error message:
```[dafny]
dafny git:(main) ✗ dafny /noVerify /compile:4 '/Users/franck/development/dafny/bug-reports/generic.dfy'

Errors compiling program:
(1832,46): error CS0246: The type or namespace name '__S' could not be found (are you missing a using directive or an assembly reference?)

(1832,31): error CS1503: Argument 1: cannot convert from 'Dafny.ISequence<__S>' to 'Dafny.ISequence<System.Numerics.BigInteger>'
```
### Preliminary assessment

The problem seems to show up with other types of collections too e.g. `map<nat, S>`.
The error disappears if we pass a parameter e.g. a `[]` to method `g`.
The type inference for the default value of the parameter may have a glitch.

### Dafny version
MacOs X version 

```
dafny git:(main) % dafny /version 
Dafny 3.4.1.40215
```

### How the issue was discovered

Issue showed up in a larger piece of code and was trimmed down to the examples in this report to isolate the problem.

prvshah51:
To make sure Substituter computes  Resolver.SubstType(expr.Type, typeMap) whether or not newExpr is null. Then, if either newExpr is non-null or if the susbstituted type is different from expr.Type, then Substitute should return a new expression with the type substitution.

Fixes #1903


robin-aws (dismissed):
Nice work debugging this, subtle bug!

Test/git-issues/git-issue-1903.dfy
```diff
@@ -0,0 +1,12 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method g<T>(x : seq<T> := []) 
+{
+   print "worked!\n";  
+}
+
+
+method {:verify true} Main() {
```
MikaelMayer, line 10:
```suggestion
method Main() { 
```

Please confirm if the change I suggest is relevant. It seems that verify is always true by default.

Test/git-issues/git-issue-1903.dfy
```diff
@@ -0,0 +1,12 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
```
MikaelMayer, line 1:
```suggestion
// RUN: %dafny /compile:3 "%s" > "%t"
```

You also want to run the file, don't you?

Test/git-issues/git-issue-1903.dfy.expect
```diff
@@ -0,0 +1,2 @@
+
+Dafny program verifier finished with 1 verified, 0 errors
```
MikaelMayer, line 2:
```suggestion
Dafny program verifier finished with 1 verified, 0 errors
worked!

```

Source/Dafny/Substituter.cs
```diff
@@ -59,7 +59,7 @@ public class Substituter {
       } else if (expr is DisplayExpression) {
         DisplayExpression e = (DisplayExpression)expr;
         List<Expression> newElements = SubstituteExprList(e.Elements);
-        if (newElements != e.Elements) {
+        if (newElements != e.Elements || !Resolver.SubstType(expr.Type, typeMap).Equals(e.Type)) {
```
MikaelMayer, line 62:
```suggestion
        if (newElements != e.Elements || !Resolver.SubstType(e.Type, typeMap).Equals(e.Type)) {
```

It took me longer than necessary to figure out that `expr` and `e`` are the same object. Is there any reason not to use the same variable?

Also, Equals() has two variants, one that considers equality modulo subset types (the one you use), and the other that considers equality including subset types (with a second argument `, true`). I'm not sure, but I would have thought the second type of equality might have been appropriate. Could you let me know your reasoning for choosing one over the other?

MikaelMayer (changes requested):
Great job finding that bug ! That was non-trivial for sure. A few suggestions but otherwise I'll approve it quickly as it's a simple solution with a test.

Source/Dafny/Substituter.cs
prvshah51: I should have used `e` there, thanks. for the latter I don't know about the other type of `Equals` are you suggesting to change it with `Equals(e.Type,true))` ?

MikaelMayer: I'm not sure about that, but yes, that's my suggestion.

Test/git-issues/git-issue-1903.dfy
prvshah51: That is true, I was using example provided in the issue thus didn't remove it I did that now.

PR #1968, opened 2022-04-04
tests: Clean up tutorial/maximum.dfy

cpitclaudel:
Following up from a recent chat


keyboardDrummer (commented):
Changes look good but it seems like they cause a test failure.

cpitclaudel:
Good catch, thanks, and sorry about that.

keyboardDrummer (approved):

PR #1977, opened 2022-04-06
feat: Verification diagnostics related ranges

MikaelMayer:
This PR adds related ranges to verification tree nodes, in order to expose more of the information offered by counter-examples to end-users.

For example, it enables identifying a path of failure that otherwise wouldn't be available.
image


MikaelMayer:
Closed for now because not maintained.
It was mostly useful to identify the failing path, but there was a lot of garbage highlighted elements otherwise.
To implement it in the future, It would require to implement a pull request from the IDE extension

PR #1979, opened 2022-04-06
feat: Numerical types in Python

fabiomadge:


Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -74,38 +74,40 @@ def IsNat(ord):
...
-        # invariant: x != 0 && x * 10^log10 == old(x)
+    def deleteFactor(f, x):
+        y = 0
         while True:
-            if i == 1:
-                return log10
-            elif i % 10 == 0:
-                log10 += 1
-                i //= 10
+            if x > 1 and x % f == 0:
```
cpitclaudel, line 95: Suggestion: move the condition into the while and move return out of the loop

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -74,38 +74,40 @@ def IsNat(ord):
...
+        return f"{sign}0.{'0' * (shift - len(digits))}{digits}"
 
     @staticmethod
-    def intLog10(i):
-        log10 = 0
-        if i == 0:
-            raise ValueError("must be called with a power of 10")
-        # invariant: x != 0 && x * 10^log10 == old(x)
+    def deleteFactor(f, x):
+        y = 0
```
cpitclaudel, line 93: Can we pick a better name?

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -74,38 +74,40 @@ def IsNat(ord):
...
-                return log10
-            elif i % 10 == 0:
-                log10 += 1
-                i //= 10
+            if x > 1 and x % f == 0:
+                y += 1
+                x //= f
             else:
-                raise ValueError("must be called with a power of 10")
+                return x == 1, x, y
```
cpitclaudel, line 99: Suggestion: return only x and y, not `x == 1`

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -74,38 +74,40 @@ def IsNat(ord):
...
+
+    @staticmethod
+    def devidesAPowerOf10(x):
+        c, rem, expA = BigRational.deleteFactor(10, x)
```
cpitclaudel, line 103: Suggestion: remove `c`

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -74,38 +74,40 @@ def IsNat(ord):
...
+        if c:
+            return True, 1, expA
+        if rem % 5 == 0 or rem % 2 == 0:
+            major, minor = (5, 2) if rem % 5 == 0 else (2, 5)
+            c, rem, expB = BigRational.deleteFactor(major, rem)
+            return c, minor**expB, expA+expB
+        return False, -1, -1
```
cpitclaudel, line 110: Suggestion: Return None, not `-1`, for `no value`

cpitclaudel: In fact, I wouldn't return the boolean: just `None` in the `False` case or a tuple in the `True` case (that's how the `Optional` type works in Python)

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -74,38 +74,40 @@ def IsNat(ord):
...
-                raise ValueError("must be called with a power of 10")
+                return x == 1, x, y
+
+    @staticmethod
+    def devidesAPowerOf10(x):
+        c, rem, expA = BigRational.deleteFactor(10, x)
+        if c:
+            return True, 1, expA
+        if rem % 5 == 0 or rem % 2 == 0:
+            major, minor = (5, 2) if rem % 5 == 0 else (2, 5)
```
cpitclaudel, line 107: Same here: since we know that at least one of the two factors will already be exhausted we can just make two calls :)

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -74,38 +74,40 @@ def IsNat(ord):
...
             else:
-                raise ValueError("must be called with a power of 10")
+                return x == 1, x, y
+
+    @staticmethod
+    def devidesAPowerOf10(x):
+        c, rem, expA = BigRational.deleteFactor(10, x)
+        if c:
+            return True, 1, expA
+        if rem % 5 == 0 or rem % 2 == 0:
```
cpitclaudel, line 106: This conditional is unnecessary, I think: `deleteFactor` returns instantly when `% f == 0`.

cpitclaudel (changes requested):

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: True!

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: Thanks!

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: Sure, but need to know which one I used later anyway.

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: Is there a nice way to keep the multi-assign?

Source/DafnyRuntime/DafnyRuntime.py
cpitclaudel: I don't think you do?

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -89,25 +89,21 @@ def __str__(self):
...
-            return True, 1, expA
-        if rem % 5 == 0 or rem % 2 == 0:
+        rem, expA = BigRational.isolateFactor(10, x)
+        if rem % 5 == 0 or rem % 2 == 0 or rem == 1:
             major, minor = (5, 2) if rem % 5 == 0 else (2, 5)
-            c, rem, expB = BigRational.deleteFactor(major, rem)
-            return c, minor**expB, expA+expB
-        return False, -1, -1
+            rem, expB = BigRational.isolateFactor(major, rem)
+            return rem == 1, minor**expB, expA+expB
```
cpitclaudel, line 105:
```suggestion
            return minor**expB, expA+expB if rem == 1 else None
```

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -89,25 +89,21 @@ def __str__(self):
...
+        return False, None, None
```
cpitclaudel, line 106:
```suggestion
        return None
```

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -89,25 +89,21 @@ def __str__(self):
...
-                x //= f
-            else:
-                return x == 1, x, y
+        while x > 1 and x % f == 0:
+            y += 1
+            x //= f
+        return x, y
 
     @staticmethod
     def devidesAPowerOf10(x):
```
cpitclaudel, line 100:
```suggestion
    def dividesAPowerOf10(x):
```

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: Is there a nice way to keep the multi-assign?

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: `minor**expB`

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -567,15 +568,18 @@ protected class ClassWriter : IClassWriter {
     protected override void EmitLiteralExpr(ConcreteSyntaxTree wr, LiteralExpr e) {
       switch (e) {
         case CharLiteralExpr:
-          wr.Write("'{0}'", (string)e.Value);
+          wr.Write($"'{(string)e.Value}'");
```
cpitclaudel, line 571: Do you need the cast? (And, does `Write` correctly support the `0`-arguments case, or will it break if `e.Value` contains curly braces?)

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -886,43 +890,63 @@ protected class ClassWriter : IClassWriter {
...
-          opString = "*"; break;
+          opString = "*";
+          truncateResult = true;
+          break;
 
-        //Fixme: Ints
         case BinaryExpr.ResolvedOpcode.Div:
-          opString = "/"; break;
+          if (resultType.IsIntegerType || resultType.IsBitVectorType || resultType.AsNewtype != null) {
+            staticCallString = "_dafny.euclidianDivision";
```
cpitclaudel, line 937: Not sure how much it matters, but in Python this would typically be styled `euclidian_division`

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -886,43 +890,63 @@ protected class ClassWriter : IClassWriter {
...
+            opString = "/";
+          }
+          break;
 
         case BinaryExpr.ResolvedOpcode.Mod:
-          opString = "%"; break;
+          if (resultType.IsIntegerType || resultType.IsBitVectorType || resultType.AsNewtype != null) {
+            staticCallString = "_dafny.euclidianModulus";
+          } else {
+            opString = "%";
```
cpitclaudel, line 947: What case does this correspond to? Real numbers?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -972,7 +996,25 @@ protected class ClassWriter : IClassWriter {
     }
 
     protected override void EmitConversionExpr(ConversionExpr e, bool inLetExprBody, ConcreteSyntaxTree wr, ConcreteSyntaxTree wStmts) {
+      var (pre, post) = ("", "");
+      if (e.E.Type.IsNumericBased(Type.NumericPersuasion.Int) || e.E.Type.IsBitVectorType) {
+        if (e.ToType.IsNumericBased(Type.NumericPersuasion.Real)) {
+          (pre, post) = ("_dafny.BigRational(", ", 1)");
+        } else if (e.ToType.IsCharType) {
+          (pre, post) = ("chr(", ")");
```
cpitclaudel, line 1004: Is `chr` the right function? (Does Dafny's 16-bits thing line up with Python's `chr`)?

Source/DafnyRuntime/DafnyRuntime.py
cpitclaudel: Sure, but why not just compute a power of 10, so that there isn't even a need for a correction?  Sorry if I'm missing something.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -97,16 +98,47 @@ def isolateFactor(f, x):
...
         if rem % 5 == 0 or rem % 2 == 0 or rem == 1:
             major, minor = (5, 2) if rem % 5 == 0 else (2, 5)
             rem, expB = BigRational.isolateFactor(major, rem)
-            return rem == 1, minor**expB, expA+expB
-        return False, None, None
+            return (minor**expB, expA+expB) if rem == 1 else None
+        return None
+
+    def __add__(self, other):
+        return BigRational(Fraction.__add__(self, other))
```
cpitclaudel, line 110:
```suggestion
        return BigRational(super().__add__(other))
```

cpitclaudel (dismissed):
LGTM, but I left some comments

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: The `Value` is a raw pointer (object), so it seems sensible here. What's the 0-arguments case? `""`?
This works:
```Dafny
method Main(){
  var x := "{";
  print x, "\n";
}
```

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: dividesAPowerOf10 -> divides_a_power_of_10
?

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: Good point, I probably added it for symmetry in `/`.

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: It should be fine, as `char`s can't overflow.

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: We apply this compensation to the numerator of the fraction.

cpitclaudel (approved):

Issue #1441, opened 2021-09-17
Feature request: explicit language features for verification debugging


robin-aws:
Follow-up from #1354.

It is a fairly well-established idiom amongst Dafny programmers to temporarily insert `assume false;` or `assert false;` in various places when trying to determine why their program doesn't verify. The effect is to "turn off verification" for a region of code, in order to gain more precise feedback on what specific proof obligations are not met, or to find regions with a high verification cost (by observing that the verifier takes a very long time to flag the violation from an `assert false;`). Doing so creates an invalid Dafny program, so as long as the verifier raises at least one violation it is still sound, but the effectiveness of this technique depends on the somewhat undefined behavior of how the verifier behaves in the presence of existing logical inconsistencies.

#1354 improves the effectiveness of this mechanism by making the effects of something like `assume false;` more useful for verification debugging, but it's still a fragile situation. I propose that a better solution would be a more explicit way to tell the verifier "don't verify this section", which would therefore have concrete and well-defined semantics. For example (without actually suggesting this syntax is ideal):

```dafny
method JustVerifyThisAlreadyDammit(n: nat) { // error, "body contains unverified blocks"
  var i := 0;
  while i < n {
    unverified { // Temporarily inserted
      SomeCodeThatWontVerifyYet(i);
    }
    i := i + 1;
  }
  assert n == 0; // error, because the unverified block never hides the loop from Boogie
}
```

The current technique reminds me of what I used to do in languages that had limited debugging support: I'd add this snippet in random places just so I had a statement to attach a breakpoint to.

```java
int breakpoint = 42;
breakpoint++;
```

There could be other solutions to this problem, so don't take my suggestion too literally. The key point is I think it's time for SOMETHING better. :)

tjhance:
> The effect is to "turn off verification" for a region of code, in order to gain more precise feedback on what specific proof obligations are not met, or to find regions with a high verification cost (by observing that the verifier takes a very long time to flag the violation from an assert false;)

I do this a lot too, but it seems to me there could be much more direct language support for this common developer scenario, e.g., if dafny had a way to print out a detailed breakdown of how much time it is spending at the granularity of assertions & preconditions.

robin-aws:
Absolutely, and we're investigating improvements in that area as well. I don't even intend this issue to take a strong stance on which solution is the best bang for the buck yet.

kjx:
(assuming this is still live)

I quite like (unverified) tag within code,
[...]

PR #1981, opened 2022-04-06
Added debugging documentation for when verification is slow.

### LINKED ISSUE #1725: Dafny repo wiki page on debugging/fixing verification

MikaelMayer:
Might fix #1725 (@robin-aws ?)

Based on my experience and @RustanLeino 's experience, I added a few sections in the User's guide to help someone for whom verification is slow.

### DIFF
--- a/docs/DafnyRef/UserGuide.md
+++ b/docs/DafnyRef/UserGuide.md
@@ -317,11 +317,11 @@ This list is not exhaustive but can definitely be useful to provide the next ste
 
 ### 24.8.2. Verification debugging when verification is slow {#sec-verification-debugging-slow}
 
-In this section, we describe methodologies to apply in the case when verification is slower than expected, does not terminate, or timeouts.
+In this section, we describe techniques to apply in the case when verification is slower than expected, does not terminate, or timeouts.
 
 #### 24.8.2.1. `assume false;` {#sec-assume-false}
 
-Assuming `false` is a powerful way to short-circuit the verifier and stop verification at a given point[^explainer-assume-false], and since the final compilation steps do not accept this command, it is safe to use it during development.
+Assuming `false` is an empirical way to short-circuit the verifier and usually stop verification at a given point[^explainer-assume-false], and since the final compilation steps do not accept this command, it is safe to use it during development.
 Another similar command, `assert false;`, would also short-circuit the verifier, but it would still make the verifier try to prove `false`, which can also lead to timeouts.
 
 [^explainer-assume-false]: `assume false` tells the Dafny verifier "Assume everything is true from this point of the program". The reason is that, 'false' proves anything. For example, `false ==> A` is always true because it is equivalent to `!false || A`, which reduces to `true || A`, which reduces to `true`.
@@ -635,7 +635,7 @@ The fundamental unit of verification in Dafny is an _assertion batch_, which con
 
 [^caveat-about-assertion-and-assumption]: Caveat about assertion and assumption: One big difference between an "assertion transformed in an assumption" and the original "assertion" is that the original "assertion" can unroll functions twice, whereas the "assumed assertion" can unroll them only once. Hence, Dafny can still continue to analyze assertions after a failing assertion without automatically proving "false" (which would make all further assertions vacuous).
 
-[^smaller-batches]: To create a smaller batch, Dafny duplicates the assertion batch, and transforms each assertion into an assumption into exactly one batch, so that assertions are verified only in one batch. This results in "easier" formulas for the verifier because it has less to prove, but it takes more overhead because every verification instance have a common set of axioms and there is no knowledge sharing between instances because they run independently.
+[^smaller-batches]: To create a smaller batch, Dafny duplicates the assertion batch, and arbitrarily transforms the clones of an assertion into assumptions except in exactly one batch, so that each assertion is verified only in one batch. This results in "easier" formulas for the verifier because it has less to prove, but it takes more overhead because every verification instance have a common set of axioms and there is no knowledge sharing between instances because they run independently.
 
 #### 24.8.3.1. Controlling assertion batches {#sec-assertion-batches-control}
 
--- a/docs/DafnyRef/UserGuide.md
+++ b/docs/DafnyRef/UserGuide.md
@@ -156,9 +156,9 @@ exit code.
 
 ## 24.8. Verification
 
-In this section, we suggest a methodology to figure out [why a single assertion might not hold](#sec-verification-debugging), we explain how to [make the verifier more focused on particular assertions](#sec-assertion-batches) to devote more time to prove them, and we also give some more examples of [useful options and attributes to control verification](#sec-command-line-options-and-attributes-for-verification).
+In this section, we suggest a methodology to figure out [why a single assertion might not hold](#sec-verification-debugging), we propose methodologies to deal with [assertions that slow a proof down](#sec-verification-debugging-slow), we explain how to [verify assertions in parallel or in a focused way](#sec-assertion-batches) to devote more time to prove them, and we also give some more examples of [useful options and attributes to control verification](#sec-command-line-options-and-attributes-for-verification).
 
-### 24.8.1. Verification debugging {#sec-verification-debugging}
+### 24.8.1. Verification debugging when verification fails {#sec-verification-debugging}
 
 Let's assume one assertion is failing ("assertion might not hold" or "postcondition might not hold"). What should you do next?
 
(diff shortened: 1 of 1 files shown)


docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
   `method m(i) returns (j: T)`<br>&nbsp;&nbsp;`  requires A(i)`<br>&nbsp;&nbsp;`  ensures B(i, j)`<br>`{`<br>&nbsp;&nbsp;`  ...`<br>`}`<br><br>`method n() {`<br>&nbsp;&nbsp;`  ...`<br><br><br>&nbsp;&nbsp;`  var x := m(a);`<br>&nbsp;&nbsp;`  assert P(x);` | `method m(i) returns (j: T)`<br>&nbsp;&nbsp;`  requires A(i)`<br>&nbsp;&nbsp;`  ensures B(i, j)`<br>`{`<br>&nbsp;&nbsp;`  ...`<br>`}`<br><br>`method n() {`<br>&nbsp;&nbsp;`  ...`<br>&nbsp;&nbsp;`  assert A(k);`<br>&nbsp;&nbsp;`  assert forall x :: B(k, x) ==> P(x);`<br>&nbsp;&nbsp;`  var x := m(k);`<br>&nbsp;&nbsp;`  assert P(x);`
 
-### 24.8.2. Assertion batches {#sec-assertion-batches}
+### 24.8.2. Verification debugging when verification is slow {#sec-verification-debugging-slow}
+
+In this section, we describe methodologies to apply in the case when verification is slower than expected, does not terminate, or timeouts.
+
+#### 24.8.2.1. `assume false;` {#sec-assume-false}
+
+Assuming `false` is a powerful way to short-circuit the verifier and stop verification at a given point[^explainer-assume-false], and since the final compilation steps do not accept this command, it is safe to use it during development.
```
robin-aws, line 324: Just FYI, I'm a little wary of documenting this so strongly since its effectiveness is dependent on undefined behavior of the verifier: https://github.com/dafny-lang/dafny/issues/1441

This looks like great content that seems to work well now, just might need a caveat like what I discuss in that issue.

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
   <br><br>`assert forall i | i == m :: P(m);` |  `assert P(m);`<br>`assert forall i | i == m :: P(i);`
   `method m(i) returns (j: T)`<br>&nbsp;&nbsp;`  requires A(i)`<br>&nbsp;&nbsp;`  ensures B(i, j)`<br>`{`<br>&nbsp;&nbsp;`  ...`<br>`}`<br><br>`method n() {`<br>&nbsp;&nbsp;`  ...`<br><br><br>&nbsp;&nbsp;`  var x := m(a);`<br>&nbsp;&nbsp;`  assert P(x);` | `method m(i) returns (j: T)`<br>&nbsp;&nbsp;`  requires A(i)`<br>&nbsp;&nbsp;`  ensures B(i, j)`<br>`{`<br>&nbsp;&nbsp;`  ...`<br>`}`<br><br>`method n() {`<br>&nbsp;&nbsp;`  ...`<br>&nbsp;&nbsp;`  assert A(k);`<br>&nbsp;&nbsp;`  assert forall x :: B(k, x) ==> P(x);`<br>&nbsp;&nbsp;`  var x := m(k);`<br>&nbsp;&nbsp;`  assert P(x);`
 
-### 24.8.2. Assertion batches {#sec-assertion-batches}
+### 24.8.2. Verification debugging when verification is slow {#sec-verification-debugging-slow}
+
+In this section, we describe methodologies to apply in the case when verification is slower than expected, does not terminate, or timeouts.
```
robin-aws, line 320: How about a first subsection that suggests verifying with more, smaller assertion batches. Personally I would suggest first trying one batch per assertion before moving on to the techniques below. You can link the reader to https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-assertion-batches for more details.

robin-aws (commented):
Looks like more excellent material, thanks for taking this on! Just left a couple of high-level suggestions for now.

docs/DafnyRef/UserGuide.md
```diff
@@ -156,9 +156,9 @@ exit code.
 
 ## 24.8. Verification
 
-In this section, we suggest a methodology to figure out [why a single assertion might not hold](#sec-verification-debugging), we explain how to [make the verifier more focused on particular assertions](#sec-assertion-batches) to devote more time to prove them, and we also give some more examples of [useful options and attributes to control verification](#sec-command-line-options-and-attributes-for-verification).
+In this section, we suggest a methodology to figure out [why a single assertion might not hold](#sec-verification-debugging), we propose methodologies to deal with [assertions that slow a proof down](#sec-verification-debugging-slow), we explain how to [verify assertions in parallel or in a focused way](#sec-assertion-batches) to devote more time to prove them, and we also give some more examples of [useful options and attributes to control verification](#sec-command-line-options-and-attributes-for-verification).
```
cpitclaudel, line 159: Why does verifying assertions in parallel devote more time to prove them?

cpitclaudel: methodologies → techniques
(A methodology (singular) is already a set of techniques, so methodologies are sets of techniques)

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
+In this section, we describe methodologies to apply in the case when verification is slower than expected, does not terminate, or timeouts.
+
+#### 24.8.2.1. `assume false;` {#sec-assume-false}
+
+Assuming `false` is a powerful way to short-circuit the verifier and stop verification at a given point[^explainer-assume-false], and since the final compilation steps do not accept this command, it is safe to use it during development.
+Another similar command, `assert false;`, would also short-circuit the verifier, but it would still make the verifier try to prove `false`, which can also lead to timeouts.
+
+[^explainer-assume-false]: `assume false` tells the Dafny verifier "Assume everything is true from this point of the program". The reason is that, 'false' proves anything. For example, `false ==> A` is always true because it is equivalent to `!false || A`, which reduces to `true || A`, which reduces to `true`.
+
+Thus, let us say a program of this shape takes forever to terminate.
```
cpitclaudel, line 329: verify, not terminate

docs/DafnyRef/UserGuide.md
cpitclaudel: I'm bothered by the fact that the compiler rejects `assume false` — it seems quite arbitrary.  I'm not sure it's worth documenting here.

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
+   }
+   assert W;
+}
+```
+
+If verification is fast, which of the two assertions `assert Z;` or `assert P;` causes the slowdown?[^answer-slowdown]
+
+[^answer-slowdown]: `assert P;`.
+
+We now hope you know enough of `assume false;` to locate an assertion that makes verification slow.
```
cpitclaudel, line 437: It's not always the case that a single assertion causes slowness, unfortunately :/  It can be multiple assertions together causing a loop.

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
+```dafny
+  assert X by {
+    assert LEMMA_PRECONDITION;
+    LEMMA();
+  }
+```
+
+Now, only `X` is available for the Dafny Verifier to prove the rest of the method.
+
+[^verifier-lost]: By default, the expression of an assertion or a precondition is added to the knowledge base of the Dafny verifier for further assertions or postconditions. However, this is not always desirable, because if the verifier has too much knowledge, it might get lost trying to prove something in the wrong direction.
```
cpitclaudel, line 468: Further *within the same basic block*

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
+
+We now hope you know enough of `assume false;` to locate an assertion that makes verification slow.
+Nest, we will describe some other strategies at the assertion level to figure out what happens and perhaps fix it.
+
+#### 24.8.2.2. `assert ... by {}` {#sec-verification-debugging-assert-by}
+
+If an assertion `assert X;` is slow, it is possible that calling a lemma or invoking other assertions can help proving it: The postcondition of this lemma, or the added assertions, could help the Dafny verifier figure out faster how to prove the result.
+
+```dafny
+  assert LEMMA_PRECONDITION;
```
cpitclaudel, line 445: This assert is unnecessary in most cases, since it's exactly the precondition of the following function.

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
+  assert i >= j by {
+    reveal greater;
+  }
+}
+```
+
+#### 24.8.2.4. Non-opaque `function method` {#sec-non-opaque-function-method}
+
+Functions are normally used for specifications, but their functional syntax is sometimes also desirable to write application code.
+However, doing so naively results in the body of a `function method Fun()` be available for every caller, which can result in the verifier to timeout or get extremely slow[^verifier-lost].
```
cpitclaudel, line 500: `can cause the verifier to time out` or `can result in the verifier timing out`

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
+```dafny
+newtype {:nativeType "char"} byte = x | 0 <= x < 256
+  var a: byte := 250;
+  var b: byte := 200;
+  var c := b - a; // OK, inferred to be a byte
+  var d := a + b; // Error: cannot prove that the result of a + b is of type `byte`.
+  var f := ((a as int + b as int) % 256) as byte // OK
+```
+
+One might consider refactoring this code into specific functions if used over and over.
```
cpitclaudel, line 535:
```suggestion
One might consider refactoring this code into separate functions if used over and over.
```

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
+
+#### 24.8.2.6. Nested loops {#sec-nested-loops}
+
+In the case of nested loops, the verifier might timeout sometimes because of the information available[^verifier-lost].
+One way to mitigate this problem, when it happens, is to isolate the inner loop byt refactoring it to another method, with suitable pre and postconditions that will usually assume and prove the invariant again.
```
cpitclaudel, line 540:
```suggestion
One way to mitigate this problem, when it happens, is to isolate the inner loop by refactoring it into a separate method, with suitable pre and postconditions that will usually assume and prove the invariant again.
```

docs/DafnyRef/UserGuide.md
```diff
@@ -315,7 +315,260 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
+ {
+   innerLoop();
+ }
+...
+method innerLoop()
+  require Y'
+  ensures Y'
+```
+
+In the nest section, when everything can be proven in a timely manner, we explain another strategy to decrease proof time by parallelizing it if needed, and making the verifier focused on certain parts.
```
cpitclaudel, line 569:
```suggestion
In the next section, when everything can be proven in a timely manner, we explain another strategy to decrease proof time by parallelizing it if needed, and making the verifier focus on certain parts.
```

docs/DafnyRef/UserGuide.md
```diff
@@ -383,7 +636,7 @@ The fundamental unit of verification in Dafny is an _assertion batch_, which con
 
 [^smaller-batches]: To create a smaller batch, Dafny duplicates the assertion batch, and transforms each assertion into an assumption into exactly one batch, so that assertions are verified only in one batch. This results in "easier" formulas for the verifier because it has less to prove, but it takes more overhead because every verification instance have a common set of axioms and there is no knowledge sharing between instances because they run independently.
```
cpitclaudel, line 637: "into an assumption into exactly one batch" this is not new but I don't understand it

docs/DafnyRef/UserGuide.md
```diff
@@ -599,8 +852,8 @@ PROVER_OPTIONS="\
   "
 
 "$DOTNET" "$BOOGIE" $BOOGIE_OPTIONS $PROVER_OPTIONS "$@"
-# Uncomment if you want to use the executable instead of the DLL
-# "$BOOGIE" $BOOGIE_OPTIONS $PROVER_OPTIONS "$@"
+#Uncomment if you want to use the executable instead of the DLL
+#"$BOOGIE" $BOOGIE_OPTIONS $PROVER_OPTIONS "$@"
```
cpitclaudel, line 856: ?

cpitclaudel (changes requested):

docs/DafnyRef/UserGuide.md
MikaelMayer: > Just FYI, I'm a little wary of documenting this so strongly since its effectiveness is dependent on undefined behavior of the verifier: https://github.com/dafny-lang/dafny/issues/1441

Interesting. What about making "assume false; X" officially recognized by the Dafny verifier (like what you suggested `unverified { X }` so that it will not emit any assert in that block?
Since it seems this works well, I'll reword to ensure I describe it as "experimental" or "know-how".

> I'm bothered by the fact that the compiler rejects assume false — it seems quite arbitrary. I'm not sure it's worth documenting here.

I think it's for the same legitimate reason as rejecting functions or methods without bodies if they are not `{:extern}`. There is nothing to back an `assume` (not just `assume false;`, but every `assume`)

docs/DafnyRef/UserGuide.md
MikaelMayer: >Why does verifying assertions in parallel devote more time to prove them?

It devotes more time if you sum all the times spent by each core used by the verifier. But I'm removing this remark as it's not very relevant.

> methodologies → techniques

Reworded. Note that I think that although a "methodology" is a "set of techniques", it's a subtype of it, because it feels like the techniques are homogeneous in some way. That's why I'll keep "methodology" for the first part, and "techniques" for the second.

docs/DafnyRef/UserGuide.md
MikaelMayer: I modified to use the plural instead of the singular.

docs/DafnyRef/UserGuide.md
MikaelMayer:
```dafny
assert SOMETHING_HELPING_TO_PROVE_LEMMA_PRECONDITION;
```

docs/DafnyRef/UserGuide.md
MikaelMayer: The space triggered the automatic numbering to complain that 2 sections were missing. Removing this space made the [regex not create a false match](https://github.com/dafny-lang/dafny/blob/master/docs/DafnyRef/Numbers.java#L16) anymore:

docs/DafnyRef/UserGuide.md
cpitclaudel: "abitrarily" is odd; why is it arbitrary?

docs/DafnyRef/UserGuide.md
MikaelMayer: THe `\h+ you mean?`That would be even worse, my workaround would not work anymore.

docs/DafnyRef/UserGuide.md
MikaelMayer: It's because it's about splitting with heuristic costs, see here:
https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-vcs_max_cost

docs/DafnyRef/UserGuide.md
cpitclaudel: No, the `h*` before the `#`

cpitclaudel: But it doesn't really matter, let's forget about this for now ^^

docs/DafnyRef/UserGuide.md
MikaelMayer: Here is my new reformulation that is less emphasizing:

"Assuming `false` is an empirical way to short-circuit the verifier and usually stop verification at a given point"

docs/DafnyRef/UserGuide.md
```diff
@@ -434,15 +434,15 @@ If verification is fast, which of the two assertions `assert Z;` or `assert P;`
 
 [^answer-slowdown]: `assert P;`.
 
-We now hope you know enough of `assume false;` to locate an assertion that makes verification slow.
+We now hope you know enough of `assume false;` to locate assertions that makes verification slow.
```
cpitclaudel, line 437:
```suggestion
We now hope you know enough of `assume false;` to locate assertions that make verification slow.
```

cpitclaudel (approved):

PR #1985, opened 2022-04-07
Feat: Added ranges to expression

MikaelMayer:
Before (possibly failing assertions are barely visible):
image

After (assertions are clearly visible)
image

Although it cannot be seen yet, related positions on postconditions also now cover an entire expression instead of just a token.

This PR:
- Adds a way for expression to expose their start token and their end token
- Offers a way to create a "RangeToken" from any expression that represents the entire range of an expression
- Also changes the tokens we use for Boogie so that they use `RangeToken` if an translator option ``ReportRange` is activated
- activates this option for the language server only for now.

In the translator, the method `GetTok` replaces `.tok` to perform the switch.

After searching for 30 minutes, I didn't find what else to change to get an entire function or method call to be entirely highlighted, any hint @RustanLeino ?

### DIFF
--- a/Source/Dafny/AST/DafnyAst.cs
+++ b/Source/Dafny/AST/DafnyAst.cs
@@ -6988,6 +6988,16 @@ namespace Microsoft.Dafny {
         }
       }
     }
+
+    [FilledInDuringResolution] private IToken rangeToken;
+    public virtual IToken RangeToken {
+      get {
+        if (rangeToken == null) {
+          rangeToken = new RangeToken(Tok, EndTok);
+        }
+        return rangeToken;
+      }
+    }
   }
 
   public class LList<T> {
--- a/Source/Dafny/AST/DafnyAst.cs
+++ b/Source/Dafny/AST/DafnyAst.cs
@@ -9028,8 +9028,8 @@ namespace Microsoft.Dafny {
 
     private RangeToken rangeToken;
 
-    // Creates a token on the entire range of the expression.
-    // Used only for error reporting.
+    /// Creates a token on the entire range of the expression.
+    /// Used only for error reporting.
     public RangeToken RangeToken {
       get {
         if (rangeToken == null) {
--- a/Source/Dafny/AST/DafnyAst.cs
+++ b/Source/Dafny/AST/DafnyAst.cs
@@ -8961,6 +8961,9 @@ namespace Microsoft.Dafny {
     // The wrapped token is the startTok
     private IToken endTok;
 
+    public IToken StartToken => WrappedToken;
+    public IToken EndToken => endTok;
+
     // Used for range reporting
     override public string val {
       get {
@@ -9086,47 +9089,37 @@ namespace Microsoft.Dafny {
       get { yield break; }
     }
 
-    private IToken startTok;
-    private IToken endTok;
+    private RangeToken rangeToken;
 
     // Creates a token on the entire range of the expression.
     // Used only for error reporting.
-    public IToken RangeToken => new RangeToken(StartTok, EndTok);
-
-    // Computes the start and end token of this expression
-    private void RecomputeStartEnd() {
-      startTok = tok;
-      endTok = tok;
-      foreach (var e in SubExpressions) {
-        if (e.tok.pos < startTok.pos) {
-          startTok = e.tok;
-        } else if (e.tok.pos > endTok.pos) {
-          endTok = e.tok;
-        }
-      }
-    }
-
-    // Never null if tok is not null
-    public IToken StartTok {
+    public RangeToken RangeToken {
       get {
-        if (startTok == null) {
-          RecomputeStartEnd();
+        if (rangeToken == null) {
+          var startTok = tok;
+          var endTok = tok;
+          foreach (var e in SubExpressions) {
+            if (e.tok.filename != tok.filename || e.IsImplicit) {
+              // Ignore auto-generated expressions, if any.
+              continue;
+            }
+
+            if (e.StartToken.pos < startTok.pos) {
+              startTok = e.StartToken;
+            } else if (e.EndToken.pos > endTok.pos) {
+              endTok = e.EndToken;
+            }
+          }
+
+          rangeToken = new RangeToken(startTok, endTok);
         }
 
-        return startTok;
+        return rangeToken;
       }
     }
 
-    // Never null if tok is not null
-    public IToken EndTok {
-      get {
-        if (endTok == null) {
-          RecomputeStartEnd();
-        }
-
-        return endTok;
-      }
-    }
+    public IToken StartToken => RangeToken.StartToken;
+    public IToken EndToken => RangeToken.EndToken;
 
     /// <summary>
     /// Returns the list of types that appear in this expression proper (that is, not including types that
--- a/Source/Dafny/AST/DafnyAst.cs
+++ b/Source/Dafny/AST/DafnyAst.cs
@@ -8957,6 +8957,22 @@ namespace Microsoft.Dafny {
     }
   }
 
+  public class RangeToken : TokenWrapper {
+    // The wrapped token is the startTok
+    private IToken endTok;
+
+    // Used for range reporting
+    override public string val {
+      get {
+        return new string(' ', endTok.pos + endTok.val.Length - pos);
+      }
+    }
+
+    public RangeToken(IToken startTok, IToken endTok) : base(startTok) {
+      this.endTok = endTok;
+    }
+  }
+
   public class NestedToken : TokenWrapper {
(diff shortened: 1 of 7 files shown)


cpitclaudel (approved):
👍 lovely

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -9070,6 +9086,48 @@ public abstract class Expression {
...
+
+    // Creates a token on the entire range of the expression.
+    // Used only for error reporting.
+    public IToken RangeToken => new RangeToken(StartTok, EndTok);
+
+    // Computes the start and end token of this expression
+    private void RecomputeStartEnd() {
+      startTok = tok;
+      endTok = tok;
+      foreach (var e in SubExpressions) {
```
keyboardDrummer, line 9100: Possibly out of scope: it would help if `SubExpressions` was a `IReadOnlyList` so you could retrieve the first and last element without having to iterate.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -9070,6 +9086,48 @@ public abstract class Expression {
...
+        if (e.tok.pos < startTok.pos) {
+          startTok = e.tok;
```
keyboardDrummer, line 9102: Shouldn't this be `e.StartTok` ?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -9070,6 +9086,48 @@ public abstract class Expression {
...
 
+    private IToken startTok;
+    private IToken endTok;
+
+    // Creates a token on the entire range of the expression.
+    // Used only for error reporting.
+    public IToken RangeToken => new RangeToken(StartTok, EndTok);
+
+    // Computes the start and end token of this expression
+    private void RecomputeStartEnd() {
```
keyboardDrummer, line 9097: Why is this called `Re` compute, isn't it only called once?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -9070,6 +9086,48 @@ public abstract class Expression {
...
+      startTok = tok;
+      endTok = tok;
+      foreach (var e in SubExpressions) {
+        if (e.tok.pos < startTok.pos) {
+          startTok = e.tok;
+        } else if (e.tok.pos > endTok.pos) {
+          endTok = e.tok;
```
keyboardDrummer, line 9104: Shouldn't this be `e.EndTok` ?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -9070,6 +9086,48 @@ public abstract class Expression {
       get { yield break; }
     }
 
+    private IToken startTok;
```
keyboardDrummer, line 9089: Why not store the RangeToken instead of two ITokens?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -9070,6 +9086,48 @@ public abstract class Expression {
...
+    private IToken endTok;
+
+    // Creates a token on the entire range of the expression.
+    // Used only for error reporting.
+    public IToken RangeToken => new RangeToken(StartTok, EndTok);
```
keyboardDrummer, line 9094: Do you think we could move all the new code to a class 

```
class SyntaxNode {
  public readonly IToken tok;

  public abstract IEnumerable<SyntaxNode> ChildNodes {get;}

  public IToken RangeToken =>

  ..
}

``` 

and then have

```
class Expression : SyntaxNode {
  public IEnumerable<SyntaxNode> ChildNodes => SubExpressions;
}

```

Or do you feel that's too early for this PR ?

Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
```diff
@@ -110,6 +111,10 @@ public ExpressionTranslator(ExpressionTranslator etran, string modifiesFrame)
         Contract.Requires(modifiesFrame != null);
       }
 
+      internal IToken GetTok(Expression e) {
```
keyboardDrummer, line 114: Why not name it `GetToken` ? Also, I'm not a fan of `e` even though this is such as short method ^^

Source/Dafny/Verifier/Translator.cs
```diff
@@ -44,6 +44,7 @@ public partial class Translator {
     public class TranslatorFlags {
       public bool InsertChecksums = 0 < DafnyOptions.O.VerifySnapshots;
       public string UniqueIdPrefix = null;
+      public bool ReportRanges = false;
```
keyboardDrummer, line 47: When do we not want to report ranges? Looks like it's turned on for the IDE but not the CLI. Why the difference?

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
```diff
@@ -582,7 +582,7 @@ modifies this
       Assert.AreEqual("This is the postcondition that might not hold.", relatedInformation[0].Message);
       Assert.AreEqual(new Range((14, 16), (14, 21)), relatedInformation[0].Location.Range);
       Assert.AreEqual("Related location", relatedInformation[1].Message);
-      Assert.AreEqual(new Range((9, 13), (9, 14)), relatedInformation[1].Location.Range);
+      Assert.AreEqual(new Range((9, 11), (9, 16)), relatedInformation[1].Location.Range);
```
keyboardDrummer, line 585: Could we have tests with more nested expressions as well?

keyboardDrummer:
Nice change!

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: You mean, we could cache the SubExpressions? That's a good point, but yes, let's leave it out of scope for this PR.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: Yes that would be even better, thanks. I'm also adding a check that the filename should be the same.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: I imagine it could be called multiple times if we modify the AST, e.g. by adding elements in a set comprehension.

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: We could as well use the bounded polymorphism pattern:

```
class SyntaxNode<T> where T: SyntaxNode {
  public readonly IToken tok;

  public abstract IEnumerable<T> Children{get;}

  public IToken RangeToken =>
  ..
}
class Expression : SyntaxNode<Expression> {
}
```

However, we should do this refactoring later, when I see a compelling reason to do so. We might as well replace the "tok" by a Format or trait to keep track of concrete syntax tree elements.

Source/Dafny/Verifier/Translator.cs
MikaelMayer: Yes, because the IDE will highlight ranges. However, in the CLI, there is a format to report positions only, so that would mean 1) to change that format to support ranges of positions, which would be cluttering 2) or we could also decide so activate this option when -showSnipptes is enabled, but it's not the option by default yet and it will require to change the way we display errors anyway.

I though we should first add these ranges in the Dafny Language Server, where they are the most useful, and then we can always do this later.

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
MikaelMayer: tests added

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -8997,6 +9026,38 @@ public abstract class Expression {
       get { yield break; }
     }
 
+    private RangeToken rangeToken;
+
+    // Creates a token on the entire range of the expression.
```
keyboardDrummer, line 9031: Nitpick: I'd prefer documentation comments `///` above members.

keyboardDrummer (approved):

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -8997,6 +9026,38 @@ public abstract class Expression {
...
+    // Used only for error reporting.
```
MikaelMayer, line 9032:
```suggestion
    /// Creates a token on the entire range of the expression.
    /// Used only for error reporting.
```

keyboardDrummer (approved):

PR #1986, opened 2022-04-08
fix: Don't use CompileName outside of the compilers

### LINKED ISSUE #1984: Dafny's verifier incorrectly uses :extern names
For example, this is rejected:

```dafny
method {:extern "test"} testInt(i: int)
method {:extern "test"} testBool(b: bool)
```

```console
$ Dafny "git-issue-1984.dfy"
git-issue-1984.dfy(5,24): Error: more than one declaration of procedure name: CheckWellformed$$_module.__default.test
git-issue-1984.dfy(5,24): Error: more than one declaration of procedure name: Call$$_module.__default.test
```

cpitclaudel:
* Source/Dafny/AST/DafnyAst.cs:

  (Declaration.SanitizedName, TypeParameter.SanitizedName,
   TupleTypeDecl.SanitizedName, NonglobalVariable.SanitizedName,
   LocalVariable.SanitizedName, Formal.SanitizedName):
  New property

  (ModuleDefinition.SanitizedName):
  New property (created by splitting off the non-extern parts of CompileName)

  (Declaration.CompileName, ModuleDefinition.CompileName):
  Use SanitizedName

  (TypeParameter.CompileName, TupleTypeDecl.CompileName,
   NonglobalVariable.CompileName, LocalVariable.CompileName):
  Alias to SanitizedName

  (ModuleExportDecl, ModuleDefinition.VisibilityScope):
  Use SanitizedName when creating the VisibilityScope

  (ModuleDefinition.FullSanitizedName, DataTypeDecl.FullSanitizedName):
  Use SanitizedName, not CompileName

  (NonglobalVariable.CompilerizeName):
  Rename to SanitizeName

  (NonglobalVariable.AssignUniqueName, LocalVariable.AssignUniqueName):
  Only assign uniqueName (compileName initialization was duplicated with the
  CompileName getter)

  (FunctionCallFinder.PrintFunctionCallGraph):
  Use SanitizedName, not CompileName

  (ModuleDefinition.RefinementCompileName,
   ModuleDefinition.FullSanitizedRefinementName, MemberDecl.FullNameInContext,
   MemberDecl.FullSanitizedRefinementName, SpecialField.FullCompileName,
   SpecialField.FullNameInContext, SpecialField.FullSanitizedRefinementName):
  Remove unused properties

* Source/Dafny/Verifier/Translator.cs (Translator.Translate):
* Source/DafnyServer/SymbolTable.cs (*):

  Use SanitizedName, not CompileName

* Source/Dafny/Resolver.cs (ResolveProgram):
  Use FullSanitizedName instead of FullCompileName when creating a signature for
  an abstract module (originally changed from .Name to .FullCompileName in
  66b9c19c9692ccece00fb6202bb0d2440999d93b)

Fixes #1984.

### DIFF
--- a/Source/Dafny/AST/DafnyAst.cs
+++ b/Source/Dafny/AST/DafnyAst.cs
@@ -2788,14 +2788,8 @@ namespace Microsoft.Dafny {
     }
 
     string compileName;
-    public string CompileName {
-      get {
-        if (compileName == null) {
-          compileName = ResolvedClass.CompileName;
-        }
-        return compileName;
-      }
-    }
+    public string CompileName => compileName ??= ResolvedClass.CompileName;
+
     public string FullCompanionCompileName {
       get {
         Contract.Requires(ResolvedClass is TraitDecl || (ResolvedClass is NonNullTypeDecl nntd && nntd.Class is TraitDecl));
@@ -3362,7 +3356,6 @@ namespace Microsoft.Dafny {
     IToken IRegion.BodyStartTok { get { return BodyStartTok; } }
     IToken IRegion.BodyEndTok { get { return BodyEndTok; } }
     string INamedRegion.Name { get { return Name; } }
-    protected string compileName;
 
     private VisibilityScope opaqueScope = new VisibilityScope();
     private VisibilityScope revealScope = new VisibilityScope();
@@ -3409,19 +3402,20 @@ namespace Microsoft.Dafny {
       return IsRevealedInScope(scope) || opaqueScope.VisibleInScope(scope);
     }
 
+    protected string sanitizedName;
+    public virtual string SanitizedName => sanitizedName ??= NonglobalVariable.SanitizeName(Name);
+
+    protected string compileName;
     public virtual string CompileName {
       get {
         if (compileName == null) {
-          string qual;
-          IsExtern(out qual, out compileName);
-          if (compileName == null) {
-            // this is the usual name
-            compileName = NonglobalVariable.CompilerizeName(Name);
-          }
+          IsExtern(out _, out compileName);
+          compileName ??= SanitizedName;
         }
         return compileName;
       }
     }
+
     public bool IsExtern(out string/*?*/ qualification, out string/*?*/ name) {
       // ensures result==false ==> qualification == null && name == null
       Contract.Ensures(Contract.Result<bool>() || (Contract.ValueAtReturn(out qualification) == null && Contract.ValueAtReturn(out name) == null));
@@ -3487,19 +3481,21 @@ namespace Microsoft.Dafny {
       }
     }
 
-    public override string CompileName {
+    public override string SanitizedName {
       get {
-        if (compileName == null) {
+        if (sanitizedName == null) {
           var name = Name;
           if (parent is MemberDecl && !name.StartsWith("_")) {
             // prepend "_" to type parameters of functions and methods, to ensure they don't clash with type parameters of the enclosing type
             name = "_" + name;
           }
-          compileName = NonglobalVariable.CompilerizeName(name);
+          sanitizedName = NonglobalVariable.SanitizeName(name);
         }
-        return compileName;
+        return sanitizedName;
       }
     }
+
+    public override string CompileName => SanitizedName; // Ignore :extern
 
     /// <summary>
     /// NonVariant_Strict     (default) - non-variant, no uses left of an arrow
@@ -3750,7 +3746,7 @@ namespace Microsoft.Dafny {
       Extends = extends;
       ProvideAll = provideAll;
       RevealAll = revealAll;
-      ThisScope = new VisibilityScope(this.FullCompileName);
+      ThisScope = new VisibilityScope(this.FullSanitizedName);
     }
 
     public ModuleExportDecl(IToken tok, string name, ModuleDefinition parent,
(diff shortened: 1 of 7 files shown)


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -4000,52 +3996,53 @@ public class ModuleDefinition : INamedRegion, IAttributeBearingDeclaration {
     }
 
     VisibilityScope visibilityScope;
+    public VisibilityScope VisibilityScope =>
```
keyboardDrummer, line 3999: With .NET source generators we should be able to write this as just 
```csharp
[Memoize] 
public VisibilityScope VisibilityScope => new VisibilityScope(this.SanitizedName)
```

But strangely it doesn't seem like there are NuGet packages with attributes like Memoize.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -5953,29 +5899,17 @@ public abstract class NonglobalVariable : IVariable {
...
-      if (uniqueName == null) {
-        uniqueName = generator.FreshId(Name + "#");
-        compileName = string.Format("_{0}_{1}", IVariable.CompileNameIdGenerator.FreshNumericId(), CompilerizeName(name));
-      }
-      return UniqueName;
+      return uniqueName ??= generator.FreshId(Name + "#");
     }
     static char[] specialChars = new char[] { '\'', '_', '?', '\\', '#' };
-    public static string CompilerizeName(string nm) {
+    public static string SanitizeName(string nm) {
```
keyboardDrummer, line 5912: Could you add a comment to this method? What does it mean to sanitize a name? Why do we need to sanitize? Apparently the verifier also needs sanitized names. Does SanitizeName remove any characters which are likely to cause issues in verification or compilation back-ends ?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -4000,52 +3996,53 @@ public class ModuleDefinition : INamedRegion, IAttributeBearingDeclaration {
...
+
     string compileName;
     public string CompileName {
       get {
         if (compileName == null) {
           var externArgs = DafnyOptions.O.DisallowExterns ? null : Attributes.FindExpressions(this.Attributes, "extern");
           if (externArgs != null && 1 <= externArgs.Count && externArgs[0] is StringLiteralExpr) {
             compileName = (string)((StringLiteralExpr)externArgs[0]).Value;
-          } else if (IsBuiltinName || externArgs != null) {
+          } else if (externArgs != null) {
```
keyboardDrummer, line 4036: This case should only happen when `externArgs.Count == 0` right? Maybe be nice to make that explicit.

keyboardDrummer (dismissed):

Source/Dafny/AST/DafnyAst.cs
cpitclaudel: Also when externArgs[1] is something else than a string, I think?  Though that should probably not be accepted.

Source/Dafny/AST/DafnyAst.cs
cpitclaudel: IIUC what this does is indeed escape some special chars that shouldn't be emitted into Boogie. (The notion of "sanitized" names already existed, so I just renamed the function for consistency.)

As for compilation versus verification, I'm less clear on that.  We already have idProtect on the compiler side, so I'm guessing SanitizeName isn't strictly needed there.

Source/Dafny/AST/DafnyAst.cs
cpitclaudel: Agreed, and I was disappointed to not find one :'(

Source/Dafny/AST/DafnyAst.cs
cpitclaudel: 6e69a31ea0703e3dc35b103a9f85cf0286f4d49b was where it was originally introduced, and that was specifically for C#.  Then the translator was changed to use "SanitizedName" (an alias of CompileName, more or less) in a8b18b984b289be43c112ca6f0fb2283a8ddc13a. IdProtect was introduced in a939eff5 to sanitize language-specific keywords.

cpitclaudel: I added a docstring.

keyboardDrummer (approved):

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: > Also when externArgs[1] is something else than a string, I think? Though that should probably not be accepted.

I'll leave it up to you to decide whether to make that more strict :-) From the documentation it doesn't look like a non-string argument is accepted.

cpitclaudel: We can fix this separately, I think; I prefer for this PR to not change existing behaviors.

keyboardDrummer: > We can fix this separately, I think; I prefer for this PR to not change existing behaviors.

That's fair

hirataqdees:
Hi, 
Does this merged issue solve the error: `modules 'X' and 'X' both have CompileName 'X_Compile'` for
`module {:extern} New.X {},  module {:extern} Old.X {}`

keyboardDrummer:
> Hi, Does this merged fix solve the error: `modules 'X' and 'X' both have CompileName 'X_Compile'` for `module {:extern} New.X {}, module {:extern} Old.X {}` ?

I don't think so, but this fix is available in the latest Dafny so please see if your issue is resolved there.

hirataqdees:
Hi Remy, thanks for letting me know.

cpitclaudel:
That issue was fixed in a different PR: https://github.com/dafny-lang/dafny/pull/1998

PR #1997, opened 2022-04-08
Fix: Emit constraints for compiled quantifier expressions, and nested subset types

### LINKED ISSUE #1604: Reference types in comprehension soundness issue
```
trait Tr { }
class A extends Tr { }
class B extends Tr { }

function method test(x: Tr): bool
  requires x is A
{
  if x is B then 1/0 == 0 else true
}

method Main() {
  var a := new A;
  var b := new B;
  var s: set<Tr> := {a, b};
  var aa := forall a': A :: a' in s ==> test(a');
  assert(aa);
  print "ok";
}
```

In the case above, the verifier passes and
1) C#'s compilation complains that `error CS1503: Argument 1: cannot convert from 'System.Collections.Generic.IEnumerable<_module.Tr>' to 'System.Collections.Generic.IEnumer
able<_module.A>'`, which is something that @fabiomadge  is working on right now, so that could explain why nobody found this error before.
2) JavaScript it silently fails and "ok" is never printed;
3) Java: `Exception in thread "main" java.lang.ArithmeticException: BigInteger divide by zero`
4) Go: `Running... [Program halted] reflect: Call using *main.B as type *main.A` and nothing else displayed

The problem comes again from the fact that the type inference infers that a' is of type A and the bounds are a set of type Tr, and whereas there is some check for the foreach statement in every compiler, there is no check for compiling the forall because it compiles to a true mapping expression.

I'd say, all these errors would be automatically fixed if I implement [my right solution.](https://github.com/dafny-lang/dafny/issues/697#issuecomment-970703632).

_Originally posted by @MikaelMayer in https://github.com/dafny-lang/dafny/pull/1522#discussion_r753572058_

MikaelMayer:
Fixes #1604


Source/Dafny/ConcreteSyntax/ConcreteSyntaxTree.cs
```diff
@@ -32,6 +32,12 @@ public class ConcreteSyntaxTree : ICanRender {
       return result;
     }
 
+    public void Empty() {
```
keyboardDrummer, line 35: Could `Clear` be a more recognisable name?

Source/Dafny/Compilers/Compiler-go.cs
```diff
@@ -1920,35 +1921,62 @@ private class SpecialNativeType : UserDefinedType {
...
+            $"_dafny.InstanceOfTrait/*1*/({tmpVarName}.(_dafny.TraitOffspring), {TypeName_Companion(trait, wPreconditions, tok)}.TraitID_)");
         } else {
-          typeTest = $"_dafny.InstanceOf({tmpVarName}, ({TypeName(boundVarType, wBody, tok)})(nil))";
-          wIf.WriteLine("{0}, {1} = {2}.({3})", boundVarName, okVar, tmpVarName, TypeName(boundVarType, wIf, tok));
-          wIf.WriteLine("if !{0} {{ continue }}", okVar);
+          var typeAssertSucceeds = idGenerator.FreshId("_typeAssertSucceeds");
+          wPreconditions.WriteLine(
+            $@"{typeAssertSucceeds} := func(param interface{{}}) bool {{ var ok bool; _, ok = param.({TypeName(boundVarType, wPreconditions, tok)}); return ok}}");
+          conditions.Add($"{typeAssertSucceeds}({tmpVarName})");
+          // When Go will support generics
```
keyboardDrummer, line 1946: I'd leave out comments such as "change X when Y happens". I don't expect they'll make change X easier when we get to Y because we'll have forgotten about the comment, and maybe the comment never gets found and then it's just bloat.

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -4831,29 +4867,64 @@ private class GetterSetterLvalueImpl : ILvalue {
...
+        // That's an issue. We are iterating values that need to be checked.
+        var preconditions = wr.Fork();
+        var conditions = GetSubtypeCondition(tmpVarName, boundVarType, tok, preconditions);
+        if (conditions == null) {
+          preconditions.Empty();
+        } else {
+          var thenWriter = EmitIf(out var guardWriter, isReturning, wr);
+          guardWriter.Write(conditions);
+          if (isReturning) {
+            // What do we put in the else branch
```
keyboardDrummer, line 4884: I don't get the comment.

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -4831,29 +4867,64 @@ private class GetterSetterLvalueImpl : ILvalue {
...
-        {
-          TypeArgs: var typeArgs,
-          ResolvedClass:
-          SubsetTypeDecl
-          {
-            TypeArgs: var typeParametersArgs,
-            Var: var var,
-            ConstraintIsCompilable: true,
-            Constraint: var constraint
+    private ConcreteSyntaxTree MaybeInjectSubtypeConstraint(string tmpVarName,
```
keyboardDrummer, line 4870: Could you add a comment with an example of what it creates?

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -4831,29 +4867,64 @@ private class GetterSetterLvalueImpl : ILvalue {
...
+      Type collectionElementType, Type boundVarType, bool inLetExprBody,
+      Bpl.IToken tok, ConcreteSyntaxTree wr, bool isReturning = false, bool elseReturnValue = false
+      ) {
+      if (IsTargetSupertype(collectionElementType, boundVarType)) {
+        // That's an issue. We are iterating values that need to be checked.
```
keyboardDrummer, line 4875: Instead of the comment, consider introducing a variable like 

`var iterationValuesNeedToBeChecked = IsTargetSupertype(collectionElementType, boundVarType)`

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1865,29 +1866,39 @@ protected class ClassWriter : IClassWriter {
       return string.Format($"{DafnyHelpersClass}.Quantifier<{bvType}>");
     }
 
-    protected override ConcreteSyntaxTree CreateForeachLoop(string tmpVarName, Type collectionElementType, string boundVarName, Type boundVarType, bool introduceBoundVar,
-      Bpl.IToken tok, out ConcreteSyntaxTree collectionWriter, ConcreteSyntaxTree wr) {
+    protected override ConcreteSyntaxTree CreateForeachLoop(string tmpVarName, Type collectionElementType, Bpl.IToken tok, out ConcreteSyntaxTree collectionWriter, ConcreteSyntaxTree wr) {
       collectionWriter = new ConcreteSyntaxTree();
       wr.Format($"foreach ({TypeName(collectionElementType, wr, tok)} {tmpVarName} in {collectionWriter})");
       var wwr = wr.NewBlock();
```
keyboardDrummer, line 1872: What does `wwr` mean? Seems better to inline this variable anyways since it's immediately returned.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1865,29 +1866,39 @@ protected class ClassWriter : IClassWriter {
...
+      return wwr;
+    }
+
+    protected override ConcreteSyntaxTree EmitDowncastVariableAssignment(string boundVarName, Type boundVarType, string tmpVarName,
+      Type collectionElementType, bool introduceBoundVar, Bpl.IToken tok, ConcreteSyntaxTree wwr) {
```
keyboardDrummer, line 1877: Could you rename `wwr` to something recognisable? Why does this method return one of its parameters?

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -1865,29 +1866,39 @@ protected class ClassWriter : IClassWriter {
...
+          typeTest = $"{tmpVarName} is {TypeName(boundVarType, wPreconditions, tok)}";
         }
         if (boundVarType.IsNonNullRefType) {
           typeTest = $"{tmpVarName} != null && {typeTest}";
         } else {
           typeTest = $"{tmpVarName} == null || {typeTest}";
         }
-        wwr = wwr.NewBlock($"if ({typeTest})");
+      } else {
+        typeTest = "true";
```
keyboardDrummer, line 1898: Could we also assign `null` here and on line 1888, and replace line 1901 with `return typeTest` ?

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -548,20 +548,52 @@ protected interface IClassWriter {
...
     ///     }
     /// where
     ///   * "[[collectionWriter]]" is the writer returned as "collectionWriter"
     ///   * "[[bodyWriter]]" is the block writer returned
+    /// </summary>
+    protected abstract ConcreteSyntaxTree CreateForeachLoop(
+      string tmpVarName, Type collectionElementType,
+      Bpl.IToken tok, out ConcreteSyntaxTree collectionWriter, ConcreteSyntaxTree wr);
+
+    private ConcreteSyntaxTree CreateGuardedForeachLoop(
```
keyboardDrummer, line 561: Could you add a C# example of what this creates in a comment? It's difficult to read the code without knowing what the goal is.

keyboardDrummer (commented):
Looks great!

It would have been easier to review if the refactoring was in a separate PR/commit, but that's OK. I can review the end result.

Source/Dafny/Compilers/SinglePassCompiler.cs
MikaelMayer: > I don't get the comment.

Removed, it was obsolete

MikaelMayer:
Ok, one of the tests revealed why this PR was intricated with the previous PR. Here is the link:

```dafny
type Small = x: int | 0 <= x < 100 && x != 3

function method F(x: int): int
  requires x != 3
{
  if x == 3 then 1/0 else 100
}

method Main() {
  var b := !exists n: Small :: F(n) != 100;
  assert b;
  print b;
}
```

The "exists", typechecks and verifies, but crashes at run-time, even with this PR.
The reason for that is that "n" is wrongly inferred of type "Small", so the compiler doesn't think there is something which needs to be done.
To really solve that problem, one has to consider type-checking "n" differently in the range, but doing so would break existing tests, because wto determine whether to type check "n" differently, we need to know if we are in a ghost context, which is exactly what https://github.com/dafny-lang/dafny/pull/1702 does.

For now, in this PR, I will ensure the compiler always emit the type constraint when it's compilable.
I will remove this check in the other PR because it'll no longer needed.

keyboardDrummer:
> Ok, one of the tests revealed why this PR was intricated with the previous PR. Here is the link:
> 
> ```
> type Small = x: int | 0 <= x < 100 && x != 3
> 
> function method F(x: int): int
>   requires x != 3
> {
>   if x == 3 then 1/0 else 100
> }
> 
> method Main() {
>   var b := !exists n: Small :: F(n) != 100;
>   assert b;
>   print b;
> }
> ```
> 
> The "exists", typechecks and verifies, but crashes at run-time, even with this PR. The reason for that is that "n" is wrongly inferred of type "Small", so the compiler doesn't think there is something which needs to be done. To really solve that problem, one has to consider type-checking "n" differently in the range, but doing so would break existing tests, because wto determine whether to type check "n" differently, we need to know if we are in a ghost context, which is exactly what #1702 does.
> 
> For now, in this PR, I will ensure the compiler always emit the type constraint when it's compilable. I will remove this check in the other PR because it'll no longer needed.

If we're compiling a comprehension, then we're never in a ghost context right? Shouldn't `BoundedPool` have a `Type` field that returns the type of values created by `CompileCollection(BoundedPool ..)` ?

MikaelMayer:
> If we're compiling a comprehension, then we're never in a ghost context right? Shouldn't BoundedPool have a Type field that returns the type of values created by CompileCollection(BoundedPool ..), so we could assign var collectionElementType = boundedPool.Type ?

Right, and after your remark, I saw that this bounded pool ought to be better computed for integer pools as well, else the subset type was taken instead. I refactored the code so that now ElementType is a method of a pool to ensure it's implemented  by every pool

keyboardDrummer (dismissed):

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -548,20 +548,66 @@ protected interface IClassWriter {
...
+    protected abstract string GetSubtypeCondition(
+      string tmpVarName, Type boundVarType, Bpl.IToken tok, ConcreteSyntaxTree wPreconditions);
+
+    /// <summary>
+    /// Emit an (already verified) downcast assignment like:
+    ///     var boundVarName:boundVarType := tmpVarName as boundVarType;
+    ///     [[bodyWriter]]
+    /// </summary>
+    /// where
+    ///   * "[[bodyWriter]]" is the block writer returned
```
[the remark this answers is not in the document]
MikaelMayer: `* "[[bodyWriter]]" is where the writer wr's position will be next`

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -548,20 +548,66 @@ protected interface IClassWriter {
...
     /// Option:
```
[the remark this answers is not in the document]
MikaelMayer:
```cs
    /// <summary>
    /// Emit an (already verified) downcast assignment like:
    /// 
    ///     var boundVarName:boundVarType := tmpVarName as boundVarType;
    ///     [[bodyWriter]]
    /// 
    /// where
    ///   * "[[bodyWriter]]" is where the writer wr's position will be next
    /// </summary>
    /// <param name="boundVarName">Name of the variable after casting</param>
    /// <param name="boundVarType">Expected variable type</param>
    /// <param name="tmpVarName">The collection's variable name</param>
    /// <param name="collectionElementType">type this variable is casted from, in case it is useful</param>
    /// <param name="introduceBoundVar">Whether or not to declare the variable, in languages requiring declarations</param>
    /// <param name="tok">A position in the AST</param>
    /// <param name="wr">The concrete syntax tree writer</param>
    ```

PR #1999, opened 2022-04-08
compiler: Use name DafnyRuntime, not Dafny

cpitclaudel:
This makes it easier to link the output of the compiler into Dafny (otherwise
the runtime pollutes the main Dafny namespace).

### DIFF
--- a/Source/DafnyRuntime/DafnyRuntime.cs
+++ b/Source/DafnyRuntime/DafnyRuntime.cs
@@ -18,7 +18,7 @@ namespace DafnyAssembly {
   }
 }
 
-namespace Dafny {
+namespace DafnyRuntime {
   using System.Collections.Generic;
   using System.Collections.Immutable;
   using System.Linq;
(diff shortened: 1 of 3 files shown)


Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -18,7 +18,7 @@ public class DafnySourceAttribute : Attribute {
   }
 }
 
-namespace Dafny {
+namespace DafnyRuntime {
```
MikaelMayer, line 21: I would be strongly in favor or having a sub-namespace rather than a concatenated namespace, so for example:
* `Dafny.Runtime`
* `Dafny.Run`
* `Dafny.VM`
* `Dafny.Library`
* `Dafny.Lib`
* `Dafny.System`
* `Dafny.Core`
* `Dafny.Include`

What do you think of that?

MikaelMayer (commented):
Great ! I clearly see how doing such a renaming would be beneficial for not colliding names in C#.
I'm just wondering whether it's the right name, ses my suggestion.

PR #2000, opened 2022-04-08
Add limited support for automatically generating Dafny traits from C# classes

cpitclaudel:
This is the first real Dafny-in-Dafny PR.  Still needs tests (draft for now).  Co-authored by @atomb. 

Just 48 issues/PRs to go to the next round number!


robin-aws:
Assigning myself as a reviewer as I'm particularly interested in this feature even beyond the Dafny-in-Dafny work. I wonder if we could clean up the external code in https://github.com/dafny-lang/dafny-reportgenerator with this? :)

cpitclaudel:
> I wonder if we could clean up the external code in https://github.com/dafny-lang/dafny-reportgenerator with this? :)

Probably not; this doesn't export functions

Source/AutoExtern/AutoExtern.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
keyboardDrummer, line 3: Is there a good place to add a description of a .NET project? I think it would be nice to easily be able to see what AutoExtern is for.

cpitclaudel: Not that I know of; I used a README instead.

cpitclaudel: It would be nice to have a project-wide XML doc comment, but I don't think that exists

cpitclaudel: Wait, I had forgotten to push the README ! :) There was one in the tutorial but not one for the code.  Pushed now; https://github.com/dafny-lang/dafny/blob/9b6b22b673c983767f924f49c88436f89d0be6d0/Source/AutoExtern/README.md

Source/AutoExtern/Program.cs
```diff
@@ -0,0 +1,414 @@
...
+  private IEnumerable<Variable> Variables =>
+    syntax.Declaration.Variables.Where(v => model.IsPublic(v))
+      .Select(v => new Variable(syntax.Declaration.Type, v, model));
+
+  public override void Pp(TextWriter wr, string indent) {
+    PpChildren(wr, indent, Variables);
+  }
+}
+
+class Property : PrettyPrintable {
```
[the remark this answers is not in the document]
cpitclaudel: That's a good point.  I have the opposite issue: my editor isn't fancy enough, so if everything is in one file I can trivially search around, rename functions, search and replace, etc; whereas otherwise I need advanced support for navigation and refactoring.  Same for errors: if I change something I can see everything that breaks from the change in one file, rather than having to visit them all or recompile.

(Also, I usually dislike splitting because it loses history, but since our merge process already does that it's not an issue here.)

Still, I wouldn't mind splitting this one: would you just move the abstract class and keep all the nodes together, or would you do one file per node class?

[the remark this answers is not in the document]
keyboardDrummer: Given that they're large enough to fill a screen, I'd put the abstract class and subclasses in separate files. However, I recognise this is subjective. I can only mention this was the norm in the two C# dev teams I've worked in, both of which were using Visual Studio with ReSharper as an IDE.

Test/lit.site.cfg
```diff
@@ -163,21 +158,19 @@ else:
...
 
-for root in potentialSolverRoots:
-    for executable in potentialSolverExecutables:
-        if not z3Path:
-            potentialPath = path.join(root, executable)
-            if path.exists(potentialPath):
-                z3Path = potentialPath
+solverPath = \
+    lit.util.which("z3", solverRoots) or \
+    lit.util.which("cvc4", solverRoots)
```
atomb, line 169: I wonder how long `cvc4` has been in here. Recent versions of Dafny and Boogie don't work with it (though Boogie often works with `cvc5`). It's probably not urgent to change this, but we may want to eventually.

customBoogie.patch
```diff
@@ -1,10 +1,10 @@
 diff --git a/Source/Dafny.sln b/Source/Dafny.sln
-index 36bb4d93b..0dea612cf 100644
+index 88200dcc4..3a75b0cc5 100644
 --- a/Source/Dafny.sln
 +++ b/Source/Dafny.sln
-@@ -30,6 +30,34 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DafnyTestGeneration", "Dafn
+@@ -34,6 +34,34 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoExtern", "AutoExtern\Au
```
atomb, line 5: This change doesn't seem right. I think this patch should mostly just add entries.

customBoogie.patch
```diff
@@ -1,10 +1,10 @@
...
  EndProject
- Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DafnyTestGeneration.Test", "DafnyTestGeneration.Test\DafnyTestGeneration.Test.csproj", "{896E7F24-FD59-4B34-A1BF-53C51DDBC9E9}"
+ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoExtern.Test", "AutoExtern.Test\AutoExtern.Test.csproj", "{A47E2F45-DEA3-4700-A82F-9506FEEB199A}"
```
atomb, line 7: Same for this change.

Source/AutoExtern/README.md
```diff
@@ -0,0 +1,94 @@
...
+want to see what it can do.
+
+## Usage
+
+```
+dotnet run <Project.csproj> <SourceFile.cs> <RootNamespace> <Template.dfy> \
+    <CSharpModel.dfy> <Output.dfy>
+```
+
+The tool takes size arguments:
```
atomb, line 22:
```suggestion
The tool takes six arguments:
```

atomb (changes requested):
This looks very nice! It looks to me that `customBoogie.patch` might have some spurious changes in it, though.

customBoogie.patch
cpitclaudel: You're looking at a second-order diff (a diff of a patch); adding entries changes the context in which the patch applies, hence the change.

customBoogie.patch
cpitclaudel: Same here, it's right

customBoogie.patch
atomb: Ah, right, of course.

atomb (dismissed):

atomb:
There's one remaining `dotnet-format` complaint, but I think that's the only thing I can see to change before merging.

keyboardDrummer (approved):

PR #2003, opened 2022-04-11
chore: Dafny Server documentation

MikaelMayer:
I just added some basic documentation about the Dafny Server.

### DIFF
--- a/docs/DafnyRef/UserGuide.md
+++ b/docs/DafnyRef/UserGuide.md
@@ -107,7 +107,52 @@ Before Dafny [implemented](https://github.com/dafny-lang/dafny/tree/master/Sourc
 
 The Dafny Server has [integration tests](https://github.com/dafny-lang/dafny/tree/master/Test/server) that serve as the basis of the documentation.
 
-The executable DafnyServer is a simple read-eval loop. For example, if you compile and run `DafnyServer.exe`, you could paste the following command:
+The server is essentially a REPL, which produces output in the same format as the Dafny CLI; clients thus
+do not need to understand the internals of Dafny's caching.  A typical editing session proceeds as follows:
+
+* When a new Dafny file is opened, the editor starts a new instance of the
+  Dafny server.  The cache is blank at that point.
+* The editor sends a copy of the buffer for initial verification.  This takes
+  some time, after which the server returns a list of errors.
+* The user makes modifications; the editor periodically sends a new copy of
+  the buffer's contents to the Dafny server, which quickly returns an updated
+  list of errors.
+
+The client-server protocol is sequential, uses JSON, and works over ASCII
+pipes by base64-encoding utf-8 queries.  It defines one type of query, and two
+types of responses:
+
+Queries are of the following form:
+
+     verify
+     <base64 encoded JSON payload>
+     [[DAFNY-CLIENT: EOM]]
+
+Responses are of the following form:
+
+     <list of errors and usual output, as produced by the Dafny CLI>
+     [SUCCESS] [[DAFNY-SERVER: EOM]]
+
+or
+
+     <error message>
+     [FAILURE] [[DAFNY-SERVER: EOM]]
+
+The JSON payload is an utf-8 encoded string resulting of the serialization of
+a dictionary with 4 fields:
+   * args:   An array of Dafny arguments, as passed to the Dafny CLI
+   * source: A Dafny program, or the path to a Dafny source file.
+   * sourceIsFile: A boolean indicating whether the 'source' argument is a
+                   Dafny program or the path to one.
+   * filename:     The name of the original source file, to be used in error
+                   messages
+
+For small files, embedding the Dafny source directly into a message is
+convenient; for larger files, however, it is generally better for performance
+to write the source snapshot to a separate file, and to pass that to Dafny
+by setting the 'sourceIsFile' flag to true.
+
+For example, if you compile and run `DafnyServer.exe`, you could paste the following command:
 
 ```dafny
 verify
@@ -139,7 +184,7 @@ btoa(JSON.stringify({
 ```
 
 Thus to decode such output, you'd manually use `JSON.parse(atob(payload))`.
-The Dafny Server is still supported, but the [Language Server](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) along with the initial [IDE extension for VSCode](https://marketplace.visualstudio.com/items?itemName=dafny-lang.ide-vscode) provides more features, such as Ghost highlighting or symbol hovering.
+The Dafny Server is still supported, but we recommend using the [Language Server](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) that is both available in Emacs and [VSCode](https://marketplace.visualstudio.com/items?itemName=dafny-lang.ide-vscode). The Language Server provides also features such as ghost highlighting or symbol hovering.
 
 ## 24.7. Using Dafny From the Command Line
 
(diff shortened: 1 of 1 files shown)


docs/DafnyRef/UserGuide.md
```diff
@@ -103,7 +103,43 @@ on the [Dafny INSTALL page in the wiki](https://github.com/dafny-lang/dafny/wiki
...
+verify
+eyJhcmdzIjpbIi9jb21waWxlOjAiLCIvcHJpbnRUb29sdGlwcyIsIi90aW1lTGltaXQ6MjAiXSwi
+ZmlsZW5hbWUiOiJ0cmFuc2NyaXB0Iiwic291cmNlIjoibWV0aG9kIEEoYTppbnQpIHJldHVybnMg
+KGI6IGludCkge1xuICBiIDo9IGE7XG4gIGFzc2VydCBmYWxzZTtcbn1cbiIsInNvdXJjZUlzRmls
+ZSI6ZmFsc2V9
+[[DAFNY-CLIENT: EOM]]
+```
+
+The interpreter sees the command `verify`, and then starts reading every line until it sees `[[DAFNY-CLIENT: EOM]]`
+The payload is a base64 encoded string that you could encode or decode using JavaScript's `atob` and `btoa` function.
```
cpitclaudel, line 122: Why mention javascript here?

docs/DafnyRef/UserGuide.md
```diff
@@ -103,7 +103,43 @@ on the [Dafny INSTALL page in the wiki](https://github.com/dafny-lang/dafny/wiki
...
+`method A(a:int) returns (b: int) {
+   b := a;
+   assert false;
+}
+`,"sourceIsFile": false}))
+=== "eyJhcmdzIjpbIi9jb21waWxlOjAiLCIvcHJpbnRUb29sdGlwcyIsIi90aW1lTGltaXQ6MjAiXSwiZmlsZW5hbWUiOiJ0cmFuc2NyaXB0Iiwic291cmNlIjoibWV0aG9kIEEoYTppbnQpIHJldHVybnMgKGI6IGludCkge1xuICBiIDo9IGE7XG4gIGFzc2VydCBmYWxzZTtcbn1cbiIsInNvdXJjZUlzRmlsZSI6ZmFsc2V9"
+```
+
+Thus to decode such output, you'd manually use `JSON.parse(atob(payload))`.
+The Dafny Server is still supported, but the [Language Server](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) along with the initial [IDE extension for VSCode](https://marketplace.visualstudio.com/items?itemName=dafny-lang.ide-vscode) provides more features, such as Ghost highlighting or symbol hovering.
```
cpitclaudel, line 142: I would go for "we recommend using the LSP plugin", without a mention of the editor, since we have supprot for LSP in Emacs as well.

If we get into a comparison we'll have to mention the fact that the LSP server is less robust, which doesn't look very good.

cpitclaudel (approved):
Looks nice, some small suggestions.  BTW, we already has this doc in boogie-friends:

The server is essentially a
REPL, which produces output in the same format as the Dafny CLI; clients thus
do not need to understand the internals of Dafny's caching.  A typical editing
session proceeds as follows:

* When a new Dafny file is opened, the editor starts a new instance of the
  Dafny server.  The cache is blank at that point.
* The editor sends a copy of the buffer for initial verification.  This takes
  some time, after which the server returns a list of errors.
* The user makes modifications; the editor periodically sends a new copy of
  the buffer's contents to the Dafny server, which quickly returns an updated
  list of errors.

The client-server protocol is sequential, uses JSON, and works over ASCII
pipes by base64-encoding utf-8 queries.  It defines one type of query, and two
types of responses:

Queries are of the following form:

     verify
     <base64 encoded JSON payload>
     [[DAFNY-CLIENT: EOM]]

Responses are of the following form:

     <list of errors and usual output, as produced by the Dafny CLI>
     [SUCCESS] [[DAFNY-SERVER: EOM]]

or

     <error message>
     [FAILURE] [[DAFNY-SERVER: EOM]]

The JSON payload is an utf-8 encoded string resulting of the serialization of
a dictionary with 4 fields:
   * args:   An array of Dafny arguments, as passed to the Dafny CLI
   * source: A Dafny program, or the path to a Dafny source file.
   * sourceIsFile: A boolean indicating whether the 'source' argument is a
                   Dafny program or the path to one.
   * filename:     The name of the original source file, to be used in error
                   messages

For small files, embedding the Dafny source directly into a message is
convenient; for larger files, however, it is generally better for performance
to write the source snapshot to a separate file, and to pass that to Dafny
by setting the 'sourceIsFile' flag to true.

docs/DafnyRef/UserGuide.md
MikaelMayer: Because that's why I used to understand what these server commands were, it's something everyone can test with.

docs/DafnyRef/UserGuide.md
```diff
@@ -103,7 +103,88 @@ on the [Dafny INSTALL page in the wiki](https://github.com/dafny-lang/dafny/wiki
...
+`method A(a:int) returns (b: int) {
+   b := a;
+   assert false;
+}
+`,"sourceIsFile": false}))
+=== "eyJhcmdzIjpbIi9jb21waWxlOjAiLCIvcHJpbnRUb29sdGlwcyIsIi90aW1lTGltaXQ6MjAiXSwiZmlsZW5hbWUiOiJ0cmFuc2NyaXB0Iiwic291cmNlIjoibWV0aG9kIEEoYTppbnQpIHJldHVybnMgKGI6IGludCkge1xuICBiIDo9IGE7XG4gIGFzc2VydCBmYWxzZTtcbn1cbiIsInNvdXJjZUlzRmlsZSI6ZmFsc2V9"
+```
+
+Thus to decode such output, you'd manually use `JSON.parse(atob(payload))`.
+The Dafny Server is still supported, but we recommend using the [Language Server](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) that is both available in Emacs and [VSCode](https://marketplace.visualstudio.com/items?itemName=dafny-lang.ide-vscode). The Language Server provides also features such as ghost highlighting or symbol hovering.
```
cpitclaudel, line 187:
```suggestion
The Dafny Server is still supported, but we recommend using the [Language Server](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) that is both available in Emacs and [VSCode](https://marketplace.visualstudio.com/items?itemName=dafny-lang.ide-vscode). The Language Server also provides features such as ghost highlighting or symbol hovering.
```

docs/DafnyRef/UserGuide.md
```diff
@@ -103,7 +103,88 @@ on the [Dafny INSTALL page in the wiki](https://github.com/dafny-lang/dafny/wiki
 
 ## 24.6. The Dafny Server
 
-TO BE WRITTEN
+Before Dafny [implemented](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) the official [Language Server Protocol](https://microsoft.github.io/language-server-protocol/), it implemented its own protocol to offer the [emacs mode](https://github.com/boogie-org/boogie-friends), which resulted in a project called [DafnyServer](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyServer).
```
cpitclaudel, line 106:
```suggestion
Before Dafny [implemented](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) the official [Language Server Protocol](https://microsoft.github.io/language-server-protocol/), it implemented its own protocol for [Emacs](https://github.com/boogie-org/boogie-friends), which resulted in a project called [DafnyServer](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyServer).
```

cpitclaudel (approved):

PR #2004, opened 2022-04-11
Incrementally publish verification diagnostics

keyboardDrummer:
### Changes

- Instead of waiting for all Boogie implementations to be verified before publishing verification diagnostics, publish them after each implementation finishes verifying.

### Testing

- Added test IncrementalVerificationDiagnostics


Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
```diff
@@ -20,6 +20,21 @@ public class ClientBasedLanguageServerTest : DafnyLanguageServerTestBase {
...
+    await client.WaitForNotificationCompletionAsync(documentItem.Uri, cancellationToken);
+    var document = await Documents.GetVerifiedDocumentAsync(documentItem);
+    await diagnosticReceiver.AwaitNextDiagnosticsAsync(cancellationToken); // Get resolution diagnostics
+    var remainingDiagnostics = expectedNumber ?? Int32.MaxValue;
+    var result = await diagnosticReceiver.AwaitNextDiagnosticsAsync(cancellationToken); // Get first verification diagnostics
+    remainingDiagnostics--;
+    while (!document!.Diagnostics.SequenceEqual(result) && remainingDiagnostics > 0) {
+      result = await diagnosticReceiver.AwaitNextDiagnosticsAsync(cancellationToken);
+      remainingDiagnostics--;
+    }
```
cpitclaudel, line 33: Can this be written as a `do` loop?

Source/DafnyLanguageServer.Test/Various/SlowVerifier.cs
```diff
@@ -1,34 +1,50 @@
...
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various;
 
-/// If any top-level declaration has an attribute `{:neverVerify}`,
-/// this verifier will return a task that only completes when cancelled
-/// which can be useful to test against race conditions
 class SlowVerifier : IProgramVerifier {
-  public SlowVerifier(ILogger<DafnyProgramVerifier> logger, IOptions<VerifierOptions> options) {
+  public SlowVerifier(ILogger<IProgramVerifier> logger, IOptions<VerifierOptions> options) {
```
cpitclaudel, line 13: Shouldn't this be `SlowVerifier`?

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -83,8 +81,6 @@ private class CounterExampleLoader {
 
       private IEnumerable<CounterExampleItem> GetCounterExamples(DafnyModel model) {
         return model.States
-          .WithCancellation(cancellationToken)
```
cpitclaudel, line 86: Why can this be removed?

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -67,71 +69,35 @@ public class DafnyProgramVerifier : IProgramVerifier {
         : Convert.ToInt32(options.VcsCores);
     }
 
-    public async Task<VerificationResult> VerifyAsync(Dafny.Program program,
+    // 256MB
+    private const int MaxStackSize = 0x10000000;
```
cpitclaudel, line 73:
```suggestion
    private const int MaxStackSize = 256 * 1024 * 1024;
```

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -55,15 +58,18 @@ public class DiagnosticErrorReporter : ErrorReporter {
...
-        new Diagnostic {
-          Severity = DiagnosticSeverity.Error,
-          Message = error.Msg,
-          Range = error.Tok.GetLspRange(),
-          RelatedInformation = relatedInformation,
-          Source = VerifierMessageSource.ToString()
-        },
-        GetDocumentUriOrDefault(error.Tok)
+        diagnostic,
+        uri
```
cpitclaudel, line 72: FWIW, I'm not sure this small refactoring helps; it's convenient to see values in context.

Source/DafnyLanguageServer/Language/IProgramVerifier.cs
```diff
@@ -14,7 +15,7 @@ public interface IProgramVerifier {
     /// <returns>The result of the verification run.</returns>
     /// <exception cref="System.OperationCanceledException">Thrown when the cancellation was requested before completion.</exception>
     /// <exception cref="System.ObjectDisposedException">Thrown if the cancellation token was disposed before the completion.</exception>
-    Task<VerificationResult> VerifyAsync(Dafny.Program program,
+    IReadOnlyList<IImplementationTask> Verify(Dafny.Program program,
```
cpitclaudel, line 18: Why a list rather than an enumerable?

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -26,10 +26,11 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
     }
 
     private void PublishDocumentDiagnostics(DafnyDocument document) {
+      Console.WriteLine($"Publishing document version {document.Version} with {document.Diagnostics.Count()} diagnostics.");
```
cpitclaudel, line 29: Was this supposed to remain here?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -69,15 +69,15 @@ IRelocator relocator
     public IObservable<DafnyDocument> OpenDocument(TextDocumentItem document) {
       var cancellationSource = new CancellationTokenSource();
       var resolvedDocument = OpenAsync(document, cancellationSource.Token);
-      var verifiedDocument = VerifyAsync(resolvedDocument, VerifyOnOpen, cancellationSource.Token);
+      var verifiedDocuments = Verify(resolvedDocument, VerifyOnOpen, cancellationSource.Token);
       var databaseEntry = new DocumentEntry(
         document.Version,
         resolvedDocument,
-        verifiedDocument,
+        verifiedDocuments.Select(Task.FromResult).DefaultIfEmpty(resolvedDocument).ToTask(cancellationSource.Token).Unwrap(),
```
cpitclaudel, line 76: When does the empty case kick in?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -180,24 +180,26 @@ IRelocator relocator
       }
 
       var cancellationSource = new CancellationTokenSource();
-      var verifiedDocumentTask = VerifyDocumentIfRequiredAsync(databaseEntry, cancellationSource.Token);
+      var verifiedDocuments = VerifyDocumentIfRequired(databaseEntry, cancellationSource.Token);
       var updatedEntry = new DocumentEntry(
         databaseEntry.Version,
         databaseEntry.ResolvedDocument,
-        verifiedDocumentTask,
+        verifiedDocuments.Select(Task.FromResult).DefaultIfEmpty(databaseEntry.ResolvedDocument).ToTask(cancellationSource.Token).Unwrap(),
```
cpitclaudel, line 187: Any chance we might share this code with the above?

cpitclaudel: Is ToTask from a third-party package / an extension?  I'm not familiar with this method.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -180,24 +180,26 @@ IRelocator relocator
...
     }
 
-    private async Task<DafnyDocument> VerifyDocumentIfRequiredAsync(IDocumentEntry databaseEntry, CancellationToken cancellationToken) {
-      var document = await databaseEntry.ResolvedDocument;
-      if (!RequiresOnSaveVerification(document)) {
-        throw new TaskCanceledException();
-      }
-      var verifiedDocumentTask = documentLoader.VerifyAsync(document, cancellationToken);
-      return await verifiedDocumentTask;
+    private IObservable<DafnyDocument> VerifyDocumentIfRequired(IDocumentEntry databaseEntry, CancellationToken cancellationToken) {
```
cpitclaudel, line 194: Why are we changing from a task to an observable here?  Are we returning multiple results?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -222,11 +224,14 @@ IRelocator relocator
...
     private record DocumentEntry(int? Version, Task<DafnyDocument> ResolvedDocument,
-      Task<DafnyDocument> VerifiedDocument,
+      Task<DafnyDocument> FullyVerifiedDocument,
       CancellationTokenSource CancellationSource) : IDocumentEntry {
       public void CancelPendingUpdates() {
         CancellationSource.Cancel();
       }
+
+      public Task<DafnyDocument> MostVerifiedDocument =>
+        FullyVerifiedDocument.IsCompletedSuccessfully ? FullyVerifiedDocument : ResolvedDocument;
```
cpitclaudel, line 234: I'm not sure about these names: resolution isn't verification.  Not sure I can think of something better.  Maybe `checked`?  We do resolution and typechecking, then verification; a verified document is "fully checked", whereas a typechecked one is only "partially checked".

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -26,6 +28,7 @@ public class TextDocumentLoader : ITextDocumentLoader {
     // 256MB
     private const int MaxStackSize = 0x10000000;
 
+    private DafnyOptions Options => DafnyOptions.O;
```
cpitclaudel, line 31: Is the next step to make this a field?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -26,6 +28,7 @@ public class TextDocumentLoader : ITextDocumentLoader {
     // 256MB
     private const int MaxStackSize = 0x10000000;
```
cpitclaudel, line 29: Why is this duplicated from above? (I know it's not from this patch.)

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -78,18 +81,17 @@ ILoggerFactory loggerFactory
...
+      return await await Task.Factory.StartNew(async () => LoadInternal(textDocument, cancellationToken), cancellationToken,
 #pragma warning restore CS1998
+        TaskCreationOptions.None, LargeStackScheduler);
     }
 
     private DafnyDocument LoadInternal(TextDocumentItem textDocument, CancellationToken cancellationToken) {
       var errorReporter = new DiagnosticErrorReporter(textDocument.Uri);
       var program = parser.Parse(textDocument, errorReporter, cancellationToken);
-      PublishDafnyLanguageServerLoadErrors(errorReporter, program);
+      IncludePluginLoadErrors(errorReporter, program);
```
cpitclaudel, line 94: Why "include" instead of "publish"?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -141,29 +148,54 @@ bool loadCanceled
...
+      var programErrorReporter = new DiagnosticErrorReporter(document.Uri);
+      document.Program.Reporter = programErrorReporter;
+      var implementationTasks = verifier.Verify(document.Program, progressReporter, cancellationToken);
+      foreach (var implementationTask in implementationTasks) {
+        implementationTask.Run();
+      }
 
-      return Task.Factory.StartNew(() => VerifyInternalAsync(document, cancellationToken), cancellationToken,
-        TaskCreationOptions.None, LargeStackScheduler).Unwrap();
+      var _ = NotifyStatusAsync(document.TextDocumentItem, implementationTasks);
```
cpitclaudel, line 161: No await here?

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -67,71 +69,35 @@ public class DafnyProgramVerifier : IProgramVerifier {
...
-      }
-    }
 
-    private async Task<bool> VerifyWithBoogieAsync(TextWriter output,
-      ExecutionEngine engine, Boogie.Program program,
-      CancellationToken cancellationToken, string programId) {
-      program.Resolve(engine.Options);
-      program.Typecheck(engine.Options);
-
-      engine.EliminateDeadVariables(program);
```
cpitclaudel, line 110: We have four instances of this call chain in Dafny; would it be worth adjusting the other callers as well?

cpitclaudel (approved):
This looks great; nice job!

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
```diff
@@ -603,10 +606,10 @@ modifies this
       await SetUp(new Dictionary<string, string>() {
         { $"{VerifierOptions.Section}:{nameof(VerifierOptions.VcsCores)}", "4" }
       });
-      for (int i = 0; i < 20; i++) {
+      for (int i = 0; i < 10; i++) {
```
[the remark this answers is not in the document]
keyboardDrummer: No, but otherwise it times out locally.

Source/DafnyLanguageServer.Test/Synchronization/SaveDocumentTest.cs
```diff
@@ -115,7 +116,6 @@ public class SaveDocumentTest : DafnyLanguageServerTestBase {
       await client.SaveDocumentAndWaitAsync(documentItem, CancellationToken);
       var savedDocument = await Documents.GetVerifiedDocumentAsync(documentItem.Uri);
       Assert.IsNotNull(savedDocument);
-      Assert.AreNotSame(openedDocument, savedDocument);
```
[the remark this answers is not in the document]
keyboardDrummer: This was a test on implementation details (checking for reference equality), and it failed with the new code but it's not important for users.

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
keyboardDrummer: Nice, I forgot they exist 😅

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: Seems like it, but I'd prefer doing that in a different PR

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: Specifically the Dafny resolver and translator need larger stacks, but this is caused by independent problems that could be resolved independently, so I've also kept the large stack workarounds independent. The code duplication is tiny.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: Publish is terminology used for sending diagnostics to the LSP clients (https://microsoft.github.io/language-server-protocol/specification#textDocument_publishDiagnostics), which is not yet being done here.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: Nope, these notifications are not waited for. They can run on their own time. The `var _ = ` is to explicitly not use the `Task`.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
keyboardDrummer: I think that requires more updates to the Boogie API.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
cpitclaudel: Good point, then maybe "collect"?  But this is minor.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
cpitclaudel: Ah, scary.  Thanks.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
cpitclaudel: I suspect the differences we have are mostly oversight rather than conscious decisions to run things slightly differently in each case, but in any case we can handle this later.

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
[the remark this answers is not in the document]
cpitclaudel: I wonder why this has gotten slower.

cpitclaudel (approved):

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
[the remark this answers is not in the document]
keyboardDrummer: I was never able to finish it before the time on my machine.

cpitclaudel (approved):

Issue #1322, opened 2021-07-23
Unintuitive overloading of the term "compile" in CLI flags


robin-aws:
This is something that's always irked me about the `dafny` CLI, but that I've just gotten used to. I assert that it's a common sharp edge for new Dafny users and worth fixing.

The current `/compile` flag has the following meaning (copied from `/help` output):

```
/compile:<n>  0 - do not compile Dafny program
    1 (default) - upon successful verification of the Dafny
        program, compile it to the designated target language
        (/noVerify automatically counts as failed verification)
    2 - always attempt to compile Dafny program to the target
        language, regardless of verification outcome
    3 - if there is a Main method and there are no verification
        errors and /noVerify is not used, compiles program in
        memory (i.e., does not write an output file) and runs it
    4 - like (3), but attempts to compile and run regardless of
        verification outcome
```

There are several problems with this:

1. The idea of "I'm going to compile SO HARD that I actually execute the program too" is not exactly intuitive. :)
2. `/compile:1` or `/compile:2` currently not only compiles the Dafny source to whatever the `/compileTarget` is, but also applies the target language compiler (for C# and Java), or at least runs the output through a target language tool (e.g. `node`) without executing any `Main` method, presumably to ensure the program is valid. This is quite subtle and surprising if the Dafny compiler doesn't actually produce a valid target language program. If you actually want to run JUST the Dafny compiler the correct set of flags is `/compile:0 /spillTargetCode:...` instead, which feels even weirder!
3. I generally dislike the style of numeric flags where different numbers have very different semantic meanings, and this is just one of the most egregious examples.

No alternative proposal just yet, and no plans to immediately fix this, just needed to get it written down after my latest adventure with this. :)

keyboardDrummer:
> if the Dafny compiler doesn't actually produce a valid target language program

Shouldn't a valid Dafny program always produce a valid target language program? Externs currently prevent this right? I think we can change how extern is compiled to prevent them form creating invalid target language programs.

> If you actually want to run JUST the Dafny compiler the correct set of flags is /compile:0 /spillTargetCode:... instead, which feels even weirder!

Why would you want to do that?

keyboardDrummer:
> The idea of "I'm going to compile SO HARD that I actually execute the program too" is not exactly intuitive. :)

I'd prefer separate top-level commands for building and running, similar to `dotnet build` and `dotnet run`.

parno:
> Why would you want to do that?

[...]

PR #2005, opened 2022-04-11
feat: /runAllTests option for executing all {:test} methods 

robin-aws:
Implemented by generating most of a `Main` method in the Dafny AST, with a small amount of target language code for recovering from halting (e.g. a failed `expect` statement). Recovering from halting is not part of the Dafny language, but already implemented internally for most target languages in order to support printing `[Program Halted] ...` at the top level.

Not yet implemented for C++ or Python as they don't yet support `expect` statements.


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -6927,6 +6927,39 @@ public Statement(IToken tok, IToken endTok)
...
+      Contract.Requires(type != null);
+      var variable = new LocalVariable(tok, tok, name, type, false);
+      variable.type = type;
+      return new VarDeclStmt(tok, tok, Util.Singleton(variable), null);
+    }
+
+    /// <summary>
+    /// Create a resolved statement for a local variable with an initial value.
+    /// </summary>
+    public static VarDeclStmt CrateLocalVariable(IToken tok, string name, Expression value) {
```
keyboardDrummer, line 6946: `Crate` => `Create` also on line 6934

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -5164,6 +5169,119 @@ private class GetterSetterLvalueImpl : ILvalue {
       }
     }
 
+    /// <summary>
+    /// Emits code to recover from halting in the code written to the returned syntax tree,
+    /// and assign the halting message to the given variable.
+    /// </summary>
+    protected abstract ConcreteSyntaxTree EmitHaltHandling(LocalVariable haltMessageVar, ConcreteSyntaxTree wr);
+
+    protected void EmitRunAllTestsMainMethod(Program program, ConcreteSyntaxTree w) {
```
keyboardDrummer, line 5178: For method in compilers it's often difficult to read them without knowing what the output should be. Could you add a C# example of the target output in a comment?

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2387,6 +2387,16 @@ protected class ClassWriter : IClassWriter {
       TrParenExpr("_dafny.SingleValue", e, wr, inLetExprBody, wStmts);
     }
 
+    protected override ConcreteSyntaxTree EmitHaltHandling(LocalVariable haltMessageVar, ConcreteSyntaxTree wr) {
+      var tryBlock = wr.NewBlock("try");
+      var catchBlockWr = wr.NewBlock("catch (e)");
```
keyboardDrummer, line 2392: Consider leaving out the `Wr` suffix.

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -563,6 +564,25 @@ private record TargetPaths(string Directory, string Filename) {
       var compiler = DafnyOptions.O.Compiler;
 
       var hasMain = Compilers.SinglePassCompiler.HasMain(dafnyProgram, out var mainMethod);
+      if (DafnyOptions.O.RunAllTests) {
```
keyboardDrummer, line 567: The PR currently seems to work both by generating Dafny code early on and by injecting extra output code in the compiler classes. Why do we need to do both?

keyboardDrummer (commented):
The `SinglePassCompiler` class is already very large (>5000 LOC) and this PR adds another feature to it. I think it would be amazing if we could implement this PR as a compiler pass.

The pass would detect the RunAllTests option and add a _Dafny_ main method that calls all the {:test} methods, all in Dafny.

I'm guessing we're generating some code that you can't generate with native Dafny, possibly related to the halt handling. Could we add a Dafny feature that's only available internally to enable generating the same code?

keyboardDrummer:
Could you update the Dafny reference manual to describe the new feature?

Source/DafnyDriver/DafnyDriver.cs
robin-aws: It has to create the `Method` AST node so that the special-cased logic dealing with the top-level entry point in `DafnyDriver` still works correctly. But I think I'll go ahead with your suggestion of an internal-only AST code for the reflective features so that everything can happen in the resolver instead.

robin-aws:
> The `SinglePassCompiler` class is already very large (>5000 LOC) and this PR adds another significant feature to it. I think it would be amazing if we could implement this PR as a compiler pass.
> 
> The pass would detect the RunAllTests option and add a _Dafny_ main method that calls all the {:test} methods, all in Dafny.
> 
> I'm guessing we're generating some code that you can't generate with native Dafny, possibly related to the halt handling. Could we add a Dafny feature that's only available internally to enable generating the same code?

I was strongly considering this earlier as well. Rather than a compilation pass, it can be handled within the resolver just as all the implicit `iterator` features currently are. I'm going to take a crack at that (and see if I can add the new code in a file other than the also-massive `Resolver.cs` :) )

Source/Dafny/DafnyOptions.cs
```diff
@@ -267,6 +268,15 @@ public enum IncludesModes {
             return true;
           }
 
+        case "runAllTests": {
+            int runAllTests = 0;
+            if (ps.GetIntArgument(ref runAllTests, 2)) {
+              RunAllTests = runAllTests != 0; // convert to boolean
```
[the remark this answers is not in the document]
robin-aws: Not meant as a TODO (and copied from other options like `/dafnyVerify`), more meant to be documentation of this being a conversion from a `0` or `1` argument to a boolean flag. Probably not worth a `GetBoolArgument` helper method?

robin-aws:
Okay that was definitely more work than I expected. :) Much cleaner though! And perhaps moving in the right direction towards being able to write a testing framework in pure Dafny in the future.

Source/Dafny/Verifier/Translator.TrStatement.cs
```diff
@@ -505,6 +505,12 @@ public partial class Translator {
         CheckWellformedWithResult(rhs, new WFOptions(null, false, false), rIe, pat.Expr.Type, locals, builder, etran);
         CheckCasePatternShape(pat, rIe, rhs.tok, pat.Expr.Type, builder);
         builder.Add(TrAssumeCmd(pat.tok, Bpl.Expr.Eq(etran.TrExpr(pat.Expr), rIe)));
+      } else if (stmt is HaltRecoveryStatement haltRecoveryStatement) {
```
keyboardDrummer: Would it be possible to add a noVerify annotation to the generated main method, and let this block `throw new NotSupportedException()` ?

Source/Dafny/Resolver/RunAllTestsMainMethod.cs
```diff
@@ -10,6 +10,10 @@ public class RunAllTestsMainMethod : IRewriter {
   public RunAllTestsMainMethod(ErrorReporter reporter) : base(reporter) {
   }
 
+  /// <summary>
+  /// Verifies that there isn't already a main method, and then creates
+  /// one with no body (to be filled in by PostResolve()).
```
keyboardDrummer, line 15: Just out of curiosity, why does this already need to happen in the PreResolve?

Source/Dafny/DafnyOptions.cs
```diff
@@ -1147,6 +1147,8 @@ verification outcome
     1 - Emits a main method in the target language that will execute every method
         in the program with the {{:test}} attribute.
         Cannot be used if the program already contains a main method.
+        Note that /compile:3 or 4 must be provided as well to actually execute
```
keyboardDrummer, line 1150: I'm looking forward to implementing this: https://github.com/dafny-lang/dafny/issues/1322#issuecomment-889023432, which could have a `test` command.

keyboardDrummer (commented):
I'm positively in love with the updated code! Not just the high level but also the details are a pleasure to read. Thanks!

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -3164,7 +3164,8 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
         } else if (DafnyOptions.O.ForbidNondeterminism) {
           Error(s.Tok, "modify statement without a body forbidden by /definiteAssignment:3 option", wr);
         }
-
+      } else if (stmt is HaltRecoveryStatement h) {
+        EmitHaltRecoveryStmt(h.Body, h.HaltMessageVar.CompileName, h.RecoveryBody, wr);
```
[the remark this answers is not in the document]
robin-aws: It IS a bit weird isn't it? All of the `Emit*` methods take individual pieces rather than whole AST nodes and I was just being consistent. I think it's so that multiple callers can emit low-level components with different configuration even if they don't have exactly the corresponding AST node in hand. I'm inclined to leave this as is for consistency unless you feel strongly about it.

Source/Dafny/Resolver/RunAllTestsMainMethod.cs
robin-aws: I initially started with needing PreResolve because I thought I'd be generating unresolved AST nodes, but then realized I needed at least the test return types to be resolved. It might be feasible to move creating the main method to PostResolve as well, but then I'd have to create a resolved method so it might be more significant refactoring (I tried just moving the existing code and hit errors). Given the `PreResolve(Program)` seems like a good hook to add for rewriters in general, I'd prefer to leave this as is now that it's working.

Source/Dafny/DafnyOptions.cs
robin-aws: I absolutely had that in mind when writing this. :) I've also been meaning to extract that idea into its own issue so it's not just buried in that discussion: https://github.com/dafny-lang/dafny/issues/2036

Source/Dafny/Verifier/Translator.TrStatement.cs
robin-aws: Good call, I like that better (and there's even precedent for doing that in a rewriter: https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Rewriter.cs#L1087)

keyboardDrummer (approved):

Source/Dafny/Compilers/SinglePassCompiler.cs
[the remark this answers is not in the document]
keyboardDrummer: Okay, I'd prefer the other way but this is fine too.

Source/Dafny/Resolver/RunAllTestsMainMethod.cs
```diff
@@ -0,0 +1,225 @@
...
+  ///
+  /// It might be possible to move this to PostResolve() if we created a resolved
+  /// Method instead, but for now this is a bit easier. It also allows us to produce
+  /// errors much earlier in the pipeline.
+  /// </summary>
+  internal override void PreResolve(Program program) {
+    Method mainMethod;
+    var hasMain = Compilers.SinglePassCompiler.HasMain(program, out mainMethod);
+    if (hasMain) {
+      Reporter.Error(MessageSource.Rewriter, mainMethod.tok, "Cannot use /runAllTests on a program with a main method");
```
cpitclaudel, line 25: Would it make sense to replace the existing main method instead?

robin-aws: I think it's generally better to catch cases of more than one main method by default, and I'd consider this an implicit case of that. I can imagine someone accidentally including source that includes an application main method, and this error would be an earlier signal their build is buggy. I've even accidentally included a dependency on a testing framework's CLI runner when I meant to depend on the package with the testing annotations, and the equivalent of this error saved me quite a bit of wasted time debugging. :)

cpitclaudel: Makes sense, but… how do I run all tests for an application that *has* a main method?

robin-aws: At least one approach is separating the source code into `src` and `test`, where the `test` code will reference most of the `src` content but not pull in the main method. The report generator does that for example: https://github.com/dafny-lang/dafny-reportgenerator. You could add `/runAllTests:1` to only the test project file here: https://github.com/dafny-lang/dafny-reportgenerator/blob/main/test/dafny-reportgenerator-test.csproj

PR #2006, opened 2022-04-11
Remove unused references and don't link DafnyRuntime into DafnyDriver

cpitclaudel:
This is in preparation for linking the output of the Dafny-in-Dafny compiler into Dafny (as a compiler plugin).  We can't do that today because Dafny.exe (compiled from DafnyDriver) already links in a copy of the runtime, even though it doesn't use it.

### DIFF
--- a/Source/IntegrationTests/IntegrationTests.csproj
+++ b/Source/IntegrationTests/IntegrationTests.csproj
@@ -22,7 +22,9 @@
 
     <ItemGroup>
         <ProjectReference Include="..\DafnyDriver\DafnyDriver.csproj" />
-        <ProjectReference Include="..\DafnyRuntime\DafnyRuntime.csproj" />
+        <ProjectReference Include="..\DafnyRuntime\DafnyRuntime.csproj">
+          <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+        </ProjectReference>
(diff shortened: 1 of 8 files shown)


Source/IntegrationTests/IntegrationTests.csproj
```diff
@@ -22,7 +22,9 @@
 
     <ItemGroup>
         <ProjectReference Include="..\DafnyDriver\DafnyDriver.csproj" />
-        <ProjectReference Include="..\DafnyRuntime\DafnyRuntime.csproj" />
+        <ProjectReference Include="..\DafnyRuntime\DafnyRuntime.csproj">
+          <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
```
MikaelMayer, line 26: I don't know the meaning of "ReferenceOutputAssembly" in these various projects. @robin-aws  can you please let me know if it's the right thing to do? You wanted to change the dependency to DafnyRuntime for a long time.

MikaelMayer (commented):
I'm asking an expert, otherwise, it looks good.

cpitclaudel:
Closing this PR for now, since @MikaelMayer pointed out to me that we can use the `Dafny` namespace for the runtime, which is different from the `Microsoft.Dafny` one used in the compiler.

PR #2018, opened 2022-04-13
Clean up lit.site.cfg and align it with xunit runner

cpitclaudel:
The set of environment variables exported by these two was not consistent.  It still isn't (`HOME` vs `DOTNET_CLI_HOME`), but this PR adds missing variables on the lit side.  Additionally, I've fixed the lit code to find Z3 (previously it found z3.exe but not z3.cmd)


keyboardDrummer:
Have do you feel about having a bash script `lit` that redirects to `dotnet --no-build --project Source/IntegrationTests ..`, and removing `lit.site.cfg` ?

keyboardDrummer (approved):

cpitclaudel:
> Have do you feel about having a bash script lit that redirects to dotnet --no-build --project Source/IntegrationTests .., and removing lit.site.cfg ?

Not sure: there are many things that lit does that xrunner doesn't do.  But maybe that's actually misleading to poor souls who get things working with lit only to realize that xrunner doesn't do the same thing ^^

keyboardDrummer:
> > Have do you feel about having a bash script lit that redirects to dotnet --no-build --project Source/IntegrationTests .., and removing lit.site.cfg ?
> 
> Not sure: there are many things that lit does that xrunner doesn't do. But maybe that's actually misleading to poor souls who get things working with lit only to realize that xrunner doesn't do the same thing ^^

I've noticed also the syntax we support is different between lit and XUnit, and some tests only run in XUnit but not in Lit, like CVSLogger.dfy

PR #2021, opened 2022-04-15
Feat: Code action support

MikaelMayer:
insertfailingassertdemo

This PR introduces the DafnyLanguageServer support for QuickFixes, which are a kind of code actions, restricted to same-file edits.
* The class `QuickFixer` in dafny plugins only requires a plugin instantiation to implement the method `public abstract QuickFix[] GetQuickFixes(IQuickFixInput input, IToken selection);` (the current selection is passed as a token).
* `QuickFix` have a title and a lazily computed list of text edits.
* `IQuickFixInput` notably contains the code and the last program.

I make use of this class in `DafnyLanguageServer` in the file `DafnyCodeActionHandler`, with two major functions called `Handle`, one is for returning a list of title + data to identity the command, the other is for resolving the code action to concrete edits.

I included 1 battery:
- A built-in starter plugin `VerificationQuickFixer` in its own file, which can currently inline a failing postcondition when there is no return statement, in method bodies only. It also work in nested blocks.

This PR includes tests with a unified representation of code, designated hover space with `>>>` and expected insertion with `[[message|insertedText]]` that should be pretty straightforward to understand and to maintain.

It also includes a test where a plugin provides quick fixes to ensure plugins can correctly provide this functionality.


keyboardDrummer:
I don't have enough Dafny verification debugging experience to comment on the use-case. Maybe @cpitclaudel or @RustanLeino can?

Source/DafnyLanguageServer/Plugins/QuickFixerHelpers.cs
```diff
@@ -0,0 +1,153 @@
...
+    var length = endTokenPos - startTokenPos;
+    if (startTokenPos < 0 || endTokenPos < startTokenPos || endTokenPos >= text.Length) {
+      return ""; // Safeguard
+    }
+
+    return text.Substring(startTokenPos, length);
+  }
+
+  /// <summary>
+  /// Given the position of a closing brace '}' and the position of the opening brace '{',
```
amaurremi, line 31: Since matching-parentheses functionality is provided in the method below, maybe one can provide only the opening parenthesis here?

amaurremi: Could you add a brief example showing what indentation this method would return for a simple code snippet?

Source/DafnyLanguageServer/Plugins/QuickFixerHelpers.cs
```diff
@@ -0,0 +1,153 @@
...
+  /// <summary>
+  /// Given an opening brace (typically where an error is detected),
+  /// finds and returns the closing brace token.
+  /// </summary>
+  /// <param name="program">The program</param>
+  /// <param name="documentUri">The document URI</param>
+  /// <param name="openingBrace">A token whose pos is the absolute position of the opening brace</param>
+  /// <returns>The token of a matching closing brace, typically the `ÈndTok` of a BlockStmt</returns>
+  public static IToken? GetMatchingEndToken(Dafny.Program program, string documentUri, IToken openingBrace) {
+    // Look in methods for BlockStmt with the IToken as opening brace
```
amaurremi, line 94: Could the brace not also be for a class or module body? If methods are the only use case, could this be documented?

Source/DafnyLanguageServer/Plugins/QuickFixerHelpers.cs
```diff
@@ -0,0 +1,153 @@
...
+
+    return null;
+  }
+
+  /// <summary>
+  /// Given an opening brace and a statement, if the statement's token is openingBrace
+  /// returns the closing brace token, else null.
+  /// Visit sub-statements recursively
+  /// </summary>
+  public static IToken? GetMatchingEndToken(IToken openingBrace, Statement stmt) {
```
amaurremi, line 122: Is it possible to implement this without the `stmt` parameter?

Source/DafnyLanguageServer.Test/_plugins/PluginsQuickFixTest.cs
```diff
@@ -0,0 +1,27 @@
+using Microsoft.Dafny;
+using Microsoft.Dafny.LanguageServer.Plugins;
+using System.Collections;
+using Microsoft.Boogie;
+using Microsoft.Dafny.LanguageServer.Language;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+namespace PluginsQuickFixTest {
+  /// <summary>
+  ///  Small plugin that detects all extern methods and verifies that there are test methods that actually invoke them.
```
amaurremi, line 10: Where is the code that detects extern methods?

amaurremi (commented):
Really great work, and I like how easy it appears to be now to add new code actions!

Source/DafnyLanguageServer/Workspace/IDocumentDatabase.cs
```diff
@@ -57,6 +57,13 @@ public interface IDocumentDatabase {
     /// <returns>An instance of the managed document, <c>null</c> if the specified document was not found.</returns>
     Task<DafnyDocument?> GetDocumentAsync(TextDocumentIdentifier documentId);
 
+    /// <summary>
+    /// Gets the last version of this document
+    /// </summary>
+    /// <param name="documentId">The ID of the document to resolve.</param>
+    /// <returns>An instance of the managed document, <c>null</c> if the specified document was not found.</returns>
+    Task<DafnyDocument?> GetLatestDocumentAsync(TextDocumentIdentifier documentId);
```
keyboardDrummer, line 65: I don't understand why you'd need this from the outside.

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+
+  /// <summary>
+  /// Returns the fixes along with a unique identifier
+  /// </summary>
+  public IEnumerable<QuickFixWithId> GetFixesWithIds(QuickFixer[] fixers, DafnyDocument document, CodeActionParams request) {
+    var ID = 0;
+    foreach (var fixer in fixers) {
+      // Maybe we could set the program only once, when resolved, instead of for every code action?
```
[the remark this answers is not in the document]
MikaelMayer: > do you mean moving the fixerInput variable out of the loop? Is that a problem to do it?

This comment is obsolete, let me remove it.

Source/DafnyLanguageServer/Plugins/QuickFixer.cs
```diff
@@ -0,0 +1,63 @@
...
+using Range = OmniSharp.Extensions.LanguageServer.Protocol.Models.Range;
+
+namespace Microsoft.Dafny.LanguageServer.Plugins;
+
+/// <summary>
+/// Plugins implement one or more QuickFixer to offer "quick code fixes",
+/// They should return very quickly, so most of the processing has to be done in the GetEdit()
+/// of every QuickFix.
+///
+/// To provide quick fixes for diagnostics related to the selection,
```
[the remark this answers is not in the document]
MikaelMayer: I just added a tutorial in DafnyLanguageServer/README.md

Source/DafnyLanguageServer/Plugins/QuickFixerHelpers.cs
MikaelMayer: I added a brief example.
I prefer to dissociate provide both, because in any case, the caller will need the end token to perform the insertion, so it has to compute it.

However, I created another helper that do both at the same time so that's it's easier to use.

Source/DafnyLanguageServer/Plugins/QuickFixerHelpers.cs
MikaelMayer: I don't think so. This method is recursive to traverse all statements. I could also use a visitor there, but I don't think it's necessary.

Source/DafnyLanguageServer.Test/_plugins/PluginsQuickFixTest.cs
MikaelMayer: > Where is the code that detects extern methods?

The comment was oudated. Updated.

Source/DafnyLanguageServer/Workspace/IDocumentDatabase.cs
MikaelMayer: > Why do you need this?

Because when I was developing this PR, there wasn't a `GetLastDocumentAsync` so I had to create one myself.
Now with the latest master there is the equivalent of the method I needed, so this conversation is obsolete.

MikaelMayer:
I'm converting this PR to a draft because the mechanism to extract a postcondition by looking at the range of the error is not reliable. For example, if the postcondition was:

```dafny
ensures P(x)
```
The inlined postcondition would be:

```dafny
assert P(x;
```

and the closing ")" would be missing. The only way to solve this consistently is to have a deeper access to failing postconditions and re-printing them, so I'll wait till https://github.com/dafny-lang/dafny/pull/2153/files is merged to re-enable this PR.

MikaelMayer:
With recent changes to master, the missing parenthesis problem is not a problem anymore, so I reopen this PR.

keyboardDrummer:
I don't think I understand the Dafny verification debugging process well enough to evaluate the UX of this feature well. How does inlining the failing ensures clause at the return location where it fails help you?

MikaelMayer:
> I don't think I understand the Dafny verification debugging process well enough to evaluate the UX of this feature well. How does inlining the failing ensures clause at the return location where it fails help you?

The Dafny verification debugging process consist first in figuring out the assert that need to be inserted so that everything else verifies. Once we only have failing explicit assertions, we can use [verification debugging techniques described here](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-verification-debugging) 
Assertions are buried everywhere, not all of them can be surfaced. I started an effort there to gather ways to convert verification errors into explicit assertions: https://github.com/dafny-lang/dafny/pull/2153

Postconditions are the simple case of such assertions. When we are on the return path, it's painful to have to navigate up to see which postcondition is failing. If we could show the failing postcondition, we might now what is wrong. Inlining the failing postcondition is one way to further split the assertion and figure out what is wrong.

Once we inline the failing postcondition inside an assert, and we reverify the method or function, it will also tell us if this postcondition did not hold only in this branch or if it was in all branches:
* If the postcondition now verifies, it means it's the last branch for which the postcondition did not hold, and we can simply apply [verification debugging techniques described here](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-verification-debugging) 
* If the postcondition still does not verify (and the assert usually don't verify either), we can similarly look for other places where to insert this failing postcondition, and continue from there.
* There is sometimes a third case like in https://github.com/dafny-lang/dafny/issues/2308 where it demonstrates that there is an extensionality axiom missing.

For example:

```dafny
method f(i: int) returns (j: int)
  decreases
  ensures i > j
{
  while * {
    if * {
      j := 2;
      return;
    }
  }
  return;
}
```
after two code actions, we will have inlined the failing asserts in two different places:
```dafny
method f(i: int) returns (j: int)
  decreases
  ensures i > j
{
  while * {
    if * {
      j := 2;
      assert i > j;
      return;
    }
  }
  assert i > j
  return;
}
```
and we will have proven the postcondition. Now, we can use weakest precondition calculus to figure out what is wrong.
Code action only prevents the need of copy-pasting for those who immediately have no clue.

robin-aws (commented):
Just started reading the feature at a high level and posting initial comments, will review the rest separately.

Source/DafnyLanguageServer/README.md
```diff
@@ -121,4 +130,122 @@ Plugins are typically used to report additional diagnostics such as unsupported
...
+  }
+}
+```
+Note that you could also override the methods `GetRewriters()` and `GetCompilers()` for other purposes, but this is out of scope for this tutorial.
+
+Then, we need to create the quickFixer `AddCommentQuickFixer` itself:
+
+```
+public class AddCommentQuickFixer : QuickFixer {
+  public override QuickFix[] GetQuickFixes(IQuickFixInput input, Range selection) {
```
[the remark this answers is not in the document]
MikaelMayer: I hear you, but that means it requires clients to include Ling in their plugin, which they don't have. Clients should be able to write plugins using languages that's don't have IEnumerable (i.e. Dafny ...)

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -0,0 +1,156 @@
+﻿using System;
```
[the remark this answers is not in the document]
MikaelMayer: Cleaned up !

Source/DafnyLanguageServer/README.md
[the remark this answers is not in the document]
keyboardDrummer: This is a C# API, so it should be idiomatic to C#.

If you want to have a Dafny API for a piece of the Dafny compiler written in C#, then you should add a Dafny layer on top of the C# API.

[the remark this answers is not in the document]
MikaelMayer: See this discussion here we had previously in the context of plugins for the first reason why it's not a good idea to use C# idioms.
https://github.com/dafny-lang/dafny/pull/1739#discussion_r802868967
In short, it is not meant to be a C# API. It's a DotNet API. And it's not only Dafny that will use it, but also F#.

The second reason I'll make is that we need all the summary of these quick fixes immediately. An `IEnumerable` would make sense if we were streaming them or if it was expensive to compute all of them. But since they are going to appear in the UI of the user, there won't be more than 10 of these items usually.
You can always use an array as an `IEnumerable` by including Linq.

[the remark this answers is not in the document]
keyboardDrummer: > In short, it is not meant to be a C# API. It's a DotNet API. And it's not only Dafny that will use it, but also F#.

Does C# vs F# change anything about IEnumerable being idiomatic?

> The second reason I'll make is that we need all the summary of these quick fixes immediately

You could argue for returning an IReadOnlyList, but I think returning an abstract type like `IEnumerable` or `IReadOnlyList` instead of a specific type like an array is still idiomatic C#/.NET here.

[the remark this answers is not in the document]
MikaelMayer: I forgot to reply, thanks for your patience and review @robin-aws  and @keyboardDrummer 

I disagree that IEnumerable (🥈) is "best" for an API than a regular array (🥇) , because it requires to import `use System.Collection.Generic`, enumerables take more space on the disk, and in other parts of the API (`GetRewriters`, `GetCompilers`, `GetQuickFixers`), where we don't need anything enumerable, we are using arrays.

However, given that you were 2️⃣  to ask for it, and it does not make that much difference at the end, and it's not at the same level as the top-level `GetRewriters` ..., I did implement the change to use `IEnumerable<>` instead of an array 🍨 . Let me know if you like it.

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+    };
+  }
+
+
+  /// <summary>
+  /// Returns the fixes along with a unique identifier
+  /// </summary>
+  public IEnumerable<QuickFixWithId> GetFixesWithIds(QuickFixer[] fixers, DafnyDocument document, CodeActionParams request) {
+    var id = 0;
+    foreach (var fixer in fixers) {
```
keyboardDrummer, line 46: Have you considered using `return fixers.SelectMany` instead of the double foreach and yield return?

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+      ),
+      WorkDoneProgress = false
+    };
+  }
+
+
+  /// <summary>
+  /// Returns the fixes along with a unique identifier
+  /// </summary>
+  public IEnumerable<QuickFixWithId> GetFixesWithIds(QuickFixer[] fixers, DafnyDocument document, CodeActionParams request) {
```
keyboardDrummer, line 44: I think the input argument `fixers` here should be of type `IEnumerable<QuickFixer>` to be more admissible.

Source/Dafny/Plugin.cs
```diff
@@ -13,18 +13,18 @@ public interface Plugin {
 }
 
 public class ConfiguredPlugin : Plugin {
-  private readonly PluginConfiguration configuration;
+  public readonly PluginConfiguration Configuration;
```
keyboardDrummer, line 16: I believe public fields are avoided in .NET, but I don't know why so I'm fine with this. The only downside I see is that it might surprise .NET developers coming into this codebase.

Source/DafnyLanguageServer/Language/VerificationQuickFixer.cs
```diff
@@ -0,0 +1,50 @@
...
+    var expression = QuickFixerHelpers.Extract(relatedInformation.Location.Range, input.Code);
+    var (beforeEndBrace, indentationExtra, indentationUntilBrace) =
+      QuickFixerHelpers.GetInformationToInsertAtEndOfBlock(input, diagnostic.Range.Start);
+    if (beforeEndBrace == null) {
+      return null;
+    }
+
+    return new QuickFix[] {
+      new InstantQuickFix(
+        "Make the failing assertion explicit",
```
keyboardDrummer, line 41: Does this insert the condition of a postcondition at a return site? Have you considered a text like `assert postcondition at return location where it fails`

Source/DafnyLanguageServer/Language/BoogieExtensions.cs
```diff
@@ -49,5 +52,27 @@ public static class BoogieExtensions {
...
       return new Position(boogieLine + LineOffset, boogieColumn + ColumnOffset);
     }
+
+    public static IToken ToToken(this Position position, string document) {
+      try {
+        return new Token() {
+          line = position.Line - LineOffset,
+          col = position.Character - ColumnOffset,
+          val = "",
+          pos = position.ToAbsolutePosition(document)
```
keyboardDrummer, line 62: Calling `ToAbsolutePosition` seems like a performance problem. Each time the user changes their cursor to a new line, `textDocument/codeAction` is called and this triggers an iteration over the entire document on the server. I think we should avoid doing any document sized operations. Wouldn't it be better not to have `PositionExtensions.ToAbsolutePosition` at all and force clients that want to access text based on ranges to work through `DocumentTextBuffer` ?

Source/DafnyLanguageServer/Plugins/PluginConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+/// If the plugin defines no PluginConfiguration, then Dafny will instantiate every sub-class
+/// of Rewriter and Compiler from the plugin (providing rewriters with an ErrorReporter in the constructor
+/// as the first and only argument, and compilers with no arguments).
+/// </summary>
+public abstract class PluginConfiguration : Dafny.Plugins.PluginConfiguration {
+  /// <summary>
+  /// Override this method to provide quick fixers
+  /// </summary>
+  /// <returns>An array of quick fixers implemented by this plugin</returns>
+  public virtual QuickFixer[] GetQuickFixers() {
```
keyboardDrummer, line 22: Minor: I would prefer the name `QuickFixProvider` because I find it less ambiguous

Source/DafnyLanguageServer/Plugins/QuickFix.cs
```diff
@@ -0,0 +1,41 @@
...
+    return edits;
+  }
+}
+
+/// <summary>
+/// A quick fix replaces a range with the replacing text.
+/// </summary>
+/// <param name="Range">The range to replace. The start is given by the token's start, and the length is given by the val's length.</param>
+/// <param name="Replacement"></param>
+public record QuickFixEdit(Range Range, string Replacement = "");
```
keyboardDrummer, line 41: Given that this API only supports actions whose edits span a single file, is the idea to start with this API and evolve it over time?

Source/DafnyLanguageServer/Plugins/QuickFix.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.Dafny.LanguageServer.Plugins;
+
+/// <summary>
+/// Describe a single suggestion made to the user, where edits can be
+/// computed lazily.
+/// If the edits can be provided instantly, you can use the derived class
+/// `new InstantQuickFix(title, edits)`
+/// </summary>
+public abstract class QuickFix {
```
keyboardDrummer, line 11: Could this name be more connected to the name `CodeAction`, to stay in line with LSP's terminology? I believe LSP uses the term `quickfix` to indicate a code action that resolves a diagnostic, but that's not necessarily the case here.

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+    var quickFixers = GetQuickFixers();
+    var fixesWithId = GetFixesWithIds(quickFixers, document, request).ToArray();
+
+    var documentUri = document.Uri.ToString();
+    documentUriToQuickFixes.AddOrUpdate(documentUri,
+      _ => fixesWithId, (_, _) => fixesWithId);
+    var codeActions = fixesWithId.Select(fixWithId => {
+      CommandOrCodeAction t = new CodeAction {
+        Title = fixWithId.QuickFix.Title,
+        Data = new JArray(documentUri, fixWithId.Id)
```
keyboardDrummer, line 74: One of the fields of a CodeAction is `diagnostics`, and filling this field causes the code action to show up grouped under the diagnostic in the Problems view. The code action provided by `VerificationQuickFixer` relates to a diagnostic, so shouldn't we connect those two using the API ?

Source/DafnyLanguageServer/Plugins/IQuickFixInput.cs
```diff
@@ -0,0 +1,19 @@
+﻿using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+namespace Microsoft.Dafny.LanguageServer.Plugins;
+
+public interface IQuickFixInput {
```
keyboardDrummer, line 5: Why not use `DafnyDocument` instead of this?

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+      return new CommandOrCodeActionContainer();
+    }
+    var quickFixers = GetQuickFixers();
+    var fixesWithId = GetFixesWithIds(quickFixers, document, request).ToArray();
+
+    var documentUri = document.Uri.ToString();
+    documentUriToQuickFixes.AddOrUpdate(documentUri,
+      _ => fixesWithId, (_, _) => fixesWithId);
+    var codeActions = fixesWithId.Select(fixWithId => {
+      CommandOrCodeAction t = new CodeAction {
```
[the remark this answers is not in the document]
MikaelMayer: I'm not sure about what you mean.
The code that handles the code action is this one:

`public override Task<CodeAction> Handle(CodeAction request...`, which explicitly requires an unresolved code action. So I don't understand why I should send a command?

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
MikaelMayer: > Have you considered using return fixers.SelectMany instead of the double foreach and yield return?

Good point.

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
MikaelMayer: > I think the input argument fixers here should be of type IEnumerable<QuickFixer> to be more admissible.

I also made the method private because it does not need to be public.

Source/Dafny/Plugin.cs
MikaelMayer: > I believe public fields are avoided in .NET, but I don't know why so I'm fine with this. The only downside I see is that it might surprise .NET developers coming into this codebase.

I made it into a get property

Source/DafnyLanguageServer/Plugins/QuickFix.cs
MikaelMayer: Yes, we could add a field Uri to this down the road, so now we can start with this simpler API.

Source/DafnyLanguageServer/Plugins/QuickFix.cs
MikaelMayer: > Could this name be more connected to the name CodeAction,

I renamed it to DafnyCodeAction / DafnyCodeActionProvider

Source/DafnyLanguageServer/Plugins/IQuickFixInput.cs
MikaelMayer: I prefer to have an interface for the API rather than a concrete implementation that might be evolving.

Also, for a suitable API, we don't want to expose plugin developers with more information than necessary, do we?

Source/DafnyLanguageServer/Language/BoogieExtensions.cs
MikaelMayer: This call will only happen when the user places their caret on a line with a verification diagnostic.
However, I refactored the code now so that it will call `ToAbsolutePösition` only when accepting a Quick fix

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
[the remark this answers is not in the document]
keyboardDrummer: I'm saying you could return a CodeAction that includes a Command, since this seems how the API is intended to be used. From the documentation: 

> A CodeAction must set either `edit` and/or a `command`. If both are supplied, the `edit` is applied first, then the `command` is executed.

Source/DafnyLanguageServer/Plugins/DafnyCodeActionHelpers.cs
```diff
@@ -0,0 +1,193 @@
...
+/// Helpers for plugins defining a DafnyCodeActionProvider
+/// </summary>
+public static class DafnyCodeActionHelpers {
+  /// <summary>
+  /// Given an LSP range and some text, extract the corresponding substring
+  /// </summary>
+  /// <param name="range">The range to extract</param>
+  /// <param name="text">The document</param>
+  /// <returns>The substring of the document in the range</returns>
+  public static string Extract(Range range, string text) {
```
keyboardDrummer, line 20: I don't think this linear time complexity method should exist. We can add an `Extract(Range range)` to `DocumentTextBuffer` though.

Source/DafnyLanguageServer/Language/VerificationDafnyCodeActionProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+      return null;
+    }
+
+    if (relatedInformation.Location.Uri != uri) {
+      return null;
+    }
+
+    var expression = DafnyCodeActionHelpers.Extract(relatedInformation.Location.Range, input.Code);
+    var (beforeEndBrace, indentationExtra, indentationUntilBrace) =
+      DafnyCodeActionHelpers.GetInformationToInsertAtEndOfBlock(input, diagnostic.Range.Start);
```
keyboardDrummer, line 34: I find the API of `DafnyCodeActionHelpers.GetInformationToInsertAtEndOfBlock` hard to understand. How about something like

```c#
static class RefactorTools {

  public DafnyCodeActionEdit? InsertBlockAtReturnLocation(
    DocumentTextBuffer text, 
    Range returnLocationRange,
    IEnumerable<string> statementsToInsert)

}
``` 
?

Source/DafnyLanguageServer/Plugins/DafnyCodeActionHelpers.cs
```diff
@@ -0,0 +1,193 @@
...
+  ///       Some code here:
+  ///       assert X;
+  ///     }
+  /// ```
+  /// 
+  /// </summary>
+  /// <param name="endToken">The position of the closing brace</param>
+  /// <param name="text">The document text</param>
+  /// <returns>(extra indentation for a statement, current indentation)</returns>
+  public static (string, string) GetIndentationBefore(IToken endToken, int startLine, int startCol, string text) {
```
keyboardDrummer, line 56: In the long run, I imagine we might want to provide a different API for creating refactorings in Dafny, one that doesn't require thinking about formatting.

Here's a bad example:

```c#
Edit AddStatementToEnd(method: DafnyMethod) {
  var original = method.Snapshot();
  method.Body = method.Body.Concat(new SomeStatement(..));
  Formatter.Format(method);
  return DiffEngine.ComputeEdit(original, method);
}
```

I'm sure Roslyn has some interesting APIs for this very same purpose. I'd be great if we could get familiar with them, although I imagine it's quite a learning curve.

Here's an example: https://github.com/dotnet/roslyn/blob/main/src/Analyzers/CSharp/CodeFixes/MisplacedUsingDirectives/MisplacedUsingDirectivesCodeFixProvider.cs

With test: https://github.com/dotnet/roslyn/blob/main/src/Analyzers/CSharp/Tests/MisplacedUsingDirectives/MisplacedUsingDirectivesCodeFixProviderTests.cs

Source/DafnyLanguageServer/Plugins/DafnyCodeActionHelpers.cs
```diff
@@ -0,0 +1,193 @@
...
+  /// <summary>
+  /// Given an opening brace of a Block (typically where an error is detected),
+  /// finds and returns the closing brace token.
+  /// </summary>
+  /// <param name="program">The program</param>
+  /// <param name="documentUri">The document URI</param>
+  /// <param name="line">The line of the opening brace</param>
+  /// <param name="col">The column of the opening brace</param>
+  /// <returns>The token of a matching closing brace, typically the `ÈndTok` of a BlockStmt</returns>
+  private static IToken? GetMatchingEndToken(Dafny.Program program, string documentUri, int line, int col) {
```
keyboardDrummer, line 133: Why not pass a `Position openingBrace` instead of a line and col ?

Source/DafnyLanguageServer/Plugins/IDafnyCodeActionInput.cs
```diff
@@ -0,0 +1,19 @@
+﻿using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+namespace Microsoft.Dafny.LanguageServer.Plugins;
+
+public interface IDafnyCodeActionInput {
```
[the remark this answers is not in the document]
MikaelMayer: Interesting point of view!
I think the convention in C# is to usually name interfaces in what purpose they serve, not about how they are implemented. A document compilation (report?) would seem like an implementation detail.
So let's keep this naming for now.

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -0,0 +1,161 @@
...
+      edits.Add(new TextEdit() {
+        NewText = toReplace,
+        Range = range
+      });
+    }
+    return edits.ToArray();
+  }
+}
+
+public class VerificationDafnyCodeActionProviderInput : IDafnyCodeActionInput {
```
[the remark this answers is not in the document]
MikaelMayer: Verification is obsolete. It dates from the time when I had the quick fixer be defined in Dafny core instead of DafnyLanguageServer
I renamed it `DafnyCodeActionInput`. And what do you mean by your question? This class uses `DafnyDocument` to implement `IDafnyCodeActionInput` (which is the interface plugins implementers are going to work with)

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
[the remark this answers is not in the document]
MikaelMayer: Looks like the documentation of the API is incomplete. Although what you say works well theoretically, it has the side-effect that it creates an error in VSCode.
image
And that's what I thought: Code actions can either give edits on the text or more general actions.
However, the API documentation you describe don't talk about the construction of code actions which is two phases, which I've seen while browing the code online
1) Creation of the code action with possible data attached
2) Resolution of the code action when it's being clicked.

In 1) there is no need to create an edit or a command if all the edits and commands are going to be computed on 2.

As a conclusion, adding a "command" would require us to change the VSCode extension to support this command, which would have to manually invoke the code action resolver... which would be useless and hard to maintain
So I'm going to revert my changes there and just use the regular data field which is there for this purpose - the data between code action creation and code action resolution.

Source/DafnyLanguageServer/Language/VerificationDafnyCodeActionProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    if (diagnostic.RelatedInformation?.FirstOrDefault() is not { } relatedInformation) {
+      return null;
+    }
+
+    if (relatedInformation.Location.Uri != uri) {
+      return null;
+    }
+
+    var expression = DafnyCodeActionHelpers.Extract(relatedInformation.Location.Range, input.Code);
```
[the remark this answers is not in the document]
MikaelMayer: > How come? Code actions are available on lines without diagnostics, are they not?

Not every code action requires extracting some substring of the code all the time. If no code action is returned, or a code action is returned that does not call this particular method, then this line is not called. I won't go into more details as I removed this call.

> This call still seems like an O(N) operation. Why don't you get a DocumentTextBuffer from IDafnyCodeActionInput and let DocumentTextBuffer store a mapping from line to index that can be used to extract text using a Range in constant time?

Well, it's a O(N*C) where C is the number of code actions and N is the size of the file, and C is expected to be very small, so I wouldn't have worried about it.
However, I implemented such a conversion in the implementation  DafnyCodeActionInput, so that the interface can expose the method Extract, and I removed the three ToToken methods that were converting relative positions to absolute positions.

Source/DafnyLanguageServer/Plugins/DafnyCodeAction.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Microsoft.Dafny.LanguageServer.Plugins;
+
+/// <summary>
+/// Describe a single suggestion made to the user, where edits can be
+/// computed lazily.
+/// If the edits can be provided instantly, you can use the derived class
+/// `new InstantDafnyCodeAction(title, edits)`
+/// </summary>
+public abstract class DafnyCodeAction {
```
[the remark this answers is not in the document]
MikaelMayer: I would like to avoid this, since I know some implementers of code actions in F# and since it's an abstract class, they would need to implement boilerplate methods such as clone, copy, etc.
But even if you make it concrete, I'm not sure how the Func<>  interfaces which are C# specific would translate to F#. So, let's avoid the complexity and keep the API as simple as possible.

Source/DafnyLanguageServer/Plugins/DafnyCodeAction.cs
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  protected DafnyCodeAction(string title, Container<Diagnostic> diagnostics) {
+    Title = title;
+    Diagnostics = diagnostics;
+  }
+
+  // Edits are all performed at the same time
+  // They are lazily invoked, so that they can be as complex as needed.
+  public abstract DafnyCodeActionEdit[] GetEdits();
```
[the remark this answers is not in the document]
MikaelMayer: > without downsides.

It forces API users to link to the library System.Collection.Generic (I have to add one more import to the file as well). I can't call this "without downside".
Nevertheless, I did as you suggested.

Source/DafnyLanguageServer/Language/VerificationDafnyCodeActionProvider.cs
MikaelMayer: Thanks for this refactoring suggestion, I applied it.
Note that I only need one statement to insert. For multiple statement, it's sufficient to call this function repeatedly and store the edits in the same order, because this is how they are going to be applied.

Source/DafnyLanguageServer/Plugins/DafnyCodeActionHelpers.cs
MikaelMayer: Yes yes ! That's also part of what I envision we will be able to do with [the Dafny formater](https://github.com/dafny-lang/dafny/pull/2399).
In any case, in the short-term after this PR, we will need to handle expressions, substitutions, etc. So formatting them might be a good idea.

Source/DafnyLanguageServer/Plugins/DafnyCodeActionHelpers.cs
MikaelMayer: Because we need to compare the line and the column to a lot of tokens, and Position is offset by tokens by 1, so we would be computing unnecessary affine transformations all the time.
This is a private method anyway, we can always upgrade this later.

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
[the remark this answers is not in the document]
keyboardDrummer: To me a name ending in `Input` seems wrong.

When I read a function signature `SomeFunction(SomeFunctionInput x)` then the name `SomeFunctionInput` doesn't give me more information than what the rest of the function signature already did. What I'd like to know is what information can I get from `x`, and the name of the type of `x` should help with that.

Instead of the name `IDafnyCodeActionInput`, I think something like `IResolvedCompilation` or `IResolvedDocument` would tell me better what information I can get from it. If it's resolved then I expect that I can get the document text, the document AST, and the resolved types.

Source/DafnyLanguageServer/Plugins/DafnyCodeAction.cs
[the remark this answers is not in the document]
keyboardDrummer: > It forces API users to link to the library System.Collection.Generic

But that's part of the .NET framework is it not? Any C# application will use that namespace. Does this relate to C# vs F#?

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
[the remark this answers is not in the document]
keyboardDrummer: > As a conclusion, adding a "command" would require us to change the VSCode extension to support this command, which would have to manually invoke the code action resolver... which would be useless and hard to maintain

That indeed doesn't seem what we want.

Source/DafnyLanguageServer/Plugins/DafnyCodeAction.cs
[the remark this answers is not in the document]
keyboardDrummer: I think it might be good to take some time to agree on what language we're writing the API for. You seem to have settled on a particular style of API that sits between what is idiomatic for C# and F#, while I'm still trying to optimise for C#.

In this particular case, wouldn't `record DafnyCodeAction(string Title, Container<Diagnostic> Diagnostics, Func<DafnyCodeActionEdit[]> GetEdits)` work well for F# as well, since there is no abstract class, and F# can use a lambda expression for the `GetEdits` argument?

Source/DafnyLanguageServer/Plugins/DafnyCodeActionHelpers.cs
keyboardDrummer: Ok,

Just want us and our users to be aware that the API might still change (a lot)

Source/DafnyLanguageServer/Plugins/DafnyCodeActionHelpers.cs
keyboardDrummer: This seems like a slow approach to find the end token. I think something like the IIntervalTree used [here](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Language/Symbols/SymbolTable.cs#L33), or a position directed search like [here](https://github.com/keyboardDrummer/Miksilo/blob/f04b13466426d5cbdd7766f4c8c6239f7241a675/languageServer/shared/src/main/scala/miksilo/languageServer/server/SourcePath.scala#L19) would be better since both are constant instead of linear complexity.

MikaelMayer: I agree we should do a better job at indexing the program to access various definitions, as we get more and more features about navigation.
For this PR, implementing a tree system is out of scope, because the verification completion plugin is a toy example and is meant to evolve; to avoid traversing the body of 10k methods, I added a check about line numbers, so that it will be very fast. I think that's the best reasonable we should aim at for now if we want this PR to be merged.

Source/DafnyLanguageServer/Plugins/PluginConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+/// If the plugin defines no PluginConfiguration, then Dafny will instantiate every sub-class
+/// of Rewriter and Compiler from the plugin (providing rewriters with an ErrorReporter in the constructor
+/// as the first and only argument, and compilers with no arguments).
+/// </summary>
+public abstract class PluginConfiguration : Dafny.Plugins.PluginConfiguration {
+  /// <summary>
+  /// Override this method to provide quick fixers
+  /// </summary>
+  /// <returns>An array of quick fixers implemented by this plugin</returns>
+  public virtual DafnyCodeActionProvider[] GetDafnyCodeActionProviders() {
```
atomb, line 22: Given that you changed `DafnyCodeActionProvider` to return an `IEnumerable`, would it make sense to do so here, too? If you don't do that, I think you could instead remove `using System.Collections.Generic` above.

Source/DafnyLanguageServer/README.md
```diff
@@ -121,4 +130,122 @@ Plugins are typically used to report additional diagnostics such as unsupported
 
 Note that all plugin errors should use the original program's expressions' token and NOT `Token.NoToken`, else no error will be displayed in the IDE.
 
-Morover, plugins should not write anything to `stdout` as it interferes with the communication protocol with the IDE.
```
atomb, line 124: Is this no longer true?

Source/DafnyLanguageServer/Language/VerificationDafnyCodeActionProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Dafny.LanguageServer.Plugins;
+using Microsoft.Dafny.Plugins;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using Range = OmniSharp.Extensions.LanguageServer.Protocol.Models.Range;
+
+namespace Microsoft.Dafny.LanguageServer.Language;
+
+/// <summary>
+/// A verification quick fixers provides quick "fixes" for verification errors.
+/// For now, it offers to inline a failing postcondition if there is no "return" keyword.
```
atomb, line 15: I don't see logic in here related to whether there's a `return` keyword or not. I'm not sure it's necessary to have that, but I'd update the comment to match the code.

Source/DafnyLanguageServer/Plugins/DafnyCodeActionProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+          result.AddRange(moreDafnyCodeActiones);
+        }
+      }
+    }
+
+    return result;
+  }
+
+  /// <summary>
+  /// You can safely assume that input.Program is not null
```
atomb, line 56: Could you update this to start with a summary of what the method does?

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -0,0 +1,213 @@
...
+  public string Code => Document.TextDocumentItem.Text;
+  public Dafny.Program Program => Document.Program;
+  public DafnyDocument Document { get; }
+
+  public Diagnostic[] Diagnostics {
+    get {
+      var result = Document.Diagnostics.ToArray();
+      if (result.Length == 0) {
+        // For anonymous documents opened in VSCode
+        result = Document.Diagnostics.ToArray();
```
atomb, line 153: I'm not sure I understand what's happening here. Does this do anything different from line 150?

Source/DafnyLanguageServer/Plugins/DafnyCodeActionHelpers.cs
atomb: Yeah, I hope we can move toward AST-based rewrites relatively soon. All this string processing makes me nervous.

atomb (changes requested):
This looks like it's in pretty good shape. Thanks! I requested a few changes, but they should be pretty quick.

Source/DafnyLanguageServer/Plugins/PluginConfiguration.cs
MikaelMayer: Good suggestion. For consistency with the original `Dafny.Plugins.PluginConfiguration` that returns `Compiler[]` and `Rewriter[]`, I prefer to keep returning an array here. So I removed `using System.Collections.Generic`

Source/DafnyLanguageServer/README.md
MikaelMayer: Yes, this is no longer true. Plugins can write to stdout now, and this information is recorded in a log file.

https://github.com/dafny-lang/dafny/blob/1472e40e4007588071b2bced71f753fb7b678f9b/Source/DafnyLanguageServer/Program.cs#L30-L32

Source/DafnyLanguageServer/Language/VerificationDafnyCodeActionProvider.cs
MikaelMayer: I rephrased it to
```
/// For now, it offers to inline a failing postcondition if its failure is
/// indicated on the '{' -- meaning there is no explicit return.
```

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
MikaelMayer: Ahah interesting ! the code there used to be

```cs
      var result = Document.Errors.GetDiagnostics(Document.Uri).ToArray();
      if (result.Length == 0) {
        // For anonymous documents opened in VSCode
        result = Document.Errors.GetDiagnostics(DocumentUri).ToArray();
      }
```
so that it would handle separatedly the document URI for anonymous documents. However, since the recent refactorings, and the fact that URI are universally used, the case split is no longer user. Thanks for pointing out.

MikaelMayer:
> Yeah, I hope we can move toward AST-based rewrites relatively soon. All this string processing makes me nervous.

Me too. It's just that, with the guarantee that we don't need to replace variables for inlining postconditions, we should be fine. But with the PR of the formatter, we'll have access to tokens, so this will be even easier to print things (and indent them correctly!)

atomb (dismissed):
Looks good!

atomb (approved):

PR #2023, opened 2022-04-15
fix: Resolution of static functions-by-method

### LINKED ISSUE #2016: static function-by-method crashes Dafny
```
module M {
  datatype D = D() {
    static function f(): (res: int) {
      5
    } by method {
      return 5;
    }
  }
}
```

crashes Dafny.

```
$ dafny /compile:0 ~/tmp/bug.dfy
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
...
```

fabiomadge:
Fixes #2016.

### DIFF
--- a/Source/Dafny/Resolver.cs
+++ b/Source/Dafny/Resolver.cs
@@ -2307,7 +2307,7 @@ namespace Microsoft.Dafny {
             extraMember.InheritVisibility(m, false);
             members.Add(extraName, extraMember);
           } else if (m is Function f && f.ByMethodBody != null) {
-            RegisterByMethod(f);
+            RegisterByMethod(f, cl);
           }
         } else if (m is Constructor && !((Constructor)m).HasName) {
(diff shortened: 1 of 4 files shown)


Source/Dafny/Resolver.cs
```diff
@@ -2317,10 +2317,9 @@ public class ModuleBindings {
       }
     }
 
-    void RegisterByMethod(Function f) {
+    void RegisterByMethod(Function f, TopLevelDeclWithMembers cl) {
       Contract.Requires(f != null && f.ByMethodBody != null);
 
-      var cl = (TopLevelDeclWithMembers)f.EnclosingClass;
```
cpitclaudel, line 2323: Is it a bug that the `enclosingclass` of `f` doesn't match cl?

fabiomadge: I don't think so, as I believe that it gets filled in later. It's null at this point.

cpitclaudel (approved):

PR #2024, opened 2022-04-15
Merge mocking changes

cassidywaldrip:
Fixes #

This branch does two things:

1. Add support for mocking and the creation of fresh objects to the Java compiler (functionality that already exists for the C# compiler).
2. Add parameters to the {:synthesize} attribute in order to differentiate between the creation of a fresh object and the creation of a mocked object for both the Java and C#.

Mocking in Java is similar to mocking in C#. The only significant difference between the two at this point is that the Mockito library for Java does not support mocking fields (C#'s Moq does). So, the grammar for mocking in Java is:

`S            = FORALL 
                 | EQUALS 
                 | S && S
EQUALS  = ID.ID (ARGLIST) == EXP // stubs a method call
                 | EQUALS && EQUALS
FORALL  = forall SKOLEMS :: EXP ==> EQUALS
ARGLIST = ID // this can be one of the skolem variables
                 | EXP // this exp may not reference any of the skolems 
                 | ARGLIST, ARGLIST
SKOLEMS = ID : ID 
                 | SKOLEMS, SKOLEMS
`

Here is an example of using the `{:synthesize}` attribute for creating a fresh method and for creating a mock.

`
class Even {

    var value: int;

    function method IsValid() : bool reads this
    {
        this.value % 2 == 0
    }

    constructor(value: int) 
        requires value % 2 == 0
        ensures this.IsValid()
    {
        this.value := value;
    }

    function method Sum(a: int, b: int) : int 
    {
        a + b
    }
}

static method {:synthesize "fresh"} freshEven() returns (e:Even) ensures fresh(e)

static method {:synthesize "mock"} MockSum() returns (e:Even) 
    ensures fresh(e) 
    ensures e.Sum(2, 2) == 3
`

The methods annotated with `{:synthesize}`  are automatically translated to the following Java:

`
public static EvenAddition_Compile.Even freshEven() {
  return new EvenAddition_Compile.Even();
}

public static EvenAddition_Compile.Even MockSum() {
  EvenAddition_Compile.Even eMock = org.mockito.Mockito.mock(EvenAddition_Compile.Even.class);
  org.mockito.Mockito.when(eMock.Sum(java.math.BigInteger.valueOf(2L), java.math.BigInteger.valueOf(2L))).thenReturn(java.math.BigInteger.valueOf(3L));
  return eMock;
}
`


.gitignore
```diff
@@ -63,4 +63,10 @@ Source/.vs
...
 package.json
-package-lock.json
\ No newline at end of file
+package-lock.json
+
+
+lecture_programs
+natural_doc
+natural_config
+Working\ Data
```
fabiomadge, line 72: Are you sure?

docs/DafnyRef/Attributes.md
```diff
@@ -343,18 +343,46 @@ recursion was explicitly requested.
 * If `{:tailrecursion true}` was specified but the code does not allow it,
 an error message is given.
 
+<<<<<<< HEAD
```
fabiomadge, line 346: Those markers shouldn't be here.

Test/DafnyTests/TestAttribute.dfy
```diff
@@ -1,280 +0,0 @@
...
-// CHECK: .*_module.__default.FailingTestUsingExpectWithMessage.*
-// CHECK: .*_module.__default.FailingTestUsingExpect.*
-// CHECK: .*_module.__default.FailingTestUsingNoLHSAssignOrHalt.*
-// CHECK: .*_module.__default.FailingTestUsingAssignOrHalt.*
-// CHECK-NOT: .*PassingTest.*
-
-include "../exceptions/VoidOutcomeDt.dfy"
-include "../exceptions/NatOutcomeDt.dfy"
-
-function method SafeDivide(a: nat, b: nat): NatOutcome {
```
fabiomadge, line 19: Where did these tests move?

Source/Dafny/Verifier/Translator.cs
```diff
@@ -11790,4 +11790,4 @@ class VariableNameVisitor : Boogie.StandardVisitor {
 
     static readonly List<Boolean> Bools = new List<Boolean> { false, true };
   }
-}
```
fabiomadge, line 11793: Do we need that?

Source/Dafny/Compilers/Synthesizer-Java.cs
```diff
@@ -0,0 +1,241 @@
...
+
+    var tmpId = compiler.idGenerator.FreshId("tmp");
+    for (int i = 0; i < forallExpr.BoundVars.Count; i++) {
+      var boundVar = forallExpr.BoundVars[i];
+      var varType = compiler.TypeName(boundVar.Type, wr, boundVar.tok);
+      bounds[boundVar] = $"org.mockito.Mockito.any({varType}.class)";
+    }
+    SynthesizeExpression(wr, binaryExpr);
+  }
+}
```
fabiomadge, line 241:
```suggestion
}

```

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -102,6 +102,8 @@ public abstract class SinglePassCompiler : Plugins.Compiler {
     protected interface IClassWriter {
       ConcreteSyntaxTree/*?*/ CreateMethod(Method m, List<TypeArgumentInstantiation> typeArgs, bool createBody, bool forBodyInheritance, bool lookasideBody);
       ConcreteSyntaxTree/*?*/ SynthesizeMethod(Method m, List<TypeArgumentInstantiation> typeArgs, bool createBody, bool forBodyInheritance, bool lookasideBody);
+      // ConcreteSyntaxTree/*?*/ FreshMethod(Method m, List<TypeArgumentInstantiation> typeArgs, bool createBody, bool forBodyInheritance, bool lookasideBody);
```
fabiomadge, line 105:
```suggestion
```

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -2690,7 +2712,6 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
     void TrStmt(Statement stmt, ConcreteSyntaxTree wr) {
       Contract.Requires(stmt != null);
       Contract.Requires(wr != null);
-
```
fabiomadge, line 2693: Let's keep that line.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -142,6 +142,11 @@ protected class ClassWriter : IClassWriter {
         throw new NotImplementedException();
       }
 
+      public ConcreteSyntaxTree CreateFreshMethod(Method m) {
+         throw new NotImplementedException();
+        // return javaSynthesizer.CreateFreshMethod(m, Writer(m.IsStatic, true, m));
```
fabiomadge, line 147:
```suggestion
```
Same for the others.

fabiomadge:
Can you explain why we want the `.devcontainer` files?

cpitclaudel:
Beyond low-level code comments, I find myself wondering: do we want this feature built-in into Dafny, or would it be better to expose the right hooks so that it could be implemented as a plugin?

Is there a pressing reason for it to be part of the standard Dafny release, rather than a separately distributed plugin?

(Sorry if I missed previous discussion of this)

cpitclaudel:
Thanks!  I saw that part.  Given that the feature is growing (to support more languages), I'm wondering whether the whole mocking / testing aspect shouldn't be developed out-of tree.

keyboardDrummer:
> > How do you see that working?
> 
> I was hoping we could use this as a way to develop a stable plugin API sufficient for this use case

My guess is that getting to an API that we would feel comfortable calling stable would require some refactoring. By keeping this consumer in our repository, we can do such refactorings without breaking them.

cpitclaudel:
Got it, thanks!

hirataqdees:
Hi Dafny team, 
We, at ARI, are looking into generating some Java tests from a Dafny model, and it would be great if this CR is reviewed/merged with the mainline! :)  

Thanks

keyboardDrummer:
> Hi Dafny team, We, at ARI, are looking into generating some Java tests from a Dafny model, and it would be great if this CR is reviewed/merged with the mainline! :)
> 
> Thanks

Could you resolve the merge conflicts in the PR?

robin-aws (changes requested):
Glad to see there's interest in this PR @hirataqdees!

@cassidywaldrip - we need to get a clearer view of what the proposed changes are here before we can review, especially the `{:synthesize}` semantics changes. There are remnants of other changes in here that are muddying the waters. Can you please update from master and resolve the conflicts?

cassidywaldrip:
Hi @robin-aws, I updated from master and resolved the conflicts. Let me know if there are other issues that come up. I'll try to be available to fix them right away.

docs/DafnyRef/Attributes.md
```diff
@@ -472,21 +476,24 @@ This is mostly helpful for debugging which assertion is taking the most time to
 ### 22.2.22. `{:synthesize}` {#sec-synthesize-attr}
 
 The `{:synthesize}` attribute must be used on methods that have no body and
-return one or more fresh objects. During compilation, 
-the postconditions associated with such a
+return one or more fresh or mocked objects. To create a fresh instance of 
+an object, use the "fresh" parameter (`{:synthesize "fresh"}`). To create a 
+mocked instance of an object, use the "mock" parameter (`{:synthesize "mock"}`).
```
robin-aws, line 481: My intention for suggesting the name `{:synthesize ... }` for this feature was to keep the semantics as abstract as possible: create an object whose behavior aligns with the specification. Using a mocking framework is just one possible implementation strategy, and one that could change in the future: alternatively you could actually synthesize source code in the target language that obeys the specification.

I'm concerned about committing to the semantics of `{:synthesize "fresh"}`, since AFAICT those semantics are to return a new instance of an object without invoking any constructor. Depending on the compilation strategy, that could easily create an object that causes errors/crashes/exceptions when used, rather than the intention (I believe) of producing a valid but underspecified object, whose fields are initialized arbitrarily.

It seems to me that it is only valid to use `{:synthesize "fresh"}` if `fresh(o)` is literally the only specified post-condition. Could you switch implementation strategies automatically based on that instead? Alternatively, could you stick with the mocking implementation instead, in the name of simplicity? (I recall we discussed this earlier but I can't recall the reason for wanting to make this distinction).

robin-aws (changes requested):
Thanks for tidying this up! I've added one comment we should address before I review the rest, and my apologies for not having all the state paged into my head from the last time we discussed it. :) On the other hand, having thought on this for a while may have changed the strength of my opinion on it.

docs/DafnyRef/Attributes.md
cassidywaldrip: I have updated this branch so that it matches the C# version of `synthesize`. I did this by using spies instead of mocks. By using spies, the need for a "fresh" argument is eliminated, as fields can be accessed and updated in a spy (they could not be with a mock). A spy allows for the same customization of behavior as a mock. The difference is that a spy comes with all the functionality of a real object by default, and that functionality can be changed, whereas a mock comes with none of the functionality of a real object by default, so everything must be added.

Test/DafnyTests/TestAttribute.dfy
```diff
@@ -114,7 +114,7 @@ class Even {
   static method AStaticMethod() {}
 }
 
-method {:synthesize} mockUnsafe() 
+method {:synthesize "mock"} mockUnsafe()
```
robin-aws, line 117: We can revert all these "mock" parameters now, correct?

docs/DafnyRef/Attributes.md
robin-aws: Excellent, I'm so much happier with that! Thanks for sticking with this for the better outcome. :)

Please do revert this documentation change since it's not needed any more, aside from the Java specific remarks below.

Source/DafnyRuntime/DafnyRuntimeJava/src/test/java/dafny/EuclideanTest.java
```diff
@@ -236,4 +236,4 @@ void testBigIntegerMod() {
     assertEquals(BigInteger.ONE, DafnyEuclidean.EuclideanModulus(BigInteger.valueOf(13), two));
     assertEquals(BigInteger.ONE, DafnyEuclidean.EuclideanModulus(BigInteger.valueOf(-13), two));
   }
-}
+}
```
robin-aws, line 239: nit: Nice to revert all these trailing newline changes to reduce the review noise.

docs/DafnyRef/Attributes.md
```diff
@@ -476,28 +476,33 @@ This is mostly helpful for debugging which assertion is taking the most time to
...
+Currently, this compilation pass is only supported in C# and Java. To use it for C#, 
+add the latest version of the Moq library to the .csproj file before
+generating the binary. 
 
 Not all Dafny postconditions can be successfully compiled - below is the
 grammar for postconditions that are supported (`S` is the start symbol, `EXPR`
 stands for an arbitrary Dafny expression, and `ID` stands for
-variable/method/type identifiers):
+variable/method/type identifiers). Note that the grammar for Java does not include 
+field accesses, as Mockito does not support mocking fields:
```
robin-aws, line 498: I assume this is still true with the spies approach?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1959,7 +1961,7 @@ protected class NullClassWriter : IClassWriter {
           } else if (m.Body == null && !(c is TraitDecl && !m.IsStatic) &&
                      !(!DafnyOptions.O.DisallowExterns && (Attributes.Contains(m.Attributes, "dllimport") || (IncludeExternMembers && Attributes.Contains(m.Attributes, "extern"))))) {
             // A (ghost or non-ghost) method must always have a body, except if it's an instance method in a trait.
-            if (Attributes.Contains(m.Attributes, "axiom") || (!DafnyOptions.O.DisallowExterns && Attributes.Contains(m.Attributes, "extern"))) {
+            if (Attributes.Contains(m.Attributes, "axiom")) {
```
robin-aws, line 1964: Why is this necessary?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1406,7 +1406,9 @@ protected class NullClassWriter : IClassWriter {
       public ConcreteSyntaxTree SynthesizeMethod(Method m, List<TypeArgumentInstantiation> typeArgs, bool createBody, bool forBodyInheritance, bool lookasideBody) {
         throw new UnsupportedFeatureException(m.tok, Feature.MethodSynthesis);
       }
-
+      public ConcreteSyntaxTree CreateFreshMethod(Method m) {
```
robin-aws, line 1409: We don't need this method any more, correct?

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -540,6 +564,48 @@ protected class ClassWriter : IClassWriter {
       }
     }
 
+    internal string Keywords(bool isPublic = false, bool isStatic = false, bool isExtern = false) {
+      return (isPublic ? "public " : "") + (isStatic ? "static " : "") + (isExtern ? "extern " : "") + (Synthesize && !isStatic && isPublic ? "virtual " : "");
+    }
+
+    internal ConcreteSyntaxTree GetMethodParameters(Method m, List<TypeArgumentInstantiation> typeArgs, bool lookasideBody, bool customReceiver, string returnType) {
+      //var parameters = GetFunctionParameters(m.Ins, m, typeArgs, lookasideBody, customReceiver);
```
robin-aws, line 572: No checking in commented out code please :)

Source/DafnyCore/Compilers/Synthesizer-Java.cs
```diff
@@ -0,0 +1,229 @@
...
+///           | EQUALS && EQUALS
+/// FORALL    = forall BOUNDVARS :: EXPR ==> EQUALS
+/// ARGLIST   = ID   // this can be one of the bound variables
+///           | EXPR // this expr may not reference any of the bound variables 
+///           | ARGLIST, ARGLIST
+/// BOUNDVARS = ID : ID 
+///           | BOUNDVARS, BOUNDVARS
+/// 
+/// </summary>
+public class JavaSynthesizer {
```
robin-aws, line 27: This generally LGTM, but it seems like there's a ripe opportunity to introduce a common base class for this and  the `CsharpSynthesizer` class to reduce the duplication. I don't want to further delay this PR now that we're better aligned on the design, but I'd appreciate it if you could time-box a bit of time to refactor - if it becomes a rathole I'd settle for a TODO comment.

robin-aws (changes requested):
Great news on not needing "mock" / "fresh" any more! I've done a proper review now and I think we're pretty close here.

(since I can't comment on deleted files) Why was DafnyTests.csproj deleted?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
cassidywaldrip: Correct!

Source/DafnyCore/Compilers/SinglePassCompiler.cs
cassidywaldrip: It is not - I must have been using it before. Changed!

Test/DafnyTests/TestAttribute.dfy
cassidywaldrip: Yes!

cassidywaldrip:
Yes- I will do this today

robin-aws:
LGTM! Just a couple final bookkeeping tasks, if you please:

1. Remove the line here stating that the Java backend doesn't support synthesis? https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Compilers/Compiler-java.cs#L47. That will update the reference manual's table of supported features per compiler.
2. Add a `docs/dev/2024.feat` file with a release note for this, something along the lines of "The Java backend now supports the {:synthesize} attribute".
3. Update the PR description to account for the design change since it was first cut (I'll use it as the commit message when we squash and merge this)

Thanks so much for the contribution!

Issue #2026, opened 2022-04-18
NullReferenceException during counterexample generation


dijkstracula:
On Dafny with /Version output `Dafny 3.5.0.40314`:

Consider the following Dafny program:

```dafny
method foo(n: nat) returns (ret: array<string>)
{
    ret := new string[n];

    var i := 0;
    while i < n
        invariant 0 <= i <= n
        invariant forall j :: 0 <= j < i ==> j % 2 == 0 ==> ret[j] == "even"
        invariant forall j :: 0 <= j < i ==> j % 2 == 1 ==> ret[j] == "odd"
    {
        if i % 2 == 0 {
            ret[i] := "odd";
        } else {
            ret[i] := "even";
        }
        i := i + 1;
    }
}
```

Dafny correctly indicates that our loop invariants are incorrect (as the implementation swapped the cases).  The VS Code plugin is able to show counterexamples, but the CLI interface has a few problems:

## 1. Outdated documentation string for the `/extractCounterexample` flag:

The CLI interface explicitly states that some additional flags need to be set.

```
$ ~/code/dafny/Binaries/Dafny Foo.dfy /help | grep -A4 '/extractCounter'
/extractCounterexample
    If verification fails, report a detailed counterexample for the first
    failing assertion. Requires specifying the /mv option as well as
    /proverOpt:0:model_compress=false and /proverOpt:0:model.completion=true.
/verificationLogger:<configuration string>
➜  fizzbuzz git:(main) ✗
```

However, the `model_compress` option is no longer known to Dafny:

```
$ ~/code/dafny/Binaries/Dafny Foo.dfy /extractCounterexample /mv model /proverOpt:0:model_compress=false
/proverOpt:0:model.completion=true
Fatal Error: ProverException: Unrecognised prover option: 0:model_compress=false
```

### Crash during counterexample generation

More seriously, a null pointer exception is raised while attempting to extract a counterexample (irrespective of whether or not those unknown flags are present):

```
$ ~/code/dafny/Binaries/Dafny Foo.dfy /extractCounterexample /mv model
/Users/ntaylor/school/phd/projects/dafny_sandbox/fizzbuzz/Foo.dfy(8,18): Error: This loop invariant might not be maintained by the loop.
/Users/ntaylor/school/phd/projects/dafny_sandbox/fizzbuzz/Foo.dfy(8,18): Related message: loop invariant violation
Execution trace:
    (0,0): anon0
[...]

PR #2027, opened 2022-04-18
Remove mention of deprecated flags in /extractCounterexample help text

dijkstracula:
Per https://github.com/dafny-lang/dafny/issues/2026 :

Previously, the help string for /extractCounterexample would make
mention of certain proverOpt flags that have seemingly since been
removed.  This patch removes mention of them from the help string
(and makes explicit that /mv takes a file path argument).

Fixes #

### DIFF
--- a/Source/Dafny/DafnyOptions.cs
+++ b/Source/Dafny/DafnyOptions.cs
@@ -1093,7 +1093,7 @@ Exit code: 0 -- success; 1 -- invalid command-line; 2 -- parse or type errors;
 /extractCounterexample
     If verification fails, report a detailed counterexample for the first
     failing assertion. Requires specifying the /mv option as well as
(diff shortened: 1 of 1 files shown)


atomb:
It turns out that the error in this message is smaller than it seems. Those prover options _are_ supported, but the message mistakenly uses `0` where it should use `O`. Running `dafny Foo.dfy /extractCounterexample /mv model /proverOpt:O:model_compress=false` on the example from #2026 works for me.

dijkstracula:
Aha, you know, I _thought_ I saw mention of `O` in the codebase and thought something seemed awry, but didn't investigate.

I'l modify this patch to address this smaller problem.

dijkstracula:
Pushed the smaller fix, with an amended commit message.

atomb (approved):
Thanks for fixing this!

PR #2037, opened 2022-04-20
Fix default Makefile target

### LINKED ISSUE #1617: Default Makefile target is broken
It looks like the phony `runtime` target was removed, but it is required by `default` and `all`.

atomb:
Fixes #1617

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,12 @@
 DIR=$(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 
-default: parser exe
+default: exe
 
 all: exe refman
 
 exe:
 	(cd ${DIR} ; dotnet build Source/Dafny.sln ) ## includes parser
 
-parser:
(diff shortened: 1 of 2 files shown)


cpitclaudel (changes requested):
The parser target can be removed: it's done automatically by the `exe` target.  Keeping the boogie target may be useful, but the prerequisite looks suspicious on line 10, and  we now put `boogie` in` ${DIR}/boogie`, not in `Source/boogie`

atomb:
> The parser target can be removed: it's done automatically by the `exe` target. Keeping the boogie target may be useful, but the prerequisite looks suspicious on line 10, and we now put `boogie` in` ${DIR}/boogie`, not in `Source/boogie`

Good catch. I re-added the `boogie` target and fixed `Scripts/use-local-boogie.sh` to work with it.

cpitclaudel (approved):

Issue #1363, opened 2021-08-21
soundness: refining datatype adding new field introduces unsoundness


erniecohen:
the following verifies in 3.2:
```
module B {
    datatype T = T() {
        lemma oneVal(t:T) ensures t == this {}
    }
     
}
module C refines B {
    datatype T = ... T(x:bool) {
        lemma oops() ensures false { oneVal(this.(x := !x));   }
    }
}
```

RustanLeino:
This is a bug. The intention is that once a `datatype` is defined, you are not allowed to change what constructors it has or what parameters the constructors take.

Issue #1963, opened 2022-04-04
Missing refinement checks between newtypes and datatypes


cpitclaudel:
Dafny incorrectly accepts the following two programs.

For the first one, the fix is either to make sure that newtype refinement preserves constraints (here `x < 0` is incompatible with `x >= 0`) or to recheck the postcondition of the lemmas in refined modules.  Fort the second one, I'm not sure how the feature is intended to work (should refinement be allowed to add constructors?

```dafny
module A {
  newtype NT = x: int | x >= 0 witness 0
  lemma P(x: NT) ensures x >= 0 {}
}

module B refines A {
  newtype NT = ... x: int | x < 0 witness -1
}

method M() ensures false {
  assert -1 >= 0 by { B.P(-1); }
}
```

```dafny
module A {
  datatype D = D
  lemma P(d: D) ensures d == D {}
}

module B refines A {
  datatype D = ... D | D'
}

method M() ensures false {
  assert B.D == B.D' by { B.P(B.D'); }
}
```

PR #2044, opened 2022-04-22
fix: Protect newtypes and datatypes from refinement

cpitclaudel:
Closes GH-1363, GH-1963.

Here's my understanding of the current refinement situation (trying to prevent similar bugs.  I have put 🆗 for places that I'm decently confident make sense, ❓ when I don't know, 🐛 for known issues, and an additional ✔️ when we checked further with @RustanLeino.

- `ModuleExport`: ellipsis comes before exports 🆗
- `ClassDecl`, `TraitDecl`: ellipsis replaces `extends` 🆗 
- `DatatypeDecl`: ellipsis comes before "|" (`DatatypeMemberDecl`) 🐛
  **FIXED, now comes instead of `|` and constructors**
- `ConstantFieldDecl`: ellipsis comes before `:=`  ❓  ✔️ 
- `NewtypeDecl`: ellipsis comes after `=`, before type and subset constraint 🐛
  **FIXED, now comes instead of `=`, type, and constraints**
- `SynonymTypeDecl`: ellipsis replaces `=` and definition 🆗 
- `MethodDecl`, `FunctionDecl`, `IteratorDecl`: ellipsis replaces signature 🆗 
- `SkeletonStmt`: ellipsis comes before `;` 🆗 
- `IfStmt`: ellipsis replaces guard 🆗
- `WhileStmt`: ellipsis replaces guard or body 🆗 
- `AssertStmt`, `AssumeStmt`: ellipsis replaces body (follows attributes) 🆗 (❓ for attributes) ✔️
- `ModifyStmt`: ellipsis replaces frame expressions ❓ 🐛 (follow up in https://github.com/dafny-lang/dafny/issues/2056)

Question: `ConstantFieldDecl` allows `const x ... := 1` to refine `const x: nat`.  But `const x := 1` is also a valid refinement of `const x: nat`, so the refinement is unnecessary.  In contrast, `ghost const x ...` is not allowed (must have type or definition), so to ghostify one must write `ghost const x: nat`.  Should we change anything here? → Moved to https://github.com/dafny-lang/dafny/issues/2057

### DIFF
--- a/Source/Dafny/RefinementTransformer.cs
+++ b/Source/Dafny/RefinementTransformer.cs
@@ -171,7 +171,7 @@ namespace Microsoft.Dafny {
 
         Contract.Assert(RefinedSig.ModuleDef != null);
         Contract.Assert(m.RefinementQId.Def == RefinedSig.ModuleDef);
-        // check that the openess in the imports between refinement and its base matches
+        // check that the openness in the imports between refinement and its base matches
         List<TopLevelDecl> declarations = m.TopLevelDecls;
         List<TopLevelDecl> baseDeclarations = m.RefinementQId.Def.TopLevelDecls;
         foreach (var im in declarations) {
@@ -196,7 +196,8 @@ namespace Microsoft.Dafny {
         PreResolveWorker(m);
       } else {
         // do this also for non-refining modules
-        CheckSuperflousRefiningMarks(m.TopLevelDecls, new List<string>());
+        CheckSuperfluousRefiningMarks(m.TopLevelDecls, new List<string>());
+        AddDefaultBaseTypeToUnresolvedNewtypes(m.TopLevelDecls);
       }
     }
 
@@ -242,7 +243,8 @@ namespace Microsoft.Dafny {
           }
         }
       }
-      CheckSuperflousRefiningMarks(m.TopLevelDecls, processedDecl);
+      CheckSuperfluousRefiningMarks(m.TopLevelDecls, processedDecl);
+      AddDefaultBaseTypeToUnresolvedNewtypes(m.TopLevelDecls);
 
       // Merge the imports of prev
       var prevTopLevelDecls = RefinedSig.TopLevels.Values;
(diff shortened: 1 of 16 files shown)


Source/Dafny/RefinementTransformer.cs
```diff
@@ -269,6 +271,19 @@ public RefinementTransformer(Program p)
...
     }
 
+    /// <summary>
+    /// Give unresolved newtypes a reasonable default type (<c>int</c>), to avoid having to support `null` in the
+    /// rest of the resolution pipeline.
+    /// </summary>
+    private void AddDefaultBaseTypeToUnresolvedNewtypes(List<TopLevelDecl> topLevelDecls) {
+      foreach (var d in topLevelDecls) {
+        if (d is NewtypeDecl { IsRefining: true, BaseType: null } decl) {
+          Reporter.Info(MessageSource.RefinementTransformer, decl.tok, $"defaulting to 'int' for unspecified base type of '{decl.Name}'");
```
RustanLeino, line 281: This happens only when the `newtype T ...` doesn't actually refine a `newtype` in a parent module, right? (Otherwise, this info message would come out for all program that refine a `newtype` and that would seem confusing to a user.)

Test/git-issues/git-issue-1963a.dfy
```diff
@@ -0,0 +1,17 @@
...
+
+/// This file checks that refinement is forbidden for newtypes
+
+module A {
+  newtype NT = x: int | x >= 0 witness 0
+  lemma P(x: NT) ensures x >= 0 {}
+}
+
+module B refines A {
+  newtype NT = ... x: int | x < 0 witness -1
```
RustanLeino, line 12: I suggest you add a comment on this line, saying that we expect an error. This makes it easier to read the test files (without looking up every error in the corresponding .expect file).

Test/git-issues/git-issue-1963b.dfy
```diff
@@ -0,0 +1,17 @@
...
+
+/// This file checks that refinement is forbidden for datatypes
+
+module A {
+  datatype D = D
+  lemma P(d: D) ensures d == D {}
+}
+
+module B refines A {
+  datatype D = ... D | D'
```
RustanLeino, line 12: Same here. Add a comment that says this line gives an error.

RustanLeino (approved):
I like the new, consistent syntax (and the unsoundness fixes).

Source/Dafny/RefinementTransformer.cs
cpitclaudel: Correct, `BaseType == null` captures that

Test/git-issues/git-issue-1963a.dfy
cpitclaudel:
```suggestion
  newtype NT = ... x: int | x < 0 witness -1 // Error: Change (or repeat) a newtype body
```

Test/git-issues/git-issue-1963b.dfy
cpitclaudel:
```suggestion
  datatype D = ... D | D' // Error: Cannot change (or repeat) constructors
```

RustanLeino (approved):

PR #2051, opened 2022-04-22
Enable packaging Dafny as a .NET tool

### LINKED ISSUE #1779: Package Dafny as a .NET Core CLI tool
https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools

This would replace our bespoke packaging and installation procedure and Homebrew, simplifying our release process and avoiding users having to do silly things like run a `allow_on_mac.sh` script. Boogie is already doing the same thing: https://github.com/boogie-org/boogie#installation

The main wrinkle is not being able to package Z3 along with it, but we can likely workaround that by having the Dafny CLI offer to download Z3 for you on the first run.

atomb:
This makes the first steps toward packaging Dafny as a .NET CLI tool.

Mostly fixes #1779, but doesn't yet auto-install Z3. We can do that as part of a later PR.


mschlaipfer:
What about the target language runtimes and supporting the `/compile:3` argument? How is that different from `z3`? Is it because the default is not `/noVerify` so `z3` is required whereas the target compilers are optional?

atomb:
> What about the target language runtimes and supporting the `/compile:3` argument? How is that different from `z3`? Is it because the default is not `/noVerify` so `z3` is required whereas the target compilers are optional?

That's a good question. I think the idea is that all the target language runtimes would be available through the packaging system for the respective languages (and the C# runtime is prepared to be available through NuGet, so C# compilation and execution should work smoothly).

Alternatively, I expect the source files for the target language runtimes could be included as data files in the Dafny package, and spit out alongside the generated code (as I think would be necessary for C++?). They're at least not platform-dependent, whereas Z3 binaries are.

robin-aws:
> Alternatively, I expect the source files for the target language runtimes could be included as data files in the Dafny package, and spit out alongside the generated code (as I think would be necessary for C++?). 

This is the approach we should use for now, including the `DafnyRuntime.cs` file even if/when it's possible to get the `DafnyRuntime.dll` artifact from nuget. The default behavior is still to prepend that code to the compiler output anyway.

I'm assuming that `dotnet tool install --global ...` works as well? That's how I'll want to document the common installation process for users, as that will add the `dafny` CLI to the PATH so you don't need the `dotnet tool run ...` part.

.github/workflows/publish-release.yml
```diff
@@ -80,3 +80,7 @@ jobs:
           dafny/Package/dafny-${{ steps.get-version.outputs.version-without-v }}*
           dafny/docs/DafnyRef/DafnyRef.pdf
         fail_on_unmatched_files: true
+    - name: Create NuGet package (for uploading)
+      run: dotnet pack --no-build dafny/Source/Dafny.sln
+    #- name: Upload package to NuGet
+    #  run: dotnet nuget push "Binaries/Dafny*.nupkg" -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
```
fabiomadge, line 86:
```suggestion
    - name: Upload package to NuGet
      run: dotnet nuget push "Binaries/Dafny*.nupkg" -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
```

Source/Dafny/DafnyPipeline.csproj
```diff
@@ -34,4 +34,28 @@
   <ItemGroup>
     <Content Include="DafnyPrelude.bpl" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Include="..\DafnyRuntime\DafnyRuntime.csproj">
+      <LogicalName>DafnyRuntime.csproj</LogicalName>
+    </EmbeddedResource>
```
fabiomadge, line 41: Would this also work if done in `DafnyRuntime`?

fabiomadge (dismissed):
How did we deal with z3?

atomb:
> How did we deal with z3?

The current plan is to have it be an implicit dependency that must be installed separately for the moment, but perhaps to add functionality to automatically download it at some point in the future.

Source/Dafny/DafnyPipeline.csproj
fabiomadge: Would adding that dependency be so bad?

RELEASE_NOTES.md
```diff
@@ -1,6 +1,7 @@
 # Upcoming
 
 - feat: The IDE will show verification errors for a method immediately after that method has been verified, instead of after all methods are verified.
+- feat: Dafny is now available on NuGet, installable with `dotnet tool install Dafny`.
```
robin-aws, line 4:
```suggestion
- feat: Dafny is now available on NuGet, installable with `dotnet tool install --global Dafny`.
```

Debatable perhaps, but it seems like the most common use case by far is putting it on your PATH so you can just run `dafny` on the command line.

I would also be more specific about the set of packages we're publishing than just "Dafny" :)

Source/Dafny/Compilers/Compiler-go.cs
```diff
@@ -68,11 +68,12 @@ private struct Import {
...
       // Keep the import writers so that we can import subsequent modules into the main one
       EmitImports(wr, out RootImportWriter, out RootImportDummyWriter);
+
+      var rt = wr.NewFile("dafny/dafny.go");
+      ReadRuntimeSystem(program, "DafnyRuntime.go", rt);
     }
 
     protected override void EmitBuiltInDecls(BuiltIns builtIns, ConcreteSyntaxTree wr) {
-      var rt = wr.NewFile("dafny/dafny.go");
-      ReadRuntimeSystem("DafnyRuntime.go", rt);
```
robin-aws, line 75: Why did this have to move?

RELEASE.md
```diff
@@ -16,11 +16,20 @@
...
    of the release date (e.g., a release on January 12th, 2022 would be
-   x.y.z.40112). Edit the internal version number in
-   `Source/version.cs`, replace the "Upcoming" header in `RELEASE_NOTES.md`
-   with the the public version number, and add a new "Upcoming" header above it.
-   Push and cut a PR, get it approved,
-   and squash and merge those changes to the `master` branch.
+   x.y.z.40112). Edit the internal version number in the following
+   places:
+
+   * `Source/version.cs`
```
robin-aws, line 22: I'd add a comment to this file noting that the other two instances have to be kept in sync, just in case.

Source/Dafny/DafnyPipeline.csproj
atomb: It seems like it would be sort of a strange thing to do, to me. The `DafnyRuntime` project is a C# library intended for use by generated C# code. It doesn't make sense to me for it include, for instance, `DafnyRuntime.js` as an embedded resource. Said differently, it makes more sense, in my view, for the compiler to bake in the code for the various runtime systems, rather than having the runtime system for one language include the runtime systems for other languages.

robin-aws (changes requested):
Today is a great day for Dafny. :)

Mostly just questions, but one other blocker: we need a copy of `release-downloads.yml` that installs the .NET tool version and runs the same smoke test. I don't see an easy way of triggering that automatically, so we should just make it a manual step of the release process for now.

Source/Dafny/DafnyPipeline.csproj
fabiomadge: I disagree, `DafnyRuntime.csproj` already includes the runtimes for all target languages and is included by `Dafny.sln`. In fact, we probably shouldn't embed it here at all.

RELEASE.md
atomb: Good thought.

Source/Dafny/DafnyPipeline.csproj
fabiomadge: I think we should either get rid of the `csproj` entirely, or use it as much as possible.

atomb: Another way to look at it: if we embed those files in `DafnyRuntime.csproj`, then the `DafnyRuntime.dll` file that users download from NuGet to link with their generated C# code will also include the text of the various other language runtimes. It wouldn't cause problems to do that, but it seems awkward to me.

atomb:
There's one thing I didn't mention earlier but that is a limitation of the current implementation. The `DafnyRuntime.h` and `DafnyRuntime.jar` files are never read by the Dafny compiler. So, although they were embedded in the `DafnyPipeline` assembly (until I removed them in a recent commit) included in the NuGet package, there was never a way to get them out. We could potentially modify how those compilers work, or add a flag to spit out those files, but that seemed like too large a functional change to include in this PR.

.github/workflows/msbuild.yml
```diff
@@ -49,6 +49,8 @@ jobs:
     - name: Check whitespace and style
       working-directory: dafny
       run: dotnet tool run dotnet-format -w -s error --check Source/Dafny.sln --exclude Dafny/Scanner.cs --exclude Dafny/Parser.cs
+    - name: Create NuGet package (just to make sure it works)
+      run: dotnet pack --no-build dafny/Source/Dafny.sln
```
fabiomadge, line 53:
```suggestion
      run: dotnet pack --no-build Source/Dafny.sln
```

fabiomadge (approved):

Source/DafnyLanguageServer/DafnyLanguageServer.csproj
```diff
@@ -8,6 +8,7 @@
     <RootNamespace>Microsoft.Dafny.LanguageServer</RootNamespace>
     <OutputPath>..\..\Binaries\</OutputPath>
     <ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>
+    <PackageVersion>3.6.0.40511</PackageVersion>
```
keyboardDrummer, line 11: What's the use of publishing this as a package?

Source/Dafny/DafnyPipeline.csproj
```diff
@@ -34,4 +34,19 @@
   <ItemGroup>
```
keyboardDrummer, line 34: Shouldn't `DafnyPipeline` be published as package for Dafny as a library use-cases?

Source/Dafny/DafnyPipeline.csproj
keyboardDrummer: I think we should have separate projects for each back-end, and each of those should include their own runtime, and then `DafnyDriver.csproj` should reference those, not DafnyPipeline.

Without making all those changes, I think `DafnyDriver.csproj` can reference `DafnyRuntime.csproj`

Source/Dafny/DafnyPipeline.csproj
atomb: I think so, but I was thinking that we'd make the modifications you mention [here](https://github.com/dafny-lang/dafny/issues/2084#issuecomment-1115870493) before doing that. If you think the library would be useful as-is then, yeah, let's include it.

Source/DafnyLanguageServer/DafnyLanguageServer.csproj
atomb: Do you know whether it's possible to set it up so that the language server gets distributed along with the main CLI? And, if so, how to do it? I think that'd be ideal, but wasn't sure if it would be possible, at least without significant refactoring.

keyboardDrummer: I'm sure it possible, and I think we should, but we should spend a little more time discussing whether we want it. For now, maybe just not publish DafnyLanguageServer because unless you see a reason why it would be useful to consume for other .NET projects.

Source/Dafny/DafnyPipeline.csproj
keyboardDrummer: I think we should publish them now. Customers can always copy things from DafnyDriver if they're missing something that's only there.

Source/Dafny/DafnyPipeline.csproj
```diff
@@ -34,4 +34,19 @@
...
     <Content Include="DafnyPrelude.bpl" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
+
+  <ItemGroup>
```
keyboardDrummer, line 38: I don't think DafnyPipeline should reference DafnyRuntime in any way. Can't we have these references in DafnyDriver?

Source/Dafny/DafnyPipeline.csproj
atomb: Yeah, I like the main proposal you're making, @keyboardDrummer. Each back-end should be a separate project.

Given the current state of the code, though, the reason for embedding those resources in `DafnyPipeline` for the moment is so that the compilers can get a reference to them through [`GetExecutingAssembly`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getexecutingassembly?view=net-6.0). I originally put them in `DafnyDriver`. It would be possible to move them back into `DafnyDriver` and instead use [`GetEntryAssembly`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=net-6.0).

I had previously only seen methods in the `Assembly` class to load by type, and types in `DafnyRuntime` aren't available in the compilers, which are currently in `DafnyPipeline`, unless we make `DafnyRuntime` a dependency of `DafnyPipeline`, which is pretty much the opposite of what you're proposing. We could potentially use [`Load`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.load?view=net-6.0) to load `DafnyRuntime` by name. I prefer something that doesn't require a name as a string, but not strongly.

Source/DafnyLanguageServer/DafnyLanguageServer.csproj
atomb: Sounds reasonable.

Source/Dafny/DafnyPipeline.csproj
atomb: I changed my mind on this yet again. The issue I came across is that the `DafnyPipeline` project, which we're making available as a library, includes the compilers. If you invoke those compilers and their assembly (or, as an option, possibly some other one that's available at runtime) doesn't contain the runtime files, they'll crash. So the runtime files can't reasonably go in `DafnyDriver`. They ultimately should go in language-specific compiler packages, as I think we all agree. But I still feel like it'd be messy to put them in the `DafnyRuntime` assembly (even though the source code is in that package). So that leaves `DafnyPipeline`, for now, until we split off the individual compilers.

keyboardDrummer (approved):

keyboardDrummer (dismissed):

RELEASE.md
```diff
@@ -16,11 +16,20 @@
...
+
+   * `Source/DafnyDriver/DafnyDriver.csproj`
+
+   * `Source/DafnyLanguageServer/DafnyLanguageServer.csproj`
+
+   Put the public version number in place of the "Upcoming" header in
+   `RELEASE_NOTES.md`, and add a new "Upcoming" header above it.
+
+   Push and cut a PR, get it approved, and squash and merge those
+   changes to the `master` branch.
```
robin-aws, line 33: You need a step after #7 to manually trigger the new `release-downloads-nuget.yml` workflow too.

Scripts/package.py
```diff
@@ -255,11 +255,12 @@ def check_version_cs(args):
     # Checking version.cs
     fp = open(path.join(SOURCE_DIRECTORY,"version.cs"))
     lines = fp.readlines()
-    qstart = lines[2].index('"')
-    qend = lines[2].index('"', qstart+1)
-    lastdot = lines[2].rindex('.',qstart)
-    v1 = lines[2][qstart+1:lastdot]
-    v2 = lines[2][lastdot+1:qend]
+    verline = lines[5]
```
robin-aws, line 258: Yuck - there has to be a better way of tracking the version number in the future :P

Source/Dafny/Compilers/Compiler-java.cs
```diff
@@ -16,6 +16,7 @@
 using System.Diagnostics;
 using System.Text.RegularExpressions;
 using System.Reflection;
+using Microsoft.Win32.SafeHandles;
```
robin-aws, line 19: Is this used somehow? Hopefully not?

Source/Dafny/Compilers/Compiler-java.cs
```diff
@@ -2195,6 +2196,19 @@ private class GenericArrayElementLvalue : ILvalue {
       }
     }
 
+    private void EmitRuntimeJar(string targetDirectory) {
```
robin-aws, line 2199: This is definitely "weird", but not a lot weirder than the old way. :) 

It's at least more consistent with the behavior of the other languages.

Source/Dafny/Compilers/Compiler-java.cs
```diff
@@ -2207,8 +2221,12 @@ private class GenericArrayElementLvalue : ILvalue {
           return false;
         }
       }
+
+      var targetDirectory = Path.GetDirectoryName(targetFilename);
+      EmitRuntimeJar(targetDirectory);
```
robin-aws, line 2226: Hmm...this should probably only happen if `DafnyOptions.O.UseRuntimeLib` is `false`. If it's `true`, the assumption is that you'll add `DafnyRuntime.jar` to the classpath yourself when you compile/run later. How hard is it to extract `DafnyRuntime.jar` from the nuget package after you've installed it?

I don't think this is blocking since existing codebases since we're still publishing the zip files as well, but we should probably make sure we start publishing `DafnyRuntime.jar` as well before we STOP publishing the zip files.

robin-aws (changes requested):

Source/Dafny/Compilers/Compiler-java.cs
atomb: Ah, IDEs! I was exploring various APIs to use to emit `DafnyRuntime.jar`, and my IDE must have inserted that while I had something temporary in place that I removed as soon as I read the docstring. :)

Source/Dafny/Compilers/Compiler-java.cs
atomb: That's an easy check to add, so I'll do it even if it's not immediately critical.

atomb: Oh, and in response to your question, I don't know of any way to extract the embedded files from an assembly other than through the .NET API, but there probably is one.

robin-aws (approved):

PR #2053, opened 2022-04-22
C# Runtime: include `BigRational` constructor taking `double`

### LINKED ISSUE #1940: Allow conversion between rationals and floats in Dafny runtime
The current Dafny runtime (for C#, at least) uses arbitrary precision rationals to represent real numbers, which provides nice precision but interacts awkwardly with external code that uses floating point. It would be convenient to allow external code to easily convert these values to and from floating-point values.

atomb:
This PR adds a constructor to the `BigRational` data type in the C# runtime that takes a `double` as an argument, representing its value exactly as a rational. This constructor isn't currently used by the compiler, but makes external code that uses floating-point values more easily able to interact with `real` values in Dafny (in a sound way).

This PR also makes the `num` and `den` fields of `BigRational` `internal` so client code should be able to read their values to construct a floating-point approximation of a `real` value. The choice to expose these fields rather than including a built-in approximation is for two reasons. First, by making clients implement their own (potentially simple) approximation, it makes it clear that it is an approximation. Second, there are multiple ways to approximate the rational -> float conversion, and different clients may have different preferences.

Fixes #1940.


Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -1552,11 +1553,69 @@ public struct BigRational {
...
+        throw new ArgumentException(
+          "Can't convert +/- infinity to a rational.");
+      }
+      if (Double.IsSubnormal(n)) {
+        throw new ArgumentException(
+          "Can't convert a subnormal value to a rational (yet).");
+      }
+
+      // Double-specific values
+      const int exptBias = 1023;
```
cpitclaudel, line 1592: Can you comment on the choice of algorithm?  The one implementation I'm familiar with is https://github.com/python/cpython/blob/da6c78584b1f45ce3766bf7f27fb033169715292/Objects/floatobject.c#L1560

atomb: I derived it mostly from first principles based on the IEEE standard. I did look [here](https://github.com/ghc/ghc/blob/master/libraries/base/GHC/Float.hs) to some extent, but didn't follow it entirely.

atomb: It'd honestly likely be faster if I followed that Haskell code more precisely, because it has no iteration and no multiplication.

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -1509,7 +1509,8 @@ public struct BigRational {
     // We need to deal with the special case "num == 0 && den == 0", because
     // that's what C#'s default struct constructor will produce for BigRational. :(
     // To deal with it, we ignore "den" when "num" is 0.
-    BigInteger num, den;  // invariant 1 <= den || (num == 0 && den == 0)
+    internal BigInteger num, den;  // invariant 1 <= den || (num == 0 && den == 0)
```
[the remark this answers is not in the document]
atomb: Oh, good call! I'll do that.

cpitclaudel:
Would you mind spot-checking the following program, and seeing if it gives an all-clear?  Testing all doubles would be prohibitive, but at least this should give us some confidence for single-precision floats:


```
class Program {

  // https://stackoverflow.com/a/27238358/695591
  public static unsafe float Int32BitsToSingle(uint value) {
    return *(float*)(&value);
  }

  public static void TestFloats(uint beg, uint end) {
    bool ok = true;
    uint counter = beg;
    double epsilon = Single.Epsilon;
    do {
      var flt = Int32BitsToSingle(counter);
      if (Single.IsNaN(flt) || Single.IsInfinity(flt) || Single.IsSubnormal(flt)) {
        continue; // Explicitly not supported
      }

      var dbl = (double) flt;
      var r = new Dafny.BigRational(dbl);
      ok = (dbl == 0 && r.den.IsZero) || Math.Abs((double) r.num / (double) r.den - dbl) <= epsilon;
      // Console.WriteLine($"{counter} -> {d} -> {r.num}/{r.den} -> {(double) r.num / (double) r.den}");

      counter = unchecked(counter + 1);
      if (counter % 1_048_576 == 0) {
        Console.WriteLine($"{beg}..{end}: {counter} ({unchecked(100.0 * (double)(counter - beg) / (double)(end - beg)):F3}%)");
      }
    } while (ok && counter != end);

    if (!ok) {
      counter = unchecked(counter - 1);
      Console.WriteLine($"Test failed for counter={counter} ({Int32BitsToSingle(counter)})");
    }
  }

  public static void TestAllFloats() {
    var NBATCHES_LOG2 = 5;
    uint BATCH_SIZE = (uint)1 << (32 - NBATCHES_LOG2);
    int NBATCHES = 1 << NBATCHES_LOG2;
    var steps = Enumerable.Range(0, NBATCHES).Select(i => ((uint) i) * BATCH_SIZE);
    Parallel.ForEach(steps, beg => TestFloats(beg, unchecked(beg + BATCH_SIZE)));
  }

  public static void Main() {
    TestAllFloats();
  }
}
```

atomb:
I ran that driver program overnight, and it didn't report any failures. There might be something wrong with the termination condition, since it stopped printing output (while still using 400% CPU) after an hour or so (with most of the final percentage reports at or close to 100%). Overall, though, I think that's probably pretty good evidence of correctness.

cpitclaudel (approved):

PR #2054, opened 2022-04-22
fix(performance): Provide capacity for the immutable array builder in ConcatSequence.ComputeElements()

robin-aws:
This method calculates the result of a tree of lazily concatenated sequences. Not providing a capacity was functionally correct, but meant the `ImmutableArray<T>.Builder` had to reallocate its internal array and copy values repeatedly as more ranges were added. Since `ConcatSequence` already tracks the total length of the concatenation, we can provide it as the initial capacity for the builder and avoid all these reallocations and copies.


atomb (approved):
Nice when a fix can be this simple!

alex-chew (approved):

PR #2061, opened 2022-04-23
docs: Minor wording change

x11x:
Changed "one of" to "among" because two things are discussed. Also tried to reword it so that it is clearer

### DIFF
--- a/docs/OnlineTutorial/guide.md
+++ b/docs/OnlineTutorial/guide.md
@@ -1823,10 +1823,10 @@ to their `reads` cousins,
 except they say what can be changed, rather than what the value of the function
 depends on. In combination with reads, modification
 restrictions allow Dafny to prove properties of code that would otherwise be
-very difficult or impossible. `reads` and `modifies` are one of the tools that
-allow Dafny to work on one method at a time, because they restrict what would
-otherwise be arbitrary modifications of memory to something that Dafny can
-reason about.
+very difficult or impossible. `reads` and `modifies` are among the tools that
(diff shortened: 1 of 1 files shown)


docs/OnlineTutorial/guide.md
```diff
@@ -1823,10 +1823,10 @@ to their `reads` cousins,
 except they say what can be changed, rather than what the value of the function
 depends on. In combination with reads, modification
 restrictions allow Dafny to prove properties of code that would otherwise be
-very difficult or impossible. `reads` and `modifies` are one of the tools that
-allow Dafny to work on one method at a time, because they restrict what would
-otherwise be arbitrary modifications of memory to something that Dafny can
-reason about.
+very difficult or impossible. `reads` and `modifies` are among the tools that
+make it possible for Dafny to work on one method at a time, because they
```
fabiomadge, line 1827: Not sure this is better.

fabiomadge:
Thanks for your suggestions! Unless you can _allow_ is unclear, I'll only merge the first one.

x11x:
Sorry for the delay. "Allow" is fine, not sure why I changed that.
Also, I realized the sense of the original sentence was that "reads and modifies are [together/combined] one of the tools that....", maybe you prefer something like that?
This is not a huge issue, it makes sense even as it was originally, it just hit my ear strange when I read it.

fabiomadge:
No worries, I just wanted to avoid surprising you by merging something that was diverging from you original intention. How about now?

fabiomadge (approved):

x11x:
Looks good

PR #2079, opened 2022-04-29
fix: calc of multiset crashes

### LINKED ISSUE #2078: Language server not robust enough and multiset keyword crash
Enter the following in VSCode:

```dafny
lemma Something(i: int)
{
  calc {
    
  }
}
```
It parses, resolves and verifies. Now, type in the keyword "multiset" inside the calc statement. Parsing fails.
Now delete "multiset", parsing fails again and everything else is stuck, the language server needs to be restarted manually.
3b22f041-911b-4db5-b2f5-2000a7ea2a1b

To solve this issue, two things.
1) The most im;portant one: The language server should work when canceling modifications.
2) Fix the underlying error.

MikaelMayer:
Fixes #2078 

`calc{multiset}`  triggers a while loop that will try to get all inner statements, but the while loop is guarded by something that checks if the next element is the start of an expression, which it is (`multiset` can be the start of an expression). However, since it is not a valid expression, it will emit an error, store a dummy expression, and continue without changing the scanner's position.

The problem does not arise with `iset` because that keyword can be used both for set display and set comprehension, so the scanner can continue in any case.

Since there is no multiset comprehension for now, I removed the lookahead code so that now multiset always expects a `(` or a `{` after it, and the scanner can move on in any case.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -22,7 +22,7 @@
 - fix: Prevent refinements from changing datatype and newtype definitions (https://github.com/dafny-lang/dafny/pull/2038)
 - feat: The new `older` modifier on arguments to predicates indicates that a predicate ensures the allocatedness of some of its arguments. This allows more functions to show that their quantifiers do not depend on the allocation state. For more information, see the reference manual section on `older`. (https://github.com/dafny-lang/dafny/pull/1936)
 - fix: Fix `(!new)` checks (https://github.com/dafny-lang/dafny/issues/1419)
-- fix: multiset keyword no more crashes the parser (https://github.com/dafny-lang/dafny/pull/2079)
+- fix: multiset keyword no longer crashes the parser (https://github.com/dafny-lang/dafny/pull/2079)
 
 # 3.5.0
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -22,6 +22,7 @@
 - fix: Prevent refinements from changing datatype and newtype definitions (https://github.com/dafny-lang/dafny/pull/2038)
 - feat: The new `older` modifier on arguments to predicates indicates that a predicate ensures the allocatedness of some of its arguments. This allows more functions to show that their quantifiers do not depend on the allocation state. For more information, see the reference manual section on `older`. (https://github.com/dafny-lang/dafny/pull/1936)
 - fix: Fix `(!new)` checks (https://github.com/dafny-lang/dafny/issues/1419)
+- fix: multiset keyword no more crashes the parser (https://github.com/dafny-lang/dafny/pull/2079)
 
 # 3.5.0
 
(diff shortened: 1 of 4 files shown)


RELEASE_NOTES.md
```diff
@@ -22,6 +22,7 @@
 - fix: Prevent refinements from changing datatype and newtype definitions (https://github.com/dafny-lang/dafny/pull/2038)
 - feat: The new `older` modifier on arguments to predicates indicates that a predicate ensures the allocatedness of some of its arguments. This allows more functions to show that their quantifiers do not depend on the allocation state. For more information, see the reference manual section on `older`. (https://github.com/dafny-lang/dafny/pull/1936)
 - fix: Fix `(!new)` checks (https://github.com/dafny-lang/dafny/issues/1419)
+- fix: multiset keyword no more crashes the parser (https://github.com/dafny-lang/dafny/pull/2079)
```
keyboardDrummer, line 25: nitpick: `no longer`

Source/Dafny/Dafny.atg
```diff
@@ -386,7 +386,8 @@ bool IsSetDisplay() {
   if (la.kind == _lbrace) return true;
   int k = scanner.Peek().kind;
   if (la.kind == _iset && k == _lbrace) return true;
-  if (la.kind == _multiset && (k == _lbrace || k == _openparen)) return true;
+  // Until there is a multiset comprehension, there is no point in checking the next item.
```
keyboardDrummer, line 389: I would remove the `//` and the `/*` comment. I don't think a future implementor of multiset comprehensions needs these.

keyboardDrummer (dismissed):

RELEASE_NOTES.md
MikaelMayer, line 25:
```suggestion
- fix: multiset keyword no longer crashes the parser (https://github.com/dafny-lang/dafny/pull/2079)
```

Source/Dafny/Dafny.atg
```diff
@@ -386,7 +386,7 @@ bool IsSetDisplay() {
   if (la.kind == _lbrace) return true;
   int k = scanner.Peek().kind;
   if (la.kind == _iset && k == _lbrace) return true;
-  if (la.kind == _multiset && (k == _lbrace || k == _openparen)) return true;
+  if (la.kind == _multiset) return true;
```
cpitclaudel, line 389: Does this function need to be renamed? Just having the token `"multiset"` isn't enough to know that we're looking at a set display, is it?

cpitclaudel (changes requested):

Source/Dafny/Dafny.atg
MikaelMayer: I had looked at the documentation, and I did not find any other uses of the keyword "multiset" than defining a set display and the multiset type. However, I don't think there is a single case when we could be either parsing an expression or a type, so it's not ambiguous.

cpitclaudel: I'm confused, `multiset<…>` is valid syntax and this function will accept it as a set display, no?
Maybe it would be enough to rename this to `ExprIsSetDisplay`, since we know from the context that we're looking at a set display.

MikaelMayer: Ok, I understand now your renaming suggestion. Let me do it.

cpitclaudel (approved):

Issue #2074, opened 2022-04-28
Null pointer exception on underconstrained closures


MikaelMayer:
```dafny
method m() {
  var f := st => set x | x in st;
}
```

To test it on the command-line:
```
echo "method m() { var f := st => set x | x in st; }" | ./Binaries/Dafny.exe -stdin
```

> <stdin>(1,17): Error: the type of this variable is underspecified
<stdin>(1,36): Error: the type of this variable is underspecified
<stdin>(1,32): Error: type of bound variable 'x' could not be determined; please
 specify the type explicitly
Unhandled exception. System.AggregateException: One or more errors occurred. (Ob
ject reference not set to an instance of an object.)
 ---> System.NullReferenceException: Object reference not set to an instance of
an object.
   at Microsoft.Dafny.Resolver.DiscoverAllBounds_Aux_SingleVar[VT](List`1 bvars,
 Int32 j, Expression expr, Boolean polarity, List`1 knownBounds) in C:\Users\mim
ayere\Documents\dafny\Source\Dafny\Resolver\BoundsDiscovery.cs:line 197

The error is here:

```csharp
          case BinaryExpr.ResolvedOpcode.InSeq:
            if (whereIsBv == 0) {
              bounds.Add(new ComprehensionExpr.SeqBoundedPool(e1, e0.Type, e1.Type.AsSeqType.Arg));
            }                                     //AsSeqType seems to return null ^^
```

MikaelMayer:
Good news, since recent changes, the issue is still there, but the LSP server is no longer stuck. I only displays the error in the diagnostics window, making it possible for the user to identify what is wrong by dissecting.

PR #2080, opened 2022-04-29
fix: Added code to catch resolver exceptions if any.

MikaelMayer:
As a prevention measure for issues such as https://github.com/dafny-lang/dafny/issues/2074
I added some code that catches exceptions coming from the Resolver so that it won't impact user experience in the IDE.
I also added some catching code for Ghost Diagnostics because they were impacted as well.
The result is a non-blocking diagnostic that does not crash the server anymore:
image
Moreover, it returns also all the other resolution errors so that users are not helpless:
image


Source/DafnyLanguageServer/Language/Symbols/DafnyLangSymbolResolver.cs
```diff
@@ -36,6 +36,9 @@ public class DafnyLangSymbolResolver : ISymbolResolver {
           // because it must not be null.
           return new CompilationUnit(program);
         }
+      } catch (Exception e) {
+        program.reporter.Error(MessageSource.Resolver, program.GetFirstTopLevelToken(), $"Please report this internal error:\n{e}");
```
cpitclaudel, line 40:
```suggestion
        program.reporter.Error(MessageSource.Resolver, program.GetFirstTopLevelToken(), $"Dafny encountered an error.  Please report it at <https://github.com/dafny-lang/dafny/issues>:\n{e}");
```

Source/DafnyLanguageServer/Language/GhostStateDiagnosticCollector.cs
```diff
@@ -20,18 +23,25 @@ public class GhostStateDiagnosticCollector : IGhostStateDiagnosticCollector {
...
-      var visitor = new GhostStateSyntaxTreeVisitor(symbolTable.CompilationUnit.Program, cancellationToken);
-      visitor.Visit(symbolTable.CompilationUnit.Program);
-      return visitor.GhostDiagnostics;
+
+      try {
+        var visitor = new GhostStateSyntaxTreeVisitor(symbolTable.CompilationUnit.Program, cancellationToken);
+        visitor.Visit(symbolTable.CompilationUnit.Program);
+        return visitor.GhostDiagnostics;
+      } catch (Exception e) {
+        logger.LogDebug(e, "encountered an exception while getting ghost state diagnostics of {Name}", symbolTable.CompilationUnit.Name);
```
cpitclaudel, line 42: This will not be visible to users, right?  Aren't these logs only written to disk in a somewhat hidden location?

MikaelMayer: Yes they are written next to DafnyLanguageServer.dll, and only if an option is activated. 
But this kind of error can happen if the language server recovered from problematic resolving, and leaving the program in a state that the Ghost Diagnostic Collector is not assuming, so that's why it's less important.

keyboardDrummer:
Can we add tests through using dependency injection to inject a broken implementation?

keyboardDrummer (approved):

Source/DafnyLanguageServer.Test/Unit/DafnyLangSymbolResolverTest.cs
```diff
@@ -0,0 +1,51 @@
...
+      new DefaultModuleDecl(), null) {
+    }
+    public override object Dereference() {
+      return this;
+    }
+  }
+
+  [TestMethod]
+  public void EnsureResilienceAgainstErrors() {
+    // Builtins is null to trigger an error.
```
keyboardDrummer, line 42: It seems like this test relies on some badly defined API of `DafnyPipeline.csproj` to trigger the exception. I think it might be nicer if the test uses a more explicit API to trigger a failure, like mocking something from `DafnyPipeline.csproj` and letting the mock throw an Exception, but I imagine `DafnyPipeline.csproj` is not setup for being mocked.

MikaelMayer: Yes you identified this test correctly.
I don't see either how to mock the DafnyPipeline, so I took the route of the minimum working test for now.

PR #2081, opened 2022-04-29
feat: Various small improvements to Python code generation

fabiomadge:


robin-aws (commented):
Still thinking through the changes necessary to support let expressions, but in the meantime here are a couple of comments.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -39,10 +39,31 @@ def _tail_call():
...
+        self.isStr = isinstance(__iterable, str) or isStr
+        super().__init__(__iterable)
+
     @property
     def Elements(self):
         return self
 
+    @property
+    def UniqueElements(self):
+        return Seq(set(self))
```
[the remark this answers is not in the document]
fabiomadge: It doesn't really matter here, but I think it's the type dafny assumes.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -158,3 +179,12 @@ def euclidian_modulus(a, b):
 @dataclass
 class HaltException(Exception):
     message: str
+
+def quantifier(vals, frall, pred):
```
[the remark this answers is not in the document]
fabiomadge: It's what Java, C#, and JS call it, but I don't deeply care.

Source/DafnyRuntime/DafnyRuntime.py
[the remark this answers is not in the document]
robin-aws: `UniqueElements` is just something that needs to be called in the target language, it doesn't have to produce a "Dafny value" at this level of abstraction. Just `set(self)` will work because it will only be used to compile something like this:

```dafny
var mySet := set x | x in mySeq
```

To something like this in Python:

```python
mySet = _dafny.Set()
for x in mySet.UniqueElements:  // <- produced by PythonCompiler.CreateForEachLoop
  mySet.add(x)
```

What you've got is correct too, since `_dafny.Seq` is also iterable, I'm just trying to save us an extra object creation. :)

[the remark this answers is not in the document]
fabiomadge: It will undoubtedly work for python(see it doesn’t matter _here_) and given that this is only used immediately in comprehensions, like you described, it doesn’t matter in general. That being said, I do like the cleaner interface of returning a sequence, but your overhead argument is compelling, so I’ll adopt it.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -39,10 +39,31 @@ def _tail_call():
     raise TailCall()
 
 class Seq(list):
+    def __init__(self, __iterable = None, isStr = False):
```
robin-aws, line 42: The testing coverage for this is implicitly in `Hello.dfy`, I assume, which only worked before because we were just emitting a base Python string?

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -4863,8 +4881,12 @@ private class GetterSetterLvalueImpl : ILvalue {
         var e = (LambdaExpr)expr;
 
         wr = CaptureFreeVariables(e, false, out var su, inLetExprBody, wr, wStmts);
+        if (su != Substituter.EMPTY && TargetLambdasRestrictedToExpressions) {
+          wStmts = wr.Fork();
+          wr = EmitReturnExpr(wr);
+        }
```
robin-aws, line 4887: Just noting for the record that as I'm better appreciating our solution so far for `TargetLambdasRestrictedToExpressions`, I definitely see more speed bumps down the road for edge cases we don't handle correctly, because we're not doing this "post hoc clean up" for every possible method call that might need it. I won't likely see an immediate alternative and don't want to block this proven progress, but I do want to revisit before we go too much further.

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: Yes, we weren't doing anything with the strings other than passing them around, so it was fine.

Source/Dafny/Compilers/SinglePassCompiler.cs
fabiomadge: Yes, those changes are unlikely to be comprehensive, but I was hesitant to introduce additional untested changes.

Source/DafnyRuntime/DafnyRuntime.py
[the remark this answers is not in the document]
robin-aws: Thanks! It's purely an optimization, and normally I wouldn't prioritize it at this stage, but this is a really easy one. :)

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: Good point, I'll try to add it in this one.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -48,6 +48,8 @@ public class PythonCompiler : SinglePassCompiler {
     const string DafnySeqClass = $"{DafnyRuntimeModule}.Seq";
     const string DafnyMapClass = $"{DafnyRuntimeModule}.Map";
     protected override string StmtTerminator { get => ""; }
+    protected override string True { get => "True"; }
+    protected override string False { get => "False"; }
```
robin-aws, line 52: 🚀

robin-aws (approved):
Approved on the condition that the next PR adds a beefy comment on `TargetLambdasRestrictedToExpressions` about how we're using `wStmts` and why we have to conditionally add return statements in the `SinglePassCompiler` superclass. :)

Source/Dafny/Compilers/Compiler-python.cs
robin-aws: Oh if we're going to do that, we should do it in the verifier instead. :)

Issue #2010, opened 2022-04-12
Address CI capacity issues (queued jobs etc.)


robin-aws:
With the amount of concurrent development on this project, CI jobs have been queueing for hours and PRs are piling up like crazy. Here are some potential ways we can improve this in the approximate order we'll investigate them:

1. Enable executing the Dafny driver multiple times in the same process, so that the xUnit test runner can avoid the overhead of the .NET VM startup on every call to `%dafny`. This will require refactoring several remaining static variables in Dafny and potentially Boogie. My initial experiments suggest this may cut the test suite execution time in half or better, especially for very small test files. It would also enable easily measuring code coverage of the integration tests.
2. Set up a branching model where the most expensive tests do not run on every PR, but instead in an "overnight build" and/or a deep regression test before releasing. It might be reasonable to only run the integration tests on a single platform for each PR (likely ubuntu since the concurrency limits on mac os runners are much lower) and cover other platforms in this separate deep test suite.
3. Investigate finding more CI capacity via upgrading the `dafny-lang` account, hooking up self-hosted runners, or connecting external build services such as AWS CodeBuild. This is a last resort, as depending on funding for these additional resources is a risk for the future of the Dafny project.

robin-aws:
Calling this resolved for now, as I hear @MikaelMayer was able to merge 5 PRs in a day which is a strong indication things are much better now. :) Will reopen with extreme prejudice if we hit more issues!

PR #2089, opened 2022-05-02
chore: Only run "deep tests" in nightly build

robin-aws:
Should address #2010, but will leave that open for a while until we verify it's enough to unblock things.

Initially, this means the large integration test suite is only run for `ubuntu-18.01` on every PR, and "deep tests" refers to running the suite on the current full set of 4 OS runners (`ubuntu-latest`,  `ubuntu-18.01`, `macos-latest` and `windows-latest`). This should work out well since the Ubuntu runners complete the suite in approximately half the time compared to Windows (~15 minutes vs. > 30 minutes), and have four times the concurrent job limit compared to Mac OS.

To back this up with mechanisms, two places in the CI use the GitHub API to check that the deep tests have run successfully:
1. Every PR will check that the last `deep-tests.yml` run on `master` passed. That means if a nightly build fails because of a PR the day before, subsequent PRs will be blocked until the deep suite is fixed (which may just involve rerunning individual jobs if there were transient failures). To merge changes intended to fix those tests themselves, PRs can add the label `skip-deep-tests-check` (since otherwise we're in a catch-22!)
2. The `publish-release.yml` workflow will verify that the last run of `deep-tests.yml` on the tagged commit ran successfully.

Other kinds of expensive and unlikely-to-break tests could be added or moved to this scope in the future. Note that `deep-tests.yml` includes a `workflow_dispatch:` trigger, meaning it can be run manually from here: https://github.com/dafny-lang/dafny/actions/workflows/deep-tests.yml

Also removed several `push` triggers for workflows: no one was paying attention to these builds, and since our branch protection prohibits direct pushes to master these were only happening immediately after merging PRs, and hence were just adding additional CI capacity load for no good reason.


.github/workflows/check-for-workflow-run.js
```diff
@@ -0,0 +1,29 @@
+// Check for a successful previous run of a workflow.
+// This could make a decent reusable action itself.
+module.exports = async ({github, context, core}, workflowID, branch, sha = null) => {
+  // This API doesn't support filtering by SHA, so we just
+  // fetch the first page and scan manually.
+  // That means if the run is fairly old it may be missed,
+  // but that should be rare.
```
atomb, line 7: I slightly worry that this could be fragile, but in the worst case it would be sufficient to re-trigger the deep tests manually, right?

.github/workflows/msbuild.yml
```diff
@@ -51,8 +49,14 @@ jobs:
     - name: Check whitespace and style
       working-directory: dafny
       run: dotnet tool run dotnet-format -w -s error --check Source/Dafny.sln --exclude Dafny/Scanner.cs --exclude Dafny/Parser.cs
-      
+  
+  check-deep-tests:
+    uses: ./.github/workflows/check-deep-tests-reusable.yml
+    if: "!(contains(github.event.pull_request.labels.*.name, 'skip-deep-tests-check'))"
```
atomb, line 55: Nice addition! Documentation changes, for example, would be nice to not put through the whole battery of tests.

atomb (approved):

.github/workflows/check-for-workflow-run.js
robin-aws: Exactly. The default page size for this API is 30 (https://docs.github.com/en/rest/actions/workflow-runs#list-workflow-runs), and if the nightly build hasn't run at all for a month than we have other problems to deal with. :)

.github/workflows/msbuild.yml
robin-aws: Ah, but notice the feature here is to skip checking that the deep tests have run successfully on the nightly build, to deal with the edge case where a previous PR broke only the deep tests and you need another PR to fix them.

I think you're thinking of a label like `skip-integration-tests` that wouldn't run the integration tests at all on a PR. That's a reasonable feature request too and we could follow up with that. I was a little nervous about it because there are some Dafny files hiding in the `docs` folder tree with lit tests that include them.

PR #2090, opened 2022-05-02
chore: Testing whitespace

fabiomadge:
Tests now observe whitespace and reenables rendering html pages from the integration test build logs.


Source/XUnitExtensions/AssertWithDiff.cs
```diff
@@ -7,7 +7,7 @@
 namespace XUnitExtensions {
   public class AssertWithDiff {
     public static void Equal(string expected, string actual) {
-      var diff = InlineDiffBuilder.Instance.BuildDiffModel(expected, actual);
+      var diff = InlineDiffBuilder.Instance.BuildDiffModel(expected, actual, false);
```
robin-aws, line 10: We'll see if the CI confirms this, but I recall needing to ignore whitespace to have this work on all platforms, because otherwise it would complain about `/r/n` not matching `/n`.

fabiomadge: You mean this is one of those instances where we want to test on all platforms?

fabiomadge:
https://github.com/dafny-lang/dafny/blob/master/Test/git-issues/git-issue-1957.dfy

Source/Dafny/Reporting.cs
```diff
@@ -195,7 +195,7 @@ public class ConsoleErrorReporter : BatchErrorReporter {
     public override bool Message(MessageSource source, ErrorLevel level, IToken tok, string msg) {
       if (base.Message(source, level, tok, msg) && ((DafnyOptions.O != null && DafnyOptions.O.PrintTooltips) || level != ErrorLevel.Info)) {
         // Extra indent added to make it easier to distinguish multiline error messages for clients that rely on the CLI
-        msg = msg.Replace(Environment.NewLine, Environment.NewLine + " ");
+        msg = msg.Replace("\n", "\n ");
```
robin-aws, line 198: Nice find!

cpitclaudel: Why was that needed?  I think the expectation here was that messages were going to use the native line ending, so it made sense to replace just that; is there an underlying problem?

cpitclaudel: (Of course this is safe, but I wonder if it's hiding something else)

fabiomadge: The problem was that we are using native line endings almost nowhere, so this just didn't do anything on Windows, leading to inconsistent output between platforms.

PR #2099, opened 2022-05-05
chore: Quick CI refactoring followups

robin-aws:
* Replace `skip-deep-tests-check` label with `run-deep-tests` label on PRs. Rather than skip this check and wait to see if you've fixed master, better to run them directly on the PR first.
* Updated `RELEASE.md` to indicate you need to manually trigger `deep-tests.yml` after merging the update to the version. The automation is strict and looks for a run on exactly the commit being tagged, and there isn't a way to trigger a workflow on a past commit, only the tip of a branch. It's possible to make the commit to download different from the one defining the workflow, but `check-deep-tests` would need to be refactored to understand that.

### DIFF
--- a/.github/workflows/msbuild.yml
+++ b/.github/workflows/msbuild.yml
@@ -52,11 +52,14 @@ jobs:
   
   check-deep-tests:
     uses: ./.github/workflows/check-deep-tests-reusable.yml
-    if: "!(contains(github.event.pull_request.labels.*.name, 'skip-deep-tests-check'))"
+    if: "!(contains(github.event.pull_request.labels.*.name, 'run-deep-tests'))"
  
   integration-tests:
     needs: check-deep-tests
     uses: ./.github/workflows/integration-tests-reusable.yml
     with:
-      all_platforms: false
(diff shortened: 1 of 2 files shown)


.github/workflows/msbuild.yml
```diff
@@ -52,11 +52,14 @@ jobs:
...
     uses: ./.github/workflows/check-deep-tests-reusable.yml
-    if: "!(contains(github.event.pull_request.labels.*.name, 'skip-deep-tests-check'))"
+    if: "!(contains(github.event.pull_request.labels.*.name, 'run-deep-tests'))"
  
   integration-tests:
     needs: check-deep-tests
     uses: ./.github/workflows/integration-tests-reusable.yml
     with:
-      all_platforms: false
+      # By default only run on one platform, but run on all platforms if the PR has the "run-deep-tests"
```
atomb, line 61:
```suggestion
      # By default run only on one platform, but run on all platforms if the PR has the "run-deep-tests"
```

atomb (dismissed):
Looks like a great change! I noticed one typo and one thing I would re-word in the comment.

fabiomadge (approved):

PR #2101, opened 2022-05-05
fix: Treat provided types as opaque

### LINKED ISSUE #2100: Malformed Boogie for export-provides of class that extends trait
In the following example, the intention with the export set of `Library` is that importers will see `Cl` as an opaque type:

``` dafny
module Library {
  export
    provides Cl

  trait Tr { }
  class Cl extends Tr { }
}

module Client {
  import Library

  method Test(cl: Library.Cl) { }
}
```

However, the program causes Dafny to generate malformed Boogie:

```
test.dfy(7,8): Error: use of undeclared function: implements$Library.Tr
1 name resolution errors detected in d_Client.bpl

d_Client.bpl(2888,6): Error: use of undeclared function: implements$Library.Tr
1 name resolution errors detected in d_Client.bpl
```

The problem is that the generated Boogie contains the declaration

``` boogie
axiom implements$Library.Tr(Tclass.Library.Cl?());
```

which should only be generated in scopes where `Cl` is known to be a class type. Also, it seems suspicious that the symbol `Tclass.Library.Cl?` is used (and that there are no other complaints from Boogie about it). That symbol should not be generated, either, unless `Cl` is known to be a class type.

RustanLeino:
Don’t generate functions/axioms about types that are opaque in the scope.

Fixes #2100

### DIFF
--- a/Source/Dafny/Verifier/Translator.ClassMembers.cs
+++ b/Source/Dafny/Verifier/Translator.ClassMembers.cs
@@ -10,30 +10,32 @@ using PODesc = Microsoft.Dafny.ProofObligationDescription;
 
 namespace Microsoft.Dafny {
   public partial class Translator {
-    void AddClassMembers(TopLevelDeclWithMembers c, bool includeAllMethods) {
+    void AddClassMembers(TopLevelDeclWithMembers c, bool includeAllMethods, bool includeInformationAboutType) {
       Contract.Requires(sink != null && predef != null);
       Contract.Requires(c != null);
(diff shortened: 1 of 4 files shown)


cpitclaudel (approved):
LGTM.  This might be a good opportunity to split that method and rename it, since the part that is now conditional does not seem to be about class members; maybe something like this:

```
void AddTypeInfoAndClassMembers(includeAllMethods) {
   AddTypeInfo()
   AddClassMembers(includeAllMethods)
}
```

PR #2105, opened 2022-05-06
chore: Enable `run-deep-tests`

fabiomadge:
The problem is, that not running `check-deep-tests`, as we do in case of  `run-deep-tests`, keeps us from running the `integration-tests` at all.


atomb (approved):
I _think_ this should be the right thing. It's probably worth trying, and fixing if it's not. :)

PR #2116, opened 2022-05-09
Sort output of `/verificationLogger:text`

atomb:
The order of elements in the lists that `/verificationLogger:text` iterates over can vary when using multiple cores. This PR sorts those lists before printing their contents to ensure deterministic output.

### DIFF
--- a/Source/DafnyDriver/TextLogger.cs
+++ b/Source/DafnyDriver/TextLogger.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using Microsoft.Boogie;
 
 namespace Microsoft.Dafny;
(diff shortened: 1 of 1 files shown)


Source/DafnyDriver/TextLogger.cs
```diff
@@ -13,13 +14,14 @@ public class TextLogger {
   }
 
   public void LogResults(List<(Implementation, VerificationResult)> verificationResults) {
-    foreach (var (implementation, result) in verificationResults) {
+    var orderedResults = verificationResults.OrderBy(vr => vr.Item1.Name);
```
keyboardDrummer, line 17: I would inline this variable

Source/DafnyDriver/TextLogger.cs
```diff
@@ -13,13 +14,14 @@ public class TextLogger {
   }
 
   public void LogResults(List<(Implementation, VerificationResult)> verificationResults) {
```
keyboardDrummer, line 16: `VerificationResult` also contains a (private) `Implementation` field, so consider exposing that so you can replace `(Implementation, VerificationResult)` with `VerificationResult`

keyboardDrummer (approved):

Source/DafnyDriver/TextLogger.cs
atomb: It'll require a Boogie change, though, so I think perhaps it's a refactor best left to a later PR.

Issue #2033, opened 2022-04-19
`xlit` CLI for easily running lit-style integration tests from the command line


robin-aws:
Many contributors are still using commands like `lit -v SomeTestFile.dfy` tool because it's much more convenient than the equivalent `dotnet test Source/IntegrationTests --filter DisplayName~SomeTestFile.dfy` command for using xUnit.

Issue #1770, opened 2022-01-28
Track which compilers support which Dafny features


robin-aws:

robin-aws:
Depends on #2033, as I'd much rather not have to implement this for `lit` as well

robin-aws:
@parno - I'm working on a solution to this now, and in particular I'd like to modify the warnings/errors the C++ compiler produces to make it easier for the testing framework to recognize known limitations of individual compilers.

I was a little surprised that in some cases the compiler will output only a warning rather than an error, and allow the target code to be emitted and possibly compiled when it is guaranteed not to be valid. For example: https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Compilers/Compiler-cpp.cs#L976 

Is this to support some workflow that somehow corrects for this post-hoc? Or is it okay for me to change these cases to hard errors instead? 

I'll be adding a mechanism for a compiler to cleanly reject a valid Dafny program because it doesn't support a given feature (the lack of which could be why the C++ compiler does what it does now, since the alternative is throwing an exception and dumping a stack trace).

parno:
> Is this to support some workflow that somehow corrects for this post-hoc? 

We found it more convenient to get a complete list of the places you needed to fix (by converting to a `u64` or making the `int` a ghost) all in one go, rather than fixing one, re-running the compiler, fixing the next, etc.  This process is sound in so far as the Dafny C++ "runtime" doesn't define the `BigNumber` type, so the C++ compiler will complain and fail.

> Or is it okay for me to change these cases to hard errors instead?

If it's more convenient for testing purposes, I think it's probably fine to convert to a hard error.

robin-aws:
Thanks @parno! To be more precise, I'll make it so these unsupported features are reported via the `Error` method, which just means the output text will change from "WARNING:" to "Error:", and the overall compilation will fail before it attempts to compile the target program (since the error count is checked here: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyDriver/DafnyDriver.cs#L597). 

This will be more consistent with other unsupported feature cases in other compilers (and even in the C++ compiler itself! e.g. https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Compilers/Compiler-cpp.cs#L942), and as you say the target program is guaranteed not to compile anyway.

parno:
Ah, got it.  That sounds great!

robin-aws:
[...]

PR #2121, opened 2022-05-09
chore: Return the appropriate status code when testing

fabiomadge:
This change enables us to detect crashes more reliably. To support tests that require the previous default, there is now `dafny_0`


fabiomadge:
I probably didn't apply the label early enough.

robin-aws:
This is awesome, and now I'm curious to see the script you used to apply the changes. :)

My only ask is that we add `%dafny_0` to `lit.site.cfg` as well, since that should be quick and I know some contributors are still using `lit` instead of the xUnit-based runner (until we deliver https://github.com/dafny-lang/dafny/issues/2033)

> I probably didn't apply the label early enough.

If having to push a dummy commit to kick the CI again in this situation becomes a recurring irritation, I believe we can set things up so that a workflow triggers when a label is applied to a PR.

fabiomadge:
I did add it to both xUnit and lit, but it's hard to miss among the literal hundreds of other files.

```Python
FILES = ["wishlist/FunctionEquality.dfy',"]

def replace(s, a, b, i):
    if i <= 0:
        return s
    start = -len(a)
    while i > 0:
        start = s.find(a, start+len(a))
        if start == -1:
            return s
        i -= 1
    return s[:start] + b + s[start+len(a):]


for relative in FILES:
    full = f'/Users/fmadge/Documents/dafny/Test/{relative}'
    f = open(full, 'r')
    before = f.read()
    f.close()
    f = open(full, 'w')
    current = before
    def check():
        f.seek(0)
        f.write(current)
        f.flush()
        return os.system(f'lit {full} &> /dev/null') == 0
    i = 0
    while not check():
        i += 1
        current = replace(before, '%dafny', '%dafny_0', i)
        if current == before:
            print(f'failed fixing {relative} and restored')
            f.close()
            f = open(full, 'w')
            f.write(before)
            break
    if not current == before:
        print(f'fixed {relative}')
    f.close()
```

cpitclaudel:
Our docs say that `/countVerificationErrors` is deprecated; instead of that, maybe we could use use `|| true` on each line that we expect to exist with a non-zero exit code?

fabiomadge:
The burden of keeping this option around doesn't seems too high, and luckily we control our own destiny in that regard.

robin-aws:
> Our docs say that `/countVerificationErrors` is deprecated; instead of that, maybe we could use use `|| true` on each line that we expect to exist with a non-zero exit code?

This is a good point: `|| true` already works with the xUnit runner (https://github.com/dafny-lang/dafny/blob/master/Test/DafnyTests/TestAttribute.dfy#L2), and is more self-documenting than `%dafny_0`. Assuming you can easily apply that suffix with your script instead, I'd prefer that simpler solution!

> I did add it to both xUnit and lit, but it's easy to miss among the hundreds of other files.

Whoops, clearly it is. :)

fabiomadge:
The script isn't prefect, so it would be easier to expand `dafny_0` in the proposed way. I see value in less complex expressions in the test files.

robin-aws (approved):
Although I have plans for a different, deeper solution to this as part of https://github.com/dafny-lang/dafny/issues/1770, this is still great forwards progress and will save time in converting test cases in that work. Thanks for taking this on!

PR #2124, opened 2022-05-10
Introduce an async type DynamicStack that can be used to prevent stac…

keyboardDrummer:
Introduce an async type DynamicStack that can be used to prevent stack overflows in recursive code

### DIFF
--- a/Source/DafnyLanguageServer.Test/Various/DynamicStack.cs
+++ b/Source/DafnyLanguageServer.Test/Various/DynamicStack.cs
@@ -1,0 +1,122 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Dafny;
+
+[AsyncMethodBuilder(typeof(DynamicStackBuilder))]
+class DynamicStack {
+  public void Run() {
+    DynamicStackBuilder.Builder.Value!.Run();
+  }
+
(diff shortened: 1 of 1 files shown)


robin-aws:
Based on a quick skim, this is a cool idea with promise. I naturally worry about the overhead of using await/async for what is actually a sequential algorithm, but I could easily be convinced with benchmarking data that C#/CLR is clever enough about optimizing that case (and this kind of processing is not where I'd worry about minor inefficiencies).

Another alternative is to use the visitor pattern more. A lot of AST processing is using that, even though there is surely room to improve the UX. Can we experiment with how that would look here? (perhaps on a smaller subset of the PrintExpression code as that's a lot of lines to change :)). On the flip side, if this approach works well it might be worth migrating existing visitor-based code to it eventually instead.

cpitclaudel:
> Another alternative is to use the visitor pattern more. 

Does the visitor pattern help with recursion depth?

keyboardDrummer:
> I naturally worry about the overhead of using await/async for what is actually a sequential algorithm

It's still executed sequentially in this PR. There are no tasks or different threads involved. For every method with `async DynamicStack`, its code is split up into parts that are separated by the await operator. A class is generated for the method, which I'll call the 'execution class', that contains a `MoveNext()` operator which executes the method part that the class is currently at and then increments and integer to move to the next part.

When an `async DynamicStack` method is called, an instance of its execution class is created and that instance is placed on the thread local execution stack. When the `await` operator is used inside a `async DynamicStack`, execution halts and the current method's execution class is placed on the execution stack.

Calling `Run()` on a dynamic stack executes everything on the execution stack until it's empty.

PR #2126, opened 2022-05-10
chore: Fixed a concurrency issue

MikaelMayer:
I recently got a CI issue in this PR:
https://github.com/dafny-lang/dafny/runs/6355961286?check_suite_focus=true

The error was the following
>Unhandled exception. System.AggregateException: One or more errors occurred. (An item with the same key has already been added. Key: /home/runner/work/dafny/dafny/dafny/Source/IntegrationTests/bin/Debug/net6.0/TestFiles/LitTests/LitTest/dafny0/ShowSnippets.dfy)
 ---> System.ArgumentException: An item with the same key has already been added. Key: /home/runner/work/dafny/dafny/dafny/Source/IntegrationTests/bin/Debug/net6.0/TestFiles/LitTests/LitTest/dafny0/ShowSnippets.dfy
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Microsoft.Dafny.DafnyConsolePrinter.GetFileLine(String filename, Int32 lineIndex)

I saw that there could be a race condition in the way we test if a key exist, then we compute a value, and then we add it, if done from two different threads.

This PR fixes it. I don't know how to create a test in the general case.

### DIFF
--- a/Source/Dafny/DafnyConsolePrinter.cs
+++ b/Source/Dafny/DafnyConsolePrinter.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


MikaelMayer:
> It's unfortunately not clear to me that this is a concurrency issue

Oh, sorry if it was not clear. In the previous version, there were two calls to the dictionary, one to test if it contained a key, and another one to add a value to this key if it was not present.
However, `fsCache.Add(filename, lines);` crashes if a key was added for the second time, and an interleaving of threads could make it so that both threads enter the "then" section because the key is not contained, and they would both try to add the value to the dictionary.

fabiomadge:
Thanks for the explanation, but I guess I wasn’t very clear in my comment. The race condition is obvious, but I‘m not 100% certain that fixing it will make the bug go away.

MikaelMayer:
After reading the doc of `GetOrAdd`, it seems that there could still be a race issue of multiple threads executing the computation code at the same time, but only the first returned value will actually be inserted and all threads will have then the same value.

However, the bug I mentioned was that ".Add" on a dictionary was failing, and there is no such exception for `GetOrAdd`, only if key is null or it overflows, but no more duplicate key exception.

PR #2129, opened 2022-05-10
Release Dafny 3.6.0

cpitclaudel:
Using this PR to run CI; will need another commit to fix release notes on master.

### DIFF
--- a/.github/workflows/check-deep-tests-reusable.yml
+++ b/.github/workflows/check-deep-tests-reusable.yml
@@ -5,6 +5,8 @@ on:
     inputs:
       sha:
         type: string
+      branch:
+        type: string
 
 jobs:
   check-deep-tests:
@@ -19,4 +21,7 @@ jobs:
       with:
         script: |
           const script = require('${{ github.workspace }}/dafny/.github/workflows/check-for-workflow-run.js')
-          console.log(script({github, context, core}, workflowID = 'deep-tests.yml', branch = 'master', sha = '${{ inputs.sha }}'))
+          console.log(script({github, context, core,
+                              workflow_id: 'deep-tests.yml',
+                              ...('${{ inputs.sha }}' ? {sha: '${{ inputs.sha }}'} : {}),
+                              ...('${{ inputs.branch }}' ? {branch: '${{ inputs.branch }}'} : {})}))
--- a/.github/workflows/check-for-workflow-run.js
+++ b/.github/workflows/check-for-workflow-run.js
@@ -1,6 +1,6 @@
 // Check for a successful previous run of a workflow.
 // This could make a decent reusable action itself.
-module.exports = async ({github, context, core}, workflowID, branch, sha = null) => {
+module.exports = async ({github, context, core, workflow_id, sha, ...config}) => {
   // This API doesn't support filtering by SHA, so we just
   // fetch the first page and scan manually.
   // That means if the run is fairly old it may be missed,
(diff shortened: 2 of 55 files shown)


cpitclaudel:
Deeps test run is at https://github.com/dafny-lang/dafny/actions/runs/2303523148

robin-aws (dismissed):

Source/version.cs
```diff
@@ -1,4 +1,4 @@
 using System.Reflection;
-// Version 3.5.0, year 2018+4 month 3 day 14 
-[assembly: AssemblyVersion("3.5.0.40314")]
-[assembly: AssemblyFileVersion("3.5.0.40314")]
+// Version 3.6.0, year 2018+4 month 5 day 10
```
robin-aws, line 2: Just a heads up that we'll need to bump this up a day, since we won't be able to actually tag and release until the 11th :)

.github/workflows/check-deep-tests-reusable.yml
```diff
@@ -19,4 +21,7 @@ jobs:
       with:
         script: |
           const script = require('${{ github.workspace }}/dafny/.github/workflows/check-for-workflow-run.js')
-          console.log(script({github, context, core}, workflowID = 'deep-tests.yml', branch = 'master', sha = '${{ inputs.sha }}'))
+          console.log(script({github, context, core},
+                             {workflow_id: 'deep-tests.yml',
+                              sha: '${{ inputs.sha }}' || null,
```
robin-aws, line 26: The `|| null` seems redundant to me, since `''` is equally falsey.

.github/workflows/check-for-workflow-run.js
```diff
@@ -1,19 +1,19 @@
 // Check for a successful previous run of a workflow.
 // This could make a decent reusable action itself.
-module.exports = async ({github, context, core}, workflowID, branch, sha = null) => {
+module.exports = async ({github, context, core}, {workflow_id, sha, ...config}) => {
```
robin-aws, line 3:
```suggestion
module.exports = async ({github, context, core, workflow_id, sha, ...config}) => {
```

Might as well pass everything in one object parameter rather than two, right?

.github/workflows/check-deep-tests-reusable.yml
cpitclaudel: We don't know whether the library treats them the same, do we?

cpitclaudel: Solved by removing the keys from the object completely when they are not given.

Source/version.cs
cpitclaudel: Done

.github/workflows/check-for-workflow-run.js
cpitclaudel: Good point!  Amended existing commit.

.github/workflows/check-deep-tests-reusable.yml
```diff
@@ -19,4 +21,7 @@ jobs:
       with:
         script: |
           const script = require('${{ github.workspace }}/dafny/.github/workflows/check-for-workflow-run.js')
-          console.log(script({github, context, core}, workflowID = 'deep-tests.yml', branch = 'master', sha = '${{ inputs.sha }}'))
+          console.log(script({github, context, core,
+                              workflow_id: 'deep-tests.yml',
+                              ...('${{ inputs.sha }}' ? {sha: '${{ inputs.sha }}'} : {}),
+                              ...('${{ inputs.branch }}' ? {branch: '${{ inputs.branch }}'} : {})}))
```
robin-aws, line 27: Cute! I suspect this isn't particularly idiomatic Javascript but not worth bikeshedding over. :)

cpitclaudel: ? It's completely standard, AFAIK.
You might argue that `('${{ inputs.branch }}' && {branch: '${{ inputs.branch }}'})` also complies with ES6, but it relies on corner cases.

robin-aws: I meant it would be more idiomatic to pass the parameters unconditionally and check for empty strings on the callee side as needed.

robin-aws (approved):

Issue #413, opened 2019-11-05
Revisit definition of Dafny strings (RFC #12)


robin-aws:
(was: "Explicitly document that Dafny strings are UTF-16")

I found this just a bit too subtle when reading the reference manual. There are a couple references to "Unicode letters" and "Unicode characters" that should really be "Unicode 16-bit code units". The grammar does spell out that a Unicode character is four hex digits, but that's a little bit buried.

Even better, I would love it if Dafny characters were actually sequences of Unicode code points, so that the built-in sequence operations could be used to perform robust per-character string manipulation.

robin-aws:
Going further than this, even if the canonical representation of strings in Dafny is using UTF16, another issue with just making `string` an alias for `seq<char>` means that it allows invalid UTF16 sequences (those that use surrogate code points incorrectly for example).

Given that Unicode is so often misunderstood and often the source of subtle bugs, I think there's a great opportunity to have a verifiably-correct implementation of Unicode in Dafny. Languages like Java and C# will often allow invalid sequences and can be lax with encoding/decoding, but I think it would be much better to be strict in Dafny.

robin-aws:
Making `string` an alias for `seq<char>` also has implications for interacting with external code: it is necessary to convert from the native representation in the target language to the runtime implementation of Dafny sequences. This can have performance consequences and also complicates the "foreign function interface": because sequences are value types in Dafny you can't refer to a possibly-null string value, which happens all the time in external code.

robin-aws:
My current plan here is to keep `string` as an alias for `seq<char>`, but to redefine `char` to mean "Unicode scalar value" ala Rust, and to therefore need 32 bits instead of 16. This means the direct value of a string is the "true" Unicode sequence. Within each target runtime, it should be possible to implement this type by wrapping the native representation of strings (as Java already does with the `StringDafnySequence` class), therefore allowing more efficient encoding of these values. This will help with both performance and the FFI - yes, doing per-character operations will be slow because the native wrappers will not be random access w.r.t. the actual Unicode scalar values, but it should be possible to optimize all the operations Unicode-compliant code will actually need to do.

indo-dev-0:
Hi!  I'm modelling bits of Unicode in Dafny.

[...]

PR #2130, opened 2022-05-10
feat: Use (wrapped) native strings instead of sequences of characters in Python

fabiomadge:


robin-aws (changes requested):
Watch out here, this is almost certainly incorrect - Python models Unicode correctly but Dafny doesn't, so you'll get inconsistent behavior. See https://github.com/dafny-lang/dafny/issues/413 (which I'm intending to fix in the near future, so better to wait on this until then)

PR #2132, opened 2022-05-11
fix: Fixed null warnings.

MikaelMayer:
Fixes recent CI warnings.

### DIFF
--- a/Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
+++ b/Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
@@ -79,7 +79,7 @@ namespace Microsoft.Dafny.LanguageServer.Workspace.ChangeProcessors {
         return originalDiagnostics.SelectMany(diagnostic =>
           MigrateDiagnostic(changeEndOffset, diagnostic)).ToList();
       }
-      private List<Position> MigratePositions(List<Position> originalRanges, (TextDocumentContentChangeEvent change, Position position) changeEndOffset) {
+      private List<Position> MigratePositions(List<Position> originalRanges, (TextDocumentContentChangeEvent change, Position? position) changeEndOffset) {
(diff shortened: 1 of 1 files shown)


Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -88,12 +88,12 @@ CancellationToken cancellationToken
...
 
       // Requires changeEndOffset.change.Range to be not null
-      private IEnumerable<Position> MigratePosition((TextDocumentContentChangeEvent change, Position afterChangeEndOffset) changeEndOffset, Position position) {
+      private IEnumerable<Position> MigratePosition((TextDocumentContentChangeEvent change, Position? afterChangeEndOffset) changeEndOffset, Position position) {
         if (changeEndOffset.change.Range!.Contains(position)) {
           return new List<Position> { };
         }
 
-        return new List<Position> { MigratePosition(position, changeEndOffset.change.Range, changeEndOffset.afterChangeEndOffset) };
+        return new List<Position> { MigratePosition(position, changeEndOffset.change.Range, changeEndOffset.afterChangeEndOffset!) };
```
cpitclaudel, line 96: Looks like a dangerous API to me: you have a function that claims to accept null but then assumes it's never passed null

cpitclaudel: Is that safe?

cpitclaudel (approved):

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
MikaelMayer: Unfortunately for the eye, yes, it is safe, because I won't call this function if changeEndOffset.change.Range is null, which means that afterChangeEndOffset cannot be null either because of the invariant I'm maintaining.

PR #2136, opened 2022-05-12
Do not republish diagnostics if they have not changed

keyboardDrummer:
### Changes

- Do not republish diagnostics if they have not changed to reduce load on the client and to make unit testing easier. This should resolve the instability in `OpeningDocumentWithMultipleVerificationCoresReturnsStableDiagnostics`
- Several renames

### Testing
- Updated the tests to reflect the new diagnostic behavior.


Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -53,11 +64,17 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
...
     private void PublishGhostDiagnostics(DafnyDocument document) {
-      languageServer.TextDocument.SendNotification(new GhostDiagnosticsParams {
+
+      var newParams = new GhostDiagnosticsParams {
         Uri = document.Uri,
         Version = document.Version,
         Diagnostics = document.GhostDiagnostics.ToArray(),
-      });
+      };
+      if (previouslyPublishedGhostDiagnostics.TryGetValue(document.Uri, out var previousParams) && previousParams.Diagnostics.Equals(newParams.Diagnostics)) {
```
cpitclaudel, line 73: Is tehre a `GetValueOrDefault` for concurrent dictionaries?

```suggestion
      if (previouslyPublishedGhostDiagnostics.GetValueOrDefault(document.Uri) == newParams.Diagnostics) {
```

Probably not…

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -53,11 +64,17 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
...
+        return;
+      }
+      previouslyPublishedGhostDiagnostics.AddOrUpdate(document.Uri, _ => newParams, (_, _) => newParams);
```
cpitclaudel, line 76:
```suggestion
      previouslyPublishedGhostDiagnostics[document.Uri] = newParams;
```

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -29,6 +35,11 @@ public class DiagnosticPublisher : IDiagnosticPublisher {
         Version = document.Version,
         Diagnostics = document.Diagnostics.ToArray(),
       };
+      if (previouslyPublishedDiagnostics.TryGetValue(document.Uri, out var previousParams) && previousParams.Diagnostics.Equals(diagnosticParameters.Diagnostics)) {
+        return;
+      }
+
+      previouslyPublishedDiagnostics.AddOrUpdate(document.Uri, _ => diagnosticParameters, (_, _) => diagnosticParameters);
```
cpitclaudel, line 42: Could we share this code with the code below?

cpitclaudel (commented):
Neat, thanks!

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
keyboardDrummer: That doesn't exist

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
keyboardDrummer: I don't want to. LSP diagnostics and Ghostness hints conceptually have little in common. The ghostness hints happen to reuse LSP's `Diagnostic` type but they only partially fill it, not sure why they don't send just a `Range` instead.

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
keyboardDrummer: Nice

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
cpitclaudel: I meant just the part that looks up in the dictionary and sends a message if there's no match; would it be bad to share that?

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
```diff
@@ -70,10 +70,10 @@ public class DiagnosticPublisher : IDiagnosticPublisher { // TODO rename to Noti
         Version = document.Version,
         Diagnostics = document.GhostDiagnostics.ToArray(),
       };
-      if (previouslyPublishedGhostDiagnostics.TryGetValue(document.Uri, out var previousParams) && previousParams.Diagnostics.Equals(newParams.Diagnostics)) {
+      if (previouslyPublishedGhostDiagnostics.TryGetValue(document.Uri, out var previousParams) && previousParams.Diagnostics.SequenceEqual(newParams.Diagnostics)) {
```
cpitclaudel, line 73: Woops, good catch.

cpitclaudel (dismissed):

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
keyboardDrummer: Well this wasn't needed for some reason, the tests already showed it was working, but this seems safer.

Source/DafnyLanguageServer/Workspace/DiagnosticPublisher.cs
keyboardDrummer: I'll have a look in the next PR (that also changes this file and publishes another cached thing)

MikaelMayer (commented):
It looks like there is one more instability that is unrelated to this change and that this PR just hit. Can you have a look?

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
```diff
@@ -47,9 +47,9 @@ decreases y
 }".TrimStart();
       var documentItem = CreateTestDocument(source);
       await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
-      var diagnostics = await GetLastVerificationDiagnostics(documentItem, CancellationToken);
+      var diagnostics = await GetLastDiagnostics(documentItem, CancellationToken);
       Assert.AreEqual(0, diagnostics.Length);
-      await AssertNoDiagnosticsAreComing();
+      await AssertNoDiagnosticsAreComing(CancellationToken);
```
MikaelMayer, line 52: Isn't the cancellation token a protected property of the parent `LanguageServerTestBase`? Couldn't you use it inside `AssertNoDiagnosticsAreComing` instead of providing it from the child?

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
```diff
@@ -20,16 +20,13 @@ public class ClientBasedLanguageServerTest : DafnyLanguageServerTestBase {
...
-    var document = await Documents.GetVerifiedDocumentAsync(documentItem);
-    await diagnosticsReceiver.AwaitNextDiagnosticsAsync(cancellationToken); // Get resolution diagnostics
-    var remainingDiagnostics = expectedNumber ?? Int32.MaxValue;
+    var document = await Documents.GetLastDocumentAsync(documentItem);
     Diagnostic[] result;
     do {
       result = await diagnosticsReceiver.AwaitNextDiagnosticsAsync(cancellationToken);
-      remainingDiagnostics--;
-    } while (!document!.Diagnostics.SequenceEqual(result) && remainingDiagnostics > 0);
+    } while (!document!.Diagnostics.SequenceEqual(result));
```
MikaelMayer, line 29: Why do you need this? It's a bit smelly to access the diagnostics of the document and compare them against what was received, because it could or could not have been updated in the meanwhile.

MikaelMayer (approved):
I'm worried about the way you extract diagnostics, but otherwise, it should already solve some instabilities.

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
keyboardDrummer: We're comparing the diagnostics of the final document, which won't be updated anymore, against the diagnostics of the document.

MikaelMayer: Very good insight, so as long as we call this method after every change, and call it until there are no more diagnostics, we should be fine.

PR #2140, opened 2022-05-16
feat: Add basic `{:extern}` support to Python

fabiomadge:


Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -1340,10 +1349,50 @@ protected class ClassWriter : IClassWriter {
       TrStmt(recoveryBody, exceptBlock);
     }
 
+    private static readonly Regex ModuleLine = new(@"^\s*assert\s+""([a-zA-Z0-9_]+)""\s*==\s*__name__\s*$");
```
robin-aws, line 1352: Can we eliminate this trick if we require that the original file already has the right name? Java already doesn't let you name the external file `ExternN.java`.

robin-aws (changes requested):

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: I think so, but I think it's convenient. It's also how it's done in go.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -688,8 +689,16 @@ protected class ClassWriter : IClassWriter {
       throw new NotImplementedException();
     }
 
-    protected override void EmitNew(Type type, IToken tok, CallStmt _, ConcreteSyntaxTree wr, ConcreteSyntaxTree wStmts) {
-      wr.Write($"{TypeName(type, wr, tok)}()");
+    protected override void EmitNew(Type type, IToken tok, CallStmt initCall, ConcreteSyntaxTree wr, ConcreteSyntaxTree wStmts) {
+      var ctor = (Constructor)initCall?.Method;  // correctness of cast follows from precondition of "EmitNew"
+      var arguments = Enumerable.Empty<string>();
+      if (ctor != null && ctor.IsExtern(out _, out _)) {
+        // the arguments of any external constructor are placed here
```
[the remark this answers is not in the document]
fabiomadge: Yes, it does, usually we don't use native constructors, but an init call, or something.

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: The assert line was already there. The propose is to ensure the next line `module_ = sys.modules[__name__]` is actually doing the right thing. This was originally added so we could always use the full name. This could actually be in the filename, as we always flatten module hierarchies. (Imposed by the SinglePassCompiler)

Test/comp/Extern5.py
```diff
@@ -0,0 +1,73 @@
+"""
+The Python compiler only supports {:extern} code on a module level, so the
+entire module must be supplied.
+"""
+
+import sys, _dafny
+
+assert "Library" == __name__
+Library = sys.modules[__name__]
```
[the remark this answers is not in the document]
robin-aws: I also see now that we generate this line in Python files created by the compiler, so it's consistent at least.

robin-aws (approved):
LGTM! This suggestion is optional: https://github.com/dafny-lang/dafny/pull/2140#discussion_r898194996

robin-aws (approved):

PR #2144, opened 2022-05-16
fix: Redirect Console.Write to the logger in LSP

MikaelMayer:
This PR immediately redirects all further calls to `Console.Write` to the logger, so that the language server won't crash anymore just because some code uses `Console.Write`.

### DIFF
--- a/Source/DafnyLanguageServer/Program.cs
+++ b/Source/DafnyLanguageServer/Program.cs
@@ -3,11 +3,11 @@ using Microsoft.Extensions.Logging;
 using OmniSharp.Extensions.LanguageServer.Server;
 using Serilog;
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Reflection;
-using System.Threading;
+using System.Text;
 using System.Threading.Tasks;
+using Serilog.Events;
 using OmniSharpLanguageServer = OmniSharp.Extensions.LanguageServer.Server.LanguageServer;
 
 namespace Microsoft.Dafny.LanguageServer {
@@ -27,6 +27,8 @@ namespace Microsoft.Dafny.LanguageServer {
             // ReSharper disable once AccessToModifiedClosure
             .WithDafnyLanguageServer(configuration, () => shutdownServer!())
         );
+        // Prevent any other parts of the language server to actually write to standard output.
+        Console.SetOut(new LogWriter());
         shutdownServer = () => server.ForcefulShutdown();
         await server.WaitForExit;
       }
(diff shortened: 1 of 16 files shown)


Source/DafnyLanguageServer/Program.cs
```diff
@@ -27,6 +27,8 @@ public class Program {
             // ReSharper disable once AccessToModifiedClosure
             .WithDafnyLanguageServer(configuration, () => shutdownServer!())
         );
+        // Prevent any other parts of the language server to actually write to standard output.
+        Console.SetOut(new LogWriter());
         shutdownServer = () => server.ForcefulShutdown();
         await server.WaitForExit;
```
camrein, line 33: The `TextWriter` is disposable, so make sure all resources are freed.
```suggestion
        // Prevent any other parts of the language server to actually write to standard output.
        await using var logWriter = new LogWriter();
        Console.SetOut(logWriter);
        shutdownServer = () => server.ForcefulShutdown();
        await server.WaitForExit;
```

Source/DafnyLanguageServer/Program.cs
```diff
@@ -60,4 +62,15 @@ public class Program {
         .AddSerilog(Log.Logger);
     }
   }
+  internal class LogWriter : TextWriter {
```
camrein, line 65: Since `LogWriter` is only used by the `Program` type, I suggest making it an inner `private` class.

Source/DafnyLanguageServer/Program.cs
```diff
@@ -60,4 +62,15 @@ public class Program {
...
+    public override void Write(char value) {
+      Log.Write(LogEventLevel.Verbose, value.ToString());
+    }
+
+    public override void Write(string? value) {
+      Log.Write(LogEventLevel.Verbose, value);
+    }
```
camrein, line 72: Two suggestions:
- Use `Log.Logger` with the suitable method for the desired log-level.
- Always use a format string to not accidentally provide format-patterns to the logging framework which can cause errors (or even worse). Furthermore, it makes it easier to scan the logs for this kind of log messages.


```suggestion
    public override void Write(char value) {
      Log.Logger.Verbose("unexpected console output: {value}", value);
    }

    public override void Write(string? value) {
      Log.Logger.Verbose("unexpected console output: {value}", value);
    }
```

PS: You may want to test my suggestions, I'm not entirely sure that I got it right.

camrein (changes requested):
That's exciting!

MikaelMayer:
The code portion I wrote on `Program.cs` is XUnit-untested because all the tests rely on a custom creation of the language server, which would not suffer the shortcoming because it's defining its own pipes.

However, I did used the compiled PluginsTest.cs as a plugin for VSCode to actually write to Console.Out during resolution. Without this PR, it fails silently (no error message). With this PR, it succeeds.

Hence, I now switch this PR state to "ready for review" because I wouldn't know how to test it further, unless you have some advice.

camrein (approved):
Fantastic! Let's hope this resolves the issue, but I'm positive that your intuition for the root-cause is right.

keyboardDrummer:
Nice!

PR #2154, opened 2022-05-20
Change default for errorTrace option so it's off

keyboardDrummer:
Error traces can be a nuisance to customers so it's better to turn them off by default.

### DIFF
--- a/Source/Dafny/DafnyOptions.cs
+++ b/Source/Dafny/DafnyOptions.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Dafny {
 
     public DafnyOptions()
       : base("Dafny", "Dafny program verifier", new DafnyConsolePrinter()) {
+      ErrorTrace = 0;
       Prune = true;
       NormalizeNames = true;
       EmitDebugInformation = false;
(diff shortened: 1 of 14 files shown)


cpitclaudel:
Great work, the output looks much less cluttered this way.  Do we have a ticket to try to use those traces for `function method`s, though?  The line numbers are wrong, but the structure is (usually) right, so there is some useful information in there that we're not exposing.

keyboardDrummer:
> Do we have a ticket to try to use those traces for function methods, though? The line numbers are wrong, but the structure is (usually) right, so there is some useful information in there that we're not exposing.

Why do you mention function methods in particular?

We do have plans for visualising counterexamples which would include the error trace information and more, wouldn't it?

cpitclaudel:
> Why do you mention function methods in particular?

For methods, we already show the error on the erroneous branch, but not for `function method`s, so that trace is the only easy way to get the info without using counterexamples.

keyboardDrummer:
> For methods, we already show the error on the erroneous branch

Only when that branch is returned from, right?

cpitclaudel:
>  Only when that branch is returned from, right?

Right.  So it helps for `if … { return x }` but not `var y; if … { y := x }; …; return y`.  For `function method`s though the most common is to have a single `if` or `switch` at the top level, IME, so the latter for appears less commonly .

PR #2159, opened 2022-05-23
chore: Clean up root directory

cpitclaudel:
- Remove unused vendored version of Coco (keep only frame files)
- Remove unused Vim and Emacs utils (now maintained out of tree)
- Move release instructions to `docs/dev`
- Move `interact.py` to `Scripts/`
- Move LaTeX style to `docs/`
- Remove `refman-release` Makefile target


Util/Emacs/README
```diff
@@ -1,2 +0,0 @@
-Emacs support for dafny is provided by the boogie-friends package, available from MELPA.
```
MikaelMayer, line 1: Why did you delete this file? It seems like a good pointer. Is there another pointer to Emacs somewhere visible?

Makefile
```diff
@@ -15,9 +15,6 @@ ${DIR}/boogie/Binaries/Boogie.exe:
 refman:
 	make -C ${DIR}/docs/DafnyRef
 
-refman-release:
```
MikaelMayer, line 18: Why are you confident about deleting this?

MikaelMayer (changes requested):
I like the idea of cleaning up the root directory, but I need some explanations.

Makefile
```diff
@@ -15,9 +15,6 @@ ${DIR}/boogie/Binaries/Boogie.exe:
...
-	make -C ${DIR}/docs/DafnyRef release
-
```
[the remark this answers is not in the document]
cpitclaudel: Correct

Util/Emacs/README
cpitclaudel: Yes: in the README :)

Makefile
cpitclaudel: It's not used

cpitclaudel: Well, I guess that's not right.  Let me revert this bit.

Makefile
```diff
@@ -15,9 +15,6 @@ ${DIR}/boogie/Binaries/Boogie.exe:
 refman:
 	make -C ${DIR}/docs/DafnyRef
```
cpitclaudel, line 17:
```suggestion

refman-release:
	make -C ${DIR}/docs/DafnyRef release
```

Makefile
```diff
@@ -17,7 +17,6 @@ refman:
 
 refman-release:
 	make -C ${DIR}/docs/DafnyRef release
```
cpitclaudel, line 19:
```suggestion
	make -C ${DIR}/docs/DafnyRef release

```

Makefile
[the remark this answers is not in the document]
cpitclaudel: I re-added it.  The `package.py` script had support for calling it, but we don't use that anymore.

MikaelMayer (approved):

PR #2162, opened 2022-05-23
Added verification trees for constructors with default values

MikaelMayer:
Follow-up on gutter status, I had forgotten the case of default parameters. This PR fixes that.

Before, the gutter was all green (but the error was still highlighted)
Now, the gutter correctly highlights where the error is:
image

### DIFF
--- a/Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
+++ b/Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
@@ -78,6 +78,25 @@ public class VerificationProgressReporter : IVerificationProgressReporter {
     foreach (var module in document.Program.Modules()) {
       foreach (var topLevelDecl in module.TopLevelDecls) {
         if (topLevelDecl is TopLevelDeclWithMembers topLevelDeclWithMembers) {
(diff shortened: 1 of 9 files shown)


Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
```diff
@@ -78,6 +78,25 @@ IDiagnosticPublisher diagnosticPublisher
     foreach (var module in document.Program.Modules()) {
       foreach (var topLevelDecl in module.TopLevelDecls) {
         if (topLevelDecl is TopLevelDeclWithMembers topLevelDeclWithMembers) {
+          if (topLevelDecl is DatatypeDecl datatypeDecl) {
```
keyboardDrummer, line 81: Does 81 need to be nested inside 80 ?

MikaelMayer: It does not need (I'll move it as another if of the same level), but since a `DatatypeDecl` is a `TopLevelDeclWithMembers`, I thought it would be more appropriate to place it inside.

keyboardDrummer (approved):

PR #2163, opened 2022-05-25
Document using sequences and multisets in `reads` clauses

atomb:


MikaelMayer (approved):
Wow that's interesting, I learned something today thanks to you, that multisets and seqs are allowed in reads clause, thanks for figuring out.

PR #2166, opened 2022-05-25
Feat: Plugin's Compiler/Verification errors should not prevent verification from running.

MikaelMayer:
A concern I got from the plugin system in Dafny is that users are confused by plugins errors when they would expect verification errors before fixing all the plugin-based errors.

And they have a point:  when a plugin raises a compilation or a verification error in `PostResolve(Program)` after the program was _fully resolved_, this should not prevent DafnyLanguageServer to verify the document.

This PR adds the ability for the Dafny Language Server to continue to verification when only verification and compilation errors are raised in the resolution phase. Only plugins would have the ability to raise verification or compilation errors in the resolution phase anyway, so this makes the plugins able to decide whether they create blocking errors or non-blocking errors.

This PR contains a test that verifies this new behavior, and fixes another plugin test.


cpitclaudel:
With this change, what mechanism do plugins have to stop verification?  Shouldn't a resolution plugin that does not want to break verification only show warnings, instead?

This may be related to compilation errors, too: maybe we should run the compiler and collect errors as part of a normal IDE run, and the plugins that you have in mind should really be compilation plugins?

amaurremi:
> This may be related to compilation errors, too: maybe we should run the compiler and collect errors as part of a normal IDE run, and the plugins that you have in mind should really be compilation plugins?

Would this have the effect that those plugins would only run after verification is done? If so, as @MikaelMayer pointed out, since verification might take a long time, one would also have to wait for a long time for compilation-related errors (such as some features being unsupported by specific compiler).

It would be nice to have the following:
- plugin-related errors are shown as errors, not warnings (because warnings are easy to oversee, and because warnings from other files aren't shown: it would be nice to retain the "imported file contains errors" type of error)
- plugin-related errors are grouped together (as opposed to interleaved with verification- and other errors)
- plugin-related errors don't block verification; it shouldn't be the case that the condition for getting verification errors is to resolve custom plugin errors.

cpitclaudel:
> Would this have the effect that those plugins would only run after verification is done?

No, it wouldn't

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -133,10 +135,12 @@ VerifierOptions verifierOptions
       Dafny.Program program,
       bool loadCanceled
     ) {
+      var parseAndResolutionDiagnostics = errorReporter.GetDiagnostics(textDocument.Uri);
       return new DafnyDocument(
         Options,
         textDocument,
-        errorReporter.GetDiagnostics(textDocument.Uri),
+        parseAndResolutionDiagnostics,
+        parseAndResolutionDiagnostics.Any(d => d.Severity == DiagnosticSeverity.Error),
```
amaurremi, line 143: should this not be `d.Severity != DiagnosticSeverity.Error`? I.e. doesn't this say that verification can be done if there is at least one diagnostic error?

amaurremi: What exactly does this method do? Is it called after resolution already happened, so diagnostic errors here will only come from plugins?

Source/Dafny/Resolver.cs
```diff
@@ -446,9 +446,13 @@ enum ValuetypeVariety {
       rewriters.Add(new InductionRewriter(reporter));
       rewriters.Add(new PrintEffectEnforcement(reporter));
 
+      var builtInRewritersCount = rewriters.Count;
+      var pluginRewriters = new List<IRewriter>() { };
```
amaurremi, line 450: What is the purpose of introducing the `pluginRewriters` variable? Is it just for clarity? I don't see it used after line `454` but maybe I'm missing something.

Source/Dafny/Resolver.cs
```diff
@@ -712,11 +716,17 @@ enum ValuetypeVariety {
         }
       }
 
-      foreach (var rewriter in rewriters) {
+      foreach (var rewriter in rewriters.Take(builtInRewritersCount)) {
+        rewriter.PostResolve(prog);
+      }
+      ErrorsCountWithoutPlugin = reporter.ErrorCount;
```
amaurremi, line 722: This code now strongly depends on rewriters starting with resolver rewriters. Is it a concern that the order would change in the future, but someone forgetting to update this code? Is it possible to get the source of errors in another way, from the errors themselves?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
MikaelMayer: I just checked, and this method creates a document with no program. It is called in case a plugin fails to load, parsing fails, or loading a document is cancelled by a new version of that document.
So first, you were write, it should have been `.All(d.Severity != DiagnosticSeverity.Error)`, but more generally, since verification is not supposed to run after creating this document, I replaced the condition with `false`

Source/Dafny/Resolver.cs
MikaelMayer: Right, since I'm now only taking `builtInRewritersCount`, I can remove these extra variables.

keyboardDrummer:
I find it strange to have the concept of a 'plugin error'. From a user perspective, they are working with a single language. I don't think that the origin of the language, whether it was defined with or without plugins, should change how errors behave.

In case the plugin removes a certain Dafny language feature and the user still uses this, I would treat that same way as if the user is using a feature that never existed in the first place, so to show a pre-verification error and stop.

Could you provide some details on what type of error the plugin is giving?

Source/Dafny/Resolver.cs
MikaelMayer: Yes. Right now, we provide plugins with an `ErrorReporter`. They can raise issues using typically `ErrorReporter.Error(MessageSource source, IToken tok, string msg)`
And, plugins would typically use `MessageSource.Resolver` as the first argument since they are plugin in as resolver issues.
If they used `MessageSource.Verifier`, currently, this message source would be ignored and the pipeline would just stop after resolution.
However, this PR opens the way that error messages using the source `MessageSource.Verifier` would not prevent the verifier from running. For that, I would remove the newly introduced property `Resolver.ErrorsCountWithoutPlugin`, and instead of `nt resolverErrors = resolver.ErrorsCountWithoutPlugin; if (resolverErrors > 0) {...` I would simply look at all non verifications errors raised by the resolver.

amaurremi (dismissed):
Do I understand correctly that the way we now indicate whether verification should be enabled or not is through the `MessageSource`? If I want verification to be blocked, I select the resolver; if I don't, I select the compiler.

If my understanding is correct this looks great to me, thanks for making this plugin behaviour work.

cpitclaudel:
> From my understanding, all compiler plugins are currently run during resolution, which makes them all resolution plugins, no matter what the plugins do.

I don't think so?  We have two kind of plugins, resolver ones and compiler ones.

cpitclaudel:
> But if you, for example, have a compiler plugin that (potentially, temporarily) compiles only a subset of Dafny, then you might want to let the user still get verification feedback before they were able to successfully compile their Dafny program.

Certainly, and that should not raise resolution errors — only compilation ones, which don't preclude verification

keyboardDrummer:
I think it would be nicer for API consumers if a custom Compiler could add checks to restrict the Dafny language to a subset, and then failing those checks wouldn't block the verifier.

With the current API, it seems like customer Compiler authors that support only a Dafny subset must inherit from both IRewriter and Compiler, which seems strange to me.

MikaelMayer:
> Let me try to explain my above comment better.
> 
> I think it would be nicer for API consumers if a custom `Compiler` could add checks to restrict the Dafny language to a subset, and then failing those checks wouldn't block the verifier.

This would be nice indeed, but I think it should be in the scope of a different PR.
Until before this PR, nothing prevented a plugin to emit directly `MessageSource.Compiler` and `MessageSource.Verifier` errors during the resolution with `IRewriter`, anyway. It's just that this PR now ensures such `MessageSource.Compiler` and `MessageSource.Verifier` errors emitted during the resolution phase (through plugins because we don't have yet other use cases) aren't blocking verification.

Moreover, there are also 5 places where we emit in the verifier "assert false" with a custom message  ([witness check](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.cs#L4476), [recursivity check](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.cs#L5891), [incomplete match expression](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.cs#L6279), [incomplete match statement](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.TrStatement.cs#L779) and [datatype recursivity](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.TrStatement.cs#L1873)) because we know statically that some property does not hold, but we moved the check to the verification so that we aren't blocking other verification checks. After this PR, it will be possible to move these "immediately false checks" to the resolver pipeline, so that even if verification is disabled, we can immediately report to the user e.g. missing match cases.

> With the current API, it seems like custom `Compiler` authors that support only a Dafny subset must inherit from both `IRewriter` and `Compiler`. That's not a big problem but it seems like an unnecessary complication.

Compiler are used only for the final compilation step, so they are not used yet by the DafnyLanguageServer. Thus the solution of
- Adding an abstract method to compiler so that they can report errors before final compilation
- Changing the architecture of DafnyLanguageServer so that it runs this method in parallel with verification.

seems like an unnecessary complication if the method provided by user-defined compilers would be fast enough to be run synchronously in the resolution phase anyway. Again, I'm not saying it's bad, it would be great to support that architecture, but it might be just killing a fly with a hammer for the cases we want to support today.

Moreover, a plugin developer do not need to create one class to inherit both "IRewriter" and "SinglePassCompiler", Indeed, authors can make use of the `PluginConfiguration` to instantiate one `Compiler`, and one `IRewriter` that has a pointer to their `Compiler` if they really wanted to put the check in the compiler method and call it through the `IRewriter` wrapper, until we offer them an API for the compiler to provide resolution feedback directly.

keyboardDrummer:
You make very good points. 

> This would be nice indeed, but I think it should be in the scope of a different PR.
Until before this PR, nothing prevented a plugin to emit directly MessageSource.Compiler and MessageSource.Verifier errors during the resolution with IRewriter, anyway. It's just that this PR now ensures such MessageSource.Compiler and MessageSource.Verifier errors emitted during the resolution phase (through plugins because we don't have yet other use cases) aren't blocking verification.

True. I do feel though that this PR is a a good place to do the refactoring we're discussing. 

> Moreover, there are also 5 places where we emit in the verifier "assert false" with a custom message ([witness check](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.cs#L4476), [recursivity check](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.cs#L5891), [incomplete match expression](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.cs#L6279), [incomplete match statement](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.TrStatement.cs#L779) and [datatype recursivity](https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Verifier/Translator.TrStatement.cs#L1873)) because we know statically that some property does not hold, but we moved the check to the verification so that we aren't blocking other verification checks. After this PR, it will be possible to move these "immediately false checks" to the resolver pipeline, so that even if verification is disabled, we can immediately report to the user e.g. missing match cases.

Great!

> Compiler are used only for the final compilation step, so they are not used yet by the DafnyLanguageServer. Thus the solution of adding an abstract method to compiler so that they can report errors before final compilation
Changing the architecture of DafnyLanguageServer so that it runs this method in parallel with verification.
seems like an unnecessary complication if the method provided by user-defined compilers would be fast enough to be run synchronously in the resolution phase anyway. Again, I'm not saying it's bad, it would be great to support that architecture, but it might be just killing a fly with a hammer for the cases we want to support today.

Alright :-)

Source/DafnyLanguageServer/Language/Symbols/DafnyLangSymbolResolver.cs
```diff
@@ -23,33 +23,36 @@ public class DafnyLangSymbolResolver : ISymbolResolver {
       this.logger = logger;
     }
 
-    public CompilationUnit ResolveSymbols(TextDocumentItem textDocument, Dafny.Program program, CancellationToken cancellationToken) {
+    public CompilationUnit ResolveSymbols(TextDocumentItem textDocument, Dafny.Program program, out bool CanDoVerification, CancellationToken cancellationToken) {
```
keyboardDrummer, line 26: the capitalisation seems wrong

keyboardDrummer (dismissed):

keyboardDrummer (approved):

PR #2183, opened 2022-05-30
fix: reads function of seq/multiset

RustanLeino:
This PR fixes two issues with advanced `reads` clauses.

One issue was with `reads F`, where `F` is an expression of type `... ~> seq<object>` or `... ~> multiset<object>`. Previously, these generated malformed Boogie. This is now fixed.

The other issue was with `reads F` where `F` is an expression of type `... ~> object`. Previously, the resolver allowed this, but the translator generated malformed Boogie. There's no strong use case for allowing such a type, so this PR fixes the issue by making the resolver forbid this case.

### DIFF
--- a/Source/Dafny/Resolver.cs
+++ b/Source/Dafny/Resolver.cs
@@ -5465,7 +5465,7 @@ namespace Microsoft.Dafny {
                   return false;  // there is not enough information
                 }
               }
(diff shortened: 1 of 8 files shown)


atomb (approved):
Nicely scoped and well-tested fix!

PR #2189, opened 2022-06-01
fix: Emit notifications in correct order

keyboardDrummer:
### Changes
1. Ensure that when multiple notification generating requests come in that operate on the same document version, such as a change followed by a save, then the notifications created by the requests are emitted in the same order as the requests came in.
2. Translated documents always emit migrated diagnostics

### Testing
Repeatedly ran "WhenUsingOnSaveMethodStaysStaleUntilSave" for 30m with no failures

1. I don't see a reliable way of testing this race condition. I don't think having a slowly resolving document would help. The best we can do is have a change followed by a save and repeat that 50 times, but I'm not enthusiastic about adding a test like that.
2. Updated `ApplyChangeBeforeVerificationFinishes` test so it longer waits to receive a not migrated diagnostic.

### DIFF
--- a/Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
+++ b/Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
@@ -7,8 +7,11 @@ using OmniSharp.Extensions.LanguageServer.Protocol.Document;
 using OmniSharp.Extensions.LanguageServer.Protocol.Models;
 using OmniSharp.Extensions.LanguageServer.Protocol.Server.Capabilities;
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Boogie;
 
 // Justification: The handler must not await document loads. Errors are handled within the observer set up by ForwardDiagnostics.
 #pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
@@ -36,6 +39,7 @@ namespace Microsoft.Dafny.LanguageServer.Handlers {
     private readonly IDocumentDatabase documents;
     private readonly ITelemetryPublisher telemetryPublisher;
     private readonly IDiagnosticPublisher diagnosticPublisher;
+    private readonly Dictionary<DocumentUri, DiagnosticsObserver> observers = new();
 
     public DafnyTextDocumentHandler(
       ILogger<DafnyTextDocumentHandler> logger, IDocumentDatabase documents,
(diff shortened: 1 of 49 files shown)


Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
```diff
@@ -72,33 +76,50 @@ public class DafnyTextDocumentHandler : TextDocumentSyncHandlerBase {
...
       private readonly IDiagnosticPublisher diagnosticPublisher;
+      private readonly Queue<IObservable<DafnyDocument>> allUpdates = new();
 
       public DiagnosticsObserver(ILogger logger, ITelemetryPublisher telemetryPublisher, IDiagnosticPublisher diagnosticPublisher) {
         this.logger = logger;
         this.telemetryPublisher = telemetryPublisher;
         this.diagnosticPublisher = diagnosticPublisher;
       }
 
+      public void AddUpdates(IObservable<DafnyDocument> updates) {
```
MikaelMayer, line 106:
```suggestion
      /// Most of the time, allUpdates contains no `IObservable<DafnyDocument>` when `AddUpdates` is called, so this class directly subscribes to the given observable of documents
      /// However, sometimes a save command can be sent before a change command is done executing. so this class would already be subscribed to a previous observable of document.
      /// This method and the next ensure that notifications are sent in order.
      public void AddUpdates(IObservable<DafnyDocument> updates) {
```

cpitclaudel (dismissed):
This looks good.  Just to clarify: the problem this class is solving is essentially merging a stream of observers as they arrive, so that their relative order is preserved, per-document.  Correct?
There was https://stackoverflow.com/questions/64841312/how-to-merge-multiple-observables-with-order-preservation-and-maximum-concurrenc that seemed relevant, but AFAICT it doesn't handle the case of having the observables arrive one by one.

But then I wonder: why are we creating multiple observables per document, in the first place?  Could we not solve the issue on the other side, by having a single observable into which all diagnostics are sent?

keyboardDrummer:
> This looks good. Just to clarify: the problem this class is solving is essentially merging a stream of observers as they arrive, so that their relative order is preserved, per-document. Correct? There was https://stackoverflow.com/questions/64841312/how-to-merge-multiple-observables-with-order-preservation-and-maximum-concurrenc that seemed relevant, but AFAICT it doesn't handle the case of having the observables arrive one by one.

Link seems relevant, but indeed the scenario is different enough that I can't just apply this.

> But then I wonder: why are we creating multiple observables per document, in the first place? Could we not solve the issue on the other side, by having a single observable into which all diagnostics are sent?

Wouldn't that introduce the problem this PR is solving, where notifications from requests that are concurrently being handled become intertwined?

cpitclaudel:
> Wouldn't that introduce the problem this PR is solving, where notifications from requests that are concurrently being handled become intertwined?

I meant one observable per document, but now I see what you're saying.  Thanks!

PR #2192, opened 2022-06-01
chore: Slowness hints tests to be more robust.

MikaelMayer:
Fixes some failing CI builds by enabling tests to change the amount at which a verification task is considered as "slow"

### DIFF
--- a/Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
+++ b/Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
@@ -2,6 +2,7 @@
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
+using Microsoft.Dafny.LanguageServer.Handlers;
 using Microsoft.Dafny.LanguageServer.IntegrationTest.Extensions;
 using Microsoft.Dafny.LanguageServer.IntegrationTest.Util;
 using Microsoft.Dafny.LanguageServer.Language;
@@ -194,7 +195,10 @@ method f(x: int) {
 
     [TestMethod, Timeout(5 * MaxTestExecutionTimeMs)]
     public async Task IndicateSlowHints() {
-      var documentItem = await GetDocumentItem(@"
+      var original = DafnyHoverHandler.RuLimitToBeOverCostly;
+      DafnyHoverHandler.RuLimitToBeOverCostly = 500000;
+      try {
+        var documentItem = await GetDocumentItem(@"
 lemma {:timeLimit 3} SquareRoot2NotRational(p: nat, q: nat)
   requires p > 0 && q > 0
   ensures (p * p) !=  2 * (q * q)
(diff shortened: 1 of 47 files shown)


Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -194,7 +195,10 @@ method Abs(x: int) returns (y: int)
 
     [TestMethod, Timeout(5 * MaxTestExecutionTimeMs)]
     public async Task IndicateSlowHints() {
-      var documentItem = await GetDocumentItem(@"
+      var original = DafnyHoverHandler.RuLimitToBeOverCostly;
+      DafnyHoverHandler.RuLimitToBeOverCostly = 500000;
```
robin-aws, line 199: Is this bounding the resource count (given the `Ru` prefix)? If so I'm confused as to why the test has been flaky.

MikaelMayer: The problem came that, when I designed the test originally, the verification of the slow to verify method was taking 15M Ru, and I thought this number was stable. However, after running it a few more times, and including looking at the error on the CI build, it looks like verification goes down to using only 7M resources sometimes, and the hard-coded limit is 10M. Hence, I manually set the limit to 500K so that it's always too slow for this test.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -20,7 +20,7 @@ public class DafnyHoverHandler : HoverHandlerBase {
     private readonly ILogger logger;
     private readonly IDocumentDatabase documents;
 
-    private const int RuLimitToBeOverCostly = 10000000;
+    public static int RuLimitToBeOverCostly = 10000000;
```
keyboardDrummer, line 23: I don't feel comfortable using mutable statics. Tests in separate files can be run concurrently and statics introduce the possibility that tests influence each other.

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -210,16 +214,20 @@ method Abs(x: int) returns (y: int)
   }
   assert {:split_here} true;
 } ", "testfileSlow.dfy", CompilationStatus.VerificationFailed);
-      await Task.Delay(100); // Just time for the diagnostics to be updated
-      await AssertHoverMatches(documentItem, (0, 22),
-        @"**Verification performance metrics for method SquareRoot2NotRational**:
+        await Task.Delay(100); // Just time for the diagnostics to be updated
```
keyboardDrummer, line 217: Off topic BUT: is it possible to replace the `await Task.Delay(100)` calls with `await GetLastDiagnostics(..)` ?

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -194,7 +195,10 @@ method Abs(x: int) returns (y: int)
 
     [TestMethod, Timeout(5 * MaxTestExecutionTimeMs)]
     public async Task IndicateSlowHints() {
```
keyboardDrummer, line 197: I don't understand this test. I looked for a 'slow hint' in the expected message but I don't see one.

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -194,7 +195,10 @@ method Abs(x: int) returns (y: int)
...
-      var documentItem = await GetDocumentItem(@"
+      var original = DafnyHoverHandler.RuLimitToBeOverCostly;
+      DafnyHoverHandler.RuLimitToBeOverCostly = 500000;
+      try {
+        var documentItem = await GetDocumentItem(@"
 lemma {:timeLimit 3} SquareRoot2NotRational(p: nat, q: nat)
```
MikaelMayer, line 202:
```suggestion
lemma {:rlimit 12000} SquareRoot2NotRational(p: nat, q: nat)
```

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
robin-aws: For anyone else curious/anxious about this, we realized the resource count was only varying because of the `{:timeLimit ...}` attribute. Naturally if verification is timing out because of a wall-clock time limit, the resource count will vary. :)

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
MikaelMayer: I changed the test name.

keyboardDrummer (approved):

PR #2195, opened 2022-06-02
feat: Domains and ranges for quantified variables

### LINKED ISSUE #2038: Sequence implementation in Go missing Elements() method
Compiling the following (from https://github.com/dafny-lang/libraries/blob/master/src/Collections/Sequences/Seq.dfy) to Go:

```dafny
  /* converts a sequence to a set */
  function method {:opaque} ToSet<T>(s: seq<T>): set<T> 
  {
    set x: T | x in s
  }
```

Results in a Go compiler error:

```
$ dafny /compileTarget:go src/Scratch.dfy

# command-line-arguments
src/Scratch-go/src/Scratch.go:47:37: s.Elements undefined (type dafny.Seq has no field or method Elements)
```

Just a simple omission of a method definition on the `Seq` structure in the Go runtime.

### LINKED ISSUE #2187: Implement quantified variable domains and ranges
Alternate syntax for quantification that will support future features that depend on ordered quantification (sequence comprehensions and `foreach` loops).

See https://github.com/dafny-lang/rfcs/pull/10 for detailed design.

robin-aws:
Resolves #2187. First phase of implementing https://github.com/dafny-lang/rfcs/pull/10.

This change adds two new features on several uses of quantified variables in the language:

1. Quantified variable *domains*, specified with the syntax `x <- C`, where C is an expression with a collection type (i.e. set, multiset, sequence, or map). Besides offering a more compact syntax for a common pattern (i.e. replacing `myLovelyButLongVariableName | myLovelyButLongVariableName in someSet` with `myLovelyButLongVariableName <- someSet`), this is forwards-compatible with the notion of *ordered quantification*, which the two features designed in the linked RFC depend on, and potentially others in the future.

2. Quantified variable *ranges*, specified with the syntax `x | E`, where E is a boolean expression. These replace the existing concept of a single `| E` range after all quantified variables (which is now bound to the last declared variable, with equivalent semantics). These are helpful for restricting the values of one quantified variable so that it can be used in the domain expression of another variable.

These features apply to the quantifier domains used in `forall` and `exists` expressions, `set` and `map` comprehensions, and `forall` statements, and the parsing for quantifier domains is now shared between these features. As a small consequence, set comprehensions now no longer require range expressions, and will default to `| true`.

This new syntax is not fully backwards-compatible: the statement `print set x | 0 <= x < 10, y`, for example, was previously parsed as `print (set x | 0 <= x < 10), (y)` but will now be parsed as `print (set x | 0 <= x < 10, y)` and rejected. The `/quantifierSyntax` option is used to help migrate this otherwise breaking change:

* `/quantifierSyntax:3` keeps the old behaviour where a `| <Range>` always terminates the list of quantified variables.
* `/quantifierSyntax:4` instead attempts to parse additional quantified variables.

Like `/functionSyntax`, this option will default to `4` instead in Dafny 4.0.

This is implemented with pure desugaring for now: the `QuantifiedDomain` production parses a list of `QuantifiedVar` objects, but then immediately rewrites a snippet such as `x <- C | P(x), y <- D(x) | Q(x, y)` into the equivalent (for now) `x | x in C && P(x) && y in D(x) && Q(x, y)`. Token wrappers are used to ensure error messages can still refer to the original syntax. This will not be equivalent once features that depend on the ordering of quantification are added, though, so a subsequent change will add the new fields to `BoundVar` instead and plumb these extra components through the whole Dafny pipeline.

On testing: I duplicated several existing test files and modified them to use the new syntax features, to get good white-box test coverage even though the implementation is very simple for now.

~~NOTE: I still have one more commit to add changes to the reference manual and a few more tests (especially of the `/quantifierSyntax` option), but a review of the rest is welcome in the meantime!~~

Also resolves #2038 (and the same issue for Java) since I hit it when converting test cases, and it was dead easy to fix.


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -6128,6 +6128,63 @@ public BoundVar(IToken tok, string name, Type type)
...
+        bvars.Add(bvar);
+
+        if (qvar.Domain != null) {
+          // Attach a token wrapper so we can produce a better error message if the domain is not a collection
+          var domainWithToken = QuantifiedVariableDomainCloner.Instance.CloneExpr(qvar.Domain);
+          var inDomainExpr = new BinaryExpr(domainWithToken.tok, BinaryExpr.Opcode.In, new IdentifierExpr(bvar.tok, bvar), domainWithToken);
+          range = range == null ? inDomainExpr : new BinaryExpr(domainWithToken.tok, BinaryExpr.Opcode.And, range, inDomainExpr);
+        }
+
+        if (qvar.Range != null) {
```
MikaelMayer, line 6179: I think at some point, we should rename "Range" to "Condition", but let's leave this out of scope for now :-)

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -3423,6 +3423,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
         var b = (ComprehensionExpr.MapBoundedPool)bound;
         TrParenExpr(su.Substitute(b.Map), collectionWriter, inLetExprBody, wStmts);
         GetSpecialFieldInfo(SpecialField.ID.Keys, null, null, out var keyName, out _, out _);
+        // TODO: Shouldn't this be UniqueElements?
```
MikaelMayer, line 3426: Why should it be so? `Elements` are already unique, aren't they?

Source/Dafny/Dafny.atg
```diff
@@ -219,6 +219,44 @@ bool IsIdentifier(int kind) {
   return kind == _ident || kind == _least || kind == _greatest || kind == _older;
 }
 
+bool IsQuantifierVariableDecl(QuantifiedVar previousVar) {
+  // Introducing per-quantified variable ranges creates some ambiguities in the grammar,
+  // since before that the range would terminate the quantifed domain. For example, the following statement:
+  //
+  // print set x: x | E, y;
```
MikaelMayer, line 226:
```suggestion
  // print set x | E, y;
```

Source/Dafny/Dafny.atg
```diff
@@ -219,6 +219,44 @@ bool IsIdentifier(int kind) {
...
+  //
+  // This would previously parse as two separate arguments to the print statement, but
+  // could now also be parsed as a single set comprehension argument with two quantified variables
+  // (and an invalid one since it would need an explicit ":: <Term>" section).
+  //
+  // Even worse:
+  //
+  // print set x: x | E, y <- C;
```
MikaelMayer, line 234:
```suggestion
  // print set x | E, y <- C;
```

Test/comp/ComprehensionsNewSyntax.dfy
```diff
@@ -0,0 +1,284 @@
...
+function method FourMore(x: int): int { x + 4 }
+
+method Quantifier() {
+  // Note, by making *two* assignments to "s" (instead of the obvious one), the direct translation
+  // into Java makes "s" NOT be "effectively final". This means that Java does not allow "s" to be
+  // used in the lambda expression into which the quantifier translates. Compilation into Java
+  // thus needs to capture the value of "s" into another variable, which these tests check on.
+  var s := [0, 1, 1, 2];
+  s := s + [3, 5, 8, 13];
+  print forall x <- s :: x < 20, " ";  // true
```
MikaelMayer, line 29: I'm so happy to see the reduction in the duplication of the variable x! Great job!

Test/comp/ComprehensionsNewSyntax.dfy
```diff
@@ -0,0 +1,284 @@
...
+  // sequentialized forall statement
+  forall a: CellA <- s {
+    a.data := 2;
+  }
+  print c.data, d.data, e.data, "\n";  // 221
+
+  // assign-such-that statement
+  d.data := 9;
+  assert d in s;
+  var u: CellA :| u in s && 7 <= u.data;
```
MikaelMayer, line 185: Could you have:
```suggestion
  var u: CellA <- s | 7 <= u.data;
```
?

Test/comp/ForallNewSyntax.dfy
```diff
@@ -0,0 +1,356 @@
+// RUN: %dafny /quantifierSyntax:4 /compile:3 /spillTargetCode:2 /compileTarget:cs "%s" > "%t"
```
MikaelMayer, line 1: Could you verify once, and then use `/compile:4` here as well?

docs/DafnyRef/Grammar.md
```diff
@@ -561,7 +561,40 @@ A ``FormalsOptionalIds`` is a formal parameter list in which the types are requi
...
+QuantifierDomain(allowLemma, allowLambda) =
+    QuantifierVarDecl(allowLemma, allowLambda) 
+    { "," QuantifierVarDecl(allowLemma, allowLambda) }
+
+QuantifierVarDecl(allowLemma, allowLambda) =
+    IdentTypeOptional
+    [ <- Expression(allowLemma, allowLambda) ]
+    { Attribute }
+    [ | Expression(allowLemma, allowLambda) ]
+````
```
MikaelMayer, line 575: Only three backticks are necessary.

Also, I think we all agreed on that and all the documentation needs to be reformatted accordingly, but could you perhaps start the effort by moving this grammar to the end of this section and replace it with a few examples? That would be less frightening.

MikaelMayer (changes requested):
A few minor things, but the feature introduced in this PR looks great, very well documented and tested, congratulations for the big effort.

Test/comp/ComprehensionsNewSyntax.dfy
cpitclaudel: Oh, I didn't realize that this would be supported in `forall`/`exists`!  I thought it was just for `set`/`seq`/`multiset`

Test/comp/ComprehensionsNewSyntax.dfy
robin-aws: I had exactly the same thought. :) Or perhaps the syntax `var u: CellA <- s :| 7 <= u.data;`

But given this wasn't in scope in the RFC, and let/assign-such-that isn't already using the `QuantifierDomain` syntax, I don't want to pull that in to this step just yet. But it's really tempting as a follow-up, especially because it then allows deterministic behavior (select the first matching value in `s`) without having to prove there exists exactly one such value.

Source/Dafny/AST/DafnyAst.cs
robin-aws: Quite possibly, "Range" is almost as overloaded as "Domain" :) But if we did change it I'd want to change it everywhere.

Source/Dafny/Compilers/SinglePassCompiler.cs
robin-aws: Actually you're right, I confused this with the context of a map comprehension. Will remove.

Test/comp/ComprehensionsNewSyntax.dfy
robin-aws: Might as well! They all have the same underlying quantifier domain semantics, so I like the simplicity of aligning the syntax.

Test/comp/ForallNewSyntax.dfy
robin-aws: Good call, this is just what the original test file was doing.

Test/comp/ComprehensionsNewSyntax.dfy
cpitclaudel: Still worth checking a bit more broadly, maybe ^^

robin-aws: I'm afraid I'm not following, checking what exactly? This was the intention in the RFC all along at least.

Test/comp/ComprehensionsNewSyntax.dfy
robin-aws: I have to admit the more I think about this the more I like it. Essentially `var` would be interpreted as another "binder" option like `forall` or `set`, and in this case it means "take the first value(s) such that...". I even like the way just `var x <- s;` reads as "let x be some element of s".

Test/comp/ComprehensionsNewSyntax.dfy
cpitclaudel: It was unexpected to me, despite having read the proposals, so I'm guessing it might be unexpected to other people who provided feedback as well; that's it :)

Test/comp/ComprehensionsNewSyntax.dfy
MikaelMayer: I did not think about the "first" being interpreted as a "uniqueness", but it makes even more sense. I love it.

docs/DafnyRef/Grammar.md
MikaelMayer: > Are you suggesting moving it to the end of Section 2, as a 2.7?

No I'm not suggesting changing numbering or the order of sections and sub-sections.
I'm only suggesting to not place the grammar at the beginning of this section, only at the end (until it will be moved to another place). The beginning of the section should consist of introductory text to the features and example covering most use cases.

So, instead of:
```
### 2.6.5. Quantifier Domains {#sec-quantifier-domains}
<frightening and unreadable grammar definitions>
<introduction text>
<examples>
<more text>
```

I strongly suggest you write:
```
### 2.6.5. Quantifier Domains {#sec-quantifier-domains}
<introduction text>
<examples>
<more text>
<frightening and unreadable grammar definitions>
```

robin-aws: Aha thanks for clarifying. I will gladly start setting a better example. :)

Source/Dafny/Dafny.atg
```diff
@@ -4702,41 +4736,59 @@ QuantifierExpression<out Expression q, bool allowLemma, bool allowLambda>
...
   .)
-  IdentTypeOptional<out bv>                    (. bvars.Add(bv); .)
-  { ","
-    IdentTypeOptional<out bv>                  (. bvars.Add(bv); .)
+  QuantifierVariableDecl<out qv, ref attrs, allowLemma, allowLambda, allowBitwiseOps>         (. qvars.Add(qv); .)
+  { IF(IsQuantifierVariableDecl(qv))
+    ","
+    QuantifierVariableDecl<out qv, ref attrs, allowLemma, allowLambda, allowBitwiseOps>       (. qvars.Add(qv); .)
   }
+  (. QuantifiedVar.ExtractSingleRange(qvars, out bvars, out range); .)
```
atomb, line 4751: Do I understand correctly that future uses of this new syntax _won't_ call `ExtractSingleRange`, but will keep the components separate? That seems like it would be necessary for the expected operational semantics of `foreach`, for example (once you get to that!).

atomb (changes requested):
This is very nice! All of the changes I requested are tiny little things (and one question).

Source/Dafny/Dafny.atg
robin-aws: That's right, the next PR will add domains and ranges to `BoundVar` instead, and `ExtractSingleRange` will likely be removed. But this PR will establish a ton of test cases that still have to work after that refactoring. :)

Test/comp/ComprehensionsNewSyntax.dfy
```diff
@@ -0,0 +1,284 @@
...
+  MapComprehension();
+  OutParamsUnderLambdas();  // white-box testing
+  SetComprehension();
+  Enumerations();
+  EnumerationsMaybeNull();
+  TestImplicitTypeTests.Test();
+}
+
+predicate method Thirteen(x: int) { x == 13 }
+predicate method Even(y: int) { y % 2 == 1 }
```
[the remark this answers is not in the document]
robin-aws: Ah good point, I removed methods from the original `Test/comp/Comprehensions.dfy` that didn't use the syntax and didn't clean these up. :)

Source/Dafny/DafnyOptions.cs
```diff
@@ -992,6 +1010,16 @@ is printed.
     experimentalPredicateAlwaysGhost - Compiled functions are written `function`.
         Ghost functions are written `ghost function`. Predicates are always ghost
         and are written `predicate`.
+/quantifierSyntax:<version>
```
keyboardDrummer, line 1013: Would be nice if we can reuse `Options.txt` here.

keyboardDrummer (approved):
Love the relatively small scope of this PR :-) makes the review manageable.

Quality of code, documentation and tests looks excellent!

RELEASE_NOTES.md
```diff
@@ -1,6 +1,7 @@
 # Upcoming
 
 - feat: The IDE will show verification errors for a method immediately after that method has been verified, instead of after all methods are verified.
+- feat: New syntax for quantified variables, allowing per-variable domains (`x <- C`) and ranges (`x | P(x), y | Q(x, y)`). See Section [2.6.5](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-quantifier-domains) of the Reference Manual.
```
MikaelMayer, line 4:
```suggestion
- feat: New syntax for quantified variables, allowing per-variable domains (`x <- C`) and ranges (`x | P(x), y | Q(x, y)`). See [the quantifier domain section](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-quantifier-domains) of the Reference Manual.
```
Let's not use hard numbers, they can get outdated (for example, 2.6.5 currently refers to numeric literals, what can it refer to in one year.

atomb (dismissed):
Nice work! I'm looking forward to the point when we can start using these new features. :)

docs/DafnyRef/Grammar.md
```diff
@@ -562,6 +562,52 @@ but the names of the parameters are optional. This is used in algebraic
...
+which binds the variable `x` to the values `0`, `1`, `2`, `3`, `4`, and `5`.
+
+Here are some more examples.
+
+- `x: byte` (where `type byte = x | 0 <= x < 256`)
+- `x: nat | x <= 5`
+- `x <- integerSet`
+- `x: nat <- integerSet`
+- `x: nat <- integerSet | x % 2 == 0`
+- `x: nat, y: nat | x < 2 && y < 2`
```
MikaelMayer, line 582:
```suggestion
- `x: nat, y: nat | x < 2 && y < 2`
- `x: nat  | x < 2, y: nat | y < x`
```
It would be good to have a simpler example just before the next one, because the next one uses vertical bars for a different meaning.

docs/DafnyRef/Grammar.md
```diff
@@ -562,6 +562,52 @@ but the names of the parameters are optional. This is used in algebraic
...
+may also include any of three optional elements:
+
+1. The optional syntax `: T` declares the type of the quantified variable.
+   If not provided, it will be inferred from context.
+
+2. The optional syntax `| E` attaches a boolean expression `E` as a *quantified variable range*,
+   which restricts the bindings to values that satisfy this expression.
+   In the example above `x <= 5` is the range attached to the `x` variable declaration.
+
+3. The optional syntax `<- C` attaches a collection expression `C` as a *quantified variable domain*.
```
MikaelMayer, line 596: Could you move this one before 2, so that it's the order in which we parse things?

MikaelMayer (changes requested):
Great job with the documentation, I like it much much better!
A few tweaks to make documentation more friendly.

docs/DafnyRef/Grammar.md
robin-aws: Yeah I like this, more of a gradual onramp of complexity :)

docs/DafnyRef/Grammar.md
robin-aws: I was on the edge of doing this already and you have tipped me over :)

MikaelMayer (approved):

PR #2198, opened 2022-06-02
chore: Fixed all the broken links in the documentation

MikaelMayer:
This effort will probably break the PDF generation of the documentation, but I think we care more about the website.

### DIFF
--- a/docs/DafnyRef/Attributes.md
+++ b/docs/DafnyRef/Attributes.md
@@ -277,7 +277,16 @@ lemma Fill_J(s: seq<int>)
 }
 ```
 
-### 22.2.8. `{:print}` {#sec-print}
+### 22.2.8. `{:opaque}` {#sec-opaque}
+Ordinarily, the body of a function is transparent to its users, but
+sometimes it is useful to hide it. If a function `foo` or `bar` is given the
+`{:opaque}` attribute, then Dafny hides the body of the function,
+so that it can only be seen within its recursive clique (if any),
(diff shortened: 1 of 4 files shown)


fabiomadge:
You're right, it does break the pdf. I had a related problem in #2188, maybe we should dump Jekyll and use pandoc for html generation as well
[DafnyRef.pdf](https://github.com/dafny-lang/dafny/files/8832945/DafnyRef.pdf)
.

MikaelMayer:
> You're right, it does break the pdf in different ways. I had a related problem in #2188. Maybe we should dump Jekyll and use pandoc for html generation as well. [DafnyRef.pdf](https://github.com/dafny-lang/dafny/files/8832945/DafnyRef.pdf)

After looking at it, since I'm using `id="theId"`  in HTML tags and I removed `~ Equation {#theId}` to create references, it makes sense that they don't work in LaTeX anymore. Is there a way we can re-add these labels so that they work for both? Do you see any other breaking changes?

fabiomadge (approved):
It's undoubtedly better than the status quo, but it would be nice, if we could enable the references in the pdf version eventually.

MikaelMayer:
> It's undoubtedly better than the status quo, but it would be nice, if we could enable the references in the pdf version eventually.

Yes, I agree, but I don't know how we could do that. A wrapper script before using pandoc to add `~ Equation ... ~` and move references when encountering an `id=""`?

PR #2202, opened 2022-06-03
Feat: Show snippets range

### LINKED ISSUE #2197: Ranges are not highlighted properly
image

The reported ranges are missing the closing square bracket, and occasionally the closing parenthesis.

MikaelMayer:
Fixes #2197 

Fixes https://github.com/dafny-lang/ide-vscode/issues/189 as well

The behavior of the command-line option `/showSnippets:1` changes from
```
ShowSnippets.dfy(17,7): Error: A precondition for this call might not hold.
   |
17 |   Never();
   |        ^ here
```
to
```
ShowSnippets.dfy(17,2): Error: A precondition for this call might not hold.
   |
17 |   Never();
   |   ^^^^^^^^
```

It works also for binary expressions containing input and output variables (also in the language server)

Previously (after activating the range feature), the variable was substituted with a box version of the output variable for the verifier, so it was loosing track of the token for y:
```
git-issue-2197.dfy(6,2): Related location
  |
6 |   y >= 1
  |     ^^^^

```
Now, when I substitute, i wrap the substitution with a `ParensExpression` that keeps the original token (which we were explicitly doing for input identifiers already). I tried to come up with a clever version of a relocating cloner, but all my attempts failed, so this is the best I got:
```
git-issue-2197.dfy(6,2): Related location
  |
6 |   y >= 1
  |   ^^^^^^

```

This PR not only enables display of ranges for the command line, but it also fixes displayed ranges for sequence select expressions, function calls expressions, method call statements, and correct handling of output variables.

I changed Expression to store a potentially null array of formatting tokens, and it's only used for now to store the closing parenthesis, so that the range extends to the end of the expression when needed. I expect the Code formatting PR to replace this array by something else that computes accurately the Range.


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -10528,13 +10550,15 @@ public class SeqSelectExpr : Expression {
     public readonly Expression Seq;
     public readonly Expression E0;
     public readonly Expression E1;
+    public readonly IToken CloseParens;
```
robin-aws, line 10553:
```suggestion
    public readonly IToken CloseParen;
```

I find `CloseParens` implies "Close Parentheses" when it should be singular (and you used `CloseParen` elsewhere too)

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -9158,27 +9158,49 @@ public abstract class Expression {
 
     private RangeToken rangeToken;
 
+    protected IToken[] FormatTokens = null;
```
robin-aws, line 9161: I'd add a comment here about what this is for.

Source/DafnyLanguageServer/Language/DafnyLangParser.cs
```diff
@@ -39,6 +39,7 @@ public sealed class DafnyLangParser : IDafnyParser, IDisposable {
           DafnyOptions.Install(DafnyOptions.Create());
           DafnyOptions.O.ApplyDefaultOptions();
           DafnyOptions.O.PrintIncludesMode = DafnyOptions.IncludesModes.None;
+          DafnyOptions.O.ShowSnippets = true;
```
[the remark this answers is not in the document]
MikaelMayer: I'm going to use this option in the Translator to determine whether to use bare tokens or range tokens. See here:
https://github.com/dafny-lang/dafny/pull/2202/files#diff-7d845d452486b0b59700ca5c133f2648f76d44616148081d51efffe32dbc6e7aR16741

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: >I'd add a comment here about what this is for.

Added
```
    // Contains tokens that did not make it in the AST but are part of the expression,
    // Enables ranges to be correct.
```

Source/DafnyLanguageServer/Language/DafnyLangParser.cs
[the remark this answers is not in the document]
robin-aws: I see, so in this case `ShowSnippets` is really a proxy for a non-existent, more general `ParseRangeTokens` option. That's fine for now, but I'd add a comment to clarify.

robin-aws (dismissed):

MikaelMayer:
Note to myself: When this is merged, we can consider updating hovering in LSP to actually display failing postconditions and failing preconditions.

robin-aws (approved):

PR #2204, opened 2022-06-04
Add a script to download all releases

cpitclaudel:
This is a small script that downloads all Dafny releases from GitHub (having a
bunch of releases on hand is convenient for getting a rough estimate of when a
bug was introduced).


atomb (dismissed):
This is very nice. I would have found this useful recently. I especially like that it automatically extracts the archives into versioned directories, which is one of the most labor-intensive aspects of doing this manually.

(In the longer term it makes me wonder whether we should do Dafny releases with a path prefix that includes the version number, as most open source projects do.)

Scripts/download_all_releases.py
```diff
@@ -0,0 +1,112 @@
...
+    except:
+        shutil.rmtree(target_dir)
+        raise
+
+def get_platform_glob() -> str:
+    system = platform.system()
+    if system == 'Linux':
+        return '*ubuntu*'
+    if system == 'Darwin':
+        return '*mac*'
```
wilcoxjay, line 83: this should be `*osx*` I think

cpitclaudel: Good catch, thanks!

cpitclaudel, line 83:
```suggestion
        return '*osx*'
```

atomb (approved):

PR #2213, opened 2022-06-07
Reuse execution engine throughout language server lifetime

### LINKED ISSUE #168: Dafny warnings are reported as errors; Dafny won't run program; breaks VSCode plug-in
I'm running current Dafny with the Dafny VSCode plugin in VSCode.

The problem is that my Dafny program verifies, but when I try to "Compile and Run" in VSCode,
 Dafny reports, "Errors compiling program into basic-dafny.exe." Here's the full compiler output. 

```
Dafny program verifier finished with 9 verified, 0 errors
Errors compiling program into basic-dafny.exe
/var/folders/5n/6hml8q9s39300wsyb2kb89b00000gn/T/4bcad2d0/5e25f480.0.cs(1792,19) : warning CS1718: A comparison made to same variable. Did you mean to compare something else?
/var/folders/5n/6hml8q9s39300wsyb2kb89b00000gn/T/4bcad2d0/5e25f480.0.cs(1854,7) : warning CS0162: Unreachable code detected
```

The problem I face is that Dafny won't run the compiled code when this happens, even though compiled code is generated, and in the "Dafny VSCode" plugin, (1) the only way to run a Dafny program is using the Compile and Run option, and (2) VSCode does not report the warnings at all. So one is left with a mysterious situation in which the code verifies, an inexplicable error is given, and the program can't be run. (It can be run if one launches a terminal window and uses mono to execute the .exe file that *is* generated by the compiler).

Suggested fixes: (1) Change "Errors compiling program into basic-dafny.exe" to "Warnings compiling program into basic-dafny.exe" when only warnings are generated, and (2) let the compiled executable run under these circumstances. If you don't want do (2) I understand but (1) could help the plug-in developers to better handle the "warnings generated" situation.

Kevin

keyboardDrummer:
Fixes #168
Alternative to this fix: https://github.com/dafny-lang/dafny/pull/2209

### Changes
Use a single ExecutionEngine and associated CheckerPool throughout the language server lifetime.

This has the downside that some memory is occupied by Z3 processes even when no verification is being run. The memory usage for an idle Z3 process seems to be about 4MB. The potential upside is that fresh verification requests can be executed slightly faster, but I don't know by how much.

Looking at alternatives to a single `ExecutionEngine`:
- One `ExecutionEngine` per `ImplementationTask`. This way the benefit of having a pool of solvers is lost, since many `ImplementationTask`'s have only a single VC.
- [What we currently do] One `ExecutionEngine` per document change. This makes it hard to know when to dispose the engine. When a new document change comes in, verification may still be running so we can't dispose of the old execution engine then. Also, we can't dispose the engine once all running verification that uses it is done, because we also want to make it possible to manually invoke verification. We could do something like dispose the engine once a newer change has come in and no tasks using it are currently running, but that's complicated and I don't see advantages to this approach.

### Tests
Add a test to detect whether z3 processes are being orphaned or not.


Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -181,8 +181,9 @@ bool loadCanceled
       notificationPublisher.SendStatusNotification(document.TextDocumentItem, CompilationStatus.VerificationStarted);
 
       var progressReporter = CreateVerificationProgressReporter(document);
-      document.VerificationTasks!.BatchCompletions.Subscribe(progressReporter.ReportAssertionBatchResult);
-      var implementationTasks = document.VerificationTasks!.Tasks;
+      var subscription = verifier.BatchCompletions.Subscribe(progressReporter.ReportAssertionBatchResult);
```
MikaelMayer, line 184: I'm not sure this is correct. On the deleted code, we correctly subscribe all notifications for only the verification tasks for a given document.
On the new code, every document verification process subscribes to all messages for all batch results.

This is not caught by a test because I don't think there is a test to verify two documents with multiple splits at the same time.

The problem is that the outputPrinter is an option of the execution engine, so if we want one execution engine only, and there can be only one printer, we need to do routing ourselves.

One solution is to, in ReportAssertionBatchResult, verify that the split result is for the current document; and if not, we just ignore the message. That's one routing option.

Also, when do you unsubscribe these progressReporter?

MikaelMayer (changes requested):
Great move towards being able again to reuse the cache !
I have a concern about subscribing to split results, see my comments.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: Nicely spotted. Pushed a new revision. The unsubscription happens on cancellation.

I spent a little bit of time today trying to refactor Boogie, splitting the options into one part that is passed to CheckerPool and cannot be changed, and another part that can be changed on each run and would take the Printer that ideally we want to change each run, but that seemed like quite a lot of work, so I discontinued and thought we could get away with something like this.

MikaelMayer (approved):
It looks good to keep reusing the same execution engine. Great job.

PR #2215, opened 2022-06-07
Add Run and cancel API

keyboardDrummer:
### Changes

- Update to latest Boogie (2.15.3)
- Publish unhandled exceptions that occur synchronously in notification handlers.
- Do not publish empty notifications for documents that have just been opened.
- Refactoring: rename IDiagnosticPublisher to INotificationPublisher
- Refactoring: moved `RequestUpdatesOnUriObserver` inside `DocumentEntry` so we don't need `Dictionary<DocumentUri, RequestUpdatesOnUriObserver> observers`
- Refactoring: move `RequestUpdatesOnUriObserver` and `DiagnosticsObserver ` to a separate file
- Migrate the Range property inside `dafny/textDocument/symbolStatus` notifications
- Fix a concurrency bug in `Translator.cs`: 
  - Fixes https://github.com/dafny-lang/dafny/issues/2237  
  - Commit: https://github.com/dafny-lang/dafny/pull/2215/commits/0a107371d0ee2d57a333a7d2784091efd617d5fa 
- Add an API for running and cancelling verifiable symbols.
  - `dafny/textDocument/verifySymbol`
  - `dafny/textDocument/cancelVerifySymbol`

### Tests

- Add tests for "Add an API for running and cancelling verifiable symbols." and "Migrate the Range property inside `dafny/textDocument/symbolStatus` notifications"
- Adapted existing tests for "Do not publish empty notifications for documents that have just been opened."


Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -26,8 +27,8 @@ public record DafnyDocument(
     IReadOnlyList<Diagnostic> ParseAndResolutionDiagnostics,
     // VerificationDiagnostics can be deduced from CounterExamples,
     // but they are stored separately because they are migrated and counterexamples currently are not.
-    IReadOnlyDictionary<ImplementationId, ImplementationView>? ImplementationViews,
-    IReadOnlyList<Counterexample> CounterExamples,
+    IReadOnlyDictionary<ImplementationId, ImplementationView>? ImplementationViewsView,
```
keyboardDrummer, line 30: In a future PR I plan to refactor `DafnyDocument`, so there is one mutable version which I'll rename to `Compilation`, and one immutable version which is a `CompilationSnapshot` (or something) that is used for publishing information. It's useful to have an immutable object so that when this object is on its way to be published, it is not mutated by ongoing processes, which is especially useful for reliable testing.

Source/Dafny/MergeOrdered.cs
```diff
@@ -16,14 +16,15 @@ namespace Microsoft.Dafny;
...
   private readonly Queue<IObservable<T>> allUpdates = new();
-  private IDisposable innerSubscription;
+  private bool idle = true;
   private bool outerCompleted;
   private readonly Subject<T> result = new();
 
   public void OnNext(IObservable<T> next) {
     lock (this) {
-      if (innerSubscription == null) {
-        innerSubscription = next.Subscribe(InnerNext, InnerError, InnerCompleted);
```
keyboardDrummer, line 26: Here `innerSubscription` was assigned which is supposed to be nulled in `InnerCompleted`, but if `next.Subscribe` calls the passed OnComplete handler before it returns, `InnerCompleted` in this case, which happens when `next` is `Observable.Empty<T>()`, then `innerSubscription` would end up never getting nulled again.

MikaelMayer (changes requested):
I could not look at all the files but I thought of publishing this initial review first and I'll continue tomorrow.

Source/Dafny.sln
```diff
@@ -34,6 +34,34 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoExtern", "AutoExtern\Au
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoExtern.Test", "AutoExtern.Test\AutoExtern.Test.csproj", "{A47E2F45-DEA3-4700-A82F-9506FEEB199A}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Boogie", "Boogie", "{60332269-9C5D-465E-8582-01F9B738BD90}"
```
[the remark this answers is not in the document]
keyboardDrummer: Strange, I think I already did that.

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
```diff
@@ -48,8 +46,6 @@ decreases y
 }".TrimStart();
       var documentItem = CreateTestDocument(source);
       await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
-      var diagnostics = await GetLastDiagnostics(documentItem, CancellationToken);
-      Assert.AreEqual(0, diagnostics.Length);
```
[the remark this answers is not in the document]
keyboardDrummer: > Do you have any motivation for not publishing diagnostics that I am not aware of?

It seems better not to send notifications that have no effect on the client.

Source/DafnyLanguageServer/DafnyRequestNames.cs
```diff
@@ -5,5 +5,7 @@ public static class DafnyRequestNames {
     public const string GhostDiagnostics = "dafny/ghost/diagnostics";
     public const string VerificationStatusGutter = "dafny/verification/status/gutter";
     public const string VerificationSymbolStatus = "dafny/textDocument/symbolStatus";
+    public const string VerifySymbol = "dafny/textDocument/verifySymbol";
```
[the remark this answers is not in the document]
keyboardDrummer: > Could you perhaps separate the two sections so that it's clear what symbols are used to send to the server, and what symbols are sent from the server to the client?

Why do you want to see that here? Note that `public const string CounterExample = "dafny/counterExample";` is an existing request that's listed among the remaining notifications.

I'm not sure what the purpose is of having these constants in a single file. I'm guessing the goal is to have a starting point for finding all the added APIs. If we split this file into two then that feature is lost and I think we might as well move all the constants to where they are used.

Maybe we can have a separate PR where we think about how to document our LSP extensions?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -99,17 +97,41 @@ VerifierOptions verifierOptions
...
-        var view = new ImplementationView(task.Implementation.tok.GetLspRange(), status, Array.Empty<Diagnostic>());
-        initialViews.Add(GetImplementationId(task.Implementation), view);
+        var status = StatusFromBoogieStatus(task.CacheStatus);
+        var id = GetImplementationId(task.Implementation);
+        if (loaded.ImplementationViewsView!.TryGetValue(id, out var existingView)) {
+#pragma warning disable VSTHRD002
+          initialViews.TryAdd(id, existingView with { Status = status });
+#pragma warning restore VSTHRD002
+        } else {
+          // TODO add diagnostics in cache results are cached, and then also make sure not to jump into the then branch.
```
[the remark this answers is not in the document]
keyboardDrummer: I've replaced it with an implementation :)

Source/DafnyLanguageServer/Workspace/RequestUpdatesOnUriObserver.cs
```diff
@@ -0,0 +1,90 @@
...
+
+  public DiagnosticsObserver(ILogger logger, ITelemetryPublisher telemetryPublisher,
+    INotificationPublisher notificationPublisher, DocumentTextBuffer document) {
+    PreviouslyPublishedDocument = new DafnyDocument(document,
+      ArraySegment<Diagnostic>.Empty,
+      false,
+      new Dictionary<ImplementationId, ImplementationView>(),
+      ArraySegment<Counterexample>.Empty,
+      ArraySegment<Diagnostic>.Empty,
+#pragma warning disable CS8625
```
keyboardDrummer, line 62: Note: I'm not happy with the structure of DocumentEntry and DafnyDocument. I hope to refactor the structure a bit in a follow-up PR and then these pragma's won't be needed.

Source/DafnyLanguageServer.Test/Various/SlowVerifier.cs
```diff
@@ -40,25 +40,23 @@ class SlowVerifier : IProgramVerifier {
 
   class NeverVerifiesImplementationTask : IImplementationTask {
     private readonly IImplementationTask original;
-    private readonly TaskCompletionSource<VerificationResult> source;
+    private readonly Subject<IVerificationStatus> source;
```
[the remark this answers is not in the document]
keyboardDrummer: That's a good analogy

Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs
```diff
@@ -13,6 +13,56 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization;
...
+    var documentItem = CreateTestDocument(source);
+    await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+    var methodHeader = new Position(0, 7);
+    client.RunSymbolVerification(new TextDocumentIdentifier(documentItem.Uri), methodHeader);
+    await WaitUntilAllStatusAreCompleted(documentItem);
+
+    ApplyChange(ref documentItem, new Range(0, 0, 0, 0), "\n");
+
+    var afterChangeDiagnostics = await diagnosticsReceiver.AwaitNextDiagnosticsAsync(CancellationToken);
+    Assert.AreEqual(1, afterChangeDiagnostics.Length);
```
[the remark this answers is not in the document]
keyboardDrummer: I think this test indeed doesn't make sense. Updated it.

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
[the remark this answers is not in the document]
MikaelMayer: A client could definitely decide to render a diagnostic window differently depending on whether they are published an empty, or if they are not published.

For example, a client could decide to display an initial placeholder message "waiting for diagnostics" and then removing it once it receives empty diagnostics.

From this (slightly ambiguous) specification:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/

it says that:
>When a file changes it is the server’s responsibility to re-compute diagnostics and push them to the client

I originally disagreed not publishing diagnostics if they were the same as before, but because it made some tests simpler, I ended up agreeing on that "hack", but I think that the spirit is that it should actually push diagnostics, at least the first time, to notify that it's alive, e.g. when opening a new file.

Source/DafnyLanguageServer/DafnyRequestNames.cs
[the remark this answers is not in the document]
MikaelMayer: > I'm not sure what the purpose is of having these constants in a single file. I'm guessing the goal is to have a starting point for finding all the added APIs. If we split this file into two then that feature is lost and I think we might as well move all the constants to where they are used.

I did not say split into two files. Both classes can be there as well. I understand your concern that the file should match the class.
Then, at least, please separate them visually and/or with a comment.

And since you observed that counterExample is a push notification (which I didn't know), please move it to the same section visually.

E.g.

```csharp
  public static class DafnyRequestNames {
    // Notifications from server to client
    public const string CompilationStatus = "dafny/compilation/status";
    public const string GhostDiagnostics = "dafny/ghost/diagnostics";
    public const string VerificationStatusGutter = "dafny/verification/status/gutter";
    public const string VerificationSymbolStatus = "dafny/textDocument/symbolStatus";
    // Requests from client to server
    public const string CounterExample = "dafny/counterExample";
    ... your code here...
}
```

Source/Dafny/Util.cs
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 using System;
+using System.Collections;
```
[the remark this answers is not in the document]
MikaelMayer: Why do you need this one?

Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
```diff
@@ -65,102 +61,54 @@ public class DafnyTextDocumentHandler : TextDocumentSyncHandlerBase {
...
       return Unit.Task;
     }
 
     public override Task<Unit> Handle(DidChangeTextDocumentParams notification, CancellationToken cancellationToken) {
       logger.LogTrace("received change notification {DocumentUri}", notification.TextDocument.Uri);
-      ForwardDiagnostics(notification.TextDocument.Uri, documents.UpdateDocument(notification));
+      try {
+        documents.UpdateDocument(notification);
+      } catch (Exception e) {
+        telemetryPublisher.PublishUnhandledException(e);
```
MikaelMayer, line 87: Great job adding this try/catch.

Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
```diff
@@ -65,102 +61,54 @@ public class DafnyTextDocumentHandler : TextDocumentSyncHandlerBase {
 
     public override Task<Unit> Handle(DidOpenTextDocumentParams notification, CancellationToken cancellationToken) {
       logger.LogTrace("received open notification {DocumentUri}", notification.TextDocument.Uri);
-      ForwardDiagnostics(notification.TextDocument.Uri, documents.OpenDocument(DocumentTextBuffer.From(notification.TextDocument)));
+      try {
+        documents.OpenDocument(DocumentTextBuffer.From(notification.TextDocument));
+      } catch (Exception e) {
+        telemetryPublisher.PublishUnhandledException(e);
```
MikaelMayer, line 67: Nice, that should be very helpful to debug

Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
```diff
@@ -65,102 +61,54 @@ public class DafnyTextDocumentHandler : TextDocumentSyncHandlerBase {
...
-      public void OnNext(DafnyDocument document) {
-        diagnosticPublisher.PublishDiagnostics(document);
-      }
+      return Unit.Task;
     }
 
     private async Task CloseDocumentAndHideDiagnosticsAsync(TextDocumentIdentifier documentId) {
       try {
         await documents.CloseDocumentAsync(documentId);
+        notificationPublisher.HideDiagnostics(documentId);
```
MikaelMayer, line 106: It's already hiding diagnostics in VSCode when I close a file, but I guess here it's more implementing this behavior on the LSP protocol side of view?

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -82,6 +86,10 @@ CancellationToken cancellationToken
         });
       }
 
+      public Range MigrateRange(Range range) {
```
MikaelMayer, line 89: Question: MigrateRange always returns a range, even if the original range was deleted. Won't this cause a problem?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -27,8 +28,8 @@ public record DafnyDocument(
     bool CanDoVerification,
     // VerificationDiagnostics can be deduced from CounterExamples,
     // but they are stored separately because they are migrated and counterexamples currently are not.
-    IReadOnlyDictionary<ImplementationId, ImplementationView>? ImplementationViews,
-    IReadOnlyList<Counterexample> CounterExamples,
+    IReadOnlyDictionary<ImplementationId, ImplementationView> ImplementationViewsView,
```
MikaelMayer, line 31: I suggest another naming that is less confusing:
```suggestion
    IReadOnlyDictionary<ImplementationId, ImplementationView> ImplementationIdToView,
```
```suggestion
    IReadOnlyDictionary<ImplementationId, ImplementationView> ImplementationViewsView,
```

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -71,6 +70,9 @@ public record DafnyDocument(
     }
 
     public int LinesCount => VerificationTree.Range.End.Line;
+    public IVerificationProgressReporter? GutterProgressReporter { get; set; }
+    public ConcurrentStack<Counterexample>? Counterexamples { get; set; }
+    public ConcurrentDictionary<ImplementationId, ImplementationView>? ImplementationViews { get; set; }
```
MikaelMayer, line 75: Same here, we could have
```suggestion
    public ConcurrentDictionary<ImplementationId, ImplementationView>? ImplementationIdToView { get; set; }
```
to not only reflect what values are, but also what keys are.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -67,23 +73,29 @@ IRelocator relocator
...
-      return resolvedDocumentTask.ToObservableSkipCancelled().Where(d => !d.LoadCanceled).
-        Concat(translatedDocument.ToObservableSkipCancelled()).
-        Concat(verifiedDocuments);
+        cancellationSource,
+        new RequestUpdatesOnUriObserver(logger, telemetryPublisher, notificationPublisher, document)
+      );
+      documents.Add(document.Uri, documentEntry);
+
+      documentEntry.Observe(
+        resolvedDocumentTask.ToObservableSkipCancelled().Where(d => !d.LoadCanceled).
```
MikaelMayer, line 93: Would it be possible, perhaps, to not just skip cancelled tasks, but also tasks which contain errors?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -134,36 +145,39 @@ IRelocator relocator
 
       databaseEntry.CancelPendingUpdates();
       var cancellationSource = new CancellationTokenSource();
-      var previousDocumentTask = databaseEntry.LatestDocument;
-      var resolvedDocumentTask = ApplyChangesAsync(previousDocumentTask, documentChange, cancellationSource.Token);
+      var updatedText = textChangeProcessor.ApplyChange(databaseEntry.TextBuffer, documentChange, CancellationToken.None);
```
MikaelMayer, line 148: Excellent ! This removes the computation of the updated text to be in a task, which was probably a cause of concurrent issues such as https://github.com/dafny-lang/dafny/issues/2246

This also means that the TextBufferItem is now shared between the `DocumentEntry` and the `DafnyDocument`, with the second possibly outdated.
Can you confirm you ensures that `UpdateDocument` itself will never fail, so that we are immediately left with the newest version of the document every time?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -134,36 +145,39 @@ IRelocator relocator
...
-      var updatedText = textChangeProcessor.ApplyChange(oldDocument.TextDocumentItem, documentChange, CancellationToken.None);
-      var oldVerificationDiagnostics = oldDocument.ImplementationViews ?? new Dictionary<ImplementationId, ImplementationView>();
+      var oldVerificationDiagnostics = oldDocument.ImplementationViewsView ?? new Dictionary<ImplementationId, ImplementationView>();
       var migratedImplementationViews = oldVerificationDiagnostics.ToDictionary(
         kv => kv.Key with {
           NamedVerificationTask =
           relocator.RelocatePosition(kv.Key.NamedVerificationTask, documentChange, CancellationToken.None)
         },
         kv => kv.Value with {
+          Range = relocator.RelocateRange(kv.Value.Range, documentChange, CancellationToken.None),
```
MikaelMayer, line 180: I think here you should remove an implementation view if the ranged was deleted.

Source/DafnyLanguageServer/Workspace/RequestUpdatesOnUriObserver.cs
MikaelMayer: Just a quick question: Why are you using ArraySegment?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -99,17 +97,41 @@ VerifierOptions verifierOptions
...
       foreach (var task in verificationTasks) {
-        var status = await StatusFromImplementationTaskAsync(task);
-        var view = new ImplementationView(task.Implementation.tok.GetLspRange(), status, Array.Empty<Diagnostic>());
-        initialViews.Add(GetImplementationId(task.Implementation), view);
+        var status = StatusFromBoogieStatus(task.CacheStatus);
+        var id = GetImplementationId(task.Implementation);
+        if (task.CacheStatus is Completed completed) {
+          var view = new ImplementationView(task.Implementation.tok.GetLspRange(), status, GetDiagnosticsFromResult(loaded, completed.Result));
+          initialViews.TryAdd(GetImplementationId(task.Implementation), view);
+        } else if (loaded.ImplementationViewsView!.TryGetValue(id, out var existingView)) {
```
MikaelMayer, line 108: Can you please add a comment to this method about how we ensure that `ImplementationViewsView` is not null? What is the method that needed to be called before, if any?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -184,126 +206,103 @@ bool loadCanceled
...
+            implementationTask.Cancel();
+          } catch (InvalidOperationException e) {
+            if (!e.Message.Contains("no ongoing run")) {
+              throw;
+            }
+          }
+        });
+        return GetVerifiedDafnyDocuments(dafnyDocument, implementationTask, observable);
+      } catch (InvalidOperationException e) {
+        if (e.Message.Contains("already completed") || e.Message.Contains("ongoing run")) {
```
MikaelMayer, line 248: `"no ongoing run".Contains("ongoing run")` is true... is that a problem?

Could you at least put these strings into constants and define these constant in an appropriate place?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -184,126 +206,103 @@ bool loadCanceled
...
+      var initial = document with {
+        ImplementationViewsView = document.ImplementationViews!.ToImmutableDictionary(),
+      };
+      return Observable.Return(initial).Concat(result);
+    }
+
+    DafnyDocument HandleStatusUpdate(DafnyDocument document, IImplementationTask implementationTask, IVerificationStatus boogieStatus) {
+      var id = GetImplementationId(implementationTask.Implementation);
+      var status = StatusFromBoogieStatus(boogieStatus);
+      var lspRange = implementationTask.Implementation.tok.GetLspRange();
```
MikaelMayer, line 270:
```suggestion
      var implementationRange = implementationTask.Implementation.tok.GetLspRange();
```

In LSP, it's mostly clear that ranges are always LSP ranges, so you can name them more precisely according to their function.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -184,126 +206,103 @@ bool loadCanceled
...
+        var verificationResult = completed.Result;
+        foreach (var counterExample in verificationResult.Errors) {
+          document.Counterexamples!.Push(counterExample);
         }
 
-        return document with {
-          ImplementationViews = implementationViews.ToImmutableDictionary(),
-          CounterExamples = counterExamples.ToArray(),
-        };
+        var itDiagnostics = GetDiagnosticsFromResult(document, verificationResult);
```
MikaelMayer, line 277:
```suggestion
        var intermediateDiagnostics = GetDiagnosticsFromResult(document, verificationResult);
```

?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -338,14 +337,14 @@ bool loadCanceled
     protected virtual VerificationProgressReporter CreateVerificationProgressReporter(DafnyDocument document) {
       return new VerificationProgressReporter(
         loggerFactory.CreateLogger<VerificationProgressReporter>(),
-        document, notificationPublisher, diagnosticPublisher);
+        document, notificationPublisher, NotificationPublisher);
```
MikaelMayer, line 340: Is there a way not to have two variables with the same name but different capitalization? This sounds like there is room for improvement.

Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
keyboardDrummer: I'm not sure, but I moved this code from a few lines below, so this change is just about maintaining the existing behavior.

Source/DafnyLanguageServer/Workspace/RequestUpdatesOnUriObserver.cs
keyboardDrummer: No good reason, changed it to Array.Empty.

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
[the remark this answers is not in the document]
keyboardDrummer: I like that you referenced the LSP spec, but I feel it doesn't answer our question.

Let's look at the practical side. The Microsoft VSCode LSP client won't display anything differently if you send empty diagnostics after opening a file. However, our VSCode extension could have the status bar use the fact that first diagnostics have not arrived to determine whether it's showing "Resolving..." or not. If we do that, then I think we can really ditch the compilationStatus API.

I'll undo this change in this PR, and then we may or may not discuss it in a future PR :-)

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
MikaelMayer: > I'll address this with an updated Boogie API so no try/catch is needed: https://github.com/boogie-org/boogie/pull/592

Well, that's more than I asked for, but thanks :-)

MikaelMayer (approved):
Well done ! I think we should merge this as soon as possible, test VSCode, and verify that everything is no longer crashing anymore.

MikaelMayer (commented):
Just a minor fix.

Source/DafnyLanguageServer/Handlers/Custom/VerificationHandler.cs
```diff
@@ -43,13 +43,12 @@ public class VerificationHandler : IJsonRpcRequestHandler<VerificationParams, bo
...
     var requestPosition = request.Position;
     var tasks = GetTasksAtPosition(translatedDocument, requestPosition).ToList();
-    var failedTaskRuns = 0;
     foreach (var taskToRun in tasks) {
       var verifiedDocuments = documentLoader.Verify(translatedDocument, taskToRun, CancellationToken.None);
       documentEntry.Observe(verifiedDocuments);
     }
 
-    return failedTaskRuns != tasks.Count;
+    return !documentEntry.Idle;
```
MikaelMayer, line 51: Won't this return true immediately because verification is running in a separate thread? What is the purpose of returning a value which might or might not be true?

keyboardDrummer: Honestly I need to do more testing on the VSCode side to see what exactly needs to be returned here.

The current approximation is to return true if more notifications are incoming, and false otherwise. If false is returned, the extension won't start a 'test run' since no notifications will arrive to finish the run.

PR #2217, opened 2022-06-07
feat: Minimize unnecessary copies in the C# runtime for sequences

### LINKED ISSUE #2216: Performance issues in the C# runtime for sequences
The following programs are slow in the C# runtime:

```
const REPEAT := 1000

method NoPhysicalEqualityOnSequences(s: seq<int>) {
  for i := 0 to REPEAT {
    var b := s == s; // No physical equality makes this take linear time
  }
}

method EqualityAtLeastLinear(s: seq<int>) {
  var s0 := [0] + s;
  var s1 := [1] + s;
  for i := 0 to REPEAT {
    var b := s0 == s1; // Full sequence copy makes this linear time
  }
}

method ElementsCopiedToPerformLengthCheck(s: seq<int>) {
  var empty: seq<int> := [];
  for i := 0 to REPEAT {
    var b := [] <= s;
  }
}

method Main() {
  var s := seq(1048576, _ => 0);

  print "ElementsCopiedToPerformLengthCheck... ";
  ElementsCopiedToPerformLengthCheck(s);
  print "done.\n";

  print "EqualityAtLeastLinear... ";
  EqualityAtLeastLinear(s);
  print "done.\n";

  print "NoPhysicalEqualityOnSequences... ";
  NoPhysicalEqualityOnSequences(s);
  print "done.\n";
}
```

Each of them does what should be a small amount of work, yet here's what I measure:

```
$ time Dafny -noVerify -compile:4 slow.dfy 

Running...

ElementsCopiedToPerformLengthCheck... done.
EqualityAtLeastLinear... done.
NoPhysicalEqualityOnSequences... done.

real	1m36.694s
user	0m0.000s
sys	0m0.015s
```

cpitclaudel:
* `Source/Dafny/Compilers/Compiler-Csharp.cs`:
* `Source/DafnyRuntime/DafnyRuntime.cs`:
  Add reference to System.Collections (for `IEnumerable`)
* `Source/DafnyRuntime/DafnyRuntime.cs`:
  (`Set<T>.Equals`, `MultiSet<T>.Equals`, `Map<U, V>.Equals`): Check for
  reference equality first (before comparing `Count`).
  (`MultiSet<T>.FromSeq`): Use iterator instead of `Elements` (`Elements`
  requires a copy of the whole sequence).
  (`ISequence<T>`): Implement `IEnumerable<T>` to avoid copies.
  (`Sequence.Empty`): Use `Array<T>.Empty`.
  (`Sequence.Create`): Use an `ImmutableArray.Builder` to avoid a copy.
  (`Sequence.Update`): Copy elements twice, not three times.
  (`Sequence.EqualUntil`): Avoid copies by using `.Select(i)`.
  (`Sequence.IsPrefixOf`, `Sequence.IsProperPrefixOf`): Avoid copies by
  using `.Count` instead of `Clone().Length`.
  (`Sequence.Elements`): Rename to `CloneAsArray` to lower risk of mistakes.
  (`Sequence.GetEnumerator`): New function.
  (`Sequence.Equals`): Check for physical equality before comparing further.
  (`Sequence.ToString`): Simplify type check for `char` sequences.
  (`Sequence.ToString`): Simplify string concatenation code.
  (`Sequence.EqualUntil`): Avoid copies by using `.Select(i)`.
  (`Sequence.Take`, `ISequence.Drop`, `ISequence.Subsequence`): Simplify and
  avoid extra copies by using the right `ImmutableArray` constructor.
  (`ConcatSequence.ComputeElements`): Avoid copying subsequences by using iterator and
  avoid final copy by calling `MoveToImmutable`.
  (`Helpers.SeqFromArray`): Copy once, not twice.

Fixes #2216

### DIFF
--- a/Source/DafnyRuntime/DafnyRuntime.cs
+++ b/Source/DafnyRuntime/DafnyRuntime.cs
@@ -860,6 +860,8 @@ namespace Dafny {
   public interface ISequence<out T> : IEnumerable<T> {
     long LongCount { get; }
     int Count { get; }
+    [Obsolete("Use CloneAsArray() instead of Elements (both perform a copy).")]
+    T[] Elements { get; }
     T[] CloneAsArray();
     IEnumerable<T> UniqueElements { get; }
     T Select(ulong index);
@@ -971,6 +973,8 @@ namespace Dafny {
     // "covariant type T occurs in invariant position" error. There do not appear to be interfaces for ImmutableArray<T>
     // that resolve this.
     internal abstract ImmutableArray<T> ImmutableElements { get; }
+
+    public T[] Elements { get { return CloneAsArray(); } }
 
     public T[] CloneAsArray() {
       return ImmutableElements.ToArray();


cpitclaudel:
Benchmarks:

Null:
```
$ time Dafny.old -noVerify -compile:2 slow.dfy
real	0m2.180s
```

Before:
```
$ time Dafny.old -noVerify -compile:4 slow.dfy
ElementsCopiedToPerformLengthCheck... done.
EqualityAtLeastLinear... done.
NoPhysicalEqualityOnSequences... done.

real	1m36.694s
```

After:
```
$ time Dafny.new -noVerify -compile:4 slow.dfy
ElementsCopiedToPerformLengthCheck... done.
EqualityAtLeastLinear... done.
NoPhysicalEqualityOnSequences... done.

real	0m2.225s
```

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -1209,11 +1196,15 @@ internal class ConcatSequence<T> : Sequence<T> {
             toVisit.Push(leftBuffer);
           }
         } else {
-          var array = seq.Elements;
-          ansBuilder.AddRange(array);
+          var sq = seq as Sequence<T>;
+          if (sq != null) {
```
fabiomadge, line 1200:
```suggestion
          if (seq is Sequence<T> sq) {
```

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -1029,25 +1049,13 @@ public abstract class Sequence<T> : ISequence<T> {
...
-      var typeCheckTmp = new T[0];
-      ImmutableArray<T> elmts = ImmutableElements;
-      if (typeCheckTmp is char[]) {
-        var s = "";
-        foreach (var t in elmts) {
-          s += t.ToString();
-        }
-        return s;
+      if (typeof(T) == typeof(char)) {
+        return String.Concat(this);
```
fabiomadge, line 1053:
```suggestion
        return string.Concat(this);
```

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -991,8 +1008,11 @@ public abstract class Sequence<T> : ISequence<T> {
       return ImmutableElements[(int)index];
     }
     public bool Equals(ISequence<T> other) {
-      int n = ImmutableElements.Length;
-      return n == other.Elements.Length && EqualUntil(this, other, n);
+      int n = Count;
+      if (ReferenceEquals(this, other)) {
+        return true;
+      }
+      return n == other.Count && EqualUntil(this, other, n);
```
fabiomadge, line 1015:
```suggestion
      return ReferenceEquals(this, other) || Count == other.Count && EqualUntil(this, other, Count);
```

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -889,11 +897,11 @@ public abstract class Sequence<T> : ISequence<T> {
 
     public static ISequence<T> Create(BigInteger length, System.Func<BigInteger, T> init) {
       var len = (int)length;
```
fabiomadge, line 899: Should this be `checked` as well?

fabiomadge (approved):

Source/DafnyRuntime/DafnyRuntime.cs
cpitclaudel: Doesn't casting from biginteger to int raise an exception if it doesn't fit?

fabiomadge (dismissed):

fabiomadge (dismissed):

Issue #2222, opened 2022-06-07
Automatic Test Gen: Default Type of Type Parameters


tbean79:
This issue is a continuation of [this one](https://github.com/dafny-lang/dafny/issues/2221).

Dafny version: 3.6.0.40511

Should the above issue be fixed, then we would expect running /generateTestMode:Block on 

```
module Test {
  method IsEvenLength<K, V>(m: map<K, V>) returns (isEven: bool)
  {
    if (|m| % 2 == 0) {
      return true;
    } else {
      return false;
    }
  }
}
```

to yield

```
method test0() returns (r0:bool)  {
r0 := Test.IsEvenLength<null, null>(map[null := null, null := null, null := null]);
}
```

This is not runnable, as Dafny reports when running the test: "Error: type of 'null' is a reference type, but it is used as map<object?, object?>".  The default type of type parameters should be object? as the error suggests.  This would require maintaining a mapping from the boogie type 'Ty' to 'object?'

Issue #2221, opened 2022-06-07
Automatic Test Gen: Type Parameter Translation


tbean79:
Dafny version: 3.6.0.40511

When running /generateTestMode:Block on this Dafny method

```
module Test {
  method IsEvenLength<K, V>(m: map<K, V>) returns (isEven: bool)
  {
    if (|m| % 2 == 0) {
      return true;
    } else {
      return false;
    }
  }
}
```

the tests generated take on this form

```
method test0() returns (r0:bool)  {
r0 := Test.IsEvenLength(null, null, map[null := null, null := null, null := null]);
}
```

The first two parameters are meant to be type parameters, but instead get lumped in with the other parameters.  This occurs because the instrumented boogie implementation signature of IsEvenLength looks like this: 

```
implementation Impl$$Test.__default.IsEvenLength(Test._default.IsEvenLength$K: Ty, Test._default.IsEvenLength$V: Ty, m#0: Map Box Box) returns (isEven#0: bool, $_reverifyPost: bool)
```

Would there be any simple means of preserving the `List<TypeParameter> TypeArgs;` of the Dafny Method so that TestMethod.cs is aware of how many there are when it builds the call to the test method?  I see this as difficult because the ProgramModifier is a visitor of the BoogieAST, not the Dafny AST.

olivier-aws:
Closing the issue as it cannot be reproduced with latest Dafny release (4.9.1).

PR #2227, opened 2022-06-08
Type Parameters in Automatically Generated Tests

Dargones:
Fixes https://github.com/dafny-lang/dafny/issues/2222 and https://github.com/dafny-lang/dafny/issues/2221 by adding support for type parameters to counterexample extraction and test generation.

Test generator will replace all type variables with `int`. This should be a valid solution, I think, because all permitted type parameter characteristics apply to`int` (i.e. it is equality-supporting, auto-initializable, nonempty, and non-heap based)  

Regarding the example in the linked issues, the tool will now generate the following:  

```Dafny
include "tmp.dfy"
module tmpUnitTests {
  import Test
  method test0() returns (r0:bool)  {
    r0 := Test.IsEvenLength<int,int>(map[0 := 1, 2 := 3, 4 := 5]);
  }
  method test1() returns (r0:bool)  {
    r0 := Test.IsEvenLength<int,int>(map[]);
  }
}
```


Dargones:
@tbean79, does this seem to address the issue?

tbean79:
@Dargones Indeed, this PR solves both issues, I was able to run generated tests in both Block and Path mode.

RELEASE_NOTES.md
```diff
@@ -5,6 +5,7 @@
 - feat: Live verification diagnostics for the language server (https://github.com/dafny-lang/dafny/pull/1942)
 - fix: Prevent the language server from crashing and not responding on resolution or ghost diagnostics errors (https://github.com/dafny-lang/dafny/pull/2080)
 - fix: Correctly specify the type of the receiver parameter when translating tail-recursive member functions to C# (https://github.com/dafny-lang/dafny/pull/2205)
+- fix: Added support for type parameters in automatically generated tests
```
MikaelMayer, line 8:
```suggestion
- fix: Added support for type parameters in automatically generated tests (https://github.com/dafny-lang/dafny/pull/2227)
```

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -456,7 +465,7 @@ public class DafnyModel {
         return new DafnyModelType("?");
       }
       var fullName = GetTrueName(typeElement);
-      if (fullName != null && fullName.Length > 7) {
+      if (fullName != null && fullName.Length > 7 && fullName.Substring(0, 7).Equals("Tclass.")) {
```
MikaelMayer, line 468: Otherwise, generics could have a weird name, right?

Source/DafnyTestGeneration/TestMethod.cs
```diff
@@ -89,7 +103,11 @@ public class TestMethod {
       }
 
       List<string> elements = new();
-      switch (variable.Type.Name) {
+      var variableType = variable.Type.ReplaceTypeVariables(defaultType);
```
MikaelMayer, line 106: This does not work for higher-kinded type variables, right? But I don't think we have them in Dafny.

Source/DafnyTestGeneration/TestMethod.cs
```diff
@@ -129,10 +147,11 @@ public class TestMethod {
           return $"map[{string.Join(", ", mappingStrings)}]";
         case var arrType when new Regex("^_System.array[0-9]*\\?$").IsMatch(arrType):
           break;
+        case var _ when (!variable.Value.StartsWith("(") && variable.Value != "null"):
+          return "DATATYPES_NOT_SUPPORTED";
```
MikaelMayer, line 151: Can you please explain me why datatypes are not supported? What was happening before?

Source/DafnyTestGeneration/Utils.cs
```diff
@@ -37,15 +32,19 @@ public static class Utils {
...
-      var classPath = new DafnyModelType(boogieName
-        .Substring(0, boogieName.Length - methodName.Length - 1))
-        .InDafnyFormat().Name
-        .Split(".")
-        .Where(m => m[0] != '_');
-      var className = string.Join(".", classPath);
-      return className.Equals("") ? methodName : $"{className}.{methodName}";
+      // Get the fully qualified name of the class\module the method is defined in:
+      var classPath = boogieName
+        .Substring(0, boogieName.Length - methodName.Length - 1);
```
MikaelMayer, line 43: Are you sure that `boogieName.Length - methodName.Length - 1 >= 0`?

MikaelMayer (changes requested):
Great job solving two problems at once !
At least update the release notes, but otherwise, I have a few clarifying questions to understand your solution.

Source/DafnyTestGeneration/TestMethod.cs
Dargones: I have code that supports algebraic datatypes on my fork and can make another PR to add them but for know I just wanted to make it easier for a potential user to figure out why test generation might not work for them. Let me know if I should remove this case from this PR / make another PR specifically for datatypes.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Yes, the boogie name of a concrete Dafny type always starts with "Tclass.", whereas a generic type does not

Dargones:
Thank you, @MikaelMayer! Let me know if there is anything else.

Source/DafnyTestGeneration/TestMethod.cs
MikaelMayer: Ok thanks, It's ok to keep this case in this PR I think. I'm happy to have a look at your other PR about datatypes.

MikaelMayer (approved):
Great work, thanks!

PR #2233, opened 2022-06-10
Do nightly releases

keyboardDrummer:
### Changes
As part of the nightly workflow, create a Dafny release.

- A tag `nightly` is automatically updated to the head of the master branch during the nightly workflow
- A single GitHub release coupled to the nightly tag is updated (or created if it doesn't exist) and given the name `nightly-YYYY-MM-DD-<shortsha>`. Example: https://github.com/keyboardDrummer/dafny/releases/tag/nightly
- The GitHub nightly release is a pre-release
- There's a NuGet package for the nightly release, which is a NuGet prerelease because it has as a version containing a `-`: `<version>-<date>-<shorthash>`. The variable `<version>` is set using the currently configured value in the .csproj files, so that would be the last stable release version.

### Testing
- Manually tested the nightly workflow: https://github.com/keyboardDrummer/dafny/actions/runs/2596837241
- DID NOT test the draft release workflow, since I didn't want to publish a new stable version to NuGet.


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  deep-integration-tests:
+    uses: ./.github/workflows/integration-tests-reusable.yml
+    with:
+      all_platforms: true
+      num_shards: 5
+  publish-release:
+    uses: ./.github/workflows/publish-release-reusable.yml
+    with:
+      sha: ${{ refs/heads/master }}
```
keyboardDrummer, line 23: Not sure if this works, might have to do:

```
    steps:
      - id: get-sha
        run: |
          echo ::set-output name=sha::$( curl -u "u:${{github.token}}" https://api.github.com/repos/some-owner/some-repo/git/ref/heads/master | jq .object.sha | tr -d '"' )
      - uses: actions/checkout@v2
        with:
          ref: ${{ steps.get-sha.outputs.sha }}
```

cpitclaudel (commented):
This looks really nice!  I have a few questions:

- Does this only release if the deep-tests workflow succeeds?  If so, maybe we also need to update release instructions to remove the manual deep-tests step?
- Will this pollute the list of releases?  If we anticipate this being an issue, we could consider keeping only a few nightly releases around (https://github.com/marketplace/actions/deploy-nightly)

.github/workflows/publish-release-reusable.yml
```diff
@@ -0,0 +1,89 @@
...
+      release_notes:
+        required: true
+        type: string
+
+env:
+  dotnet-version: 6.0.x # SDK Version for building Dafny
+  
+jobs:
+  check-deep-tests:
+    uses: ./.github/workflows/check-deep-tests-reusable.yml
```
robin-aws, line 24: I'm happy with the rename of `deep-tests.yml` to `nightly.yml`, but you'll also need to rename and fix the action that explicitly checks for the `deep-tests.yml` workflow file name: https://github.com/dafny-lang/dafny/blob/master/.github/workflows/check-deep-tests-reusable.yml#L25

robin-aws (changes requested):
We were 95% of the way already weren't we :)

.github/workflows/nightly.yml
keyboardDrummer: > You should be able to just use `${{ github.sha }}` as in https://github.com/dafny-lang/dafny/blob/master/.github/workflows/publish-release.yml#L16

How would that be defined for a cronjob? It's not a sha that's triggering the workflow.

keyboardDrummer:
> * Will this pollute the list of releases?  If we anticipate this being an issue, we could consider keeping only a few nightly releases around (https://github.com/marketplace/actions/deploy-nightly)

That sounds nice, but that package stores the different releases as separate assert bundles inside a single GH release, which I think we don't want. I've opted to only store a single nightly release for implementation simplicity.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,47 @@
...
+    with:
+      all_platforms: true
+      num_shards: 5
+
+  determine-vars:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Get short sha
+        id: short-sha
+        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
```
robin-aws, line 26: Don't you have to `checkout` first for this to work?

.github/workflows/nightly.yml
robin-aws: Ah right. :P

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,47 @@
...
+      name: nightly-${{ steps.date.outputs.date }}-${{ steps.short-sha.outputs.sha_short }}
+
+  publish-release:
+    uses: ./.github/workflows/publish-release-reusable.yml
+    with:
+      sha: ${{ refs/heads/master }}
+      name: ${{ needs.determine-vars.outputs.name }}
+      tag_name: nightly
+      prerelease: true
+      release_nuget: false
```
robin-aws, line 44: Fine for v1, but it would be really nice to publish these to nuget as well, so we can move towards that being the only way to consume Dafny. See https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages.

robin-aws (commented):
Have you been able to dry-run these on your fork? That's how I tested the initial versions.

.github/workflows/publish-release-reusable.yml
```diff
@@ -0,0 +1,118 @@
+name: Publish Draft Release
```
robin-aws, line 1:
```suggestion
name: Publish Release
```

.github/workflows/publish-release-reusable.yml
```diff
@@ -0,0 +1,118 @@
...
+          dafny/Package/dafny-${{ inputs.name }}*
+          dafny/docs/DafnyRef/DafnyRef.pdf
+        fail_on_unmatched_files: true
+
+    - name: Create release NuGet package (for uploading)
+      if: ${{ !inputs.prerelease }}
+      run: dotnet pack --no-build dafny/Source/Dafny.sln
+    - name: Create prerelease NuGet package (for uploading)
+      if: ${{ inputs.prerelease }}
+      run: dotnet pack --version-suffix ${{ inputs.name }} --no-build dafny/Source/Dafny.sln
```
robin-aws, line 115: It wasn't obvious to me that Nuget automatically considers any version with a suffix as a prerelease, so probably worth a comment here. Very nice that it's this simple though!

robin-aws (commented):
Looks awesome! Just a couple of nits.

robin-aws (approved):

PR #2240, opened 2022-06-13
Fix MergeOrdered and empty observable

keyboardDrummer:
Fixes a bug that would ignore subsequent inner observables after MergeOrdered received an empty inner observable, which in Dafny would cause notifications to stop occurring after saving a file that did not have verify on save turned on.


MikaelMayer (approved):
Fantastic bug catching and testing at two different levels. Approved.

PR #2245, opened 2022-06-13
Use Dafny names instead of Boogie names in log output

### LINKED ISSUE #1734: Improve display name of "test results" when using `/verificationLogger:...`

atomb:
Integrate support for using Dafny names in Boogie through the `verboseName` attribute added in [this PR](https://github.com/boogie-org/boogie/pull/586). As a result, Dafny names now appear instead of Boogie names in the output of `/trace` and `/verificationLogger`, and are used as the argument to `/proc`.

Existing tests have been updated to reflect the new output.

**Note:** This implementation broke the assumptions of some code in the test generator, and therefore this PR includes a fix, which originated in [this other PR](https://github.com/atomb/dafny/pull/2). It may be easier to review that PR first.

Fixes #1734


Source/Dafny/Verifier/Translator.cs
```diff
@@ -7919,6 +7932,34 @@ enum MethodTranslationKind { SpecWellformedness, Call, CoCall, Implementation, O
...
+      Contract.Requires(fullName != null);
+      switch (kind) {
+        case MethodTranslationKind.SpecWellformedness:
+          return fullName + " (well-formedness)";
+        case MethodTranslationKind.Call:
+          return fullName + " (call)";
+        case MethodTranslationKind.CoCall:
+          return fullName + " (co-call)";
+        case MethodTranslationKind.Implementation:
+          return fullName + " (implementation correctness)";
```
MikaelMayer, line 7945:
```suggestion
          return fullName + " (correctness)";
```

MikaelMayer (changes requested):
Great job  thanks !

Source/Dafny/Verifier/Translator.ClassMembers.cs
```diff
@@ -1476,6 +1481,7 @@ public partial class Translator {
 
       var name = MethodName(m, kind);
       var proc = new Boogie.Procedure(m.tok, name, new List<Boogie.TypeVariable>(), inParams, outParams, req, mod, ens, etran.TrAttributes(m.Attributes, null));
+      AddVerboseName(proc, m.FullDafnyName, kind);
```
MikaelMayer, line 1484: You couldn't refactor this one?

MikaelMayer (dismissed):
Great addition.

Source/Dafny/Verifier/Translator.ClassMembers.cs
atomb: The benefit in this case (for `Procedure` instead of `Implementation`) is less clear, because there are only two instances of it and they differ in many dimensions.

However, reading this made me realize that there were some instances of the previous refactoring that I missed. I'll include those, too.

MikaelMayer (dismissed):

Source/Dafny/Verifier/Translator.ClassMembers.cs
```diff
@@ -714,10 +714,8 @@ public partial class Translator {
       if (EmitImplementation(m.Attributes)) {
         // emit impl only when there are proof obligations.
         QKeyValue kv = etran.TrAttributes(m.Attributes, null);
-        Boogie.Implementation impl = new Boogie.Implementation(m.tok, proc.Name,
-          new List<Boogie.TypeVariable>(), inParams, outParams,
-          localVariables, stmts, kv);
-        sink.AddTopLevelDeclaration(impl);
+        Boogie.Implementation impl = AddImplementationWithVerboseName(m.tok, proc,
```
cpitclaudel, line 717: Why is it called `VerboseName`? (it's the name that's displayed to the user, right?  I think of "verbose" as the thing that has all the information but isn't very readable)

Source/Dafny/Verifier/Translator.ClassMembers.cs
```diff
@@ -893,9 +891,11 @@ public partial class Translator {
       };
       var ens = new List<Boogie.Ensures>();
 
-      var proc = new Boogie.Procedure(f.tok, "OverrideCheck$$" + f.FullSanitizedName, new List<Boogie.TypeVariable>(),
+      var name = MethodName(f, MethodTranslationKind.OverrideCheck);
+      var proc = new Boogie.Procedure(f.tok, name, new List<Boogie.TypeVariable>(),
         Util.Concat(Util.Concat(typeInParams, inParams_Heap), inParams), outParams,
         req, mod, ens, etran.TrAttributes(f.Attributes, null));
+      AddVerboseName(proc, f.FullDafnyName, MethodTranslationKind.OverrideCheck);
```
cpitclaudel, line 898: I wonder if we could merge the calls to `MethodName` and `AddVerboseName`, since it seems that a call to the first must always be followed by a call to the second

cpitclaudel: (Especially given that we repeat the `MethodTranslationKind`

Source/Dafny/Verifier/Translator.cs
```diff
@@ -7919,6 +7924,34 @@ enum MethodTranslationKind { SpecWellformedness, Call, CoCall, Implementation, O
...
+        case MethodTranslationKind.SpecWellformedness:
+          return fullName + " (well-formedness)";
+        case MethodTranslationKind.Call:
+          return fullName + " (call)";
+        case MethodTranslationKind.CoCall:
+          return fullName + " (co-call)";
+        case MethodTranslationKind.Implementation:
+          return fullName + " (correctness)";
+        case MethodTranslationKind.OverrideCheck:
+          return fullName + " (override check)";
```
cpitclaudel, line 7939: I would refactor this as a (static const) dictionary from MethodTranslationKind to string and then
```
return $"{fullname} ({kindDescription[kind]})"
```

Source/Dafny/Verifier/Translator.cs
```diff
@@ -7919,6 +7924,34 @@ enum MethodTranslationKind { SpecWellformedness, Call, CoCall, Implementation, O
...
+    }
+
+    private static void AddVerboseName(Bpl.NamedDeclaration boogieDecl, string dafnyName, MethodTranslationKind kind) {
+      var name = MethodVerboseName(dafnyName, kind);
+      boogieDecl.AddAttribute("verboseName", new object[] { name });
+    }
+
+    private static void CopyVerboseName(Bpl.NamedDeclaration targetDecl, Bpl.NamedDeclaration sourceDecl) {
+      targetDecl.AddAttribute("verboseName", new object[] { sourceDecl.VerboseName });
+    }
```
cpitclaudel, line 7953: Use AddVerboseName here?  After that I would think you could even directly inline the call.

Source/DafnyPipeline.Test/IntraMethodVerificationStability.cs
```diff
@@ -190,7 +190,7 @@ class C {
     public async void EqualProverLogWhenAddingUnrelatedProgram() {
 
       var options = DafnyOptions.Create();
-      options.ProcsToCheck.Add("*SomeMethod");
+      options.ProcsToCheck.Add("*SomeMethod *");
```
cpitclaudel, line 193: What is the meaning of this `*`?

Source/DafnyTestGeneration/PathBasedModifier.cs
```diff
@@ -10,7 +10,7 @@ namespace DafnyTestGeneration {
   public class PathBasedModifier : ProgramModifier {
 
     // prefix given to variables indicating whether or not a block was visited
-    private const string blockVarNamePrefix = "$$visited$$_";
+    private const string BlockVarNamePrefix = "$$visited$$_";
```
cpitclaudel, line 13: is this uppercase because it's a const?

Source/DafnyTestGeneration/ProgramModifier.cs
```diff
@@ -268,7 +268,7 @@ private class AnnotationVisitor : StandardVisitor {
         node.Blocks[0].cmds.Insert(0, GetAssumeCmd(types));
 
         // record parameter values:
-        var values = new List<string> { "\"Impl\"", $"\"{implName}\"" };
+        var values = new List<string> { "\"Impl\"", $"\"{node.VerboseName.Split(" ")[0]}\"" };
```
cpitclaudel, line 271: I don't understand this part

Test/allocated1/dafny0/Superposition.dfy.expect
```diff
@@ -1,18 +1,18 @@
 
-Verifying Impl$$M0.C.M ...
+Verifying M0.C.M (implementation correctness) ...
   [4 proof obligations]  verified
 
-Verifying CheckWellformed$$M0.C.Q ...
+Verifying M0.C.Q (well-formedness) ...
```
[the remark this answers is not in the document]
cpitclaudel: 👍

Test/dafny0/snapshots/Snapshots0.run.dfy.expect
```diff
@@ -2,7 +2,7 @@ Processing command (at Snapshots0.v0.dfy(4,10)) assert Lit(false);
   >>> DoNothingToAssert
 
 Dafny program verifier finished with 1 verified, 0 errors
-Processing call to procedure Call$$_module.__default.bar in implementation Impl$$_module.__default.foo (at Snapshots0.v1.dfy(3,6)):
+Processing call to procedure bar (call) in implementation foo (correctness) (at Snapshots0.v1.dfy(3,6)):
```
cpitclaudel, line 5: I find the parentheses a bit confusing in this example.  Might be a problem for later, though.

cpitclaudel (commented):
Nice work, and I'm glad to see how few changes are actually needed on the Dafny side.  I left some stylistic comments.  It might be nice to have one test for this feature specifically (I was surprised to see that most existing tests seemed accidental in connection with this feature — they are tests from the legacy server and the cache)

Source/Dafny/Verifier/Translator.ClassMembers.cs
atomb: I originally called it `DisplayName` but changed it after [this conversation](https://github.com/boogie-org/boogie/pull/586#discussion_r898182589). I'm not especially attached to any particular name.

Source/DafnyTestGeneration/PathBasedModifier.cs
atomb: Yes, I think so. This originally came from [the other PR](https://github.com/atomb/dafny/pull/2), but that's what I expect was the motivation.

Source/DafnyTestGeneration/ProgramModifier.cs
atomb: Also from the other PR, but an earlier version. You can see a discussion of it [here](https://github.com/atomb/dafny/pull/1#discussion_r900251083).

Source/Dafny/Verifier/Translator.cs
atomb: I don't think that'd work, because `AddVerboseName` calls `MethodVerboseName`, which we don't want to do here. But I think there is a refactoring that could make sense here. I'll do it and you can let me know what you think.

Source/Dafny/Verifier/Translator.ClassMembers.cs
cpitclaudel: OK, I don't have strong feelings either.

keyboardDrummer: The name we came up with after the Boogie PR and liked most was `Description`

Source/DafnyTestGeneration/ProgramModifier.cs
Dargones: The idea is to get the fully qualified name of a Dafny method from Boogie. Doing this allows getting rid of [this method](https://github.com/dafny-lang/dafny/blob/479bdda4f676bf12d743c15e03db1c43dc05d9f6/Source/DafnyTestGeneration/Utils.cs#L34-L47) so I think it should make things much more stable.

Test/dafny0/snapshots/Snapshots0.run.dfy.expect
atomb: No, the module path is included when explicit. (See the changes in `Superposition.dfy.expect`, for example.)

Source/Dafny/Verifier/Translator.ClassMembers.cs
atomb: It's a little less clear how to abstract this cleanly. The result of `MethodName` is used only in the call to `new Bpl.Procedure`, but not every call site for `new Bpl.Procedure` has a preceding `MethodName` call.

Source/Dafny/Verifier/Translator.ClassMembers.cs
atomb: And I'd be open to a future PR changing it to that. We'd have to change it in Boogie first, though.

cpitclaudel (approved):
Great work, thanks.

I'm approving this as-is, but I'd like to suggest one more change.  I think it would probably be possible to go one step further and change `VerboseName` in Boogie to be an arbitrary object (we just need it to support `ToString`.  Then, based on this, we could use a richer object on the Dafny side (probably a record containing the name and the method kind), and we could chose what to print depending on the context.  Just a suggestion ^^

atomb:
> Great work, thanks.
> 
> I'm approving this as-is, but I'd like to suggest one more change. I think it would probably be possible to go one step further and change `VerboseName` in Boogie to be an arbitrary object (we just need it to support `ToString`. Then, based on this, we could use a richer object on the Dafny side (probably a record containing the name and the method kind), and we could chose what to print depending on the context. Just a suggestion ^^

That would match the approach of the `ProofObligationDescription` type, which seems nice. One downside would be that it would be harder to specify in a Boogie source file. I'm not sure if that's a critical requirement, though.

cpitclaudel:
> One downside would be that it would be harder to specify in a Boogie source file. I'm not sure if that's a critical requirement, though.

In a Boogie source file we would fall back to using plain strings, which would be fine, I think?

PR #2249, opened 2022-06-14
fix: Errors in translation no longer result in a server crash

MikaelMayer:
This fix prevent any errors in the verifier/translator to permanently crash the language server.

I added a unit test, `DocumentDatabaseTest`, to only test that, if the `TextDocumentLoader.PrepareVerificationTasksAsync` throws an exception, then an error is reported and the resolved document is returned.

### DIFF
--- a/Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
+++ b/Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
@@ -8,7 +8,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reactive.Linq;
 using System.Reactive.Threading.Tasks;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Boogie;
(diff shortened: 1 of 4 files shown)


Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -214,24 +213,32 @@ IRelocator relocator
...
+          ImplementationViews = withVerificationTasks.ImplementationViews!.ToDictionary(
+            kv => kv.Key,
+            kv =>
+              kv.Value with {
+                Diagnostics = resolvedDocument.ImplementationViews.GetValueOrDefault(kv.Key)?.Diagnostics ??
+                              kv.Value.Diagnostics
+              }
+          ),
+        };
+      } catch (Exception e) {
```
keyboardDrummer, line 237: Since this is an `async` method, any exceptions are already caught in the `Task`. I think it might be better to handle those exceptions at a higher level, where the task is handled.

Where we call `translatedDocument.ToObservableSkipCancelled()`, any exception in the Task is translated to an OnError event in the observable, and that may subsequently block all further notifications, since the `Subject` inside `MergeOrdered` will, like any `Subject`, stop sending OnNext events after it has sent an OnError event. Let's fix that!

MikaelMayer: > Where we call `translatedDocument.ToObservableSkipCancelled()`, any exception in the Task

I don't think it's true for "any exception", only "OperationCancelledException".

keyboardDrummer: I'm suggesting to do this: https://github.com/dafny-lang/dafny/pull/2254/commits/24d6f13d6a3a9d08c577771d1b87dedd053f0c42

Without that change, any exception other than `OperationCancelledException `, that occurs in Tasks on which ToObservableSkipCancelled is called, will corrupt the language server so it doesn't send notifications any more.

MikaelMayer: I like this commit, that makes sense.

PR #2252, opened 2022-06-14
Fix concurrency issue in Translator

keyboardDrummer:
Fix concurrency issue in Translator


MikaelMayer (approved):
Thanks ! That should do it.

PR #2266, opened 2022-06-16
fix: Fix naming of formals in Translator

### LINKED ISSUE #2263: Dafny translates a file from the standard library to a syntacticly invalid Boogie code
Version of Dafny (current HEAD): `92736033599c34dc3c5cd9f0f46e3dccbd5a3210`
Version of Standard Library (also current HEAD): `44f5891e7b5b21a5f13f139b22fb5559b46de36f`

How to reproduce:

1. Generate Boogie code (note that Dafny successfully verifies the file): `dotnet dafny-head/Binaries/Dafny.dll libraries/src/Collections/Sequences/Seq.dfy /print:Seq.bpl /compile:0`

1. Attempt to verify generated code with Boogie 3.15 (version currently used by Dafny): `dotnet boogie-2.15.0/Source/BoogieDriver/bin/Debug/net6.0/BoogieDriver.dll /infer:j /proverOpt:O:auto_config=false /proverOpt:O:type_check=true /proverOpt:O:smt.case_split=3 /proverOpt:O:smt.qi.eager_threshold=100 /proverOpt:O:smt.delay_units=true /proverOpt:O:smt.arith.solver=2 Seq.bpl`

Error message:

```
Seq.bpl(20222,43): error: the type of the last parameter is unspecified
1 parse errors detected in Seq.bpl
```

Here is the line mentioned in the error message: 
`function Reads1#Handle(#$T0: Ty, #$R: Ty, HandleType) : HandleType;`

Dargones:
Fixes #2263, which is due to inconsistent naming of formal parameters in automatically generated function handles.

Explanation: The syntax of `MkTyParamFormals` was changed with commit https://github.com/dafny-lang/dafny/commit/81cf582cce3fdeb482ed2746da6a812df8023514 from 

```cs
List<Bpl.Variable> MkTyParamFormals(List<TypeParameter> args, bool named = true)
```
 to

```cs
List<Variable> MkTyParamFormals(List<TypeParameter> args, bool includeWhereClause, bool named = true)
```

But the call to the method [on this line](https://github.com/dafny-lang/dafny/blob/fcc912f9b117115cbc8b2273800397e88d3c1e53/Source/Dafny/Verifier/Translator.cs#L6855) was left unchanged, so the method is now being called with `named` field set to the default `true`. [Elsewhere](https://github.com/dafny-lang/dafny/blob/fcc912f9b117115cbc8b2273800397e88d3c1e53/Source/Dafny/Verifier/Translator.cs#L6882) the formal parameters of the a function handle don't get a name. The result is a mix-match of named and unnamed formals that causes the bug reported in the linked issue.

I don't completely understand the meaning of `includeWhereClause` in this context but I think `false` is the correct value for this call? 


robin-aws:
Thanks @Dargones! This LGTM in general, although I also don't fully understand the `includeWhereClause` parameter so I'm not 100% sure `false` is correct.

In the meantime, can you add a new test case to expose the issue somehow? I'd like to get to the bottom of why this only shows up when you manually verify the Boogie code separately - it might point to the Dafny pipeline failing to pass an option to catch this error more strictly.

Dargones:
Thank you, @robin-aws. I have added a test case and it succeeds when I run it with `lit` directly but fails here because, I think, it is executed via `IntergrationTests.csproj`. It might be possible to add `BoogieDriver` as a dependency to `IntegrationTests.csproj` but I am not completely sure, so I am wondering what the right way to do this would be. Is there a directory for tests that are executed directly with `lit`? 

This test also adds additional overhead of having to build Boogie but I think it could be useful for more extensive testing of the Dafny -> Boogie file -> Verify with Boogie pipeline.

More about the bug: the reason the verification succeeds when running Dafny alone is because all parameters have names, one of these names happens to be an empty string. However, when this empty string name is printed to file, Boogie cannot parse it back.

More specifically, the bug seems to be triggered by adding a reads clause for a function type parameter, hence the test case:

```Dafny
function Test(f: (int ~> bool)): (b:bool) reads f.reads { true }
```

As for the `includeWhereClause` parameter, I think `false` is the correct value, since I couldn't find `where` clauses on function parameters in any of the

robin-aws:
Great, thanks for the explanation. I fully agree being able to automatically test the Dafny -> Boogie file -> Verify with Boogie workflow is extremely useful, given many contributors will use that to debug issues with the translator (I know I have :).

Rather than building Boogie from source we could just install the corresponding release of the `boogie` dotnet tool from nuget in our CI. In fact the easiest solution is just to add it to our `dotnet-tools.json`.

> As for the `includeWhereClause` parameter, I think `false` is the correct value, since I couldn't find `where` clauses on function parameters in any of the

I think you accidentally a few words :)

Dargones:
I believe I have been able to fix the testing process (pending integration tests rerun, which was successful on my fork). @robin-aws - let me know if there is anything else I could do to merge this in!

atomb (dismissed):
I love the fix/test ratio on this one. ;)

Dargones:
Thank you for approving, @atomb! I have synced the PR with master so it should be good to go

PR #2270, opened 2022-06-19
chore: Enable another batch of tests supported by the Python compiler

fabiomadge:


Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -1443,28 +1456,11 @@ protected class ClassWriter : IClassWriter {
...
       }
-      if (runAfterCompile) {
-        Contract.Assert(callToMain != null); // this is part of the contract of CompileTargetProgram
-        // Since the program is to be run soon, nothing further is done here. Any compilation errors (that is, any errors
-        // in the emitted program--this should never happen if the compiler itself is correct) will be reported as 'python'
-        // will run the program.
-        return true;
-      } else {
-        // compile now
-        return SendToNewPythonProcess(dafnyProgramName, targetProgramText, null, targetFilename, otherFileNames,
```
robin-aws, line 1454: Isn't removing this weakening the assurance of our compiler a bit? i.e. if we emit syntactically invalid Python code in some case somehow, before we would get an error when passing `/compile:1`, but now we won't.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -1477,9 +1473,9 @@ protected class ClassWriter : IClassWriter {
 
       try {
         using var pythonProcess = Process.Start(psi);
-        pythonProcess.StandardInput.Close();
-        pythonProcess.WaitForExit();
-        return pythonProcess.ExitCode == 0;
+        pythonProcess?.StandardInput.Close();
```
robin-aws, line 1476: Is this a case of `Process.Start` declaring a return type of `Process?` when it should be `Process`? If so, it would be good to explicitly cast to a non-null type once instead.

Otherwise, I would think we'd want to propagate an error if we didn't actually start the process.

Test/comp/DowncastClone.dfy
```diff
@@ -3,8 +3,9 @@
 // RUN: %dafny /noVerify /compile:4 /compileTarget:js "%s" >> "%t"
 // RUN: %dafny /noVerify /compile:4 /compileTarget:go "%s" >> "%t"
 // RUN: %dafny_0 /noVerify /compile:4 /compileTarget:java "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:py "%s" >> "%t"
```
robin-aws, line 6: Can we switch this one and TailRecursion.dfy to `%testdafny` instead?

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: Yes, but we don't have a good separation between run and compile, so I'd rather not "compile" when no run is requested.

Test/comp/DowncastClone.dfy
fabiomadge: Neither works because of a bug in `%testdafny`.

Source/Dafny/Compilers/Compiler-python.cs
robin-aws: I agree the comment "compile now" from other compilers is misleading. I think of it as a linter, to ensure the code is at least valid, even if we aren't going to run it. If you don't want to invoke the target compiler/runtime like this, then you do `/compile:0 /spillTargetCode:1` instead.

You could make the argument that this API is confusing for non-compiled languages, perhaps successfully :), but even then we should fix/change that consistently at the level of `Compiler` and the meaning of the CLI flags. For now we should keep Python consistent.

Test/comp/DowncastClone.dfy
robin-aws: Thanks for trying anyway! I'll address the issue with the `Feature` code as part of completing #1770.

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: It's not just the comment, it's our [docs](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-controlling-compilation) as well. It's not just a linter. There is no limit to how long execution of the generated program runs for, breaking any reasonable expectation of compiling.

fabiomadge: I would prefer a difference between 1 and 3.

fabiomadge: In JS we can get away with not writing main, but I don't think we can do that for Python.

robin-aws: Okay, it's a good point that the docs don't state that we evaluate the program with the target interpreter for JS.

Why don't you pull this change into a separate PR that also fixes JavaScript? We can discuss whether that's a good (possibly breaking) change independently.

fabiomadge: JS is different, as I was saying. Your requested change would mean ripping out the compile tests as well. They run the program twice without my changes.

cpitclaudel: I might be widely off the mark, but why don't we just compile without executing when execution isn't requested?

https://stackoverflow.com/questions/2042426/compile-python-py-file-without-executing

robin-aws: I don't want to block this any further so I cut https://github.com/dafny-lang/dafny/issues/2567 to track this (and implicitly replied to you in there :) )

robin-aws (approved):
LGTM! (aside for the test failure that showed up on the merge from master)

I've cut https://github.com/dafny-lang/dafny/issues/2567 to track the discussion on `/compile:3` separately, so I'm happy for now.

keyboardDrummer (approved):

PR #2282, opened 2022-06-21
chore: Test for latest version on NuGet

fabiomadge:


atomb (approved):
Thanks for doing this! It's sad that we need to account for the "Fetch latest release version string" failing, but it's definitely better than what we had before.

PR #2283, opened 2022-06-22
Reduce size of generated tuple type definitions

keyboardDrummer:
Since this PR contains refactorings, please review using the commits. The refactoring commits are all at the start so it should be easy to skip just those and review the remaining changes as one big diff.

### Changes
- Refactoring: move TupleTypeDecl from DafnyAST.cs to its own file
- Change the CompileName TupleTypeDecl so it ignores any ghost arguments of the tuple
- Change SinglePassCompiler so datatypes with the same CompileName only generate one target language type.
- Change `BuiltIns` so tuples with exactly two non-ghost arguments are not compiled

### Testing
This PR only affects the size of the generated code, which we currently do not test, so there are no test changes. The goal of this PR is to prepare for another PR where we no longer generate tuples but instead only support them to some fixes size (say 10) by putting them in the runtime. With this PR, we only need 10 predefined tuples in the runtime, instead of 2^10, and we support more of them (all tuples with 10 non-ghost arguments, instead of all tuples with 10 arguments in total)


Source/Dafny/AST/TupleTypeDecl.cs
```diff
@@ -79,6 +79,5 @@ private TupleTypeDecl(ModuleDefinition systemModule, List<TypeParameter> typeArg
 
   public override string SanitizedName =>
     sanitizedName ??= $"Tuple{BuiltIns.ArgumentGhostnessToString(ArgumentGhostness)}";
-  public override string CompileName =>
-    compileName ??= SanitizedName;
```
cpitclaudel: Why do we want this mismatch between the CompileName and the SanitizedName?

cpitclaudel:
> Refactoring: move TupleTypeDecl from DafnyAST.cs to its own file

I would prefer to avoid this part, since DafnyAst.cs is an API that we expose to Dafny-in-Dafny.

> Change BuiltIns so tuples with exactly two non-ghost arguments are not compiled

Where is this part?

cpitclaudel:
> The goal of this PR is to prepare for another PR where we no longer generate tuples but instead only support them to some fixes size (say 10) by putting them in the runtime.

Can you say more about the motivation for this?  Tuples are regular datatypes with an anonymous constructor, so it seems odd to special-case them.

> With this PR, we only need 10 predefined tuples in the runtime, instead of 2^10, and we support more of them (all tuples with 10 non-ghost arguments, instead of all tuples with 10 arguments in total)

I don't understand this part.  Where does the 2^10 part come from?  (And more importantly, which problem are we solving?)

keyboardDrummer:
> Why do we want this mismatch between the CompileName and the SanitizedName?

AFAIK SanitizedName is used for the verification back-end, in which the ghost arguments in tuples are relevant, while CompileName is only for the code generation back-ends, where only the non-ghost tuple arguments are relevant.

>> Refactoring: move TupleTypeDecl from DafnyAST.cs to its own file
> I would prefer to avoid this part, since DafnyAst.cs is an API that we expose to Dafny-in-Dafny.

My Rider has a degraded UX in DafnyAST.cs. Most IDE suggestions are disabled entirely and the IDE generally slows down. What's the specific issue for DinD?

>> Change BuiltIns so tuples with exactly two non-ghost arguments are not compiled
> Where is this part?

That's this change:
```
-        var attributes = dims == 2 && !argumentGhostness.Contains(true) ? DontCompile() : null;
+        var attributes = argumentGhostness.Count(x => !x) == 2 ? DontCompile() : null;
```

> Can you say more about the motivation for this? Tuples are regular datatypes with an anonymous constructor, so it seems odd to special-case them.

What's special about tuples is that they're Dafny native types, like sets and sequences, except there's an infinite about of different tuple type constructors, for each combination of ghost and non-ghost argument types.

What special casing are you referring to?

> I don't understand this part. Where does the 2^10 part come from? (And more importantly, which problem are we solving?)

Dafny tuple types can have each argument be ghost or non-ghost, so that means there's 2^10 different tuple types of dimension 10.

cpitclaudel:
> My Rider has a degraded UX in DafnyAST.cs. Most IDE suggestions are disabled entirely and the IDE generally slows down. What's the specific issue for DinD?

We export only DafnyAST.cs; if there are part of it that depend on other definitions, we have to model them by hand.

Thanks for the other clarifications!

keyboardDrummer:
> Although maybe `TupleTypeDecl` shouldn't exist at all, and it should only be a regular datatype declaration?

Maybe, but that seems out of scope for this PR.

Tuples have a cool feature that I don't think we have for regular datatypes, since they may contain both ghost and non-ghost values, while the tuple as a whole stays non-ghost. Maybe we can generalise this. For an instantiated datatype, you could get the de-ghosted version where all ghost fields and methods have been removed, so you could have a runtime instance of this datatype. Not sure how useful that would be though.

>> Refactoring: move TupleTypeDecl from DafnyAST.cs to its own file
> I would prefer to avoid this part, since DafnyAst.cs is an API that we expose to Dafny-in-Dafny.

Could you add a piece of Dafny-inDafny to `dafny-lang/dafny` so that this API usage is tested? I can't reliably fix what I'm breaking without a test.

>> Can we make it so DinD takes a list of files that it exports, instead of just DafnyAST.cs ?
> Sure, we'd need to patch Source/AutoExtern (which is already merged) to support that.

Are you suggesting to do that as part of or before this PR?

cpitclaudel:
> Could you add a piece of Dafny-inDafny to dafny-lang/dafny so that this API usage is tested? I can't reliably fix what I'm breaking without a test.

I will do that.

> Were you suggesting to do that as part of or before this PR? Anyways here is a PR to add that: https://github.com/dafny-lang/dafny/pull/2292

Thanks!

cpitclaudel (approved):

Issue #1729, opened 2022-01-19
Compiler.ReportError uses hardcoded comment syntax


cpitclaudel:
The following method probably shouldn't use `/* … */` comment syntax unconditionally:

```csharp
    protected static void ReportError(ErrorReporter reporter, Bpl.IToken tok, string msg, ConcreteSyntaxTree/*?*/ wr, params object[] args) {
      Contract.Requires(msg != null);
      Contract.Requires(args != null);

      reporter.Error(MessageSource.Compiler, tok, msg, args);
      wr?.WriteLine("/* {0} */", string.Format("Compilation error: " + msg, args));
    }
```

Reported as part of https://github.com/dafny-lang/dafny/pull/1708

PR #2284, opened 2022-06-22
Move tuples to runtime instead of generating them

keyboardDrummer:
Since this PR contains refactorings, please review using the commits. The refactoring commits are all at the start so it should be easy to skip just those and review the remaining changes as one big diff.

### Changes
- Stop generating tuple type definitions on the fly. Instead put these definitions in the Dafny runtimes for various target languages, up to tuples of size 20. Moving tuple types into the runtime has the benefit that independently compiled Dafny programs can be further compiled in the target language toolchain without running into duplicate type definition errors.

### Testing
- Modify comp/Dt.dfy so it uses the largest tuple size that C# and Java support (size 20)
- Add comp/TupleMaximumSize.dfy that tests the error messages coming from the Java and C# compilers when going over the maximum tuple size


cpitclaudel:
I'm not sure this is the right way to go (I don't know of any other functional language that handles tuples like that). 

- Tuple types are not the only type that we generate on the fly: there's also function types, which we already translate to the corresponding native types.  Maybe we could do the same with tuples?  So in C# we would translate Dafny's tuple type to `Tuple<>`, for example
- Alternatively, we could do what other languages do, and treat `(a, b, c)` as `(a, (b, c))` (but this might not work well with our `.1`, `.2`, etc accessors)

keyboardDrummer:
> I'm not sure this is the right way to go (I don't know of any other functional language that handles tuples like that).
> 
> * Tuple types are not the only type that we generate on the fly: there's also function types, which we already translate to the corresponding native types.  Maybe we could do the same with tuples?  So in C# we would translate Dafny's tuple type to `Tuple<>`, for example
> * Alternatively, we could do what other languages do, and treat `(a, b, c)` as `(a, (b, c))` (but this might not work well with our `.1`, `.2`, etc accessors)

I think for C# we should use native tuple types, but I don't think we need to do that in this PR. For Java, I don't think there are native tuples or tuples in the standard library, so the approach of Dafny putting them in the Java runtime seems reasonable. For tuples above the maximum size in Java, Dafny could switch to using untyped arrays or tuples containing tuples (as you suggest), but also I don't think we need to do that in this PR.

The goal of this PR though is to stop generating tuple type definitions as that causes issues in diamond dependency scenarios, and having a maximum tuple size of 20 for C# and Java seems doesn't seem like it would cause an immediate issues.

cpitclaudel:
> The goal of this PR though is to stop generating tuple type definitions as that causes issues in diamond dependency scenarios

That makes sense.  Don't you have a similar issue with other parts of the runtime, though?  Things like `ArrayHelpers`, the type `nat`, or `DowncastClone` on `Func`s? (Or maybe the last one isn't an issue because it's extension methods?)

Maybe what we need is a Dafny file that's "the runtime" and that's included automatically, which would contain the definitions for tuples as well as the definition of `nat`?  And then your `/library:` mechanism could be used to exclude that file from compilation?

cpitclaudel:
Something like this:

```dafny
type nat = n: int | n >= 0
datatype {:nativeTuple} Tuple0 = Tuple0_()
datatype {:nativeTuple} Tuple1<A> = Tuple1_(0: A)
datatype {:nativeTuple} Tuple2<A, B> = Tuple2_(0: A, 1: B)
…
```

keyboardDrummer:
> > The goal of this PR though is to stop generating tuple type definitions as that causes issues in diamond dependency scenarios
> 
> That makes sense. Don't you have a similar issue with other parts of the runtime, though? Things like `ArrayHelpers`, the type `nat`, or `DowncastClone` on `Func`s? (Or maybe the last one isn't an issue because it's extension methods?)
> 

Any generated functions, such as DowncastClone or `ArrayHelpers.InitNewArray1`, can be scoped to the assembly that they're used in (in C# by making the classes internal), since they don't need to be compared between assemblies.

I didn't run into an issue with `nat` in https://github.com/dafny-lang/dafny/pull/2286, but I think it should be moved to the DafnyRuntime as well.

> Maybe what we need is a Dafny file that's "the runtime" and that's included automatically, which would contain the definitions for tuples as well as the definition of `nat`? And then your `/library:` mechanism could be used to exclude that file from compilation?

There's already an option, `/useRuntimeLib`, that excludes the runtime from being included in the generated code.

keyboardDrummer:
> Thanks, I'm sold. But I don't know how to review this PR. The tuple implementations are thousands of lines, and they are correctness-critical. 
> Did you generate them by copying them from a Dafny-generated file?

I did 

> Additionally, any change to the Java and C# code generators will have to update them so that their representation matches the new representation.

I think most changes won't require updating the tuple definitions, but in any case our regression suite should detect the absence of an update. Also, I think this PR makes the situation for tuples the same as it already is for seq/set/map/multiset.

keyboardDrummer:
> > Also, I think this PR makes the situation for tuples the same as it already is for seq/set/map/multiset.
> 
> The API of multiset/set is fixed, right (they do not have the same API as other datatypes).
> 
> We have docs describing changes to datatypes that would affect this code, I think; could you at least check in the code-generation script in `Scripts/` so as to make our lives easier in the future?

I generated this code by removing the `DoNotCompile()` attribute from tuple datatypes. The way to automatically generate it is by defining a `dafnyRuntime.dfy` file and generating (part of) the runtime code from that.

cpitclaudel:
> The way to automatically generate it is by defining a dafnyRuntime.dfy file, adding those tuples to it, and generating (part of) the runtime code from that.

Got it, thanks.  Maybe we can track this as a todo in an issue somewhere.

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -123,6 +123,10 @@ public class CsharpCompiler : SinglePassCompiler {
     }
 
     protected override void EmitBuiltInDecls(BuiltIns builtIns, ConcreteSyntaxTree wr) {
+      if (builtIns.MaxNonGhostTupleSizeUsed > 20) {
+        Reporter!.Error(MessageSource.Compiler, builtIns.MaxNonGhostTupleSizeToken, "C# back-end does not support tuples with more than 20 arguments.");
```
robin-aws, line 127: Any reason this doesn't just call `Error(...)`? https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/Compilers/SinglePassCompiler.cs#L76

Only nit-picking here because it will make adapting this to the changes I have pending for https://github.com/dafny-lang/dafny/issues/1770 easier.

Source/Dafny/AST/BuiltIns.cs
```diff
@@ -0,0 +1,287 @@
...
+    ObjectDecl = new TraitDecl(Token.NoToken, "object", SystemModule, new List<TypeParameter>(), new List<MemberDecl>(), DontCompile(), false, null);
+    SystemModule.TopLevelDecls.Add(ObjectDecl);
+    // add one-dimensional arrays, since they may arise during type checking
+    // Arrays of other dimensions may be added during parsing as the parser detects the need for these
+    UserDefinedType tmp = ArrayType(1, Type.Int, true);
+    // Arrow types of other dimensions may be added during parsing as the parser detects the need for these.  For the 0-arity
+    // arrow type, the resolver adds a Valid() predicate for iterators, whose corresponding arrow type is conveniently created here.
+    CreateArrowTypeDecl(0);
+    // Note, in addition to these types, the _System module contains tuple types.  These tuple types are added to SystemModule
+    // by the parser as the parser detects the need for these.
```
robin-aws, line 54: Out of date comment now :)

robin-aws (changes requested):
LGTM! Just some nits to clean up.

Regarding this todo:

>> The way to automatically generate it is by defining a dafnyRuntime.dfy file, adding those tuples to it, and generating (part of) the runtime code from that.

> Got it, thanks. Maybe we can track this as a todo in an issue somewhere.

How about just adding that to a `README` under `DafnyRuntime`?

Source/Dafny/Compilers/Compiler-Csharp.cs
keyboardDrummer: I saw that but it didn't seem to do anything useful unless you have a ConcreteSyntaxTree to write to. Maybe you can make Reporter inaccessible in your PR to prevent accidentally missing the logic you're adding?

robin-aws: Oh I can easily handle it if it doesn't make sense to call it here.

Now that I look closer, `Error(...)` is semi-broken because it assumes `/* ... */` is valid comment syntax in the target language. I'm not convinced it's worth the trouble to output errors in the target program in addition to reporting them, but I'll deal with that in my own PR.

cpitclaudel: > Now that I look closer, Error(...) is semi-broken because it assumes /* ... */ is valid comment syntax

https://github.com/dafny-lang/dafny/issues/1729

robin-aws: Nice, thanks! Assigned it to myself, will do something about it as I finish the work on #1770

Source/DafnyRuntime/dafnyRuntime.dfy
```diff
@@ -0,0 +1,23 @@
+method HasTuples() {
+  var has0 := (0);
```
robin-aws, line 2:
```suggestion
  var has0 := ();
```

The generated files look right anyway as I believe `()` gets declared implicitly anyway.

robin-aws (approved):

PR #2292, opened 2022-06-23
Enable AutoExtern to consume from multiple source files

keyboardDrummer:
### Changes

Enable AutoExtern to consume from multiple source files

### Tests

Updated both AutoExtern tests to use this new feature

### DIFF
--- a/Source/AutoExtern/Program.cs
+++ b/Source/AutoExtern/Program.cs
@@ -124,7 +124,7 @@ internal class AST : PrettyPrintable {
     this.model = model;
   }
 
-  public static IEnumerable<AST> FromFile(string projectPath, IEnumerable<string> sourceFiles, string cSharpRootNS) {
+  public static IEnumerable<AST> FromFiles(string projectPath, IEnumerable<string> sourceFiles, string cSharpRootNS) {
     // https://github.com/dotnet/roslyn/issues/44586
     MSBuildLocator.RegisterDefaults();
     var workspace = Microsoft.CodeAnalysis.MSBuild.MSBuildWorkspace.Create();
--- a/Source/AutoExtern/Program.cs
+++ b/Source/AutoExtern/Program.cs
@@ -124,7 +124,7 @@ internal class AST : PrettyPrintable {
     this.model = model;
   }
 
-  public static AST FromFile(string projectPath, string filePath, string cSharpRootNS) {
+  public static IEnumerable<AST> FromFile(string projectPath, IEnumerable<string> sourceFiles, string cSharpRootNS) {
     // https://github.com/dotnet/roslyn/issues/44586
     MSBuildLocator.RegisterDefaults();
     var workspace = Microsoft.CodeAnalysis.MSBuild.MSBuildWorkspace.Create();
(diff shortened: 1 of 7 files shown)


Source/AutoExtern/Program.cs
```diff
@@ -398,10 +400,16 @@ public static class Program {
     return template;
   }
 
-  public static string GenerateDafnyCode(string projectPath, string filePath, string cSharpRootNS) {
-    var ast = AST.FromFile(projectPath, filePath, cSharpRootNS);
+  public static string GenerateDafnyCode(string projectPath, IList<string> sourceFiles, string cSharpRootNS) {
     var wr = new StringWriter();
-    ast.Pp(wr, "");
+    var asts = AST.FromFile(projectPath, sourceFiles, cSharpRootNS).ToList();
```
cpitclaudel, line 405:
```suggestion
    var asts = AST.FromFiles(projectPath, sourceFiles, cSharpRootNS).ToList();
```

Source/AutoExtern/Program.cs
```diff
@@ -398,10 +400,16 @@ public static class Program {
...
+    var last = asts.Last();
+    foreach (var ast in asts) {
+      ast.Pp(wr, "");
+      if (ast != last) {
+        wr.WriteLine();
+      }
+    }
```
cpitclaudel, line 412:
```suggestion
    foreach (var ast in asts) {
      ast.Pp(wr, "");
      wr.WriteLine();
    }
```

Source/AutoExtern/Program.cs
```diff
@@ -124,7 +124,7 @@ internal class AST : PrettyPrintable {
     this.model = model;
   }
 
-  public static AST FromFile(string projectPath, string filePath, string cSharpRootNS) {
+  public static IEnumerable<AST> FromFile(string projectPath, IEnumerable<string> sourceFiles, string cSharpRootNS) {
```
cpitclaudel, line 127:
```suggestion
  public static IEnumerable<AST> FromFiles(string projectPath, IEnumerable<string> sourceFiles, string cSharpRootNS) {
```

cpitclaudel (dismissed):
LGTM, thanks.  Feel free to apply the suggestions or not.

Source/AutoExtern/Program.cs
keyboardDrummer: Why this suggestion?

cpitclaudel: It simplifies the code and it doesn't cost anything.  Not a big deal.

cpitclaudel (approved):

Source/AutoExtern/Program.cs
cpitclaudel: Also, it's marginally prettier, in that it doesn't depend on the sequence being non-empty? 🤷

keyboardDrummer: That's nice. I didn't want to have a double line in the output that's why I made this change.

cpitclaudel: OK, works for me :)

PR #2295, opened 2022-06-24
Tell the client which solverPath is used

keyboardDrummer:
### Changes

Tell the LSP client which solverPath is used

### Testing

Didn't add tests since this is a tiny logging feature but the current tests assure that this change does not cause issues.


cpitclaudel (commented):
We have an interface for telemetry events in `Workspace/ITelemetryPublisher`, maybe this should go in there?

cpitclaudel (approved):

cpitclaudel:
Thanks for making the change!

PR #2297, opened 2022-06-24
oldest and newest parameter for java in integration tests.

prvshah51:


keyboardDrummer (approved):

keyboardDrummer (approved):

robin-aws (changes requested):
Glad things still work on a much newer JDK, but no longer running on 1.8 means we might accidentally start emitting Java code that requires a newer Java version and break users.

We can ADD building on Java 18 if we want assurance that we support that too, however. I'd suggest another dimension on the `integration-tests-reusable.yml` workflow, something like `target-language-version: ["oldest", "newest"]`. We can have each target language define their oldest supported version independently. We'll probably want to only run with "newest" in the nightly build mind you.

.github/workflows/integration-tests-reusable.yml
```diff
@@ -66,10 +75,16 @@ jobs:
       run: |
         sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 60
     - uses: actions/setup-python@v1
-    - name: Set up JDK 1.8
+    - name: Set up oldest supported JDK
+      if: matrix.target-language-version == "oldest"
```
robin-aws, line 79: Turns out double quotes are a syntax error (https://github.com/dafny-lang/dafny/actions/runs/2756979457) - string literals have to use single quotes: https://docs.github.com/en/actions/learn-github-actions/expressions#literals

robin-aws (approved):
Thanks! It'll be great to hook up `oldest` for the other languages in the near future too.

PR #2298, opened 2022-06-24
Warning when constructor name is used as variable in match #1600

### LINKED ISSUE #1600: Warning when constructor name is used as variable in match
Consider the following program, which contains a simple mistake:

``` dafny
datatype Color = Red | Green | ShadesOfGray(nat)

predicate Monochromatic(c: Color) {
  match c
  case ShadesOfGray => true
  case anythingElse => false
}
```

For this program, Dafny issues a "redundant branch" warning for the second `case`. Why? Because the _constructor_ `ShadesOfGray` takes a parameter, and since the _pattern_ `ShadesOfGray` does not list a parameter, Dafny interprets the `ShadesOfGray` in the first pattern as a variable (which will match any `c`, just like the variable `anythingElse` in the second pattern would match anything that does not match in a prior `case`).

It would be nice if Dafny issued a warning in this case. I can imagine

    Warning: "ShadesOfGray" on line 5 is used as a variable, but it is also the name of a constructor that can be written here. If you intended to write a pattern that uses this constructor, add the parameters, like `ShadesOfGray(_)`.

Such a warning could be emitted once all resolution/type inference/type checking has been done. The check would then traverse through all `case` patterns. If inside a pattern it finds a variable `X` of type `T`, where `T` is a datatype that contains a constructor (with 1 or more parameters) named `X`, then the warning above is generated. Alternatively, this check could be implemented earlier, at the point where the Resolver today distinguishes between a nullary constructor (like `Red`) and a variable.

tobiaj10:
The code will produce a warning whenever a variable in a case expression starts with a capital letter. Fixes #1600 

I added a couple tests to a new file `/Test/linters/constructorMatch.dfy`


Source/Dafny/Rewriter.cs
```diff
@@ -118,6 +119,36 @@ public abstract class IRewriter {
       Contract.Requires(program != null);
     }
   }
+  // intern project 
+
+  class tobiRewriter : IRewriter {
```
keyboardDrummer, line 124: Please rename the class to something that matches what it does.

Source/Dafny/Rewriter.cs
```diff
@@ -118,6 +119,36 @@ public abstract class IRewriter {
       Contract.Requires(program != null);
     }
   }
+  // intern project
```
keyboardDrummer, line 122: Please remove this comment.

Source/Dafny/Rewriter.cs
keyboardDrummer, line 124: Could you add a comment above this class which contains a code examples of when this class shows an error?

Source/Dafny/Rewriter.cs
```diff
@@ -118,6 +119,36 @@ public abstract class IRewriter {
...
   }
+  // intern project 
+
+  class tobiRewriter : IRewriter {
+    internal override void PostResolve(Program program) {
+      base.PostResolve(program);
+      foreach (var moduleDefinition in program.Modules()) {
+        foreach (var topLevelDecl in moduleDefinition.TopLevelDecls.OfType<TopLevelDeclWithMembers>()) {
+          foreach (var function in topLevelDecl.Members.OfType<Function>()) {
+            if (function.Body is NestedMatchExpr matchExpr) {
```
keyboardDrummer, line 130: This only matches on match expressions at the root level of a function, not in expressions that occur in arbitrary places.

Source/Dafny/Rewriter.cs
```diff
@@ -118,6 +119,36 @@ public abstract class IRewriter {
...
+              var idPatterns = matchExpr.Cases.Select(matchCase => matchCase.Pat).OfType<IdPattern>();
+              foreach (var idPattern in idPatterns) {
+                if (char.IsLower(idPattern.Id[0]) == false && idPattern.Arguments == null) {
```
keyboardDrummer, line 133: `<expression> == false` can be simplified by `!<expression>`

keyboardDrummer, line 133: Could you extract `idPattern.Arguments == null` into a variable so the name of the variable can indicate why you are doing this check.

Source/Dafny/Rewriter.cs
```diff
@@ -118,6 +119,36 @@ public abstract class IRewriter {
...
+                  var lineNum = idPattern.Tok.line;
+                  string message = "Variable name '" + idPattern + "' on line " + lineNum +
```
keyboardDrummer, line 135: You can make this easier to read by using an interpolated string.

Also, I wouldn't put the location in the message since it's already passed to the `.Warning(` call which'll use it for some form of visualisation.

Test/linters/constructorMatch.dfy.expect
```diff
@@ -0,0 +1,5 @@
+/Users/tobiaja/Documents/dafny/Test/linters/constructorMatch.dfy(10,4): Warning: this branch is redundant
```
keyboardDrummer, line 1: This expect file seems incorrect since it contains the current but not the newly desired behavior.

Test/linters/constructorMatch.dfy
```diff
@@ -0,0 +1,64 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+module WithWarning {
```
keyboardDrummer, line 4: I like that you added the With and Without warning modules :)

Source/Dafny/Rewriter.cs
```diff
@@ -118,14 +120,58 @@ public abstract class IRewriter {
...
+    private readonly ErrorReporter reporter;
+    public ConstructorMatchWarningVisitor(ErrorReporter reporter) {
+      this.reporter = reporter;
+    }
+
+    protected override bool VisitOneExpr(Expression expr, ref Unit st) {
+      if (expr is NestedMatchExpr matchExpr) {
+        var idPatterns = matchExpr.Cases.Select(matchCase => matchCase.Pat).OfType<IdPattern>();
+        foreach (var idPattern in idPatterns) {
+          var variableChecker = idPattern.Arguments;
```
keyboardDrummer, line 159: I don't understand the name `variableChecker`. With my previous comment I was thinking of something like 
`var isVariable = idPattern.Argument == null`

Source/Dafny/Rewriter.cs
```diff
@@ -118,14 +120,58 @@ public abstract class IRewriter {
...
+          if (!char.IsLower(idPattern.Id[0]) && variableChecker == null) {
+            var lineNum = idPattern.Tok.line;
+            this.reporter.Warning(MessageSource.Rewriter, matchExpr.tok, $"Variable name '{idPattern}' on line {lineNum} starts with a capital letter");
```
keyboardDrummer, line 162: Could we change this message to say what's wrong: "variable name {idPattern} should begin with a lowercase letter"

Test/linters/constructorMatch.dfy.expect
```diff
@@ -0,0 +1,10 @@
+/Users/tobiaja/Documents/dafny/Test/linters/constructorMatch.dfy(11,12): Warning: this branch is redundant 
+/Users/tobiaja/Documents/dafny/Test/linters/constructorMatch.dfy(17,12): Warning: this branch is redundant 
+/Users/tobiaja/Documents/dafny/Test/linters/constructorMatch.dfy(25,16): Warning: this branch is redundant 
+/Users/tobiaja/Documents/dafny/Test/linters/constructorMatch.dfy(9,17): Warning: Variable name 'ShadesOfGray' on line 10 starts with a capital letter
```
keyboardDrummer, line 4: Note that the line number is mentioned twice, first in `(9,17)` and then in `on line 10`

keyboardDrummer (commented):
Should we also add a warning that reports if any datatype constructors are spelled with a lowercase, to match this warning?

Source/Dafny/Rewriter.cs
```diff
@@ -127,6 +127,7 @@ public abstract class IRewriter {
       method MonochromaticMethod(c: Color) returns (x: bool) {
       return match c
             case ShadesOfGray => true
+                 ^^^ Warning: Variable name 'ShadesOfGray' on line 129 starts with a capital letter
```
keyboardDrummer, line 130: This example warning doesn't match your updated warning.

Source/Dafny/Rewriter.cs
```diff
@@ -118,14 +120,59 @@ public abstract class IRewriter {
...
     }
   }
+  /*
+   * Code Example that displays a Warning
+   * 
+   * datatype Color = Red | Green | ShadesOfGray(nat)
+      method MonochromaticMethod(c: Color) returns (x: bool) {
+      return match c
+            case ShadesOfGray => true
+                 ^^^ Warning: Variable name 'ShadesOfGray' on line 129 starts with a capital letter
```
MikaelMayer, line 130: Can you please synchronize this commented error message with this line:
https://github.com/dafny-lang/dafny/pull/2298/files#diff-a6c43acea85285427a4a3b5de38c83abb9f9b966f95d9b992be559ccf376191aR163

Test/linters/constructorMatch.dfy.expect
```diff
@@ -0,0 +1,10 @@
+constructorMatch.dfy(11,12): Warning: this branch is redundant 
+constructorMatch.dfy(17,12): Warning: this branch is redundant 
+constructorMatch.dfy(25,16): Warning: this branch is redundant 
+constructorMatch.dfy(9,17): Warning: Variable name 'ShadesOfGray' should begin with a lowercase letter
```
MikaelMayer, line 4: There is no variable `ShadesOfGray` on line 9, column 17 of constructorMatch. Can you please figure out what's wrong, if any?

Test/linters/constructorMatch.dfy.expect
```diff
@@ -0,0 +1,10 @@
...
+constructorMatch.dfy(15,8): Warning: Variable name 'ShadesOfGray' should begin with a lowercase letter
```
MikaelMayer, line 5: Same remark here.

MikaelMayer (changes requested):
Great job, almost there !

PR #2300, opened 2022-06-25
fix: Substitute for heap labels in TrSplitExpr

### LINKED ISSUE #2299: Two-state predicate checked incorrectly
Using the following program, one can prove `false`:

``` dafny
class Twostate {
  var z: int

  method Test()
    modifies this
  {
    z := z + 1;
    label L: z := z - 1;
    assert Increase@L(); // BOGUS: this should not verify
    assert false;
  }

  twostate predicate Increase()
    reads this
  {
    old(z) <= z
  }
}
```

However, by changing the marked assertion above to `assert Increase@L() == true;`, an error is reported, as expected. This suggests that the problem lies with the internal inlining of predicates that's performed by `TrSplitExpr` in `Translator.cs`.

RustanLeino:
This PR fixes two things related to heap labels. One gave rise to unsound verification, and the other to malformed Boogie code.

Fixes #2299

### DIFF
--- a/Source/Dafny/Substituter.cs
+++ b/Source/Dafny/Substituter.cs
@@ -9,18 +9,20 @@ namespace Microsoft.Dafny {
   /// particular, the substituter does not copy parts of an expression that are used only for well-formedness checks.
   /// </summary>
   public class Substituter {
-    public readonly Expression receiverReplacement;
-    public readonly Dictionary<IVariable, Expression/*!*/>/*!*/ substMap;
-    public readonly Dictionary<TypeParameter, Type/*!*/>/*!*/ typeMap;
+    protected readonly Expression receiverReplacement;
+    protected readonly Dictionary<IVariable, Expression> substMap;
+    protected readonly Dictionary<TypeParameter, Type> typeMap;
+    protected readonly Label oldLabel;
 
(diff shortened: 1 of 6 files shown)


Source/Dafny/Substituter.cs
```diff
@@ -9,18 +9,20 @@ namespace Microsoft.Dafny {
...
-    public readonly Dictionary<TypeParameter, Type/*!*/>/*!*/ typeMap;
+    protected readonly Expression receiverReplacement;
+    protected readonly Dictionary<IVariable, Expression> substMap;
+    protected readonly Dictionary<TypeParameter, Type> typeMap;
+    protected readonly Label oldLabel;
 
     public static readonly Substituter EMPTY = new Substituter(null, new Dictionary<IVariable, Expression>(), new Dictionary<TypeParameter, Type>());
 
-    public Substituter(Expression receiverReplacement, Dictionary<IVariable, Expression/*!*/>/*!*/ substMap, Dictionary<TypeParameter, Type> typeMap) {
+    public Substituter(Expression receiverReplacement, Dictionary<IVariable, Expression> substMap, Dictionary<TypeParameter, Type> typeMap, Label oldLabel = null) {
```
keyboardDrummer, line 19: Consider renaming `oldLabel` to `oldHeapLabel`

keyboardDrummer (approved):

keyboardDrummer (approved):

PR #2306, opened 2022-06-27
Make ShutdownTest.LanguageServerShutsDownIfParentDies stable

keyboardDrummer:
Resolve the test instability that failed this run: https://github.com/dafny-lang/dafny/runs/7071695232?check_suite_focus=true

### DIFF
--- a/Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
+++ b/Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
@@ -71,12 +71,21 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
 
       // Wait for the language server to kill itself by waiting until it closes the output stream.
       await process.StandardOutput.ReadToEndAsync();
-      await Task.Delay(400); // Give the process some time to die
 
-      Assert.ThrowsException<ArgumentException>(() => {
-        var languageServer = Process.GetProcessById(int.Parse(languageServerProcessId));
-        languageServer.Kill();
-      }, "Language server should have killed itself if the parent is gone.");
+      for (int attempts = 0; attempts < 10; attempts++) {
+        // Give the language server time to kill itself
+        await Task.Delay(100);
+
+        try {
+          Process.GetProcessById(int.Parse(languageServerProcessId));
+        } catch {
+          return;
+        }
+      }
+
+      var languageServer = Process.GetProcessById(int.Parse(languageServerProcessId));
+      languageServer.Kill();
+      Assert.Fail("Language server should have killed itself if the parent is gone.");
     }
 
     private static async Task<Process> StartLanguageServerRunnerProcess() {
--- a/Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
+++ b/Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
@@ -72,9 +72,9 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
       // Wait for the language server to kill itself by waiting until it closes the output stream.
       await process.StandardOutput.ReadToEndAsync();
 
-      for (int attempts = 0; attempts < 10; attempts++) {
+      for (int attempts = 0; attempts < 500; attempts++) {
         // Give the language server time to kill itself
-        await Task.Delay(100);
+        await Task.Delay(10);
 
         try {
           Process.GetProcessById(int.Parse(languageServerProcessId));
--- a/.github/workflows/integration-tests-reusable.yml
+++ b/.github/workflows/integration-tests-reusable.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Populate OS list (all platforms)
         id: populate-os-list-all
         if: inputs.all_platforms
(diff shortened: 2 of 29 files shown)


Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -71,12 +71,21 @@ public class ShutdownTest {
...
-        languageServer.Kill();
-      }, "Language server should have killed itself if the parent is gone.");
+      for (int attempts = 0; attempts < 10; attempts++) {
+        // Give the language server time to kill itself
+        await Task.Delay(100);
+
+        try {
+          Process.GetProcessById(int.Parse(languageServerProcessId));
+        } catch {
+          return;
```
fabiomadge, line 82: Shouldn't this be a `continue`?

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -71,12 +71,21 @@ public class ShutdownTest {
...
-      Assert.ThrowsException<ArgumentException>(() => {
-        var languageServer = Process.GetProcessById(int.Parse(languageServerProcessId));
-        languageServer.Kill();
-      }, "Language server should have killed itself if the parent is gone.");
+      for (int attempts = 0; attempts < 10; attempts++) {
+        // Give the language server time to kill itself
+        await Task.Delay(100);
+
+        try {
+          Process.GetProcessById(int.Parse(languageServerProcessId));
```
fabiomadge, line 80: And a `break` here?

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
```diff
@@ -71,12 +71,21 @@ public class ShutdownTest {
...
+        } catch {
+          return;
+        }
+      }
+
+      var languageServer = Process.GetProcessById(int.Parse(languageServerProcessId));
```
fabiomadge, line 86: Couldn't we do this assignment in the loop?

fabiomadge (commented):
Why is that preferable to `Task.Dalay(1000)`?

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
keyboardDrummer: If this try doesn't throw an exception then the current iteration has failed, so we need to keep trying.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
keyboardDrummer: The catch is the happy case, in which the language server has killed itself and the test should pass.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
keyboardDrummer: This is the cleanup code once we've decided the language server won't kill itself so we're going to manually do it.

keyboardDrummer:
> Why is that preferable to `Task.Delay(1000)`?

Because that takes a long time.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
fabiomadge: Sure, but why the assignment.

fabiomadge:
> Because that takes a long time. The philosophy is, if you're waiting for an event that has no event listener, then keep looping while you wait for it until your test times out, instead of waiting an arbitrary time, since the latter produce unstable tests.

But why is 10*100 more stable than 1000? I'm not convinced this added complexity buys us much.

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
keyboardDrummer: The value is used on the next line. I'm not sure I understand the question.

fabiomadge: Yes, but you also generate it in L80. Might as well use that.

keyboardDrummer: Okay. I just looked into doing that but it creates warnings around assignment that I'd rather not override.

keyboardDrummer:
> Yes, but this is not what this PR is proposing. 

Changed it to wait 10ms for 500 times.

> I don't think a second is too much of a problem.

I think a second is too much, because we'll have thousands of unit tests. Having them take a second when they only take 100ms on average seems wasteful.

fabiomadge:
> I think a second is too much, because we'll have thousands of unit tests. Having them take a second when they only take 100ms on average seems wasteful.

Sure, but we're only discussing a single test. Maybe we could come up with some abstraction, if that is a pattern we expect to deploy often. At least in case of the unit tests, we have some parallelism which would alleviate some of the problems of excessive waiting.

fabiomadge (approved):

Source/DafnyLanguageServer.Test/Various/ShutdownTest.cs
fabiomadge: Thanks for looking into it.

PR #2312, opened 2022-06-28
Fixes gutter when cache is enabled

MikaelMayer:
Fixes https://github.com/dafny-lang/ide-vscode/issues/187

The issue was that caching would prevent reports of implementations being finished to verify for the gutter reporting mechanism. I added it, and a test that I verified would not work in the absence of the fix.

Changes
* It also report all the results that were cached as if they were resolved immediately.


Source/DafnyLanguageServer.Test/GutterStatus/CachedLinearVerificationGutterStatusTester.cs
```diff
@@ -0,0 +1,51 @@
...
+
+  [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+  public async Task EnsureCachingDoesNotHideErrors() {
+    await SetUp(new Dictionary<string, string>() {
+      { $"{VerifierOptions.Section}:{nameof(VerifierOptions.VerifySnapshots)}", "1" }
+    });
+    await VerifyTrace(@"
+ .  S [S][ ][I][S][S][ ]:method test() {
+ .  S [O][O][o][Q][O][O]:  assert true;
+ .  S [=][=][-][~][=][=]:  assert false;
```
MikaelMayer, line 47: Before the fix, the last [O] and [=] would have been replaced by [ ] because the result of individual cached splits were not reported the second time.

Source/DafnyLanguageServer.Test/GutterStatus/CachedLinearVerificationGutterStatusTester.cs
```diff
@@ -0,0 +1,51 @@
...
+
+  [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+  public async Task EnsureCachingDoesNotMakeSquigglyLinesToRemain() {
+    await SetUp(new Dictionary<string, string>() {
+      { $"{VerifierOptions.Section}:{nameof(VerifierOptions.VerifySnapshots)}", "1" }
+    });
+    await VerifyTrace(@"
+ .  S  S  |  I  $  | :method test() {
+ .  S  |  |  I  $  | :  assert true;
+ .  S  S  |  I  $  | :  //Next:
```
MikaelMayer, line 35: Before the fix, this text would not terminate because it would never get back to the | icon

Source/DafnyLanguageServer/Language/IVerificationProgressReporter.cs
```diff
@@ -20,5 +20,6 @@ public interface IVerificationProgressReporter {
     void ReportEndVerifyImplementation(Implementation implToken, Boogie.VerificationResult verificationResult);
     void ReportImplementationsBeforeVerification(Implementation[] implementations);
     void ReportAssertionBatchResult(AssertionBatchResult batchResult);
+    void ReportAssertionBatchResult(Implementation implementation, VCResult batchResult);
```
keyboardDrummer, line 23: Why are there two `ReportAssertionBatchResult` ? Seems like you only need the new one.

keyboardDrummer (approved):

PR #2314, opened 2022-06-29
fix: Correctly infer type of numeric arguments when the type is a subset type of a newtype

### LINKED ISSUE #2307: Type checker fails to infer types of constants as subset types of newtypes
Given a subset type `T` of a newtype with base type `U`, type checking fails on assigning an appropriate constant of type `U` to a variable of type `T`:
```dafny
newtype int0 = x: int | true
type int1 = x: int0 | true
const x: int1 := 0; // type for constant 'x' is 'int1', but its initialization value type is 'int'
```
(thanks @cpitclaudel for the minimal example)

RustanLeino:
Fixes #2307 

### DIFF
--- a/Source/Dafny/AST/DafnyAst.cs
+++ b/Source/Dafny/AST/DafnyAst.cs
@@ -1861,7 +1861,7 @@ namespace Microsoft.Dafny {
       Contract.Requires(builtIns != null);
       var j = JoinX(a, b, builtIns);
       if (DafnyOptions.O.TypeInferenceDebug) {
-        Console.WriteLine("DEBUG: Meet( {0}, {1} ) = {2}", a, b, j);
+        Console.WriteLine("DEBUG: Join( {0}, {1} ) = {2}", a, b, j);
       }
       return j;
     }
--- a/Source/Dafny/AST/DafnyAst.cs
+++ b/Source/Dafny/AST/DafnyAst.cs
@@ -2090,6 +2090,18 @@ namespace Microsoft.Dafny {
       Contract.Requires(b != null);
       Contract.Requires(builtIns != null);
 
+      a = a.NormalizeExpandKeepConstraints();
+      b = b.NormalizeExpandKeepConstraints();
+      if (a is IntVarietiesSupertype) {
+        return b is IntVarietiesSupertype || b.IsNumericBased(NumericPersuasion.Int) || b.IsBigOrdinalType || b.IsBitVectorType ? b : null;
+      } else if (b is IntVarietiesSupertype) {
+        return a.IsNumericBased(NumericPersuasion.Int) || a.IsBigOrdinalType || a.IsBitVectorType ? a : null;
+      } else if (a is RealVarietiesSupertype) {
+        return b is RealVarietiesSupertype || b.IsNumericBased(NumericPersuasion.Real) ? b : null;
+      } else if (b is RealVarietiesSupertype) {
+        return a.IsNumericBased(NumericPersuasion.Real) ? a : null;
+      }
+
       var towerA = GetTowerOfSubsetTypes(a);
       var towerB = GetTowerOfSubsetTypes(b);
       if (towerB.Count < towerA.Count) {
@@ -2129,16 +2141,8 @@ namespace Microsoft.Dafny {
       }
       Contract.Assert(towerA.Count == 1 && towerB.Count == 1);
 
-      if (a is IntVarietiesSupertype) {
-        return b is IntVarietiesSupertype || b.IsNumericBased(NumericPersuasion.Int) || b.IsBigOrdinalType || b.IsBitVectorType ? b : null;
-      } else if (b is IntVarietiesSupertype) {
-        return a.IsNumericBased(NumericPersuasion.Int) || a.IsBigOrdinalType || a.IsBitVectorType ? a : null;
-      } else if (a.IsBoolType || a.IsCharType || a.IsBigOrdinalType || a.IsTypeParameter || a.IsInternalTypeSynonym || a is TypeProxy) {
+      if (a.IsBoolType || a.IsCharType || a.IsBigOrdinalType || a.IsTypeParameter || a.IsInternalTypeSynonym || a is TypeProxy) {
         return a.Equals(b) ? a : null;
-      } else if (a is RealVarietiesSupertype) {
-        return b is RealVarietiesSupertype || b.IsNumericBased(NumericPersuasion.Real) ? b : null;
-      } else if (b is RealVarietiesSupertype) {
-        return a.IsNumericBased(NumericPersuasion.Real) ? a : null;
       } else if (a.IsNumericBased()) {
         return a.Equals(b) ? a : null;
       } else if (a.IsBitVectorType) {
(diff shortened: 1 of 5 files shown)


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -1798,8 +1798,7 @@ public enum AutoInitInfo { MaybeEmpty, Nonempty, CompilableValue }
       Contract.Requires(type != null);
       type = type.NormalizeExpandKeepConstraints();
       List<Type> tower;
-      var udt = type as UserDefinedType;
-      if (udt != null && udt.ResolvedClass is SubsetTypeDecl) {
+      if (type is UserDefinedType udt && udt.ResolvedClass is SubsetTypeDecl) {
         var sst = (SubsetTypeDecl)udt.ResolvedClass;
```
cpitclaudel, line 1802: I think you can merge these two clauses together into a single pattern matching:
```suggestion
      if (type is UserDefinedType {ResolvedClass: SubsetTypeDecl sst} udt) {
```

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -2061,14 +2060,12 @@ public enum AutoInitInfo { MaybeEmpty, Nonempty, CompilableValue }
 
       var joinNeedsNonNullConstraint = false;
       Type j;
-      if (a is UserDefinedType && ((UserDefinedType)a).ResolvedClass is NonNullTypeDecl) {
+      if ((a as UserDefinedType)?.ResolvedClass is NonNullTypeDecl aClass) {
```
cpitclaudel, line 2063:
```suggestion
      if (a is UserDefinedType { ResolvedClass: NonNullTypeDecl aClass }) {
```

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -2061,14 +2060,12 @@ public enum AutoInitInfo { MaybeEmpty, Nonempty, CompilableValue }
...
         joinNeedsNonNullConstraint = true;
-        var nnt = (NonNullTypeDecl)((UserDefinedType)a).ResolvedClass;
-        j = MeetX(nnt.RhsWithArgument(a.TypeArgs), b, builtIns);
-      } else if (b is UserDefinedType && ((UserDefinedType)b).ResolvedClass is NonNullTypeDecl) {
+        j = MeetX(aClass.RhsWithArgument(a.TypeArgs), b, builtIns);
+      } else if ((b as UserDefinedType)?.ResolvedClass is NonNullTypeDecl bClass) {
```
cpitclaudel, line 2066:
```suggestion
      } else if (b is UserDefinedType { ResolvedClass: NonNullTypeDecl bClass }) {
```

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -2093,6 +2090,18 @@ public enum AutoInitInfo { MaybeEmpty, Nonempty, CompilableValue }
...
+      a = a.NormalizeExpandKeepConstraints();
+      b = b.NormalizeExpandKeepConstraints();
+      if (a is IntVarietiesSupertype) {
+        return b is IntVarietiesSupertype || b.IsNumericBased(NumericPersuasion.Int) || b.IsBigOrdinalType || b.IsBitVectorType ? b : null;
+      } else if (b is IntVarietiesSupertype) {
+        return a.IsNumericBased(NumericPersuasion.Int) || a.IsBigOrdinalType || a.IsBitVectorType ? a : null;
+      } else if (a is RealVarietiesSupertype) {
+        return b is RealVarietiesSupertype || b.IsNumericBased(NumericPersuasion.Real) ? b : null;
+      } else if (b is RealVarietiesSupertype) {
+        return a.IsNumericBased(NumericPersuasion.Real) ? a : null;
```
cpitclaudel, line 2102: Nitpick: I find the asymmetry here confusing.  Is it semantically relevant that the check on `a` goes first?  I'm not fluent in what the two variants (`x is XVarietiesSupertype` vs. `x.IsNumericaBased(NumericPersuasion.X)`) mean, but I wonder if there would be a more symmetrical way to phrase this check.

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -2132,16 +2141,8 @@ public enum AutoInitInfo { MaybeEmpty, Nonempty, CompilableValue }
...
-      } else if (a.IsBoolType || a.IsCharType || a.IsBigOrdinalType || a.IsTypeParameter || a.IsInternalTypeSynonym || a is TypeProxy) {
+      if (a.IsBoolType || a.IsCharType || a.IsBigOrdinalType || a.IsTypeParameter || a.IsInternalTypeSynonym || a is TypeProxy) {
         return a.Equals(b) ? a : null;
-      } else if (a is RealVarietiesSupertype) {
-        return b is RealVarietiesSupertype || b.IsNumericBased(NumericPersuasion.Real) ? b : null;
-      } else if (b is RealVarietiesSupertype) {
-        return a.IsNumericBased(NumericPersuasion.Real) ? a : null;
       } else if (a.IsNumericBased()) {
         return a.Equals(b) ? a : null;
       } else if (a.IsBitVectorType) {
```
cpitclaudel, line 2148: Could you fold these two cases in the one above?

Source/Dafny/Resolver.cs
```diff
@@ -6838,7 +6838,11 @@ public CheckTypeInference_Visitor(Resolver resolver, ICodeContext codeContext)
           }
         } else if (CheckTypeIsDetermined(expr.tok, expr.Type, "expression")) {
           if (expr is UnaryOpExpr uop) {
-            uop.ResolveOp(); // Force resolution eagerly at this point to catch potential bugs
+            // The CheckTypeInference_Visitor has already visited uop.E, but uop.E's may be undetermined. If that happened,
+            // then an error has already been reported.
+            if (CheckTypeIsDetermined(uop.E.tok, uop.E.Type, "expression")) {
+              uop.ResolveOp(); // Force resolution eagerly at this point to catch potential bugs
+            }
```
cpitclaudel, line 6845: Are you sure about this part?  resolution for unary ops is lazy, as opposed to binary ops that are resolved explicitly when the resolver traverses the AST.

cpitclaudel (approved):
Looking good, left some nitpicky comments

Source/Dafny/AST/DafnyAst.cs
RustanLeino: I'm leaving this as is. The identity of values of type `RealVarietiesSupertype` [or, analogously, `IntVarietiesSupertype`] is not important. These types just say "this is a placeholder that stands for any real-valued [int-valued, respectively] type". So, the only asymmetry is that if both `a` and `b` satisfy `is RealVarietiesSupertype`, then `b` (not `a`) is returned, but this choice is arbitrary.

Source/Dafny/Resolver.cs
RustanLeino: Frankly, I'm _not_ sure about this change. It should be fine like it is now, for the reason explain in the comment on lines 6841-6842. The thing I'm not sure about is why the resolver started crashing here when I made the other changes. The `uop.ResolveOp()` method expects that `uop.E` has a legal type (and it crashes if it detects otherwise). I tried to provoke this crash in the previous version of Dafny, but wasn't able to. I also tried to see how the other changes I implemented could cause a change that brings control here, but was stumped.

So, I feel good about the new code here, but I don't understand why this wasn't a problem before.

cpitclaudel (approved):

PR #2317, opened 2022-06-29
Fixes #2308

### LINKED ISSUE #2308: Cannot prove tautology with map, opaque function and transparent function.
Consider the following code:

```dafny
function {:opaque} f(t: int): int { t }
function summarize(input: map<int, int>): int { |input| }

lemma lemme(input: map<int, int>, i: int, m: map<int, int>) {
    reveal f();
    assert  i == summarize(map k <- input :: f(k));
    assert  i == summarize(map k <- input :: f(k));
}
```
The last two asserts are failing, although they are identical. Changing the fuel of any function above does not help.
Note that the problem goes away either:
* If you remove the call to `summarize`
* If you replace the call `f(k)` by `k`
* If you remove `{:opaque}` from `f` and the `reveal f();` (just removing the `reveal f();` does not solve the problem)
* If you replace`i == summarizeChildren(` by `m == `

**Looking at Boogie**
The reveal statement generates no code in Boogie. That's weird.

As in #1631, the fuel is the only difference. However, contrary to #1631; all opaque functions are revealed and the problem is still there.
```boogie
    assert {:subsumption 0} i#0
       == _module.__default.summarize(StartFuelAssert__module._default.summarize, 
        Map#Glue((lambda $w#4: Box :: $IsBox($w#4, TInt) && Map#Domain(input#0)[$w#4]), 
          (lambda $w#4: Box :: 
            $Box(_module.__default.f(StartFuelAssert__module._default.f, $Unbox($w#4): int))), 
          TMap(TInt, TInt)));
    assume i#0
       == _module.__default.summarize(StartFuel__module._default.summarize, 
        Map#Glue((lambda $w#5: Box :: $IsBox($w#5, TInt) && Map#Domain(input#0)[$w#5]), 
          (lambda $w#5: Box :: 
            $Box(_module.__default.f(StartFuel__module._default.f, $Unbox($w#5): int))), 
          TMap(TInt, TInt)));
```

**Why it matters**
- Normally the second assert is in a postcondition, and I want to prove it.
- Normally, the first assert should have a variation of the second one that guides the verifier to prove the second one. But, if the tautology cannot be proven, any other variation does not help either.

MikaelMayer:
Fixes #

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 # Upcoming
 
 - fix: Fix bug in translation of two-state predicates with heap labels. (https://github.com/dafny-lang/dafny/pull/2300)
+- fix: Tautologies involving map comprehensions can now be proven (https://github.com/dafny-lang/dafny/pull/2317)
 - fix: Check that arguments of 'unchanged' satisfy enclosing reads clause. (https://github.com/dafny-lang/dafny/pull/2302)
 
 # 3.7.1
--- a/Source/Dafny/DafnyPrelude.bpl
+++ b/Source/Dafny/DafnyPrelude.bpl
@@ -1315,6 +1315,16 @@ axiom (forall<U, V> m: Map U V, m': Map U V::
   { Map#Equal(m, m') }
     Map#Equal(m, m') ==> m == m');
 
+// extensionality 2
+axiom (forall a: [Box]bool, b: [Box]Box, t0: Ty, t1: Ty :: 
+  { Map#Glue(a, b, TMap(t0, t1)) } 
+    (forall a2: [Box]bool, b2: [Box]Box, t2: Ty, t3: Ty ::
+    Map#Glue(a, b, TMap(t0, t1)) == Map#Glue(a2, b2, TMap(t2, t3))
+     <==> (forall u: Box :: a[u] == a2[u])
+       && (forall u: Box :: b[u] == b2[u])
+       && t0 == t2
+       && t1 == t3));
+
 function Map#Disjoint<U, V>(Map U V, Map U V): bool;
 axiom (forall<U, V> m: Map U V, m': Map U V ::
   { Map#Disjoint(m, m') }
--- a/Test/git-issues/git-issue-2308.dfy
+++ b/Test/git-issues/git-issue-2308.dfy
@@ -1,0 +1,11 @@
+// RUN: %dafny_0 /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+function {:opaque} f(t: int): int { t }
+function summarize(input: map<int, int>): int { |input| }
+
+lemma lemme(input: map<int, int>, i: int, m: map<int, int>) {
+    reveal f();
+    assert  i == summarize(map k <- input :: f(k));
+    assert  i == summarize(map k <- input :: f(k));
+}
--- a/Test/git-issues/git-issue-2308.dfy.expect
+++ b/Test/git-issues/git-issue-2308.dfy.expect
@@ -1,0 +1,3 @@
+git-issue-2308.dfy(9,14): Error: assertion might not hold
+
+Dafny program verifier finished with 0 verified, 1 error


MikaelMayer:
@RustanLeino , do you have an insight of the harm adding this axiom could cause? Are there alternatives?

RustanLeino:
The proposed axiom is not sound. For two reasons. Thinking about it also highlights other things to revisit in the axiomatization of maps, including another existing unsoundness.

## First problem

`Map#Glue` is used (in the Boogie encoding of Dafny) to create a Dafny map value. In Dafny, one map value can have several different Dafny types. For example, the _value_ `map[false := 7]` has type `map<bool, int>` and simultaneously has type `map<bool, nat>`. This is similar to how the value `7` is simultaneously an `int` and a `nat`. So, equality among map values should not consult the type used to construct the value.

(To revisit: The type argument to `Map#Glue` was probably useful at some point. I'm not sure if it's still useful. In fact, from a quick glance at it, I wasn't even sure it's encoded correctly. So, we should revisit `Map#Glue` at some point.)

As stated, the proposed axiom lets you prove `false` as follows:

``` dafny
function ToInt(x: nat): int { x }
function ToNat(x: nat): nat { x }

ghost method Test() {
  var m: map<int, int>;
  var n: map<int, nat>;
  m := map i | 0 <= i < 10 :: ToInt(i); // this uses Map#Glue where the third argument is TMap(TInt, TInt)
  n := map i | 0 <= i < 10 :: ToNat(i); // this uses Map#Glue with TMap(TInt, Tclass._System.nat())
  assert m == n; // this holds, because the equality of map values does not depend on the type used at construction
  // Because of the proposed axiom, one can now infer that TInt == Tclass._System.nat(), which is false. Hence:
  assert false; // provable
}
```

This unsoundness can be mitigated by (getting rid of `t2` and `t3` and) changing the proposed axiom to:

``` boogie
axiom (forall a: [Box]bool, b: [Box]Box, t0: Ty, t1: Ty :: 
  { Map#Glue(a, b, TMap(t0, t1)) } 
  (forall a2: [Box]bool, b2: [Box]Box :: 
    Map#Glue(a, b, TMap(t0, t1)) == Map#Glue(a2, b2, TMap(t0, t1))
       <==> (forall u: Box :: a[u] == a2[u])
         && (forall u: Box :: b[u] == b2[u])));
```

This modified axiom still solves the issue reported in #2308.

## Second problem

There's another soundness problem with the proposed axiom. It is that the axiom says the second argument to two equal `Map#Glue` values are necessarily the same. But they are only guaranteed to be the same for the values in the map's domain.

The following example illustrates:

``` dafny
function F(x: nat): nat { x }
function G(x: nat): nat { if x < 100 then x else x + 1 }

ghost method Test'() {
  var m := map i | 0 <= i < 10 :: F(i);
  var n := map i | 0 <= i < 10 :: G(i);
  assert m == n;
  assert false;
}
```

As given here, the verifier is not creative enough to leverage the unsound axiom. So, to show that the axiom is unsound, you have to add

``` boogie
assert Map#Elements(m#0)[$Box(200)] == $Box(200);
assert Map#Elements(n#0)[$Box(200)] == $Box(201);
```

to the Boogie that Dafny generates, before the translation of the `assert false;`. Both of these assertions are provable, and together with the new axiom then lets you prove `false`.

The following change to the axiom improves the situation:

``` boogie
axiom (forall a: [Box]bool, b: [Box]Box, t0: Ty, t1: Ty :: 
  { Map#Glue(a, b, TMap(t0, t1)) } 
  (forall a2: [Box]bool, b2: [Box]Box :: 
    Map#Glue(a, b, TMap(t0, t1)) == Map#Glue(a2, b2, TMap(t0, t1))
       <==> (forall u: Box :: a[u] == a2[u])
         && (forall u: Box :: a[u] ==> b[u] == b2[u]))); // notice the antecedent here 
```

This axiom also fixes the particular issue mentioned in Issue #2308.

## Yikes!

Disturbingly, I'm now realizing that the _existing_ axioms for `Map#Equals` and `Map#Elements` are unsound, due to the same "second problem" shown above. I reported this unsoundness as https://github.com/dafny-lang/dafny/issues/2375.

RustanLeino (changes requested):
The axiom is unsound, see my comment above.

cpitclaudel:
Note that my original "morphism" comment in the other thread had `<==`, not `<==>`; I don't think you'd ever want to prove function equalities from map equalities (I believe this is the source of the second unsoundness?)

MikaelMayer:
Closing as unsound improvement.

PR #2363, opened 2022-07-05
feat: Add `-diagnosticsFormat` command-line flag

### LINKED ISSUE #2084: Simplify using Dafny as a library
Currently, to use Dafny as a software library, one needs to use `DafnyPipeline.csproj` and then replicate most of the code in https://github.com/dafny-lang/dafny/blob/master/Source/DafnyDriver/DafnyDriver.cs in order verify a program. 

Dafny's CLI project, `DafnyDriver.csproj` is not published as a library (i.e., `<OutoutType>Library</OutputType>`). Furthermore, only the `Main` methods in DafnyDriver.cs are exposed with `public`, so one would need to pass an arguments string if it were available as a library in its current state.

cpitclaudel:
This PR adds `-diagnosticsFormat:json` to the CLI.  The output looks like this:

```
{"range":{"filename":"diagnosticsFormats.dfy","range":{"start":{"line":8,"character":0}}},"severity":2,"message":"module-level const declarations are always non-instance, so the \u0027static\u0027 keyword is not allowed here","source":"Parser","relatedInformation":[]}
{"range":{"filename":"diagnosticsFormats.dfy","range":{"start":{"line":9,"character":17}}},"severity":1,"message":"Error: result of operation might violate newtype constraint for \u0027byte\u0027","source":"Verifier","relatedInformation":[]}
{"range":{"filename":"diagnosticsFormats.dfy","range":{"start":{"line":12,"character":16}}},"severity":1,"message":"Error: A precondition for this call might not hold.","source":"Verifier","relatedInformation":[{"location":{"filename":"diagnosticsFormats.dfy","range":{"start":{"line":11,"character":35}}},"message":"Related location: This is the precondition that might not hold."}]}

Dafny program verifier finished with 1 verified, 2 errors
```

This format is consistent with a bunch of other tools (see below).  The intent is that client can look for `{` in column 0 and parse one object per such line.

Design notes:

- Many compilers have this option, but it's not named consistently.  Other tools
  use `-fdiagnostics-format=json`, `--json`, `--outputjson`, `--formatter=json`,
  `--format=json`, `--error-format=json`, `--output-format=json`,
  `--message-format=json`, etc.
- Output format is one message per line.  This makes it easy to parse output
  even if it's mixed with other non-JSON output.  This is a common
  approach (e.g. `rustc` does this), but it's not universal (GCC groups all
  messages into an array).  Streaming results allows clients to consume them as
  they come.

Implementation:

* `Source/Dafny/DafnyOptions.cs`
  (`DiagnosticsFormats`): New enum.
  (`DiagnosticsFormat`): Default to `PlainText`.
  (`ParseOption`): Recognize new `-diagnosticsFormat` flag.
  (`HelpBody`): Document it.
* `Source/DafnyDriver/DafnyDriver.cs`
  (ThreadMain): Delay initialization of `ErrorReporter`.
* `Source/Dafny/JsonDiagnostics.cs`: New file.  Capture Dafny and boogie message
  and print them as single-line JSON objects.
* `Test/cli/diagnosticsFormats.dfy`: New test.
* `docs/DafnyRef/UserGuide.md`: Document `-diagnosticsFormat`.

Closes #2084 (by adressing the underlying use case), closes #2229 (by superseding it).  CC @mschlaipfer


MikaelMayer:
This is a great addition to the Dafny ecosystem which will surely make Dafny more modular.
A few remarks before I dive in the PR review:
1. I would like you to detect if the token is a RangeToken, and in this case add an "end" field to the range above, so that the range has both "start" and "end"
2. Since it's API, I think we should also offer
  *. The absolute offset of the position. Necessary so that programmatic API consumers don't need to think of lines/column when parsing our error messages.
  *. The file name. There might be several files and just the position is not enough. There are also included files, and related errors.
3. Related positions should be, as for the Language Server, nested under the main error, not a top-level error. In your example, the message containing "Related location: " is a top-level one... Look at how LSP does it:
https://github.com/dafny-lang/dafny/blob/8792af1b11a053fa1a8c1d81b8a50bf8a4aec7bb/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs#L55-L77

cpitclaudel:
> I would like you to detect if the token is a RangeToken, and in this case add an "end" field to the range above, so that the range has both "start" and "end"

Done, can you give me an example of code with a RangeToken so I can test the code?

> Since it's API, I think we should also offer. The absolute offset of the position. 

I don't think we should deviate from LSP here, but I can do it if you want.

> The file name. 

Yep, good catch.

> Related positions should be, as for the Language Server, nested under the main error, not a top-level error.

I will look into that.

cpitclaudel:
>      I will look into that.

Done.

cpitclaudel:
>> Since it's API, I think we should also offer. The absolute offset of the position.

> I don't think we should deviate from LSP here, but I can do it if you want.

I changed my mind.  I'd love to do that, but we don't have the data.  Tokens have a UTF-16 offset, which is not very usable by clients.  Until we move away from that, I don't think we should include that info in the output.

cpitclaudel:
> Can you update the PR description?

With the example output?  I can!  It's also in the diff of the PR (there's a test)

cpitclaudel:
Ready for review, I think

RELEASE_NOTES.md
```diff
@@ -1,5 +1,7 @@
 # Upcoming
 
+- feat: New option `-diagnosticsFormat:json` to print Dafny error messages as JSON objects (one per line). (https://github.com/dafny-lang/dafny/pull/2363)
+
```
MikaelMayer, line 4:
```suggestion
```

Source/Dafny/DafnyOptions.cs
```diff
@@ -990,6 +1016,10 @@ is printed.
 /printTooltips
     Dump additional positional information (displayed as mouse-over tooltips by
     the VS plugin) to stdout as 'Info' messages.
+/diagnosticsFormat:<text|json>
+    Chose how to report errors, warnings, and info messages.
```
MikaelMayer, line 1020:
```suggestion
    Choose how to report errors, warnings, and info messages.
```

MikaelMayer:
> I changed my mind. I'd love to do that, but we don't have the data. Tokens have a UTF-16 offset, which is not very usable by clients. Until we move away from that, I don't think we should include that info in the output.

An UTF-16 offset? I don't think so. I have been consistently using `pos` to extract strings from program and it works well, either reading a buffer or the C# string. See https://github.com/dafny-lang/dafny/pull/2434

Source/Dafny/JsonDiagnostics.cs
```diff
@@ -0,0 +1,114 @@
...
+using Microsoft.Boogie;
+using VCGeneration;
+
+namespace Microsoft.Dafny;
+
+record DiagnosticMessageData(MessageSource source, ErrorLevel level, IToken tok, string? category, string message, List<ErrorInformation.AuxErrorInfo>? related) {
+  private static JsonObject SerializePosition(IToken tok) {
+    return new JsonObject {
+      ["line"] = tok.line,
+      ["character"] = tok.col - 1
```
MikaelMayer, line 18: Why put a character indexed by zero, but a line indexed by 1?
Let's be consistent
```suggestion
      ["character"] = tok.col
```

Test/cli/diagnosticsFormats.dfy.expect
```diff
@@ -0,0 +1,24 @@
+diagnosticsFormats.dfy(8,0): Warning: module-level const declarations are always non-instance, so the 'static' keyword is not allowed here
+diagnosticsFormats.dfy(9,17): Error: result of operation might violate newtype constraint for 'byte'
+diagnosticsFormats.dfy(12,16): Error: A precondition for this call might not hold.
+diagnosticsFormats.dfy(11,35): Related location: This is the precondition that might not hold.
+
+Dafny program verifier finished with 1 verified, 2 errors
+{"range":{"filename":"diagnosticsFormats.dfy","range":{"start":{"line":8,"character":0}}},"severity":2,"message":"module-level const declarations are always non-instance, so the \u0027static\u0027 keyword is not allowed here","source":"Parser","relatedInformation":[]}
```
MikaelMayer, line 7: To test for the "end", you have to provide the argument name `/showSnippets:1`

Test/cli/diagnosticsFormats.dfy.expect
```diff
@@ -0,0 +1,24 @@
...
+
+Dafny program verifier finished with 1 verified, 2 errors
+diagnosticsFormats.dfy(8,0): Warning: module-level const declarations are always non-instance, so the 'static' keyword is not allowed here
+diagnosticsFormats.dfy(7,8): Info: newtype byte resolves as {:nativeType "byte"} (Detected Range: 0 .. 256)
+diagnosticsFormats.dfy(9,17): Error: result of operation might violate newtype constraint for 'byte'
+diagnosticsFormats.dfy(12,16): Error: A precondition for this call might not hold.
+diagnosticsFormats.dfy(11,35): Related location: This is the precondition that might not hold.
+
+Dafny program verifier finished with 1 verified, 2 errors
+{"range":{"filename":"diagnosticsFormats.dfy","range":{"start":{"line":8,"character":0}}},"severity":2,"message":"module-level const declarations are always non-instance, so the \u0027static\u0027 keyword is not allowed here","source":"Parser","relatedInformation":[]}
```
MikaelMayer, line 19:
```suggestion
{"location":{"filename":"diagnosticsFormats.dfy","range":{"start":{"line":8,"character":0}}},"severity":2,"message":"module-level const declarations are always non-instance, so the \u0027static\u0027 keyword is not allowed here","source":"Parser","relatedInformation":[]}
```
I suggest we use "range" only for the start and the end, and "location" for something that holds a file name and a range, like LSP is doing.

MikaelMayer (changes requested):
A few items, otherwise, great start to make Dafny become more library-friendly.

Test/cli/diagnosticsFormats.dfy.expect
cpitclaudel: I don't understand: `/showSnippets` is for showing the error in context, and we don't want to do that in JSON mode, right?

Source/Dafny/JsonDiagnostics.cs
cpitclaudel: That's what the CLI does, and we want the same output in JSON and non-json mode.

Test/cli/diagnosticsFormats.dfy.expect
MikaelMayer: `/showSnippets` enables Dafny to create RangeTokens instead of just Tokens.
For the JSON mode, if `/showSnippets` is activated, it would be reasonable to me that we don't add the snippet as part of the error message (but we could... that could be useful), but that way we can report error using ranges.

MikaelMayer:
> Yep, strings are UTF-16 in C#, so if you use these offsets into strings in C# it will work, but it wouldn't work in other languages that use bytes or utf-8. It would be weird to leak that.

But isn't "col" already leaking this information?

cpitclaudel:
> But isn't "col" already leaking this information?

😱 you're right :'(
I will add `pos` then, it doesn't make things worse.

Test/cli/diagnosticsFormats.dfy.expect
cpitclaudel: Good catch, will fix, that was the intended thing and I got it wrong :'(

Test/cli/diagnosticsFormats.dfy.expect
MikaelMayer: Well the only reason was that I'm too lazy (or don't think it's worth it?) to change the 1200+ files of our integration tests  And it could be a breaking change (however, I don't think this would be a big deal)

Test/cli/diagnosticsFormats.dfy.expect
cpitclaudel: Fixed

cpitclaudel:
> I will add pos then, it doesn't make things worse.

Done

Test/cli/diagnosticsFormats.dfy.expect
cpitclaudel: Oh, good point regarding the test files.

cpitclaudel: I've added `-showSnippets:1` to the test and documented it in UserGuide.md

RELEASE_NOTES.md
```diff
@@ -1,7 +1,6 @@
 # 3.7.3
 
 - feat: New option `-diagnosticsFormat:json` to print Dafny error messages as JSON objects (one per line). (https://github.com/dafny-lang/dafny/pull/2363)
```
MikaelMayer, line 3: Move this feat in the Upcoming section?

Source/Dafny/JsonDiagnostics.cs
```diff
@@ -14,6 +14,7 @@ namespace Microsoft.Dafny;
 record DiagnosticMessageData(MessageSource source, ErrorLevel level, IToken tok, string? category, string message, List<ErrorInformation.AuxErrorInfo>? related) {
   private static JsonObject SerializePosition(IToken tok) {
     return new JsonObject {
+      ["pos"] = tok.pos,
```
MikaelMayer, line 17: A few CI failures line 1 to fix?

Source/Dafny/JsonDiagnostics.cs
```diff
@@ -74,7 +75,7 @@ record DiagnosticMessageData(MessageSource source, ErrorLevel level, IToken tok,
     var auxRelated = related?.SelectMany(SerializeAuxInfo) ?? Enumerable.Empty<JsonNode>();
     var innerRelated = SerializeInnerTokens(tok);
     return new JsonObject {
-      ["range"] = SerializeToken(tok),
+      ["location"] = SerializeToken(tok),
```
MikaelMayer, line 78: A few CI failures line 93?

MikaelMayer (changes requested):
Looks good, almost there.

MikaelMayer (dismissed):
Great job

MikaelMayer (approved):

PR #2366, opened 2022-07-05
chore: Mechanism for compilers to declare unsupported features

robin-aws:
This change adds a `Feature` enumeration and a virtual `Compiler.UnsupportedFeatures` property, which allows compilers to declare what Dafny features they are known not to support. These sets are now used to generate the compiler feature support table previously added to the reference manual: https://dafny.org/dafny/DafnyRef/DafnyRef#2498-supported-features-by-target-language. [This commit](https://github.com/dafny-lang/dafny/commit/d36605490948e5715352b5ad5a88977a6bcb147f) copies the existing content into a separate file, and [this one](https://github.com/dafny-lang/dafny/commit/16581771654f798e4135b73f605ab12eb989ce6e) updates it to match the generated content (and only changes a single row).

Having these known gaps declared in this way also enables a more rigorous approach to testing, where any valid, executable Dafny test file can be run against all current compilers, and a compiler can cleanly reject an otherwise valid program that contains known unsupported features. Doing this across the current test suite will undoubtably reveal more gaps.

To progress towards this goal and provide at least some testing of the `Compiler.UnsupportedFeatures` data, this change also adds a simple `TestDafny` CLI, exposed to integration tests as `%testdafny`, and converts a few test files to use it instead of the current pattern. This utility runs the given test file against each compiler in sequence, and will only allow a particular compiler to fail to compile a particular test file if it reports an error in the following pattern (and the named feature is present in its `UnsupportedFeatures` set):

```
File.dfy(1:2): Error: Feature not supported for this compilation target: <Feature description>
```

I haven't yet gone through the test suite to convert all compatible tests, because it still needs a way to adjust for expected non-determinism such as element ordering when `print`-ing `set` values (not to mention this PR is large enough as it is!)

There are now two convenient ways for a compiler to cleanly reject an unsupported feature, both explicitly accepting a `Feature` value:

1. Calling the `SinglePassCompiler.UnsupportedFeatureError(...)` helper method will emit an error with the expected format.
2. Throwing an `UnsupportedFeatureException` will terminate the compilation process and emit a similar error (thanks to a catch block in the calling `DafnyDriver` logic). This is useful for contexts where it is awkward to work around the required feature and continue compiling.

I largely populated the `UnsupportedFeatures` sets (and hence the previously-added reference manual table) by converting all cases of throwing `NotImplementedException` to either a `UnsupportedFeatureException` or a `cce.UnreachableException()` (the latter being the right choice if a valid Dafny program should not be able to trigger it). This makes for a better user experience in these cases, since the output will now be a single clean error message instead of an exception and stack trace.

This mostly addresses #1770, although I'd like to follow up with converting most executable files in the test suite to using `%testdafny` for higher coverage before we consider that complete.


robin-aws:
@cpitclaudel: Just a heads up that rather than defining independent "feature predicates" to determine whether a program uses a particular feature, as I proposed in https://github.com/dafny-lang/dafny/issues/1770, I'm instead letting the compilers define these predicates implicitly, by allowing them to report errors/throw exceptions when they encounter the features in their control flow. I realized this was far simpler, and doesn't stop us from lifting this logic out later while maintaining the same structure.

cpitclaudel (changes requested):
Very nice work.  I have left comments in the PR.

Source/Dafny/Compilers/Compiler-cpp.cs
```diff
@@ -893,10 +921,10 @@ protected class ClassWriter : IClassWriter {
       } else if (xType is CharType) {
         return "char";
       } else if (xType is IntType || xType is BigOrdinalType) {
-        Warn("BigInteger used", tok);
+        UnsupportedFeatureError(tok, Feature.UnboundedIntegers);
         return "BigNumber";
```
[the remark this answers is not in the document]
robin-aws: Nope, `UnsupportedFeatureError` just reports the compile error and returns normally. It only throws an exception if the feature is not in the `UnsupportedFeatures` set (as that indicates a true internal error).

Test/metatests/ConsistentWhenSupported.dfy
```diff
@@ -0,0 +1,26 @@
+
+// TIRED
```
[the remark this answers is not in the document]
robin-aws: Attempt at a meme reference. :) Fixing.

Source/TestDafny/TestDafny.cs
```diff
@@ -0,0 +1,243 @@
+﻿using System.Reflection;
+using System.Text;
+using CommandLine;
+using Microsoft.Dafny;
+using Microsoft.Dafny.Plugins;
+using XUnitExtensions;
+using XUnitExtensions.Lit;
+
+namespace TestDafny;
```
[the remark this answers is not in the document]
robin-aws: I just added a README.md, let me know if it provides enough motivation and explanation.

Source/TestDafny/TestDafny.cs
```diff
@@ -0,0 +1,243 @@
...
+        if (!IsAllowedOutputLine(compiler, line)) {
+          return false;
+        }
+      }
+    }
+
+    return true;
+  }
+
+  private static bool IsAllowedOutputLine(Compiler compiler, string line) {
```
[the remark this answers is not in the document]
robin-aws: See above.

Test/comp/Hello.dfy
```diff
@@ -1,12 +1,7 @@
-// RUN: %dafny /compile:3 /spillTargetCode:2 /compileTarget:cs "%s" > "%t"
-// RUN: %dafny /compile:3 /spillTargetCode:2 /compileTarget:js "%s" >> "%t"
-// RUN: %dafny /compile:3 /spillTargetCode:2 /compileTarget:go "%s" >> "%t"
-// RUN: %dafny /compile:3 /spillTargetCode:2 /compileTarget:java "%s" >> "%t"
-// RUN: %dafny /compile:3 /spillTargetCode:2 /compileTarget:py "%s" >> "%t"
-// RUN: %diff "%s.expect" "%t"
+// RUN: %testdafny "%s"
```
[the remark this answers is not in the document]
robin-aws: That could be a useful option independently! But that wouldn't be as good a complete solution by itself: assuming you would still pass `/compile:3` to actually run every compiled program in sequence, you'd still have to duplicate the expected output N times in the `*.expect` file.

The nice thing is that IF we add that option in the future, `%testdafny` could be refactored to use it (and account for diffing against the `*.expect` file correctly.) without having to change any tests, and we'd probably get a nice speed up by not invoking `dafny` multiple times.

Source/Dafny/Compilers/Compiler-cpp.cs
```diff
@@ -582,7 +610,7 @@ public class CppCompiler : SinglePassCompiler {
           wr.WriteLine("typedef {0} {1};", nt_name_def, nt.Name);
         }
       } else {
-        throw NotSupported(String.Format("non-native newtype {0}", nt));
+        throw new UnsupportedFeatureException(nt.tok, Feature.NonNativeNewtypes, String.Format("non-native newtype {0}", nt));
```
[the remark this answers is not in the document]
robin-aws: All of them besides C++. C++ only supports newtypes whose bounds happen to match one of the types named in `Compiler.SupportedNativeTypes`.

Source/Dafny/Compilers/Compiler-cpp.cs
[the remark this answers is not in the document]
cpitclaudel: Ah, nice, thanks

Source/TestDafny/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+dotnet run --project Source/TestDafny/TestDafny.csproj -- for-each-compiler --help
+```
+
+## Known limitations
+
+This tool does not yet have a way to account for non-deterministic output. In particular, it is common for existing test cases
+to print `set<T>` values, and the order that elements are printed in is not consistent across runtimes (nor do the language semantics
+even guarantee that a given backend will pick a consistent ordering). I intend to add something similar to `OutputCheck` directives
+to `*.expect` files to address this in the future.
```
cpitclaudel, line 37: I don't think we should be building testing tools ourselves: if lit is enough for LLVM, shouldn't it be enough for Dafny?

Source/TestDafny/README.md
```diff
@@ -0,0 +1,37 @@
+# TestDafny (a.k.a %testdafny)
+
+This internal, testing-only utility is used to execute a single Dafny source file against each compile target option and assert that the output
```
cpitclaudel, line 3: I would call it CheckDafnyCompilersOutput, or DafnyCompilers.Test, and potentially special-case it in the lit-runner instead of making it a separate binary (if we decide that we're dropping the usual lit),

Source/TestDafny/README.md
```diff
@@ -0,0 +1,37 @@
...
+is the same as the corresponding `<test file>.expect` file. This is a very common testing approach implemented manually across the integration test suite with multiple
+lit commands. See [this test file](../../Test/metatests/ConsistentWhenSupported.dfy) for a comparison of the manual approach
+compared to using this utility.
+
+Note the manual approach has several downsides:
+
+1. New compilers have to be manually added to every relevant test file.
+2. The expected output must be duplicated for each compile target, making it hard to perceive inconsistencies.
```
cpitclaudel, line 11: I'm surprised about this part: can't we write the following?

```
dafny /compileTarget:cs /compile:3 > %t.cs-output
DIFF %t.cs-output ….expect
dafny /compileTarget:java /compile:3 > %t.java-output
DIFF %t.java-output ….expect
dafny /compileTarget:go /compile:3 > %t.go-output
DIFF %t.go-output ….expect
…
```

Source/TestDafny/README.md
```diff
@@ -0,0 +1,37 @@
...
+3. When a test fails, it is difficult to determine from the output which compile target failed.
```
cpitclaudel, line 12: Not with the above, I think?

Same if we added `Dafny /compileTarget:all`; we would print `Running xyz.java`, `running xyz.cs`, etc.

Source/TestDafny/README.md
robin-aws: That's a valid point, and would be a slightly better version of the existing pattern. I'll rephrase slightly.

Source/TestDafny/README.md
robin-aws: As discussed offline: I didn't want to make this only work for the xUnit runner since that would definitely break using `lit`, which some developers are (reasonably) still using since I haven't finished on #2033 yet. Making it an internal-only project we can run with `dotnet run` meant I didn't have to pull that into scope yet. 

I'm going to make the lit substitution key `%testDafnyForEachCompiler` instead which seems like it should satisfy your justifiable naming concern. :)

Source/TestDafny/README.md
robin-aws: Agreed we should be cautious about building whole testing frameworks ourselves. I see this as morally equivalent to the kinds of domain-specific helper methods or stub interface implementations one tends to write within established testing frameworks. I think of "a valid input program should behave the same for each supported backend" as a reusable assertion we need within the lit framework, and that isn't easily expressed in a reusable way without packaging it up as an internal CLI.

Source/TestDafny/README.md
robin-aws: Yup, there are definitely other solutions to #2 and #3. I added a bit after the list to acknowledge that.

cpitclaudel (approved):

PR #2369, opened 2022-07-06
chore: Run single-platform integration tests even if nightly build failed

robin-aws:
There's no good reason to not run the integration tests in this situation: the results can still help tell you if your PR is in good shape while the nightly build gets fixed.

Our previous attempt to fix this didn't work because if `check-deep-tests` failed, the `integration-tests` job was skipped no matter what it's `if` clause evaluated to. Instead this change removes the `need` dependency entirely. All jobs still have to pass to unblock the PR anyway.

### DIFF
--- a/.github/workflows/msbuild.yml
+++ b/.github/workflows/msbuild.yml
@@ -53,13 +53,12 @@ jobs:
       run: dotnet pack --no-build dafny/Source/Dafny.sln
 
   check-deep-tests:
+    if: ! contains(github.event.pull_request.labels.*.name, 'run-deep-tests'))
     uses: ./.github/workflows/check-deep-tests-reusable.yml
     with:
       branch: master
 
   integration-tests:
-    needs: check-deep-tests
-    if: always() && (needs.check-deep-tests.result == 'success' || contains(github.event.pull_request.labels.*.name, 'run-deep-tests'))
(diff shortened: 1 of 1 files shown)


fabiomadge:
I think conserving resources for the people working on fixing the deep test issue is actually a good idea.

.github/workflows/msbuild.yml
```diff
@@ -53,13 +53,12 @@ jobs:
       run: dotnet pack --no-build dafny/Source/Dafny.sln
 
   check-deep-tests:
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'run-deep-tests') }}
```
fabiomadge, line 56:
```suggestion
    if: !contains(github.event.pull_request.labels.*.name, 'run-deep-tests')
```

robin-aws: Nope, that's what I tried first, but then the value looks like a YAML tagged value: https://yaml.org/spec/1.2.2/#691-node-tags :)

fabiomadge: It worked before and the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif) also claim so.

fabiomadge: My bad, I just tried it. 😞

PR #2372, opened 2022-07-07
Turn off NuGet publish until secrets.NUGET_API_KEY works

keyboardDrummer:
Currently the nightly run is failing because automatically publishing to NuGet is failing because `secrets.NUGET_API_KEY` seems to be empty: https://github.com/dafny-lang/dafny/runs/7228527386?check_suite_focus=true
Note that a secret `NUGET_API_KEY` has been configured for `dafny-lang/dafny`.

Automatically publishing to NuGet has not worked before, so let's comment it out until we get it working

### DIFF
--- a/.github/workflows/publish-release-reusable.yml
+++ b/.github/workflows/publish-release-reusable.yml
@@ -114,6 +114,7 @@ jobs:
       if: ${{ inputs.prerelease }}
       # NuGet will consider any package with a version-suffix as a prerelease
       run: dotnet pack --version-suffix ${{ inputs.name }} --no-build dafny/Source/Dafny.sln
+#
 #    - name: Upload package to NuGet
 #      if: ${{ inputs.release_nuget }}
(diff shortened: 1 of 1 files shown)


MikaelMayer (approved):
Not publishing nightly releases of the nuget package won't even harm anyone in the short term, so this is a good compromise for now.

robin-aws (changes requested):
I'm trying to figure out why the API key isn't getting inserted now.

robin-aws:
I found that I hadn't enabled access to the org-level secret from this repository. I've fixed that, but retrying the failed job is still not working. I suspect that's because the workflow gets a snapshot of the secrets at the time of running and hence retrying doesn't pick up the permissions change.

I've kicked off a manual run of the nightly build instead: https://github.com/dafny-lang/dafny/actions/runs/2631946159. Let's see if it goes through.

robin-aws:
Still not working. I'll experiment with a side project instead and grudgingly allow this for now. :)

robin-aws (approved):

PR #2391, opened 2022-07-11
Rework notification publishing order

keyboardDrummer:
Note, this PR depends on a yet unpublished Boogie version: https://github.com/boogie-org/boogie/pull/605

### Problem

Currently we have some LSP server internal state that is updated concurrently, specifically the verification result state, so verification diagnostics, run status, and counterexamples.

We synchronise updates to this state. However, the notification publishing that happens after occurs asynchronously, which can cause the notifications to be published out of order, which can cause a particular verification task to first show up as Completed and then as Running.

OK:
```
-----> Time
Change1: => sync S0->S1, copy S1 =======> Publish S1 copy
Change2: ========> sync S1->S2, copy S2 ========> Publish S2 copy
```

BAD, since the intermediate state S1 is published after the final state S2:
```
-----> Time
ChangeA: ==> sync S0->S1, copy S1 ================> Publish S1 copy
ChangeB: =========> sync S1->S2, copy S2 ===> Publish S2 copy
```

What we need is for the publications to happen in the same order as the internal document updates.

### Change
Use a separate thread to apply and publish verification updates:
```
-----> Time
Change1: ====> queue S1 update
Change2: ===========================> queue S2 update
VerificationEventLoop: =======> dequeue update, S0->S1, publish S1 ====> dequeue update, S1->S2, publish S2
```

### Testing
Add the test `ManyConcurrentVerificationRuns`, which ensures that when starting many verification runs of individual tasks, the final status of all tasks is that they're completed.


Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -64,7 +64,7 @@ private class CounterExampleLoader {
           logger.LogDebug("got no counter-examples for document {DocumentUri}", document.Uri);
           return new CounterExampleList();
         }
-        var counterExamples = GetLanguageSpecificModels(document.Counterexamples)
+        var counterExamples = GetLanguageSpecificModels(document.Counterexamples.ToArray())
```
cpitclaudel, line 67: ?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -26,17 +34,16 @@ public record DafnyDocument(
...
     Dafny.Program Program,
     SymbolTable SymbolTable,
     bool WasResolved,
-    IReadOnlyList<IImplementationTask>? VerificationTasks = null,
     bool LoadCanceled = false
   ) {
+    public IScheduler UpdateScheduler { get; } = new EventLoopScheduler();
+    public Subject<DafnyDocument> VerificationUpdates { get; } = new();
+
+    public IReadOnlyList<IImplementationTask>? VerificationTasks { get; set; }= null;
```
cpitclaudel, line 46:
```suggestion
    public IReadOnlyList<IImplementationTask>? VerificationTasks { get; set; } = null;
```

Source/DafnyLanguageServer/Workspace/DiagnosticsObserver.cs
```diff
@@ -0,0 +1,59 @@
...
+    if (e is TaskCanceledException) {
+      OnCompleted();
+    } else {
+      logger.LogError(e, "error while handling document event");
+      telemetryPublisher.PublishUnhandledException(e);
+    }
+  }
+
+  public void OnNext(DafnyDocument document) {
+    lock (this) { // TODO do we need this lock?
```
cpitclaudel, line 49: Locks on `this` are always a bit scary

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -64,7 +63,7 @@ IRelocator relocator
 
     public async Task<bool> CloseDocumentAsync(TextDocumentIdentifier documentId) {
       if (documents.Remove(documentId.Uri, out var databaseEntry)) {
-        databaseEntry.Observer.Dispose();
+        // databaseEntry.Observer.Dispose();
```
cpitclaudel, line 66: ?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -187,7 +193,7 @@ IRelocator relocator
         newDocument = newDocument with { LastChange = lastChange };
         if (newDocument.SymbolTable.Resolved) {
           var resolvedDocument = newDocument with {
-            ImplementationIdToView = migratedImplementationViews,
+            ImplementationIdToView = new (migratedImplementationViews),
```
cpitclaudel, line 196: ?

cpitclaudel (commented):
I'm having trouble following the complete PR, but to get started I left `?` comments in a few places

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
keyboardDrummer: From ConcurrentStack to Array

cpitclaudel: Could you change the signature of `GetLanguageSpecificModels` to take an `IEnumerable`?

Source/DafnyLanguageServer/Handlers/Custom/VerificationHandler.cs
```diff
@@ -43,12 +43,10 @@ public class VerificationHandler : IJsonRpcRequestHandler<VerificationParams, bo
     var translatedDocument = await documentEntry.TranslatedDocument;
     var requestPosition = request.Position;
     var tasks = GetTasksAtPosition(translatedDocument, requestPosition).ToList();
```
cpitclaudel, line 45: I wonder why this code calls `ToList`

Source/DafnyLanguageServer/Handlers/Custom/VerificationHandler.cs
```diff
@@ -43,12 +43,10 @@ public class VerificationHandler : IJsonRpcRequestHandler<VerificationParams, bo
...
-    foreach (var taskToRun in tasks) {
-      var verifiedDocuments = documentLoader.Verify(translatedDocument, taskToRun, CancellationToken.None);
-      documentEntry.Observe(verifiedDocuments);
```
cpitclaudel, line 48: Why do we not need to observe the task any more?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -72,8 +79,23 @@ public record DafnyDocument(
...
+
+    /// <summary>
+    /// Creates a clone of the DafnyDocument
+    /// </summary>
+    public DafnyDocument Snapshot() {
+      var result = new DafnyDocument(TextDocumentItem, ParseAndResolutionDiagnostics, CanDoVerification, GhostDiagnostics,
+        Program, SymbolTable, WasResolved, LoadCanceled)
+      {
+        VerificationTree = VerificationTree,
+        Counterexamples = new(Counterexamples),
```
cpitclaudel, line 93: Is this thread-safe?  I would have thought you'd need a lock around this to make sure that the various data structures are in sync with each other.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -96,7 +95,10 @@ IRelocator relocator
 
       if (VerifyOnOpen) {
         Verify(documentEntry, cancellationSource.Token);
+      } else {
+        documentEntry.EndVerification();
```
cpitclaudel, line 99: Why do we mark verification as complete in this case? / What is this for?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
```diff
@@ -111,22 +113,22 @@ IRelocator relocator
...
+
         if (!RequiresOnSaveVerification(document) || !document.CanDoVerification) {
-          return Observable.Empty<DafnyDocument>();
+          entry.EndVerification();
+          return;
         }
 
-        return documentLoader.VerifyAllTasks(document, cancellationToken);
-      }, TaskScheduler.Current).ToObservable().Merge();
-      documentEntry.Observe(updates);
```
cpitclaudel, line 129: Is this removal for the same reason as above?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
cpitclaudel: I think I understand this part better now, but I'm surprised that this sort of cloning doesn't need a lock (to make sure that the stuff that's getting cloned is all snapshot together

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -103,46 +102,40 @@ VerifierOptions verifierOptions
...
       }
 
-      var result = loaded with {
-        CounterexamplesCollector = new ConcurrentStack<Counterexample>(),
-        ImplementationIdToViewCollector = initialViews,
-        VerificationTasks = verificationTasks,
-        ImplementationIdToView = initialViews.ToImmutableDictionary(),
-      };
+      loaded.ImplementationIdToView = initialViews;
+      loaded.VerificationTasks = verificationTasks;
```
cpitclaudel, line 132: Is it better to mutate this object in place than to return a new one?  (Is that what the old code did?)

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -221,57 +209,60 @@ bool loadCanceled
...
 
-      var result = statusUpdates.Select(boogieStatus => HandleStatusUpdate(document, implementationTask, boogieStatus)).Replay();
-      result.Connect(); // Immediately start processing the status updates.
+      entry.RestartVerification();
+      statusUpdates.ObserveOn(dafnyDocument.UpdateScheduler).Subscribe(
+        update => HandleStatusUpdate(entry, dafnyDocument, implementationTask, update),
+        () => {
+          if (dafnyDocument.VerificationTasks!.All(t => t.IsIdle)) {
+            entry.EndVerification();
+          }
```
cpitclaudel, line 259: Is it safe to iterate on the tasks without locking here?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -280,39 +271,27 @@ bool loadCanceled
...
-      result.DefaultIfEmpty(document).LastAsync().Subscribe(finalDocument => {
-        // All unvisited trees need to set them as "verified"
-        if (!cancellationToken.IsCancellationRequested) {
-          SetAllUnvisitedMethodsAsVerified(document);
-        }
-
-        document.GutterProgressReporter!.ReportRealtimeDiagnostics(true, finalDocument);
-      });
+      // TODO something against publishing stale state
+      document.VerificationUpdates.OnNext(document);
```
cpitclaudel, line 294: TODO?

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -352,9 +331,8 @@ bool loadCanceled
         document, statusPublisher, NotificationPublisher);
     }
 
-    private async Task NotifyStatusAsync(TextDocumentItem item, IObservable<DafnyDocument> documents, CancellationToken cancellationToken) {
-      var finalDocument = await documents.ToTask(cancellationToken);
-      var errorCount = finalDocument.ImplementationIdToView.Values.Sum(r => r.Diagnostics.Count(d => d.Severity == DiagnosticSeverity.Error));
+    private void NotifyStatusAsync(TextDocumentItem item, DafnyDocument document, CancellationToken cancellationToken) {
+      var errorCount = document.ImplementationIdToView.Values.Sum(r => r.Diagnostics.Count(d => d.Severity == DiagnosticSeverity.Error));
```
cpitclaudel, line 335: How safe is it to iterate on a concurrent dictionary like this without locking?

keyboardDrummer: I think `.Values` makes a copy while keeping a lock. Also, this code is run after all verification tasks are completed and the dictionary is no longer updated.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: The tasks are only set once during the translation phase. The `.IsIdle` value of the tasks can change after that, but once the final running task reaches this OnComplete handler, no `.IsIdle` changes are still occurring.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: The old code had both mutable fields related to verification and sometimes made copies of them, but I think this only gave a false sense of security. In this PR, `DafnyDocument` is designed to be mutated concurrently by verification tasks.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: This is for tests that want to know whether to expect any more updates.

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
keyboardDrummer: Removed the ToArray call since `document.Counterexamples` is a List now. `document` comes from DiagnosticObserver.LastPublishedDocument which contains a snapshot of the document that's not mutated, so no concurrency issues should occur here.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: This part has been changed now.

Source/DafnyLanguageServer/Handlers/Custom/VerificationHandler.cs
keyboardDrummer: Because now there's a single VerificationUpdates Observable per document that's already subscribed to by the publisher.

The problem was that `documentEntry.Observe` kept the observations from calls to it in order, so then if you verify task1 and then task2, but task2 completes before task1, you'll still see the notifications of task1 before task2.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
cpitclaudel: I wonder if they could use the telemetry notification that we added in https://github.com/dafny-lang/dafny/commit/5641c36fc08c347a081bad8b0260ca770dfc0059 ? (Or maybe it would be good to test that notification using the EndVerification event

keyboardDrummer: I'm removing that telemetry event because I don't think it fires at the right time in this PR. It relied on the Completed event from an observable, but the `DafnyDocument.VerificationUpdates` Observable in this PR is not completed even when the LSP server is idle, because the user can still manually trigger more verification tasks.

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
keyboardDrummer: This is only called inside a lock now.

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
cpitclaudel: That's unfortunate; will there be any way to detect completion of verification for a given file?  I'm thinking of programmatic uses of Dafny, where you want to gather errors and warnings for a given file using the LSP server.

keyboardDrummer: "completion of verification", you mean where everything that could be verified has been? You can call `VerifyAllTasksAsync` and then `await DocumentEntry.Idle` to do that.

cpitclaudel: This is from C#, but I meant for an LSP client.  Suppose that I want to say "the document has been parsed, resolved, and verified; what can I depend on?  Only the custom verification status messages, right?

It would be nice to have a "standard" telemetry message.  And it should be possible to compute it correctly, since we have state update with our custom status event.

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -16,16 +16,27 @@ namespace Microsoft.Dafny.LanguageServer.Handlers.Custom {
...
       this.documents = documents;
+      this.documentLoader = documentLoader;
     }
 
     public async Task<CounterExampleList> Handle(CounterExampleParams request, CancellationToken cancellationToken) {
       try {
         var document = await documents.GetLastDocumentAsync(request.TextDocument);
+
         if (document != null) {
+          var verificationTasks = document!.VerificationTasks;
```
cpitclaudel, line 32: You can save one level of nesting with something like `if (document?.VerificationTasks is {} vts) {` or `if (document is { VerificationTasks: {} vts })`?

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -16,16 +16,27 @@ namespace Microsoft.Dafny.LanguageServer.Handlers.Custom {
...
+          if (verificationTasks != null) {
+            var entry = documents.Documents[request.TextDocument.Uri];
```
cpitclaudel, line 34: I don't understand these two levels of nesting: `GetLastDocumentAsync` looks up `TextDocument` using its URI, but then we look up `TextDocument.Uri` again in the result?

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -16,16 +16,27 @@ namespace Microsoft.Dafny.LanguageServer.Handlers.Custom {
...
+            foreach (var task in verificationTasks) {
+              documentLoader.Verify(entry, document, task, CancellationToken.None);
```
cpitclaudel, line 36: Will this rerun verification if it already ran, or only if it hasn't run yet?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -79,18 +77,17 @@ public record DafnyDocument(
 
     public int LinesCount => VerificationTree.Range.End.Line;
     public IVerificationProgressReporter? GutterProgressReporter { get; set; }
-    public ConcurrentStack<Counterexample> Counterexamples { get; set; } = new();
-    public ConcurrentDictionary<ImplementationId, ImplementationView> ImplementationIdToView { get; set; } = new();
+    public List<Counterexample> Counterexamples { get; set; } = new();
+    public Dictionary<ImplementationId, ImplementationView> ImplementationIdToView { get; set; } = new();
```
cpitclaudel, line 81: Why is it safe to use a regular collection here?

cpitclaudel (commented):
Looks very nice.  The additional fixes are relatively simple!
I have one high-level question: with the new model, can we still answer queries while waiting for verification?  If so, is it because verification messages come as notifications, and hence are not part of this request / response sequence that we're keeping ordered?

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
keyboardDrummer: > Only the custom verification status messages, right?

What's wrong with using these? The "standard" telemetry message would require custom logic to detect as well right?

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
keyboardDrummer: The type of `entry` is `DocumentEntry`, which is different from `DafnyDocument`. `GetLastDocumentAsync` is an async call that makes sure the document is translated, then later we need `entry` to verify the individual tasks. This all looks better in the refactored PR: https://github.com/dafny-lang/dafny/pull/2478/files#diff-d628271b7999a61822dcdc03e39b669393f7f504799c979c7fa7b1bf8f4911f9R27

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
keyboardDrummer: Only if it hasn't run.

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
keyboardDrummer: It's 
```c#
if (A) 
{ 
  if (B) { 
    C() 
  } 
  D() 
}
```
How would I get rid of one `if` ?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
keyboardDrummer: It's only updated by the thread managed by `DafnyDocument.UpdateScheduler`

Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs
cpitclaudel: Correct.  But most LSP clients I've worked with make it marginally simpler to handle telemetry messages than custom ones, and we have changed the custom messages in the past.  Having a consistent telemetry message for that is nice.  Not a blocker.

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
cpitclaudel: Woops, thanks

cpitclaudel (approved):
Congrats, nice PR.

PR #2405, opened 2022-07-14
doc: misc non-semantic typos in RM, part 2

davidcok:
Fixes non-semantic typos in the RM, sections from Statements on


docs/DafnyRef/UserGuide.md
```diff
@@ -220,13 +215,13 @@ default, but can be skipped by using the `-noVerify` or `-dafnyVerify:0` option
...
 - The particular target language used is chosen by `-compileTarget`
-- Whether or not the dafny tool attempts to run the compiled code is controlled by `-compile`
+- Whether or not the `dafny` tool attempts to run the compiled code is controlled by `-compile`
 
 The dafny tool terminates with these exit codes:
 
 * 0 -- success
 * 1 -- invalid command-line arguments
-* 2 -- parse or types errors
+* 2 -- syntax, parse or type errors
```
keyboardDrummer, line 224: What's the difference between a syntax and a parse error? I think we can drop `parse` from this list.

docs/DafnyRef/UserGuide.md
```diff
@@ -401,14 +396,14 @@ This list is not exhaustive but can definitely be useful to provide the next ste
...
 
 #### 24.8.2.1. `assume false;` {#sec-assume-false}
 
 Assuming `false` is an empirical way to short-circuit the verifier and usually stop verification at a given point[^explainer-assume-false], and since the final compilation steps do not accept this command, it is safe to use it during development.
 Another similar command, `assert false;`, would also short-circuit the verifier, but it would still make the verifier try to prove `false`, which can also lead to timeouts.
 
-[^explainer-assume-false]: `assume false` tells the Dafny verifier "Assume everything is true from this point of the program". The reason is that, 'false' proves anything. For example, `false ==> A` is always true because it is equivalent to `!false || A`, which reduces to `true || A`, which reduces to `true`.
+[^explainer-assume-false]: `assume false` tells the `dafny` verifier "Assume everything is true from this point of the program". The reason is that, 'false' proves anything. For example, `false ==> A` is always true because it is equivalent to `!false || A`, which reduces to `true || A`, which reduces to `true`.
```
keyboardDrummer, line 406: Why do you prefer `dafny` over Dafny here?

keyboardDrummer (commented):
Very nice!

docs/DafnyRef/Topics.md
```diff
@@ -681,10 +681,12 @@ and HOL [@Krauss:PhD].
 
 ## 23.5. Functions in Dafny
 
-In this section, I explain with examples the support in
+This section explains with examples the support in
 Dafny for well-founded functions, extreme predicates,
 and proofs regarding these.
 
+TODO: This topic appears to replicate the previous section
```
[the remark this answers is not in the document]
davidcok: Yes - along with many other TODOs - a placeholder for editing to be done that is too much for this PR

docs/DafnyRef/UserGuide.md
davidcok: You are right that it could be either -- I was thinking that the verifier was part of the tool, but alternately the Dafny language system has a verifier though the language itself does not

docs/DafnyRef/Topics.md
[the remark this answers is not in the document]
keyboardDrummer: But this TODO shouldn't be in the published documentation right?

docs/DafnyRef/UserGuide.md
keyboardDrummer: Okay, but from the user's perspective there's not first a lexing and then a parsing phase, there's just a step that determines if the input program matches the Dafny grammar, right?

davidcok:
In the long run no, but for now there are very many, per Mikael’s Quip document. And part of my task is to tackle them one by one (just did one in another PR)

I’d like to have the RM be complete and consistent with Dafny 4 when Dafny 4 is released, though I don’t know the time frame for that yet.

davidcok:
Well, my view is that the language requires that all the explicit and implicit assertions be true; it does not say how that is established.

The tool does its best to establish those assertions, but if you read through all the attributes and command-line options there are a lot of heuristics to the “mechanics”.
And if the mechanics improve in the future, that makes the tool more usable but does not change the language. 
One may have to adapt how one uses the language depending on the capabilities of the tool.

And for that matter there could be different tools with different capabilities - that also does not change the language

The fact that the language includes attributes that are advice to the tool is muddying the line, in my view.

keyboardDrummer:
> In the long run no, but for now there are very many, per Mikael’s Quip document. And part of my task is to tackle them one by one (just did one in another PR) I’d like to have the RM be complete and consistent with Dafny 4 when Dafny 4 is released, though I don’t know the time frame for that yet.

I guess we can leave them, although another way of working would be to have the TODOs somewhere separately, like in GitHub issues, so the published documentation doesn't contain information that isn't useful to the reader such as. "TODO This section should be rewritten"

davidcok:
Yes, though that step also include name and type resolution.
Perhaps I’m reflecting the fact that when I taught compilers (and the way the Dafny RM presents the language), tokenization, parsing and resolution were seen as separate compiler phases, followed by optimizations and code generation

keyboardDrummer (dismissed):

keyboardDrummer (dismissed):

robin-aws (approved):

davidcok:
As I go through them, I’ll either address the TODO or, if it is really a biggee, make an Issue of it.

PR #2409, opened 2022-07-14
doc: RM - aligning RM with LambdaSpec grammar

davidcok:
Fixes a mismatch between the RM and the grammar for LambdaSpec; removes a TODO


MikaelMayer (approved):
It feels better to get rid of allowLabel, thanks a lot.
Next, I'm looking forward putting these grammars at the end of each section (or in a separate file) to replace with examples.

davidcok:
Actually I’d rather extend RequiresClause with a allowAttributes argument and reuse it here, but that involves changing the grammar.

PR #2410, opened 2022-07-14
chore: Generate html version of the Reference Manual

fabiomadge:
Fixes #

### DIFF
--- a/.github/workflows/refman.yml
+++ b/.github/workflows/refman.yml
@@ -1,4 +1,4 @@
-name: Build DafnyRef.pdf
+name: Build DafnyRef
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


davidcok (changes requested):
I'm not clear on what this adds. We already serve the documentation on GitHub Pages and the pdf is part of the release package.

Is this building html pages from pdf or from markdown?
Is it intending to include the html in the release package?

I'm not sure that the content of docs are adequately synched with the release to warrant packaging them with the release just yet.

fabiomadge:
This wasn't meant for public consumption. This build only makes sure everything still works, we never publish the result. But it's nice to know what effect a PR has on the finished product, so we upload the PDF we have anyway. https://github.com/dafny-lang/dafny/actions/runs/2671601320 We don't do the same for html, so I figured it would be nice to have and not too expensive when reusing the same worker, but it doesn't work on macOS, so no luck.

PR #2417, opened 2022-07-14
fix: Add missing triggers to Map#Card axioms in Boogie prelude

### LINKED ISSUE #2380: Incompleteness in map cardinality equality
The following without uncommenting the commented line fails to verify:
```dafny
method M(t: map<int, int>, k: int)
  requires k in t
{
  var lt := |t|;
  var lt' := |t - {k}|;
  // var _ := |(t - {k}).Keys|;
  assert lt == lt' + 1;
}
```

Uncommenting the line makes it verify, because it triggers an axiom that is needed.

cpitclaudel:
Closes #2380.

@atomb , can you show me how to run a performance regression run on this?  I'm not sure how it will impact our customers.


Test/git-issues/git-issue-2380.dfy
```diff
@@ -0,0 +1,15 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method M(t: map<int, int>, k: int)
+  requires k in t
+{
+  assert |t| == |t - {k}| + 1;
```
MikaelMayer, line 7: Could you assert something about the cardinality of the values and of the items? That way, you can also test your two other axioms

cpitclaudel: Good idea!  Done.

MikaelMayer (approved):
Fantastic ! Thanks for doing it.

cpitclaudel:
Thanks for the approval!  Let's hold off on merging this PR until @prvshah51 helps me make sure that it doesn't cause performance regressions, though :)

MikaelMayer:
I'm launching a regression test to see if we can merge this one.

MikaelMayer:
Although these axioms are useful and sound, @atomb and I found that existing proofs would be slow down by up to 5%, so we don't know what to do at this point other than stall and not include these axioms.

atomb (approved):
After taking into account other recent changes, this now usually improves performance.

PR #2419, opened 2022-07-15
fix: Issue #2415

### LINKED ISSUE #2415: Initializing a new array with an initializer function not documented
There are three cases for array initialization, and the reference manual doesn't mention the third (using an initializer function, as in sequence constructions like `seq(5, i => i * 2)`):

```dafny
var a := new int[5];                // Elements initialized with arbitrary values
var b := new int[] [1, 2, 3, 4, 5]; // Elements initialized with the given values
var b := new int[5] (i => i * 2);   // Elements initialized with the given initializer function

```

davidcok:
Fixes #2415

### DIFF
--- a/docs/DafnyRef/Types.md
+++ b/docs/DafnyRef/Types.md
@@ -226,7 +226,7 @@ The language includes a literal for each integer, like
 using the prefix "`0x`", as in `0x0`, `0xD`, and `0x7c1` (always with
 a lower case `x`, but the hexadecimal digits themselves are case
 insensitive).  Leading zeros are allowed.  To form negative literals,
-use the unary minus operator, as in `-12`, but not -(12).
+use the unary minus operator, as in `-12`, but not `-(12)`.
 
 There are also literals for some of the reals.  These are
 written as a decimal point with a nonempty sequence of decimal digits
@@ -280,8 +280,8 @@ There are also operators on each numeric type:
   `*`            | multiplication (times)
   `/`            | division (divided by)
   `%`            | modulus (mod)  -- int only
-  -----------------|------------------------------------
-`-`            | negation (unary minus)
+-----------------|------------------------------------
+  `-`            | negation (unary minus)
 
 The binary operators are left associative, and they associate with
 each other in the two groups.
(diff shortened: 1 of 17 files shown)


docs/DafnyRef/Types.md
```diff
@@ -2796,8 +2796,13 @@ The initial values of the array elements are arbitrary values of type
...
 a := new T[] [t1, t2, t3, t4];
 ```
+The initial values can also be initialized using a lambda expressions, as follows:
+```
+a := new int[5](i => i*i);
+```
+
 The length of an array is retrieved using the immutable `Length`
-member.  For example, the array allocated above satisfies:
+member.  For example, the array allocated with `a := new int[5];` satisfies:
```
MikaelMayer, line 2805:
```suggestion
member.  For example, the array allocated with `a := new T[n];` satisfies:
```

MikaelMayer (changes requested):
Thanks. I did not know about the lambda initialization for arrays. Very useful.

robin-aws (changes requested):

MikaelMayer (approved):
you might want to consider @robin-aws 's suggestion as well.

davidcok:
Yes - I added some txt for that also.

MikaelMayer (approved):

robin-aws (approved):

cpitclaudel:
I believe this was already documented here:

`````
## 21.16. Array Allocation
````grammar
ArrayAllocation_ =
  "new" [ Type ] "[" [ Expressions ] "]"
  [ "(" Expression(allowLemma: true, allowLambda: true) ")"
  | "[" [ Expressions ] "]"
  ]
````

This allocates a new single or multi-dimensional array as explained in
section [Section 15](#sec-array-types).
The initialization portion is optional. One form is an
explicit list of values, in which case the dimension is optional:
```dafny
var a := new int[5];
var b := new int[5][2,3,5,7,11];
var c := new int[][2,3,5,7,11];
var d := new int[3][4,5,6,7]; // error
```
The comprehension form requires a dimension and uses a function of
type `nat -> T` where `T` is the array element type:
```dafny
var a := new int[5](i => i*i);
```
`````

MikaelMayer:
6 sections later ! We should probably add a link between these sections.

PR #2428, opened 2022-07-15
feat: Add `multiset`s and some type descriptors to the Python compiler

### LINKED ISSUE #2393: Type descriptors for Python
Type descriptors, which we use when compiling generic methods/functions, are missing in Python.

fabiomadge:
Fixes #2393.


Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -39,57 +44,169 @@ def _break(name):
...
+        while True:
+            s = queue.pop()
+            missing = complete - s
+            if missing == set():
+                break
+            for e in missing:
+                n = s.copy()
+                n.add(e)
+                yield Set(n)
+                queue.appendleft(n)
```
cpitclaudel, line 102: https://docs.python.org/3/library/itertools.html#itertools-recipes has a recipe for powerset that's short and sweet

fabiomadge: Thanks!

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -460,12 +460,53 @@ protected class ClassWriter : IClassWriter {
...
+
+      var customName = false;
+      var name = type.NormalizeExpandKeepConstraints() switch {
+        // unresolved proxy; just treat as bool, since no particular type information is apparently needed for this type
+        BoolType or TypeProxy => "bool",
+        IntType or BitvectorType => "int",
+        RealType => "real",
+        UserDefinedType udt => udt.ResolvedClass switch {
+          TypeParameter tp => TypeParameterDescriptor(tp),
+          ClassDecl => "null",
```
robin-aws, line 482: Wouldn't this be better as an exception too? Perhaps an `UnsupportedFeatureException`?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -460,12 +460,53 @@ protected class ClassWriter : IClassWriter {
...
 
     protected override string TypeDescriptor(Type type, ConcreteSyntaxTree wr, IToken tok) {
-      throw new UnsupportedFeatureException(Token.NoToken, Feature.RuntimeTypeDescriptors);
+      Contract.Requires(type != null);
+      Contract.Requires(tok != null);
+      Contract.Requires(wr != null);
+
+      var customName = false;
+      var name = type.NormalizeExpandKeepConstraints() switch {
+        // unresolved proxy; just treat as bool, since no particular type information is apparently needed for this type
```
robin-aws, line 476: Did you actually encounter this in testing or is it just for completeness? I'd prefer `cce.UnreachableException()` if it's a case that indicates a bug, even if it's upstream in the resolver.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -1231,13 +1281,17 @@ protected class ClassWriter : IClassWriter {
         case BinaryExpr.ResolvedOpcode.Mod:
           staticCallString = $"{DafnyRuntimeModule}.euclidian_modulus"; break;
 
-        case BinaryExpr.ResolvedOpcode.Lt:
```
robin-aws, line 1234: I don't see this moving somewhere else, is it just not supported now?

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -39,57 +45,159 @@ def _break(name):
...
     raise TailCall()
 
-class Seq(list):
+class Seq(tuple):
     def __init__(self, __iterable = None, isStr = False):
         if __iterable is None:
             __iterable = []
-        self.isStr = isinstance(__iterable, str) or isStr
-        super().__init__(__iterable)
+        self.isStr = isStr or isinstance(__iterable, str) or (isinstance(__iterable, Seq) and __iterable.isStr) or (not isinstance(__iterable, GeneratorType) and all(isinstance(e, str) and len(e) == 1 for e in __iterable) and len(__iterable) > 0)
```
robin-aws, line 52: Definitely needs some line breaks for readability

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -39,57 +45,159 @@ def _break(name):
...
     def UniqueElements(self):
-        return set(self)
+        return frozenset(self)
 
     def __repr__(self) -> str:
         if self.isStr:
-          return ''.join(self)
-        return super().__repr__()
+            return ''.join(self)
+        return '[' + ', '.join(map(repr, self)) + ']'
```
robin-aws, line 65: Now that I look at this, `__repr__` should be an expression that recreates the Python `Seq` value, not something that looks roughly like the Dafny sequence display.

`__str__` can be whatever you like though, so that would be this version if desired.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -39,57 +45,159 @@ def _break(name):
...
+
+    def intersection(self, other):
+        return MultiSet(self & other)
+
+    def __sub__(self, other):
+        return MultiSet(super().__sub__(other))
+
+    @property
+    def keys(self):
+        return (key for key in self if self[key] > 0)
```
robin-aws, line 134: You need to override `__eq__` and `__hash__` for this class as well, since `Counter` doesn't consider a missing key equivalent to one mapped to `0`:

```python
>>> c = Counter()
>>> c[4] = 0
>>> c2 = Counter()
>>> c == c2
False
```

robin-aws (changes requested):
Nice! Can you remove `Feature.Multisets` from the `UnsupportedFeatures` set too? :)

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: No, that's how we handle it in the other targets as well.

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: Is supported by `SinglePassCompiler`.

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: `hash` is fine, as it uses keys, but I'll look into updating `__eq__`.

Test/comp/Collections.dfy
```diff
@@ -49,14 +50,109 @@ method Sets() {
...
+  }
+
+  method lt(a: T, b: T) returns (r: bool) {
+    var aInt := SetAsInt(a);
+    var bInt := SetAsInt(b);
+    r := |a| < |b| || (|a| == |b| && aInt <= bInt);
+  }
+}
+
+abstract module PrintOrderedSet {
```
robin-aws, line 90: This is nice btw, perhaps we can extract this into a shared location in the test suite later.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -1382,7 +1456,7 @@ protected class ClassWriter : IClassWriter {
     protected override void EmitSetBuilder_Add(CollectionType ct, string collName, Expression elmt, bool inLetExprBody,
         ConcreteSyntaxTree wr) {
       var wStmts = wr.Fork();
-      wr.WriteLine($"{collName}.add({Expr(elmt, inLetExprBody, wStmts)})");
+      wr.WriteLine($"{collName} = {collName}.union({DafnySetClass}([{Expr(elmt, inLetExprBody, wStmts)}]))");
```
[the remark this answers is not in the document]
fabiomadge: Sets are now immutable.

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: null actually expands to `_dafny.defaults.null`, which eventually returns `None`.

fabiomadge: I'm pretty sure, this is covered by the tests.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -197,4 +313,11 @@ def quantifier(vals, frall, pred):
...
 
 def AllChars():
-    return [chr(i) for i in range(0x10000)]
+    return (chr(i) for i in range(0x10000))
+
+class defaults:
+    bool = staticmethod(lambda: False)
+    int = staticmethod(lambda: 0)
+    real = staticmethod(BigRational)
+    null = staticmethod(lambda: None)
```
robin-aws, line 322: Consider "class" (or rather a legal munging of that), since this is a catch-all type descriptor used for any class. "null" sounds too much like a misspelling of `None` instead.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -460,12 +459,53 @@ protected class ClassWriter : IClassWriter {
...
+        UserDefinedType udt => udt.ResolvedClass switch {
+          TypeParameter tp => TypeParameterDescriptor(tp),
+          ClassDecl => "null",
+          TupleTypeDecl tpl => TupleTypeDescriptor(tpl, udt.TypeArgs, udt.tok),
+          NewtypeDecl => NewtypeDescriptor(udt),
+          _ => throw new cce.UnreachableException()
+        },
+        _ => throw new cce.UnreachableException()
+      };
+      return (customName ? "" : DafnyDefaults + ".") + name;
```
robin-aws, line 488: I think this is a bit too confusing to be worth it. Consider just duplicating the reference inline instead.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -460,12 +459,53 @@ protected class ClassWriter : IClassWriter {
...
+      WriteFormals(formals, isStatic && !NeedsCustomReceiver(member), ref sep, wr);
       return wr.NewBlockPy("):", close: BlockStyle.Newline);
     }
 
     protected override string TypeDescriptor(Type type, ConcreteSyntaxTree wr, IToken tok) {
-      throw new UnsupportedFeatureException(Token.NoToken, Feature.RuntimeTypeDescriptors);
+      Contract.Requires(type != null);
+      Contract.Requires(tok != null);
+      Contract.Requires(wr != null);
+
```
robin-aws, line 472: Perhaps worth a comment here that our representation of a type descriptor is just a lambda that produces the default value (as opposed to a separate type as in at least some target languages)

robin-aws (approved):

PR #2434, opened 2022-07-18
feat: Better LSP related location error messages

MikaelMayer:
Improvement: instead of "Related location", explicit what these related locations are about:
image

Fixes https://github.com/dafny-lang/ide-vscode/issues/211
On hovering a postcondition, the main message is now "This postcondition might not hold on a return path". https://github.com/dafny-lang/dafny/pull/2433 takes cares of changing "related location" to "return path".

Fixes https://github.com/dafny-lang/ide-vscode/issues/197 by changing all Uri to string conversion to the standard `Uri.toString() / DocumentUri.Parse()` so that there is no loss of conversion, and we don't assume a valid file name. Except for includes.

Fixes  https://github.com/dafny-lang/ide-vscode/issues/193 by providing which postconditions are failing explicitely

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+# Upcoming
+
+- feat: Snippet in the error messages of the language server. (https://github.com/dafny-lang/dafny/pull/2434)
+
 # 3.7.2
 
 - fix: Hovering over variables and methods inside cases of nested match statements work again. (https://github.com/dafny-lang/dafny/pull/2334)
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,6 @@
 # Upcoming
 
-- feat: Snippet in the error messages of the language server. (https://github.com/dafny-lang/dafny/pull/2434)
+- feat: *Less code navigation when verifying code*: In the IDE, failing postconditions and preconditions error messages now immediately display the sub-conditions that Dafny could not prove. Both on hover and in the Problems window. (https://github.com/dafny-lang/dafny/pull/2434)
 
 # 3.7.2
 
--- a/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
+++ b/Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
@@ -4,8 +4,10 @@ using OmniSharp.Extensions.LanguageServer.Protocol;
 using OmniSharp.Extensions.LanguageServer.Protocol.Models;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using VCGeneration;
 
 namespace Microsoft.Dafny.LanguageServer.Language {
@@ -27,7 +29,8 @@ namespace Microsoft.Dafny.LanguageServer.Language {
     /// <remarks>
     /// The uri of the entry document is necessary to report general compiler errors as part of this document.
     /// </remarks>
-    public DiagnosticErrorReporter(DocumentUri entryDocumentUri) {
+    public DiagnosticErrorReporter(string documentSource, DocumentUri entryDocumentUri) {
+      this.entryDocumentsource = documentSource;
       this.entryDocumentUri = entryDocumentUri;
     }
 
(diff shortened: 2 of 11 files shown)


RELEASE_NOTES.md
```diff
@@ -1,3 +1,7 @@
+# Upcoming
+
+- feat: Snippet in the error messages of the language server. (https://github.com/dafny-lang/dafny/pull/2434)
```
atomb, line 3: This description isn't entirely clear to me, and could probably do more to show how nice this is to have.

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -82,25 +85,66 @@ public class DiagnosticErrorReporter : ErrorReporter {
...
+    public static string FormatRelated(string related) {
+      return $"Could not prove: {related}";
+    }
+
+    private IEnumerable<DiagnosticRelatedInformation> CreateDiagnosticRelatedInformationFor(IToken token, string message) {
+      var (tokenForMessage, inner) = token is NestedToken nestedToken ? (nestedToken.Outer, nestedToken.Inner) : (token, null);
+      if (tokenForMessage is RangeToken range) {
+        var rangeLength = range.EndToken.pos + range.EndToken.val.Length - range.StartToken.pos;
+        if (message == PostConditionFailingMessage) {
+          var postcondition = entryDocumentsource.Substring(range.StartToken.pos, rangeLength);
```
atomb, line 100: Ah, nice, so you get the exact text from the document!

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -82,25 +85,66 @@ public class DiagnosticErrorReporter : ErrorReporter {
...
+          message = $"This postcondition might not hold: {postcondition}";
+        } else if (message == "Related location") {
+          // We can do better than that.
```
atomb, line 103: I agree. :)

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -82,25 +85,66 @@ public class DiagnosticErrorReporter : ErrorReporter {
...
+
+    public static string FormatRelated(string related) {
+      return $"Could not prove: {related}";
+    }
+
+    private IEnumerable<DiagnosticRelatedInformation> CreateDiagnosticRelatedInformationFor(IToken token, string message) {
+      var (tokenForMessage, inner) = token is NestedToken nestedToken ? (nestedToken.Outer, nestedToken.Inner) : (token, null);
+      if (tokenForMessage is RangeToken range) {
+        var rangeLength = range.EndToken.pos + range.EndToken.val.Length - range.StartToken.pos;
+        if (message == PostConditionFailingMessage) {
```
atomb, line 99: I'm not thrilled with this way of detecting the type of message, but I suspect it may be necessary for now, and that we'd have to do a lot of refactoring to be able to do it in a nicer way.

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -82,25 +85,66 @@ public class DiagnosticErrorReporter : ErrorReporter {
...
+          var postcondition = entryDocumentsource.Substring(range.StartToken.pos, rangeLength);
+          message = $"This postcondition might not hold: {postcondition}";
+        } else if (message == "Related location") {
+          // We can do better than that.
+          var tokenUri = tokenForMessage.GetDocumentUri();
+          var tokenUriSystemPath = tokenUri.GetFileSystemPath();
+          var entryDocumentUriSystemPath = entryDocumentUri.GetFileSystemPath();
+          if (tokenUri == entryDocumentUri || tokenUriSystemPath == entryDocumentUriSystemPath ||
+              tokenUriSystemPath == "/" + entryDocumentUriSystemPath) {
```
atomb, line 108: Oof. Is there a way we can normalize the system path to avoid having to check all of these cases?

atomb (changes requested):
Looks like a very useful feature! The main thing I'd change is to make it clearer in the changelog that it's a very useful feature. ;)

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
MikaelMayer: Oh right. I had been fighting to enable a simpler code but I forgot this case. Let me fix it.

RELEASE_NOTES.md
MikaelMayer: What about the new description?

atomb (approved):
Looks great!

PR #2439, opened 2022-07-18
Cok refman more small edits

davidcok:
Fixes various clarification edits in the RM


docs/DafnyRef/Types.md
```diff
@@ -870,18 +870,18 @@ of type `T`, integer-based numeric `i` satisfying `0 <= i < |s|`, and
...
- `e in s`            | `bool` | sequence membership
- `e !in s`           | `bool` | sequence non-membership
- `s[lo..hi]`         | `seq<T>`| subsequence
- `s[lo..]`           | `seq<T>` | drop
- `s[..hi]`           | `seq<T>` | take
- `s[`_slices_`]`   | `seq<seq<T>>` | slice
- `multiset(s)`       | `multiset<T>`| sequence conversion to a `multiset<T>`
+ expression         | precedence | result type | description
+ -------------------|:---:|:---:|----------------------------------------
+ `e in s`           | 4 | `bool` | sequence membership
```
MikaelMayer, line 875: Did you get these numbers from Parser.cs ? If they, they are subject to change as it's generated code.

docs/DafnyRef/UserGuide.md
```diff
@@ -287,21 +143,19 @@ The dafny tool terminates with these exit codes:
 Errors in earlier phases of processing typically hide later errors.
 For example, if a program has parsing errors, verification or compilation will
 not be attempted.
-The option `-countVerificationErrors:0` forces the tool to always end with a 0
```
MikaelMayer, line 290: Was this removed?

MikaelMayer (commented):
It feels always good to have a better structure, thanks.
Two questions

docs/DafnyRef/Types.md
davidcok: The numbers come from the reference manual itself. They are implicit in the tables of operations, but GitHub markdown does not render the horizontal rules that make the separation between operators clear. So I added a precedence number that is consistent across all the tables . It is possible that these could change with new operators being added, but they would have to have a whole new precedence group. So it would be rare. The Dafny parser builds precedence hard-coded into the grammar productions.

MikaelMayer (approved):

PR #2443, opened 2022-07-19
Constructor parenthesis warning

tobiaj10:
Fixes #
I implemented a new approach for a Warning for constructors


Source/IntegrationTests/IntegrationTests.csproj
```diff
@@ -41,6 +41,12 @@
         <Content Include="..\..\node_modules\**\*.*" LinkBase="TestFiles\LitTests\LitTest\node_modules">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </Content>
+        <Content Update="..\..\Test\linters\constructorCaseWithoutParenthesis.dfy">
```
MikaelMayer, line 44: Why do you need these lines? I haven't seen someone modify IntegrationTests.csproj in a while

Test/linters/constructorMatch.dfy.expect
```diff
@@ -0,0 +1,9 @@
+constructorMatch.dfy(18,12): Warning: this branch is redundant 
+constructorMatch.dfy(24,12): Warning: this branch is redundant 
+constructorMatch.dfy(32,16): Warning: this branch is redundant 
+constructorMatch.dfy(10,17): Warning: Constructor name 'nigeria' should contain a parenthesis at the end
```
MikaelMayer, line 4:
```suggestion
constructorMatch.dfy(10,17): Warning: Constructor 'nigeria()' requires parentheses
```

Test/linters/constructorMatch.dfy.expect
```diff
@@ -0,0 +1,9 @@
...
+constructorMatch.dfy(46,22): Warning: Constructor name 'Nigeria' should contain a parenthesis at the end
```
MikaelMayer, line 5:
```suggestion
constructorMatch.dfy(46,22): Warning: Constructor 'nigeria()' requires parentheses
```

Source/Dafny/Rewriter.cs
```diff
@@ -118,14 +120,61 @@ public abstract class IRewriter {
...
+    public ConstructorWarningVisitor(ErrorReporter reporter) {
+      this.reporter = reporter;
+    }
+    // Implements warning for constructors in match and ensures constructor name has a set parenthesis at the end 
+    protected override bool VisitOneExpr(Expression expr, ref Unit st) {
+      if (expr is NestedMatchExpr matchExpr) {
+        var matchExprCases = matchExpr.Cases;
+        foreach (var caseExpr in matchExprCases) {
+          if (caseExpr.Pat is IdPattern idPattern) {
+            var isConstructor = idPattern.Arguments != null;
```
MikaelMayer, line 161: How can you determine if something is a constructor there? I thought you would be looking in a dictionary of constructor name. I think the failing CI  tests come from false positives.

MikaelMayer (changes requested):
This is a great move forwards more precise errors. Keep going !

Source/Dafny/Rewriter.cs
keyboardDrummer: This checks whether the idPattern was determined to be a constructor by the resolver, in which case `idPattern.Arguments` is given a non-null value.

keyboardDrummer:
Since this PR implements a breaking change, I think it should be hidden behind a CLI flag. We can make turning it on the default for Dafny 4.0

Source/Dafny/Rewriter.cs
```diff
@@ -118,14 +120,61 @@ public abstract class IRewriter {
...
+            if (idPattern.HasParenthesis == false && isConstructor) {
+              this.reporter.Warning(MessageSource.Rewriter, idPattern.Tok,
+                $"Constructor name '{idPattern}' should contain a parenthesis at the end");
```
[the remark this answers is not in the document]
cpitclaudel: `requires` is not right, since the parentheses are not necessary (they are just a suggestion)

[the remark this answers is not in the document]
tobiaj10: I agree with clement because it is intended to be a suggestion

Test/linters/constructorMatch.dfy
```diff
@@ -0,0 +1,78 @@
...
+            case anythingElse => false;
+    }
+
+    function method MonochromaticFunction(c: Color) : bool {
+        match c
+            case ShadesOfGray(_) => true
+            case anythingElse => false
+    }
+    method MonochromaticMethodloop(c: Color) returns (x: bool)  {
+        while c == ShadesOfGray(3)
```
keyboardDrummer, line 63: What is the point of this while? There's no nested match expression

keyboardDrummer:
You should resolve the conflicts so that the regression tests can run again.

Test/linters/constructorCaseWithoutParenthesis.dfy
```diff
@@ -0,0 +1,87 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
```
keyboardDrummer, line 1: This test need to use `/noParenthesisConstructor`

Source/IntegrationTests/IntegrationTests.csproj
keyboardDrummer, line 44: Please remove these changes

Test/linters/constructorCaseWithoutParenthesis.dfy
```diff
@@ -0,0 +1,87 @@
...
+                case anythingElse => false;
+        }
+        return false; 
+    }
+   
+
+}
+
+module WithoutWarning {
+     datatype Color = Red | Green | ShadesOfGray(nat) | MixColors(nat, nat)
```
keyboardDrummer, line 40: Could you make the indentation in this file consistent? This line starts with 5 indentation. Most of this file uses indentation steps of 4, but I'd rather see indentation steps of 2.

Source/Dafny/DafnyOptions.cs
```diff
@@ -518,6 +519,11 @@ public enum IncludesModes {
           PrintTooltips = true;
           return true;
 
+        case "noParenthesisConstructor":
```
keyboardDrummer, line 522: What's still missing in DafnyOptions is an additional to the Dafny help command. Look for the text "---- Warning selection" and try to add something similar as what has been done for "warnShadowing". Speaking of "warnShadowing", it might be good to rename "noParenthesisConstructor" to "warnMissingConstuctorParenthesis"

Source/Dafny/Rewriter.cs
```diff
@@ -115,14 +118,69 @@ public abstract class IRewriter {
...
+            case Green => true
+                 ^^^ Warning: Constructor name 'Green' should contain a parenthesis at the end           
+            case anythingElse => false;
+        }
+        
+    * datatype Color = Red | Green | ShadesOfGray(nat)
+      method MonochromaticMethod(c: Color) returns (x: bool) {
+      return match c
+            case ShadesOfGray => true
+                  ^^^ Warning: Variable name 'ShadesOfGray' on line 129 starts with a capital letter
```
keyboardDrummer, line 136: This example is about capitals. It shouldn't be in this PR.

Source/Dafny/Rewriter.cs
```diff
@@ -115,14 +118,69 @@ public abstract class IRewriter {
...
+    }
+    public ConstructorWarning(ErrorReporter reporter) : base(reporter) {
+    }
+  }
+  class ConstructorWarningVisitor : TopDownVisitor<Unit> {
+    private readonly ErrorReporter reporter;
+    public ConstructorWarningVisitor(ErrorReporter reporter) {
+      this.reporter = reporter;
+    }
+    // Implements warning for constructors in match which ensures constructor name has parenthesis at the end and Begins with an uppercase letter.
```
keyboardDrummer, line 160: Capitals are mentioned on this line.

Source/Dafny/Rewriter.cs
```diff
@@ -115,14 +118,69 @@ public abstract class IRewriter {
       Contract.Requires(program != null);
     }
   }
-
+  /*
+   * Code Example that displays the Warnings
```
keyboardDrummer, line 122: Can you move all this new code to a new file?

Test/linters/constructorCaseWithoutParenthesis.dfy.expect
```diff
@@ -0,0 +1,11 @@
+constructorCaseWithoutParenthesis.dfy(9,6): Warning: this branch is redundant 
+constructorCaseWithoutParenthesis.dfy(10,6): Warning: this branch is redundant 
+constructorCaseWithoutParenthesis.dfy(15,6): Warning: this branch is redundant 
+constructorCaseWithoutParenthesis.dfy(16,6): Warning: this branch is redundant 
+constructorCaseWithoutParenthesis.dfy(24,9): Warning: this branch is redundant 
+constructorCaseWithoutParenthesis.dfy(25,9): Warning: this branch is redundant 
+constructorCaseWithoutParenthesis.dfy(9,11): Warning: Constructor name 'Green' should contain a parenthesis at the end
```
keyboardDrummer, line 7: I think you want to use `parentheses` instead of `a parenthesis`, to be plural instead of singular. More information: https://www.grammarbook.com/blog/commas/all-about-parentheses/#:~:text=The%20singular%20form%20is%20parenthesis,So%20let's%20keep%20it%20simple.

In particular, it says: "a parenthesis is one of a pair of curved marks that look like this: ( ), and parentheses are both marks."

Also, I find "at the end" is a little ambiguous. The end of what? And I think we can be more specific about when this warning occurs.

Something like "A constructor name in a case pattern should be followed by parentheses."

Source/Dafny/Rewriter.cs
```diff
@@ -115,14 +118,69 @@ public abstract class IRewriter {
...
+    }
+    // Implements warning for constructors in match which ensures constructor name has parenthesis at the end and Begins with an uppercase letter. 
+    protected override bool VisitOneExpr(Expression expr, ref Unit st) {
+      if (expr is NestedMatchExpr matchExpr) {
+        var matchExprCases = matchExpr.Cases;
+        foreach (var caseExpr in matchExprCases) {
+          if (caseExpr.Pat is IdPattern idPattern) {
+            var isConstructor = idPattern.Arguments != null;
+            var isVariable = idPattern.Arguments == null;
+            if (idPattern.HasParenthesis == false && isConstructor) {
```
keyboardDrummer, line 168: `idPattern.HasParenthesis == false` can be rewritten to `!idPattern.HasParenthesis`

Source/Dafny/Rewriter.cs
```diff
@@ -115,14 +118,69 @@ public abstract class IRewriter {
...
+      this.reporter = reporter;
+    }
+    // Implements warning for constructors in match which ensures constructor name has parenthesis at the end and Begins with an uppercase letter. 
+    protected override bool VisitOneExpr(Expression expr, ref Unit st) {
+      if (expr is NestedMatchExpr matchExpr) {
+        var matchExprCases = matchExpr.Cases;
+        foreach (var caseExpr in matchExprCases) {
+          if (caseExpr.Pat is IdPattern idPattern) {
+            var isConstructor = idPattern.Arguments != null;
+            var isVariable = idPattern.Arguments == null;
```
keyboardDrummer, line 167: The variable `isVariable` seems unused.

Source/Dafny/ConstructorWarning.cs
```diff
@@ -0,0 +1,53 @@
...
+  // Implements warning for constructors in match which ensures constructor is followed by parenthesis. 
+  protected override bool VisitOneExpr(Expression expr, ref Unit st) {
+    if (expr is NestedMatchExpr matchExpr) {
+      var matchExprCases = matchExpr.Cases;
+      foreach (var caseExpr in matchExprCases) {
+        if (caseExpr.Pat is IdPattern idPattern) {
+          var isConstructor = idPattern.Arguments != null;
+          if (!idPattern.HasParenthesis && isConstructor) {
+            this.reporter.Warning(MessageSource.Rewriter, idPattern.Tok,
+              $"Constructor name '{idPattern}' should be followed by parenthesis");
```
keyboardDrummer, line 46: `parenthesis` is still singular.

Source/Dafny/DafnyOptions.cs
```diff
@@ -1082,6 +1082,8 @@ is printed.
 
 /warnShadowing  Emits a warning if the name of a declared variable caused another variable
     to be shadowed
+/warnMissingConstructorParenthesis Emits a warning when a constructor name is not followed
```
keyboardDrummer, line 1085: I think we should be more specific, "when a constructor name in a case pattern is not followed by parentheses", note the plural "parentheses" as well.

keyboardDrummer (dismissed):

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -12434,6 +12434,7 @@ public class LitPattern : ExtendedPattern {
   }
 
   public class IdPattern : ExtendedPattern {
+    public bool HasParenthesis { get; }
```
MikaelMayer, line 12437: According to the documentation of `Arguments`, you should be able to have:
```suggestion
    public bool HasParenthesis => Arguments != null
```

That way you can remove `HasParenthesis` from the constructor

Source/Dafny/ConstructorWarning.cs
```diff
@@ -0,0 +1,53 @@
+using System.Linq;
+using System.Reactive;
+using Microsoft.Dafny;
+
+/*
+   * Code Example that displays the Warnings
+   *
+   * datatype Color = Red | Green | ShadesOfGray(nat)
```
MikaelMayer, line 8: For your example below, besides the option `/warnMissingConstructorParenthesis` could you also always add the warning in the case when:
```suggestion
   * datatype Color = Red | Green() | ShadesOfGray(nat)
```
then there is ALWAYS a warning if there are missing parentheses in pattern `Green` ?

Source/Dafny/ConstructorWarning.cs
```diff
@@ -0,0 +1,53 @@
...
+  private readonly ErrorReporter reporter;
+  public ConstructorWarningVisitor(ErrorReporter reporter) {
+    this.reporter = reporter;
+  }
+  // Implements warning for constructors in match which ensures constructor is followed by parenthesis. 
+  protected override bool VisitOneExpr(Expression expr, ref Unit st) {
+    if (expr is NestedMatchExpr matchExpr) {
+      var matchExprCases = matchExpr.Cases;
+      foreach (var caseExpr in matchExprCases) {
+        if (caseExpr.Pat is IdPattern idPattern) {
```
MikaelMayer, line 42: Can you please traverse the patterns recursively, so that it's not just a top-level pattern rule?

MikaelMayer (changes requested):
Great job! I feel better with this PR.
I require at least to fix the recursivity on patterns.

Source/Dafny/AST/DafnyAst.cs
keyboardDrummer: `Arguments != null` means it's interpreted as a constructor by Dafny, but it may still not have any parenthesis in the syntax, which is exactly when this warning is shown.

Source/Dafny/ConstructorWarning.cs
keyboardDrummer: This implies a bit more work, and I suggest we leave that for another PR for the sake of time.

Source/Dafny/Rewriter.cs
```diff
@@ -4,6 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
```
keyboardDrummer, line 6: The changes in this file don't seem required. Could you undo them?

Source/Dafny/DafnyOptions.cs
```diff
@@ -518,6 +519,11 @@ public enum IncludesModes {
           PrintTooltips = true;
           return true;
 
+        case "warnMissingConstructorParenthesis":
+          DisallowConstructorCaseWithoutParenthesis = true;
+          return true;
+
```
keyboardDrummer, line 525: There's an unnecessary newline here.

Source/Dafny/ConstructorWarning.cs
```diff
@@ -0,0 +1,63 @@
...
+      if (isConstructor) {
+        foreach (var nestedPattern in idPattern.Arguments) {
+          CheckPattern(nestedPattern);
+        }
+        if (!idPattern.HasParenthesis) {
+          this.reporter.Warning(MessageSource.Rewriter, idPattern.Tok,
+            $"Constructor name '{idPattern}' should be followed by parentheses");
+        }
+      }
+
```
keyboardDrummer, line 59: Whiteline can be removed.

Source/Dafny/ConstructorWarning.cs
```diff
@@ -0,0 +1,63 @@
...
+    if (expr is NestedMatchExpr matchExpr) {
+      var matchExprCases = matchExpr.Cases;
+      foreach (var caseExpr in matchExprCases) {
+        CheckPattern(caseExpr.Pat);
+      }
+    }
+    return base.VisitOneExpr(expr, ref st);
+  }
+  private void CheckPattern(ExtendedPattern pattern) {
+    if (pattern is IdPattern idPattern) {
```
keyboardDrummer, line 48: Could you invert this `if` condition to reduce nesting? There should be a quick action for that in Rider.

keyboardDrummer (approved):

MikaelMayer (approved):
Looks great, thanks a lot for the effort!

PR #2444, opened 2022-07-19
feat: Add #nullable disable to generated C# code to avoid warnings

### LINKED ISSUE #2386: Add #nullable disable to generated Dafny files to avoid warnings

cpitclaudel:
Closes #2386.


MikaelMayer (approved):
Feels like someone is using compiled c# code for a real project :-)

atomb:
I just noticed that this leads to generating code that fails to compile with projects requiring a C# version less than 8.0, such as https://github.com/aws/aws-encryption-sdk-dafny.

cpitclaudel:
Ah, snap, let's revert this.

PR #2446, opened 2022-07-19
feat: Don't force :extern members to be implemented in derived classes

### LINKED ISSUE #2212: Feature request: ignore {:extern} method in trait
Currently Dafny forces class `C` to implement trait method `T.m`, even though we specified it as implemented externally. 
```
trait T {
  method {:extern} m() returns (res: int)
}

class C extends T {
  method n() returns (res: int)  {
    var v := m();
    return v;
  }
}
```
One work around is adding `method {:extern} m() returns (res: int)` to class `C` as well, but it's a pain if we have massive classes and need to do this over and over.

cpitclaudel:
Closes #2212.


Source/Dafny/Resolver.cs
```diff
@@ -9385,6 +9385,9 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
               // member is a field or a fully defined function or method
             } else if (cl is TraitDecl) {
               // there are no expectations that a field needs to repeat the signature of inherited body-less members
+            } else if (Attributes.Contains(member.Attributes, "extern")) {
+              // Extern functions do not need to be reimplemented.
+              // TODO: When `:extern` is separated from `:compile false`, this should become `:compile false`.
```
fabiomadge, line 9390: Should we add this to the SinglePassCompiler as well?

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -1786,13 +1786,17 @@ protected class NullClassWriter : IClassWriter {
               EmitSetterParameter(sw);
             }
           } else if (member is Function fn) {
-            Contract.Assert(fn.Body != null);
-            var w = classWriter.CreateFunction(IdName(fn), CombineAllTypeArguments(fn), fn.Formals, fn.ResultType, fn.tok, fn.IsStatic, true, fn, true, false);
-            EmitCallToInheritedFunction(fn, w);
+            if (!Attributes.Contains(fn.Attributes, "extern")) {
```
fabiomadge, line 1789: Can we merge that with L1788?

Test/comp/ExternCopyFromTrait.dfy
```diff
@@ -1,22 +1,86 @@
...
+/// Hence, the only use of this feature in C# is interfaces with explicit
+/// implementations.  The same problem means that this feature is not very
+/// useful for Java either: Dafny translates the `extends` keywords to Java's
+/// `implements`, not `extends`, and so `Asker` below must be an interface in
+/// Java.
+
+module {:extern "M1"} M1 {
+  trait {:extern "Asker"} {:compile false} Asker {
+    method Exclaim() // An abstract method, to be overridden
+    method {:extern} Speak() // A method implemented by the base class
```
fabiomadge, line 52:
```suggestion
    method {:extern} Speak() // A method implemented by the trait
```

Test/comp/ExternCopyFromTrait.dfy
```diff
@@ -1,22 +1,86 @@
...
+/// still want to model external C#-style inheritance.  To do so we declare a
+/// trait corresponding to a C# interface, list all of its methods, and annotate
+/// non-abstract ones as using `:extern`.  This lets Dafny know that it
+/// shouldn't insist on such methods being re-implemented in derived classes.
+///
+/// Note that the above mentions "interfaces" when talking about C#.  That is
+/// because the feature below is not flexible enough to accommodate abstract
+/// classes, as it does not add the `override` keyword on implementations.
+/// Hence, the only use of this feature in C# is interfaces with explicit
+/// implementations.  The same problem means that this feature is not very
```
fabiomadge, line 44:
```suggestion
/// implementations. The same problem means that this feature is not very
```

fabiomadge (dismissed):

Source/Dafny/Compilers/SinglePassCompiler.cs
cpitclaudel: No: for functions with the `:extern` attribute, we would fall into the Contract.Assert case instead of doing nothing

fabiomadge: Aren't
```c#
else if (member is Function fn) {
            if (!Attributes.Contains(fn.Attributes, "extern")) {
```
and
```c#
else if (member is Function fn && !Attributes.Contains(fn.Attributes, "extern")) {
```
semantically equivalent?

cpitclaudel: No, not when there's another branch below:

```
if (a) {
  if (b) {
    print "ab";
  }
} else {
  print Boom;
}
```

When `a = true, b = false`, the above exits silently, whereas the following prints `"Boom"`


```
if (a &&b) {
  print "ab";
} else {
  print Boom;
}
```

cpitclaudel: image

fabiomadge: Oh, I see what you mean now. Sorry!

fabiomadge (approved):

PR #2457, opened 2022-07-19
ci: Build and release process cleanups after 3.7.3

atomb:
* Update release process documentation
* Include license expressions in packages for NuGet, to avoid warnings
* Always upload `.nupkg` files as artifacts, for debugging

### DIFF
--- a/.github/workflows/publish-release-reusable.yml
+++ b/.github/workflows/publish-release-reusable.yml
@@ -124,7 +124,6 @@ jobs:
 
     - name: Make NuGet package available as an artifact
       uses: actions/upload-artifact@v2
-      if: always()
       with:
         name: nuget-packages-${{ matrix.os }}
         path: dafny/Binaries/*.nupkg
--- a/.github/workflows/publish-release-reusable.yml
+++ b/.github/workflows/publish-release-reusable.yml
@@ -122,6 +122,13 @@ jobs:
       # NuGet will consider any package with a version-suffix as a prerelease
       run: dotnet pack --version-suffix ${{ inputs.name }} --no-build dafny/Source/Dafny.sln
 
+    - name: Make NuGet package available as an artifact
+      uses: actions/upload-artifact@v2
+      if: always()
+      with:
+        name: nuget-packages-${{ matrix.os }}
+        path: dafny/Binaries/*.nupkg
+
     - name: Upload package to NuGet
       if: ${{ inputs.release_nuget }}
       # Need --skip-duplicate for cases where we release a new Dafny
--- a/.github/workflows/xunit-tests.yml
+++ b/.github/workflows/xunit-tests.yml
@@ -24,8 +24,8 @@ jobs:
             z3: z3-4.8.5-x64-win
             chmod: false
             coverage: false
-          - os: macos-10.15
-            suffix: osx-10.14.1
+          - os: macos-11
+            suffix: osx-11
             z3: z3-4.8.5-x64-osx-10.14.2
             chmod: true
             coverage: false
--- a/.github/workflows/xunit-tests.yml
+++ b/.github/workflows/xunit-tests.yml
@@ -24,8 +24,8 @@ jobs:
             z3: z3-4.8.5-x64-win
             chmod: false
             coverage: false
-          - os: macos-11
-            suffix: osx-11
+          - os: macos-latest
+            suffix: osx
             z3: z3-4.8.5-x64-osx-10.14.2
             chmod: true
             coverage: false
--- a/docs/dev/RELEASE.md
+++ b/docs/dev/RELEASE.md
@@ -153,14 +153,6 @@ with git commands and concepts is helpful.
 
 5. Before releasing a new version of the VSCode plugin, make sure to add to the release notes in `CHANGELOG.md`
 
-6. Select a new version number `$VER` for the plugin (e.g. "2.4.0") and change the value of the constant `version` in the file `package.json`
-   to `$VER` and commit the changes.
+6. Follow the release process detailed in the plugin's repository:
 
-7. Then tag and push the changes:
-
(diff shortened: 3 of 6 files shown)


.github/workflows/publish-release-reusable.yml
```diff
@@ -122,6 +122,13 @@ jobs:
       # NuGet will consider any package with a version-suffix as a prerelease
       run: dotnet pack --version-suffix ${{ inputs.name }} --no-build dafny/Source/Dafny.sln
 
+    - name: Make NuGet package available as an artifact
+      uses: actions/upload-artifact@v2
+      if: always()
```
MikaelMayer, line 127: Why do you need this line?

docs/dev/RELEASE.md
```diff
@@ -98,64 +107,36 @@ with git commands and concepts is helpful.
...
         brew reinstall dafny
 
    and then execute `dafny` on a sample file to see if it has the
    correct version number. Even better is to try this step on a
    different machine than the one on which the `dafny.rb` file was edited
 
-9. If everything works you can, at your leisure do
-
-        git branch -d <the branch name>
-
```
MikaelMayer, line 158: Why deleting this comment? I thought it was useful to know you can delete a local branch.

docs/dev/RELEASE.md
```diff
@@ -98,64 +107,36 @@ with git commands and concepts is helpful.
...
 ## Updating and releasing a new version of VSCode plugin
```
MikaelMayer, line 140: I created the release process of the VSCode plugin here because I did not know it was part of the Dafny release process
https://github.com/dafny-lang/ide-vscode/blob/master/CONTRIBUTING.md

Nothing to do right now, but we should consider putting the information at only one place (probably the VSCode repository)

.github/workflows/publish-release-reusable.yml
atomb: We probably don't!

docs/dev/RELEASE.md
atomb: The new process I described uses a tool that avoids the need for a local branch (or, I believe, creates one and then deletes it, automatically), so there's nothing to delete.

docs/dev/RELEASE.md
atomb: Ah, good to know! I think it might be worth replacing that whole section with a link to the doc in the `ide-vscode` repo.

docs/dev/RELEASE.md
```diff
@@ -153,14 +153,6 @@ with git commands and concepts is helpful.
...
-6. Select a new version number `$VER` for the plugin (e.g. "2.4.0") and change the value of the constant `version` in the file `package.json`
-   to `$VER` and commit the changes.
+6. Follow the release process detailed in the plugin's repository:
 
-7. Then tag and push the changes:
-
-       git tag v<$VER>
-       git push origin v<$VER>
-
-8. Create a pull request and, once merged, a GitHub action will automatically run in reaction to the tag being pushed to master, which will create a new release in the VSCode marketplace.
```
MikaelMayer, line 183: The new CONTRIBUTING.md gives a workflow that works even if there are multiple commits in this new branch (i.e. update the changelog), so I feel better doing this link. Thanks !

MikaelMayer (dismissed):
Great cleanup !

docs/dev/RELEASE.md
atomb: Yeah, I followed the text that I just deleted for the 3.7.3 release (even though I thought it seemed weird to create a tag on the branch before merging it) and it led to a release being pushed to the marketplace before reviewers had approved the PR, which wasn't what I wanted. Following the instructions in the `ide-vscode` repo would have avoided that.

docs/dev/RELEASE.md
```diff
@@ -98,64 +107,36 @@ with git commands and concepts is helpful.
...
+3. Expect comments from the reviewers. If changes are needed, do 4-6
+   again. Eventually the reviewers will accept and merge the PR.
 
-8. Test the installation by running
+4. Test the installation by running
 
         brew reinstall dafny
 
    and then execute `dafny` on a sample file to see if it has the
    correct version number. Even better is to try this step on a
```
robin-aws, line 137: While we're touching this, let's do `dafny /version` instead :)

.github/workflows/xunit-tests.yml
```diff
@@ -24,8 +24,8 @@ jobs:
             z3: z3-4.8.5-x64-win
             chmod: false
             coverage: false
-          - os: macos-10.15
-            suffix: osx-10.14.1
+          - os: macos-11
+            suffix: osx-11
             z3: z3-4.8.5-x64-osx-10.14.2
```
davidcok, line 29: This change should work, but I note that publish_release_rusable.yml uses macOS-latest. Should we do that here?

Also note that there is no build of Z3 4.8.5 for a more recent macOS. I hope it does not stop working with the deprecation of 10.14.

atomb: Yeah, we could go to `macos-latest`, which is currently `macos-11`. I'll do that.

I've generally had good luck running old binaries on newer versions of macOS, so I'm not too worried at the moment. And that's the same Z3 binary I'm currently using on my macOS 12 laptop.

davidcok (approved):

atomb:
Oh, bother, it's now requiring a job that no longer exists (since it probably doesn't make sense to call it `osx-10.14.1` anymore). Someone with admin access to the repo will need to change that. @robin-aws?

davidcok:
I just removed that one from the required list and saved the change, though I'm not sure why I was allowed to

atomb:
> I just removed that one from the required list and saved the change, though I'm not sure why I was allowed to

Thanks!

PR #2458, opened 2022-07-20
Adds text for various TODOs in the RM

davidcok:
Adds text for various TODOs: reveal statements, ORDINALS, type parameter inference, type parameter variance, real literals with exponents not supported


docs/DafnyRef/Expressions.md
```diff
@@ -913,7 +913,24 @@ var k := 10 / x; // error, may divide by 0.
 var m := if x != 0 then 10 / x else 1; // ok, guarded
 ```
 
-TO BE WRITTEN - binding form
+The `if` expression also permits a binding form.
+In this case the condition of the if is an existential asking
```
MikaelMayer, line 917:
```suggestion
In this case the condition of the `if` is an existential asking
```

docs/DafnyRef/Expressions.md
```diff
@@ -913,7 +913,24 @@ var k := 10 / x; // error, may divide by 0.
...
+"does there exist a value satisfying a given predicate?".
```
MikaelMayer, line 918:
```suggestion
"does there exist a value satisfying the given predicate?".
```

docs/DafnyRef/Expressions.md
```diff
@@ -913,7 +913,24 @@ var k := 10 / x; // error, may divide by 0.
...
+If not, the else branch is evaluated. But if so, then an
+(arbitrary) value that does satisfy the given predicate is
+bound to the given variable and that variable is in scope in 
+the then-branch of the expression.
+
+For example, in the expression
+```
```
MikaelMayer, line 925:
```suggestion
```dafny
```
I'm not sure it changes anything but it could help in the future for syntax highlighting.

docs/DafnyRef/Expressions.md
```diff
@@ -913,7 +913,24 @@ var k := 10 / x; // error, may divide by 0.
...
+if x: int :| x*x == s then x else 0
+```
+if `s` is indeed a square, say `25`, then x is bound to either `5` or `-5`
```
MikaelMayer, line 928: From an engineer's perspective where everything is deterministic, this is weird. My interpretation is that the value is either 5 or -5 in ghost code so that it's just a predicate, but what happens in non-ghost code? Can you write an explanation?

Also, could you perhaps use another example than something non-linear? Non-linearity give headaches to z3 in general.
For example, I suggest:

```dafny
predicate P(x: int) {
  x == 5 || x == -5
}
method main() {
  assert P(5);
  var y := if x: int :| P(x) then x else 0;
  assert y == 5 || y == -5;
}
```

Here if you change `x: int` into `x: nat`, you can remove the `y == -5`;
In the case above, if I add a range, then the `if` expression is no longer ghost:
```dafny
 var y := if x: nat :| 0 <= x < 10 && P(x) then x else 0;
```
Also, since the expression hides an existential quantifier, one needs to write a trigger to prove anything concerning its value, that's why I wrote `assert P(5);` above, without which it fails.

docs/DafnyRef/Statements.md
```diff
@@ -1537,7 +1537,14 @@ much as lemmas might be used in mathematical proofs.
 
 Using `...` as the argument of the statement is part of module refinement, as described in [Section 22](#sec-module-refinement).
 
-TO BE WRITTEN - assert by statements
+In the `by` form of the `assert` statement, there is an additional block of statements.
+Those statements are typically a sequence of [lemmas](#sec-lemmas) or a [calc statement](#sec-calc-statement).
```
MikaelMayer, line 1541: Also, I found it typical to assert other assertions (when refactoring proofs), ghost if statements, and reveal statements.

Also, could you give an example of an assert..by statement?

docs/DafnyRef/Statements.md
```diff
@@ -1537,7 +1537,14 @@ much as lemmas might be used in mathematical proofs.
 
 Using `...` as the argument of the statement is part of module refinement, as described in [Section 22](#sec-module-refinement).
 
-TO BE WRITTEN - assert by statements
+In the `by` form of the `assert` statement, there is an additional block of statements.
```
MikaelMayer, line 1540:
```suggestion
In the `by` form of the `assert` statement, there is an additional block of statements that provide the Dafny verifier with additional proof steps.
```

docs/DafnyRef/Statements.md
```diff
@@ -1739,8 +1746,92 @@ RevealStmt =
     ";"
 ````
 
+The `reveal` stastement makes available to the solver information that is otherwise opaque (that is, not visible).
```
MikaelMayer, line 1749: Can you add a link to opaque?

docs/DafnyRef/Statements.md
```diff
@@ -1739,8 +1746,92 @@ RevealStmt =
...
+```
+The first assertion fails. Without the label `x:`, the second would succeed because after a failing assertion, the 
+assertion is assumed in the context of the rest of the program.  But with the label, the first assertion is hidden from
+the rest of the program. That assertion can be _revealed_ by adding a `reveal` statement:
+
+```dafny
+method m(i: int) {
+  assert x: i == 0; // Fails
+  reveal x;
+  assert i == 0; // Now succeeds
```
MikaelMayer, line 1769: Likewise to your example of hiding preconditions, can you use `assert i == 0 by { reveal x; }` or describe that it is possible to scope the reveal using this method?

docs/DafnyRef/Statements.md
```diff
@@ -1739,8 +1746,92 @@ RevealStmt =
...
+  assert x < z by { reveal L; }
+}
+```
+
+The reason to possibly hide a precondition is the same as the reason to hide assertions: 
+sometimes less information is better for the solver as it helps the solver focus attention on 
+relevant information.
+
+Section 7 of [http://leino.science/papers/krml276.html](http://leino.science/papers/krml276.html) provides 
+an extended illustration of this technique.
```
MikaelMayer, line 1813:
```suggestion
an extended illustration of this technique to make all the dependencies of an `assert` explicit.
```

docs/DafnyRef/Statements.md
```diff
@@ -1739,8 +1746,92 @@ RevealStmt =
...
+
+### 20.20.3. Revealing function bodies
+
+Normally function bodies are transparent and available for constructing proofs of assertions that use those functions.
+However, sometimes it is helpful to mark a function _opaque_ and treat it as an uninterpreted function, whose properties are
```
MikaelMayer, line 1818:
```suggestion
However, sometimes it is helpful to mark a function [`{:opaque}`](#sec-opaque) and treat it as an uninterpreted function, whose properties are
```

docs/DafnyRef/Statements.md
```diff
@@ -1739,8 +1746,92 @@ RevealStmt =
...
+just its specifications.  This action limits the information available to the logical reasoning engine and may make a proof 
+possible where there might be information overload otherwise.
+
+But then there may be specific instances where the definition of that opaque function is needed. In that situation, the
+body of the function can be _revealed_ using the reveal statement. Here is an example:
+```dafny
+function {: opaque} f(i: int): int { i + 1 }
```
MikaelMayer, line 1825:
```suggestion
function {:opaque} f(i: int): int { i + 1 }
```
The style used until now is to not put a space between the colon and the attribute name.

docs/DafnyRef/Statements.md
```diff
@@ -1739,8 +1746,92 @@ RevealStmt =
...
+
+method m(int i) {
+  assert f(i) == i + 1;
+}
+```
+Without the `{:opaque}` attribute, the assertion is valid; with the attribute it cannot be proved because the body if the
```
MikaelMayer, line 1831:
```suggestion
Without the [`{:opaque}`](#sec-opaque) attribute, the assertion is valid; with the attribute it cannot be proved because the body if the
```

docs/DafnyRef/Statements.md
```diff
@@ -1739,8 +1746,92 @@ RevealStmt =
...
+function is not visible. However if a `reveal f();` statement is inserted before the assertion, the proof succeeds.
+Note that the psuedo-function-call in the `reveal' statement is written without arguments.
```
MikaelMayer, line 1833: Excellent remark !

docs/DafnyRef/Topics.md
```diff
@@ -2,9 +2,37 @@
...
+```
+In the latter case, Dafny knows that the already defined types `set` and `List` each take one type parameter
+so it fills in `<T>` (using some unique type parameter name) and then determines the the function itself needs
+a type parameter `<T>` also.
+
+Dafny also accepts
+```dafny
+function Elements<T>(list: List): set
+```
+In this case, the function already has a type parameter list. `List and `set` are each known to need type parameters,
```
MikaelMayer, line 27:
```suggestion
In this case, the function already has a type parameter list. `List` and `set` are each known to need type parameters,
```

docs/DafnyRef/Types.md
```diff
@@ -232,6 +232,7 @@ There are also literals for some of the reals.  These are
 written as a decimal point with a nonempty sequence of decimal digits
 on both sides, optionally prefixed by a `-` character.
 For example, `1.0`, `1609.344`, `-12.5`, and `0.5772156649`.
+Real literals using exponents are not supported in Dafny (at present).
```
MikaelMayer, line 235:
```suggestion
Real literals using exponents are not supported in Dafny. For now, you'd have to write your own function for that, e.g. 
```dafny
// realExp(2.37, 100) computes 2.37e100
function method realExp(r: real, e: int) {
  if e == 0 then r
  else if e < 0 then realExp(r/10, e+1)
  else realExp(r*10, e)
}
```
```

File: docs/DafnyRef/Types.md [hunk 16]
```diff
@@ -420,7 +421,19 @@ But `11` is not a valid `bv3` literal.
...
+- a value of type `ORDINAL` may be explicitly converted to a `nat` using `as nat` if it is known to be less than some `nat` value 
+- non-negative numeric literals may be considered `ORDINAL` literals
+- `ORDINAL`s may be compared, using `== != < <= > >=`
+- two `ORDINAL`s may be added and the result is `>=` either one of them
+- two `ORDINAL`s may be subtracted if the LHS is `>=` the RHS (so the result is non-negative) and the RHS value is equal to some `nat` value. That is, for `ORDINAL` `x`, `assert x - x == 0;` fails
+ but `assert x < 1000000000 ==> x-x == 0;` succeeds.
+- `*`, `/` and `%` are not defined for `ORDINAL`s
+- there is no literal in Dafny that represents the `ORDINAL` value that is larger than any `nat` (typically written $\omega$ in mathematics)
+
+In Dafny, `ORDINAL`s are used primarily in conjunction with extreme functions and lemmas.
```
MikaelMayer, line 436: Can you please add a link to extreme functions and lemmas?

docs/DafnyRef/Types.md
```diff
@@ -633,7 +646,26 @@ Here are some examples:
...
+`+`      | co-variant | yes
+`-`      | contra-variant | not necessarily
+`*`      | co-variant | not necessarily
+`!`      | non-variant | not necessarily
+
+- _co-variance_ (`A<+T>` or `A<*T>`) means that if `U` is a subtype of `V` then `A<U>` is a subtype of `A<V>`
+- _contra-variance_ (`A<-T>`) means that if `U` is a subtype of `V` then `A<V>` is a subtype of `A<U>`
+- _non-variance_ (`A<T>` or `A<!T>`)  means that if `U` is a different type than `V` then there is no subtyping relationship between `A<U>` and `A<V>`
+
+_Cardinality preserving_ has to do with whether the type being defined may be referred to in the type argument.
```
MikaelMayer, line 666:
```suggestion
_Cardinality preserving_ means that the cardinal of the set of all values denoted by the type parameter is not strictly less than the set of all values being defined by the type using this type parameter.
For example

    type T<X> = X -> bool

is illegal and returns the error message `formal type parameter 'X' is not used according to its variance specification (it is used left of an arrow) (perhaps try declaring 'X' as '!X')`
The meaning of that is there are strictly more predicate on X than X itself, which [could cause soundness issues](http://leino.science/papers/krml280.html).

To fix it, we use the variance `!`:

    type T<!X> = X -> bool

This states that `T` does not preserve the cardinality of `X`, meaning there could be strictly more values of type `T<E>` than `E` for any E.
```

MikaelMayer (changes requested):
Wow, these topics were awaited by users for a long time, thanks for this detailed write-up !
I added a few suggestions.

docs/DafnyRef/Types.md
davidcok: Done

docs/DafnyRef/Statements.md
davidcok: Reworded

docs/DafnyRef/Expressions.md
davidcok: Used your example and reworded.

docs/DafnyRef/Types.md
```diff
@@ -232,7 +232,14 @@ There are also literals for some of the reals.  These are
...
 on both sides, optionally prefixed by a `-` character.
 For example, `1.0`, `1609.344`, `-12.5`, and `0.5772156649`.
-Real literals using exponents are not supported in Dafny (at present).
+Real literals using exponents are not supported in Dafny. For now, you'd have to write your own function for that, e.g. 
+```dafny
+// realExp(2.37, 100) computes 2.37e100
+function method realExp(r: real, e: int) {
+  if e == 0 then r
+  else if e < 0 then realExp(r/10, e+1)
+  else realExp(r*10, e)
```
MikaelMayer, line 241:
```suggestion
function method realExp(r: real, e: int): real decreases if e > 0 then e else -e {
  if e == 0 then r
  else if e < 0 then realExp(r/10.0, e+1)
  else realExp(r*10.0, e-1)
}
```
I tried my previous function in Dafny and it had several typo and one verification error (the function was not terminating on the final else branch)

MikaelMayer (approved):
Let's merge this ! Thank you very much for contributing to the documentation.

davidcok:
Merged - fortunately the non-deterministic tests decided to pass

Issue #2350, opened 2022-07-01
Counterexample generation fails: integer overflow


sorawee:
Version 3.7.0 (Dafny VSCode extension version 2.5.0)

Consider:

```
ghost const NAT_MAX := 0x7fff_ffff
ghost const NAT64_MAX := 0x7fff_ffff_ffff_ffff

newtype nat32 = x | 0 <= x <= NAT_MAX
newtype nat64 = x | 0 <= x <= NAT64_MAX

function method plus(a: nat32, b: nat32): nat32 {
    var al := a as nat64;
    var bl := b as nat64;
    a + b
}
```

The counterexample generation errors with:

```
CounterExample request failed: Error: Internal Error - System.OverflowException: Value was either too large or too small for an Int32.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at Microsoft.Boogie.Model.Number.AsInt()
   at DafnyServer.CounterexampleGeneration.DafnyModel.RegisterReservedInts() in /Users/soraweep/.vscode/extensions/dafny-lang.ide-vscode-2.5.0/out/resources/3.7.0/custom/arm64/dafny/Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs:line 191
   at DafnyServer.CounterexampleGeneration.DafnyModel..ctor(Model model) in /Users/soraweep/.vscode/extensions/dafny-lang.ide-vscode-2.5.0/out/resources/3.7.0/custom/arm64/dafny/Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs:line 93
   at Microsoft.Dafny.LanguageServer.Handlers.Custom.DafnyCounterExampleHandler.CounterExampleLoader.GetLanguageSpecificModel(Model model) in /Users/soraweep/.vscode/extensions/dafny-lang.ide-vscode-2.5.0/out/resources/3.7.0/custom/arm64/dafny/Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs:line 79
[...]

PR #2461, opened 2022-07-21
Counterexamples - fix integer parsing bug and improve code quality

Dargones:
This pull request does not add any new functionality to counterexamples but instead makes several code quality improvements and bug fixes. Here is a summary:

- Correctly handle unbounded integers, sequence indices, etc. (fixes https://github.com/dafny-lang/dafny/issues/2350)
- Use the native `Microsoft.Dafny.Type` in place of `DafnyModelType` to improve readability and reduce the amount of string processing.
- Escape special characters when outputting the counterexample and use UTF codes for characters outside the 32 to 126 range (this prevents some parsing issues and also disambiguates between similarly looking characters such as semicolon and greek question mark) 
- Fix edge cases in object field name and datatype destructor name identification
- Remove dead code left from the original counterexample tool

I also added relevant tests and made a change to test generation to support the move from DafnyModelType to Microsoft.Dafny.Type


Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
```diff
@@ -1003,61 +1003,74 @@ class Cla__ss {
...
+        datatype D = C(i:int) {{ 
+          predicate p() {{true}}
+        }}
+
+        method test(d: D) {{
+          if (d.p()) {{
+            assert d.i != 123;
+          }}
+        }}
+      }}".TrimStart();
```
MikaelMayer, line 1039:
```suggestion
      var source = @"
      module M {
        datatype D = C(i:int) {
          predicate p() {true}
        }

        method test(d: D) {
          if (d.p()) {
            assert d.i != 123;
          }
        }
      }".TrimStart();
```
Can you please remove the interpolant "$" on the other tests where it's not necessary? That will improve legibility.

Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
```diff
@@ -1003,61 +1003,74 @@ class Cla__ss {
...
+      var documentItem = CreateTestDocument(source);
+      await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+      var counterExamples = (await RequestCounterExamples(documentItem.Uri)).ToArray();
+      Assert.AreEqual(1, counterExamples.Length);
+      Assert.IsTrue(counterExamples[0].Variables.ContainsKey("c:M.C?"));
+      Assert.IsTrue(counterExamples[0].Variables["c:M.C?"] is
+        "(c1 := '\\u1023', c2 := '\\u1023')" or
+        "(c2 := '\\u1023', c1 := '\\u1023')");
     }
+
```
MikaelMayer, line 1074:
```suggestion
```

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -137,7 +139,7 @@ public class DafnyModel {
         } else if (fn.Name.StartsWith("#") && fn.Name.IndexOf('.') != -1 && fn.Name[1] != '#') {
           foreach (var tpl in fn.Apps) {
             var elt = tpl.Result;
-            datatypeValues.Add(elt, tpl);
+            datatypeValues[elt] = tpl;
```
MikaelMayer, line 142: Before, you would be ensuring an exception if the element already existed. Can you please explain me why you are confident we don't need to check this?

Source/DafnyServer/CounterExampleGeneration/DafnyModelState.cs
```diff
@@ -10,40 +10,42 @@
...
+    internal DafnyModelState(DafnyModel model, Model.CapturedState state) {
       Model = model;
       State = state;
       VarIndex = 0;
       vars = new();
       varMap = new();
-      varNamesMap = new();
-      skolems = new List<DafnyModelVariable>(SkolemVars());
+      varNameCount = new();
+      boundVars = new(BoundVars());
```
MikaelMayer, line 43: Thanks for this much more meaningful variable name.

Source/DafnyServer/CounterExampleGeneration/DafnyModelState.cs
```diff
@@ -142,18 +145,16 @@ public class DafnyModelState {
         if (val == null) {
           continue; // This variable has no value in the model, so ignore it.
         }
+
         var vn = DafnyModelVariableFactory.Get(this, val, v, duplicate: true);
-        if (curVars.ContainsKey(v)) {
-          Model.RegisterLocalValue(vn.Name, val);
-        }
```
MikaelMayer, line 148: I don't know what this was doing before, but can you please explain me why you could delete it safely?

Source/DafnyServer/CounterExampleGeneration/DafnyModelVariable.cs
```diff
@@ -6,6 +6,8 @@
 using System.Linq;
 using System.Text.RegularExpressions;
 using Microsoft.Boogie;
+using Microsoft.Dafny;
```
MikaelMayer, line 9: Can you please fix these nullable annotations warnings above that the CI raised?

MikaelMayer (changes requested):
A few changes and questions, otherwise, I'm looking forward to enhance counter-examples!

Source/DafnyServer/CounterExampleGeneration/DafnyModelVariable.cs
```diff
@@ -192,51 +200,58 @@ internal SeqVariable(DafnyModelState state, Model.Element element, string name,
...
       this.seqLength = seqLength;
     }
 
-    public void AddAtIndex(DafnyModelVariable e, int? index) {
+    public void AddAtIndex(DafnyModelVariable e, string? index) {
       if (index == null) {
         return;
       }
-      seqElements[(int)index] = e;
+      seqElements[index] = e;
```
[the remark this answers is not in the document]
Dargones: Done

Source/DafnyServer/CounterExampleGeneration/DafnyModelVariable.cs
Dargones: Have added `#nullable enable` to fix this

Source/DafnyServer/CounterExampleGeneration/DafnyModelState.cs
Dargones: This remained from the original version that existed before I refactored the counterexample extraction code. I think its purpose must have been some form of field name preprocessing but at this point it is dead code. The only thing that [`DafnyModel.RegisterLocalValue`](https://github.com/dafny-lang/dafny/blob/dc6e7f086a2fe4860d8885ad7a0ab80ac13a2b8e/Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs#L824-L830) does is update the `localValue` dictionary, which is never used elsewhere, so I removed it. The two methods that `RegisterLocalValue` calls, which are `CompareFieldNames` and `GetNumber`, are static, non side-effecting and operate on strings, so I also remove these as part of this PR.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Type aliasing can lead to an element being associated with different datatype functions in the Model so we should not be throwing exceptions in this case. My goal is also to prevent exceptions from arising as much as possible - in some cases, for instance, the tool might not be able to figure out the type of a variable and return UnknownType but this will not break the execution and some counterexample will still be extracted, even if a portion of it will not have complete information. 

On related note, would VSCode extension crash if there were an exception in counterexample extraction? Provided this PR gets merged, I don't think there should be situations where it should throw and exception but adding a try-catch somewhere might still make sense to prevent issues in the future. Do you know what’s the best place for this would be.

Dargones:
Thank you, @MikaelMayer! Have adopted the changes.

Source/DafnyServer/CounterExampleGeneration/DafnyModelState.cs
MikaelMayer: Fair enough, thanks for the explanation.

Source/DafnyServer/CounterExampleGeneration/DafnyModel.cs
MikaelMayer: If the language server crashes because of an unhandled exception, many times the VSCode extension will not be responsive anymore.
So I understand why you want to avoid crashes and I'm with you on this refactoring.
I'm just wondering whether if you could store the information somewhere if for any reason the same key was assigned two different values. At least for example, in a log file, or at best in some log message about the counter-example sent via the counterexample API (so that it's easier to debug, and to handle later when the UI for counterexamples will change). I think the log file would be fine.

In general it's better to add the try-catch to the closest place where the exception might be thrown, if not, where it makes the most sense.

Dargones: Adding a logging functionality would make sense! I would keep that separate from this PR, if possible, but I already have some logging functionality for test generation on my fork, so I might be adding it in the near future.

Dargones:
Have now fixed all nullability warnings. Let me know if you would like me to make any other changes!

MikaelMayer (dismissed):
Great job ! Thanks for your patience.
Also, can you please confirm this enhancement is not going to break clients?

MikaelMayer (approved):

Dargones:
Thank you, @MikaelMayer! I don't add any new functionality, so it shouldn't break anything.

PR #2471, opened 2022-07-22
doc: Fixes #2462 -- adds information about ORDINALs

### LINKED ISSUE #2462: doc: Document ORDINAL.IsLimit
Add documentation about IsLimit, IsNat, IsSucc, Offset to the ORDINAL type section of the RM.

davidcok:
Fixes #2462 -- corrects and adds information about ORDINALs


cpitclaudel (approved):

docs/DafnyRef/Types.md
```diff
@@ -430,16 +430,30 @@ OrdinalType_ = "ORDINAL"
...
+The natural numbers are ordinals.
+Any ordinal has a successor ordinal (equivalent to adding `1`).
+Some ordinals are _limit_ ordinals, meaning they are not a successor to any other ordinal;
+the natural number `0` and  $\omega$ are limit ordinals.
+
+The _offset_ of an ordinal is the number of successor operations it takes to reach it from a limit ordinal.
+
+The Dafny type `ORDINAL` has these member functions:
+- `o.IsLimit` -- true if `o` is a limit ordinal (including `0`)
+- `o.IsSucc` -- true if `o` is a successor to something, so `o.IsSucc <==< !o.IsLimit`
```
MikaelMayer, line 445: Could you add a pointer to the doc of `<==<` so that I could know what this symbol mean? Or explain it on the fly.

docs/DafnyRef/Types.md
```diff
@@ -430,16 +430,30 @@ OrdinalType_ = "ORDINAL"
 ````
 
 Values of type `ORDINAL` behave like `nat`s in many ways, with one important difference:
-there is an `ORDINAL` value that is larger than any `nat`.
-- a value of type `nat` may be explicitly converted to an `ORDINAL` using `as ORDINAL`
-- a value of type `ORDINAL` may be explicitly converted to a `nat` using `as nat` if it is known to be less than some `nat` value 
-- non-negative numeric literals may be considered `ORDINAL` literals
+there are `ORDINAL` values that are larger than any `nat`. The smallest of these non-nat ordinals is
+represented as $\omega$ in mathematics, though there is no literal expression in Dafny that represents this value.
```
MikaelMayer, line 434: Except maybe; `var w :| !w.IsNat && (forall z:: z.IsNat || w <= z);` ?

MikaelMayer (commented):
Thanks for making the doc much much clearer!
A link to <==< would be good.

docs/DafnyRef/Types.md
davidcok: Sorry. <==>

docs/DafnyRef/Types.md
davidcok: I thought of that as well, but  actually meant a Literal, like 5 or true. Though I think it would be
var w :| !w.IsNat && (forall z:: !z.IsNat || w <= z)

cpitclaudel (approved):

MikaelMayer (approved):
Great job! I understand better ordinals now.

PR #2476, opened 2022-07-22
fix: Incorrect error message for modifies clause (closes #2384)

### LINKED ISSUE #2384: Confusing error message in override check for `modifies` clause
The following example causes Dafny to complain about reads:

```
trait T {
  method M()
}

class C {}

class C' extends T {
  var c: C
  constructor (c: C) { this.c := c; }
  method M() // expression might read an object not in the parent trait context's reads clause
    modifies this.c
  {
  }
}
```

There are two problems here:

1. The error mentions an expression, but it doesn't point to one (it's attached to `M`)
2. The error talks about a `reads clause`, but the problem is with a `modifies` clause.  There are no explicit read clauses in this program, so if the problem is that the `modifies` expression reads too much, it would be much better to have a custom message to report that.

h/t @seebees

cpitclaudel:
The bug was introduced by a typo in 53765c537.  This PR further improves the
message by not mentioning "expressions".

### DIFF
--- a/Source/Dafny/Verifier/ProofObligationDescription.cs
+++ b/Source/Dafny/Verifier/ProofObligationDescription.cs
@@ -378,20 +378,22 @@ public class YieldEnsures : ProofObligationDescription {
 public class TraitFrame : ProofObligationDescription {
   public override string SuccessDescription =>
     isModify
-      ? "expression abides by trait context's modifies clause"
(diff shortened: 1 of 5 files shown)


atomb (approved):
Good catch!

RustanLeino (approved):

PR #2478, opened 2022-07-22
Simplify concurrent updates in the IDE and separate concern of handling a single document revision from other concerns

keyboardDrummer:
### Changes

- Updates to a single version of a document are done synchronously. Operations that are running concurrently, such as verification, will schedule a synchronous update to DafnyDocument. Updates to DafnyDocument also synchronously trigger publication, followed by storing a copy of the document that then becomes available to IDE request handlers.
- Updates for different document versions can still fire concurrently, but `DocumentObserver` protects against this by ignoring updates from documents of a version older than the latest one it has seen.
- Separate the concern of handling a single document revision by moving all entry code related to that into a new type `CompilationManager`.
- Invariant: migration of previous document state only affects the state visible to the client, but does not affect the compilation of the latest document.
- Invariant: migration code will only migrate DafnyDocuments that have already been published to the client

### Testing
This is a refactoring, no tests have been added


Source/DafnyLanguageServer/Workspace/DocumentObserver.cs
```diff
@@ -0,0 +1,82 @@
...
+      var previousDiagnostics = LastPublishedDocument.LoadCanceled
+        ? new Diagnostic[] { }
+        : LastPublishedDocument.ParseAndResolutionDiagnostics;
+      var internalErrorDiagnostic = new Diagnostic() {
+        Message =
+          "Dafny encountered an internal error. Please report it at <https://github.com/dafny-lang/dafny/issues>.\n" +
+          exception,
+        Severity = DiagnosticSeverity.Error,
+        Range = LastPublishedDocument.Program.GetFirstTopLevelToken().GetLspRange(),
+        Source = "Crash"
```
MikaelMayer, line 57: Please recover the source from the exception so that it's a standard value...? one of MessageSource.

Source/DafnyLanguageServer/Workspace/CompilationManager.cs
```diff
@@ -0,0 +1,346 @@
...
+  public Task<DafnyDocument> ResolvedDocument { get; }
+  public Task<DafnyDocument> TranslatedDocument { get; }
+
+  public CompilationManager(IServiceProvider services,
+    VerifierOptions verifierOptions,
+    DocumentTextBuffer textBuffer,
+    Range? lastChange,
+    bool shouldVerify,
+    ImmutableList<Position> migratedLastTouchedVerifiables,
+    VerificationTree? migratedVerificationTree) {
```
cpitclaudel, line 57:
```suggestion
    VerificationTree? migratedVerificationTree)
  {
```

Source/DafnyLanguageServer/Workspace/CompilationManager.cs
```diff
@@ -0,0 +1,346 @@
...
+  private readonly IProgramVerifier verifier;
+
+  private DafnyOptions Options => DafnyOptions.O;
+  private VerifierOptions VerifierOptions { get; }
+
+  public DocumentTextBuffer TextBuffer { get; }
+  private readonly Range? lastChange;
+  private readonly IServiceProvider services;
+  private readonly ImmutableList<Position> migratedLastTouchedVerifiables;
+  private bool shouldVerify;
```
cpitclaudel, line 40: Is this the only field that's directly mutated in this class?

Source/DafnyLanguageServer/Workspace/CompilationManager.cs
```diff
@@ -0,0 +1,346 @@
...
+
+    if (shouldVerify) {
+      var _ = VerifyEverythingAsync();
+    } else {
+      MarkVerificationFinished();
+    }
+  }
+
+  public void Verify() {
+    if (shouldVerify) {
```
cpitclaudel, line 84: The name `shouldVerify` confused me.  Can we rename the constructor argument `shouldVerify` to something like `verifyImmediately` and the field `shouldVerify` to `verificationAlreadyStarted`, or something similar?

Source/DafnyLanguageServer/Workspace/CompilationManager.cs
```diff
@@ -0,0 +1,346 @@
...
+    this.shouldVerify = shouldVerify;
+    this.migratedVerificationTree = migratedVerificationTree;
+    this.lastChange = lastChange;
+    cancellationSource = new();
+
+    ResolvedDocument = ResolveAsync();
+    TranslatedDocument = TranslateAsync();
+
+    if (shouldVerify) {
+      var _ = VerifyEverythingAsync();
```
cpitclaudel, line 77: Do we not need to mark verification started here too?

Source/DafnyLanguageServer/Workspace/CompilationManager.cs
```diff
@@ -0,0 +1,346 @@
...
+      MarkVerificationFinished();
+    }
+  }
+
+  public void Verify() {
+    if (shouldVerify) {
+      return;
+    }
+
+    shouldVerify = true;
```
cpitclaudel, line 88: There can't be a race condition on this because this class is used from a single thread; correct?  It might be good to document it as not being thread-safe?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -31,20 +33,18 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
...
     bool LoadCanceled = false
   ) {
-    public IScheduler UpdateScheduler { get; } = new EventLoopScheduler();
-    public Subject<DafnyDocument> VerificationUpdates { get; } = new();
 
     public IReadOnlyList<IImplementationTask>? VerificationTasks { get; set; } = null;
 
     public IEnumerable<Diagnostic> Diagnostics => ParseAndResolutionDiagnostics.Concat(
-      ImplementationIdToView.SelectMany(kv => kv.Value.Diagnostics));
+      ImplementationIdToView?.SelectMany(kv => kv.Value.Diagnostics) ?? Enumerable.Empty<Diagnostic>());
```
cpitclaudel, line 47: Why do we allow `ImplementationIdToView` to be null?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -77,19 +77,19 @@ public record DafnyDocument(
 
     public int LinesCount => VerificationTree.Range.End.Line;
     public IVerificationProgressReporter? GutterProgressReporter { get; set; }
-    public List<Counterexample> Counterexamples { get; set; } = new();
-    public Dictionary<ImplementationId, ImplementationView> ImplementationIdToView { get; set; } = new();
+    public List<Counterexample>? Counterexamples { get; set; }
+    public Dictionary<ImplementationId, ImplementationView>? ImplementationIdToView { get; set; }
```
cpitclaudel, line 81: Why do we allow these things to be null?

Source/DafnyLanguageServer/Workspace/DafnyDocument.cs
```diff
@@ -77,19 +77,19 @@ public record DafnyDocument(
...
     public DafnyDocument Snapshot() {
-      var result = new DafnyDocument(TextDocumentItem, ParseAndResolutionDiagnostics, CanDoVerification, GhostDiagnostics,
-        Program, SymbolTable, WasResolved, LoadCanceled) {
+      var result = new DafnyDocument(TextDocumentItem, ParseAndResolutionDiagnostics, SymbolTable, CanDoVerification, GhostDiagnostics,
+        Program, WasResolved, LoadCanceled) {
         VerificationTree = VerificationTree,
-        Counterexamples = Counterexamples.ToList(),
-        ImplementationIdToView = new(ImplementationIdToView),
-        LastTouchedMethodPositions = LastTouchedMethodPositions,
+        Counterexamples = Counterexamples == null ? null : new(Counterexamples),
```
cpitclaudel, line 90:
```suggestion
        Counterexamples = Counterexamples?.ToList(),
```

Maybe?

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -0,0 +1,180 @@
...
+  private readonly DocumentOptions documentOptions;
+  private readonly VerifierOptions verifierOptions;
+  private readonly DocumentObserver observer;
+  public CompilationManager CompilationManager { get; private set; }
+  private IDisposable observerSubscription;
+
+  private bool VerifyOnOpen => documentOptions.Verify == AutoVerification.OnChange;
+  private bool VerifyOnChange => documentOptions.Verify == AutoVerification.OnChange;
+  private bool VerifyOnSave => documentOptions.Verify == AutoVerification.OnSave;
+  public Task<DafnyDocument?> LastDocumentAsync => CompilationManager.LastDocument!;
```
cpitclaudel, line 35: When can `LastDocument` be null?

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -0,0 +1,180 @@
...
+
+  public DocumentManager(
+    IServiceProvider services,
+    DocumentOptions documentOptions,
+    VerifierOptions verifierOptions,
+    DocumentTextBuffer document) {
```
cpitclaudel, line 41:
```suggestion
    DocumentTextBuffer document)
  {
```

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -0,0 +1,180 @@
...
+      };
+    }
+
+    var migratedViews = document.ImplementationIdToView?.Select(kv => {
+      var value = kv.Value.Status < PublishedVerificationStatus.Error
+        ? kv.Value with {
+          Diagnostics = migratedImplementationViews != null &&
+                        migratedImplementationViews.TryGetValue(kv.Key, out var previousView)
+            ? previousView.Diagnostics
+            : kv.Value.Diagnostics
```
cpitclaudel, line 126:
```suggestion
          Diagnostics = migratedImplementationViews?.GetValueOrDefault(kv.Key)?.Diagnostics
            ?? kv.Value.Diagnostics
```

Maybe?

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -0,0 +1,180 @@
...
+        }
+        : kv.Value;
+      return new KeyValuePair<ImplementationId, ImplementationView>(kv.Key, value);
+    }) ?? migratedImplementationViews;
+    document.ImplementationIdToView = migratedViews == null ? null : new(migratedViews);
```
cpitclaudel, line 131: It would be nice to simplify that code by using empty dictionaries instead of `null`, if possible.

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -0,0 +1,180 @@
...
+      }
+    }
+    return result;
+  }
+
+  public void Save() {
+    if (!VerifyOnSave) {
+      return;
+    }
+    CompilationManager.Verify();
```
cpitclaudel, line 157:
```suggestion
    if (VerifyOnSave) {
      CompilationManager.Verify();
    }
```

cpitclaudel (changes requested):
Thanks for this deep refactoring work.  I've left some comments, but I think it's basically good to go.  Still, here are a few high-level comments:

- We are using `TaskCanceledException` in multiple places.  Is that the right thing?  Shouldn't it be `OperationCanceledException`?
- I'm a bit uncomfortable about refactorings that also happen to fix bugs, though here we have a sense of what kind of bugs the refactoring prevents, even we haven't pinpointed the source of some of the concurrency issues that we're seeing.
- I'm having trouble keeping in mind what is mutable and what isn't.  Given the current design of the AST (resolution mutates), it would seem natural to me that further operations (translation, verification) would further mutate the original document (stepping each document through a number of states).
- For large PRs like this, I would find it very useful to get a summary of what the code changes do concretely (not their intent).  I find that summaries like in https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=289b457cac1439ac5f9bb6ce1143d91b8d52da91 or in https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=6dd4c5b953b666a9079910665d770ed56ecbf65d make it much easier to understand what's going on in a diff, especially in the presence of significant code moves.

Again, congrats on the large refactoring!

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -71,11 +69,11 @@ private class CounterExampleLoader {
       }
 
       public CounterExampleList GetCounterExamples() {
-        if (!document.Counterexamples.Any()) {
+        if (!document.Counterexamples!.Any()) {
           logger.LogDebug("got no counter-examples for document {DocumentUri}", document.Uri);
           return new CounterExampleList();
         }
-        var counterExamples = GetLanguageSpecificModels(document.Counterexamples)
+        var counterExamples = GetLanguageSpecificModels(document.Counterexamples!.ToArray())
```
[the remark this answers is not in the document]
keyboardDrummer: Removed it

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -16,28 +16,26 @@ namespace Microsoft.Dafny.LanguageServer.Handlers.Custom {
...
-          var verificationTasks = document!.VerificationTasks;
+        var documentManager = documents.GetDocumentManager(request.TextDocument);
+        if (documentManager != null) {
+          var translatedDocument = await documentManager.CompilationManager.TranslatedDocument;
+          var verificationTasks = translatedDocument.VerificationTasks;
           if (verificationTasks != null) {
-            var entry = documents.Documents[request.TextDocument.Uri];
             foreach (var task in verificationTasks) {
-              documentLoader.Verify(entry, document, task, CancellationToken.None);
+              documentManager.CompilationManager.Verify(translatedDocument, task);
```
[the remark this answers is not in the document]
keyboardDrummer: It manages the equivalent of a single Dafny CLI call, or a single 'compilation'. The only back-end used is the verification one though. Does that mean we shouldn't call it compilation? An alternative name would be `ParseResolveAndVerifyManager` or `StaticAnalysisManager`.

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -16,28 +16,26 @@ namespace Microsoft.Dafny.LanguageServer.Handlers.Custom {
...
     public async Task<CounterExampleList> Handle(CounterExampleParams request, CancellationToken cancellationToken) {
       try {
-        var document = await documents.GetLastDocumentAsync(request.TextDocument);
-
-        if (document != null) {
-          var verificationTasks = document!.VerificationTasks;
+        var documentManager = documents.GetDocumentManager(request.TextDocument);
+        if (documentManager != null) {
+          var translatedDocument = await documentManager.CompilationManager.TranslatedDocument;
+          var verificationTasks = translatedDocument.VerificationTasks;
```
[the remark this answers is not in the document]
keyboardDrummer: > Is the TranslatedDocument the result of the translation to Boogie?

It's a document that also contains the results of the translation to Boogie. `BoogieDocument` implies that it's only Boogie, but it's also Boogie.

Source/DafnyLanguageServer/Handlers/Custom/VerificationHandler.cs
```diff
@@ -22,31 +22,23 @@ public record VerificationParams : TextDocumentPositionParams, IRequest<bool>;
...
 
-    var translatedDocument = await documentEntry.TranslatedDocument;
+    var translatedDocument = await documentManager.CompilationManager.TranslatedDocument;
     var requestPosition = request.Position;
-    var tasks = GetTasksAtPosition(translatedDocument, requestPosition);
-    var anyAreRunning = tasks.Any(taskToRun =>
-      documentLoader.Verify(documentEntry, translatedDocument, taskToRun, CancellationToken.None));
-
-    return anyAreRunning;
+    var tasks = GetTasksAtPosition(translatedDocument, requestPosition).ToList();
```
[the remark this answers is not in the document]
keyboardDrummer: It's not, removed.

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
keyboardDrummer: It couldn't, changed.

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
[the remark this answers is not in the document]
cpitclaudel: Thanks, that makes sense.  Maybe `PipelineManager`?  Otherwise `CompilationManager` is fine.

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
keyboardDrummer: Nice

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
keyboardDrummer: Right now a document that has been resolved will use the migrated `ImplementationIdToView` from the previous document version without modifications, while a document that has been translated and has just created its own `ImplementationIdToView`, will combine this with the migrated one in a specific way. Using this combine function with an empty dictionary and a migrated one would give bad results.

In the future I would like to have types for the specific phases a DafnyDocument can be in, so one `TranslatedDocument` type, for which there's be a non-null field `ImplementationIdToView`, and a `ResolvedDocument` type which won't have the `ImplementationIdToView` field.

Source/DafnyLanguageServer/Workspace/DocumentObserver.cs
keyboardDrummer: It's an optional field. I've removed it.

keyboardDrummer:
> * We are using `TaskCanceledException` in multiple places.  Is that the right thing?  Shouldn't it be `OperationCanceledException`?

When catching exceptions, I think we should use `OperationCanceledException`. Made a relevant change in the last revision.

> * I'm a bit uncomfortable about refactorings that also happen to fix bugs, though here we have a sense of what kind of bugs the refactoring prevents, even we haven't pinpointed the source of some of the concurrency issues that we're seeing.

Yes, I think that's out of scope here. Ideally we have tests that reliably trigger the concurrency issues, and then we can use those to evaluate our code. This PR is really about readability.

There's one specific reason why this PR helps with test instability, and that's because it changes the implementation of the _test_ `ReorderingVerificationGutterStatusTester` so it no longer relies on the slowness of the verifier for things to happen in the correct order. Instability in `ReorderingVerificationGutterStatusTester` was one you reported.

> * I'm having trouble keeping in mind what is mutable and what isn't.  Given the current design of the AST (resolution mutates), it would seem natural to me that further operations (translation, verification) would further mutate the original document (stepping each document through a number of states).

Both a `Program` and `DafnyDocument` are mutable. I don't think there's much in this PR that isn't mutable. One confusing thing is that `DafnyDocument` is (and already was) a record with a primary constructor, which gives the appearance of immutability, especially because sometimes copies of a DafnyDocument are made using the `with` syntax. I think it might be better to turn `DafnyDocument` back into a class, but I'll leave that for a future PR.

> * For large PRs like this, I would find it very useful to get a summary of what the code changes do concretely (not their intent).  I find that summaries like in https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=289b457cac1439ac5f9bb6ce1143d91b8d52da91 or in https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=6dd4c5b953b666a9079910665d770ed56ecbf65d make it much easier to understand what's going on in a diff, especially in the presence of significant code moves.

Interesting! I'll keep it in mind.
 
> Again, congrats on the large refactoring!

Thanks :)

cpitclaudel (approved):
Thanks for the replies!

PR #2484, opened 2022-07-22
Allow `{:options "/noNLarith"}`

atomb:
This will make it possible to verify the [standard library](https://github.com/dafny-lang/libraries/) without a special command line.

### DIFF
--- a/Source/Dafny/DafnyOptions.cs
+++ b/Source/Dafny/DafnyOptions.cs
@@ -1299,6 +1299,7 @@ class ErrorReportingCommandLineParseState : Bpl.CommandLineParseState {
 class DafnyAttributeOptions : DafnyOptions {
   public static readonly HashSet<string> KnownOptions = new() {
     "functionSyntax",
+    "noNLarith",
     "quantifierSyntax"
   };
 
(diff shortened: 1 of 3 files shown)


cpitclaudel (changes requested):
I wish it were that easy, but this won't be sufficient.  This setting controls the value of `Options.O.DisableNLArith`, and that value will be restored to its default as soon as the parser exists your annotated module.  Later on when code in the translator and the trigger collector looks up the value of the option, it will see the global default (the value passed on the command line), which is not what you want.

The fix is to stash the value of that option somewhere during parsing (probably on the module itself?) and make sure that downstream code consults that instead of `Options.O.DisableNLArith`.  

Maybe it's possible to just stash the complete options object and make that current during translation / resolution / etc. of the module (anywhere that might consult the option).

atomb:
> Maybe it's possible to just stash the complete options object and make that current during translation / resolution / etc. of the module (anywhere that might consult the option).

Ah, that's too bad. I figured what you describe above was what was already happening, rather than just keeping it for parsing. I'll change this to a draft for now until I (or someone else) has a moment to implement all of that.

atomb:
Given that the work in this PR is both small and doesn't solve the problem, and that we'll probably need to start from scratch when we implement this functionality eventually, I'm going to close it.

PR #2485, opened 2022-07-24
feat: `codatatype`s for Python

fabiomadge:
Fixes #2392

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/dafny-lang/dafny/blob/master/LICENSE.txt).</small>

PR #2486, opened 2022-07-24
chore: Enable additional tests for Python

fabiomadge:
Requires #2270, #2428, and #2485


Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -320,9 +349,15 @@ public class PythonCompiler : SinglePassCompiler {
...
-      w.WriteLine($"default = staticmethod(lambda: {d})");
+      if (nt.WitnessKind == SubsetTypeDecl.WKind.Compiled) {
+        w.WriteLine("@staticmethod");
+        var block = w.NewBlockPy("def default():");
+        var wStmts = block.Fork();
+        block.Write("return ");
+        TrExpr(nt.Witness, block, false, wStmts);
+      } else {
+        w.WriteLine($"default = staticmethod(lambda: {TypeInitializationValue(udt, wr, nt.tok, false, false)})");
+      }
```
robin-aws, line 360: Can this be in a shared `DefineDefaultMethod` method?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -613,13 +668,22 @@ protected class ClassWriter : IClassWriter {
...
+                      if (decl.Class is ArrayClassDecl arr) {
+                        return $"{DafnyArrayClass}.empty({arr.Dims})";
+                      }
+                      return "None";
+                    }
+                    Contract.Assert(udt.TypeArgs.Any() && ArrowType.IsTotalArrowTypeName(td.Name));
                     var rangeDefaultValue = TypeInitializationValue(udt.TypeArgs.Last(), wr, tok, usePlaceboValue,
                       constructTypeParameterDefaultsFromTypeDescriptors);
-                    var arguments = udt.TypeArgs.Comma((_, i) => $"x{i}");
+                    var arguments = udt.TypeArgs.SkipLast(1).Comma((_, i) => $"x{i}");
```
robin-aws, line 686: Took me a bit to figure out why we needed the `SkipLast`, I assume it's because at this point we know udt is a total arrow and hence we're skipping the result type? Probably worth a quick comment.

robin-aws (dismissed):
Nice work! Just naming and clean-up suggestions.

I think we can remove `TupleInitialization` from the set of unsupported features for Python now, right? AFAICT that was already completed in a previous change, but I didn't spot it (and we don't have a mechanism that makes us remove those yet :)

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -14,22 +14,31 @@ def __get__(self, instance, owner):
         return classmethod(self.fget).__get__(None, owner)()
 
 def print(value):
-    if value is None:
-        builtins.print("null", end="")
+    builtins.print(str(value), end="")
+
+def str(value) -> builtins.str:
```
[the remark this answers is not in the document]
fabiomadge: This becomes `_dafny.str` in the generated code. Your concern only affects the runtime.

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: What would you like to be shared? The first one needs the block so we have something for `wStmts` to point to. The other way is more succinct.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -223,15 +225,40 @@ public class PythonCompiler : SinglePassCompiler {
...
+                                   from dtor in ctor.Destructors
+                                   where dtor.EnclosingCtors[0] == ctor
+                                   select dtor.CorrespondingFormals[0] into arg
+                                   where !arg.IsGhost
+                                   select IdProtect(arg.CompileName)) {
+          w.WriteLine("@property");
+          w.NewBlockPy($"def {destructor}(self):")
+            .WriteLine($"return self._get().{destructor}");
+        }
+      }
```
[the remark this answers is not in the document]
fabiomadge: Bad merge. Thanks!

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -176,15 +176,16 @@ public class PythonCompiler : SinglePassCompiler {
...
       var methodWriter = wr.NewBlockPy(header: $"class {IdProtect(name)}{baseClasses}:");
 
-      var needsConstructor = cls is TopLevelDeclWithMembers decl
-                             && decl.Members.Any(m => !m.IsGhost && ((m is Field && !m.IsStatic) || m is Constructor));
-      ConcreteSyntaxTree constructorWriter = null;
-      if (needsConstructor) {
-        var block = methodWriter.NewBlockPy(header: "def  __init__(self):", close: BlockStyle.Newline);
-        constructorWriter = block.Fork();
-        block.WriteLine("pass");
+      var args = typeParameters.Where(NeedsTypeDescriptor).Comma(tp => tp.CompileName);
```
cpitclaudel, line 179: Let's add a variable for `typeParameters.Where(NeedsTypeDescriptor)` to make it obvious that they can't get out of sync

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -243,8 +244,8 @@ public class PythonCompiler : SinglePassCompiler {
           .WriteLine("self.d = self.c()")
           .WriteLine("self.c = None");
         get.WriteLine("return self.d");
-        w.NewBlockPy("def __str__(self) -> str:")
-          .WriteLine("return str(self._get())");
+        w.NewBlockPy("def __dafnystr__(self) -> str:")
+          .WriteLine($"return {DafnyRuntimeModule}.str(self._get())");
```
cpitclaudel, line 248: You have `_dafny` in some places and `DafnyRuntimeModule` in others

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -262,7 +263,9 @@ public class PythonCompiler : SinglePassCompiler {
         var ctorName = IdProtect(ctor.CompileName);
 
         // Class-level fields don't work in all python version due to metaclasses.
-        var argList = ctor.Destructors.Select(d => $"(\'{IdProtect(d.CompileName)}\', {TypeName(d.Type, wr, d.tok)})").Comma();
+        // Adding a more restrictive type would be desirable, but Python expects their definition to precede this.
+        var argList = ctor.Destructors.Where(d => !d.IsGhost)
+          .Select(d => $"(\'{IdProtect(d.CompileName)}\', Any)").Comma();
```
cpitclaudel, line 268: Do the quotes really need escaping?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -262,7 +263,9 @@ public class PythonCompiler : SinglePassCompiler {
         var ctorName = IdProtect(ctor.CompileName);
 
         // Class-level fields don't work in all python version due to metaclasses.
-        var argList = ctor.Destructors.Select(d => $"(\'{IdProtect(d.CompileName)}\', {TypeName(d.Type, wr, d.tok)})").Comma();
+        // Adding a more restrictive type would be desirable, but Python expects their definition to precede this.
+        var argList = ctor.Destructors.Where(d => !d.IsGhost)
```
cpitclaudel, line 267: Could we maybe give a name to `Destructors.Where(d => !d.IsGhost)`?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -320,9 +323,15 @@ public class PythonCompiler : SinglePassCompiler {
...
-
-      w.WriteLine($"default = staticmethod(lambda: {d})");
+      if (nt.WitnessKind == SubsetTypeDecl.WKind.Compiled) {
+        w.WriteLine("@staticmethod");
+        var block = w.NewBlockPy("def default():");
+        var wStmts = block.Fork();
+        block.Write("return ");
+        TrExpr(nt.Witness, block, false, wStmts);
+      } else {
+        w.WriteLine($"default = staticmethod(lambda: {TypeInitializationValue(udt, wr, nt.tok, false, false)})");
```
cpitclaudel, line 333: `staticmethod(lambda: … )` isn't the more readable — why not use the same form for both?
(Or use a property in the first case and a constant on the second case)

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -331,10 +340,15 @@ public class PythonCompiler : SinglePassCompiler {
...
-      w.WriteLine("@staticmethod");
-      w.NewBlockPy("def default():").WriteLine($"return {d}");
+      if (sst.WitnessKind == SubsetTypeDecl.WKind.Compiled) {
+        w.WriteLine("@staticmethod");
+        var block = w.NewBlockPy("def default():");
+        var wStmts = block.Fork();
+        block.Write("return ");
+        TrExpr(sst.Witness, block, false, wStmts);
+      } else {
+        w.WriteLine($"default = staticmethod(lambda: {TypeInitializationValue(udt, wr, sst.tok, false, false)})");
```
cpitclaudel, line 350: This is duplicated from above, right?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -503,34 +516,50 @@ protected class ClassWriter : IClassWriter {
       Contract.Requires(wr != null);
 
       return type.NormalizeExpandKeepConstraints() switch {
+        var x when x.IsBuiltinArrowType => $"{DafnyDefaults}.pointer",
         // unresolved proxy; just treat as bool, since no particular type information is apparently needed for this type
         BoolType or TypeProxy => $"{DafnyDefaults}.bool",
```
cpitclaudel, line 521: Would we be able to catch errors earlier if we returned a dummy object in the `TypeProxy` case?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -1060,10 +1116,26 @@ protected class ClassWriter : IClassWriter {
...
+              }
+              var sep = "(";
+              EmitTypeDescriptorsActuals(ForTypeDescriptors(typeArgs, member, false), member.tok, w, ref sep);
+              if (customReceiver) {
+                w.Write(sep + "self");
+                sep = ", ";
+              }
+              if (sep != "(") {
+                w.Write(")");
+              }
```
cpitclaudel, line 1137: I would extract this parenthesizing logic into a separate function

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -14,22 +14,31 @@ def __get__(self, instance, owner):
...
+    if hasattr(value, '__dafnystr__'):
+        return value.__dafnystr__()
+    elif value is None:
```
cpitclaudel, line 22: An alternative to recycling None would be to use a singleton object `NULL`

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -14,22 +14,31 @@ def __get__(self, instance, owner):
...
+        return "null"
     elif isinstance(value, bool):
-        builtins.print("true" if value else "false", end="")
+        return "true" if value else "false"
+    elif isinstance(value, tuple):
+        return '(' + ', '.join(map(str, value)) + ')'
+    elif isinstance(value, FunctionType):
```
cpitclaudel, line 28: Alternatively, we could wrap Dafny functions into an object

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -86,6 +95,23 @@ def set(self, key, value):
...
+        return Array([Array.empty(dims-1)])
+
+    def __dafnystr__(self) -> builtins.str:
+        return '[' + ', '.join(map(str, self)) + ']'
+
+    def __len__(self):
+        l = super().__len__()
+        # Hack to enable "empty" matrices
+        if l == 1 and isinstance(self[0], Array) and len(self[0]) == 0:
+            return 0
```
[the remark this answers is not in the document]
cpitclaudel: Alternatively, you could define a singleton `UNIT` with overridden `length` / dimensions and use that as the "dimension-0" object

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -165,8 +191,8 @@ def __contains__(self, item):
         return self[item] > 0
 
 class Map(dict):
-    def __str__(self) -> str:
-        return 'map[' + ', '.join(map(lambda i: f'{i[0]} := {i[1]}', self.items)) + ']'
+    def __dafnystr__(self) -> builtins.str:
+        return 'map[' + ', '.join(map(lambda i: f'{str(i[0])} := {str(i[1])}', self.items)) + ']'
```
cpitclaudel, line 195: I find `str` risky as a name choice.  Are we sure we will remember to use `builtins.str` everywhere in the prelude?  I think it would be safer to use `dstr()` or `to_str`, and leave `str` to have its usual meaning?

cpitclaudel (approved):
This looks generally good to me, but I don't feel very confident in my assessment.  It would be much easier to look at the generated code than at the generator code ^^

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: This more explicit workaround is unique to the js compiler, for now. The others must catch it more implicitly. changing `NormalizeExpandKeepConstraints` may not be the worst idea

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: In `DatatypeCtor`?

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: I don't see a reasonably generic way. Those `sep`s are annoying to work with.

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -319,36 +320,28 @@ public class PythonCompiler : SinglePassCompiler {
...
+      var udt = UserDefinedType.FromTopLevelDecl(d.tok, d);
+      w.WriteLine("@staticmethod");
+      var block = w.NewBlockPy("def default():");
+      var wStmts = block.Fork();
+      block.Write("return ");
+      if (witnessKind == SubsetTypeDecl.WKind.Compiled) {
+        TrExpr(witness, block, false, wStmts);
       } else {
-        w.WriteLine($"default = staticmethod(lambda: {TypeInitializationValue(udt, wr, nt.tok, false, false)})");
+        block.Write(TypeInitializationValue(udt, wr, d.tok, false, false));
```
cpitclaudel, line 334: This looks good, but this is another thing that we should do in Dafny in the long run; these two cases should be the same (we should auto-fill the witness as the TypeInitializationValue)

PR #2492, opened 2022-07-26
Try running with z3 4.10.1

cpitclaudel:


atomb:
It looks like there are some long-running XUnit tests. I tried the integration tests locally with 4.10.1 and only 49/1269 failed with a solver time limit of 60s. So things are looking a little better than I expected.

atomb:
I tried the XUnit tests locally, too, with the small modification of removing the `model_compress=true` option to Z3, since i doesn't understand it anymore. With this change, 288/291 of the language server tests pass, along with all of the `AutoExtern` and `DafnyPipeline` tests. One of the `TestGeneration` tests hangs, however. We're closer to having new Z3 working than I thought!

cpitclaudel:
Closing this, since we've gathered the data we wanted

PR #2494, opened 2022-07-26
Optimize shift/rotate for more efficient solving

### LINKED ISSUE #2174: Simple lemma involving bit-vector operations cannot be proved
Dafny reports "A postcondition might not hold on this return path" after taking a while when verifying
```
lemma lm(val: bv16, idx: nat)
	requires 0 <= idx <= 16 - 8
	requires 0 <= val < (1 << 8)
	ensures val as bv32 << idx == (val << idx) as bv32 {}
```
The proof can go through quickly if a smaller bit-vector is used or `idx` is casted to bit-vector type.

### LINKED ISSUE #2230: Having trouble applying lemma involving bit-vectors
The following snippet cannot be verified.
```dafny
lemma MergeLShift(v: bv128, i: nat, j: nat)
  requires i <= 128 && j <= 128 && i + j <= 128
  ensures v << i << j == v << i + j

method M(i: nat)
  requires i <= 64
{
  ghost var half: bv128 := 0xffff_ffff_ffff_ffff;
  MergeLShift(half, 64, 64 - i);
  assert half << 64 - i << 64 == half << (128 - i);
}
```
Changing `nat`s to `bv32` would make it work, reminding me of #2174. 

It seems Dafny is having problem with `64-i`. For example, this lemma is inconclusive as well. Again, changing `nat`s to `bv32` works.
```dafny
lemma whyyy(v: bv128, a: nat, b: nat, i: nat)
  requires a <= 64 && b <= 64 && a == 64 - i
  ensures ((v << 64 - i) << b) == ((v << a) << b)
{}
```
On a separate note, no matter with `nat` or `bv32`, lemma `MergeLShift` itself is inconclusive.

atomb:
When translating `x << y` to Boogie, where `x` has width `w`, rather than immediately casting `y` to the type of `x`, instead cast `y` to a bit vector with width `log2(w)` and then cast, a second time, to a bit vector of width `w`. This ensures that the solver knows that most of the bits of the shift amount are zero.

Fixes #2174
Fixes #2230

### DIFF
--- a/Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
+++ b/Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
@@ -991,11 +991,11 @@ namespace Microsoft.Dafny {
 
             case BinaryExpr.ResolvedOpcode.LeftShift: {
                 Contract.Assert(0 <= bvWidth);
-                return TrToFunctionCall(GetToken(expr), "LeftShift_bv" + bvWidth, translator.BplBvType(bvWidth), e0, translator.ConvertExpression(GetToken(expr), e1, e.E1.Type, e.Type), liftLit);
+                return TrShiftRotateExpr("LeftShift", e, e0, e1, e.E1.Type, bvWidth, liftLit);
               }
             case BinaryExpr.ResolvedOpcode.RightShift: {
                 Contract.Assert(0 <= bvWidth);
-                return TrToFunctionCall(GetToken(expr), "RightShift_bv" + bvWidth, translator.BplBvType(bvWidth), e0, translator.ConvertExpression(GetToken(expr), e1, e.E1.Type, e.Type), liftLit);
+                return TrShiftRotateExpr("RightShift", e, e0, e1, e.E1.Type, bvWidth, liftLit);
               }
             case BinaryExpr.ResolvedOpcode.BitwiseAnd: {
                 Contract.Assert(0 <= bvWidth);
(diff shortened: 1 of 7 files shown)


Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
```diff
@@ -1422,17 +1422,27 @@ public ExpressionTranslator(ExpressionTranslator etran, string modifiesFrame)
...
+      // an easier SMT problem. This is always valid because the value of the
+      // RHS must be less than w.
+      private Expr TrShiftRotateExpr(string fnName, Expression expr, Expr lhs, Expr rhs, Type rhsType, int w, bool liftLit) {
+        var intermediateBvWidth = (int)(new BigInteger(w - 1)).GetBitLength();
+        translator.AddBitwidth(intermediateBvWidth);
+        var tok = GetToken(expr);
+        var rhsConvertedType = new BitvectorType(intermediateBvWidth);
+        var rhsConverted = translator.ConvertExpression(tok, rhs, rhsType, rhsConvertedType);
+        var rhsFinal = translator.ConvertExpression(tok, rhsConverted, rhsConvertedType, expr.Type);
+        return TrToFunctionCall(tok, fnName + "_bv" + w, translator.BplBvType(w), lhs, rhsFinal, liftLit);
```
keyboardDrummer, line 1436: Why does the RHS need to be the same type as the LHS in Boogie?

Also, what happens if the RHS value doesn't fit into the intermediate type? Is there a check someone else checks constrains this value? Can we refer to that check here to indicate why this is safe.

Source/Dafny/Verifier/Translator.cs
```diff
@@ -700,6 +701,40 @@ internal class PredefinedDecls {
       return new Bpl.IdentifierExpr(tok, var.AssignUniqueName(currentDeclaration.IdGenerator), TrType(var.Type));
     }
 
+    private void AddBitwidth(int w) {
+      if (generatedBitwidths.Contains(w)) {
```
keyboardDrummer, line 705: Consider writing this as `if (!generatedBitwidths.Add(w)) { return; }` and dropping the `generatedBitwidths.Add(w);` at the end of this method.

Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
atomb: The RHS needs to be the same type as the LHS because Boogie's bit vector operations are exactly those in SMT-Lib, and the [SMT-Lib standard](https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml) constrains them that way. I'm not sure what the rationale is behind it.

I'm glad you brought up whether the value fits in the intermediate type, though! I had thought that this was being checked elsewhere, and it sort of is, but there's actually an off-by-one between that check and what I need here. 

Dafny already emits a check that you don't shift (or rotate) more than the bit vector width. But you **can** shift by exactly the bit vector width (turning it into all zeros), and that one value won't fit into the bit vector of the size used here. So it needs to be one more bit.

cpitclaudel:
Very nice, and good catch @keyboardDrummer .

I'm surprised that we need to encode this explicitly.  I would expect that the check that we do elsewhere would give that information to the SMT solver.
Is is possible that the check is done in a way that the SMT solver doesn't understand?  If so, could we change that check to be implemented as a Dafny-level cast, but with a custom error message?

This is particularly important for rotate, because we've been having discussions about *removing* that check in the case of `rotate` (since rotation is well defined for all widths). If it's silently relied on in here and we remove it, we'll get unsoundness.  If we actually add a checked cast at the Dafny level, then we would catch such issues.

atomb:
I think the issue is that Dafny checks the range of the input to the cast at the integer level, but efficiency of these proofs depends on knowing information about the bits in a bit vector. If Dafny were to somehow translate properties (such as ranges) it knows about integers to the equivalent properties on bit vectors, whenever it did a cast, then I think the solver could use them. But it seems hard to do that in general.

You make a good point about having a custom message. Normally, when casting from an integer to a bit vector, Dafny inserts a check that the value will fit, but I believe it does that before the point where this code runs. I could (and think I will) add another check of that form that would be, for the moment, somewhat redundant (but probably easy to prove because of that) but that would guard against unsoundness in the case where we removed the range check for rotates.

cpitclaudel:
I'm also thinking that we could do this as a rewriter pass.  Specifically: could we rewrite `a >> b` into `a >> (b as bvN)`, and would that achieve the same as this PR?

atomb:
Yes, I think you're right about that. Do you think that'd be a preferable approach? I haven't dug into the rewriter much.

atomb:
In comparing this with rewriting, I think it makes sense to compare `TrShiftRotateExpr` with the equivalent rewrite rule. The rest of the content of this PR would likely be the same, I think. (Some of the changes consolidate some duplicated code that should probably be consolidated regardless.)

cpitclaudel:
> Do you think that'd be a preferable approach? I haven't dug into the rewriter much.

I'm scared of almost all changes to the verifier.  The cast would have caught the off-by-one error, for example (because the verifier would have complained about a a bad cast)

atomb:
I've successfully written a rewriter pass the has the same effect, with the benefit of making the modification before type checking or inserting side condition assertions. The downside is that it required making the children of `BinaryExpr` mutable (and suggests we do that for the rest of the AST, which is a big design decision). Is that tradeoff worth it?

atomb:
The rewrite-based approach is in draft PR #2520, for comparison.

atomb:
I'm closing this in favor of #2520.

PR #2501, opened 2022-07-27
Run `dotnet pack` before `dotnet publish`

### LINKED ISSUE #2423: Dafny NuGet package has overly-strict .NET framework dependencies
The `Dafny` package on NuGet depends strictly on version 6.0.7 of `Microsoft.NETCore.App`, even though I can't think of any reason it shouldn't be able to work with other point releases of 6.0. Since this is one of the core .NET frameworks, it can be awkward to have to install yet another version just to get Dafny to run.

atomb:
My hypothesis is that `dotnet publish` is adding a too-strict runtime version dependency that is then inherited by the created `.nupkg` files.

Attempts to fix #2423, but we won't know for sure until the next nightly build runs.

### DIFF
--- a/.github/workflows/publish-release-reusable.yml
+++ b/.github/workflows/publish-release-reusable.yml
@@ -116,8 +116,6 @@ jobs:
     # and that perturbs the path enough that `python` above would resolve to an older built-in Python version.
     # Additionally, since the refman build scripts expect to find Dafny in its usual Binaries/ folder (not in
     # a platform-specific directory), we build Dafny once here.
-    - name: Build Dafny
-      run: dotnet build dafny/Source/Dafny.sln
     - name: Build reference manual
       run: |
         eval "$(/usr/libexec/path_helper)"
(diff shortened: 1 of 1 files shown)


robin-aws (changes requested):
This seems like a good change! I can't comment on the line in the file changes, but you should also remove the "Build Dafny" step as it shouldn't be necessary any more (and was the original reason the nightly nuget publishing wasn't working AFAICT - your reordering the steps means it should just do nothing now, but still better to clean it up).

atomb:
> This seems like a good change! I can't comment on the line in the file changes, but you should also remove the "Build Dafny" step as it shouldn't be necessary any more (and was the original reason the nightly nuget publishing wasn't working AFAICT - your reordering the steps means it should just do nothing now, but still better to clean it up).

Good catch. Related to that, I was wondering whether the C++ and `npm` steps were necessary anymore, since we're not testing the compiler as part of this workflow (unless `package.py` does a little of that?).

robin-aws (approved):
Thanks, LGTM!

I wouldn't worry about taking out those steps that might not be necessary. They are likely just copied from other workflows, but I don't think it's worth the time to try to optimize that. I'd probably prefer to find a way to factor out "install dependencies" as a reusable workflow instead (although I'm not actually sure of the best way to do that offhand :)

PR #2509, opened 2022-07-28
Another attempt at building NuGet package first

### LINKED ISSUE #2423: Dafny NuGet package has overly-strict .NET framework dependencies

atomb:
The `dotnet publish` process that runs as part of `package.py` adds a too-strict runtime version dependency that is then inherited by the created `.nupkg` files. By building the `.nupkg` files first, we avoid this problem.

I've run the nightly build process manually on this branch, and confirmed that it succeeded and produced `.nupkg` files with the appropriate version constraints.

Fixes #2423

### DIFF
--- a/.github/workflows/publish-release-reusable.yml
+++ b/.github/workflows/publish-release-reusable.yml
@@ -84,7 +84,7 @@ jobs:
     - name: Make NuGet package available as an artifact
       uses: actions/upload-artifact@v2
       with:
-        name: nuget-packages-${{ matrix.os }}
+        name: nuget-packages
         path: dafny/Binaries/*.nupkg
 
     - name: Upload package to NuGet
--- a/.github/workflows/publish-release-reusable.yml
+++ b/.github/workflows/publish-release-reusable.yml
@@ -71,17 +71,13 @@ jobs:
     - uses: actions/setup-node@v1
     - run: npm install bignumber.js
 
-    - name: Build Dafny
-      run: |
-        dotnet build dafny/Source/Dafny.sln
-        rm dafny/Binaries/*.nupkg
     - name: Create release NuGet package (for uploading)
       if: ${{ !inputs.prerelease }}
-      run: dotnet pack --no-build dafny/Source/Dafny.sln
+      run: dotnet pack dafny/Source/Dafny.sln
     - name: Create prerelease NuGet package (for uploading)
       if: ${{ inputs.prerelease }}
       # NuGet will consider any package with a version-suffix as a prerelease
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-release-reusable.yml
```diff
@@ -70,6 +70,32 @@ jobs:
         java-version: 1.8
     - uses: actions/setup-node@v1
     - run: npm install bignumber.js
+
+    - name: Build Dafny
+      run: |
+        dotnet build dafny/Source/Dafny.sln
+        rm dafny/Binaries/*.nupkg
```
fabiomadge, line 77: Is that really necessary?

.github/workflows/publish-release-reusable.yml
```diff
@@ -94,7 +122,7 @@ jobs:
     # and that perturbs the path enough that `python` above would resolve to an older built-in Python version.
     # Additionally, since the refman build scripts expect to find Dafny in its usual Binaries/ folder (not in
     # a platform-specific directory), we build Dafny once here.
-    - name: Build Dafny
+    - name: Re-build Dafny
       run: dotnet build dafny/Source/Dafny.sln
```
fabiomadge, line 126: I think this step is obsolete.

atomb: I thought so, too, but the reference manual build fails without it. It may be possible to move that build up _before_ running `package.py`, though.

.github/workflows/publish-release-reusable.yml
atomb: I think it'd actually be okay to remove this whole step and then remove `--no-build` from the later `dotnet pack` commands. However, if we have `dotnet build` then we do, unfortunately, have to remove the `.nupkg` files it creates.

.github/workflows/publish-release-reusable.yml
atomb: Actually, given the comment in the workflow about the path to `python`, I don't really want to move that earlier.

fabiomadge (approved):

.github/workflows/publish-release-reusable.yml
fabiomadge: Right, I guess that is the new feature matrix's fault. It's regenerated when building the manual.

fabiomadge (approved):

PR #2510, opened 2022-07-28
feat: Allow assume statements annotated with {:axiom} to be compiled

cpitclaudel:
* `Source/Dafny/AST/DafnyAst.cs`
  (`AttributedToken`): New class to hold a token with attributes.
  (`AssignSuchThatStmt`, `AssignOrReturnStmt`): Use it.
* `Source/Dafny/Cloner.cs` (`CloneStmt`): Clone attributed tokens.
* `Source/Dafny/Resolver.cs` (`ResolveAssignOrReturnStmt`):
  Copy attributes from `AttributedToken` to newly created statements.
* `Source/Dafny/Dafny.atg`
  (`UpdateStmt`, `VarDeclStatement`): Allow attributes on `assume` keyword.
* `Source/Dafny/Compilers/SinglePassCompiler.cs`
  (`RejectAssume`): New function to check for `:axiom`-less `assume`.
  (`CheckHasNoAssumes_Visitor.VisitOneStmt`): Use it.
  (`TrStmt`): Remove unnecessary `assume` check (redundant with visitor).
* `Source/Dafny/Makefile`: Fix case of `parser` and `scanner` frames.


robin-aws (changes requested):
Thanks for tackling this!

I was a bit surprised by the `AttributedToken` approach, although I understand that allows you to share more code between `:- assume` and `:| assume` uses. It took me a bit to realize that it's not used consistently, though, because `assume` statements already have statement-level attributes and that's where the parser puts the `{:axiom}` instead. I also understand you could argue that in an `assume` statement the `assume` keyword determines what kind of AST node you have in the first place, whereas for `:- assume` and `:| assume` you could argue the keyword is a secondary feature.

Consider either attaching the `{:axiom}` on `assume` statements to the token as well for consistency, or drop a comment or two (perhaps on `AttributedToken` itself) to clarify the inconsistency.

We also need at least one test case that successfully compiles because it only uses `assume {:axiom}`, just in case the presence of those makes any compilers trip over themselves. :)

cpitclaudel:
> because assume statements already have statement-level attributes and that's where the parser puts the {:axiom} instead

Not quite: it's because the `:axiom` is not attached to the parent statement but to a subpart of it.   Even if `:-` statements has attributes, they would go before the `:-`, as for `:|` (`var x: int {:xyz} :| true;` is supported today).

I will add a comment.

cpitclaudel:
Rebased (to adjust to the split of AST.cs) and added a comment and a test

robin-aws (dismissed):

robin-aws (approved):

PR #2519, opened 2022-07-29
fix: Work around #2429 (incorrect Boogie type in sequence update)

### LINKED ISSUE #2429: Internal error when extending trait with defined type param
Getting an internal error for the following program:
```
predicate P<T>(s: seq<T>)
trait Spec<T> {
  var base: seq<T>
  method M(t: T) ensures P(base[0 := t])
}

class String {}
class Impl extends Spec<String> {
  method M(t: String)
}
```
Error:
```
/tmp/tmpjuhZi7.tmp.dfy(4,37): Error: invalid type for argument 2 in application of Seq#Update: ref (expected: T)
1 type checking errors detected in /tmp/tmpjuhZi7.tmp__module.bpl

/tmp/tmpjuhZi7.tmp__module.bpl(3057,107): Error: invalid type for argument 2 in application of Seq#Update: ref (expected: T)
1 type checking errors detected in /tmp/tmpjuhZi7.tmp__module.bpl
```

cpitclaudel:
When Dafny translates a class, the override check that it performs mentions the
old ensures clauses, applied to its new arguments.  For example, the following
code:

```dafny
trait T           { method M(a: int) returns (b: int) ensures b > a     }
class C extends T { method M(a: int) returns (b: int) ensures b > a + 1 }
```

… leads to the following check:

```boogie
implementation OverrideCheck(this: ref, a#0: int) returns (b#0: int) {
  assume b#0 > a#0 + 1;
  assert b#0 > a#0;
}
```

Problems pop up when argument types differ between the old and new ensures
clauses, which happens in the presence of type parameters:

```dafny
predicate P<A>(s: seq<A>)
trait T<A> {
  method M(a: A) ensures P([a][0 := a])
}

predicate Q<A>(s: seq<A>)
class C extends T<object> {
  method M(a: object) ensures Q([a][0 := a])
}
```

When translating the original ensures clause (`ensures P([a][0 := a])`) to Boogie
in the override check, Dafny uses the old types in some places, and hence
assumes that `[a]` has type `seq<A>`, not type `seq<object>`.  This causes it to
make a mistake about whether to insert a box, and generate the following Boogie
code:

```boogie
assume _module.__default.Q(Tclass._System.object(),
  Seq#Update(Seq#Build(Seq#Empty(): Seq Box, $Box(a#0)), LitInt(0), $Box(a#0)));
assert _module.__default.P(_module.T$A,
  Seq#Update(Seq#Build(Seq#Empty(): Seq Box, $Box(a#0)), LitInt(0), a#0));
```

Notice how the `assert` line differs from the `assume` one: the former (with the
`Q`) correctly boxes `a#0` twice: once in the sequence construction and once in
the update.  In contrast, the latter (with `P`, corresponding to the trait's
ensures), incorrectly omits the box on the second `a#0` (the argument to
`Seq#Update`).

The reason for the discrepancy is that the translation for `Seq#Build` uses the
type of the element to decide whether to insert a box (which evaluates to
`object` in both statements), whereas the translation for `Seq#Update` uses the
`Arg` field in the type of the sequence itself to decide whether to box `a#0`.
That type evaluates to `object` in the `assume` but to `Q` in the `assert`, and
hence no box is inserted.

This commit cheats by using the type of the parameter, instead of the `Arg`
field of the type of the sequence.  There are surely more places with this
issue, and the "right" fix is most likely to substitute the type parameter `A`
in the type of the ensures before generating the override check, but this is
left for later.

Closes #2429.

### DIFF
--- a/Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
+++ b/Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
@@ -551,20 +551,18 @@ namespace Microsoft.Dafny {
           Boogie.Expr seq = TrExpr(e.Seq);
           var seqType = e.Seq.Type.NormalizeExpand();
           if (seqType is SeqType) {
-            Type elmtType = cce.NonNull((SeqType)seqType).Arg;
             Boogie.Expr index = TrExpr(e.Index);
             index = translator.ConvertExpression(GetToken(e.Index), index, e.Index.Type, Type.Int);
-            Boogie.Expr val = BoxIfNecessary(GetToken(expr), TrExpr(e.Value), elmtType);
+            Boogie.Expr val = BoxIfNecessary(GetToken(expr), TrExpr(e.Value), e.Type);
             return translator.FunctionCall(GetToken(expr), BuiltinFunction.SeqUpdate, predef.BoxType, seq, index, val);
           } else if (seqType is MapType) {
             MapType mt = (MapType)seqType;
             Boogie.Type maptype = predef.MapType(GetToken(expr), mt.Finite, predef.BoxType, predef.BoxType);
-            Boogie.Expr index = BoxIfNecessary(GetToken(expr), TrExpr(e.Index), mt.Domain);
-            Boogie.Expr val = BoxIfNecessary(GetToken(expr), TrExpr(e.Value), mt.Range);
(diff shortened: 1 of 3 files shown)


cpitclaudel:
I have completed the investigation of the issue in https://github.com/dafny-lang/dafny/pull/2522 , so I'm comfortable closing this PR in favor of that one.

PR #2520, opened 2022-07-29
fix: Improve the performance of proofs involving bit vector shifts

atomb:
When encountering `x << y`, where x has width w, rewrite it to `x << (y as bvu)` where `u` is `log2(w)`, and similarly for (or `x >> y`. This ensures that the solver knows that most of the bits of the shift amount are zero.

This is an alternative implementation of #2494

Fixes https://github.com/dafny-lang/dafny/issues/2174
Fixes https://github.com/dafny-lang/dafny/issues/2230

By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/dafny-lang/dafny/blob/master/LICENSE.txt).


cpitclaudel (commented):
I like this version a lot :)

atomb:
Me, too. My only hesitation has to do with the general philosophy of how we do rewriting. Should we generally make all the AST fields mutable so that rewriting passes can change them as needed? I like the fact that they're immutable now, but maybe it's a reasonable price to pay.

atomb:
This approach has the unfortunate result of turning "shift amount must not exceed the width of the result (n)" error messages into "value to be converted might not fit in bv(log2 n)".

cpitclaudel:
> This approach has the unfortunate result of turning "shift amount must not exceed the width of the result (n)" error messages into "value to be converted might not fit in bv(log2 n)".

Right, but as we discussed on the other thread, could we attach a custom message to this?

atomb:
> > This approach has the unfortunate result of turning "shift amount must not exceed the width of the result (n)" error messages into "value to be converted might not fit in bv(log2 n)".
> 
> Right, but as we discussed on the other thread, could we attach a custom message to this?

It's possible. I'll look into what that'd take.

atomb:
Attaching a custom message, the way things are currently structured, would be awkward. However, I've realized that the current situation isn't that `shift amount must not exceed the width of the result (n)` gets _replaced_ by `value to be converted might not fit in bv(log2 n)`, but that the second message is emitted _in addition_ (and there are some cases where only the former is emitted --- specifically when the shift amount might be in between `n` and `2*n - 1`, inclusive).

atomb:
The most recent version has a slight tweak. Take the following Dafny function:

```dafny
function F(x: bv128, y: int): bv128
  requires 0 < y
{
  x << y
}
```

Dafny now outputs the following messages:
```
bv-shift-large.dfy(4,4): Error: shift amount must not exceed the width of the result (128)
bv-shift-large.dfy(4,7): Error: when converting shift amount to a bit vector, the value to be converted might not fit in bv8
```

I could imagine a user being confused about why the shift amount is being converted to a bit vector (and why `bv8`?), but overall it's not terrible.

Source/Dafny/Resolver/BitvectorOptimization.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Microsoft.Dafny;
+
+public class BitvectorOptimization : IRewriter {
+  public BitvectorOptimization(ErrorReporter reporter) : base(reporter) { }
+
+  internal override void PostResolveIntermediate(ModuleDefinition m) {
+    var visitor = new BitvectorOptimizationVisitor();
+    foreach (var decl in ModuleDefinition.AllCallables(m.TopLevelDecls)) {
+      visitor.Visit(decl);
```
cpitclaudel, line 17: Will this miss cover everything (including consts?)

cpitclaudel (approved):
I like it!

Source/Dafny/Resolver/BitvectorOptimization.cs
```diff
@@ -0,0 +1,43 @@
...
+
+public class BitvectorOptimizationVisitor : BottomUpVisitor {
+  private bool IsShiftOp(BinaryExpr.Opcode op) {
+    return op is BinaryExpr.Opcode.LeftShift or BinaryExpr.Opcode.RightShift;
+  }
+
+  private Expression ShrinkBitVectorShiftAmount(Expression expr, BitvectorType originalType) {
+    var width = new BigInteger(originalType.Width);
+    var intermediateType = new BitvectorType((int)width.GetBitLength());
+    var newExpr = new ConversionExpr(expr.tok, expr, intermediateType, "when converting shift amount to a bit vector, the ");
```
keyboardDrummer, line 30: This error message doesn't occur in the new test cases, and I didn't find it in existing ones. Shouldn't we test for this message?

Source/Dafny/AST/DafnyAst.cs
```diff
@@ -10821,11 +10821,13 @@ public TypeUnaryExpr(IToken tok, Expression expr, Type toType)
...
 
   public class ConversionExpr : TypeUnaryExpr {
-    public ConversionExpr(IToken tok, Expression expr, Type toType)
+    public readonly string messagePrefix;
+    public ConversionExpr(IToken tok, Expression expr, Type toType, string msgPrefix = "")
       : base(tok, expr, toType) {
       Contract.Requires(tok != null);
       Contract.Requires(expr != null);
       Contract.Requires(toType != null);
+      messagePrefix = msgPrefix;
```
keyboardDrummer, line 10830: Why use inconsistent names? I would expect `this.messagePrefix = messagePrefix`

Source/Dafny/Resolver/BitvectorOptimization.cs
atomb: I'll take a deeper look to make sure.

atomb: Hmm. Constants are `ICallable` but `AllCallables` has specific logic to skip them? That's weird. Using `AllItersAndCallables` seems better. It doesn't skip constants and also includes `IteratorDecl` entries.

cpitclaudel (approved):

cpitclaudel:
I'm realizing now that I forgot to ask for a release notes entry :/

atomb:
> I'm realizing now that I forgot to ask for a release notes entry :/

Oh, right! I'll create a new PR with that.

PR #2525, opened 2022-08-01
fix: Name collision in `datatypes` of some compilers

### LINKED ISSUE #2441: Generated Java vulnerable to name confusion
### **What is happening**?

I have `ADataType` that depends on `AnotherDataType` that can only be `A_CONSTANT`.

I compile that from Dafny to Java.
I try to compile the Generated Java.
Java is an unhappy.

### **How can any one reproduce this?**
Download this: [July18.zip](https://github.com/dafny-lang/dafny/files/9135427/July18.zip)
Unzip it: `mkdir July18; unzip July18.zip -d July18`
Glance at the `build.zsh` to ensure that I am not hacking your computer: `less July18/build.zsh`.
See that I depend on `dafny` and `mvn`.
Run the build: `cd July18; ./build.zsh`.
See the Compile error.

### **What is expected?**
I expect valid Dafny to become Valid Java.

### **I have a fix suggestion**
When generating Java, fully qualify Dafny generated Java objects.

The issue is that I have been a jerk and ~~ambiguously~~ poorly named the field
of my `ADataType` to be identical to the field's Datatype's name.

Thus Java is upset, 
because it appears to be using a non-static field
in a static method.

Which is not Dafny's intent,
it is trying to use the Datatype's static method.

Which can be achieved by changing the manner the `Default` method is called,
by fully qualifying it as `Example.Types.AnotherDataType.Default`.

### **What is my environment**:
I am using: `Dafny 3.6.0.40511` on a Mac OS X ( `11.6.6 (20G624)` ).
The `pom.xml` specifies Java-11.

fabiomadge:
Fixes #2441.


Test/git-issues/git-issue-2441.dfy
```diff
@@ -0,0 +1,7 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
```
robin-aws, line 1: Does this work for every target language now? Can we use `%testDafnyForEachCompiler`?

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -652,26 +652,27 @@ public class CsharpCompiler : SinglePassCompiler {
...
       string PrintInvocation(Formal f, int i) {
-        var prefix = customReceiver
-          ? datatype.IsRecordType || !f.HasName ? $"(({DtCtorDeclarationName(ctor)}{typeArgs}) _this)." : "_this.dtor_"
-          : "";
-        var (name, type) = ($"{prefix}{FormalName(f, i)}", f.Type);
+        var name = customReceiver
+          ? datatype.IsRecordType || !f.HasName
+            ? $"(({DtCtorDeclarationName(ctor)}{typeArgs}) _this).{FieldName(f, i)}"
+            : $"_this.{DestructorGetterName(f, ctor, i)}"
+          : FieldName(f, i);
```
robin-aws, line 659: I find nested tertiary operators hard to read, I think it's worth making this a few separate statements.

robin-aws (changes requested):
I like this solution a little better than https://github.com/dafny-lang/dafny/pull/2532 just because I think it's a slightly better improvement to readability of the generated code.

Test/git-issues/git-issue-2441.dfy
fabiomadge: No, python fail, but the fix is in a different PR. Go fails because of an unrelated issue.

Source/Dafny/Compilers/Compiler-Csharp.cs
fabiomadge: It's indented and everything.

robin-aws (approved):

PR #2530, opened 2022-08-02
Don't publish a diagnostics with a negative line and column number

### LINKED ISSUE #2529: Language server sends invalid diagnostic when a file is empty
For an empty .dfy file, the language server sends the diagnostic:

```
[Trace - 12:38:04 PM] Received notification 'textDocument/publishDiagnostics'.
Params: {
    "uri": "file:///.../test.dfy",
    "version": 2,
    "diagnostics": [
        {
            "range": {
                "start": {
                    "line": -1,
                    "character": -1
                },
                "end": {
                    "line": -1,
                    "character": 29
                }
            },
            "severity": 2,
            "source": "Parser",
            "message": "File contains no code: DafnyExamples/test.dfy",
            "relatedInformation": []
        }
    ]
}
```

Note the negative line and characters numbers, which are illegal. VSCode ignores this diagnostic.

keyboardDrummer:
Don't publish a diagnostics with a negative line and column number when a file is empty

Fixes #2529

### DIFF
--- a/Source/Dafny/AST/DafnyAst.cs
+++ b/Source/Dafny/AST/DafnyAst.cs
@@ -78,12 +78,12 @@ namespace Microsoft.Dafny {
     public int pos { get; set; } // Used by coco, so we can't rename it to Pos
 
     /// <summary>
-    /// One indexed
+    /// One-indexed
     /// </summary>
     public int col { get; set; } // Used by coco, so we can't rename it to Col
 
     /// <summary>
-    /// One indexed
+    /// One-indexed
     /// </summary>
     public int line { get; set; } // Used by coco, so we can't rename it to Line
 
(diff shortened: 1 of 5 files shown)


Source/Dafny/AST/DafnyAst.cs
```diff
@@ -56,28 +56,35 @@ public interface IToken : Microsoft.Boogie.IToken {
...
 
-    public Token() : this(0, 0) {
-    }
+    public Token() : this(0, 0) { }
 
     public Token(int linenum, int colnum) {
       this.line = linenum;
       this.col = colnum;
-      this.val = "anything so that it is nonnull";
+      this.val = "";
```
cpitclaudel, line 71: I'm fine with this, but note that having this odd value is somewhat nice, beause it lets you see clearly that this is the default value

cpitclaudel (approved):
LGTM

Source/Dafny/AST/DafnyAst.cs
MikaelMayer: An empty val, even if given a correct line and column, might result in no error displayed in VSCode. Can you please double check?

MikaelMayer (changes requested):

MikaelMayer (approved):

PR #2534, opened 2022-08-02
Add top-level commands to Dafny and redesign the CLI UI

### LINKED ISSUE #2036: Introduce top level commands commands for Dafny
Related discussion in: https://github.com/dafny-lang/dafny/issues/1322#issuecomment-889023432

### Goal
Update the Dafny CLI so it _can_ accept as its first argument a "command". When no command is passed, the CLI remains unchanged. Commands correspond to use-cases. When a command is used, other options may no longer be available since the command specifies them. The options `/compile` and `/spillTargetCode` are not available when using commands.

- `dafny verify <inputs>`, runs verification, but does not emit anything. Useful in CI and when developing without an IDE.
  - Similar to the current `dafny /compile:0`
- `dafny run <inputs>`, runs the Dafny program. Useful for running Dafny applications during development.
  - Similar to the current `dafny /compile:3`. 
  - `dafny run --no-verify` runs the program without verifying it, similar to `dafny /noVerify /compile:4`
  - The option `--debug=<port>` will start a Debug server protocol server that listens at the specified port.
  - By default, no intermediate generated files are visible, although there may be an option to show them, like gcc's [save temps](https://renenyffenegger.ch/notes/development/languages/C-C-plus-plus/GCC/options/save-temps/index)
  - By default uses a Dafny back-end that provides a mix of fast compilation, fast execution, and few required dependencies. .NET is a sensible default since .NET is already a required dependency for calling the Dafny CLI.
  - When input files other than .dfy are specified, such as `.cs`, `.dll`, `.java`, `.jar`, these may be used to implement methods annotated with `{:extern}` in the Dafny sources. Only input files that use the same back-end type may be used. The option `--target` may be used to force using a particular back-end.
 
- `dafny build [--os] [--arch] <inputs>`, creates a binary that can either be executed directly on the specified platform and architecture, or using `dafny run` when a platform and architecture independent build is made. Useful when creating a Dafny-only application or library.
  - Example: `dafny build --os=win --arch=x64 <sourceFile>`, will create a `.exe` binary that can be executed on a 64 bit Windows machine.
  - Example: `dafny build <sourceFile>` will create a file that can be executed on any machine using `dafny run <builtFile>`. We leave the format of the built file unspecified for now.
  - Similar to the current `dafny /compile:1`. 
  - `dafny build --no-verify` builds the program without verifying it, similar to `dafny /noVerify /compile:2`
  - By default uses a Dafny back-end that provides fast execution, and few required dependencies. .NET is a sensible default since .NET is already a required dependency for calling the Dafny CLI. 
  - The back-end may be configured using `--target`

- `dafny translate <target> <inputs>`, compiles to target language source and build files. The end result can immediately be invoked by a target language build tool. Useful for creating libraries to be used by an application written in another language.
  - Similar to the current `dafny /compile:0 /spillTargetCode:2` .
  - `dafny translate --no-verify` skips verification and produces target source code, similar to `dafny /noVerify /compile:0 /spillTargetCode:3`.
  - Targets are defined as subcommands of `translate`, and as such may define additional options and arguments. For example, `java` may take an additional argument `[gradle|maven]` to specify for which build system to generate files for, so a user could invoke `dafny translate java gradle <inputFile>`. Since targets are subcommands, help can be shown for them using `dafny translate <target> --help`.
  - For C#, _should_ generate a `.csproj` file that references the Dafny Runtime NuGet package and the required .NET version.
  - For JavaScript, _should_ generate a `package.json` that specifies a Dafny Runtime NPM package as a dependency, and possibly a `jsconfig.json` to specify which ECMAScript version is required.
  - For Python, _should_ generate a `pyproject.toml` that specifies the required python version and the Python Dafny Runtime as a dependency.

#### A translate option that doesn't require a package manager
The current `dafny translate` produces files that rely on a package manager in the target language. For users who don't want to use a package system, Dafny could provide an option to produce sources that can be easily included as generated source in another project. This would mean the Dafny runtime is emitted as source, and possibly it would be good to package all the sources together in a single large source file for easy inclusion.

#### Combined commands
Users who want to get the output of different Dafny commands together, can do so by running several commands separately. For example, a customer who wants .cs sources, a .dll, and to run that .dll, can do:
```
dafny translate cs inputFile.dfy --out=translateFolder
dotnet build translateFolder/inputFile.csproj --out=buildFolder
dotnet buildFolder/inputFile.dll
```

Alternatively, Dafny could offer combined commands, like `translate-build-run`.

### Out of scope
- `dafny package/publish`, commands related to publishing build artifacts.
- `dafny [test|check]` runs tests. Useful when developing and in CI.
  - Similar to the current `dafny /compile:0`
  - Passing `--no-verify` will skip verification
- Discussion of what hidden options are available for helping developers of Dafny

### Implementation hints
We expect a straightforward implementation

### LINKED ISSUE #2360: Dafny 4 proposal: Make `-xyz` mean the same as `-xyz:1` when `-xyz` accepts only `:0` and `:1`
We have `-showSnippets:<0|1>` but `-printTooltips` (no `:1`, my fault) and `-funcCallGraph` (not my fault).  This is confusing and unnecessary.

Also, let's move to named options (as in `-printIncludes:<None|Immediate|Transitive>`) instead of numeric ones (`-compile:<0-4>`).

keyboardDrummer:
Fixes #2036
Fixes #2360

# Changes
- Introduce three commands for the Dafny CLI: `verify`, `build`, `run`, as defined in #2036
- When commands are used, use a redesigned UI for the CLI
  - Use proper GNU options 
    - (`-` and `--` for short and long names, instead of `-` and `/` for long names)
    - Use `=` instead of `:` for specifying the value of an option in a single term
  - `--<option>` will mean the same as `--<option>:1` for boolean options
  - Boogie options can only be specified by starting with the prefix `boogie-`, although some Boogie options are exposed at the Dafny level:
    - `vscCores` is exposed as `cores` 
    - `useBaseNameForFileName` is exposed
    - `timeLimit` is exposed as `verificationTimeLimit`
  - Rename several options
    - `compileTarget` is renamed to `target`
    - `dprint` is renamed to `print`
- There is a new mechanism for adding options, that allows adding options to both the new and old CLI UI. `showSnippets` is an existing option that has been refactored to make use of this new mechanism.
- Disable the use of "compile" "spillTargetCode" and "dafnyVerify" when using commands.
- `dafny <command> --help` shows which options there are for that specific command.

Output from `dafny --help`:

```
Usage: dafny [dafny-options] [command] [command-options] [arguments]

Execute a Dafny command.

  ---- dafny-options  --------------------------------------------------------

  -h|--help         Show command line help.
  --version         Display Dafny version in use.
  --attrHelp        Print a message about supported declaration attributes

  ---- Dafny commands  -------------------------------------------------------

  verify            Verify the program.
  build             Generate source files in the target language.
  run               Run the program.
```

Output from `dafny run --help`

```
Usage: dafny [dafny-options] [command] [command-options] [arguments]

Execute a Dafny command.

  ---- dafny-options  --------------------------------------------------------

  -h|--help         Show command line help.
  --version         Display Dafny version in use.
  --attrHelp        Print a message about supported declaration attributes

  ---- General options -------------------------------------------------------

  --cores=<value>
      Run the Dafny CLI using <n> cores. Defaults to 1.
  --help=<value>
      Display this help text
  --useBaseNameForFileName=<value>
      When parsing use basename of file for tokens instead
      of the path supplied on the command line

  All the .dfy files supplied on the command line along with files recursively
  included by 'include' directives are considered a single Dafny program;
  however only those files listed on the command line are verified.
  
  Exit code: 0 -- success; 1 -- invalid command-line; 2 -- parse or type errors;
             3 -- compilation errors; 4 -- verification errors
  
  ---- Input configuration ---------------------------------------------------
  
  --dprelude=<file>
      choose Dafny prelude file
  --stdin
      Read standard input and treat it as an input .dfy file.

  
etc...
```

### Future work
- Replace all option definitions in DafnyOptions by options defined through a subclass of `ICommandLineOption`.
- Expose `--boogie-print` as `dafny build --target=boogie`
- Allow options that take many numeric values to also take string based values that better describe what the value does. Note that this won't be a breaking change.
- Expose more Boogie options as first class Dafny options, so they don't require using `--boogie-`
- Make Dafny's integrated CLI documentation more interactive, so it's easier to navigate.
- Change `--noVerify` so it does not invoke Boogie.

### Implementation

There's now two UIs for the CLI, the old (`-` and `/`) and the new (`--`). Both styles can use the same option definitions. There's now two styles of defining options, the old and a new style. The new style is more declarative and flexible, it allows specifying the option only for the new or old style, or both.

### Testing
- `comp/separateCompilation/usesTimesTwo.dfy`  and `git-issues/github-issue-305.dfy` use `dafny build`
- `comp/Arrays.dfy` and `comp/AsIs-Compile` use `dafny verify` and `dafny run`
- Most of the tests using commands use longname options, and `comp/AsIs-Compile` also uses shortname options
- `dafny0/AssumptionVariables1` uses boogie options through `--boogie-<optionName>`
- `git-issue-2197.dfy` tests the boolean option `showSnippets` without explicit truthy value, while `ShowSnippets.dfy` tests it with explicit thruthy value.


cpitclaudel:
It would be nice to address https://github.com/dafny-lang/dafny/issues/2360 at the same time.  Because of the ambiguity that @RustanLeino noted earlier, we won't be able to do it for the main command line, so if we're building a new command line it would be the perfect opportunity to do other cleanups.

Other candidates would be to do proper gnu options (`-` for short and `--` for long options), and probably to rename / delete a bunch of old options.

keyboardDrummer:
> It would be nice to address #2360 at the same time. Because of the ambiguity that @RustanLeino noted earlier, we won't be able to do it for the main command line, so if we're building a new command line it would be the perfect opportunity to do other cleanups.
> 
> Other candidates would be to do proper gnu options (`-` for short and `--` for long options), and probably to rename / delete a bunch of old options.

I will also do #2360 and proper gnu options in this PR.

cpitclaudel:
Thanks for doing this!  I will look at the implementation soon.
Some references that may be useful:

- [GNU Standards for Command Line Interfaces](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html)
- [List of all GNU options](https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table), relevant to avoid confusion / promote consistency
- [POSIX guidelines for CLI](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html)
- [Summary of the POSIX guidelines](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html)

docs/DafnyRef/Expressions.md
```diff
@@ -975,13 +975,13 @@ and [expressions](#sec-match-expression).
 `CasePattern`s are used
 in `LetExpr`s and `VarDeclStatement`s.
 The `ExtendedPattern` differs from `CasePattern` in allowing literals,
-symbolic constants, and disjunctive (“or”) patterns.
+symbolic constants, and disjunctive (???or???) patterns.
```
cpitclaudel, line 978: Small encoding issue here

docs/DafnyRef/Attributes.md
```diff
@@ -510,7 +510,7 @@ BOUNDVARS = ID : ID
 ### 23.2.23. `{:options OPT0, OPT1, ... }` {#sec-attr-options}
 
 This attribute applies only to modules. It configures Dafny as if
-`OPT0`, `OPT1`, … had been passed on the command line.  Outside of the module,
+`OPT0`, `OPT1`, ??? had been passed on the command line.  Outside of the module,
```
cpitclaudel, line 513: Small encoding issue here

Source/Dafny/Options/TargetOption.cs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    return new ParsedOption(compiler);
+  }
+
+  public override void PostProcess(DafnyOptions options) {
+    options.Compiler = Get(options);
+    base.PostProcess(options);
+  }
+}
```
cpitclaudel, line 47: Is your IDE configured to remove final newlines?

Test/git-issues/git-issue-645.dfy
```diff
@@ -1,4 +1,4 @@
 // dafny should emit exit value 1
-// RUN: %dafny_0 /xyz    > "%t"
+// RUN: %dafny_0 /xyz    > "%t" || true
```
cpitclaudel, line 2: I though `dafny_0` was introduced to avoid having to add `||true`.  I like `||true`, but then do we still need `%dafny_0`?

Source/Dafny/Options/TargetOption.cs
```diff
@@ -0,0 +1,47 @@
...
+  public override string ShortName => "t";
+  public override string Category => "Compilation options";
+
+  public override string Description => @"
+cs (default) - Compilation to .NET via C#
+go - Compilation to Go
+js - Compilation to JavaScript
+java - Compilation to Java
+py - Compilation to Python
+cpp - Compilation to C++
```
cpitclaudel, line 23: Should this string be dynamically generated based on available compiler plugins?

cpitclaudel (commented):
I've looked at the implementation.  It seems reasonably, but I would much prefer to not do the command line parsing ourselves.  Does `System.CommandLine` have all the we need, or is it missing something crucial?  Given that it's part of the standard library of .net I would love to move to that if we can.

Beyond that, some small things:
- I would like to replace numeric enum flags everywhere. For `-xyz:1` we would move to `--xyz` and  `--no-xyz`, or to `--xyz=false`, or `--xyz=off`, or something like that (which you pretty much did, thanks!), and for things like `/compile:<n>` we could move to named enums instead of numeric ones (I think we're on the same page, just putting it here.
- We might remove the list of codegen backends from `/compileTarget` (or `--target`) and instead have a separate flag (or section in the help output) to print the list, dynamically generated form the currently loaded codegen backends.

docs/DafnyRef/Expressions.md
keyboardDrummer: Oops, sorry!

Source/Dafny/Options/TargetOption.cs
keyboardDrummer: I don't think so. What's the deal with those? I've never paid attention to them.

Seems like we could add 

```
[*]
insert_final_newline = true
```

to our `.editorconfig` to let IDEs add these newlines.

keyboardDrummer:
> I've looked at the implementation. It seems reasonable (and good!), but I would much prefer to not do the command line parsing ourselves. Does `System.CommandLine` have all the we need, or is it missing something crucial? 

It looks great. However, I'm doubtful we can use any off the shelf solution as long as we support both the old and new CLI UI and share code between those and with Boogie. Once we drop the old CLI UI, I'd love to clean up the code and use this. Alternatively we could not share code between the old and new UI, and already use this for the new UI, but I think that only increases the total amount of work required.

> Given that it's part of the standard library of .net I would love to move to that if we can.

It's in a NuGet package. Would you still call that part of the standard library? (just curious)

> * I would like to replace numeric enum flags everywhere. For `-xyz:1` we would move to `--xyz` and  `--no-xyz`, or to `--xyz=false`, or `--xyz=off`, or something like that (which you pretty much did, thanks!), and for things like `/compile:<n>` we could move to named enums instead of numeric ones (I think we're on the same page, just putting it here.

Why don't you want 1 and 0 for boolean options? Wouldn't it be useful to be able to specify `--some-boolean-option=$FOO`, where it might be slightly nicer if $FOO can be `1/0` instead of `on/off`? Is there an ambiguity difference between the two sets?

I think changing the argument values of Boogie options is out of scope for this PR. Note that we can always add support for string based argument values later, without dropping support for numeric values, so then there's no breaking change and it doesn't need to be done in this PR.

For numeric value options in Dafny, here's what I suggest:

```
/optimizeResolution:<n>
    0 - Resolve and translate all methods
    1 - Translate methods only in the call graph of current verification target
    2 (default) - As in 1, but only resolve method bodies in non-included Dafny sources
```
Doesn't seem useful. Also judging by the back-end code, 1 and 2 do exactly the same thing. I suggest we drop support in the new CLI.

```
/induction:<n>
    0 - never do induction, not even when attributes request it
    1 - only apply induction when attributes request it
    2 - apply induction as requested (by attributes) and also
        for heuristically chosen quantifiers
    3 - apply induction as requested, and for
        heuristically chosen quantifiers and lemmas
    4 (default) - apply induction as requested, and for lemmas
/inductionHeuristic:<n>
    0 - least discriminating induction heuristic (that is, lean
        toward applying induction more often)
    1,2,3,4,5 - levels in between, ordered as follows as far as
        how discriminating they are:  0 < 1 < 2 < (3,4) < 5 < 6
    6 (default) - most discriminating
```
Would probably be better if it supported flags `--induction=(attributes|heuristics|lemma)+`, but that might require some back-end code changes. I suggest we leave this for now.

`/arith` Its description reads "(Experimental switch. Its options may change.)". It accepts 10 possible values. To prevent our users from learning names that will change, I suggest we keep the numeric enum flags.

`/definiteAssignment`: "ignore", "default", "all", "deterministic"
`/deprecation`. The third option "also point out where there's new simpler syntax" only has a single usage, and such a feature seems better suited somewhere else. I suggest we don't support the third value in the new UI and make this option have a value.

`/allocated`: "nowhere", "nonghost", "heapful", "default", "antecedents"

> * We might remove the list of codegen backends from `/compileTarget` (or `--target`) and instead have a separate flag (or section in the help output) to print the list, dynamically generated form the currently loaded codegen backends.

Seems like that only effects the help, and doesn't need to be part of this PR. I'd prefer to restrict the scope as much as possible since it's already bigger than I'd like.

Test/git-issues/git-issue-645.dfy
keyboardDrummer: dafny_0 adds the argument `/countVerificationErrors:0`. In this PR that argument no longer swallows errors in command line arguments, but only in verification errors.

davidcok: I thought this option was deprecated already. If not, it certainly needs renaming. I think the only use was to not have expected verification errors in tests cause the test to fail.

PR #2538, opened 2022-08-02
Fixed bodyless methods gutter icons

### LINKED ISSUE #2533: Weird gutter icons
I'm writing specifications around extern methods, and the highlighting on the left seems broken:

sample code:

```dafny
method test() {
    assert false;
}

method {:extern} test3(a: nat, b: nat)
  ensures true
  ensures test2(a - b)
  ensures true
  ensures test2(a - b)
  ensures true

predicate method test2(x: nat) {
    assert false; true
}
```

Gutter icons:
image

MikaelMayer:
Fixes #2533 
For that, I also added EndToken for top-level declarations, and split the Range of VerificationTree into a Range and a token. The token is the one that is used for finding the place of a tree and is included in the range, but it is not its start anymore.

It seems that some tests have changed because of this refactoring but I have no idea why. Their results seem better than before though.

### DIFF
--- a/Source/DafnyLanguageServer.Test/GutterStatus/SimpleLinearVerificationGutterStatusTester.cs
+++ b/Source/DafnyLanguageServer.Test/GutterStatus/SimpleLinearVerificationGutterStatusTester.cs
@@ -116,11 +116,11 @@ public class SimpleLinearVerificationGutterStatusTester : LinearVerificationGutt
   [TestMethod/*, Timeout(MaxTestExecutionTimeMs)*/]
   public async Task EnsuresAddingNewlinesMigratesPositions() {
     await VerifyTrace(@"
- .  S [S][ ][I][S][S][ ][I][S][S][ ]:method f(x: int) {
- .  S [S][ ][I][S][S][ ][I][S][S][ ]:  //Next1:\n  //Next2:\n  
- .  S [=][=][I][S][S][ ][I][S][S][ ]:  assert x == 2; }
-            [-][~][=][=][I][S][S][ ]:
-                        [-][~][=][=]:");
+ .  S [S][ ][I][S][ ][I][S][ ]:method f(x: int) {
+ .  S [S][ ][I][S][ ][I][S][ ]:  //Next1:\n  //Next2:\n  
+ .  S [=][=][I][S][ ][I][S][ ]:  assert x == 2; }
+            [-][~][=][I][S][ ]:
+                     [-][~][=]:");
   }
 
   [TestMethod/*, Timeout(MaxTestExecutionTimeMs)*/]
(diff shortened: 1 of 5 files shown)


Source/DafnyLanguageServer.Test/GutterStatus/SimpleLinearVerificationGutterStatusTester.cs
```diff
@@ -145,4 +145,41 @@ public class SimpleLinearVerificationGutterStatusTester : LinearVerificationGutt
...
+ .  S [S][ ]:  ensures true
+ .  S [=][=]:  ensures test2(a - b)
+ .  S [S][ ]:  ensures true
+ .  S [O][O]:  ensures test2(a - b)
+ .  S [S][ ]:  ensures true
+    |  |  | :
+ .  |  |  | :predicate method test2(x: nat) {
+ .  |  |  | :  true
+ .  |  |  | :}");
+  }
```
cpitclaudel, line 184: These tests look really nice ❤️

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -156,8 +156,10 @@ public record VerificationTree(
      // Used to re-trigger the verification of some diagnostics.
      string Identifier,
      string Filename,
-     // The range of this node.
-     Range Range
+     // The range of this node. Used to rendering
```
cpitclaudel, line 159: Small grammar issue

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -156,8 +156,10 @@ public record VerificationTree(
...
+     Range Range,
+     // The position of this node as believed by Boogie (i.e. the symbol name)
```
cpitclaudel, line 161: I don't understand this comment

Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
```diff
@@ -133,13 +137,14 @@ INotificationPublisher notificationPublisher
...
-          var verificationTreeRange = subsetTypeDecl.tok.GetLspRange(subsetTypeDecl.BodyEndTok);
+          var verificationTreeRange = subsetTypeDecl.StartToken.GetLspRange(subsetTypeDecl.EndToken);
           var verificationTree = new TopLevelDeclMemberVerificationTree(
             $"subset type",
             subsetTypeDecl.Name,
             subsetTypeDecl.CompileName,
             subsetTypeDecl.tok.Filename,
-            verificationTreeRange);
+            verificationTreeRange,
+            subsetTypeDecl.tok.GetLspPosition());
```
cpitclaudel, line 147: (When) will these positions be different from the first token of the `verificationTreeRange`?

cpitclaudel (changes requested):
Looks good to me modulo a small typo in a comment, plus a question about the relation between `Position` and `Ranges`.  Could you give an intuition as a comment about why the position isn't the same as the beginning of the range any more?  Is it because the range covers more?

It would be great to have a small example before the definition of `VerificationTree`.

Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
MikaelMayer: I'm going to explain it in an example in a new PR, but typically, the tok is the symbol name, whereas the range extends to the left (keywords) and to the right of this symbol name (definition)

cpitclaudel (approved):
Excellent, thanks!

PR #2546, opened 2022-08-04
Fix bug in nested match 

### LINKED ISSUE #2539: /warnMissingConstructorParenthesis warns in expressions but not in statements
The following example code produces two warnings, not four, when compiled with `/warnMissingConstructorParenthesis`:

```
datatype T = A | B

method M(t: T) { // No warnings
  match t
    case A => print "A";
    case B => print "B";
}

method M'(t: T) { // 2 warnings
  print match t
    case A => "A"
    case B => "B";
}
```

```
bug.dfy(11,9): Warning: Constructor name 'A' should be followed by parentheses
bug.dfy(12,9): Warning: Constructor name 'B' should be followed by parentheses
```

Additionally, maybe the flag should be `/warnMissingConstructorParentheses`?

tobiaj10:
Fixes #2539


Source/Dafny/ConstructorWarning.cs
```diff
@@ -44,6 +44,15 @@ class ConstructorWarningVisitor : TopDownVisitor<Unit> {
     }
     return base.VisitOneExpr(expr, ref st);
   }
+  protected override bool VisitOneStmt(Statement stmt, ref Unit st) {
+    if (stmt is NestedMatchStmt matchStmt) {
+      var matchStmtCases = matchStmt.Cases;
```
cpitclaudel, line 49: Since this variable is used just once, you can inline it: either by changing the `foreach` below to refer to `matchStmt.cases` or by changing the pattern match above to bind `Cases` (look up the documentation on pattern matching if you're not sure how to do that)

Test/linters/constructorCaseWithoutParenthesis.dfy
```diff
@@ -1,10 +1,16 @@
-// RUN: %dafny /compile:0 /warnMissingConstructorParenthesis "%s" > "%t"
+// RUN: %dafny /compile:0 /warnMissingConstructorParentheses "%s" > "%t"
```
cpitclaudel, line 1: I would suggest renaming this test file as well.

cpitclaudel (changes requested):
Looks great, I've suggested two small improvements.

PR #2547, opened 2022-08-04
Do not publish verification symbol status for unresolved documents

keyboardDrummer:
Fix: prevents the LSP client from thinking things are verified when really they're not even resolving.

### DIFF
--- a/Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
+++ b/Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
@@ -93,7 +93,7 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
         statusPublisher.SendStatusNotification(textDocument, CompilationStatus.ParsingFailed);
         return CreateDocumentWithEmptySymbolTable(loggerFactory.CreateLogger<SymbolTable>(), textDocument,
           errorReporter.GetDiagnostics(textDocument.Uri), program,
-          wasResolved: true, loadCanceled: false);
+          wasResolved: false, loadCanceled: false);
       }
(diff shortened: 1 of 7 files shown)


cpitclaudel (approved):

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -93,7 +93,7 @@ INotificationPublisher notificationPublisher
         statusPublisher.SendStatusNotification(textDocument, CompilationStatus.ParsingFailed);
         return CreateDocumentWithEmptySymbolTable(loggerFactory.CreateLogger<SymbolTable>(), textDocument,
           errorReporter.GetDiagnostics(textDocument.Uri), program,
-          wasResolved: true, loadCanceled: false);
+          wasResolved: false, loadCanceled: false);
```
MikaelMayer, line 96: Oh I see why you can change it from true to false. I remember when I added this variable, it used to mean `resolutionFinished: true`, meaning the resolution phase was completed. But it looks like this field isn't used anywhere anymore so feel free to reuse it for your means.

MikaelMayer (approved):
I don't have anything against, but can you please confirm me that it won't break the following scenario:
- Gutter icons are sent anyway (with migrated verification information displayed in gray) even if there is a parse or resolution error
- Are the test symbols migrated with that? Otherwise, they could appear on top of gutter icons.

keyboardDrummer:
> * Are the test symbols migrated with that? Otherwise, they could appear on top of gutter icons.

Yes, VSCode migrates them automatically.

PR #2566, opened 2022-08-09
Fix verification status for function by methods, and two other small issues

keyboardDrummer:
### Changes
1. Consider a function by method as a single verifiable
1. Stop sending duplicate verification status notifications
1. Fix a warning producing piece of code in Dafny.atg

### Tests
Added tests for 1 and 2.

### DIFF
--- a/Source/Dafny/Dafny.atg
+++ b/Source/Dafny/Dafny.atg
@@ -2608,7 +2608,7 @@ FunctionSpec<.List<AttributedExpression> reqs, List<FrameExpression> reads, List
   .
 
 /*------------------------------------------------------------------------*/
-PredicateResult<string name, out Formal? result>
+PredicateResult<string name, out Formal result>
 = (. Type/*!*/ returnType = new BoolType();
      result = null;
   .)
--- a/Source/Dafny/Resolver.cs
+++ b/Source/Dafny/Resolver.cs
@@ -2334,7 +2334,7 @@ namespace Microsoft.Dafny {
       var receiver = f.IsStatic ? (Expression)new StaticReceiverExpr(tok, cl, true) : new ImplicitThisExpr(tok);
       var fn = new FunctionCallExpr(tok, f.Name, receiver, tok, tok, f.Formals.ConvertAll(Expression.CreateIdentExpr));
       var post = new AttributedExpression(new BinaryExpr(tok, BinaryExpr.Opcode.Eq, r, fn));
-      var method = new Method(tok, f.Name, f.HasStaticKeyword, false, f.TypeArgs,
+      var method = new Method(f.tok, f.Name, f.HasStaticKeyword, false, f.TypeArgs,
         f.Formals, new List<Formal>() { resultVar },
         f.Req, new Specification<FrameExpression>(new List<FrameExpression>(), null), new List<AttributedExpression>() { post }, f.Decreases,
         f.ByMethodBody, f.Attributes, null, true);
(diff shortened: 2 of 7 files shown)


Source/Dafny/Dafny.atg
```diff
@@ -2608,7 +2608,7 @@ FunctionSpec<.List<AttributedExpression> reqs, List<FrameExpression> reads, List
   .
 
 /*------------------------------------------------------------------------*/
-PredicateResult<string name, out Formal? result>
+PredicateResult<string name, out Formal result>
```
MikaelMayer, line 2611: I don't think you can remove the question mark, unless now predicates always return a formal ^^

Source/Dafny/Resolver.cs
```diff
@@ -2334,7 +2334,7 @@ public class ModuleBindings {
       var receiver = f.IsStatic ? (Expression)new StaticReceiverExpr(tok, cl, true) : new ImplicitThisExpr(tok);
       var fn = new FunctionCallExpr(tok, f.Name, receiver, tok, tok, f.Formals.ConvertAll(Expression.CreateIdentExpr));
       var post = new AttributedExpression(new BinaryExpr(tok, BinaryExpr.Opcode.Eq, r, fn));
-      var method = new Method(tok, f.Name, f.HasStaticKeyword, false, f.TypeArgs,
+      var method = new Method(f.tok, f.Name, f.HasStaticKeyword, false, f.TypeArgs,
```
MikaelMayer, line 2337:
```suggestion
      var method = new Method(tok, f.Name + "_by_method", f.HasStaticKeyword, false, f.TypeArgs,
```

I understand why you are doing that: That way, this method has the same symbol position as the function itself.
However, can you elaborate on why this results in a better user experience?

In my short-sighted mind, I pretty much see the by-method as syntactic sugar so 
that I don't need to write a separate method. So, I wouldn't mind that it results in two different verification symbols, even if the by-method symbol points to the "method" keyword - but it should be displayed with f.name + "'s method"

MikaelMayer (commented):
I'm wondering whether we could keep the by-method token to have two different verification symbols, but display them in a better way.
See the discussion I started below.

Source/Dafny/Dafny.atg
keyboardDrummer: The `?` syntax is not enabled for this project, so having the `?` creates a warning. Since the `?` syntax is not enabled, not having it does not imply that it cannot be null.

Source/Dafny/Resolver.cs
keyboardDrummer: Interesting idea, but your proposed change doesn't achieve that. No names are passed to the IDE by the symbolStatus API, only ranges, so we should only return ranges that map to names in the IDE.

Source/Dafny/Dafny.atg
MikaelMayer: Ok good point thanks.

Source/Dafny/Resolver.cs
MikaelMayer: Yes, my change does not achieve that, I was just suggesting a starting point.

Wouldn't it be the right moment to now evolve your API as I suggested in the past, to have a separate "DisplayName" from the "Range" in your API, and not assume that the range points to a valid name that you have to painfully and manually extract in the IDE?

keyboardDrummer: I prefer the current UI. We could discuss this design but I suggest we first fix the bug that the current UI has.

MikaelMayer (approved):
Great job fixing these issues !

Issue #2390, opened 2022-07-11
Implement lazy sequence concatenation in remaining backends


robin-aws:
This is currently only implemented for C# and Java, but is a pretty critical optimization.

I believe it is feasible to implement this logic once in Dafny code instead of N times in each target language.

robin-aws:
Implementation strategy is as described in #1775

robin-aws:
Here are more details on the implementation strategy:

Sequences are currently implemented in each runtime via one or more target language classes or similar datatypes, such as ISequence<T>, Sequence<T>, ArraySequence<T>, ConcatSequence<T> etc. in C#. I have [a similar common implementation in Dafny code](https://github.com/robin-aws/dafny/blob/dafny-runtime-in-dafny/Source/DafnyRuntimeDafny/src/sequence.dfy#L15), which will be compiled to each target language as externs with similar names. Then it is only a matter of tweaking code generation for each language to target the Dafny-generated types instead.

The remaining challenges that must be addressed are:

1. *Circular dependency*: the Dafny `Sequence<T>` code requires a fully-built `dafny` tool, but the core `DafnyPipeline` project currently depends on the `DafnyRuntime` project. The solution will likely be to build any Dafny runtime code with a previously-released version of Dafny, and to check in the generated code (with a CI mechanism to regenerate it and verify it matches).
2. *Monomorphization for numeric types*: the existing hand-rolled sequence implementations all at least provide versions of methods like `Select` and `Count` that accept and/or produce unbounded integer types, such as `BigInteger`s in C#. Several also provide multiple [overloads](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntime.cs#L879-L887) or alternate [versions](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntime.cs#L861-L862) that use more primitive types such as `int` or `long`. When compiling Dafny source code that uses `newtype` declarations that align with native types, these overloads help avoid the overhead of casting to and from such `BigInteger` types. The solution here will be for each compiler to track the set of native numeric types to use for indexing, and to generate these overloads for each such type.
[...]

PR #2399, opened 2022-07-12
Feat: Dafny format file.dfy and IDE extension

MikaelMayer:
This PR makes it possible to "format" one or many Dafny files on the command-line, which for now means only changing the indentation of every single line. This is a step towards #1658

Supported syntax:

* `dafny format file1.dfy file2.dfy file3.dfy...` Format each file provided in arguments.
* `dafny format -print:- file1.dfy` Format each file but instead of writing files, output them formatted in stdout
* `dafny format .` Look up for all the Dafny files recursively in the current repository and format them.

# Insights

Coming up with an opinionated formatter wasn't an easy task. We want the guarantee that the reprinted program should be the same as the previous one, modulo whitespace.
Our approach in this PR consists of three phases:

1. Convert tokens to a double linked list so that we can traverse them from the beginning to the end and store the first token in the program (Scanner.frame, and all AST nodes)
2. Traverse the program on pre-resolved nodes, and decide on the indentation of each token when it's associated to a declaration, a node, a statement, etc.
3. Reprint the tokens in their original order by replacing their leading and trailing whitespace by their correctly indented counterpart

Step 3. has been entirely written in Dafny, and offers the interesting guarantee that the final reprinted program will contain all the "val" of every token reachable from the first token provided in input.

## What is the indentation of a "token"?

This PR considers that each token has to be associated with 3 types of indentation.
1. The indentation about comments that are before this token
2. The indentation of this token itself if it starts a new line
3. The indentation of comments that are after this token.

Because the token printer will traverse the tokens in order, the indentation 1. is used only for the trivia associated to the token in the leading whitespace of that token, like `/* Comment about X */` in the following example, which is in the leading trivia of `const`

```
datatype Y :=
   | C
     // Comment about C
   | D
     // Comment about D

/* Comment about X */
const X :=
```
Note that in this same example, the indentation of `Comment about D` is possible because of the indentation stored in 3.

The indentation of 2. makes is possible to have the token to be a delimiter of two indented regions, like this:

```dafny
if X then
  Y
else // Here "else" has its own indentation
  Z
```

# How this PR has been tested?

* I added the test Formatter.cs in the pipeline with 64 different tests, where almost each test starts from a totally unindented program, verifies the new indent, and reindent it once more to ensure the indentation is idempotent.
* I added the test FormatterTest.cs in the language server to ensure it behaves as expected.

# Rules implemented:

- Indentation is based on spaces only. Tabs are replaced with spaces. One indentation level consists of 2 spaces

# Interesting cases

[...]

PR #794, opened 2020-08-13
Migrate to .NET Core v2

keyboardDrummer:
This PR is a modified version of [this similar one](https://github.com/dafny-lang/dafny/pull/778). The addition of this PR is that a forked version of Boogie is used as a submodule, with a change that makes Boogie easier to build so no manual steps are required to get Dafny to build after doing a `git clone --recurse-submodules`.

### Changes
- Convert all .csproj files from .NET framework to .NET Core.
- Include Boogie as a git submodule. The referenced Boogie is a modified version of v2.4.21 that does not reference the GitVersionTask package. Without this package it's much easier to build Boogie.

### Implementation details
- Remove reference to `Mono.Cecil` since it is no longer needed.
- Make some method calls in Dafny more explicit to resolve ambiguous references caused by changes in the standard library.
- Remove `Source/Dafny/cce.cs` in favor of [this Boogie version](https://github.com/boogie-org/boogie/blob/master/Source/CodeContractsExtender/cce.cs) which seems slightly better. This solves ambiguous references that occurred in the .NET Core build. It's not clear why didn't occur before.
- The `ModelViewer` package from Boogie, which is used by `DafnyServer` has not been ported to .NET Core, and I assume it can not be because it uses .NET framework libraries. It's unclear whether the `ModelViewer` package has a future in Boogie, since they want to get rid of their .NET framework support altogether. The files used from `ModelViewer` have been copied into `DafnyServer` to resolve the issue.
- Rewrite the `CompileTargetProgram` method in `Compiler-CSharp.cs` so it uses Roslyn's `Microsoft.CodeAnalysis.CSharp` to Compile C#, because the existing method of compiling with `System.CodeDom` throws a platform not supported exception for Ubuntu in .NET Core. See [this thread](https://github.com/dotnet/runtime/issues/18768) for more information.
- Update the lit configuration so it uses `dotnet run` to get a platform independent way to run, instead of using Mono to run `Dafny.exe`.

#### No more .NET Framework support
This PR adds support for the .NET Core framework but removes it for the .NET framework. Supporting both platforms at the same time would be difficult, since this PR contains source level changes. To support both platforms at the same time we'd have to duplicate some files and customize each platform to use the right file, but in that case I would suggest considering a fork for .NET Framework support.

### Testing
- Builds and runs on Windows using Visual studio or the dotnet CLI tool
- Builds and runs on MacOS using JetBrains Rider or the dotnet CLI tool

PR #2568, opened 2022-08-09
chore: Factor out projects without runtimes

robin-aws:
Factors out two new projects that don't depend on the `DafnyRuntime` project or `DafnyRuntime.*` runtime artifacts, and removes the DLL-level dependencies on `DafnyRuntime.dll`. It also has the happy side-effect of making directory, csproj, and assembly names consistent.

The four main changes are:

* Rename `Dafny` folder/`DafnyPipeline` project to `DafnyCore`, remove reference to `DafnyRuntime` and `DafnyRuntime.*` embedded resources
* Create new `DafnyPipeline` folder/project, referencing `DafnyCore` and adding `DafnyRuntime.*` embedded resources
  * Note `DafnyPipeline.dll` does NOT depend on `DafnyRuntime.dll`, however! It has a project-level dependency to ensure `DafnyRuntime` is built first, but only embeds the runtime artifacts.
* Relabel `DafnyDriver` as `DafnyDriver.dll` instead of `Dafny.dll`, remove reference to `DafnyRuntime`
  * I'll be able to use this entry point to build Dafny source code into the runtimes when implementing #2390
* Create new `Dafny` folder/project, referencing `DafnyDriver` and `DafnyPipeline` (and hence pulling in embedded runtimes)

Note that we are still only publishing `Dafny.dll` and `DafnyPipeline.dll` as Nuget packages, and they still transitively contain the same content. We could publish `DafnyDriver` and/or `DafnyCore` in the future as well.

Removing `DafnyRuntime` as a DLL dependency from all of these projects initially caused the `Test/dafny0/DafnyLibClient.dfy` test to break. This is because the driver was attempting to `Assembly.Load` each DLL passed to it, in order to read any `DafnySourceAttribute` value out of it if present. This has never been a good idea, since an arbitrary DLL will have dependencies not satisfied by the invoking context, and not finding `DafnyRuntime.dll` was just one potential error this could cause. I had previously used [Cecil](https://www.mono-project.com/docs/tools+libraries/libraries/Mono.Cecil/) to read DLL files without loading them, but this was lost when we moved to .NET Core v2 since that library was not supported. This time I've used the built-in `System.Reflection.Metadata.MetadataReader` and related types to decode enough of the DLL to read this attribute.


Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -2217,14 +2217,12 @@ private class GenericArrayElementLvalue : ILvalue {
...
       var stream = assembly.GetManifestResourceStream(jarName);
-      if (stream is not null) {
-        var fullJarName = $"{targetDirectory}/{jarName}";
-        FileStream outStream = new FileStream(fullJarName, FileMode.Create, FileAccess.Write);
-        stream.CopyTo(outStream);
-        outStream.Close();
-      }
+      var fullJarName = $"{targetDirectory}/{jarName}";
+      FileStream outStream = new FileStream(fullJarName, FileMode.Create, FileAccess.Write);
+      stream!.CopyTo(outStream);
```
atomb, line 2224: This change has the side effect of creating an empty `JavaRuntime.jar` if it's not found in the assembly, as opposed to doing nothing, which is perhaps better behavior? It might be even nicer to emit an error message in that case, but I didn't do that in the original code, either. :)

Source/DafnyDriver/DafnyDriver.csproj
```diff
@@ -1,18 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-      <OutputType>Exe</OutputType>
-      <AssemblyName>Dafny</AssemblyName>
+      <OutputType>Library</OutputType>
```
atomb, line 4: One nice impact of this is that a client can now get _all_ of Dafny as a library (though only conveniently if we publish this library to NuGet).

Source/IntegrationTests/IntegrationTests.csproj
```diff
@@ -21,8 +21,7 @@
     </ItemGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\DafnyDriver\DafnyDriver.csproj" />
-        <ProjectReference Include="..\DafnyRuntime\DafnyRuntime.csproj" />
```
atomb, line 25: Based on the test output, it seems that this project still depends dynamically on `DafnyRuntime`.

atomb (changes requested):
This is a very nice improvement. :) I noticed a few tiny things that could be fixed (which should fix at least one of the test failures) but it otherwise seems great.

Source/DafnyDriver/DafnyDriver.csproj
robin-aws: Yes, and I think I'm going to have to define an additional "DafnyWithoutRuntimes" Exe project that depends on this in order to run Dafny on my `DafnyRuntime.dfy` code in the next PR. My intent remains to keep this and that project internal-only for now though.

Source/DafnyCore/Compilers/Compiler-java.cs
robin-aws: Ah my intention was just to cleanly make this an exception rather than a silent failure, but I can throw right away without that messy side effect.

Note that I'm going to make it an exception rather than an error message as `SinglePassCompiler.ReadRuntimeSystem` does (and I intend to make that throw as well in the next PR). At least right now, the publicly released tools should always have the runtimes attached, so I'd rather crash loudly if we've made a publishing mistake.

robin-aws: In fact I'm touching `SinglePassCompiler.ReadRuntimeSystem` already anyway, I'm going to make it throw in this PR after all - no time like the present!

Source/IntegrationTests/IntegrationTests.csproj
robin-aws: Specific tests are depending on linking to `DafnyRuntime.dll`, yup. Unfortunately one of them reveals that `DafnyMain` calls `Assembly.LoadFile` on any `dll` file passed to the CLI, in order to read the contents of any `DafnySourceAttribute`: https://github.com/dafny-lang/dafny/blob/master/Source/Dafny/DafnyMain.cs#L67. This will fail if its dependencies are not available (`DafnyRuntime.dll` or any other dependency!).

I originally fixed this by loading the assembly reflectively using Cecil, but unfortunately we lost that when switching to .NET Core: https://github.com/dafny-lang/dafny/pull/794/files#diff-0ba2e4842e187b0efcf6a6b5fd6fca73b546ece27e54074265eab734f826240dL45-R47. Even better, there's an `Assembly.ReflectionOnlyLoad` method, but it turns out it always did a regular load instead, and now always throws an exception instead (which is definitely better, but still feels like a bait-and-switch :P )

robin-aws: Eventually figured out the supported replacement for `Assembly.ReflectionOnlyLoad` - see updated PR description and release note.

Source/DafnyCore/DafnyCore.csproj
```diff
@@ -30,36 +29,8 @@
       <PackageReference Include="Boogie.ExecutionEngine" Version="2.15.7" />
   </ItemGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="..\DafnyRuntime\DafnyRuntime.csproj" />
```
MikaelMayer, line 34: I'm fine removing the reference to the DafnyRuntime, however, in [one of my PRs](https://github.com/dafny-lang/dafny/pull/2399), I am making use of the DafnyRuntime.cs which should then become a dependency of DafnyCore in your setting. Can we just add an external dependency to this file so that it's included at compile time?
That was we don't need a dependency of all the possible runtimes as you suggest but I have still it working for us.

robin-aws: Your PR will actually work just fine after this change. Since you're not passing `/useRuntimeLib` when compiling your Dafny code through `dafny.msbuild`, the compiled output will have the embedded `DafnyRuntime.cs` file prepended, and you won't need to depend on `DafnyRuntime.dll`. 

That means that the version of the runtime you build your part of `DafnyCore` with doesn't affect the version that a `dafny` end user will end up depending on in their compiled code. It's admittedly a little weird and confusing to have the C# runtime code embedded like that though. :)

MikaelMayer: Oh I see, it's in the generated code. That's fine I think ^^

Source/Dafny/Dafny.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
keyboardDrummer, line 1: Could we merge this project with DafnyDriver?

Source/DafnyDriver/DafnyDriver.csproj
```diff
@@ -1,18 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
```
keyboardDrummer, line 1: From the PR description:

> Relabel DafnyDriver as DafnyDriver.dll instead of Dafny.dll, remove reference to DafnyRuntime
I'll be able to use this entry point to build Dafny source code into the runtimes when implementing https://github.com/dafny-lang/dafny/issues/2390

Why do you need this project? Can't you use DafnyCore instead? Are you planning on setting the output type here to Exe again so you can run it?

robin-aws: I'm planning on adding an empty Exe project that references `DafnyDriver`, something like `DafnyWithoutRuntimes`, so I can easily invoke the Dafny compiler from something like a Makefile. I don't know of any way to invoke logic in `DafnyCore` directly, but I'd be happier if there was!

Source/Dafny/Dafny.csproj
```diff
@@ -0,0 +1,36 @@
...
+  </PropertyGroup>
+
+  <!-- Working around some stange behavior in dotnet publish: https://github.com/dotnet/sdk/issues/10566 -->
+  <PropertyGroup Condition="$(RUNTIME_IDENTIFIER) != ''">
+    <RuntimeIdentifier>$(RUNTIME_IDENTIFIER)</RuntimeIdentifier>
+  </PropertyGroup>
+    
+  <ItemGroup>
+    <ProjectReference Include="..\DafnyDriver\DafnyDriver.csproj" />
+    <ProjectReference Include="..\DafnyPipeline\DafnyPipeline.csproj" />
```
atomb, line 30: Does this need `DafnyPipeline`?

Source/DafnyCore/DafnyMain.cs
```diff
@@ -64,27 +66,72 @@ public class DafnyFile {
...
         SourceFileName = Path.GetTempFileName();
         File.WriteAllText(SourceFileName, sourceText);
 
       } else {
         throw new IllegalDafnyFile();
       }
+    }
+
+    private static string GetDafnySourceAttributeText(string dllPath) {
+      using var dllFs = new FileStream(dllPath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
```
atomb, line 81: Does this result in a reasonable error if `dllPath` is a non-existent file? Or does the earlier code in `DafnyFile` ensure that it exists?

atomb (approved):
This looks good! I had a couple of questions that could lead to changes, but I'm approving it to avoid delay in the (likely) case where they don't require any.

Source/Dafny/Dafny.csproj
robin-aws: Absolutely, since that's how the runtime files are included. I understand why you ask, because the minimal contents of `Dafny` should just be `DafnyDriver + runtimes`, but `DafnyPipeline` is exactly the assembly that embeds the runtimes. Most importantly the compilers load them from the `DafnyPipeline.dll` assembly explicitly rather than whatever the currently executing assembly is.

Source/DafnyCore/DafnyMain.cs
robin-aws: It is no worse than the old version that tried to `Assembly.LoadFile` it: you still get a `System.IO.FileNotFoundException`, just with a slightly different stack trace.

It's worth making that cleaner since it's a user error that shouldn't lead to a crash, but it may be non-trivial to fix. Throwing `IllegalDafnyFile` will probably lead to a different less helpful error so it may need some refactoring. I cut #2719 as a follow up.

Source/Dafny/Dafny.csproj
atomb: Right, of course!

PR #2572, opened 2022-08-10
fix: Actually create the target-runtime-version matrix dimension

robin-aws:
I noticed that the changes in #2297 didn't actually take effect, because we didn't actually define the new matrix dimension. *Mea culpa* for assuming the CI run on the PR just didn't use the new version. :)


robin-aws:
Still not working yet: we're getting the new dimension of the matrix but neither "Set up newest/oldest JDK" is triggering.

fabiomadge:
What's the status on this? As of now, this is still broken.

robin-aws:
This PR is now correctly running the tests with two different versions of Java. Before merging I was only hoping to trigger a failure by intentionally depending on something that's not available on Java 8 to verify it's actually testing what we want (I worry that the runners come with a default Java installation that might still be picked up unintentionally).

fabiomadge:
I don't expect the tests to pass because of the incompatibility with Java 18 on certain OSs. (#3344)

PR #2578, opened 2022-08-11
Fix document/verifySymbol handler so it works when symbols have multiple tasks

keyboardDrummer:
### Changes

Fix document/verifySymbol handler so it works when symbols have multiple tasks

### Testing

Added test

### DIFF
--- a/Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs
+++ b/Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs
@@ -16,6 +16,24 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization;
 public class VerificationStatusTest : ClientBasedLanguageServerTest {
 
   [TestMethod]
+  public async Task ManuallyRunMethodWithTwoUnderlyingTasks() {
+    var source = @"
+method Foo() returns (x: int) ensures x / 2 == 1; {
+  return 2;
+}";
+    await SetUp(new Dictionary<string, string> {
+      { $"{DocumentOptions.Section}:{nameof(DocumentOptions.Verify)}", nameof(AutoVerification.Never) }
+    });
+    var documentItem = CreateTestDocument(source);
+    await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+    var fooSymbol = (await RequestDocumentSymbol(documentItem)).Single();
+    var fooPosition = fooSymbol.SelectionRange.Start;
+    var runSuccess = await client.RunSymbolVerification(documentItem, fooPosition, CancellationToken);
+    Assert.IsTrue(runSuccess);
+    await WaitForStatus(fooSymbol.SelectionRange, PublishedVerificationStatus.Correct, CancellationToken);
+  }
+
+  [TestMethod]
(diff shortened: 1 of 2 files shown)


Source/DafnyLanguageServer/Handlers/Custom/VerificationHandler.cs
```diff
@@ -37,8 +37,12 @@ public class VerificationHandler : IJsonRpcRequestHandler<VerificationParams, bo
 
     var translatedDocument = await documentManager.CompilationManager.TranslatedDocument;
     var requestPosition = request.Position;
-    return GetTasksAtPosition(translatedDocument, requestPosition).
-      Any(taskToRun => documentManager.CompilationManager.Verify(translatedDocument, taskToRun));
+    var someTasksAreRunning = false;
+    var tasksAtPosition = GetTasksAtPosition(translatedDocument, requestPosition);
+    foreach (var taskToRun in tasksAtPosition) {
+      someTasksAreRunning |= documentManager.CompilationManager.Verify(translatedDocument, taskToRun);
```
MikaelMayer, line 43: Interesting use of bitwise or to ensure the right-hand side is always evaluated!

Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs
```diff
@@ -15,6 +15,24 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization;
...
+    await SetUp(new Dictionary<string, string> {
+      { $"{DocumentOptions.Section}:{nameof(DocumentOptions.Verify)}", nameof(AutoVerification.Never) }
+    });
+    var documentItem = CreateTestDocument(source);
+    await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+    var fooSymbol = (await RequestDocumentSymbol(documentItem)).Single();
+    var fooPosition = fooSymbol.SelectionRange.Start;
+    var runSuccess = await client.RunSymbolVerification(documentItem, fooPosition, CancellationToken);
+    Assert.IsTrue(runSuccess);
+    await WaitForStatus(fooSymbol.SelectionRange, PublishedVerificationStatus.Correct, CancellationToken);
```
MikaelMayer, line 33: The two tasks are the well-formedness and the correctness, right?
How does this prove that you are running the second task? What would have it generated as an output before?

MikaelMayer (commented):
Just a question to clarify my understanding.

Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs
keyboardDrummer: Before it never published the 'Correct' notification, because it only does that once _all_ underlying tasks have completed.

Source/DafnyLanguageServer/Handlers/Custom/VerificationHandler.cs
keyboardDrummer: 😅 I didn't even realise there's another risk of lazy evaluation here, but at least the test confirms this works.

MikaelMayer (approved):

PR #2580, opened 2022-08-11
Implemented a warning to prevent varaibles and constructors from having the same name

tobiaj10:


Source/Dafny/Resolver.cs
```diff
@@ -12800,6 +12800,10 @@ private class ClonerKeepLocalVariablesIfTypeNotSet : Cloner {
         // Check if the head of the pattern is a constructor or a variable
         if (ctors.TryGetValue(idpat.Id, out ctor)) {
           /* =[3]= */
+          if (ctor != null && idpat.Arguments == null && ctor.Formals.Count != 0) {
+            this.reporter.Warning(MessageSource.Resolver, pat.Tok,
+              $"Variable name '{pat}' has the same name as an exisiting constructor");
```
keyboardDrummer, line 12805: `exisiting` has a typo, but let's say `a datatype constructor that's in scope at this location`

Test/linters/constructorVariableName.dfy
```diff
@@ -0,0 +1,42 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+datatype Color = Red | GreyScale(x: nat) | Black()
```
keyboardDrummer, line 4: Your tests should also show that the warning does not trigger if the same name constructor is not in scope, for example if it's in a different module that is not imported.

Source/Dafny/Resolver.cs
```diff
@@ -12800,6 +12800,10 @@ private class ClonerKeepLocalVariablesIfTypeNotSet : Cloner {
         // Check if the head of the pattern is a constructor or a variable
         if (ctors.TryGetValue(idpat.Id, out ctor)) {
           /* =[3]= */
+          if (ctor != null && idpat.Arguments == null && ctor.Formals.Count != 0) {
```
keyboardDrummer, line 12803: The condition of this `if` and the next one are quite similar. Consider combining the two:

```c#
if (ctor != null && idpat.Arguments == null) {
  if (ctor.Formals.Count == 0) {
    // nullary constructor without () -- so convert it to a constructor
    idpat.MakeAConstructor();
  } 
  else {
    this.reporter.Warning(MessageSource.Resolver, pat.Tok,
    $"Variable name '{pat}' has the same name as a datatype constructor that's in scope at this location");
  }
}


```

keyboardDrummer (approved):

keyboardDrummer:
Looks great! I left one refactoring comment.

PR #2594, opened 2022-08-16
feat: Dafny Main method to accept optional seq<string> argument

### LINKED ISSUE #1116: Dafny 4.0 suggestion: Compiled Dafny programs should accept command-line arguments, return an exit code
Compiled Dafny programs should accept command-line arguments. To do so, the allowed Main methods should be permitted to take a string[] argument.

MikaelMayer:
Fixes #1116

Ability to write in a file `program.dfy` (it works for all previous cases of main methods)

```dafny
def Main(args: seq<string>) {
  if |args| == 3 {
    print args[0], " says ", args[1], " and ", args[2];
  } else {
    print "I need two arguments";
  }
}
```

and then run it, either with

```
> dafny /compile:3 program.dfy -- hello world
dotnet says hello and world
> dafny /compile:3 /compileTarget:java program.dfy -- hello world
java says hello and world
> dafny /compile:3 /compileTarget:js program.dfy -- hello world
<full path to the javascript file> says hello and world
```

or by compiling it first
```
> dafny /compile:1 program.dfy
> dotnet run program.dll -- hello world
dotnet says hello and world
```

Design decisions:
- The argument's array is truly the arguments passed in the command line, and includes the executable if there is one on position 0. One problem is that in Java and dotnet, there is no executable name passed in arguments, so it's for consistency. with the other languages.
- Support for one `seq<string>` because in C++, Dafny arrays don't have a length, so we need to handle sequences, otherwise no one would be able to handle arguments in C++
- Dafny's CLI option to input arguments is `-- arg1 arg2` in the command line to pass two extra arguments after the option `/compile:4`
- Internally, the option is named `DafnyOptions.MainArgs` and is a `List<string>`, and it has an automatic conversion to string-like arguments for processes using `DanfyOptions.ArgsStringExtra`
- Support for C#, Java, JavaScript, Go, Python and C++, both with /compile:3 and running the compiled version made with `/compile:2`
- Added tests for the error message when the type of the argument is not supported
- Added documentation in the options and in the reference manual.


Source/Dafny/DafnyOptions.cs
```diff
@@ -309,6 +310,15 @@ public enum IncludesModes {
             return true;
           }
 
+        case "mainArgs":
+        case "MainArgs": {
```
davidcok, line 314: Is there a reason to include this uppercase version? I think the convention for Dafny option is lowerCamelCase.

Source/Dafny/Compilers/SinglePassCompiler.cs
```diff
@@ -1670,8 +1671,19 @@ protected class NullClassWriter : IClassWriter {
         }
       }
       if (!m.Ins.TrueForAll(f => f.IsGhost)) {
-        reason = "the method has non-ghost parameters";
-        return false;
+        var nonGhostFormals = m.Ins.Where(f => !f.IsGhost).ToList();
```
davidcok, line 1674: Just a curiosity question: is this change related to the CLI arguments or is it a fix that is piggy-backing along?

MikaelMayer: Related to the CLI  arguments? What do you mean?

Source/Dafny/DafnyOptions.cs
MikaelMayer: I saw in some places support for lowerCamelCase and UpperCamelCase, so I'd think I'd mimick it. No strong reason to do it that way though.

cpitclaudel:
- Should we use a `seq` instead of an `array`?
- Can we have `/arg:arg1` `/arg:arg2` instead of `/mainArgs:arg1 arg2`?  This way there's no separate quoting?  Or even `/mainArgs arg1 arg2 arg3 …`?

Source/Dafny/Compilers/SinglePassCompiler.cs
davidcok: This PR is about support for CLI arguments for main in Dafny programs. I don't understand the relationship of this change to that goal.

MikaelMayer: Oh ok. So yes this change relates to the CLI arguments.
This change in the context of resolving the "main" method which has some specificities. Previously, a main method could accept only ghost parameters, an no non-ghost parameters.
Now, a main method should be able to receive at most one non-ghost parameter, which should be an `array<string>`. All this code is there to ensure it.

davidcok:
Usually a repeated CLI option overrules an earlier one. To me that argues (a bit) against /arg
I'm still OK with /mainArgs (or /args) with a quoted list if necessary.

I see the main purpose of this feature being to support creating executables from Dafny that when run standalone can take arguments (like "real" programs). The use case of having arguments when run from "dafny run ..." should be supported, but is less important to me and I'm willing to keep it simple with the quoted argument list.

MikaelMayer:
> Usually a repeated CLI option overrules an earlier one. To me that argues (a bit) against /arg

We already broke this "usually" with plugins. It's possible to specify multiple plugins via repeated `/plugin:` option.

cpitclaudel:
> The second one is hardly possible, because we can't predict the number of arguments.

We could take everything after `/mainArgs`

MikaelMayer:
> We could take everything after `/mainArgs`, same as what happens when you call `python script.py args…` or `dotnet run -- …`

Nice suggestion. I decided to use the `--` separator so that we copy what dotnet run is doing. I prefer it over an argument name.

cpitclaudel:
> I decided to use the -- separator so that we copy what dotnet run is doing. I prefer it over an argument name.

This is a bit dangerous though, since `--` can also be used to indicate the end of arguments, and we already have a potentially unbounded list of files (`.dfy` files).  But maybe it's OK, since files can always be disambiguated by using `./--xyz.dfy` instead of `xyz.dfy`.  Still, it's worth checking in with @keyboardDrummer to make sure that his CLI refactoring will work with that.

keyboardDrummer:
Accepting the arguments after `--` will work well with the CLI change PR that I'm working on.

MikaelMayer:
To answer @cpitclaudel  and @keyboardDrummer 's concerns

'--' is not used by dotnet at all. It's one of the arguments of the `string[]` that our application receives.
There is indeed a POSIX convention for this, but it is used only to distinguish hypothetical arguments from options.

Dafny already already knows what is an "argument" (from POSIX convention) and what is an option. there is no ambiguity possible. To my knowledge, "POSIX arguments" for Dafny are file names and have an extension and this is the thing that can be unbounded, hence I don't think they can be mistaken for options.  Dafny does not need to use "--" to separate options from (possibly unbounded) file names

So, as long as no file is named "--" there won't be any problem. We can always write, at the end, two unbounded lists of arguments using "--"

```
dotnet dafny.dll --option1:x file1.dfy file2.dfy --option3:y file3.dfy ..... -- arg1 arg2 arg3 ....
```

and there is no ambiguity.

Any other concerns?

cpitclaudel:
>  there is no ambiguity possible. […]  So, as long as no file is named "--" there won't be any problem.

What happens if I have a file called `-print:input.dfy`?  In usual GNU conventions I would use `Dafny -- -print:input.dfy` to process that file without it being mistaken for a command line flag.

Of course I could use `Dafny ./-print:input.dfy`, too, which is why I wrote above that maybe this isn't an issue.

MikaelMayer:
> Of course I could use `Dafny ./-print:input.dfy`, too, which is why I wrote above that maybe this isn't an issue.

Ok, you got me on this one. I thought that ":" was never allowed in a file name but apparently that's just a Windows rule.

Yes, I think the workaround you suggest suffices for our case so that we don't need to complexify the command-line interface for such rare (pathological) cases.

cpitclaudel:
We discussed and concluded that it would be OK to split on `--args`.  This would be a special delimiter after which nothing counts as Dafny argument, so we would preprocess the command line.

Should it be `--args` or `--run-args`/`--exec-args`/… ?

RELEASE_NOTES.md
```diff
@@ -1,9 +1,9 @@
 # Upcoming
 
+- feat: Command-line arguments available from Main() in Dafny programs (https://github.com/dafny-lang/dafny/pull/2594)
```
cpitclaudel, line 3:
```suggestion
- feat: Command-line arguments are now available from `Main` in Dafny programs, using `Main(args: seq<string>)` (https://github.com/dafny-lang/dafny/pull/2594)
```

Source/Dafny/Compilers/Compiler-Csharp.cs
```diff
@@ -222,11 +222,11 @@ public class CsharpCompiler : SinglePassCompiler {
       }
     }
 
-    protected override ConcreteSyntaxTree CreateStaticMain(IClassWriter cw) {
+    protected override ConcreteSyntaxTree CreateStaticMain(IClassWriter cw, ref string argsParameterName) {
```
cpitclaudel, line 225: Why `ref`?

Source/Dafny/Compilers/Compiler-go.cs
```diff
@@ -196,7 +202,11 @@ private struct Import {
       importWriter.WriteLine("{0} \"{1}\"", id, path);
 
       if (!import.SuppressDummy) {
-        importDummyWriter.WriteLine("var _ {0}.{1}", id, DummyTypeName);
+        if (id == "os") {
+          importDummyWriter.WriteLine("var _ = os.Args");
+        } else {
+          importDummyWriter.WriteLine("var _ {0}.{1}", id, DummyTypeName);
+        }
```
cpitclaudel, line 209: This looks odd — what happens if someone has a Dafny module called `os`?

Source/Dafny/Compilers/Compiler-go.cs
```diff
@@ -3603,7 +3613,7 @@ private class SpecialNativeType : UserDefinedType {
         verb = string.Format("build -o \"{0}\"", output);
       }
 
-      var args = string.Format("{0} \"{1}\"", verb, targetFilename);
+      var args = string.Format("{0} \"{1}\"", verb, targetFilename) + DafnyOptions.O.ArgsStringExtra;
```
cpitclaudel, line 3616: This isn't safe; arguments shouldn't be concatenated as strings

Source/Dafny/Compilers/Compiler-java.cs
```diff
@@ -342,7 +342,13 @@ private enum JavaNativeType { Byte, Short, Int, Long }
...
       companion = modName + companion;
       Coverage.EmitSetup(wBody);
-      wBody.WriteLine($"{DafnyHelpersClass}.withHaltHandling({companion}::__Main);");
+      wBody.WriteLine(@"@SuppressWarnings(""unchecked"")");
+      wBody.WriteLine($"dafny.TypeDescriptor<dafny.DafnySequence<? extends Character>> type = dafny.DafnySequence.<Character>_typeDescriptor(dafny.TypeDescriptor.CHAR);");
+      wBody.WriteLine($"dafny.Array<dafny.DafnySequence<? extends Character>> dafnyArgs = dafny.Array.newArray(type, args.length + 1);");
+      wBody.WriteLine($"dafnyArgs.set(0, dafny.DafnySequence.asString(\"java\"));");
+      wBody.WriteLine($"for (int i = 0; i < args.length; i++) dafnyArgs.set(i + 1, dafny.DafnySequence.asString(args[i]));");
+      wBody.WriteLine($"dafny.DafnySequence<? extends dafny.DafnySequence<? extends Character>> result = dafny.DafnySequence.fromArray(type, dafnyArgs);");
+      wBody.WriteLine($"{DafnyHelpersClass}.withHaltHandling(() -> {{ {companion}.__Main(result); }} );");
```
cpitclaudel, line 351: Please move this to DafnyRuntime

Source/Dafny/Compilers/Compiler-java.cs
```diff
@@ -3995,9 +4001,9 @@ private class GenericArrayElementLvalue : ILvalue {
     protected override bool IssueCreateStaticMain(Method m) {
       return true;
     }
-    protected override ConcreteSyntaxTree CreateStaticMain(IClassWriter cw) {
+    protected override ConcreteSyntaxTree CreateStaticMain(IClassWriter cw, ref string argsParameterName) {
```
cpitclaudel, line 4004: Why ref?

cpitclaudel (changes requested):
Syntax looks good to me, but I would much prefer to add a function each runtime, rather than having ad-hoc code in each compiler

Source/Dafny/Compilers/Compiler-Csharp.cs
MikaelMayer: Removed. I used it few commits ago, but now it's useless.

Source/Dafny/Compilers/Compiler-go.cs
MikaelMayer: Then there will be a compilation failure. What would be the alternative?

Source/Dafny/Compilers/Compiler-go.cs
MikaelMayer: That's unfortunately the only way to launch a `ProcessInfo`:

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.-ctor?view=net-6.0#system-diagnostics-processstartinfo-ctor(system-string-system-string)

cpitclaudel: Thankfully that's not right :) Here's an example of how to do it: https://github.com/dafny-lang/dafny/blob/5058d5a7dab45c1743ce8a254619ad030b26b0d5/Source/AutoExtern.Test/Tests.cs#L16-L31

Source/Dafny/Compilers/Compiler-go.cs
cpitclaudel: A clean error produced by Dafny; but what is bothering me is the special case on the string `os`.  We could store tuples in the list of modules (maybe `("os", "Args")` and `(id, DummyTypeName)`)

Source/Dafny/Compilers/Compiler-go.cs
MikaelMayer: Oh that's very interesting. I changed the existing behavior to also adapt to this.

Source/Dafny/Compilers/Compiler-go.cs
MikaelMayer: What do you mean by "we could store tuples in the list of modules"?
One could make this error go away by suppressing these dummy calls.
But overall, I think that if this is a problem down the road, the Go compiler will have to compile a Dafny module named "os" to something else, since "os" is the base package in Go anyway.

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2467,7 +2467,7 @@ protected class ClassWriter : IClassWriter {
         }
         nodeProcess.StandardInput.Write(targetProgramText);
         if (callToMain != null && DafnyOptions.O.RunAfterCompile) {
-          nodeProcess.StandardInput.WriteLine("require('process').argv = [\"\", " + string.Join(",", DafnyOptions.O.MainArgs.Select(ToStringLiteral)) + "];");
+          nodeProcess.StandardInput.WriteLine("require('process').argv = [\"node\", " + string.Join(",", DafnyOptions.O.MainArgs.Select(ToStringLiteral)) + "];");
```
cpitclaudel, line 2470: Don't we need Javascript quoting here (instead of Dafny quoting?)

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -1697,13 +1697,17 @@ protected class ClassWriter : IClassWriter {
...
         CreateNoWindow = true,
         UseShellExecute = false,
         RedirectStandardInput = true,
         RedirectStandardOutput = false,
         RedirectStandardError = false,
       };
+      psi.ArgumentList.Add(targetFilename);
+      foreach (var arg in DafnyOptions.O.MainArgs) {
+        psi.ArgumentList.Add(arg);
+      }
```
cpitclaudel, line 1710: Nice :)

Source/Dafny/Compilers/Compiler-js.cs
cpitclaudel: And more importantly, can't we pass these arguments on node's command line directly?

cpitclaudel (approved):

Source/Dafny/Compilers/Compiler-js.cs
MikaelMayer: This is JavaScript quoting (ToStringLiteral is a compiler to JavaScript function)
More importantly, the problem is that passing anything in command line (e.g. node arg1 arg2) will attempt to load the file, so this is also why the javascript compiler's run method is pushing everything in the standard input.

cpitclaudel (approved):

PR #2596, opened 2022-08-16
Remove the use of Boogie's XML output

### LINKED ISSUE #1950: Remove the use of Boogie's XML output
Dafny currently implements `/verificationLogger:trx` and `/verificationLogger:csv` by reading the XML output generated by Boogie. The new [`/verificationLogger:text`](https://github.com/dafny-lang/dafny/pull/1947) implementation uses a more direct internal data structure, instead. The two previous loggers could be adapted to use this same data structure, simplifying them and removing the awkward incompatibility between `/verificationLogger` and `/xml`.

atomb:
Now that verification results are aggregated into a data structure returned to Dafny, using Boogie's XML output is no longer necessary.

Fixes #1950

This is a refactoring that should still pass existing tests.

### DIFF
--- a/Source/DafnyDriver/VerificationResultConvertor.cs
+++ b/Source/DafnyDriver/VerificationResultConvertor.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Dafny {
   ///  * CSV files, which are easier to parse and summarize. 
   ///  * human-readable text output.
   /// </summary>
-  public static class BoogieXmlConvertor {
+  public static class VerificationResultConvertor {
 
     public static TestProperty ResourceCountProperty = TestProperty.Register("TestResult.ResourceCount", "TestResult.ResourceCount", typeof(int), typeof(TestResult));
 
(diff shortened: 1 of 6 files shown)


Source/DafnyDriver/VerificationResultConvertor.cs
```diff
@@ -0,0 +1,126 @@
...
+
+  /// <summary>
+  /// Utility to translate verification results into test logger events,
+  /// allowing us to deliver them through common loggers on the .NET
+  /// platform. For now we support three formats:
+  ///  * TRX files, which can be understood and visualized by various .NET tools.
+  ///  * CSV files, which are easier to parse and summarize. 
+  ///  * human-readable text output.
+  /// </summary>
+  public static class VerificationResultConvertor {
```
keyboardDrummer, line 28: From the name I can see that it does something with VerificationResults. I think in this case the conversion target (TestLoggerEvents) is more specific than the source (VerificationResult), so I'd prefer to base the name on that, or put both in the name.

keyboardDrummer (approved):
Much cleaner 🤗, thanks!

Source/DafnyDriver/VerificationResultConvertor.cs
atomb: Yeah, that's a good point. One thing that makes me hesitate, though, is that one of the loggers doesn't go through the `ITestLogger` interface. Although I'm thinking that maybe `VerificationResultLogger` is a better name in either case.

keyboardDrummer (approved):

PR #2598, opened 2022-08-16
Removing obsolete 'protected' from keyword lists

davidcok:
Fixes part of #2527

Still need to remove 'protected' from Dafny.atg and ./Dafny/Compilers/Compiler-*.cs


davidcok:
Branch made off of wrong root -- switched to cok-2527b

PR #2599, opened 2022-08-16
Removing 'protected' from syntax highlighting keyword lists

### LINKED ISSUE #2527: Make sure protected is no longer a keyword in the RM or in the code

davidcok:
Partially fixes #2527
Still need to remove 'protected' from Dafny.atg and Compilers/Compiler-*.cs


docs/dafny.sty
```diff
@@ -8,9 +8,9 @@
   morekeywords={class,datatype,codatatype,newtype,type,iterator,trait,extends,
       bool,char,nat,int,real,object,set,iset,multiset,seq,string,map,imap,array,array2,array3,
       bv0,bv1,bv2,bv3,bv4,bv5,bv6,bv7,bv8,bv12,bv16,bv20,bv24,bv32,bv48,bv64,bv128,
-      function,predicate,copredicate,inductive,
-      ghost,var,static,protected,refines,
-      method,lemma,constructor,colemma,twostate,
+      for,function,predicate,copredicate,inductive,
```
MikaelMayer, line 11: Do you really need to reserve the "for" as a keyword everywhere? Reserving a keyword is a breaking change.

Same for "nameonly"

MikaelMayer (commented):
I think we need more input there. @robin-aws ?

docs/dafny.sty
davidcok: Removed those from this PR. (Note that this is just for syntax coloring -- not in the parser)

MikaelMayer (approved):

PR #2604, opened 2022-08-17
Let `DafnyRuntime` target more frameworks

### LINKED ISSUE #2481: DafnyRuntime nuget package should be compatible with older .NET frameworks
We've ended up requiring net6.0 by default, but the actual code is known to work with older frameworks: it is inlined when compiling if `/useRuntimeLib` is not used, and https://github.com/aws/aws-encryption-sdk-dafny builds this way, for example, and supports .NET/.NET Core 3.1.

Should just be a matter of fixing the DafnyRuntime.csproj file AFAICT, with no actual code changes.

atomb:
Allow the `DafnyRuntime` package to work with `netstandard2.0` and `net462`, as well as the previous `net6.0`. This should make Dafny target code that links with a pre-built `DafnyRuntime.dll` more portable.

Fixes #2481 


.github/workflows/msbuild.yml
```diff
@@ -45,7 +45,10 @@ jobs:
         ref: v${{ steps.regex-match.outputs.group1 }}
     - name: Build Dafny with local Boogie
       working-directory: dafny
-      run: dotnet build Source/Dafny.sln
+      run: |
+        dotnet build Source/Dafny.sln
+        dotnet build -f net462 Source/DafnyRuntime/DafnyRuntime.csproj
+        dotnet build -f netstandard2.0 Source/DafnyRuntime/DafnyRuntime.csproj
```
fabiomadge, line 51: Does this generate 3 runtimes?

atomb: Yeah, one for each of those frameworks, so that folks downloading from NuGet can get the one that works for them. My understanding is that this is the best way to make .NET libraries available for many versions, but I'd love to hear about a better way if there is one.

.github/workflows/msbuild.yml
```diff
@@ -45,7 +45,10 @@ jobs:
         ref: v${{ steps.regex-match.outputs.group1 }}
     - name: Build Dafny with local Boogie
       working-directory: dafny
-      run: dotnet build Source/Dafny.sln
+      run: |
+        dotnet build Source/Dafny.sln
+        dotnet build -f net462 Source/DafnyRuntime/DafnyRuntime.csproj
```
keyboardDrummer, line 50: Are these useful frameworks to target? Who would consume a netstandard2.0 or a net462 library?

atomb: This was partly informed by [the dependencies here](https://github.com/aws/aws-encryption-sdk-dafny/blob/mainline/aws-encryption-sdk-net/Source/AWSEncryptionSDK.csproj#L5) with the adjustment that the current runtime doesn't build with `net452` but does go work as far as `netstandard2.0`. I also took guidance from [here](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#net-standard-not-deprecated).

keyboardDrummer: If Standard allows sharing code with Framework and Core, then why would you build for Standard AND something else? 

From the linked docs:
> Most widely used libraries will multi-target for both .NET Standard 2.0 and .NET 5+. Supporting .NET Standard 2.0 gives you the most reach, while supporting .NET 5+ ensures you can leverage the latest platform features for customers that are already on .NET 5+.

atomb: Hmm. Re-reading that page, I think you may be right. It seems as though if we built for `netstandard2.1` we might also need a separate build to support Framework (e.g., `net462`), but if we build for `netstandard2.0` we don't need that?

keyboardDrummer: Yes, looks like it. But pff, that 2.1 vs 2.0 matters so much is confusing!

keyboardDrummer (approved):

PR #2607, opened 2022-08-17
fix: Resolve and translate StmtExpr inside 'old' correctly

### LINKED ISSUE #2597: Old expressions don't always properly nest into StmtExpr's
Dafny does not allow an `old` to be used inside another `old`. More generally, the context inside an `old` is considered a single-state context, which means several things (`old`, `fresh`, calls to `twostate` functions, ...) that are not allowed.

Alas, for the statement portion of a `StmtExpr` inside an `old`, Dafny's resolver doesn't "remember" that there was an `old` around it, and in the case of lemma calls inside such a `StmtExpr`, Dafny's verifier uses the current state, rather than the `old` state. (This is not a soundness issue, but it is confusing to a user, because the program text doesn't say what one would think it says.)

Here are repros. For resolution:

``` dafny
class C {
  var data: int

  method ResolutionIssues()
    // requires old(data) == 3 // 'old' expressions are not allowed in this context
    requires calc { 0; { var d := old(data); } 0; } true // bogus: 'old' should not be allowed here, either
  {
    //var a := old(old(data)); // an 'old' is not allowed inside another 'old
    var b := old(
      calc {
        0;
      ==  { var a := old(data); } // bogus: 'old' should not be allowed here, either
        0;
      }
      0
      );
  }
}
```

For verification:

``` dafny
class C {
  var data: int

  lemma DataIs(x: int)
    requires data == x
  {
  }

  method VerificationIssues()
    requires data == 3
    modifies this
  {
    data := 100;
    ghost var x;
    x := old(assert data == 3; 0); // this assertion passes
    x := old(DataIs(3); 0); // bogus: this assertion does not pass (but it does if you replace 3 by 100)
  }
}
```

RustanLeino:
Depends on #2606
Fixes #2597

Previously, some statements in a `StmtExpr` had ignored any enclosing `old`, and the resolution of `StmtExpr`s paid no attention to an enclosing `old`.


Source/Dafny/Resolver.cs
```diff
@@ -15397,16 +15400,23 @@ public class ResolveOpts {
       } else if (expr is StmtExpr) {
         var e = (StmtExpr)expr;
         int prevErrorCount = reporter.Count(ErrorLevel.Error);
+
+        var previousInOneStateStatementContext = inOneStateStatementContext;
+        if (!IsTwoStateContext(opts)) {
```
keyboardDrummer, line 15405: Have you considered doing something like what `ResolveOpts` is for expressions, but then for statements?

```c#
record ResolveStatementContext(bool IsOneState, ICodeContext CodeContext);
public void ResolveStatement(Statement stmt, ResolveStatementContext context); 
```

instead of having the field `inOneStateStatementContext` in `Resolver`

That way you can't forget to restore the previous `inOneStateStatementContext` state, and it might make it easier to parallelise things in the future.

Source/Dafny/Verifier/Translator.cs
```diff
@@ -6260,7 +6260,37 @@ private class WFOptions {
 
       } else if (expr is StmtExpr) {
         var e = (StmtExpr)expr;
+        // If we're inside an "old" expression, then "etran" will know how to translate
```
keyboardDrummer, line 6263: Given the size of `Translator.cs`, and that the code for checking wellformed of expression is about 1200 lines, I suggest we move all of that to `Translator.ExpressionWellformed.cs`. Would you mind doing that in a follow-up PR? I think not getting full Rider IDE support makes it significantly harder to write idiomatic code.

Source/Dafny/Verifier/Translator.cs
```diff
@@ -6260,7 +6260,37 @@ private class WFOptions {
...
+        // expressions. However, here, we're also having to translate e.S, which is a
+        // Statement. Since statement translation (in particular, CallStmt's) works
+        // directly on the global variable $Heap, we change its value here.
+        // (Note, this would be a bad translation if the Statement could introduce
+        // control flow that would miss the "$Heap := tmpHeap;" assignment that
+        // resets the value of $Heap. Luckily, StmtExpr does not allow any such
+        // Statement.)
+        Bpl.IdentifierExpr tmpHeap = null;
```
keyboardDrummer, line 6271: `CheckWellformedWithResult` is already very long, so it would be good to extract a method here. 

Consider introducing:
```c#
WithFreshHeap(string heapVarSuffix, Token token, ExpressionTranslator etrans, BoogieStmtListBuilder builder, Action<BoogieStmtListBuilder> addStatements) {
  var suffix = CurrentIdGenerator.FreshId(heapVarSuffix);
  var tmpHeapVar = new Bpl.LocalVariable(token, new Bpl.TypedIdent(token, "Heap$" + suffix, predef.HeapType));
  locals.Add(tmpHeapVar);
  var tmpHeap = new Bpl.IdentifierExpr(token, tmpHeapVar);
  var generalEtran = new ExpressionTranslator(this, predef, token);
  var theHeap = (Bpl.IdentifierExpr /*TODO: this cast is dubious*/)generalEtran.HeapExpr;

  // tmpHeap := $Heap;
  builder.Add(Bpl.Cmd.SimpleAssign(token, tmpHeap, theHeap));
  // $Heap := etran.$Heap;
  builder.Add(Bpl.Cmd.SimpleAssign(token, theHeap, etran.HeapExpr));
  
  addStatements(builder);
  
  // $Heap := tmpHeap;
  builder.Add(Bpl.Cmd.SimpleAssign(token, theHeap, tmpHeap);
}

File: Source/Dafny/Verifier/Translator.cs [hunk 4]
```diff
@@ -6260,7 +6260,37 @@ private class WFOptions {
...
+        Bpl.IdentifierExpr theHeap = null;
+        if (etran.UsesOldHeap) {
+          var suffix = CurrentIdGenerator.FreshId("StmtExpr#");
+          var tmpHeapVar = new Bpl.LocalVariable(e.S.Tok, new Bpl.TypedIdent(e.S.Tok, "Heap$" + suffix, predef.HeapType));
+          locals.Add(tmpHeapVar);
+          tmpHeap = new Bpl.IdentifierExpr(e.S.Tok, tmpHeapVar);
+          var generalEtran = new ExpressionTranslator(this, predef, e.S.Tok);
+          theHeap = (Bpl.IdentifierExpr /*TODO: this cast is dubious*/)generalEtran.HeapExpr;
```
keyboardDrummer, line 6279: Please resolve the TODO

Source/Dafny/Verifier/Translator.cs
```diff
@@ -6260,7 +6260,37 @@ private class WFOptions {
 
       } else if (expr is StmtExpr) {
         var e = (StmtExpr)expr;
+        // If we're inside an "old" expression, then "etran" will know how to translate
+        // expressions. However, here, we're also having to translate e.S, which is a
+        // Statement. Since statement translation (in particular, CallStmt's) works
```
keyboardDrummer, line 6265: Can a CallStmt occur inside a StmtExpr? If not, why is it mentioned here?

Source/Dafny/Verifier/Translator.cs
keyboardDrummer, line 6263: Can you help me understand better why we need to temporarily clear the heap here?

Source/Dafny/Resolver.cs
RustanLeino: Yes, that would be a good idea. Currently, options for statement translation uses these "global" fields and options for expression translation are passed as a parameter. However, I would suggest we do such a refactoring as a separate PR.

Source/Dafny/Verifier/Translator.cs
RustanLeino: Good idea. And I have the same problem with getting Rider's IDE to respond to this long file.

Source/Dafny/Verifier/Translator.cs
RustanLeino: Such a cast is done in many places in the translator. Each one is dubious (but, apparently, we are not seeing any examples where these casts actually fail). I could refactor all occurrences of this pattern to one place, but that one place would still have the comment. Would that be an improvement? Can you think of some better solution?

Source/Dafny/Verifier/Translator.cs
RustanLeino: Yes, a `StmtExpr` is an `Expression` preceded by one of about 5 different kinds of statements (`assert`, `assume`, `reveal`, lemma call, and `calc`). The lemma call was a problem if `old` was placed around the `StmtExpr`. Furthermore, the hints in a `calc` can contain `CallStmt`s, and these were a problem, too. When thinking through various possibly problematic statements, only `CallStmt` seemed problematical, but the fix here takes take of any statement.

Source/Dafny/Verifier/Translator.cs
RustanLeino: A Boogie procedure operates on global variables. Dafny uses just one global variable, namely `$Heap`. (Actually, there's another one, too, called `$Tick`, but I don't think it's used anymore or could be deleted.) If a Dafny call statement sits inside an `old`, we would want the precondition of the Boogie procedure to use `old($Heap)` instead of `$Heap`. A Boogie procedure has no facility to parameterize the global variables it uses. Therefore, this PR changes the global variable `$Heap` to the desired value before any such Boogie procedure call is made.

This would not be a good idea if the procedure could change the heap. In fact, if that could happen, then we shouldn't allow the Dafny code to make such a call inside `old`. But lemma don't have any net effect on `$Heap`, so we're okay. After the calls, this PR thus changes the global variable `$Heap` back to what it previously had been.

So, essentially, the temporary assignment to `$Heap` is a way to pass in a different value for `$Heap` to any procedure calls that occur in the `Statement` part of the `StmtExpr`.

Source/Dafny/Verifier/Translator.cs
RustanLeino: Will do.

Source/Dafny/Verifier/Translator.cs
keyboardDrummer: Sorry, I should have asked a more specific question. I would expect us to set the global heap variable to some value that reflects the old heap, but it seems we're just clearing the heap. Am I reading this incorrectly, or how is _clearing_ the heap sufficient?

Source/Dafny/Resolver.cs
keyboardDrummer: If you leave the current global options for Statement as they are, and only make the change I suggested, would you still consider it a refactoring? To me it seems like a choice between adding a parameter to a (recursive)method or introducing a globalish variable, and then I prefer the former.

Source/Dafny/Verifier/Translator.cs
keyboardDrummer: A TODO implies something has to be done, but if you merge it, then the moment when TODO that is not clear, and I'd say at that point it's never. If you want to allocate time to resolving this, I think creating a GH issue will at least prompt someone to consider picking it up, or deciding not to and closing the issue.

If you don't want to allocate time, but just help future developers read the code, I'd suggest changing this into a comment, `// we don't know why but this cast never fails in practice`, and then indeed it would help centralising the cast.

Source/Dafny/Resolver.cs
RustanLeino: Thanks for the suggestion! I did the change as new commits of this PR. This cleaned up a lot of stuff in how this was done before. Now, there's a `ResolutionContext` that gets passed through both statements and expressions.

RustanLeino: Btw, `ResolutionContext` is what previously was called `ResolveOpts`, which previously had been used only for expression resolution. I expanded it to work with statements, too.

Source/Dafny/Verifier/Translator.cs
RustanLeino: I did it in this PR. The code move is a separate commit (and contains only the code move).

RustanLeino: This PR generates:

```
tmpHeap := $Heap;
$Heap := etran.$Heap;
...
$Heap := tmpHeap;
```

The RHS of the first assignment to `$Heap` (which I sketched as `etran.$Heap` here and in the code) is whatever the current `etran` thinks the `.HeapExpr` is. For example, this `etran.HeapExpr` may be `old($Heap)` or `$SpecialHeapForSomeSpecialPurpose`.

So, the heap is not empty, but is what the expression translator thinks the heap's value is.

RustanLeino:
Thanks for your very good comments, as always, @keyboardDrummer . I've address them all (as commits directly in this PR).

Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
```diff
@@ -18,6 +18,18 @@ internal class ExpressionTranslator {
         // below), but also, the debugger may invoke HeapExpr and that will cause an increment as well.
         get { Statistics_HeapUses++; return _the_heap_expr; }
       }
+
+      /// <summary>
+      /// Return HeapExpr as an IdentifierExpr.
+      /// CAUTION: This getter should be used only if the caller "knows" that HeapExpr really is an IdentifierExpr.
+      /// </summary>
+      public Boogie.IdentifierExpr HeapCastToIdentifierExpr {
```
keyboardDrummer, line 26: Honestly this I would just inline since the comments do not provide more information than `(Boogie.IdentifierExpr)HeapExpr` already would.

Source/Dafny/Verifier/Translator.cs
keyboardDrummer: Even in a separate commit, unless that commit is the first or last in the list, it makes it harder to review the PR.

When trying to comment on part of the diff I also got the message "This diff is outdated, please refresh and try again."

Source/Dafny/Resolver.cs
```diff
@@ -14628,49 +14630,110 @@ class LazyString_OnTypeEquals {
...
-        this.isPostCondition = isPostCondition;
-        this.InsideOld = insideOld;
+      public ResolutionContext(ICodeContext codeContext, bool isTwoState)
+        : this(codeContext, isTwoState, false, false, false, false) {
+      }
+
+      /// <summary>
+      /// Return a ResolutionContext appropriate for the body of "codeContext".
+      /// </summary>
+      public static ResolutionContext FromTopLevel(ICodeContext codeContext) {
```
keyboardDrummer, line 14663: The inconsistency in the names, naming the method From _TopLevel_ followed by the type `ICodeContext`, and the comment "codeContext", make it seem you don't like the name `ICodeContext`. Please follow-up with a rename.

Source/Dafny/Resolver.cs
```diff
@@ -14628,49 +14630,110 @@ class LazyString_OnTypeEquals {
...
+          isTwoState = false;
+        } else if (codeContext is Function && !(codeContext is TwoStateFunction)) {
+          isTwoState = false;
+        } else {
+          isTwoState = true;
+        }
+        return new ResolutionContext(codeContext, isTwoState);
+      }
+
+      public ResolutionContext WithGhost(bool isGhost) {
```
keyboardDrummer, line 14675: Why not use a record instead of a class for `ResolutionContext`? that way you can use the `with` keyword instead of having to write these `With` methods.

Source/Dafny/Verifier/Translator.cs
keyboardDrummer: I see, sorry I think I should have just read better :)

Source/Dafny/Verifier/Translator.ExpressionWellformed.cs
```diff
@@ -0,0 +1,1295 @@
...
+      // (Note, this would be a bad translation if the Statement could introduce
+      // control flow that would miss the "$Heap := tmpHeap;" assignment that
+      // resets the value of $Heap. Luckily, StmtExpr does not allow any such
+      // Statement.)
+
+      void addStatements() {
+        TrStmt(stmtExpr.S, builder, locals, etran);
+      }
+
+      void withFreshHeap(string heapVarSuffix, IToken token) {
```
keyboardDrummer, line 1153: Maybe we could change this into:

```c#
BuildWithHeapAs(BoogieStmtListBuilder builder, 
  Bpl.Expr newHeap, 
  string heapStorageSuffix, 
  IToken token, 
  Action<BoogieStmtListBuilder> build) 
{
        var suffix = CurrentIdGenerator.FreshId(heapStorageSuffix);
        var currentHeapVar = new Bpl.LocalVariable(token, new Bpl.TypedIdent(token, "Heap$" + suffix, predef.HeapType));
        locals.Add(currentHeapVar);
        var tmpHeap = new Bpl.IdentifierExpr(token, currentHeapVar);
        var theHeap = new ExpressionTranslator(this, predef, token).HeapCastToIdentifierExpr;

        // currentHeap := $Heap;
        builder.Add(Bpl.Cmd.SimpleAssign(token, currentHeapVar, theHeap));
        // $Heap := newHeap;
        builder.Add(Bpl.Cmd.SimpleAssign(token, theHeap, newHeap));

        build(builder);

        // $Heap := currentHeap;
        builder.Add(Bpl.Cmd.SimpleAssign(token, theHeap, currentHeapVar));
}
```

And make that a top-level method.

keyboardDrummer (dismissed):
I'm approving. If you want to make additional changes, consider doing them in a follow-up PR since this one is rather bulky.

Source/Dafny/Verifier/Translator.cs
RustanLeino: I don't understand the diff being outdated. But I can move the new commits to a new PR.

Source/Dafny/Verifier/Translator.ExpressionWellformed.cs
RustanLeino: Good suggestions. Done.

Source/Dafny/Resolver.cs
RustanLeino: I did not know about the `with` keyword in C#. I'll try it.

keyboardDrummer (approved):

Issue #320, opened 2019-07-15
Dafny should warn when `old` is used on an expression that does not access the heap


mschlaipfer:
The following post-condition

`data[..] == old(data)[.. old(pos)] + [a] + old(data)[old(pos) + 1 ..]`

is translated to
```
Seq#Equal(Seq#FromArray($Heap, read($Heap, this, _24_Streams.StringWriter.data)), 
          Seq#Append(Seq#Append(Seq#Take(Seq#FromArray($Heap, read(old($Heap), this, _24_Streams.StringWriter.data)), 
                read(old($Heap), this, _24_Streams.StringWriter.pos)), 
              Seq#Build(Seq#Empty(): Seq Box, $Box(a#0))), 
            Seq#Drop(Seq#FromArray($Heap, read(old($Heap), this, _24_Streams.StringWriter.data)), 
              read(old($Heap), this, _24_Streams.StringWriter.pos) + 1)))
```

but it should be translated to 

```
Seq#Equal(Seq#FromArray($Heap, read($Heap, this, _24_Streams.StringWriter.data)), 
          Seq#Append(Seq#Append(Seq#Take(Seq#FromArray(old($Heap), read(old($Heap), this, _24_Streams.StringWriter.data)), 
                read(old($Heap), this, _24_Streams.StringWriter.pos)), 
              Seq#Build(Seq#Empty(): Seq Box, $Box(a#0))), 
            Seq#Drop(Seq#FromArray(old($Heap), read(old($Heap), this, _24_Streams.StringWriter.data)), 
              read(old($Heap), this, _24_Streams.StringWriter.pos) + 1)))
```

The difference is in the first arguments to `FromArray` on the RHS of the equals, which are missing the call to `old`.

mschlaipfer:
Created a minimal example https://gist.github.com/mschlaipfer/34f713e2866b9b16ebe0a59803502caf

`C.f` and `C.h` are the same. `C.g` is the version analog to containing `old(pos)`. The first assertion fails, whereas the second one passes.

mschlaipfer:
Tried to debug this, it seems to be related to line 14473 in Translator.cs, where the `HeapExpr` does not have the old for `g` (from the gist example).

mschlaipfer:
The currently implemented behavior is as intended, but we should produce a warning when `old` does not apply any change in the Boogie encoding, as in this case.

acioc:
We should warn on this

davidcok:
[...]

PR #2610, opened 2022-08-17
feat: Generate a warning when 'old' has no effect

### LINKED ISSUE #1989: Linter rule: `old` should only refer to heap values
Dafny does not provide a warning when using the `old` keyword on immutable values. This doesn't make sense and can lead to misunderstanding. Pointing this out in an analysis exposed via the IDE would clear this up.

For example, the following code verifies, and it's not immediately obvious why.

```dafny
module Test {

  class C {
  }

  function arrSet(a: array<C>) : set<object>
    reads a
  {
    set i : int | 0 <= i < a.Length :: a[i]
  }

  method test(a: array<C>, i : int, x : C)
    modifies a
    requires 0 <= i < a.Length
    requires x !in arrSet(a)
  {
    a[i] := x;
    assert(arrSet(a) == arrSet(old(a)));
  }
}
```

The reason that `assert(arrSet(a) == arrSet(old(a)));` verifies is that `old(a)` refers to the reference, and not to the values in `a`. This is [documented](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-old-expression), but a linter pass warning about using `old` on an immutable value would be helpful.

RustanLeino:
Depends on PRs #2606 and #2607
Fixes #1989

If the argument to `old` has no dereferences of the mutable heap, generate a warning saying that the `old` has no effect. This has been a common source of errors and confusion.


mschlaipfer:
Thanks for implementing this!

keyboardDrummer:
Could you resolve the conflicts with master?

Source/Dafny/Resolver.cs
```diff
@@ -10622,6 +10624,8 @@ public enum FrameExpressionUse { Reads, Modifies, Unchanged }
     void CreateIteratorMethodSpecs(IteratorDecl iter) {
       Contract.Requires(iter != null);
 
+      var tok = new AutoGeneratedToken(iter.tok);
```
keyboardDrummer, line 10627: Could we run the old has no effect detection after we generate the iterator method specs? This works as well, but I'm curious what the best ordering of operations is. Maybe `CreateIteratorMethodSpecs` should not be part of the main resolution phase, but instead be part of the code-gen phase? 

Can `FreeVariableUtil.ComputeFreeVariables` not be called before resolution? Maybe we should specify that in a comment on the before.

Side note: the AutoGeneratedToken concept seems strange to me. In generated code I wouldn't expect warnings/errors, and I wouldn't expect any non-nullish tokens there.

Source/Dafny/Linter.cs
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: MIT
+//
+//-----------------------------------------------------------------------------
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive;
+
+namespace Microsoft.Dafny {
+
+  class Linter : IRewriter {
```
keyboardDrummer, line 14: Please use a more specific name while this class does a specific thing.

keyboardDrummer (approved):

Source/Dafny/Resolver.cs
RustanLeino: The linter is run after resolution. At that time (and not before that time), we have information about heap usage and variable definitions. For example, we know whether `s[i]` is referring to a sequence element or an array element. The free-variable machinery already has the ability to detect heap usage, so it made sense to call it from the linter.

There are a number of desugaring tasks (`match`, datatype update, iterators, ...) that are being done quite early. Creating new AST nodes is easier pre-resolution, because then the resolution pass will fill in type information. To my taste, some of that desugaring is done too early, or perhaps shouldn't be done at all. It would be good to have a discussion and redesign of all of that sometime. But for iterators, the desugaring does need to be done early, because the iterator desugaring introduces names that the program is allowed to call. So, this means iterator desugaring is necessarily done before the linter runs.

One could consider making the iterator desugaring more clever, so that it never generates useless `old` expressions. But this is difficult to determine before resolution (cf. my `s[i]` example above).

Therefore, to avoid linter warnings from such generated code, we need some way to keep track of which expressions have been automatically generated. We already had a mechanism for this, which is to put some information into the tokens. This is not perfect, and I'm not sure that we're using nested token wrappers correctly. For linter warnings, any such imperfections seem okay, since at worst we'd generate bad warnings or omit some warnings.

keyboardDrummer: Thanks for the info. This approach seems good :)

keyboardDrummer (approved):

davidcok:
Added a sentence in the RM about this warning.

davidcok:
Does this solve #320 as well?

Issue #2487, opened 2022-07-25
Create a FAQ and common Error message explain action web page


davidcok:

davidcok:
Progress 8/8/2022: #dafny, #dafny-development and #yucca mined through 7/31/2022. A bit more than 100 candidate FAQs identified, with about 60 or so having draft answers.

davidcok:
All FAQs through August are complete.
Yet TODO:
- update FAQs for August to now
- comprehensive evaluation of Error messages

PR #2623, opened 2022-08-22
FAQ entries - part E

davidcok:
Portion of #2487


docs/HowToFAQ/FAQSplitHere.md
```diff
@@ -0,0 +1,34 @@
...
+---
+
+## Question
+
+How do `{:split_here}` and `{:focus}` work to divide up a verification condition?
+
+## Answer
+
+Analysis of a method for the purpose of verification identifies a large number of assertions that must be proved.
+These assertions are grouped into batches.
```
keyboardDrummer, line 12: I think it might be good to lead with the reason for grouping assertions into batches.

`Dafny uses an SMT solver to verify assertions. The SMT solver may become better or worse at verifying an assertion if you also ask it to also verify another assertion. Dafny allows you to split up a group of assertions into batches, where each batch is sent to the SMT solver separately, so the verification of each batch is not influenced by the others.`

docs/HowToFAQ/FAQSplitHere.md
```diff
@@ -0,0 +1,34 @@
...
+
+One default way of operating is to combine all assertions into one batch, make one verification condition out of it,
+and have the prover tackle the whole thing at once. This method allows the prover to use intermediate results to
```
keyboardDrummer, line 15: Should we call an SMT solver a prover, just because we're using it to prove things?

docs/HowToFAQ/FAQSplitHere.md
```diff
@@ -0,0 +1,34 @@
...
+## Question
+
+How do `{:split_here}` and `{:focus}` work to divide up a verification condition?
+
+## Answer
+
+Analysis of a method for the purpose of verification identifies a large number of assertions that must be proved.
+These assertions are grouped into batches.
+
+One default way of operating is to combine all assertions into one batch, make one verification condition out of it,
```
keyboardDrummer, line 14: "make one verification condition out of it" consider replacing this with "leading to a single [prover/solver] run"

docs/HowToFAQ/FAQSplitHere.md
```diff
@@ -0,0 +1,34 @@
...
+and have the prover tackle the whole thing at once. This method allows the prover to use intermediate results to 
+more efficiently prove all of the assertions. However, even when this is more efficient than the combination of proving each
+assertion by itself (with start-up costs and the like for each one), it can also take quite a while to give a final result, possibly even timing-out.
```
keyboardDrummer, line 17: "it can also take quite a while to give a final result", the situation is worse than just the total solving time becoming bigger right? The solving time of one assertion may also become bigger because of the presence of another one, because there is more "clutter" in the solver's state.

docs/HowToFAQ/FAQTriggers.md
```diff
@@ -0,0 +1,22 @@
...
+```dafny
+assert vals != [];
+assert MatchingResult() == (forall v :: v in vals ==> false);
+assert !MatchingResult();
+```
+
+## Answer
+
+The problem here is the trigger on the quantification expression.
+Dafny sees the `forall` quantifier and but uses it only when there is a `v in vals` fact in the context for some v (this is what the annotation on the forall iin the VSCode IDE means: Selected triggers: {v in vals}).  So until you give it a concrete fact to "trigger" on, it doesn't use your quantifier.
```
keyboardDrummer, line 19: "and but"

keyboardDrummer, line 19: "iin"

keyboardDrummer, line 19: "So until you give it a concrete fact to "trigger" on, it doesn't use your quantifier." shouldn't we show how to do that for this example? Linking to a paper for the final answer for more information seems fine, but as the final answer to this simple question it seems heavy handed.

docs/HowToFAQ/FAQTriggers.md
```diff
@@ -0,0 +1,22 @@
...
+
+[Pages 2--4 of this paper](https://pit-claudel.fr/clement/papers/dafny-trigger-selection-CAV16.pdf)
+give a description of how triggers work.
```
keyboardDrummer, line 22: `gives`?

docs/HowToFAQ/FAQMatchOnSet.md
```diff
@@ -0,0 +1,12 @@
...
+title: How do I pattern match against a head and tail of a sequence or against a set?
+---
+
+## Question
+
+How do I pattern match against a head and tail of a sequence or against a set?
+
+## Answer
+
+You can't. Match [expressions](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-match-expression) and [statements](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-match-statement) operate on `datatype` values and not on other Dafny types like sets, sequences, and maps. If statements, perhaps with [binding guards](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-binding-guards), may be an alternative.
```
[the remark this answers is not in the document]
davidcok: Aah. I forgot to add the change to the RM that adds the link to this particular PR. Done now.
(The 2010-... link is automatically generated, but changes if the section numbers change.)

docs/HowToFAQ/FAQRecord.md
```diff
@@ -0,0 +1,31 @@
...
+
+How does one define a record?
+
+## Answer
+
+The Dafny `datatype` feature lets you define simple or complex records, including recursive ones.
+
+A simple record might be defined as
+```dafny
+datatype ABCD = A | B | C | D
```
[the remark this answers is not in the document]
davidcok: Actually this is a good point. I'll add an example of this sort as well. And I agree that it is closer to the customary use of "record"

docs/HowToFAQ/FAQRecord.md
```diff
@@ -0,0 +1,31 @@
+---
+title: How does one define a record?
+---
+
+## Question
+
+How does one define a record?
```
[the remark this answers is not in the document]
davidcok: No -- but I added your Date example to the enumeration examples I already had.

docs/HowToFAQ/FAQSplitHere.md
```diff
@@ -0,0 +1,34 @@
+---
+title: How do `{:split_here}` and `{:focus}` work to divide up a verification condition?
+---
+
+## Question
+
+How do `{:split_here}` and `{:focus}` work to divide up a verification condition?
```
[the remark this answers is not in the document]
davidcok: Well, the descriptions are in the section on attributes, which describes each attribute one by one.
And there are lots of questions on Slack which are essentially "how does X work?" -- some before and some after an update of the RM.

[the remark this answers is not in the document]
davidcok: The term is used in the RM, at least to describe one group of CLI options. But I reworded the one use here.

docs/HowToFAQ/FAQSplitHere.md
davidcok: Possibly, but it is also definitely the case that two different assertions can benefit from the prover internally establishing (internal) lemmas etc. just once, in a combined VC attempt. At least the original engineering folklore for provers was that it was usually in total more efficient to attempt all assertions at once -- because of similar work needing to be done only once. However, I don't know if that has been measured recently with current designs of proof tools. It would be a nice exercise.

docs/HowToFAQ/FAQTriggers.md
davidcok: Fixed.

davidcok: Fixed

docs/HowToFAQ/FAQTriggers.md
davidcok: Fixed.

docs/HowToFAQ/FAQSplitHere.md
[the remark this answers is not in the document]
keyboardDrummer: Okay, but still, why would the answer to "How does feature X work?" be something other than "Read the reference manual on feature X [here]"  ?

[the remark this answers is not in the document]
davidcok: Well, because a curt 'RTFM' is a bit rude and not entirely helpful :-) I did put in links to the relevant sections of the RM, but also decided to put in a short summary of the difference.

docs/HowToFAQ/FAQTriggers.md
davidcok: I'll make a note to find and add an example.

docs/HowToFAQ/FAQSplitHere.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Answer
+
+Analysis of a method for the purpose of verification identifies a large number of assertions that must be proved.
```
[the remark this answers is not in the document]
davidcok: Thanks for much better wording.

docs/HowToFAQ/FAQSplitHere.md
davidcok: Reworked to include this text

docs/HowToFAQ/FAQSplitHere.md
davidcok: Done

docs/HowToFAQ/FAQSplitHere.md
davidcok: OK

docs/HowToFAQ/FAQSplitHere.md
```diff
@@ -8,12 +8,13 @@ How do `{:split_here}` and `{:focus}` work to divide up a verification condition
 
 ## Answer
 
-Analysis of a method for the purpose of verification identifies a large number of assertions that must be proved.
-These assertions are grouped into batches.
+Verifying a method requires proving a large number of assertions.
+Dafny uses a backend prover (an SMT solver) to verify assertions. The prover may become better or worse at verifying an assertion if you also ask it to also verify another assertion.
```
keyboardDrummer, line 12: double use of "also"

keyboardDrummer (dismissed):

docs/HowToFAQ/FAQSplitHere.md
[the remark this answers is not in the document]
keyboardDrummer: Okay, but the RM should be the best place to read about these features right? I'm a little worried that if we're duplicating information we'll have a race between the two locations, each improving over time but none be strictly better than the other.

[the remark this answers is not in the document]
keyboardDrummer: "at least to describe one group of CLI options"

Some of our options and their related help originate from Boogie, and I think in our CLI redesign we'll stop doing that.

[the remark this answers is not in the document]
davidcok: Yes the RM should be the "final word"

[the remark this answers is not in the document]
davidcok: So I won't enumerate them here...

docs/HowToFAQ/FAQSplitHere.md
davidcok: Fixed

docs/HowToFAQ/FAQTriggers.md
davidcok: Added a reference and txt after discussion with Rustan.

keyboardDrummer (dismissed):

keyboardDrummer (approved):

Issue #2493, opened 2022-07-26
Dafny 4.0: Need syntax to refer to the top-level module


davidcok:
I don't see an issue for something I wanted two years ago.

The following was Rustan's solution to (someone else's) question:
```
module util {
  const x: nat := 0
}

module util_ {
  import util
}

module java {
  module util {
    import util_
    const y: nat := util_.util.x // Because you can't write util.x
    method test() { assert y == 0; }
  }
}
```

My proposal was to have a way to represent a sequence of module names starting with the root module, namely an initial `.`
so that we could write in the above `const y: nat := .util.x`, as if the root module were named the empty string.
An alternate solution with a more visible name would be `_.util.x`

cpitclaudel:
I'd love to have something like this.
Related issue, but not overlapping: https://github.com/dafny-lang/dafny/issues/2012

PR #2626, opened 2022-08-22
FAQ Entries - Part B

davidcok:
Portion of #2487


docs/HowToFAQ/FAQFailingPost.md
```diff
@@ -0,0 +1,30 @@
+---
+title: I can assert a condition right before a return, so why does the postcondition fail?
```
cpitclaudel, line 2: I didn't immediately what the title of this one meant ("postcondition fail")

docs/HowToFAQ/FAQFailingPost.md
```diff
@@ -0,0 +1,30 @@
...
+---
+
+## Question
+
+I can assert a condition right before a return, so why does the postcondition fail?
```
cpitclaudel, line 7: Same here

docs/HowToFAQ/FAQFailingPost.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Answer
+
+There can be lots of reasons why a postcondition might fail to verify, but if you can prove the same predicate just before a return, 
+a typical reason is that there are other exit paths from the method or function. There might be other `return` statements, but a
+harder to spot exit path is the `:-` let-or-fail assignment.
+Here is a sketch of that kind of problem:
+
+```
```
cpitclaudel, line 16:
```suggestion
```dafny
```

docs/HowToFAQ/FAQFailingPost.md
```diff
@@ -0,0 +1,30 @@
...
+method test(MyClass o) returns (r: Result<int>)
+  modifies o;
+  ensures o.ok == true;
```
cpitclaudel, line 19: Could we make this a self-contained example?

cpitclaudel: (Possibly by importing the standard library)

docs/HowToFAQ/FAQFunctionUnroll.md
```diff
@@ -0,0 +1,36 @@
+---
+title: A function seems to work just once. How do I get it to apply a bunch of times?
+---
+
+## Question
+
+A function seems to work just once. How do I get it to apply a bunch of times?
+Here is an example:
+```
+{% include_relative FAQFunctionUnroll0.dfy %}
```
cpitclaudel, line 10: The problem with having separate Dafny files is that the markdown file is harder to read on its own.  Doesn't matter for this PR though.

docs/HowToFAQ/FAQFunctionUnroll.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+The latter two lemmas will not prove without uncommenting the application of the lemma for one less iteration.
+How can I get this to prove automatically?
+
+## Answer
+
+Function bodies are transparent. That is, when there is a call to a function in Dafny code, Dafny will insert inline
```
cpitclaudel, line 18: This is not strictly true; not sure whether it matters in the context of this explanation

docs/HowToFAQ/FAQFunctionUnroll.md
```diff
@@ -0,0 +1,36 @@
...
+the definition of the function. But for recursive functions, it generally only does this once, as just once is
```
cpitclaudel, line 19: It's twice, not once; the function needs to be unfolded twice for sequences of length one, and three times for sequences of length 2, right?

docs/HowToFAQ/FAQFunctionUnroll.md
```diff
@@ -0,0 +1,36 @@
...
+The latter two lemmas will not prove without uncommenting the application of the lemma for one less iteration.
+How can I get this to prove automatically?
+
+## Answer
+
+Function bodies are transparent. That is, when there is a call to a function in Dafny code, Dafny will insert inline
+the definition of the function. But for recursive functions, it generally only does this once, as just once is 
+usually sufficient to form an inductive proof. And always unrolling multiple times can reduce performance.
+
+You can request that Dafny unroll a function multiple times by using the (unintuitively named) `{:fuel}` attribute.
```
cpitclaudel, line 22: I think `fuel` is a pretty common name for this; maybe we can explain why it's called this?

docs/HowToFAQ/FAQFunctionUnroll.md
```diff
@@ -0,0 +1,36 @@
...
+The value of the attribute is the number of times the function will be unrolled. This is still a fixed upper limit,
+but it does the trick in this case:
+```
+{% include_relative FAQFunctionUnroll1.dfy %}
+```
+
+A function will also keep unrolling if it has an argument that is a literal `nat` and that argument decreases
```
cpitclaudel, line 29: It's not just nats; it's any literals (modulo bugs that we have)
This would be a good place to link to the paper on computing with SMT, maybe: https://www.microsoft.com/en-us/research/publication/computing-smt-solver/

docs/HowToFAQ/FAQFunctionUnroll0.dfy
```diff
@@ -0,0 +1,27 @@
+function LengthSum(s: seq<string>): nat {
```
cpitclaudel, line 1: Could we simplify the example to use just `Sum` instead of `LengthSum`?

docs/HowToFAQ/FAQFunctionUnroll2.dfy
```diff
@@ -0,0 +1,33 @@
...
+  if |s| == 0 then 0
+  else |s[0]| + LengthSum(s[1..])
+}
+
+function LengthSum_(s: seq<string>, n: nat): nat
+  ensures LengthSum(s) == LengthSum_(s,n)
+  ensures |s| > 0 && n > 0 ==> LengthSum_(s,n) == |s[0]| + LengthSum_(s[1..], n-1)
+{
+  if |s| == 0 || n == 0 then LengthSum(s)
+              else |s[0]| + LengthSum_(s[1..], n-1)
```
cpitclaudel, line 11: Indentation looks wrong here?

docs/HowToFAQ/FAQFunctionUnroll.md
```diff
@@ -0,0 +1,36 @@
...
+on each recursive call. So for this example we can write a helper function that takes a `nat` argument
+and give it a literal value that tells it how often to unroll.
```
cpitclaudel, line 31:
```suggestion
and give it a literal value that tells it how much to unroll.
```

docs/HowToFAQ/FAQFunctionUnroll.md
```diff
@@ -0,0 +1,36 @@
...
+```
+{% include_relative FAQFunctionUnroll2.dfy %}
+```
+With this solution, the number of unrollings can be set at the call site, rather than in the function definition.
```
cpitclaudel, line 35: But that number is still limited by other internal factors (like the limit on quantifiers instantiations); now sure if it's worth mentioning here.

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,58 @@
+---
+title: If I have an assertion about a class and a loop that doesn't mention (read, modify)  the class, why does dafny fail to establish the assertion after the loop?
+---
+
+## Question
+
+If I have an assertion about a class and a loop that doesn't mention (read, modify)  the class, why does dafny fail to establish the assertion after the loop?
```
cpitclaudel, line 7: Why do we duplicate the title in every entry?

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Answer
+
+TODO - This may need adjustment of the modifies/assignmetn targets discussion
+
+The short answer is that you need an appropriate combination of modifies clause and loop invariants to prove assertions about the end result of a loop.
+
+In Dafny's way of reasoning about a loop (which is typical of verifier systems), the loop invariant is the only thing you can assume at the top of every iteration. In other words, at the top of each loop iteration, it is as if all variables have arbitrary values, modulo the loop invariant. Sometimes, the word _havoc_ is used to describe this: the verifier havocs all variables, that is, gives all variables arbitrary values, and then assumes the loop invariant.
```
cpitclaudel, line 15: I find this style confusing: first we state something that is wrong, and then we say "not quite"

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,58 @@
...
+
+If that’s all the verifier did, life would be a pain, because you’d have to write a loop invariant about all variables in your program, even those that have nothing to do with the loop.
+Verifier actually does better. Instead of havocing all variables, it suffices to havoc the assignment targets within the body of the loop, including anything that might be modified by methods called in the loop body. That is, the verifier uses a simple syntactic scan of the loop body to see which variables may possibly be assigned, and then it havocs only those variables. This saves users from having to write loop invariants about all other variables. Only invariants about variables
```
cpitclaudel, line 18: And then here again we say something that is wrong (havocs only those variables), and then refine again later

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,58 @@
...
+that are modified are needed.
+
+What about the heap? For this purpose, the heap is considered to be one variable. So, the heap is either an assignment target of the loop or it isn’t. This means that if your loop body assigns to anything in the heap, the heap becomes an assignment target. Also, if the loop body allocates another object, that too is a change of the heap, so the heap becomes an assignment target. Furthermore, Dafny’s rule about a method is that a method is allowed to change the state of any object in the method’s modifies clause, and it is also allowed to allocate new objects and to modify their state. Therefore, if your loop body calls a method, then the heap my well become an assignment target.
+ So, then, does this mean your loop invariant needs to speak about everything in the heap whenever the heap is an assignment target of the loop?
```
cpitclaudel, line 22: And here again.

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,58 @@
...
+
+But when it is not, you need to be explicit about the `modifies` clause and the loop invariants. In particular
+- write a `modifies` clause for the loop that includes all objects whose fields might be assigned in the loop body (or listed in the modifies clauses of anything the loop body calls)
+- then write a loop invariant that includes assertions about any variable that listed in the modifies clause. Very typically, the invariant will give a value for a havoced variable showing its relationship to the loop index.
+For example, a loop that sets the elements of an array to some initial value might loop like this:
+```dafny
+method init(a: array<int>) 
+  modifies a
+  ensures forall j: int :: 0 <= j < a.Length ==> a[j] == j;
```
cpitclaudel, line 33:
```suggestion
  ensures forall j | 0 <= j < a.Length :: a[j] == j;
```

(There are other examples of this pattern, which I think would be worth fixing too)

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,58 @@
...
+{
+  var i: int := 0;
+
+  while (i < a.Length)
```
cpitclaudel, line 37: Could we use a for loop here?

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,58 @@
...
+- The loop says that it modifies the same array. In fact that modifies clause could be omitted
+because it would be inherited from the enclosing context
+- The loop also modifies `i`, but as `i` is a local variable of non-reference type, it need not be listed in the modifies clause.
+- The loop also has an invariant, which has two conjuncts
+   - One conjunct states what may be assumed about `i`. Even though `i` is not in the modifies clause 
+    it is an assignment target, so we need to say what may be assumed about it (prior to the loop test)
+   - The other conjunct says what is known about the elem,ents of `a`, which depends on `i`, that is, omn how many iterations of the loop have been executed
+- And then from that loop invariant and the fact that the loop has now terminated (which implies that `i == a.Length`), Dafny can prove the method's postcondition. 
+
+Even when Dafny can infer an appropriate modifies clause, it cannot infer loop invariants. The user always needs to supply those (until verification technology improves)
```
cpitclaudel, line 58: Arguably, the syntactic analysis pass above is a kind of invariant inference?

docs/HowToFAQ/FAQMethodSequence.md
```diff
@@ -0,0 +1,16 @@
...
+title: Is there a way to write `if foo().equals(bar()) { x } else { y }` where `foo` and `bar` are methods?
+---
+
+## Question
+
+Is there a way to write `if foo().equals(bar()) { x } else { y }` where `foo` and `bar` are methods?
+
+## Answer
+
+No. Methods have side-effects and so Dafny does not allow them in expressions.
```
cpitclaudel, line 11: We should explain what an expression and a statement are, here.

docs/HowToFAQ/FAQMethodSequence.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Is there a way to write `if foo().equals(bar()) { x } else { y }` where `foo` and `bar` are methods?
```
cpitclaudel, line 2: I think this question is too specific.  What is `.equals`?  What are `x` and `y`?

docs/HowToFAQ/FAQModuleImport.md
```diff
@@ -0,0 +1,22 @@
...
+Why is my import opened statement not working in the following example:
+```
+{% include_relative FAQModuleImport.dfy %}
+```
+
+## Answer
+
+Although nested modules are nested inside other modules, they should really be thought of as their own module.
+There is no particular benefit to module `A.B` from being nested inside module `A` --- it does not see any declarations of `A` other than sibling modules of `B`.
+The benefit is to module `A`, for which the nested module `B` is a namespace that can group some related declarations together.
```
cpitclaudel, line 16: I don't understand this part

docs/HowToFAQ/FAQModuleImport.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Why is my import opened statement not working?
```
cpitclaudel, line 2: Assuming that people will scan through a list of FAQ titles, I'm not sure that this title will help users find this Q&A.  Maybe "Why do children modules not inherit the imports of their parents?"

(There's nothing specific to `opened` here, is there?

docs/HowToFAQ/FAQModuleNames.md
```diff
@@ -0,0 +1,19 @@
...
+## Question
+
+How do I disambiguate module names in an example like this:
+```dafny
+{% include_relative FAQModuleNames0.dfy %}
+```
+
+## Answer
+
+There is no direct syntax to do what is wanted here.
```
cpitclaudel, line 14: Might be worth a link to the corresponding issues https://github.com/dafny-lang/dafny/issues/2493 and https://github.com/dafny-lang/dafny/issues/2493

docs/HowToFAQ/FAQMultClauses.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Question
+
+When a lemma has multiple ensures clauses, I’m finding that they interact, when I expected them to be independent.  For example, commenting out one of them can make another one not verify.  How should I think about this?
+
+## Answer
+
+Multiple ensures clauses, such as `ensures P ensures Q ensures R` is equivalent to the conjunction, in order, of the ensures predicates: `ensures P && Q && R`.
+The order is important if an earlier predicate is needed to be sure that a later one is well defined. For example the two predicates in `ensures a != null ensures a.Length > 0`
+must be in that order because `a.Length` is well-defined only if `a != null`.
```
cpitclaudel, line 13: Arrays cannot be null; may we can give an example with an index and an array?

docs/HowToFAQ/FAQMultClauses.md
```diff
@@ -0,0 +1,18 @@
...
+
+In addition, sometimes one ensures clause serves as a lemma for the next one. Without the earlier clause, Dafny can have trouble proving the second one, even though it is valid.
```
cpitclaudel, line 15: Lemma has a specific meaning in this context, so this might cause confusion?

docs/HowToFAQ/FAQSeqTrait.md
```diff
@@ -0,0 +1,29 @@
...
+  var a: A := new A;
+  var sa: seq<A> := [ a ];
+  var b := new B;
+  var sb : seq<B> := [b ];
+  var st : seq<T> := sa + sb;
+  var so : seq<object> := sa + sb;
+}
+```
+
+In fact, Dafny will generally infer a common type for you in the case of sequence concatentation.
```
cpitclaudel, line 29:
```suggestion
In fact, Dafny will generally infer a common type for you in the case of sequence concatetation.
```

cpitclaudel: Maybe we should add something about using a union type otherwise?

docs/HowToFAQ/FAQSingleton.md
```diff
@@ -0,0 +1,42 @@
+---
+title: How do I make a singleton instance of a class for repeated later use?
+---
+
+## Question
+
+How do I make a singleton instance of a class for repeated later use?
```
cpitclaudel, line 7: Isn't the answer to this question that it can't be done?

docs/HowToFAQ/FAQSingleton.md
```diff
@@ -0,0 +1,42 @@
...
+In general, one might need to allocate an object, do some initialization,
+but then treat it as const from then on.
```
cpitclaudel, line 9: Is that really called a singleton?

cpitclaudel (changes requested):
Great set of questions.  I left comments where I thought the wording or title could be improved, or where as a reader I would have wanted more details.

docs/HowToFAQ/FAQMultClauses.md
davidcok: Reworded.

docs/HowToFAQ/FAQFunctionUnroll.md
davidcok: Just removed the gratuitous parenthetical clause.

docs/HowToFAQ/FAQFunctionUnroll0.dfy
davidcok: Done

docs/HowToFAQ/FAQFunctionUnroll2.dfy
davidcok: Fixed

docs/HowToFAQ/FAQLoopModifies.md
davidcok: Many do, but many also give more detail in the Question section , like the code for the problem example.

docs/HowToFAQ/FAQSingleton.md
davidcok: I think so. In this cases it is the language the questioner used. I suppose one might make a distinction between a class of which a program only ever allocates a single instance (as here) vs. a class in which every subsequent attempt to allocate a class instance always returns the same original instance (perhaps the technical definition of the pattern).

I didn't think the distinction was worth berating here.

docs/HowToFAQ/FAQSingleton.md
davidcok: I think what the questioner was asking for can be done. If what you mean is that can it be that every attempt to allocate a class instance (`new C`) always results in the same instance (as opposed to a container delegating to a common instance) then I suppose it can't be done in Dafny.

docs/HowToFAQ/FAQMultClauses.md
davidcok: Sure they can, that is variables of array?<> type can be null:
```
method ma(a: array?<int>) {
  if a.Length == 0 { } // ERROR - a might be null
}
```

docs/HowToFAQ/FAQLoopModifies.md
davidcok: we could, but I left Rustan's example.

docs/HowToFAQ/FAQLoopModifies.md
davidcok: In this case I copied Rustan's answer to the question.

I reread it and found for myself the development of the answer helpful and not confusing.. But if you do think users would find the discussion confusing, I'll do a full rewrite.

docs/HowToFAQ/FAQLoopModifies.md
davidcok: As above

docs/HowToFAQ/FAQLoopModifies.md
davidcok: As above.

docs/HowToFAQ/FAQLoopModifies.md
davidcok: Indeed. At least for those variables that are not modified.

docs/HowToFAQ/FAQMethodSequence.md
davidcok: Well, that is how the questioner put it, with some pseudo code. But to your point I took out the example and left the Q's point.

docs/HowToFAQ/FAQModuleImport.md
davidcok: I'm not sure of your confusion. For an enclosing module A, a nested module A.B serves to group related declarations within B, taking them out of A's direct namespace, but leaving them easily accessible (without needing an import) within A.

docs/HowToFAQ/FAQModuleImport.md
davidcok: Reworded the question.
And I'll make a note to go through the whole list for similar overly specific questions.

docs/HowToFAQ/FAQModuleNames.md
davidcok: Added a link.

docs/HowToFAQ/FAQMethodSequence.md
davidcok: Expanded the answer

docs/HowToFAQ/FAQFunctionUnroll.md
davidcok: OK. But I agree it is not worth that level of internal detail.

docs/HowToFAQ/FAQFunctionUnroll.md
davidcok: Reworded. Tell me if this is better or not.

docs/HowToFAQ/FAQFunctionUnroll.md
davidcok: Not sure. But reworded to be less specific.

docs/HowToFAQ/FAQFunctionUnroll.md
davidcok: Added the reference.

davidcok: And fixed the text.

docs/HowToFAQ/FAQFailingPost.md
davidcok: Adjusted the question

docs/HowToFAQ/FAQFailingPost.md
davidcok: Ditto

docs/HowToFAQ/FAQFailingPost.md
davidcok: Added a reference to the Result type in the library, but I'd rather not include the whole datatype because the code is rather lengthy.

docs/HowToFAQ/FAQFunctionUnroll.md
```diff
@@ -15,22 +15,24 @@ How can I get this to prove automatically?
...
+A function will also keep unrolling if it has an argument that is a literal and that argument decreases
 on each recursive call. So for this example we can write a helper function that takes a `nat` argument
-and give it a literal value that tells it how often to unroll.
+and give it a literal value that tells it how much to unroll.
 ```
 {% include_relative FAQFunctionUnroll2.dfy %}
 ```
 With this solution, the number of unrollings can be set at the call site, rather than in the function definition.
 
+[This paper](https://www.microsoft.com/en-us/research/publication/computing-smt-solver/) gives some technical insight into how recursive functions are handled in situations liek these examples.
```
cpitclaudel, line 37:
```suggestion
[This paper](https://www.microsoft.com/en-us/research/publication/computing-smt-solver/) gives some technical insight into how recursive functions are handled in situations like these examples.
```

docs/HowToFAQ/FAQFunctionUnroll0.dfy
```diff
@@ -1,27 +1,27 @@
-function LengthSum(s: seq<string>): nat {
+function Sum(s: seq<string>): nat {
```
cpitclaudel, line 1: I was suggesting to use a sequence of numbers

docs/HowToFAQ/FAQFunctionUnroll0.txt
```diff
@@ -1,6 +1,6 @@
-FAQFunctionUnroll0.dfy(16,0): Error: A postcondition might not hold on this return path.
-FAQFunctionUnroll0.dfy(15,23): Related location: This is the postcondition that might not hold.
-FAQFunctionUnroll0.dfy(24,0): Error: A postcondition might not hold on this return path.
-FAQFunctionUnroll0.dfy(23,23): Related location: This is the postcondition that might not hold.
+/Users/davidcok/projects/dafny/dafny/docs/HowToFAQ/FAQFunctionUnroll0.dfy(16,0): Error: A postcondition might not hold on this return path.
```
cpitclaudel, line 1: Missing usebasenameforfilename

docs/HowToFAQ/FAQLoopModifies.md
cpitclaudel: I'd like to get a second opinion, but it's not worth blocking the whole PR on.  Can we get maybe @MikaelMayer ?

docs/HowToFAQ/FAQSeqTrait.md
```diff
@@ -26,4 +26,4 @@ method m() {
 }
 ```
 
-In fact, Dafny will generally infer a common type for you in the case of sequence concatentation.
+In fact, Dafny will generally infer a common type for you in the case of sequence concatetation.
```
cpitclaudel, line 29:
```suggestion
In fact, Dafny will generally infer a common type for you in the case of sequence concatenation.
```

cpitclaudel (changes requested):
Just a few more tiny things

docs/HowToFAQ/FAQFunctionUnroll.md
cpitclaudel: (Side note: It's harder to see the follow-up if you mark the comment resolved)

docs/HowToFAQ/FAQFunctionUnroll.md
cpitclaudel: I am sure: it is twice

docs/HowToFAQ/FAQFunctionUnroll0.dfy
cpitclaudel: I mean Sum of numbers, not of lengths

docs/HowToFAQ/FAQModuleImport.md
cpitclaudel: I just meant that the text was not clear to me (I don't know what it means for a module to "see" another module, and I'm not sure that a module counts as a "declaration")

docs/HowToFAQ/FAQMultClauses.md
cpitclaudel: Sure, but that's `array?`, not `array`.  But because in previous versions `array` (without `?`) could be null as well, we find leftover docs from time to time that still check for `null` arrays; so I was suggesting to use a value type instead to not add to the confusion.

docs/HowToFAQ/FAQSingleton.md
cpitclaudel: Maybe I'm misunderstanding the code below, but I don't think we can do a container delegating to a common instance either?

docs/HowToFAQ/FAQMultClauses.md
davidcok: I clarified that `a` is an `array?`

docs/HowToFAQ/FAQFailingPost.md
davidcok: Added that, with additional text

docs/HowToFAQ/FAQModuleImport.md
davidcok: Revised

docs/HowToFAQ/FAQSingleton.md
davidcok: Removed this FAQ for now, to be investigated further.

docs/HowToFAQ/FAQFunctionUnroll.md
davidcok: Fixed

docs/HowToFAQ/FAQFunctionUnroll0.dfy
davidcok: Done.

docs/HowToFAQ/FAQFunctionUnroll0.txt
davidcok: Fixed

docs/HowToFAQ/FAQModuleNames.md
```diff
@@ -17,3 +17,5 @@ If this structure cannot be changed, then the following somewhat ugly code is a
 ```dafny
 {% include_relative FAQModuleNames.dfy %}
 ```
+
+There is discussion of defining synttax that names the top-level module, which would make an easy way to solve the above problem. See [this issue](https://github.com/dafny-lang/dafny/issues/2493).
```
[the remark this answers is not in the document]
davidcok: Fixed

davidcok:
Removed LoopModifies for separate discussion

docs/HowToFAQ/FAQModuleNames.md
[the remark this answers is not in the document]
cpitclaudel: I can't see the fix; am I looking at the wrong code?

cpitclaudel, line 21:
```suggestion
There is discussion of defining syntax that names the top-level module, which would make an easy way to solve the above problem. See [this issue](https://github.com/dafny-lang/dafny/issues/2493).
```

docs/_includes/footer.html
```diff
@@ -1,4 +1,5 @@
 <footer class="site-footer h-card">
+FOOTER
```
cpitclaudel, line 2: Is this change on purpose?

cpitclaudel (changes requested):
Looks good, just the FOOTER change and the synttax typo

cpitclaudel (approved):

PR #2630, opened 2022-08-23
fix: Read hard-coded line from `version.cs`

prvshah51:

### DIFF
--- a/Scripts/package.py
+++ b/Scripts/package.py
@@ -255,7 +255,7 @@ def check_version_cs(args):
     # Checking version.cs
     fp = open(path.join(SOURCE_DIRECTORY,"version.cs"))
     lines = fp.readlines()
-    verline = lines[5]
+    verline = lines[6]
     qstart = verline.index('"')
     qend = verline.index('"', qstart+1)
     lastdot = verline.rindex('.',qstart)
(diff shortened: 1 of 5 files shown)


fabiomadge (approved):

Scripts/package.py
```diff
@@ -253,14 +253,10 @@ def pack(args, releases):
...
-    verline = lines[5]
-    qstart = verline.index('"')
-    qend = verline.index('"', qstart+1)
-    lastdot = verline.rindex('.',qstart)
-    v1 = verline[qstart+1:lastdot]
-    v2 = verline[lastdot+1:qend]
+    for line in open(path.join(SOURCE_DIRECTORY,"version.cs")):
+        res = re.match(r'^\s*\[assembly:\s+AssemblyVersion\("([0-9]+.[0-9]+.[0-9]+).([0-9]+)"\)\]\s*$', line)
+        if res:
+            (v1, v2) = res.groups()
```
cpitclaudel, line 259: Let's use `with open(…) as fp:` instead of `open`.  Then there's no need to `close`

prvshah51: I have made changes on other PR [here](https://github.com/dafny-lang/dafny/pull/2634).

PR #2642, opened 2022-08-25
Strongly typed dafny document

keyboardDrummer:
### Changes

- DafnyDocument has been split up into `Compilation`, `CompilationAfterParsing`, `CompilationAfterResolution` and `CompilationAfterTranslation`, which allows several fields to become non-null, enabling dropping several `!` annotations and null checks
- Removed several fields, `DafnyDocument.WasResolved`, `DafnyDocument.LoadCanceled`, `DafnyDocument.CanDoVerification`.
- CompilationManager, which should be agnostic to document changes and thus migration, no longer migrates last touched positions.


Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -0,0 +1,175 @@
...
+  /// <summary>
+  /// Internal representation of a specific version of a Dafny document.
+  ///
+  /// Only one instance should exist of a specific version.
+  /// Asynchronous compilation tasks use this instance to synchronise on
+  ///
+  /// When verification starts, no new instances of Compilation will be created for this version.
+  /// There can be different verification threads that update the state of this object.
+  /// </summary>
+  public class Compilation {
```
MikaelMayer, line 26: Let's synchronize the comment that mentions "Dafny document", the variables that are named "document" everywhere and the name of the class that only says "compilation"
I like the name "compilation" because the way you decline it after each phase, but that name is also ambiguous because compilation can also be a process, whereas here you want to return the result.
What about changing `Compilation` to any of these more unambiguous
* `Document` (yes back, to the origins...) --
* `DafnyDocument` (yes back, to the origins...) --
* `CompilationResult`
* `IntermediateDocument`
* `Result`
* `Outcome`

Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -0,0 +1,175 @@
...
+      return Snapshot(new CompilationView(TextDocumentItem, Array.Empty<Diagnostic>(),
+        SymbolTable.Empty(TextDocumentItem), new Dictionary<ImplementationId, ImplementationView>(),
+        false, Array.Empty<Diagnostic>(),
+        new DocumentVerificationTree(TextDocumentItem)));
+    }
+
+    /// <summary>
+    /// Creates a clone of the DafnyDocument
+    /// </summary>
+    public virtual CompilationView Snapshot(CompilationView previousView) {
```
MikaelMayer, line 47: Since all your method named `Snapshot` return a `CompilationView`, I suggest you choose between `View` and `Snapshot`. I prefer `Snapshot` because it's less ambiguous than `View` (a view of a data structure can change if the underlying data structure change), but I understand that verification trees are updated anyway so `View` might be preferred.

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
```diff
@@ -1,16 +1,35 @@
...
+    bool ImplementationsWereUpdated,
+    IEnumerable<Diagnostic> GhostDiagnostics,
+    VerificationTree VerificationTree
+    ) {
+
+    public DocumentUri Uri => TextDocumentItem.Uri;
+    public int? Version => TextDocumentItem.Version;
+
+    public IEnumerable<Diagnostic> Diagnostics =>
+      ResolutionDiagnostics.Concat(ImplementationViews.Values.Select(v => v.Diagnostics).SelectMany(x => x));
```
MikaelMayer, line 31: Probably for another PR, but with this architecture it seems much easier to fix the problem of resolution diagnostics lost in the middle of previous verification diagnostics. You could simply detect resolution diagnostics, and if so, change the level in implementation diagnostics to "Warning", right?

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
```diff
@@ -51,12 +64,19 @@ public class NotificationPublisher : INotificationPublisher {
       languageServer.TextDocument.SendNotification(DafnyRequestNames.VerificationSymbolStatus, notification);
     }
 
-    private static FileVerificationStatus? GetFileVerificationStatus(DafnyDocument document) {
-      if (document.ImplementationIdToView == null || document.VerificationTasks == null) {
+    private static FileVerificationStatus? GetFileVerificationStatus(CompilationView document) {
+      if (!document.ImplementationsWereUpdated) {
+        /*
+         CompilationAfterResolution.Snapshot() gets migrated ImplementationViews.
```
MikaelMayer, line 70: Could you elaborate a bit? I'm not sure I understand that point. ImplementationViews are not migrated directly?

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
```diff
@@ -51,12 +64,19 @@ public class NotificationPublisher : INotificationPublisher {
...
+         It has to get migrated Diagnostics inside ImplementationViews, otherwise we get incorrect diagnostics.
+         However, migrating the ImplementationId's may mean we lose verifiable symbols, which we don't want at this point. TODO: why not?
```
MikaelMayer, line 72: TODO here. I'm not sure I can help since I don't understand this paragraph. Could you clarify a bit?

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
```diff
@@ -51,12 +64,19 @@ public class NotificationPublisher : INotificationPublisher {
...
+         To prevent publishing file verification status unless the current document has been translated,
+         the field ImplementationsWereUpdated was added.
+         */
         return null;
       }
-      return new FileVerificationStatus(document.Uri, document.Version,
-        GetNamedVerifiableStatuses(document.ImplementationIdToView));
+      return new FileVerificationStatus(document.TextDocumentItem.Uri, document.TextDocumentItem.Version,
+        GetNamedVerifiableStatuses(document.ImplementationViews));
```
MikaelMayer, line 79: You didn't like the name ImplementationIdToViews? :-) I thought it was more explicit than a set-based name.

Source/DafnyLanguageServer/Workspace/DocumentObserver.cs
```diff
@@ -64,14 +56,14 @@ class DocumentObserver : IObserver<DafnyDocument> {
...
     lock (lastPublishedDocumentLock) {
-      if (document.Version < LastPublishedDocument.Version) {
+      if (view.Version < LastPublishedDocument.Version) {
         return;
       }
 
-      notificationPublisher.PublishNotifications(LastPublishedDocument, document);
-      LastPublishedDocument = document.Snapshot(); // Snapshot before storing.
+      notificationPublisher.PublishNotifications(LastPublishedDocument, view);
+      LastPublishedDocument = view; // Snapshot before storing.
```
MikaelMayer, line 66: The comment is obsolete, right?

Source/DafnyLanguageServer/Workspace/DocumentObserver.cs
```diff
@@ -42,19 +38,15 @@ class DocumentObserver : IObserver<DafnyDocument> {
...
-    var previousDiagnostics = LastPublishedDocument.LoadCanceled
-      ? new Diagnostic[] { }
-      : LastPublishedDocument.ParseAndResolutionDiagnostics;
     var internalErrorDiagnostic = new Diagnostic {
       Message =
         "Dafny encountered an internal error. Please report it at <https://github.com/dafny-lang/dafny/issues>.\n" +
         exception,
       Severity = DiagnosticSeverity.Error,
-      Range = LastPublishedDocument.Program.GetFirstTopLevelToken().GetLspRange()
+      Range = new Range(0, 0, 0, 1)
```
MikaelMayer, line 46: Hopefully this works. Have you tested it?

Source/DafnyLanguageServer/Workspace/DocumentManagerDatabase.cs
```diff
@@ -71,18 +72,18 @@ public class DocumentManagerDatabase : IDocumentDatabase {
       return false;
     }
 
-    public Task<DafnyDocument?> GetResolvedDocumentAsync(TextDocumentIdentifier documentId) {
+    public Task<CompilationView?> GetResolvedDocumentAsync(TextDocumentIdentifier documentId) {
```
MikaelMayer, line 75: I feel weird that two similar named methods `GetResolvedDocumentAsync` and `GetLastDocumentAsync` now return two completely different classes. Maybe you could rename these methods so that it's obvious they return things that are different?

Source/DafnyLanguageServer.Test/Synchronization/DeclarationLocationMigrationTest.cs
```diff
@@ -82,7 +81,6 @@ class C {
       );
       var document = await Documents.GetResolvedDocumentAsync(documentItem.Uri);
       Assert.IsNotNull(document);
-      Assert.IsFalse(document.SymbolTable.Resolved);
```
MikaelMayer, line 85: How do you justify removing this assertion? Don't we want to test that the document's symbol table was not resolved?

MikaelMayer (changes requested):
Great job, that will make it easier to split the document from its state for publication that mixes up with the previous view. I'm sharing some points of discussion so that we can move it forward. Thanks

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
```diff
@@ -1,16 +1,35 @@
...
+using System.Net.Mime;
 using Microsoft.Extensions.Logging;
-using Serilog.Core;
 using Microsoft.Dafny.LanguageServer.Language;
+using Microsoft.Dafny.LanguageServer.Language.Symbols;
 using Microsoft.Extensions.Options;
+using OmniSharp.Extensions.LanguageServer.Protocol;
 
 namespace Microsoft.Dafny.LanguageServer.Workspace {
+  public record CompilationView(
```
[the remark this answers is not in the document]
keyboardDrummer: > Could you please move this record to its own file?

Good idea.

> Correct?

Yes. I'm not sure what the best name is! I chose 'View' because it's what is exposed externally to the client about the state of compilation. It could also be something that abstracts away from compilation, like `VisibleIdeState`

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
keyboardDrummer: Thanks, I plan to make that change after this PR.

Source/DafnyLanguageServer.Test/Synchronization/DeclarationLocationMigrationTest.cs
keyboardDrummer: Why would you want to test this? It seems like it's testing implementation instead of behavior, which makes the test brittle. For migrated symbol tables, .Resolved is always false but it's never read so the value doesn't matter.

Source/DafnyLanguageServer/Workspace/Compilation.cs
keyboardDrummer: Renamed `Compilation` et al. to `Document` et al. , `CompilationManager` to `Compilation`, and `CompilationView` to `IdeState`

Source/DafnyLanguageServer/Workspace/DocumentObserver.cs
keyboardDrummer: Tested on VSCode (after your comment), and it works. I added a unit test to verify this stays 0,0,0,1.

I've also tested this on VSCode with master, that emits a -1,-1,-1,-1 range, and it does _not_ work.

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -223,12 +220,11 @@ method Abs(x: int) returns (y: int)
       );
     }
 
-    private async Task<TextDocumentItem> GetDocumentItem(string source, string filename, CompilationStatus expectedStatus) {
+    private async Task<TextDocumentItem> GetDocumentItem(string source, string filename) {
       source = source.TrimStart();
       var documentItem = CreateTestDocument(source, filename);
       await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
-      var lastStatus = await WaitUntilDafnyFinishes(documentItem);
-      Assert.AreEqual(expectedStatus, lastStatus);
```
MikaelMayer, line 231: I like that you have a new way more robust to get the last status; however, you removed one test here. Is there are good reason why you removed it?

Source/DafnyLanguageServer.Test/Various/CompilationStatusNotificationTest.cs
```diff
@@ -97,10 +97,6 @@ method Abs(x: int) returns (y: int)
       await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
       await AssertProgress(documentItem, CompilationStatus.ResolutionStarted);
       await AssertProgress(documentItem, CompilationStatus.CompilationSucceeded);
-      await AssertProgress(documentItem, CompilationStatus.VerificationStarted);
-      await AssertProgress(documentItem, CompilationStatus.VerificationStarted, "0/1 Abs");
-      await AssertProgress(documentItem, CompilationStatus.VerificationStarted, "1/1 (Abs finished)");
-      await AssertProgress(documentItem, CompilationStatus.VerificationSucceeded);
```
MikaelMayer, line 103: What happened to this test? Just wanted to confirm that you disabled this behavior so we are not left with untested code.

Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -1,175 +1,312 @@
...
+  private readonly INotificationPublisher notificationPublisher;
+  private readonly IProgramVerifier verifier;
+
+  private DafnyOptions Options => DafnyOptions.O;
+  private VerifierOptions VerifierOptions { get; }
+
+  public DocumentTextBuffer TextBuffer { get; }
+  private readonly IServiceProvider services;
+
+  // TODO CompilationManager shouldn't be aware of migration
```
MikaelMayer, line 43: Can you elaborate a bit more on this TODO?

Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -1,175 +1,312 @@
...
+  private async Task<DocumentAfterParsing> ResolveAsync() {
+    try {
+      var parsedCompilation = await documentLoader.LoadAsync(TextBuffer, cancellationSource.Token);
+
+      // TODO, let gutter icon publications also used the published CompilationView.
+      var compilationView = parsedCompilation.InitialIdeState();
+      compilationView = compilationView with {
+        VerificationTree = migratedVerificationTree ?? compilationView.VerificationTree
       };
+      notificationPublisher.PublishGutterIcons(compilationView, false);
```
MikaelMayer, line 83:
```suggestion
      var state = parsedCompilation.InitialIdeState();
      state = state with {
        VerificationTree = migratedVerificationTree ?? compilationView.VerificationTree
      };
      notificationPublisher.PublishGutterIcons(state, false);
```

Source/DafnyLanguageServer/Workspace/Document.cs
```diff
@@ -0,0 +1,174 @@
...
+using Microsoft.Extensions.Logging;
+using Range = OmniSharp.Extensions.LanguageServer.Protocol.Models.Range;
+
+namespace Microsoft.Dafny.LanguageServer.Workspace {
+
+  /// <summary>
+  /// Internal representation of a specific version of a Dafny document.
+  ///
+  /// Only one instance should exist of a specific version.
+  /// Asynchronous compilation tasks use this instance to synchronise on
```
MikaelMayer, line 21: Missing full stop or missing end of phrase?

Source/DafnyLanguageServer/Workspace/Document.cs
```diff
@@ -0,0 +1,174 @@
...
+      GhostDiagnostics = ghostDiagnostics;
+    }
+
+    public IReadOnlyList<Diagnostic> ParseAndResolutionDiagnostics { get; }
+    public SymbolTable SymbolTable { get; }
+    public IReadOnlyList<Diagnostic> GhostDiagnostics { get; }
+
+    public override IEnumerable<Diagnostic> Diagnostics => ParseAndResolutionDiagnostics;
+
+    public override IdeState ToIdeState(IdeState previousState) {
```
MikaelMayer, line 96: Excellent, that makes much more sense now that we use a document to update an existing IdeState

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -43,24 +43,24 @@ public class DocumentManager {
...
-      null,
       VerifyOnOpen,
-      ImmutableList.Create<Position>(),
+      ChangedRanges,
       null);
-    observerSubscription = CompilationManager.DocumentUpdates.Subscribe(observer);
+    observerSubscription = CompilationManager.DocumentUpdates.Select(d => d.NotMigratedSnapshot()).Subscribe(observer);
   }
 
+  private const int MaxRememberedChanges = 100;
```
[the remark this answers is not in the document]
MikaelMayer: Oh maybe I haven't been clear enough. Your new code still keeps looking only at the last 100 changes, but now further restricts to the last 5 verifiables. So if you modify the last verifiable 100 times, you'd get only 1 verifiable at the end, which is what I wanted to avoid. My intention is that you keep the last 5 verifiables no matter the number of changes

In other words, rather than storing and migrating the last 100 changes, you only keep the positions of the last 5 modified verifiables.

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
MikaelMayer: About this? I just want clarification. Also:
```suggestion
         DocumentAfterResolution.Snapshot() gets migrated ImplementationViews.
```

MikaelMayer (changes requested):
We are getting there ! I really like the new names, they make more sense. The bigger point is about remembering the last edited methods, otherwise, just a few minor updates.

Source/DafnyLanguageServer.Test/Various/CompilationStatusNotificationTest.cs
keyboardDrummer: I removed the code that emits VerificationStarted and VerificationSucceeded, so there wasn't anything left to test.

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
keyboardDrummer: The `ImplementationsWereUpdated` boolean is surprising. This comment is trying to argue why it exists. An IdeState that has not been translated yet gets migrated ImplementationViews so that it can publish correct diagnostics, but this would cause the verification symbol status notification to be published as well, since due to the migration it's literal value has changed since last time, even though semantically, nothing has changed and the IDE doesn't need a new notification. This boolean prevents that notification from being emitted.

I think we could try to remove the boolean, but I didn't want to spend time on that now.

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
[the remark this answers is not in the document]
keyboardDrummer: Are you sure about what the new code does? Notice the `.Distinct()` call in `ChangedVerifiables = freshlyChangedVerifiables.Concat(ChangedVerifiables).Distinct().Take(MaxRememberedChangedVerifiables).ToList()`. I believe this will remember up to 5 _different_ verifiables.

That it also remembers change ranges means that if you make multiple changes quickly, without waiting for translation to occur, it'll still remember those changes and may determine up to 5 last changed verifiables from that.

Source/DafnyLanguageServer.Test/Various/CompilationStatusNotificationTest.cs
keyboardDrummer: The behavior was removed, so there was nothing left to test.

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
keyboardDrummer: I find the need for the boolean `ImplementationsWereUpdated` surprising. I think we may be able to get rid of it but I didn't want to spend time on that in this PR. I added what I know about it in this comment.

Notifications are only sent if they're different than the previous notification, but currently a migrated notification, that is semantically the same as the previous one, is also considered different. This boolean prevents verification symbol notification from being sent before translation has been done, since they are never useful to the client.

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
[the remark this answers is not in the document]
MikaelMayer: Oh I see. I hadn't seen the concat with the previous ChangedVerifiables. That's correct then !

MikaelMayer (approved):
Great job, it will be easier to reason about all these documents.

PR #2646, opened 2022-08-26
feat: `for`-loops for the Python compiler

### LINKED ISSUE #2631: Support for-loops and the continue statement in the Python compiler

fabiomadge:
Fixes #2631. Builds on #2486.


Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -30,15 +30,20 @@ def string_of(value) -> str:
     else:
         return str(value)
 
+
```
fabiomadge, line 33:
```suggestion
```

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -822,7 +819,28 @@ protected class ClassWriter : IClassWriter {
...
+      }
+      if (endVarName != null) {
+        wr.Write($", {endVarName}");
+        if (!goingUp) {
+          wr.Write("-1");
+        }
+      }
+      if (!goingUp) {
+        wr.Write(", -1");
+      }
```
cpitclaudel, line 836: IO think this would be more readable with the four cases spelled out separately; WDYT?

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -47,10 +51,27 @@ def label(name: str = None):
...
+@contextmanager
+def c_label(name: str = None):
+    try:
+        yield
+    except Continue as g:
+        if g.target != name:
+            raise g
+    except Break as g:
+        raise g
+    except TailCall as g:
```
cpitclaudel, line 66: Why do you catch and reraise these?

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -47,10 +51,27 @@ def label(name: str = None):
...
+        raise g
 
 def _break(name):
     raise Break(target=name)
 
+def _continue(name):
+    raise Continue(target=name)
```
cpitclaudel, line 73: I wonder if it might be better to inline these functions?

cpitclaudel (changes requested):

cpitclaudel (approved):

PR #2654, opened 2022-08-29
fix: "Failing precondition" on tooltip instead of "return path"

### LINKED ISSUE #2653: A precondition for this call might not hold ... return path?
I found this strange warning for the following code, when hovering the call `Test2(0)`

```
method Test1() {
    Test2(0);
}

method Test2(i: int)
  requires i > 0 {

}
```

image

It should have say "required precondition" instead of "return path".
And even better, it should actually state the required precondition rather than pointing to it.

MikaelMayer:
Fixes #2653.
Also made the tests more robust by removing the need of Task.Delay


Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -229,6 +242,9 @@ method Abs(x: int) returns (y: int)
       await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
       var lastStatus = await WaitUntilDafnyFinishes(documentItem);
       Assert.AreEqual(expectedStatus, lastStatus);
+      foreach (var document in Documents.Documents) {
+        await document.LastDocument;
```
MikaelMayer, line 246: @keyboardDrummer  do your confirm it's the right way to wait for the final translated and verified document to be available?

keyboardDrummer (dismissed):

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
MikaelMayer: Great, that's what I wanted to know.

PR #2656, opened 2022-08-30
Fix 'Sequence contains no elements' race condition in DafnyHoverHandler

keyboardDrummer:
Fixes unstable test failure that occurred in https://github.com/dafny-lang/dafny/runs/8097837657?check_suite_focus=true


Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -132,7 +135,7 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
         }
 
         information += "  \n";
-        if (assertionBatchesToReport.Count == 1) {
+        if (orderedAssertionBatches.Count == 1) {
```
cpitclaudel, line 138: Is this change actually needed? (This is mostly curiosity, I think it's correct in any case)

keyboardDrummer: This change prevents `if (assertionBatchesToReport.Count == 1) { ...; orderedAssertionBatches.First()` }` from throwing a "Sequence has no elements" exception, does it not?

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -122,7 +125,7 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
       }
 
       var assertionBatchesToReport =
-        node.AssertionBatches.Values.OrderByDescending(a => a.ResourceCount).Take(3).ToList();
+        orderedAssertionBatches.OrderByDescending(a => a.ResourceCount).Take(3).ToList();
```
MikaelMayer, line 128: Oh nice finding. Indeed, it could be that between the two the list had grown !

MikaelMayer (approved):

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
keyboardDrummer: I think it would be better to make an immutable copy of relevant information from `VerificationTree` in `DocumentAfterTranslation.ToIdeState`, which downstream consumers like this hover handler can then use.

PR #2658, opened 2022-08-30
fix: Missing related position in failing precondition

### LINKED ISSUE #2657: Crash when "in multiset" in failing precondition
Dafny returns an incorrect error message for this code:

```
lemma Lemma(x: int, p: multiset<int>)
  requires x in p
{}

method Test() {
  Lemma(1, multiset{});
}
```

Dafny returns:

```
git-issue.dfy(6,7): Error: A precondition for this call might not hold.
(0,-1): Related location: This is the precondition that might not hold.
```

Because the related location is wrong, this also crashes the language server
```
Unhandled exception. System.UriFormatException: Given uri is null or empty
at OmniSharp.Extensions.LanguageServer.Protocol.DocumentUri.Parse(String value, Boolean strict)
at Microsoft.Dafny.LanguageServer.Language.DiagnosticErrorReporter.CreateLocation(IToken token) in 
```

MikaelMayer:
Fixes #2657 by adding a token in a generated boogie expression.


Source/Dafny/Verifier/Translator.ExpressionTranslator.cs
```diff
@@ -1882,7 +1882,9 @@ public ExpressionTranslator(ExpressionTranslator etran, string modifiesFrame)
             return disjunction;
           }
         }
-        return Boogie.Expr.Gt(Boogie.Expr.SelectTok(tok, TrExpr(s), elmtBox), Boogie.Expr.Literal(0));
+        var result = Boogie.Expr.Gt(Boogie.Expr.SelectTok(tok, TrExpr(s), elmtBox), Boogie.Expr.Literal(0));
+        result.tok = tok;
+        return result;
```
RustanLeino, line 1887: This seems like it may be wrong for all calls into `Boogie.Expr....`. A possibility would be to fix all of them by changing line 172 of `$Boogie/Source/Core/AbsyExpr.cs` to replace `Token.NoToken` with `e0.tok`.

RustanLeino (approved):

RustanLeino (approved):

RustanLeino (approved):

PR #2663, opened 2022-08-30
feat: Enable branch-coverage analysis in Python

fabiomadge:
Builds on #2661.


Test/comp/BranchCoverage.py
```diff
@@ -0,0 +1,22 @@
+import sys
+
+assert "DafnyProfiling" == __name__
```
cpitclaudel, line 3: Shouldn't this say `BranchCoverage`?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -25,6 +25,12 @@ public static class PythonExtensions {
 
 namespace Microsoft.Dafny.Compilers {
   public class PythonCompiler : SinglePassCompiler {
+    public override void OnPreCompile(ErrorReporter reporter, ReadOnlyCollection<string> otherFileNames) {
+      base.OnPreCompile(reporter, otherFileNames);
+      if (DafnyOptions.O.CoverageLegendFile != null) {
+        Imports.Add("DafnyProfiling");
```
cpitclaudel, line 31: Should this go in `DafnyRuntime`?

cpitclaudel (changes requested):

Test/comp/BranchCoverage.py
fabiomadge: This is hard coded in the SinglePassCompiler: https://github.com/dafny-lang/dafny/blob/65ae35bfc0ab237975b6a2594c55d67adf2a8eee/Source/Dafny/Compilers/SinglePassCompiler.cs#L5357

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: I think the idea is that you get to design the logging output, by making it an extern.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -374,11 +374,13 @@ def IntegerRange(lo, hi):
 class Doubler:
     def __init__(self, start):
         self.start = start
+        self.i = start
     def __iter__(self):
-        return self
+        return Doubler(self.start)
```
cpitclaudel, line 379: I think that's not right either, because now `iter(iter(x))` doesn't behave the same as `iter(x)` (IOW, your object behaves as a mix between an iterator and an iterable).  Does the following work?

```
class Doubler:
    def __init__(self, start):
        self.start = start

    def __iter__(self):
        start = self.start
        while True:
            yield start
            start *= 2
```

(The one thing it's missing is the `__call__ == __next__`, but I'm not sure why we want to have that)

Reference here: https://stackoverflow.com/questions/45685538/whats-the-advantage-of-using-yield-in-iter

Test/comp/BranchCoverage.py
cpitclaudel: But then shouldn't the file be called `DafnyProfiling.py`?

fabiomadge: Oh, I see. We rename the file according to the assert. This gives use more flexibility.

cpitclaudel (approved):

PR #2671, opened 2022-09-01
Adding a section of error explanations

davidcok:


docs/HowToFAQ/ERROR_DuplicateImportName.md
```diff
@@ -0,0 +1,17 @@
+---
+title: "Duplicate name of import: ..."
+---
+
+This error results from importing two modules with the same name. For example
+```
+import A.Util
+import B.util
+```dafny
```
keyboardDrummer, line 9: Can you place the `dafny` tag on the first triple tick? And it seems to be missing on the fixed example.

docs/HowToFAQ/ERROR_InsufficientReads.md
```diff
@@ -0,0 +1,48 @@
...
+Example: This code
+```
+{% include_relative ERROR_InsufficientReads.dfy %}
+```
+produces this output:
+```
+{% include_relative ERROR_InsufficientReads.txt %}
+```
+
+This error message indicates that a nested call of a function needs a bigger `reads` set than its enclosing function provides.
```
keyboardDrummer, line 14: Why is it insufficient? Does `A.Valid` not have an empty reads clause? Doesn't `reads things` indicate that all members of elements in things can be read? How would you make the reads clause sufficient?

Edit: a I see that `{:autocontracts}` is what causes it to fail. Is this a good example to help with understanding reads clauses?

docs/HowToFAQ/ERROR_InsufficientReads1.dfy
```diff
@@ -0,0 +1,23 @@
...
+    constructor() {
+        // no-op
+    }
+}
+
+class B {
+    var things: set<A>
+
+    predicate Valid()
+        reads things
```
keyboardDrummer, line 15: this still has an error because `reads this` is missing

docs/HowToFAQ/ERROR_InsufficientReads1.dfy
```diff
@@ -0,0 +1,23 @@
+class A {
```
keyboardDrummer, line 1: This file doesn't seem to be referenced

docs/HowToFAQ/ERROR_NoTriggers.md
```diff
@@ -0,0 +1,54 @@
+---
+title: "Warning: /!\ No terms found to trigger on."
+---
+
+This warning occurred with the following code:
+```dafny
+predicate ExistsSingleInstance(s : string32, delim : string32)
+  {
+    exists i : nat :: i <= |s| - |delim| && forall x : nat :: x < |delim| ==> s[i + x] == delim[x] &&
```
keyboardDrummer, line 9: This mostly unnamed long expressions are difficult to read. Why did you include an example that contains two types of bugs? I think having one bug per example is easier to learn from.

docs/HowToFAQ/ERROR_NoTriggers.md
```diff
@@ -0,0 +1,54 @@
...
+  (i <= |s| - |delim| && forall x : nat :: x < |delim| ==> s[i + x] == delim[x]) &&
+  forall j : nat :: (j <= |s| - |delim| && forall y : nat :: y < |delim| ==> s[i + y] == delim[y]) ==> i == j
+```
+That is, I’m guessing you wanted to say there exists a unique `i` such that
+`exists i : nat :: i <= |s| - |delim| && forall x : nat :: x < |delim| ==> s[i + x] == delim[x]`.
+
+
+If so, then here is a way to both fix the meaning of the formula and fix the “no trigger” warning:
+```dafny
+predicate SingleInstance(s: string, delim: string, i: nat)
```
keyboardDrummer, line 40: Could you rename `i` to `at` ?

docs/HowToFAQ/ERROR_NoTriggers.md
```diff
@@ -0,0 +1,54 @@
...
+exists i : nat ::
+  (i <= |s| - |delim| && forall x : nat :: x < |delim| ==> s[i + x] == delim[x]) &&
+  forall j : nat :: (j <= |s| - |delim| && forall y : nat :: y < |delim| ==> s[i + y] == delim[y]) ==> i == j
+```
+That is, I’m guessing you wanted to say there exists a unique `i` such that
+`exists i : nat :: i <= |s| - |delim| && forall x : nat :: x < |delim| ==> s[i + x] == delim[x]`.
+
+
+If so, then here is a way to both fix the meaning of the formula and fix the “no trigger” warning:
+```dafny
```
keyboardDrummer, line 39: Could the example start out with nicely written code like here, but still include the no trigger found problem?

docs/HowToFAQ/ERROR_ProverError1.md
```diff
@@ -0,0 +1,5 @@
+---
+title: "Prover error: Unexpected prover response (getting info about 'unknown' response): (:reason-unknown 'Overflow encountered when expanding old_vector')"
+---
+
+This error is caused by a Z3 bug. The best you can do is to try to change the verification condition sent to Z3 by splitting it up, using various options and attributes like `{:split_here}`, `{:focus}`, `/vcsSplitOnEveryAssert`, `/vcsMaxSplits`, and `/randomSeed`.
```
keyboardDrummer, line 5: I'd say it's a Dafny bug since we're using a too old Z3 version ^,^

Also towards the user I think it's easier to talk in terms of Dafny than the underlying technology used.

docs/HowToFAQ/ERROR_DuplicateImportName.md
davidcok: Fixed

docs/HowToFAQ/ERROR_InsufficientReads.md
davidcok: I left the autocontract because it was part of the original question. The remaining problem was that Valid() needs `reads this, things`, not just `reads things`

docs/HowToFAQ/ERROR_InsufficientReads1.dfy
davidcok: Fixed that too

docs/HowToFAQ/ERROR_ProverError1.md
davidcok: Reworded. And hopefully the problem will be moot with a new Z3 soon.

docs/HowToFAQ/ERROR_NoTriggers.md
davidcok: The goal was to include user questions as they posed them. But I agree -- I removed the part about bad operator precedence.

docs/HowToFAQ/ERROR_NoTriggers.md
davidcok: Renamed it to pos

docs/HowToFAQ/ERROR_NoTriggers.md
davidcok: Yes - did that.

keyboardDrummer (approved):

PR #2673, opened 2022-09-01
Resolve unit test instability

keyboardDrummer:
### Changes

- Run the DafnyLanguageServer test two times, to make it less likely that we check in unstable code in the future.
- Stop using `ReplaySubject` in Compilation since its OnNext method calls observers asynchronously which is exactly what we don't want in `ResolveAsync` and `TranslateAsync` since then these Tasks will complete before `observer.LastPublishedState` has been updated, and that can cause `GetSnapshotAfterResolutionAsync` to return an `IdeState` with a version that's lower than `Compilation.ResolvedDocument.Version` !! which is bad. To compensate for the lack of replaying, Compilation now has a Start() method and won't publish updates until Start() is called, which `DocumentManager` does after it subscribes to `Compilation.DocumentUpdates`.
- Stop checking the `IImplementationTask.IsIdle` field for all tasks to determine if verification is finished, since it may become true before the observable returned by `IImplementationTask.TryRun` has completed and sent its last OnNext event, which would mean `Compilation.LastDocument` could return something that'll still mutate. Instead `Compilation` now maintains a counter to see how many verification tasks are still running.
- Give DocumentManager a `workCompletedForCurrentVersion` field to determine whether it has started all verification tasks it's planning to start. This prevents `DocumentManager.GetLastDocumentAsync` from returning if a verification task finishes before `DocumentManager` manages to queue another one.


Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -29,6 +29,7 @@ public class DafnyCounterExampleHandler : ICounterExampleHandler {
           }
 
           var state = await documentManager.GetIdeStateAfterVerificationAsync();
+          logger.LogWarning("counter-examples retrieved IDE state");
```
MikaelMayer, line 32: should you keep this warning log?

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -26,29 +26,37 @@ public class DocumentManager {
...
     DocumentTextBuffer document) {
     this.services = services;
+    this.logger = services.GetRequiredService<ILogger<DocumentManager>>();
     this.documentOptions = documentOptions;
     this.verifierOptions = verifierOptions;
     this.relocator = services.GetRequiredService<IRelocator>();
     this.textChangeProcessor = services.GetRequiredService<ITextChangeProcessor>();
 
-    observer = new DocumentObserver(services.GetRequiredService<ILogger<DocumentObserver>>(),
+    observer = new IdeStateObserver(services.GetRequiredService<ILogger<IdeStateObserver>>(),
```
MikaelMayer, line 59: Good idea to rename !

MikaelMayer (approved):
I don't have a sense of why it solves instabilities, but the code is subtantially the same, so if it's less unstable, that works for me !

cpitclaudel (approved):
Looks reasonable to me, thanks!

PR #2680, opened 2022-09-01
Issue #2616 - documenting < on datatypes

### LINKED ISSUE #2616: Document < on datatype values

davidcok:
Fixes #2616


docs/DafnyRef/Types.md
```diff
@@ -3654,6 +3654,31 @@ inductive datatype for trees may be updated as follows:
...
+  assert !(x < x);
+  assert z < x; // FAILS
+}
+```
+`x` is a datatype value that holds a `T` variant, which holds a `I` variant, which holds an integer `0`.
+The value `x.t` is a portion of the datatype structure denoted by `x`, so `x.t < x` is true.
+Datatype values are immutable mathematical values, so the value of `y` is identical to the value of
+`x.t`, so `y < x` is true also, even though `y` is constructed from the ground up, rather than as
+a portion of `x`. However, `z` is different than either `y` or `x.t` and consequently `z < x` is not provable.
+Furthermore, `<` does not include `==`, so `x ==x` is false.
```
RustanLeino, line 3676:
```suggestion
Furthermore, `<` does not include `==`, so `x < x` is false.
```

docs/DafnyRef/Types.md
```diff
@@ -3654,6 +3654,31 @@ inductive datatype for trees may be updated as follows:
...
 ```
 
+The operator `<` is defined for two operands of the same datataype.
+It means _is properly contained in_. For example, in the code
+```dafny
+datatype X = T(t: X) | I(i: int)
+method comp() {
+  var x: X := T(I(0));
+  var y: X := I(0);
+  var z: X := I(1);
```
RustanLeino, line 3664: I suggest leaving off the `: X` to reduce clutter. With the definitions of `T` and `I` just above it, I don't think there's much chance of confusion.

RustanLeino (approved):

PR #2688, opened 2022-09-02
feat: Python iterators

fabiomadge:


fabiomadge:
```Python
class Iter:
    def  __init__(self, X):
        self.X = X
        self.a: _dafny.Array = _dafny.Array.empty(1)
        self.stop: TypeVar('X') = self.X()
        self.x: TypeVar('X') = self.X()
        self._iter = None
        pass

    def __dafnystr__(self) -> str:
        return "Iter"
    def ctor__(self, a, stop):
        self.a = a
        self.stop = stop
        self._iter = self.TheIterator()
    def MoveNext(self):
        try:
            next(self._iter)
            return True
        except StopIteration:
            return False
    def TheIterator(self):
        _1_i_: int
        _1_i_ = 0
        with _dafny.label("0"):
            while (_1_i_) < (len(((self).a))):
                with _dafny.c_label("0"):
                    if (_dafny.euclidian_modulus(_1_i_, 2)) == (0):
                        (self).x = ((self).a)[_1_i_]
                        yield
                    (self).x = ((self).a)[_1_i_]
                    if ((self).x) == ((self).stop):
                        raise _dafny.Break("0")
                    if (_dafny.euclidian_modulus(_1_i_, 2)) == (1):
                        yield
                    _1_i_ = (_1_i_) + (1)
```

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -205,7 +205,43 @@ public class PythonCompiler : SinglePassCompiler {
     }
 
     protected override ConcreteSyntaxTree CreateIterator(IteratorDecl iter, ConcreteSyntaxTree wr) {
-      throw new UnsupportedFeatureException(Token.NoToken, Feature.Iterators);
+      var cw = CreateClass(IdProtect(iter.EnclosingModuleDefinition.CompileName), IdName(iter), false,
+        IdName(iter), iter.TypeArgs, iter, null, iter.tok, wr) as ClassWriter;
+      Debug.Assert(cw != null, nameof(cw) + " != null");
```
cpitclaudel, line 210:
```suggestion
      Contracts.Assert(cw != null, nameof(cw) + " != null");
```

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -205,7 +205,43 @@ public class PythonCompiler : SinglePassCompiler {
...
+      var constructorWriter = cw.ConstructorWriter;
+      var w = cw.MethodWriter;
+      // here come the fields
+      Constructor ct = null;
+      foreach (var member in iter.Members) {
+        switch (member) {
+          case Field { IsGhost: false } f:
+            DeclareField(IdName(f), false, false, f.Type, f.tok, PlaceboValue(f.Type, constructorWriter, f.tok, true), constructorWriter);
+            break;
+          case Constructor constructor:
```
cpitclaudel, line 220: I think Rustan has a PR for ghost constructors, let's make sure that it doesn't conflict with this?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -205,7 +205,43 @@ public class PythonCompiler : SinglePassCompiler {
...
+        wCtor.WriteLine("self.{0} = {0}", IdName(p));
+      }
+      wCtor.WriteLine("self._iter = self.TheIterator()");
+
+      var wMoveNext = w.NewBlockPy("def MoveNext(self):");
+      wMoveNext.NewBlockPy("try:")
+        .WriteLine("next(self._iter)")
+        .WriteLine("return True");
+      wMoveNext.NewBlockPy("except (StopIteration, TypeError) as e:")
+        .WriteLine("return False");
```
cpitclaudel, line 242: Could this be put in a base class that iterators inherit from?

cpitclaudel (changes requested):

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: Yes, but it's not quite as easy as I would like it to be. The `CreateClass` wants the superclasses as `Type`s. There's also the argument to be made the the expected number of iterators per program is likely to be <1.

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: That one is only concerned with `datatype`s.

cpitclaudel (approved):

PR #2696, opened 2022-09-05
Almost fully support goto definition

keyboardDrummer:
### Changes

Add much more complete support for "Go to definition" in the LSP server. 

### Caveats

Go to definition within patterns of match constructs is not working properly, due to resolution currently happening _after_ those patterns have been rewritten. A PR is in the works to resolve this and provide go to definition there as well. I will leave updating the release notes for then.

The existing support for this feature relies on files in the directory `Source/DafnyLanguageServer/Language/Symbols`, where a mirror of the Dafny AST is constructed that seems to do its own name resolution. This PR implements go to definition more cheaply, by exploiting the existing AST and gathering resolved reference information from there. However, IMO this approach is still too expensive and should be replaced in the future. Ideally, the Dafny AST node types each indicate what references they have in a declarative fashion, and then the Dafny compiler is able to extract constraints from this which it uses to resolve the reference, and can use to record the resolved references. Implementing "go to definition" is then free, while in this PR there's still code for gathering the resolved references from the AST. 

This free approach requires re-architecting Resolver.cs and since we have ongoing changes to it, it will defer that change until later.

### Testing

Several go to definition cases have been added as automatic tests.

The PR has partly been manually tested by taking a large Dafny codebase and trying to resolve all references there. I am positive there are still edge case Dafny references for which "Go to definition" does not work, but this PR already provides a significant improvement in "Go to definition" behavior.

One large advantage of the free "go to definition" approach described above is that "Go to definition" will work for any reference that can be resolved, since there's no "Go to definition" specific code. With that approach, we won't need an extensive test suite for go to definition.


Scripts/code.sh
```diff
@@ -0,0 +1,4 @@
+# This script starts VSCode so that it uses the Dafny language server from this repository.
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
```
MikaelMayer, line 2: Long awaited script, thanks for committing it!

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -22,7 +22,26 @@ namespace Microsoft.Dafny {
   [System.AttributeUsage(System.AttributeTargets.Field)]
   public class FilledInDuringResolutionAttribute : System.Attribute { }
 
-  public class Program {
+  public interface IHasUsages : IDeclarationOrUsage {
```
MikaelMayer, line 25: Could you please move IHasUsage after IDeclarationOrUsage, and IDeclarationOrUsage after Node? It makes more sense to me to describe something linearly if possible, starting at the top ancestors

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -452,6 +473,9 @@ public abstract class IVariableContracts : IVariable {
       Contract.Ensures(Contract.Result<string>() != null);
       throw new NotImplementedException();
     }
+
+    public IEnumerable<INode> Children => throw new NotImplementedException();
+    public IToken NameToken => throw new NotImplementedException();
```
MikaelMayer, line 478:
```suggestion
    public IEnumerable<INode> Children => throw new NotImplementedException();
    public IToken NameToken => throw new NotImplementedException();
```
```suggestion
    public abstract IEnumerable<INode> Children { get; }
    public abstract IToken NameToken { get; }
```

Source/DafnyCore/AST/Expressions.cs
```diff
@@ -808,6 +808,8 @@ public abstract class Expression {
     var le = this as StringLiteralExpr;
     return le == null ? null : le.Value as string;
   }
+
+  public virtual IEnumerable<INode> Children => SubExpressions;
```
MikaelMayer, line 812: In my formatting PR, I am adding a field PreResolveSubExpressions for every ConcreteExpression (the expression that support rewriting), because I'm doing the formatting based on that tree and not the resolved one.
I had to duplicate several SubExpressions because they were slightly different.

Are your Children pre-resolve sub expressions? If so I'd suggest you rename `Children` explicitly `PreResolveChildren` or something like that, so that it's unambiguous which nodes you will be referring to.
And that will help me in my PR too.

Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs
```diff
@@ -28,20 +25,265 @@ public class DefinitionTest : ClientBasedLanguageServerTest {
...
+    }
+    x := x + 1;
+  }
 }
+".TrimStart();
+
+      var documentItem = CreateTestDocument(source);
+      await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+      var whileReference = (await RequestDefinition(documentItem, (4, 8)).AsTask()).Single();
+      Assert.AreEqual(new Range((2, 2), (2, 7)), whileReference.Location!.Range);
```
MikaelMayer, line 44: Thank you for adding these tests! I will surely make go to definition more robust in the long term.

I have another idea. Although it is more work for you, I think it will make it easier to read and maintain those tests.  Let me know if you like it.

Let's say you define the following

```cs
var sourceAndRequests = @"
method HasLoop() {
  var x := 1;
//^^^^^=1
  while(true) {
    if (x > 2) {
//      ^?1
      break;
    }
    x := x + 1;
  }
}
".TrimStart();
```

Then you could extract the source code, the request, and the expected answer directly from the same source. Of course you have to fix the line numbers when extracting everything, but to alleviate this concern, you can borrow some of the code for the hover tests which define tests this way:

https://github.com/dafny-lang/dafny/blob/62a6d582fbed84e405952483accbf095933c1acb/Source/DafnyLanguageServer.Test/Lookup/HoverTest.cs#L78-L85

and extract it like this:

https://github.com/dafny-lang/dafny/blob/62a6d582fbed84e405952483accbf095933c1acb/Source/DafnyLanguageServer.Test/Lookup/HoverTest.cs#L58-L72

I know it's a bit challenging and you might have other tasks to do though.

Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs
```diff
@@ -28,20 +25,265 @@ public class DefinitionTest : ClientBasedLanguageServerTest {
...
+
+      var getItCall = (await RequestDefinition(documentItem, (9, 23)).AsTask()).Single();
+      Assert.AreEqual(new Range((1, 9), (1, 14)), getItCall.Location!.Range);
+
+      var doItCall = (await RequestDefinition(documentItem, (10, 12)).AsTask()).Single();
+      Assert.AreEqual(new Range((4, 9), (4, 13)), doItCall.Location!.Range);
+
+      var xVar = (await RequestDefinition(documentItem, (10, 17)).AsTask()).Single();
+      Assert.AreEqual(new Range((9, 6), (9, 7)), xVar.Location!.Range);
     }
```
MikaelMayer, line 287: Is there a way to jump to another file? If so, can we test this?

Source/DafnyServer/DafnyHelper.cs
```diff
@@ -89,7 +89,7 @@ class DafnyHelper {
     public void Symbols() {
       ServerUtils.ApplyArgs(args, DafnyOptions.O);
       if (Parse() && Resolve()) {
-        var symbolTable = new SymbolTable(dafnyProgram);
+        var symbolTable = new LegacySymbolTable(dafnyProgram);
```
MikaelMayer, line 92: Good renaming. I like very much your new symbol table implementation.

Test/dafny0/EqualityTypes.dfy.expect
```diff
@@ -56,12 +56,12 @@ EqualityTypes.dfy(313,19): Error: type parameter 1 (X) passed to function H must
 EqualityTypes.dfy(315,19): Error: type parameter 0 (V) passed to type AClass must support equality (got Stream<int>)
 EqualityTypes.dfy(316,19): Error: type parameter 1 (X) passed to function H must support equality (got Stream<real>)
 EqualityTypes.dfy(319,31): Error: type parameter 1 (X) passed to function H must support equality (got Stream<real>)
-EqualityTypes.dfy(321,41): Error: type parameter 0 (V) passed to type AClass must support equality (got Stream<real>)
-EqualityTypes.dfy(323,28): Error: type parameter 0 (V) passed to type AClass must support equality (got Stream<int>)
+EqualityTypes.dfy(321,15): Error: type parameter 0 (V) passed to type AClass must support equality (got Stream<real>)
```
MikaelMayer, line 59: How come there was an error before? Good job fixing it.
I would have preferred that it points to position 22 instead (the first type parameter), but I guess pointing to the type constructor is already an improvement compared to pointing to the second type parameter.

Test/dafny0/UserSpecifiedTypeParameters.dfy.expect
```diff
@@ -1,6 +1,6 @@
 UserSpecifiedTypeParameters.dfy(26,10): Error: the type of this variable is underspecified
-UserSpecifiedTypeParameters.dfy(26,24): Error: type of type parameter could not be determined; please specify the type explicitly
-UserSpecifiedTypeParameters.dfy(26,24): Error: type of type parameter could not be determined; please specify the type explicitly
+UserSpecifiedTypeParameters.dfy(26,16): Error: type of type parameter could not be determined; please specify the type explicitly
```
MikaelMayer, line 2: Here the position 24 was more appropriate, especially for a quick fix. Could you please revert that?

Test/git-issues/git-issue-968.dfy.expect
```diff
@@ -1,2 +1,2 @@
-git-issue-968.dfy(11,14): Error: type of type parameter could not be determined; please specify the type explicitly
+git-issue-968.dfy(11,11): Error: type of type parameter could not be determined; please specify the type explicitly
```
MikaelMayer, line 1: Here as well I prefer the position of the error to be on the "." where the type parameter should be inserted.

MikaelMayer (changes requested):
Wonderful, I'm looking forward to being able to jump to definition soon. That's going to be sooo useful.
The only blocking change request is the position in the test files. I think the previous position was better in terms of error reporting.

Source/DafnyCore/AST/Expressions.cs
keyboardDrummer: They're not pre-resolve, they're resolved.

Test/git-issues/git-issue-968.dfy.expect
keyboardDrummer: It was after the `.` not before the `.`, where the type parameter is expected.

Test/dafny0/UserSpecifiedTypeParameters.dfy.expect
keyboardDrummer: 24 points to `H`, while 16 points to `MyClass`. The error is about missing type argument for `MyClass`, so why do you prefer the previous message?

Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs
keyboardDrummer: It is tested in `DefinitionOfFunctionInvocationOfFunctionDeclaredInForeignDocumentReturnsLocation`

MikaelMayer: I'm sorry, I cannot find this function here. Did it exist in the past? If so, last time I checked, it's not working on my machine.

Source/DafnyCore/AST/Expressions.cs
```diff
@@ -3942,6 +4022,7 @@ public DatatypeUpdateExpr(IToken tok, Expression root, List<Tuple<IToken, string
 /// (Ironically, AutoGeneratedExpression, which is like the antithesis of concrete syntax, inherits from ConcreteSyntaxExpression, which perhaps
 /// should rather have been called SemanticsNeutralExpressionWrapper.)
 /// </summary>
+// TODO replace this with AutoGeneratedToken.
```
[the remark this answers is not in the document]
keyboardDrummer: I'll remove this, but it seems the same concept is implemented in two different ways.

Source/DafnyCore/AST/Tokens.cs
```diff
@@ -55,6 +56,7 @@ public record Token : IToken {
 
   public int kind { get; set; } // Used by coco, so we can't rename it to Kind
 
+  public string ActualFilename => Filename;
```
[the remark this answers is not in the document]
keyboardDrummer: `Filename` something contains some extra information that might be useful for a user but causes it to not be an actual filename.

Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs
keyboardDrummer: It's here: https://github.com/dafny-lang/dafny/blob/a2b69626b9d17a648e6c494ce17a5f8fa31bbc51/Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs#L313

Test/dafny0/EqualityTypes.dfy.expect
keyboardDrummer: > I would have preferred that it points to position 22 instead (the first type parameter)

Would be better yes. Note that it was pointing to `H` before, which is well beyond the second type parameter.

Test/dafny0/UserSpecifiedTypeParameters.dfy.expect
MikaelMayer: Oh you're right. I forgot that characters positions start at 0, so I thought I was pointing to the dot, which is exactly the place where to insert type parameters. In this case, this fix is better.

Source/DafnyCore/AST/Tokens.cs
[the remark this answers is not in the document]
MikaelMayer: My question is, where is a case where ActualFilename is not equal to Filename? Is it overridden somewhere?

Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs
MikaelMayer: > However, I'd prefer making such a change in a separate PR. I can do it as a follow-up to this one and use it in the DefinitionTest.cs file.

Yes I would really appreciate a follow-up PR for that.

MikaelMayer (approved):
This PR is good to merge ! Thanks.

Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs
keyboardDrummer: It's good that you call it out :-) I've put it on my list, thanks.

PR #2700, opened 2022-09-06
New landing page

davidcok:


robin-aws (changes requested):
If I'm following correctly, it looks like we want to align on a single landing page at the top-level https://dafny.org URL, which already exists and will be updated in https://github.com/dafny-lang/dafny-lang.github.io/pull/1. But we also have https://dafny.org/dafny/, which this PR is replacing with the identical content.

It makes sense that we don't want to break the existing second URL. But let's replace it with a client-side redirect to https://dafny.org here instead, so we don't have the content duplicated - even if that's temporary it could be very confusing, and having content duplicated across two different repositories is especially dangerous. https://stackoverflow.com/questions/9276817/301-redirect-for-site-hosted-at-github discusses several options.

robin-aws:
> 



> If I'm following correctly, it looks like we want to align on a single landing page at the top-level https://dafny.org URL, which already exists and will be updated in [dafny-lang/dafny-lang.github.io#1](https://github.com/dafny-lang/dafny-lang.github.io/pull/1). But we also have https://dafny.org/dafny/, which this PR is replacing with the identical content.
> 
> It makes sense that we don't want to break the existing second URL. But let's replace it with a client-side redirect to https://dafny.org here instead, so we don't have the content duplicated - even if that's temporary it could be very confusing, and having content duplicated across two different repositories is especially dangerous. https://stackoverflow.com/questions/9276817/301-redirect-for-site-hosted-at-github discusses several options.

After offline conversation, it's now clear that the intention is for https://dafny.org to be a snapshot of https://dafny.org/dafny/. So the content is identical for now but the latter will change on every commit to `dafny-lang/dafny`, whereas the former will only be updated as part of the release process. I'll comment on both PRs accordingly.

robin-aws:
This LGTM assuming that it stays as an otherwise identical copy of https://github.com/dafny-lang/dafny-lang.github.io/pull/1, so let's merge that first and then I'll approve this once it's updated with any last minute changes.

robin-aws:
One final thought: I don't think `dafny.org/dafny` is the ideal path for "latest documentation for Dafny corresponding to the latest unreleased changes", I'd probably prefer something like `dafny.org/docs/development` or something. But I get that we also want to replace the existing `dafny.org/dafny` page so we can deal with that later, at which point I might prefer making that path just redirect to `dafny.org`.

davidcok:
Agreed on the dafny.org/dafny -- changing that is part of the plan to incrementally improve all this, without breaking things in the process

robin-aws (approved):

PR #2701, opened 2022-09-07
feat: Full `array` support in Python

### LINKED ISSUE #2697: Full `array` support for Python

fabiomadge:
Fixes #2697.

### DIFF
--- a/Source/Dafny/Compilers/Compiler-python.cs
+++ b/Source/Dafny/Compilers/Compiler-python.cs
@@ -646,7 +646,7 @@ namespace Microsoft.Dafny.Compilers {
                     }
                     if (td is NonNullTypeDecl decl) {
                       if (decl.Class is ArrayClassDecl arr) {
-                        return $"{DafnyArrayClass}.empty({arr.Dims})";
+                        return $"{DafnyArrayClass}(None, {arr.Dims})";
                       }
                       return "None";
                     }
@@ -896,19 +896,12 @@ namespace Microsoft.Dafny.Compilers {
     protected override void EmitNewArray(Type elmtType, IToken tok, List<Expression> dimensions, bool mustInitialize,
         ConcreteSyntaxTree wr, ConcreteSyntaxTree wStmts) {
       var initValue = mustInitialize ? DefaultValue(elmtType, wr, tok, true) : "None";
-      if (dimensions.Count == 1) {
-        // handle the common case of 1-dimensional arrays separately
-        wr.Write($"{DafnyArrayClass}([{initValue} for _ in range");
-        TrParenExpr(dimensions[0], wr, false, wStmts);
-        wr.Write("])");
-      } else {
-        wr.Write($"{DafnyRuntimeModule}.newArray({initValue}");
-        foreach (var dim in dimensions) {
-          wr.Write(", ");
-          TrExpr(dim, wr, false, wStmts);
-        }
-        wr.Write(")");
+      wr.Write($"{DafnyArrayClass}({initValue}");
+      foreach (var dim in dimensions) {
+        wr.Write(", ");
+        TrExpr(dim, wr, false, wStmts);
       }
+      wr.Write(")");
     }
 
     protected override void EmitLiteralExpr(ConcreteSyntaxTree wr, LiteralExpr e) {
@@ -1103,11 +1096,8 @@ namespace Microsoft.Dafny.Compilers {
           break;
         case SpecialField.ID.ArrayLength:
         case SpecialField.ID.ArrayLengthInt:
-          preString = "len(";
-          postString = ")";
-          if (idParam != null && (int)idParam > 0) {
-            postString = string.Concat(Enumerable.Repeat("[0]", (int)idParam)) + postString;
-          }
+          var dim = idParam is int d and > 0 ? d : 0;
+          compiledName = $"length({dim})";
           break;
         default:
           Contract.Assert(false); // unexpected ID
@@ -1187,11 +1177,9 @@ namespace Microsoft.Dafny.Compilers {
     }
 
     protected override ConcreteSyntaxTree EmitArraySelect(List<string> indices, Type elmtType, ConcreteSyntaxTree wr) {
-      Contract.Assert(indices != null && 1 <= indices.Count);
+      Contract.Assert(indices is { Count: >= 1 });
       var w = wr.Fork();
-      foreach (var index in indices) {
-        wr.Write($"[{index}]");
-      }
+      wr.Write($"[{indices.Comma()}]");
       return w;
     }
 
--- a/Source/Dafny/Compilers/Compiler-python.cs
+++ b/Source/Dafny/Compilers/Compiler-python.cs
@@ -648,7 +648,7 @@ namespace Microsoft.Dafny.Compilers {
                     }
                     if (td is NonNullTypeDecl decl) {
                       if (decl.Class is ArrayClassDecl arr) {
-                        return $"{DafnyArrayClass}(None, {arr.Dims})";
+                        return $"{DafnyArrayClass}(None, {Enumerable.Repeat("0", arr.Dims).Comma()})";
                       }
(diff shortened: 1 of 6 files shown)


Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -898,19 +898,12 @@ protected class ClassWriter : IClassWriter {
...
-        wr.Write($"{DafnyRuntimeModule}.newArray({initValue}");
-        foreach (var dim in dimensions) {
-          wr.Write(", ");
-          TrExpr(dim, wr, false, wStmts);
-        }
-        wr.Write(")");
+      wr.Write($"{DafnyArrayClass}({initValue}");
+      foreach (var dim in dimensions) {
+        wr.Write(", ");
+        TrExpr(dim, wr, false, wStmts);
```
cpitclaudel, line 904: I imagine this would have come up somewhere else, but how do we guarantee that this TrExpr has enough parentheses to not capture the commas around it?

Source/Dafny/Compilers/Compiler-python.cs
```diff
@@ -1105,11 +1098,8 @@ protected class ClassWriter : IClassWriter {
           break;
         case SpecialField.ID.ArrayLength:
         case SpecialField.ID.ArrayLengthInt:
-          preString = "len(";
-          postString = ")";
-          if (idParam != null && (int)idParam > 0) {
-            postString = string.Concat(Enumerable.Repeat("[0]", (int)idParam)) + postString;
-          }
+          var dim = idParam is int d and > 0 ? d : 0;
+          compiledName = $"length({dim})";
```
cpitclaudel, line 1102: Shouldn't this be `self.length`?
```suggestion
          compiledName = $"self.length({dim})";
```

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -107,22 +108,62 @@ def __lt__(self, other):
...
+        def __dafnystr__(self) -> str:
+            return '[' + ', '.join(map(string_of, self)) + ']'
+
+    def __init__(self, initValue, *dims):
+        fresh = not isinstance(initValue, Array.Box)
+        self.arr = initValue if fresh else initValue.content
+        self.dims = list(dims)
+        if fresh:
+            for i in reversed(self.dims):
+                self.arr = Array.Box([copy.deepcopy(self.arr) for _ in range(i)])
```
cpitclaudel, line 122: I'm not sure the `deepcopy` is sound: won't it clone the elements of the array as well?

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -107,22 +108,62 @@ def __lt__(self, other):
...
+            return self.arr[key]
+        arr = self.arr
+        dims = self.dims
+        if not isinstance(key, Iterable):
+            key = (key,)
+        elif len(dims) < len(key):
+            return None
+        for i in key:
+            if dims[0] == 0:
+                return None
```
cpitclaudel, line 147: I think it's better to raise an exception here

cpitclaudel: Or even better, just remove this and the reassignment of dims below

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -107,22 +108,62 @@ def __lt__(self, other):
...
+            arr = arr[i]
+            dims = dims[1:]
```
cpitclaudel, line 149: I think this line can be removed…

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -107,22 +108,62 @@ def __lt__(self, other):
...
+        return arr if not dims else Array(arr, dims)
```
cpitclaudel, line 150: … as long as this line checks `len(key)` against `len(dims)`

cpitclaudel: But, does the `else` branch ever get used by the compiler?  If not, maybe we can remove that complexity

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -107,22 +108,62 @@ def __lt__(self, other):
...
+        if not isinstance(key, Iterable):
+            self.arr[key] = value
+            return
+        if len(self.dims) < len(key):
+            return
+        arr = self.arr
+        dims = self.dims
+        (*init, last) = key
+        for i in init:
+            if dims[0] == 0:
```
cpitclaudel, line 162: Same here, I'd say no need to check here…

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -107,22 +108,62 @@ def __lt__(self, other):
...
+                return
+            arr = arr[i]
+            dims = dims[1:]
```
cpitclaudel, line 165: … which means we can skip this line, which is a nice speed bonus

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -107,22 +108,62 @@ def __lt__(self, other):
...
+
+    def __setitem__(self, key, value):
+        if not isinstance(key, Iterable):
+            self.arr[key] = value
+            return
+        if len(self.dims) < len(key):
+            return
+        arr = self.arr
+        dims = self.dims
+        (*init, last) = key
```
cpitclaudel, line 160: I'd be wary of too much allocation in a "hot" method like this

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -107,22 +108,62 @@ def __lt__(self, other):
...
+            return None
+        for i in key:
+            if dims[0] == 0:
+                return None
+            arr = arr[i]
+            dims = dims[1:]
+        return arr if not dims else Array(arr, dims)
+
+    def __setitem__(self, key, value):
+        if not isinstance(key, Iterable):
```
cpitclaudel, line 153: This bit is not consistent with the prelude of `__getitem__` above

cpitclaudel (changes requested):

Source/Dafny/Compilers/Compiler-python.cs
fabiomadge: I can't come up with something sufficiently ambiguous that we would actually generate.

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge: I agree, we don't need this here, but I'm pretty sure it doesn't harm us either, as initialisations of reference types seem to be a two step process. First init to null and then override in a loop.

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -134,36 +132,25 @@ def __len__(self):
         return self.length(0)
 
     def __getitem__(self, key):
-        if isinstance(key, slice):
+        if isinstance(key, slice) or not isinstance(key, Iterable):
```
cpitclaudel, line 135:
```suggestion
        if not isinstance(key, Iterable):
```

I believe these two are equivalent

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -134,36 +132,25 @@ def __len__(self):
...
             return
-        if len(self.dims) < len(key):
-            return
         arr = self.arr
-        dims = self.dims
-        (*init, last) = key
-        for i in init:
-            if dims[0] == 0:
+        j = len(key)
+        for i in key:
```
cpitclaudel, line 148: Alternatively, you could use `for idx, i in enumerate(key)`

cpitclaudel (approved):
LGTM, I suggested two small changes

cpitclaudel (approved):
🏃🐍 :shipit:

PR #2709, opened 2022-09-07
Remove `UnsupportedFeatureException`s in Python

fabiomadge:
Removes dead code for compiling constant member variables (not supported) and adds support for named functions for sequence construction.


robin-aws (approved):
Awesome, LGTM!

I was confused at first about why you were able to remove the `isStatic` parameter until I remembered the language doesn't support static fields at all. For the record that tidbit would have been great to have in the PR description to help guide reviewers. :)

PR #2712, opened 2022-09-08
Add plugin to create wrapper functions/methods to check contracts at runtime

### LINKED ISSUE #2235: feat: Add resolver plugin that automatically creates test wrappers for contracts
The Dafny reference manual, in the [documentation for the `expect` statement](https://github.com/dafny-lang/dafny/blob/master/docs/DafnyRef/Statements.md#1918-expect-statement-sec-expect-statement), suggests that it can be used as a way to dynamically check the contract of a method.

However, doing this manually in Dafny as-is introduces the possibility that the `expect` statement and the `ensures` clause could get out of sync. Instead, we could have a plugin that would automatically generate a wrapper to check the `ensures` clauses (and possibly even `requires` clauses) of a method using `expect`, guaranteeing that the dynamic checks would always match the stated contracts.

Further comments on this issue will elaborate a proposed design.

atomb:
This PR adds a flag to allow certain function and method contracts to be tested at runtime. For the moment, it focuses on `{:extern}` code, but could be applied to other code in the future. Some details of the design are described in the comment thread of issue #2235.

It is usable with the following:
```
/testContracts:<AllExterns|ExternsInTests>
    Enable run-time testing of compiled function or method contracts in
    certain situations, currently focused on :extern code.

    AllExterns - Check contracts on every call to a function or
        method marked :extern, regardless of where it occurs.
    Externs - Check contracts on every call to a function or method
        marked :extern when it occurs in one marked :test.
```

I'm not entirely happy that it required removing a few `readonly` qualifiers from AST fields, but that seems unavoidable without major rewriter infrastructure refactoring.

Fixes #2235.


atomb:
The following Dafny and C# files, together, demonstrate the basic operation of this for checking calls to externs from tests.

The Dafny code (`check-extern.dfy`):

```dafny
method {:extern} Foo(x: int) returns (y: int)
  requires 1 < x < 10;
  ensures y >= 10;

method FooCaller(x: int) returns (y: int)
  requires 1 < x < 10;
  ensures y >= 10;
{
  y := Foo(x);
}

method {:test} FooTest()
{
  var y := Foo(3);
  expect y == 30;
}

function method {:extern} Bar(x: int): (y: int)
  requires 1 < x < 10;
  ensures y >= 10;

function method BarCaller(x: int): (y: int)
  requires 1 < x < 10;
  ensures y >= 10;
{
  Bar(x)
}

method {:test} BarTest()
{
  var y := Bar(3);
  expect y == 30;
}

```

And the C# code (`FooBar.cs`):

```csharp
using System.Numerics;

namespace _module {

  public partial class __default {
    public static BigInteger Foo(BigInteger x)
    {
      return BigInteger.Zero;
    }
    public static BigInteger Bar(BigInteger x)
    {
      return BigInteger.Zero;
    }
  }
}
```

These can be run together with the following command:
```sh
dafny /compile:3 /runAllTests:1 /testContracts:Tests /useBaseNameForFileName check-extern.dfy FooBar.cs
```

This yields the following:
```
Wrote textual form of target program to check-extern.cs
Running...

FooTest: FAILED
	check-extern.dfy(3,12): Runtime failure of ensures clause from check-extern.dfy:3:13
BarTest: FAILED
	check-extern.dfy(20,12): Runtime failure of ensures clause from check-extern.dfy:20:13
Test failures occurred: see above.
```

atomb:
I think this PR is largely complete. The one key open question in my mind is whether the current set of `/testContracts` parameters is the right one. It could be that `ExternsInTests` is too specific to be very useful (compiling only your test code with `/AllExterns` could be the more common use case. And it could be than an `All` option, to always test all (compilable) contracts  would be valuable. I'm pretty convinced that `/AllExterns` is a useful mode, however, and perhaps the only one we need to support in this PR.

MikaelMayer:
I'm ready to review it ! Let me have a look.

Source/DafnyCore/Resolver/ExpectContracts.cs
```diff
@@ -0,0 +1,291 @@
...
+  /// <param name="callStmt">The call statement to include.</param>
+  /// <returns>The newly-created block statement.</returns>
+  private BlockStmt MakeContractCheckingBody(IEnumerable<AttributedExpression> requires,
+    IEnumerable<AttributedExpression> ensures, Statement callStmt) {
+    var expectRequiresStmts = requires.Select(req =>
+      CreateContractExpectStatement(req, "requires"));
+    var expectEnsuresStmts = ensures.Select(ens =>
+      CreateContractExpectStatement(ens, "ensures"));
+    var callStmtList = new List<Statement>() { callStmt };
+    var bodyStatements = expectRequiresStmts.Concat(callStmtList).Concat(expectEnsuresStmts);
```
MikaelMayer, line 64: That's interesting. So your approach is to rewrite any extern method call to inline the requires call and the expects clause.
Do you need to check the requires? Isn't that going to be checked by Dafny statically?
For the ensures, it's great, because it means one can write a test fuzzer without any expect, and the method's postcondition will still be checked. Could you please add this idea in documentation of /testContracts?

Source/DafnyCore/Resolver/ExpectContracts.cs
```diff
@@ -0,0 +1,291 @@
...
+      var callStmt = origFunc.Result?.Name is null
+        ? (Statement)new VarDeclStmt(tok, tok, locs, new UpdateStmt(tok, tok, lhss, rhss))
+        : (Statement)new UpdateStmt(tok, tok, lhss, rhss);
+
+      var body = MakeContractCheckingBody(origFunc.Req, origFunc.Ens, callStmt);
+
+      if (origFunc.Result?.Name is null) {
+        body.AppendStmt(new ReturnStmt(tok, tok, new List<AssignmentRhs> { new ExprRhs(localExpr) }));
+      }
+      newFunc.ByMethodBody = body;
```
MikaelMayer, line 133: Clever ! You transform the ByMethod body of an extern function with these expect statements. Love it.

Source/DafnyCore/Resolver/ExpectContracts.cs
```diff
@@ -0,0 +1,291 @@
...
+      }
+    }
+
+    // Generate a wrapper for each of the members identified above.
+    foreach (var (topLevelDecl, decl) in membersToWrap) {
+      GenerateWrapper(topLevelDecl, decl);
+    }
+  }
+
+  private class CallRedirector : TopDownVisitor<MemberDecl> {
```
MikaelMayer, line 173: Can you please add a comment about this class?

MikaelMayer (changes requested):
Great job ! This will be very valuable. Would it be possible that /runAllTests automatically triggers ExternsInTest to be set by default? This would not completely be a breaking change, as you'd only issue warnings.

Also I suggested a few doc upgrades

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -1425,6 +1444,16 @@ verification outcome.
     option is useful in a diamond dependency situation, to prevent code
     from the bottom dependency from being generated more than once.
 
+/testContracts:<AllExterns|ExternsInTests>
```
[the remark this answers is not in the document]
atomb: Yeah, I like that naming. I'll switch to it.

Source/DafnyCore/Resolver.cs
```diff
@@ -546,6 +550,11 @@ enum ValuetypeVariety {
             // Now we're ready to resolve the cloned module definition, using the compile signature
 
             ResolveModuleDefinition(nw, compileSig);
+
+            foreach (var rewriter in rewriters) {
+              rewriter.PostCompileCloneAndResolve(nw);
```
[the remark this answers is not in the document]
atomb: It took an in-depth debugging session with @RustanLeino to understand that that was necessary! It's likely to be useful for other rewriter plugins that operate (even partly) post-resolution and need to make sure their effects persist in the compiled code.

Source/DafnyCore/Resolver/ExpectContracts.cs
atomb: For externs (of the form that this is focused on, where the implementation is written in the target language), that's true. Dafny will prove that the requires clause is satisfied. However, if you were to apply a fuzzer to the target language code, the existence of a branch on a requires clause is likely to make it easier to come up with inputs that satisfy that requires clause. Arguably, perhaps, it might make sense to check all of the requires clauses together in one `expect` statement, but check the ensures clauses separately, as the current code does.

I think any changes like that would be driven by how well it works in practice, though.

I'll add some reference manual documentation of `/testContracts` that goes into this.

MikaelMayer: > However, if you were to apply a fuzzer to the target language code, the existence of a branch on a requires clause is likely to make it easier to come up with inputs that satisfy that requires clause.

If you were to apply a fuzzer to the target language code, should you not gracefully exit if the requires is not satisfied?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -99,7 +99,7 @@ public enum PrintModes {
     public enum ContractTestingMode {
       None,
       AllExterns,
```
MikaelMayer, line 101: Could you please rename this so that it matches the option?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -1444,15 +1444,16 @@ verification outcome.
     option is useful in a diamond dependency situation, to prevent code
     from the bottom dependency from being generated more than once.
 
-/testContracts:<AllExterns|ExternsInTests>
+/testContracts:<Externs|TestedExterns>
     Enable run-time testing of compiled function or method contracts in
```
MikaelMayer, line 1448: I just realize one important thing. You are not testing "compiled functions", you are testing "compilable (function or method) contracts". Could you please rephrase it like `extern function or method contracts that can be compiled, at the call site`? I think it's important to mention the call site, because, until this comment, I thought you were wrapping the functions themselves, but you are actually rewriting every call to that function or method.

Source/DafnyCore/Resolver/ExpectContracts.cs
atomb: Ideally, yes, though I think that the current structure would at least give you enough to get started. If you run a fuzzer, and filter out the cases of failing requires clauses, you'll get a useful outcome.

In the long run, I'd like this code to integrate with target language fuzzing/verification APIs, using `assume`-like things for `requires` clauses and `assert`-like things for `ensures` clauses. I think that's a bit down the road, though.

Source/DafnyCore/DafnyOptions.cs
atomb: Ah, yeah, my original intent of the word "compiled" was just to say that this only applied to functions and methods that are being compiled. I think your use of the word is better, though! So I'll change the text accordingly.

docs/DafnyRef/UserGuide.md
```diff
@@ -1668,3 +1668,27 @@ periods.
   compiled assembly rather than including `DafnyRuntime.cs` in the build
   process.
 
+
+* `-testContracts:<mode>` - test certain function and method contracts
```
davidcok, line 1672: "test certain ... contracts" -- which ones? all non-ghost ones? all non-ghost ones that are executable?

atomb: Ah, I updated the help text with what I think is better wording and forgot to update the RM! What do you think of the wording in [`DafnyOptions.cs`](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/DafnyOptions.cs#L1323-L1333)?

PR #2715, opened 2022-09-08
Formatter pass

MikaelMayer:
This PR used the formatter being developed in https://github.com/dafny-lang/dafny/pull/2399 to format all the dafny files

### DIFF
--- a/Source/DafnyCompilers/AST.dfy
+++ b/Source/DafnyCompilers/AST.dfy
@@ -1,350 +1,350 @@
-include "CSharpDafnyASTModel.dfy"
+  include "CSharpDafnyASTModel.dfy"
 include "Library.dfy"
 
 module DafnyCompilerCommon.AST {
-    import Lib.Math
-    import Lib.Seq
-    import Microsoft.Boogie
+  import Lib.Math
+  import Lib.Seq
+  import Microsoft.Boogie
+  import C = CSharpDafnyASTModel
+
+  module Types {
     import C = CSharpDafnyASTModel
 
-    module Types {
-      import C = CSharpDafnyASTModel
+    type Path = seq<string>
 
-      type Path = seq<string>
+    datatype ClassType = ClassType(className: Path, typeArgs: seq<Type>)
 
-      datatype ClassType = ClassType(className: Path, typeArgs: seq<Type>)
+    datatype CollectionKind =
+      | Seq
+      | Set
+      | Multiset
+      | Map(keyType: Type)
 
-      datatype CollectionKind =
-        | Seq
-        | Set
-        | Multiset
-        | Map(keyType: Type)
-
-      datatype Type =
-        | Unit
-        | Bool
-        | Char
-        | Int
-        | Real
-        | BigOrdinal
-        | BitVector(width: nat)
-        | Collection(finite: bool, kind: CollectionKind, eltType: Type)
-        | Function(args: seq<Type>, ret: Type) // TODO
-        | Class(classType: ClassType)
+    datatype Type =
+      | Unit
+      | Bool
+      | Char
+      | Int
+      | Real
+      | BigOrdinal
+      | BitVector(width: nat)
+      | Collection(finite: bool, kind: CollectionKind, eltType: Type)
+      | Function(args: seq<Type>, ret: Type) // TODO
+      | Class(classType: ClassType)
+    {
+      // TODO: remove?
+      predicate method NoLeftFunction()
       {
-        // TODO: remove?
-        predicate method NoLeftFunction()
-        {
-          match this {
-            case Unit => true
-            case Bool => true
-            case Char => true
-            case Int => true
-            case Real => true
-            case BigOrdinal => true
-            case BitVector(width: nat) => true
-            case Collection(finite: bool, kind: CollectionKind, eltType: Type) =>
-              && eltType.NoLeftFunction()
-              && match kind {
-                case Seq => true
-                case Set => true
-                case Multiset => true
-                case Map(kt) => kt.NoLeftFunction()
-              }
-            case Function(args: seq<Type>, ret: Type) => false
-            case Class(classType: ClassType) => false
-          }
-        }
-
-        // TODO: remove?
-        predicate method WellFormed() {
-          match this {
-            case Unit => true
-            case Bool => true
-            case Char => true
-            case Int => true
-            case Real => true
-            case BigOrdinal => true
-            case BitVector(width: nat) => true
-            case Collection(finite: bool, kind: CollectionKind, eltType: Type) =>
-              && eltType.WellFormed()
-              // This condition is overly restrictive: we will do the general case later.
-              // For instance, maps can contain keys which don't have a decidable equality,
-              // and sequences can contain elements which also don't have a decidable equality
-              // (in which case we don't have a decidable equality over the sequences, but it
-              // is fine).
-              && eltType.NoLeftFunction()
-              && match kind {
-                case Seq => true
-                case Set => eltType.NoLeftFunction()
-                case Multiset => eltType.NoLeftFunction()
-                case Map(kt) => kt.WellFormed() && kt.NoLeftFunction()
-              }
-            case Function(args: seq<Type>, ret: Type) =>
-              && (forall i | 0 <= i < |args| :: args[i].WellFormed())
-              && ret.WellFormed()
-            case Class(classType: ClassType) =>
-              && (forall i | 0 <= i < |classType.typeArgs| :: classType.typeArgs[i].WellFormed())
-          }
+        match this {
+          case Unit => true
+          case Bool => true
+          case Char => true
+          case Int => true
+          case Real => true
+          case BigOrdinal => true
+          case BitVector(width: nat) => true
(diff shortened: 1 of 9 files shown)


Source/DafnyCompilers/AST.dfy
```diff
@@ -1,350 +1,350 @@
-include "CSharpDafnyASTModel.dfy"
+  include "CSharpDafnyASTModel.dfy"
```
davidcok, line 1: Why is the include indented?

Source/DafnyCompilers/AST.dfy
```diff
@@ -1,350 +1,350 @@
...
+      | Int
+      | Real
+      | BigOrdinal
+      | BitVector(width: nat)
+      | Collection(finite: bool, kind: CollectionKind, eltType: Type)
+      | Function(args: seq<Type>, ret: Type) // TODO
+      | Class(classType: ClassType)
+    {
+      // TODO: remove?
+      predicate method NoLeftFunction()
```
davidcok, line 36: Expected 2 spaces less indent here

davidcok: and the rest of the predicate

Source/DafnyCompilers/AST.dfy
```diff
@@ -1,350 +1,350 @@
...
 
-    module BinaryOps {
-      datatype Logical =
-        Iff // And, Or, and Imp are in LazyOp
-      datatype Eq =
+  module BinaryOps {
+    datatype Logical =
+      Iff // And, Or, and Imp are in LazyOp
+    datatype Eq =
         EqCommon | NeqCommon
```
davidcok, line 104: 2 spaces less indent?
Better would be formatted onto the same line as 'datatype'

davidcok: And less indent for the subsequent declarations also

Source/DafnyCompilers/AST.dfy
MikaelMayer: Congratulations, you found a bug 🥈

Source/DafnyCompilers/AST.dfy
MikaelMayer: I'm not sure I get it. This portion looks correctly indented to me.

Source/DafnyCompilers/AST.dfy
MikaelMayer: I won't remove newlines in the scope of this PR, but I removed the extra indent in this particular case

Source/DafnyCompilers/AST.dfy
MikaelMayer: Fixed in the new version

Source/DafnyCompilers/AST.dfy
davidcok: FIne

Source/DafnyCompilers/AST.dfy
davidcok: Agreed. My misreading I guess

Source/DafnyCompilers/AST.dfy
```diff
@@ -2,349 +2,349 @@ include "CSharpDafnyASTModel.dfy"
...
+  module UnaryOps {
+    datatype UnaryOp =
+      | BVNot
+      | BoolNot
+      | SeqLength
+      | SetCard
+      | MultisetCard
+      | MapCard
+      | MemberSelect(name: string)
+        // Ghost operators
```
mschlaipfer, line 172: wrong indent?

Source/DafnyCompilers/CompilerRewriter.dfy
```diff
@@ -95,58 +95,58 @@ module CompilerRewriter {
...
+        // - all the children of `e'` deeply satisfy the post of `f`
+        // Then:
+        // - `e'` satisfies the pre of `f`
       {
         forall e, e' ::
-           && Exprs.ConstructorsMatch(e, e')
-           && f.requires(e)
-           && Deep.AllChildren_Expr(e', post)
-           ==> f.requires(e')
+        && Exprs.ConstructorsMatch(e, e')
```
mschlaipfer, line 110: seems better before?

Source/DafnyCompilers/CompilerRewriter.dfy
```diff
@@ -95,58 +95,58 @@ module CompilerRewriter {
...
+        //
+        // Given `e`, if:
+        // - all the children of `e` deeply satisfy the post of `f`
+        // - `e` satisfies the pre of `f`
+        // Then:
+        // - `f(e)` deeply satisfies the post of `f`
       {
         forall e: Expr | Deep.AllChildren_Expr(e, post) && f.requires(e) ::
-          Deep.All_Expr(f(e), post)
+                                                           Deep.All_Expr(f(e), post)
```
mschlaipfer, line 127: what happens here?

Source/DafnyCompilers/CompilerRewriter.dfy
```diff
@@ -348,8 +348,8 @@ module CompilerRewriter {
 
     // TODO: not sure it was worth making this opaque
     predicate {:opaque} GEqCtx(
-      eq_value: (WV,WV) -> bool, ctx: Context, ctx': Context
-    )
+        eq_value: (WV,WV) -> bool, ctx: Context, ctx': Context
+      )
```
mschlaipfer, line 352: is this better?

Source/DafnyCompilers/Interp.dfy
```diff
@@ -1,4 +1,4 @@
-include "Library.dfy"
+  include "Library.dfy"
```
mschlaipfer, line 1: wrong indent?

Source/DafnyCompilers/Interp.dfy
```diff
@@ -272,18 +272,18 @@ module Interp {
...
-              assert !bop.BV? && !bop.Datatypes?;
-              InterpBinaryOp(e, bop, argvs[0], argvs[1])
-            case TernaryOp(top: TernaryOp) =>
-              InterpTernaryOp(e, top, argvs[0], argvs[1], argvs[2])
-            case Builtin(Display(ty)) =>
-              InterpDisplay(e, ty.kind, argvs)
-            case FunctionCall() =>
-              InterpFunctionCall(e, env, argvs[0], argvs[1..])
-          })
+                         case UnaryOp(op: UnaryOp) =>
```
mschlaipfer, line 275: there's a code highlighting bug

Source/DafnyCompilers/Interp.dfy
```diff
@@ -468,41 +468,41 @@ module Interp {
...
-        Success(V.Int(|v0.st|))
-      case MultisetCard => :- Need(v0.Multiset?, Invalid(expr));
-        Success(V.Int(|v0.ms|))
-      case MapCard => :- Need(v0.Map?, Invalid(expr));
-        Success(V.Int(|v0.m|))
+    case BVNot => :- Need(v0.BitVector?, Invalid(expr));
+                     Success(V.BitVector(v0.width, Math.IntPow(2, v0.width) - 1 - v0.value))
+    case BoolNot => :- Need(v0.Bool?, Invalid(expr));
+                       Success(V.Bool(!v0.b))
+    case SeqLength => :- Need(v0.Seq?, Invalid(expr));
```
mschlaipfer, line 475: could we have a linebreak after `=>`? Why is `Success` aligned with the RHS of `:-`?

mschlaipfer (commented):
Nice progress!

Source/DafnyCompilers/AST.dfy
MikaelMayer: Very good finding, but that's expected.
1) We plan to have docstring to be comments like this one right after a constructor, so it's correctly aligned.
2) The comments just below `// |` are aligned with the `|` above because the Dafny formatter recognizes that this is a commented case of a constructor.

Source/DafnyCompilers/CompilerRewriter.dfy
MikaelMayer: You found a bug, congrats ! 🥉

Source/DafnyCompilers/CompilerRewriter.dfy
MikaelMayer: I'll add this to my final survey as I like the one on the right better. That way,  arguments don't have the same indentation as the body of the predicate/function, which makes it easy to determine the limit between the two.

Source/DafnyCompilers/Interp.dfy
MikaelMayer: Bug for which I pushed a fix on Friday, but I did not update this. Now it's done. Thanks

Source/DafnyCompilers/Interp.dfy
MikaelMayer: With the newest version, the code diff is no longer a problem, can you have a look and let me know if this is ok?
I still think the indentation of the match case should be more aligned.

Source/DafnyCompilers/Interp.dfy
MikaelMayer: The scope of the formatter does not include adding newlines.
Success is now aligned with :- and nots its RHS

PR #2716, opened 2022-09-08
fix: Use the right >/</>=/<= comparison operators for bitvectors in JS

### LINKED ISSUE #2672: JavaScript compiler: incorrect bignum comparison
Consider:

```
method Main() {
  print 24 as bv16 <= 1507 as bv16, "\n";
  assert 24 as bv16 <= 1507 as bv16;
}
```

This should print "true", which is what Go, Java, C#, and Python output. However, JavaScript prints "false".

cpitclaudel:
Closes #2672.


Test/git-issues/git-issue-2672.dfy
```diff
@@ -0,0 +1,82 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:cs "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:js "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:go "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:java "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:py "%s" >> "%t"
+// RUN: %diff "%s.expect" "%t"
```
robin-aws, line 7: Can you use `%testDafnyForEachCompiler %s` instead? :)

Source/Dafny/Compilers/Compiler-js.cs
```diff
@@ -2007,38 +2007,46 @@ protected class ClassWriter : IClassWriter {
...
 
         case BinaryExpr.ResolvedOpcode.Lt:
-          if (e0.Type.IsNumericBased() && AsNativeType(e0.Type) == null) {
+          if (AsNativeType(e0.Type) != null) {
+            opString = "<";
+          } else if (e0.Type.IsNumericBased() || e0.Type.IsBitVectorType) {
             callString = "isLessThan";
           } else {
-            opString = "<";
+            Contract.Assert(false);
```
robin-aws, line 2015: AFAICT we always `throw new cce.UnreachableException();` after such a contract assertion. That ensures control flow doesn't just continue in release builds.

robin-aws (changes requested):

Test/git-issues/git-issue-2672.dfy
cpitclaudel: Can you remind me how to generate the output with that flag?

robin-aws: It should just be a single copy of the program's output, not including the "Dafny program verifier ..." line. So in this case just:

```
true true true true true true true true true
false false false false false false false false false
false false false false false false false false false
true true true true true true true true true
true true true true true true true true true
false false false false false false false false false
true true true
false false false
```

It occurs to me that having a flag on the `TestDafny` tool to generate the expect file would be super handy, let me know if I can add that. :)

cpitclaudel: Thanks!  Unfortunately it doesn't correctly detect unsupported features, so it fails on this example; I've reported the error at https://github.com/dafny-lang/dafny/issues/2718 , and I'm currently preparing a PR.  We can update this test file once that PR is merged, but I don't think we need to block on it.

robin-aws (approved):

robin-aws (approved):

Issue #2620, opened 2022-08-22
Invalid UTF


sorawee:
From https://mnaoumov.wordpress.com/2014/06/14/stripping-invalid-characters-from-utf-16-strings/:

`"\ud800b"` is invalid because "here \ud800 is lead surrogate but it is followed by b letter which is not a low surrogate."

But 

```
method Main() {
  print "\ud800b";
}
```

works fine, and outputs `�b`

Ideally, this should result in a verification error.

Reported by @jswrenn

cpitclaudel:
Dafny strings are not valid unicode strings, currently, so I would say "not a bug"; we are in the process of revamping those strings in https://github.com/dafny-lang/rfcs/pull/13/

robin-aws:
Hi Sorawee! The new `--unicode-char` option, which will be on by default in Dafny 4.0, mostly addresses this gap. When it is enabled, the `\uXXXX` style of unicode escapes that identifies UTF-16 code points is no longer accepted, in favor of the `\U{X...X}` style that identifies Unicode scalar values instead. That means it's no longer possible to specify invalid strings like the one in the example. See https://dafny.org/latest/DafnyRef/DafnyRef#sec-characters for further details.

Even better, Dafny supports arbitrary Unicode characters directly in string literals now, so you no longer have to use Unicode escapes for such characters.

Closing this as I believe we've done everything we're planning to do in this area, but feel free to reopen!

PR #2717, opened 2022-09-08
Feat: Support for full UTF8 parsing by default.

MikaelMayer:
This PR enables support for parsing any UTF-8 string in a source file and in comments.

Is it a fix for #2620 ?


robin-aws (changes requested):
So glad the fix was as simple as I suspected! :)

Let's add at least one integration test too. Something like:

```dafny
method Main() {
  print "Mikaël Mayer fixed UTF8 parsing!", "\n";
}
```

I'd like to verify whether this is enough to support arbitrary characters in string literals (I suspect it is but I'm not sure).

We also need to update the reference manual since it touches on this: https://dafny.org/dafny/DafnyRef/DafnyRef#sec-unicode.

robin-aws:
> Is it a fix for https://github.com/dafny-lang/dafny/issues/2620 ?

Nope, because that issue talking about invalid uses of `\u` escape sequences. But you may be fixing the separate issue that directly included non-ASCII characters are truncated by the parser:

```dafny
method Main() {
  var s := "Unicode is just so 😎";
  print s, "\n";
}
```

This currently prints "Unicode is just so ð" (in C# at least), because the parser only retains the first byte (0xF0) of the emoji character's UTF8 encoding and then interprets it as a UTF16 code unit.

MikaelMayer:
I'm able to make it work with ë (the utf-8 sequence compiles the value of the two byte sequence to 240, which is interpreted as the `ë` in ASCII)
 but not the smiley. I tried my best for 30 minutes to change the scanner so that the smiley would pass, but 1) although it's one character, it takes two columns on an editor and 2) it would require not casting from (int) to (char), and I haven't found the good recipe.
So, do we want smileys for now, or should I just be happy with this change for now?

robin-aws:
No don't worry about smileys for now. :) I just wanted to understand the extent of the fix. If ë works please do add a test to lock that down, but that's enough.

robin-aws (changes requested):

Test/dafny0/PrintUTF8Fails.dfy.expect
```diff
@@ -0,0 +1,2 @@
+PrintUTF8Fails.dfy(5,8): Error: invalid LogicalExpression
```
robin-aws, line 1: Oh wow that's worse than I thought 😎

docs/DafnyRef/Grammar.md
```diff
@@ -89,10 +89,11 @@ let you reconstruct the original grammar.
 
 ## 2.1. Dafny Input {#sec-unicode}
 
-Dafny source code files are readable text encoded as UTF-8 Unicode
-(because this is what the Coco/R-generated scanner and parser read).
+Dafny source code files are readable text encoded as UTF-8 Unicode, where each decoded character has to fit
+a single UTF-16 character. This means that every character has to be assignable to a C# 'char', which
+has 65536 values. This is what the Coco/R-generated scanner and parser read.
```
robin-aws, line 94:
```suggestion
Dafny source code files are readable text encoded in UTF-8, where each decoded character has to be from the Basic Multilingual Plane and therefore encodable with a single UTF-16 code unit. This is what the Coco/R-generated scanner and parser read.
```

docs/DafnyRef/Grammar.md
```diff
@@ -89,10 +89,11 @@ let you reconstruct the original grammar.
...
 All program text other than the contents of comments, character, string and verbatim string literals
-consists of printable and white-space ASCII characters,
+consists of printable and white-space UTF-8 Unicode characters.
```
robin-aws, line 96:
```suggestion
consists of printable and white-space ASCII characters.
```

AFAICT the old version is still true, since the parser logic requires it even if the scanner is more liberal now.

Also UTF stands for Unicode Transformation Format, so UTF-8 Unicode is almost as redundant as "ATM Machine" :)

RELEASE_NOTES.md
```diff
@@ -1,5 +1,6 @@
 # Upcoming
 
+- feat: Support for parsing UTF8 characters in code and comments if their UTF16 value is on maximum 2 bytes (https://github.com/dafny-lang/dafny/pull/2717)
```
robin-aws, line 3:
```suggestion
- feat: Support for parsing Basic Multilingual Plane characters from UTF-8 in code and comments (https://github.com/dafny-lang/dafny/pull/2717)
```

Nit picking again, I know, but "UTF-8 character" is a bit confusing, since UTF-8 is a way to encode characters using sequences of code units (bytes).

robin-aws (changes requested):
Thanks for continuing to entertain my nit-picking.

robin-aws (approved):
❤️ (even if we can't quite parse that yet)

docs/DafnyRef/Grammar.md
```diff
@@ -89,10 +89,9 @@ let you reconstruct the original grammar.
 
 ## 2.1. Dafny Input {#sec-unicode}
 
-Dafny source code files are readable text encoded as UTF-8 Unicode
-(because this is what the Coco/R-generated scanner and parser read).
+Dafny source code files are readable text encoded in UTF-8, where each decoded character has to be from the Basic Multilingual Plane and therefore encodable with a single UTF-16 code unit. This is what the Coco/R-generated scanner and parser read.
 All program text other than the contents of comments, character, string and verbatim string literals
-consists of printable and white-space ASCII characters,
+consists of printable and white-space ASCII characters.
```
davidcok, line 94: The change from , to . is wrong here -- it leaves a sentence fragment on the next line.

MikaelMayer: Can you fix it in your next PR please? I can review that. Thanks for reporting.

davidcok: Added to cok-docs-TODOs, PR#2817

PR #2725, opened 2022-09-12
Chore: Fixed HTML rendering of least and greatest predicates approximations

MikaelMayer:
This fixes a serious problem in the web documentation where the negative space was displayed like a "!" in inline TeX and as expected in TeX blocks. This seems to be the case because because Markdown does some preprocessing and replaces `\!` with `!`. With the `\kern` command, we don't have this problem.

Before:
image

After:
image

20 places were concerned for "f" and 12 for "g".


RustanLeino (approved):
Good catch! This would indeed have been very confusing before.

PR #2739, opened 2022-09-13
fix: Include standard header with every unsupported feature message

### LINKED ISSUE #2718: for-each-compiler does not recognize unsupported features reported with a custom message
Trying to test this code with `%testDafnyForEachCompiler` fails:

```dafny
type sreal = r: real | r > 3
```

The output is the following:

```
Output:
Verifying...
Executing on C#...
Executing on JavaScript...
Executing on Go...
Executing on Java...
Executing on Python...
Executing on C++...
Execution failed, for reasons other than known unsupported features. Output:

test.dfy(3,8): Error: non-native newtype sreal
```

cpitclaudel:
Fixes #2718.


Test/metatests/EarlyExitForUnsupportedFeatures.dfy.oldway.expect
```diff
@@ -0,0 +1,3 @@
+
+Dafny program verifier did not attempt verification
+EarlyExitForUnsupportedFeatures.dfy(6,8): Error: Feature not supported for this compilation target: Non-native numeric newtypes (non-native newtype positive)
```
robin-aws, line 3: Just for the record, I'm not super proud of using `:` no less than 3 times in a single "sentence", but I can live with it as I can't think of a better way to cram all this information into a single error message. :)

robin-aws (dismissed):
Nice refactoring, and thanks for getting to this before I could!

cpitclaudel:
Assigning @robin-aws to finish the test updates (as discussed in person).

PR #2742, opened 2022-09-14
fix: Various issues in the Python compiler

fabiomadge:


Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -843,11 +855,11 @@ private class ClassWriter : IClassWriter {
       var wStmts = wr.Fork();
       wr.Write($"raise {DafnyRuntimeModule}.HaltException(");
       if (tok != null) {
-        wr.Write($"{DafnyRuntimeModule}.Seq(\"{Dafny.ErrorReporter.TokenToString(tok)}: \") + ");
+        wr.Write($"\"{ErrorReporter.TokenToString(tok)}: \" + {DafnyRuntimeModule}.string_of(");
```
robin-aws, line 858: Just curious, is this an optimization or fixing a bug of some kind?

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1096,6 +1107,8 @@ private class ClassWriter : IClassWriter {
       }
       if (dtv.Ctor.EnclosingDatatype is not TupleTypeDecl) {
         wr.Write($"{DtCtorDeclarationName(dtv.Ctor)}");
+      } else if (dtv.Ctor.Destructors.Count(d => !d.IsGhost) == 1) {
+        arguments += ",";
```
robin-aws, line 1111: This feels like the wrong place to handle commas, and confused as heck about what this is trying to do. Won't you end up with `([...],)` below?

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1522,17 +1537,12 @@ private class ClassWriter : IClassWriter {
...
 
-      resultType = resultType.NormalizeExpand();
-      var thenExpr = Expr(thn, inLetExprBody, wStmts);
-      var castedThenExpr = resultType.Equals(thn.Type.NormalizeExpand()) ? thenExpr : Cast(resultType, thenExpr);
-      var elseExpr = Expr(els, inLetExprBody, wStmts);
-      var castedElseExpr = resultType.Equals(els.Type.NormalizeExpand()) ? elseExpr : Cast(resultType, elseExpr);
-      wr.Format($"({castedThenExpr} if {Expr(guard, inLetExprBody, wStmts)} else {castedElseExpr})");
+      ConcreteSyntaxTree Tree(Expression e) => Expr(e, inLetExprBody, wStmts);
+
+      wr.Format($"({Tree(thn)} if {Tree(guard)} else {Tree(els)})");
```
robin-aws, line 1545: Nice cleanup :)

Source/DafnyCore/Compilers/Compiler-python.cs
fabiomadge: Which is exactly what you want, as the no comma version doesn't create a tuple.

fabiomadge: It's something we need to deal with as we use native tuples.

Source/DafnyCore/Compilers/Compiler-python.cs
fabiomadge: Both. Before we used `+` on a `str` and an arbitrary python object to create a string `Seq`. This would only be used when catching the exception, so no need for the dafny wrapper. Instead of fixing this mess, I got rid of it.

Source/DafnyCore/Compilers/Compiler-python.cs
robin-aws: RIGHT, please do add a comment - it looks so weird it feels like a typo. :)

robin-aws (approved):

PR #2744, opened 2022-09-14
Add DafnyRuntime.Tests to Dafny.sln

atomb:
The tests for the DafnyRuntime package were inadvertently left out of the top level Dafny.sln, and therefore weren't being run in CI. This adds the test package so that it will run along with all other tests.

### DIFF
--- a/.github/workflows/xunit-tests.yml
+++ b/.github/workflows/xunit-tests.yml
@@ -66,6 +66,8 @@ jobs:
       run: dotnet test --no-restore --verbosity normal --logger trx Source/DafnyTestGeneration.Test
     - name: Run AutoExtern Tests
       run: dotnet test --no-restore --verbosity normal --logger trx Source/AutoExtern.Test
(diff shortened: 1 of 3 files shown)


robin-aws:
You just have to fix `customBoogie.patch` accordingly. I've found it easier to manually shift the new lines earlier in the `Dafny.sln` file to make that easier. :)

robin-aws (changes requested):
Shoot, another wrinkle (and apologies for not spotting this earlier) - our CI workflow runs each unit test project in a separate step, so you have to add this project explicitly there too: https://github.com/dafny-lang/dafny/blob/master/.github/workflows/xunit-tests.yml#L59-L68

We do that because we have to run `IntegrationTests` (i.e. all the Lit tests) separately, so we can't just do `dotnet test Source/Dafny.sln` to run all tests with a single command. There's probably a better way to group things to solve this though...

robin-aws (approved):

PR #2745, opened 2022-09-14
Solution to issue 2265 related to erroneous substitution under bindin…

### LINKED ISSUE #2265: ArgumentNullException in Translator.CheckDefiniteAssignment
repro:
```dafny
function Succ(x: int): int
{
  x + 1
}

function FailingLemma(): nat
{
  var st : set<int> :| forall x | x in st :: Succ(x) > 0;
  assert true by {
    if x: nat :| Succ(x) > 1 {
      assert Succ(x) > 0;
    }
  }
  0
}
```

output:
```
$ dafny /version
Dafny 3.6.0.40511
$ dafny repro.dfy 
Unhandled exception. System.AggregateException: One or more errors occurred. (Value cannot be null. (Parameter 'key'))
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.Collections.Generic.Dictionary`2.FindValue(TKey key)
   at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
   at Microsoft.Dafny.Translator.CheckDefiniteAssignment(IdentifierExpr expr, BoogieStmtListBuilder builder)
   at Microsoft.Dafny.Translator.CheckWellformedWithResult(Expression expr, WFOptions options, Expr result, Type resultType, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.CheckWellformed(Expression expr, WFOptions options, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.CheckWellformedWithResult(Expression expr, WFOptions options, Expr result, Type resultType, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.CheckWellformed(Expression expr, WFOptions options, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.CheckWellformedWithResult(Expression expr, WFOptions options, Expr result, Type resultType, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.CheckWellformed(Expression expr, WFOptions options, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmt_CheckWellformed(Expression expr, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran, Boolean subsumption, Boolean lValueContext)
   at Microsoft.Dafny.Translator.TrPredicateStmt(PredicateStmt stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmt(Statement stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmtList(List`1 stmts, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmt(Statement stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmt2StmtList(BoogieStmtListBuilder builder, Statement block, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrIfStmt(IfStmt stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmt(Statement stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmtList(List`1 stmts, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmt(Statement stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrPredicateStmt(PredicateStmt stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.TrStmt(Statement stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.CheckWellformedWithResult(Expression expr, WFOptions options, Expr result, Type resultType, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.CheckWellformedLetExprWithResult(LetExpr e, WFOptions options, Expr result, Type resultType, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran, Boolean checkRhs)
   at Microsoft.Dafny.Translator.CheckWellformedWithResult(Expression expr, WFOptions options, Expr result, Type resultType, List`1 locals, BoogieStmtListBuilder builder, ExpressionTranslator etran)
   at Microsoft.Dafny.Translator.AddWellformednessCheck(Function f)
   at Microsoft.Dafny.Translator.AddFunction_Top(Function f, Boolean includeAllMethods)
   at Microsoft.Dafny.Translator.AddClassMembers(TopLevelDeclWithMembers c, Boolean includeAllMethods)
   at Microsoft.Dafny.Translator.AddTypeDecl(RevealableTypeDecl d)
   at Microsoft.Dafny.Translator.DoTranslation(Program p, ModuleDefinition forModule)
   at Microsoft.Dafny.Translator.Translate(Program p, ErrorReporter reporter, TranslatorFlags flags)+MoveNext()
   at Microsoft.Dafny.DafnyDriver.Translate(ExecutionEngineOptions options, Program dafnyProgram)+MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Microsoft.Dafny.DafnyDriver.ProcessFiles(IList`1 dafnyFiles, ReadOnlyCollection`1 otherFileNames, ErrorReporter reporter, Boolean lookForSnapshots, String programId)
   at Microsoft.Dafny.DafnyDriver.ThreadMain(String[] args)
   at Microsoft.Dafny.DafnyDriver.<>c__DisplayClass6_0.<Main>b__0()
Aborted
```

Related to #2264, this was the kind of exception I was getting that led me to #2264

jtristan:
…g guards

Fixes #2265 

### DIFF
--- a/Source/DafnyCore/Substituter.cs
+++ b/Source/DafnyCore/Substituter.cs
@@ -14,9 +14,11 @@ namespace Microsoft.Dafny {
     protected readonly Dictionary<TypeParameter, Type> typeMap;
     protected readonly Label oldHeapLabel;
 
-    public static readonly Substituter EMPTY = new Substituter(null, new Dictionary<IVariable, Expression>(), new Dictionary<TypeParameter, Type>());
+    public static readonly Substituter EMPTY = new Substituter(null, new Dictionary<IVariable, Expression>(),
+      new Dictionary<TypeParameter, Type>());
 
-    public Substituter(Expression receiverReplacement, Dictionary<IVariable, Expression> substMap, Dictionary<TypeParameter, Type> typeMap, Label oldHeapLabel = null) {
+    public Substituter(Expression receiverReplacement, Dictionary<IVariable, Expression> substMap,
+      Dictionary<TypeParameter, Type> typeMap, Label oldHeapLabel = null) {
       Contract.Requires(substMap != null);
       Contract.Requires(typeMap != null);
       this.receiverReplacement = receiverReplacement;
@@ -24,11 +26,13 @@ namespace Microsoft.Dafny {
       this.typeMap = typeMap;
       this.oldHeapLabel = oldHeapLabel;
     }
+
     public virtual Expression Substitute(Expression expr) {
       Contract.Requires(expr != null);
       Contract.Ensures(Contract.Result<Expression>() != null);
 
-      Expression newExpr = null;  // set to non-null value only if substitution has any effect; if non-null, the .Type of newExpr will be filled in at end
+      Expression
+        newExpr = null; // set to non-null value only if substitution has any effect; if non-null, the .Type of newExpr will be filled in at end
 
       if (expr is StaticReceiverExpr) {
         var e = (StaticReceiverExpr)expr;
@@ -93,6 +97,7 @@ namespace Microsoft.Dafny {
             anyChanges = true;
           }
         }
+
         if (anyChanges) {
           newExpr = new MapDisplayExpr(expr.tok, e.Finite, elmts);
         }
@@ -101,8 +106,10 @@ namespace Microsoft.Dafny {
         Expression substE = Substitute(mse.Obj);
         MemberSelectExpr fseNew = new MemberSelectExpr(mse.tok, substE, mse.MemberName);
         fseNew.Member = mse.Member;
-        fseNew.TypeApplication_AtEnclosingClass = mse.TypeApplication_AtEnclosingClass.ConvertAll(t => Resolver.SubstType(t, typeMap));
-        fseNew.TypeApplication_JustMember = mse.TypeApplication_JustMember.ConvertAll(t => Resolver.SubstType(t, typeMap));
+        fseNew.TypeApplication_AtEnclosingClass =
+          mse.TypeApplication_AtEnclosingClass.ConvertAll(t => Resolver.SubstType(t, typeMap));
+        fseNew.TypeApplication_JustMember =
+          mse.TypeApplication_JustMember.ConvertAll(t => Resolver.SubstType(t, typeMap));
         fseNew.AtLabel = mse.AtLabel ?? oldHeapLabel;
         newExpr = fseNew;
       } else if (expr is SeqSelectExpr) {
@@ -139,10 +146,11 @@ namespace Microsoft.Dafny {
         if (receiver != e.Receiver || newArgs != e.Args ||
             newTypeApplicationAtEnclosingClass != e.TypeApplication_AtEnclosingClass ||
             newTypeApplicationJustFunction != e.TypeApplication_JustFunction) {
-          FunctionCallExpr newFce = new FunctionCallExpr(expr.tok, e.Name, receiver, e.OpenParen, e.CloseParen, newArgs, e.AtLabel ?? oldHeapLabel);
-          newFce.Function = e.Function;  // resolve on the fly (and set newFce.Type below, at end)
-          newFce.CoCall = e.CoCall;  // also copy the co-call status
-          newFce.CoCallHint = e.CoCallHint;  // and any co-call hint
+          FunctionCallExpr newFce = new FunctionCallExpr(expr.tok, e.Name, receiver, e.OpenParen, e.CloseParen, newArgs,
+            e.AtLabel ?? oldHeapLabel);
+          newFce.Function = e.Function; // resolve on the fly (and set newFce.Type below, at end)
+          newFce.CoCall = e.CoCall; // also copy the co-call status
+          newFce.CoCallHint = e.CoCallHint; // and any co-call hint
           newFce.TypeApplication_AtEnclosingClass = newTypeApplicationAtEnclosingClass;
           newFce.TypeApplication_JustFunction = newTypeApplicationJustFunction;
           newFce.IsByMethodCall = e.IsByMethodCall;
@@ -157,10 +165,12 @@ namespace Microsoft.Dafny {
 
       } else if (expr is DatatypeValue) {
         DatatypeValue dtv = (DatatypeValue)expr;
-        var newArguments = SubstituteExprList(dtv.Bindings.Arguments); // substitute into the expressions, but drop any binding names (since those are no longer needed)
+        var newArguments =
+          SubstituteExprList(dtv.Bindings
+            .Arguments); // substitute into the expressions, but drop any binding names (since those are no longer needed)
         if (newArguments != dtv.Bindings.Arguments) {
           DatatypeValue newDtv = new DatatypeValue(dtv.tok, dtv.DatatypeName, dtv.MemberName, newArguments);
-          newDtv.Ctor = dtv.Ctor;  // resolve on the fly (and set newDtv.Type below, at end)
+          newDtv.Ctor = dtv.Ctor; // resolve on the fly (and set newDtv.Type below, at end)
           newDtv.InferredTypeArgs = Util.Map(dtv.InferredTypeArgs, tt => Resolver.SubstType(tt, typeMap));
           // ^ Set the correct type arguments to the constructor
           newExpr = newDtv;
@@ -185,8 +195,10 @@ namespace Microsoft.Dafny {
           if (fefe != fe) {
             anythingChanged = true;
           }
+
           fr.Add(fefe);
         }
+
         if (anythingChanged) {
           newExpr = new UnchangedExpr(e.tok, fr, e.At) { AtLabel = e.AtLabel ?? oldHeapLabel };
         }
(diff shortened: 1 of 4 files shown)


Source/DafnyCore/Substituter.cs
```diff
@@ -930,7 +979,73 @@ public class Substituter {
...
+      // For quantifiers and setComprehesion we want to make sure that we don't introduce name clashes with
+      // the enclosing scopes.
+
+      var q = e as QuantifierExpr;
+      if (q != null && q.SplitQuantifier != null) {
+        if (forceSubstituteOfBoundVars) {
+          return Substitute(q.SplitQuantifierExpression);
+        } else {
+          return SubstituteComprehensionExpr((ComprehensionExpr)q.SplitQuantifierExpression, false);
+        }
```
cpitclaudel, line 1006: As written, I think this code could be simplified to the following, because if `q.SplitQuantifierExpression` is a `ComprehensionExpr`, then the call to Substitute will immediately call `SubstituteComprehensionExpr(…, true)`.

```csharp
return SubstituteComprehensionExpr((ComprehensionExpr)q.SplitQuantifierExpression, forceSubstituteOfBoundVars);
```

However, I'm not sure this code is correct, because `q.SplitQuantifierExpression` is a conjunction of quantifiers, not a quantifier itself.  I would expect this code to raise an exception if it encounters a split quantifier — not just your new code, but also the old code.

This is because `Substitute(q.SplitQuantifierExpression);` was always wrong: it takes a quantifier and returns a conjunction or disjunction of quantifiers, so when applied to the `BindingGuard` of an `If` it replaces an `ExistsExpr` with a `BinaryExpr`, which confuses later parts of the code.  Here's a repro of this:

```
function method f(x: int): int { 10 - x * x }

function method BindingGuardTest(): int {
  var x: nat := 1;
  assert true by {
    if i :| 0 <= i < 10 && (f(i) == f(i+1) || f(i) == f(i+2)) {
      // print i;
    }
  }
  2
}
```

which outputs:

```
System.InvalidCastException: Unable to cast object of type 'Microsoft.Dafny.BinaryExpr' to type 'Microsoft.Dafny.ExistsExpr'.
   at Microsoft.Dafny.Translator.TrIfStmt(IfStmt stmt, BoogieStmtListBuilder builder, List`1 locals, ExpressionTranslator etran) 
```

For the bottom branch the fix would be to ignore the split expressions, I think.  But for the top branch if we want to preserve the splits then I think we have to apply the substitution on each of them.  @RustanLeino , thoughts?

Source/DafnyCore/Substituter.cs
```diff
@@ -930,7 +979,73 @@ public class Substituter {
...
+
+    /// <summary>
+    /// Substitution in a comprehension expression. 
+    /// Parameter 'forceSubstituteOfBoundVars' should be set to false if and only if
+    /// the expression is a binding guard, in which case a bound variable is introduced.
+    /// Such a variable must not be substituted. 
+    /// </summary>
+    protected Expression SubstituteComprehensionExpr(ComprehensionExpr expr, bool forceSubstituteOfBoundVars) {
+
+      Expression newExpr = null;
```
cpitclaudel, line 994: Could this be a `ComprehensionExpr` instead of an `Expr`?

Source/DafnyCore/Substituter.cs
```diff
@@ -930,7 +979,73 @@ public class Substituter {
...
+          newExpr = new ExistsExpr(expr.tok, e.BodyEndTok, newBoundVars, newRange, newTerm, newAttrs);
+        } else if (expr is LambdaExpr) {
+          var l = (LambdaExpr)expr;
+          newExpr = new LambdaExpr(e.BodyStartTok, e.BodyEndTok, newBoundVars, newRange,
+            l.Reads.ConvertAll(SubstFrameExpr), newTerm);
+        } else {
+          Contract.Assert(false); // unexpected ComprehensionExpr
+        }
+
+        ((ComprehensionExpr)newExpr).Bounds = newBounds;
```
cpitclaudel, line 1037: I think you won't need the cast here if you change the declaration of `newExpr`)

Source/DafnyCore/Substituter.cs
```diff
@@ -930,7 +979,73 @@ public class Substituter {
...
+      }
+
+      // undo any changes to substMap (could be optimized to do this only if newBoundVars != e.BoundVars)
+      foreach (var bv in e.BoundVars) {
+        substMap.Remove(bv);
+      }
+
+      return newExpr;
```
cpitclaudel, line 1045: Could we add an assertion before this to say when `newExpr` can be null?
I think it's something like this?

```
Contract.Assert(!forceSubstituteOfBoundVars ==> newExpr != null)
```

I was confused when I first read through the code to see that this was used directly in the `if` case but that it may also return null.

Test/git-issues/git-issue-2265.dfy
```diff
@@ -0,0 +1,18 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+module DefaultModule {
+  class DefaultClass {
+    static function BrokenFunction(): nat {
+      var y := 0;
+			assert true by {
```
cpitclaudel, line 8: The indentation is wrong here

cpitclaudel (changes requested):
Great fix!  A few suggestions:

- There are still a few whitespace-only changes in unrelated parts of the code that would be nice to undo
- I think this PR is missing a fix for the `if-case` pattern that @RustanLeino sent an example of in the original issue discussion (`if x :| P(x) => …`)
- The split-quantifiers case isn't quite right, I think — but the original code was broken too (!)
- I've left other small comments in the code

Source/DafnyCore/Substituter.cs
jtristan: I am going to apply the simplification that you suggest. As for the issue with the split quantifiers, since it is a different issue, I will open a new issue to report it.

Source/DafnyCore/Substituter.cs
jtristan: We could but here is my concern: it is not actually a required invariant, it just happens to be implemented that way to match the rest of the code.

jtristan:
With respect to the additional issue raised by Rustan involving a case statement, I decided to create a new issue because even if the fix might be small and simple, I would first like to make sure I understand the semantics, and I will add a new corresponding test.

Source/DafnyCore/Substituter.cs
jtristan: I created a new issue #2748 to continue this discussion.

Source/DafnyCore/Substituter.cs
```diff
@@ -262,58 +275,26 @@ public class Substituter {
...
           cases.Add(newCaseExpr);
         }
+
         if (anythingChanged) {
           var newME = new MatchExpr(expr.tok, src, cases, e.UsesOptionalBraces);
           newME.MissingCases.AddRange(e.MissingCases);
           newExpr = newME;
         }
 
       } else if (expr is ComprehensionExpr) {
```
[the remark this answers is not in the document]
jtristan: I will remain consistent with the rest of the code.

Source/DafnyCore/Substituter.cs
jtristan: Not sure. ```SplitQuantifierExpression``` returns an Expression, and not a ComprehensionExpression.

Source/DafnyCore/Substituter.cs
```diff
@@ -930,7 +979,73 @@ public class Substituter {
...
       return attrs;
     }
+
+    /// <summary>
+    /// Substitution in a comprehension expression. 
+    /// Parameter 'forceSubstituteOfBoundVars' should be set to false if and only if
+    /// the expression is a binding guard, in which case a bound variable is introduced.
+    /// Such a variable must not be substituted. 
+    /// </summary>
+    protected Expression SubstituteComprehensionExpr(ComprehensionExpr expr, bool forceSubstituteOfBoundVars) {
```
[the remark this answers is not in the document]
jtristan: Not sure. SplitQuantifierExpression returns an Expression, and not a ComprehensionExpression.

cpitclaudel (changes requested):
Thanks!  Can you revert the whitespace-only changes?  Then we can merge! :)

cpitclaudel (approved):

cpitclaudel (approved):

PR #2753, opened 2022-09-17
fix: Remove "IsNullableRefType" constraint to avoid unsound conversions

### LINKED ISSUE #2752: Soundness bug due to subset type
```
type foo = x: object? | x != null witness *

function method m(): foo {
  null
}

method Main() ensures false {
  var y := m();
  assert y == null;
  assert y != null;
}
```

cpitclaudel:
Fixes #2752.

Not for merging.  I wouldn't be surprised if it doesn't pass CI, but even if it did we need to debate how much the resolver should be doing.  The core of issue #2752 is that type inference is too lax, and hence that we assign a subset type without checking it:

```
type EmptyQ = x: object? | false witness *
const xq := null as EmptyQ
```

The reason this code work is that we set `null` to be of type `InferredTypeProxy`, with constraint `IsNullableRefType`.  We then add constraints `Assignable ? EmptyQ` and `Assignable EmptyQ ?`. Then, since `EmptyQ` satisfies the constraint `IsNullableRefType`, we unify `?` and `EmptyQ`.

What I'm not sure about is why it wouldn't be OK to remove the constraint entirely and instead infer the type of `null` as `object?`, so this PR tries to do that.

### DIFF
--- a/Source/DafnyCore/Resolver.cs
+++ b/Source/DafnyCore/Resolver.cs
@@ -5046,8 +5046,7 @@ namespace Microsoft.Dafny {
               var u = Types[1].NormalizeExpand();
               if (CheckTypeInference_Visitor.IsDetermined(t) &&
                   (fullstrength
-                   || !ProxyWithNoSubTypeConstraint(u, resolver)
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Replaced by #2754

PR #2755, opened 2022-09-18
RM edits in response to #2498

### LINKED ISSUE #2498: lemma vs. ghost method
The RM says that a lemma is a ghost method.
In places where a method may be called, it does not matter whether the lemma is named a ghost method.

But some situations, at least calc and assert-by statements, require the method to be called a lemma and reject it if it is called a ghost method.

1) do we want to change Dafny so that `lemma` and `ghost method` are entirely equivalent?, or
2) do we like the distinction because `lemma` emphasizes the purpose of the method at hand?

davidcok:
Fixes #2498


docs/DafnyRef/Types.md
```diff
@@ -2154,6 +2154,22 @@ This is done by declaring a method with the `lemma` keyword.
 Lemmas are implicitly ghost methods and the `ghost` keyword cannot
 be applied to them.
 
+Syntactically, lemmas can be placed where ghost methods can be placed, but they serve 
+a significantly different function. First of all, a lemma is forbidden to have 
+`modifies` clause: it may not change anything about even the ghost state; ghost methods
+may have `modifies clauses` and may change ghost (but not non-ghost) state.
```
RustanLeino, line 2160:
```suggestion
may have `modifies` clauses and may change ghost (but not non-ghost) state. 
```

RustanLeino (changes requested):
Looks good, modulo one typo.

RustanLeino (approved):

Issue #2425, opened 2022-07-15
Dafny 4.0 suggestion: `modify X {}` should become `modifies X {}`


davidcok:
The modify statement has two forms:

modify frame-expression; -- this havocs the fields of the given expressions

modify { frame-expression } { statements } -- in this case the 'modify' is actually a 'modifies' clause whose scope is the succeeding block statement. The description in the RM even calls it a modifies clause.

Given that 'modifies' is also a keyword, I think that the second form is misnamed and confusing. It ought to use the keyword 'modifies'. 

I suggest that 'modifies' be permitted in this place, that this second form be, grammatically, a form of a block statement rather than a modify statement, and that 'modify in this second form eventually  be deprecated.

MikaelMayer:
I wholeheartedly agree as myself I found it confusing. I renamed the title of this issue to specify it could be a change to integrate in Dafny 4, since it's a breaking change.

robin-aws:
I'm definitely game for this, and the migration is so simple I think it's reasonable to just drop `modify` in Dafny 4.0. We can still recognize `modify` but reject it with a clear error message.

RustanLeino:
(First some general comments, then my overall thought on the matter.)

The `modify` statement was introduced for one purpose, which was to allow a refinement module to modify some previously declared state (see http://leino.science/papers/krml248.pdf). The idea is that an abstract module writes `modify W;`, and that a refinement module thereof makes the statement more concrete by writing `modify ... { S }`. Since the token `...` is used to stand for the original `W`, the statement also permits the statement `modify W { S }`.

The keyword `modify` for the statement goes with the specification that says that a method `modifies` something. This is analogous with `return` and `returns`, and with `yield` and `yields`.

The syntax of `modify` could be improved using the keyword `by` when a body is present. Not only would that make it read better, but it would also make it look more like `assert P by { S }`, which was introduced much after the `modify` statement. To me, it seems fine to have the statements `assert P;` and `assert P by { S }` -- that is, there seems to be no reason to make the latter `asserts P by { S }` or `ensures P by { S }`.

We have received some request for a more general statement with specification clauses, like the block specification in VCC. Such a statement could, like @davidcok proposes above, have a `modifies` clause. I'm sympathetic to such a block statement, but it would be nice for it to be motivated by strong use cases. For consistency with other statements, I think it should then probably also start with another keyword. For example,

```
do
  requires Pre
  modifies W
  ensures Post
{
  S
}
```

[...]

PR #2756, opened 2022-09-19
Deprecating statement refinement syntax - #2765

### LINKED ISSUE #2765: Deprecate statement level refinements
This is extracting a portion of Issue #2425 to just the matter of deprecating refinement of statements, such as if, while, assert, assume.

davidcok:
Fixes #2765, which was part of #2425


Makefile
```diff
@@ -35,6 +35,7 @@ clean:
 	(cd ${DIR}; cd Source; rm -rf Dafny/bin Dafny/obj DafnyDriver/bin DafnyDriver/obj DafnyRuntime/obj DafnyRuntime/bin DafnyServer/bin DafnyServer/obj DafnyPipeline/obj DafnyPipeline/bin DafnyCore/obj DafnyCore/bin)
 	(cd ${DIR} ; dotnet build Source/Dafny.sln -v:q --nologo -target:clean )
 	make -C ${DIR}/Source/DafnyCore -f Makefile clean
+	(cd ${DIR}/Source/Dafny && rm -rf Scanner.cs Parser.cs obj )
```
MikaelMayer, line 38: Why do you need the -r and the -f flag? -r is for recursive, and -f is for force, and I don't think you need either of these.

MikaelMayer (commented):
I'm good with it, but I would prefer to have @RustanLeino provide the final review for that because I don't know the ins and outs of statement refinements

Makefile
davidcok: The -r is because obj is a directory; the -f is because the arguments might not exist.
The reason for this addition is that the location of the Scanner.cs and Parser.cs files changed at some point. If one switches branches, does a `make clean exe`, errors or an incorrect build will result because of the obsolete Scanner or Parser. I've been caught be this many times and I often have forgotten the problem

davidcok:
OK with having Rustan review it, Actually Anjali has called a halt on deprecating stuff, so there will be some strategic conversation before this is merged.

RustanLeino (changes requested):
Looks good. I suggest that this PR also deprecate the `modify` statement.

davidcok:
Did you mean to deprecate both senses of the modify statement?

davidcok:
Mikael - is it the `modify X` (i.e. havoc) that you use, or the `modify x { }`  which ought to be `modifies` and which Rustan proposes to perhaps eventually become do requires... modifies... ensures... { ... }?

RustanLeino:
Okay, let's treat the `modify` statement separately from this PR. I will approve this PR as is.

RustanLeino (approved):

PR #2759, opened 2022-09-19
Fix: substitution of binding guards in alternative statement correcte…

### LINKED ISSUE #2747: Erroneous substitution for binding guard with case statement
Follow up to issue #2265 with partial solution in PR #2745 

```
function AnotherBrokenFunction(): nat {
  var y := 0;
  assert true by {
    if
    case x: bool :| true =>
      assert x;
  }
  0
}
```

jtristan:
…d to prevent substitution of bound variable.

Fixes #2747


Test/git-issues/git-issue-2747.dfy
```diff
@@ -0,0 +1,12 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+function AnotherBrokenFunction(): nat {
+  var y := 0;
+  assert true by {
+    if
+    case x: bool :| true =>
```
cpitclaudel, line 8: I would suggest changing the example so that it doesn't emit a trigger-related warning; maybe introduce a `predicate P(x: bool)` and change `true` to `P(x)`?

We could even do that for the other example as well.

cpitclaudel (approved):
Looks good.  I left one suggested change.

Test/git-issues/git-issue-2747.dfy
jtristan: It would be nice but unfortunately, it leads to yet another (unrelated) error, and I need to do some digging to see if it has been reported yet.

PR #2763, opened 2022-09-20
tests: Add a tutorial on parser combinators

cpitclaudel:
Tagging @codyroux, if you want to comment.


Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,265 @@
...
+/// ```ocaml
+/// let parentheses' () =
+///   (either
+///      (concat (char '(') (concat parentheses' (char ')')))
+///      epsilon)
+/// let parentheses () =
+///   (concat parentheses' eos)
+/// ```
+///
+/// Dafny will reject this definintion definition because thunking isn't enough
```
codyroux, line 19: `definition definition`?

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,265 @@
...
+/// to convince it that the `parentheses'` function terminates.
+///
+/// Instead, we need to use stronger specifications to enforce progress, as
+/// shown in the solution below:
+
+module {:options "-functionSyntax:4"} Parsers {
```
codyroux, line 25: Since this is a tutorial, maybe we should say something about the `-functionSyntax:4` pragma.

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,265 @@
...
+    }
+  }
+
+/// A combinator is a function from a position to a parse result.
+
+  type Parser<+T> = nat --> ParseResult<T>
+
+/// The following is a convenient trick when defining many functions that share
+/// an argument, without having to pass it around everywhere and without having
+/// to specify that multiple calls use the same value of it: we define a
```
codyroux, line 56: "of it" is a bit awkward. "use it with the same value" maybe?

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,265 @@
...
+
+  datatype Engine = Engine(input: string) {
+
+/// `Concat` takes two parsers and returns a lambda:
+
+    static function Concat0<L, R>(left: Parser<L>, right: Parser<R>)
+      : (p: Parser<(L, R)>)
+    {
+      (pos: nat)
+        requires left.requires(pos)
```
codyroux, line 68: If this is a tutorial, this `requires` yoga might warrant some explanations.

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,265 @@
...
+        requires left(pos).Failure? ==> right.requires(pos)
+      =>
+        match left(pos)
+          case Failure => right(pos).MapResult(r => Right(r))
+          case Success(pos, l) => Success(pos, Left(l))
+    }
+
+/// Next, we define `Char`, which checks for a specific character and advances
+/// the cursor by one.  Its post-condition indicates that the cursor has
+/// advanced and that there was some input left.
```
codyroux, line 138: "was" -> "is"

Test/tutorial/parser_combinators.dfy
codyroux: Also: "thunking" as a verb might be a bit arcane, "pushing the recursion under a function definition" maybe?

Test/tutorial/parser_combinators.dfy
cpitclaudel: `was` was right, but I have now changed the postcondition to avoid this confusing statement.

cpitclaudel:
Thanks for the excellent feedback, Cody!  I think I've addressed everything.

codyroux (approved):

Test/tutorial/parser_combinators.dfy.expect
```diff
@@ -0,0 +1,6 @@
+parser_combinators.dfy(28,27): Error: cannot use naked function in recursive setting. Possible solution: eta expansion.
+parser_combinators.dfy(29,33): Error: cannot prove termination; try supplying a decreases clause
+
+Dafny program verifier finished with 8 verified, 2 errors
+"((()))": 3 nested parentheses
+"((())": no valid parse
```
MikaelMayer, line 6: Can you possibly add the example "((()))a" as well?

MikaelMayer (changes requested):
This is a great tutorial, and very useful ! Good job for putting it together, I've learned a lot.
A few things that can be improved to make it in its best shape.

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,294 @@
+/// # Parser combinators in Dafny
+///
+/// Challenge: define parse combinators.  As a usability test, use them to parse
+/// expressions of the form `(ⁿ)ⁿ` (`n` opening parentheses followed by `n`
+/// closing parentheses).
+///
+/// This challenge is tricky because we can't write what we'd usually write in
+/// Haskell or OCaml:
+///
```
jtristan, line 9: For people who might want to understanding this tutorial and do not yet know about parsing combinators (or need a refresher), we should provide some good references and explain a little more.

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,294 @@
...
+///      (concat (char '(') (concat (fun () -> parentheses' ()) (char ')')))
+///      epsilon)
+/// let parentheses () =
+///   (concat parentheses' eos)
+/// ```
+///
+/// Dafny will reject this definition because it cannot prove that the
+/// `parentheses'` function terminates.  This is because Dafny analyses
+/// anonymous functions (“lambdas”) modularly: every time a lambda is created,
+/// as with `(fun () -> parentheses' ())`, Dafny checks that it can be call in
```
jtristan, line 22: Ambiguous pronoun.

jtristan, line 22: Imprecise passive voice.

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,294 @@
...
+/// any context.  To see why, consider the function `Apply` below and the
+/// following two uses of it:
+
+function method Apply(f: () -> int): int { f() }
+
+function F0(): int { Apply(F0) }
+function F1(): int { Apply(() => F1()) }
+
+/// Dafny rejects `F0` because it does not allow using “naked” recursive
+/// functions, so one must “eta-expand” (or “thunk”) it, as done in `F1`; but
```
jtristan, line 32: Jargon.

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,294 @@
...
+
+/// (The `-functionSyntax:4` enables Dafny's new function syntax, where
+/// `function` replaces `function method`)
+///
+/// ## Utility types
+///
+/// `Either<L, R>` represents the result of the `Either` combinator.  For `And`,
+/// we just use built-in pairs.
+
+  datatype Either<+L, +R> = Left(l: L) | Right(r: R)
```
jtristan, line 51: Is this tutorial going to make use of subtyping? If not, I would suggest avoiding advanced programming concepts which are orthogonal to the tutorial.

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,294 @@
...
+/// success that moves the cursor.  An alternative encoding would record the
+/// `delta` by which an operation moved the cursor, rather than its new position
+/// (the verification works the same way).
+
+  datatype ParseResult<+T> = Failure | Success(pos: nat, t: T) {
+    predicate IsFailure() { Failure? }
+    function PropagateFailure<U>(): ParseResult<U> requires Failure? { Failure() }
+    function Extract(): (nat, T) requires Success? { (pos, t) }
+    function MapResult<T'>(f: T -> T'): ParseResult<T'> {
+      match this
```
jtristan, line 63: ??? I am lost, can you explain?

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,294 @@
...
+///   ensures forall input :: (precondition of the parser) ==> p.requires(input)
+/// ```
+///
+/// This does not say that the parser must accept all inputs: only that the
+/// condition `(precondition of the parser)` is sufficient to call the opaque
+/// function `p` that `Concat` returns.
+///
+/// With that in mind, we can define `Either` in the same way:
+
+    static function {:opaque} Either<L, R>(
```
jtristan, line 142: Is opaque strictly necessary?

jtristan, line 142: Why the static? I think I am confused, I don't see that we are defining a class (and I know that there is a default one, but why would be use it for a tutorial? )

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,294 @@
...
+/// implements a simple semantic action (here, counting parentheses).
+
+    function {:opaque} Parentheses'(pos: nat): ParseResult<nat>
+      decreases |input| - pos
+    {
+      var rec := pos' requires pos < pos' <= |input| => Parentheses'(pos');
+      Either(Concat(Char('('), Concat(rec, Char(')'))), Epsilon)(pos)
+        .MapResult<nat>(
+          var count := (r: Either<(char, (nat, char)), ()>) =>
+            match r
```
jtristan, line 198: I like extra braces for the match. Even if they're not necessary, I think it is good practice for nested matches.

jtristan (dismissed):
I only did a superficial pass for now focused mostly on the writing style and the choice of subset of Dafny you are using for the tutorial.

Test/tutorial/parser_combinators.dfy
cpitclaudel: I've added some explanatory text

Test/tutorial/parser_combinators.dfy
cpitclaudel: I prefer to keep the `+` in there, because that's the "right" definition of `Either`.  Is that what you are referring to?

Test/tutorial/parser_combinators.dfy
cpitclaudel: I agree for nested matches, but isn't it verbose for non-nested matches?

Test/tutorial/parser_combinators.dfy
```diff
@@ -0,0 +1,294 @@
...
+/// reasoning.  This will cause verification-performance issues down the line,
+/// so we use the following alternative definition instead:
+
+    static function {:opaque} Concat<L, R>(
+      left: Parser<L>,
+      right: Parser<R>
+    )
+      : (p: Parser<(L, R)>)
+      ensures forall pos: nat |
+        && left.requires(pos)
```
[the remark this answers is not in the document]
cpitclaudel: I added a note about this, but I'd rather not define a separate predicate: I don't want users to think that it's necessary.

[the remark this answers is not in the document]
MikaelMayer: ok if you added a note about it then.

Test/tutorial/parser_combinators.dfy
MikaelMayer: I prefer without parentheses here.

Test/tutorial/parser_combinators.dfy
cpitclaudel: Here, yes, for performance.  I added some explanation above.

cpitclaudel: Because it doesn't use `this` (you can have member functions in datatypes was well, not just in classes)

MikaelMayer (approved):
Fantastic !

Issue #1429, opened 2021-09-11
(soundness) datatype doesn't handle type params in fields for equality support 


tjhance:
Related to https://github.com/dafny-lang/dafny/issues/1427 although this one doesn't rely on any external types, so I think it's a separate bug. Also, with this one it's pretty easy to produce a self-contained crashing example:

```dafny
datatype Option<V> = Some(v: V) | None

datatype X = X(ghost i: int)

datatype Y = Y(s: Option<X>)

method Main()
{
  var y1 := Y(Some(X(0)));
  var y2 := Y(Some(X(1)));

  // Dafny verifier thinks y1 == y2 is false
  // But it evaluates to true at runtime
  if y1 == y2 {
    var c := 1 / 0;
  }
}
```

Here, `X` shouldn't support runtime equality, and so `Y` shouldn't either - but Dafny doesn't detect this, and the auto-generated equality function is incorrect.

RustanLeino:
In addition to fixing this problem in the Resolver, I suggest that the compiler be changed to not produce any `Equals` method for `X` and `Y`, since `X` and `Y` never support equality at run time.

cpitclaudel:
Merging this with #1427 and #2070 based on my and @jtristan 's investigation: there's no neatly separate fix for each of these issues on its own, so it's better to centralize the discussion in one place.

Issue #2070, opened 2022-04-27
Missing recursion in check for `(==)`


cpitclaudel:
Dafny correctly rejects this:

```
datatype D<A> = D(a: A)
type DT(==)<A> = D<A> // Error: A might not support ==
datatype D'<A> = d'(d: D<A>)
type D'T(==)<A> = D'<A> // Error: A might not support ==
```

but it does reject the second line here:

```
datatype D<A> = D(a: A)
type DT(==)<A> = D<A> // Error: A might not support ==
datatype D'<A> = d'(d: D<D<A>>)
type D'T(==)<A> = D'<A> // Incorrectly accepted
```

This is because the check here is not recursive:

https://github.com/dafny-lang/dafny/blob/b5ad10a2f184f966e9d621732b59ca1ebed324ab/Source/Dafny/Resolver.cs#L9709-L9727

cpitclaudel:
This is related to #1427 and #1429, but not exactly equivalent, because it involves marking datatype parameters, rather than the initial check on datatype arguments.

cpitclaudel:
Merging this issue with #1427

PR #2768, opened 2022-09-20
Debug Equality

### LINKED ISSUE #1427: datatype doesn't correctly compute equality-support
In the following, T doesn't have equality support, so X shouldn't either:

```dafny
type T

method m(t1: T, t2: T) {
  var x := t1 == t2; // correctly gives error
}

datatype X = X(t: T)

method m2(x1: X, x2: X) {
  var y := x1 == x2; // doesn't error but should
}
```

As far as I can tell, it only seems to apply to opaque types (e.g., if T is a datatype with a ghost field or something, then X will correctly compute that it doesn't have equality support).

jtristan:
Fixes #1427  #1429


jtristan:
- The two soundness bugs addressed by this PR are due to types incorrectly determined to potentially support equality, when they definitely cannot. 
- In essence, the fix consists in making the test of equality non-support more stringent, as follows

1. If a datatype D refers to an opaque type that has not been declared to support equality, then D cannot support equality
2. The test is made recursive, considering **both type operators and type operands**

- For this fix to be complete, we also needed to update the calculation of the datatype graph so that it contains edges not only for type operators, but also for type operands.

Source/DafnyCore/Resolver.cs
```diff
@@ -9744,6 +9744,9 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
           dependencies.AddVertex(idt);
           foreach (Formal p in ctor.Formals) {
             AddDatatypeDependencyEdge(idt, p.Type, dependencies);
+            foreach (Type t in p.Type.TypeArgs) {
+              AddDatatypeDependencyEdge(idt, t, dependencies);
+            }
```
cpitclaudel, line 9749: Is this needed only for datatypes?  (not for codatatypes?)

Source/DafnyCore/Resolver.cs
```diff
@@ -9941,6 +9944,51 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
       return true;
     }
 
+    bool CheckIfEqualityIsDefinitelyNotSupported(IndDatatypeDecl dt, Graph<IndDatatypeDecl/*!*/>/*!*/ dependencies) {
+
```
cpitclaudel, line 9948:
```suggestion
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9941,6 +9944,51 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+      var scc = dependencies.GetSCC(dt);
+
+      foreach (var ctor in dt.Ctors) {
+
```
cpitclaudel, line 9952:
```suggestion
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9941,6 +9944,51 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+          return true;
+        }
+
+        foreach (var arg in ctor.Formals) {
+          var anotherIndDt = arg.Type.AsIndDatatype;
+          if (arg.IsGhost ||
+              (anotherIndDt != null && anotherIndDt.EqualitySupport == IndDatatypeDecl.ES.Never) ||
+              arg.Type.IsCoDatatype ||
+              arg.Type.IsArrowType ||
+              (arg.Type.IsOpaqueType && !arg.Type.SupportsEquality)) {
```
cpitclaudel, line 9963: What does `arg.Type.SupportsEquality` say for codatatypes and arrow types?  If it already says `false`, could we simplify the above to 
```suggestion
              !arg.Type.SupportsEquality) {
```
?

cpitclaudel, line 9963: Also, I think using `arg.Type` here isn't safe; you want to expand the type first to make sure there's no subset type hiding the opaque type.

Source/DafnyCore/Resolver.cs
```diff
@@ -9941,6 +9944,51 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+          if (anotherIndDt != null && !scc.Contains(anotherIndDt)) {
+            CheckIfEqualityIsDefinitelyNotSupported(anotherIndDt, dependencies);
+          }
+
+          foreach (var type in arg.Type.TypeArgs) {
+            var anotherIndDt_arg = type.AsIndDatatype;
+            if ((anotherIndDt_arg != null && anotherIndDt_arg.EqualitySupport == IndDatatypeDecl.ES.Never) ||
+                type.IsCoDatatype ||
+                type.IsArrowType ||
+                (type.IsOpaqueType && !arg.Type.SupportsEquality)) {
```
cpitclaudel, line 9976: This code seems duplicated from above, let's refactor it?

Source/DafnyCore/Resolver.cs
```diff
@@ -9941,6 +9944,51 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+              arg.Type.IsArrowType ||
+              (arg.Type.IsOpaqueType && !arg.Type.SupportsEquality)) {
+            return true;
+          }
+
+          if (anotherIndDt != null && !scc.Contains(anotherIndDt)) {
+            CheckIfEqualityIsDefinitelyNotSupported(anotherIndDt, dependencies);
+          }
+
+          foreach (var type in arg.Type.TypeArgs) {
```
cpitclaudel, line 9971: I think this check is too strict: we should only consider the cases where the type argument is needed for equality support

Source/DafnyCore/Resolver.cs
```diff
@@ -9941,6 +9944,51 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+            var anotherIndDt_arg = type.AsIndDatatype;
+            if ((anotherIndDt_arg != null && anotherIndDt_arg.EqualitySupport == IndDatatypeDecl.ES.Never) ||
+                type.IsCoDatatype ||
+                type.IsArrowType ||
+                (type.IsOpaqueType && !arg.Type.SupportsEquality)) {
+              return true;
+            }
+
+            if (anotherIndDt_arg != null && !scc.Contains(anotherIndDt_arg)) {
+              CheckIfEqualityIsDefinitelyNotSupported(anotherIndDt_arg, dependencies);
```
cpitclaudel, line 9981: The result of this call is not used.  Does it have side effects?

Source/DafnyCore/Resolver.cs
```diff
@@ -9941,6 +9944,51 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+          if (arg.IsGhost ||
+              (anotherIndDt != null && anotherIndDt.EqualitySupport == IndDatatypeDecl.ES.Never) ||
+              arg.Type.IsCoDatatype ||
+              arg.Type.IsArrowType ||
+              (arg.Type.IsOpaqueType && !arg.Type.SupportsEquality)) {
+            return true;
+          }
+
+          if (anotherIndDt != null && !scc.Contains(anotherIndDt)) {
+            CheckIfEqualityIsDefinitelyNotSupported(anotherIndDt, dependencies);
```
cpitclaudel, line 9968: The result of this call is not used.  Does it have side effects?

cpitclaudel (changes requested):
This is a great improvement but a few things are still missing.

First, the check is a bit too strict: not all type parameters participate in equality, and hence not we do not need to check for equality (or add a dependency edge) between all type parameters.  Your code rejects this, but I think it should be accepted?

```dafny
type DoesNotSupportEq = int -> int
type AlwaysSupportsEq<A> = int
datatype SupportsEqDespiteParameter =
  SupportsEq(a: AlwaysSupportsEq<DoesNotSupportEq>)
type T_SupportsEqDespiteParameter(==) =
  SupportsEqDespiteParameter
```

On the other hand, the check is not strict enough wrt opaque types, because it does not skip past subtype constraints.  Your code accepts this, but I think it should be rejected:

```
type foo
type foo' = f: foo | true witness *
datatype NeverEq' = NeverEq'(f: foo')
type T_NeverEq'(==) = NeverEq'
```

I believe this PR covers the same issues as https://github.com/dafny-lang/dafny/pull/2066, so it would be good to see what is done there (and probably import the tests, too).

In particular, I think you may want to use the `NecessaryForEqualitySupportOfSurroundingInductiveDatatype` field of type arguments, but only in the case where the datatype is marked as `ConsultTypeArgs`.

Source/DafnyCore/Resolver.cs
jtristan: That's an interesting point. I need to think more carefully about the different ways in which it could happen. In the example you give, where the type parameter is not used, I wonder: should we actually accept such a program in the first place? I guess that it doesn't hurt to have an unused type parameter, but isn't it always a typo?

Source/DafnyCore/Resolver.cs
jtristan: Major mistake, needs to be fixed.

Source/DafnyCore/Resolver.cs
jtristan: Agreed.

cpitclaudel:
Another thing I noticed while experimenting with this code is that it breaks the build in Dafny-in-Dafny; I minimized the issue down to this:

```
module Types {
  datatype ClassType =
    ClassType(typeArgs: Type)

  datatype Type =
    | Bool
    | Class(classType: ClassType)
}

module Exprs {
  import Types

  datatype BuiltinFunction =
    | Display(ty: Types.Type)
}
```

The new code enters an infinite loop on this code.  It would be good to add it as a test in the final PR.

Source/DafnyCore/Resolver.cs
cpitclaudel: Refining a non-parametric type with a non-parametric one would be one non-typo example (but maybe not a very interesting one).

There's already code to track which type parameters are needed for equality, btw (see `NecessaryForEqualitySupportOfSurroundingInductiveDatatype`)

jtristan: Is there a type in your first sentence? :)

cpitclaudel: There was, fixed!

Source/DafnyCore/Resolver.cs
jtristan: That's what I did, initially, but this turns out to be incorrect, I think. The reason being that some types are initialized with SupportsEquality set to false, and because of recursive definitions, this Boolean may not have been resolved yet. Thoughts?

Source/DafnyCore/Resolver.cs
jtristan: This is a good question, and I do not know. Let me summarize:

1. With respect to equality support specifically, it is indeed not needed because codatatypes do not support equaity, so there's no need to dig deeper.
2. The dependency graph computation has a comment: "_The dependencies of interest among codatatypes are just the top-level types of parameters_" 

So I think that it is best not to change the computation for the codatatypes and remember that the name of dependency graph is slightly misleading because not all dependencies are included.

Source/DafnyCore/Resolver.cs
jtristan: I also like that by writing the code this way, even if it might be less efficient, the code reads more like its specification of what it should do.

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
       return true;
     }
 
-    bool CheckIfEqualityIsDefinitelyNotSupported(IndDatatypeDecl dt, Graph<IndDatatypeDecl/*!*/>/*!*/ dependencies) {
+    /// <summary>
+    /// The following two functions `BasicCheckIfEqualityIsDefinitelyNotSupported` and `CheckIfEqualityIsDefinitelyNotSupported`
+    /// are mutually recursive and their purpose is to determine if a datatype cannot possibly support equality.
+    /// The entry point is always functoin `CheckIfEqualityIsDefinitelyNotSupported`
```
cpitclaudel, line 9950:
```suggestion
    /// The entry point is always function `CheckIfEqualityIsDefinitelyNotSupported`
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+    /// Function `BasicCheckIfEqualityIsDefinitelyNotSupported` cheks all other types
+    /// 
+    /// As a first approximation, a datatype definely cannot support equality if any of the following holds:
+    /// Any of its constructors is ghost
+    /// Any of the formal parameters of Any of its constructors is ghost
+    /// It is defined in terms of another type that definitely does not support equality, that is:
+    /// A datatype that does not support equality
+    /// An opaque type that has not been declared as supporting equality
+    /// A codatatype
+    /// An arrow type
```
cpitclaudel, line 9961: Given this list, shouldn't definitely supporting equality or not be computed by a member function defined in the `Type` interface, rather than a recursive function with type cases?

Source/DafnyCore/Resolver.cs
cpitclaudel:
```suggestion
    /// The entry point is always function `CheckIfEqualityIsDefinitelyNotSupported`
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+    ///
+    /// These checks need to be made recursively, for example through type synonyms (including subset types), but with the following remarks:
+    /// We need to pay attention to recursively defined datatypes (which form a strongly connected component in the datatype dependency graph)
+    /// We assume that datatypes coming from other modules have already been fully resolved
```
cpitclaudel, line 9965:
```suggestion
    /// - We need to pay attention to recursively defined datatypes (which form a strongly connected component in the datatype dependency graph).
    /// - We assume that datatypes coming from other modules have already been fully resolved.
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+    ///
+    /// </summary>
+    bool BasicCheckIfEqualityIsDefinitelyNotSupported(Type type, Graph<IndDatatypeDecl/*!*/>/*!*/ dependencies, List<IndDatatypeDecl> scc, ModuleDefinition M) {
```
cpitclaudel, line 9968:
```suggestion
    bool BasicCheckIfEqualityIsDefinitelyNotSupported(Type type, Graph<IndDatatypeDecl> dependencies, List<IndDatatypeDecl> scc, ModuleDefinition M) {
```

I don't think it's useful to have these annotations given that they are not checked

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+
+      // The simplest case: these types never support equality
+      if (type.IsArrowType || type.IsCoDatatype || (type.IsOpaqueType && !type.SupportsEquality)) {
```
cpitclaudel, line 9971: Is the type here expected to be normalized / expanded?  If so, wouldn't it be enough to write `!type.SupportsEquality`?

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+        return true;
+      }
+
+      // If the type is a datatype:
+      // Either it has already been computed (either because it comes from another module or because it came before
+      // in topological order with respect to the dependency graph)
+      // Or, we need to recursively check, unless it is in the same SCC
+      var asIDT = type.AsIndDatatype;
+      if (asIDT != null) {
```
cpitclaudel, line 9980:
```suggestion
      if (type.AsIndDatatype is {} asIDT) {
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+        if (asIDT.EnclosingModuleDefinition != M) {
+          if (asIDT.EqualitySupport == IndDatatypeDecl.ES.Never) {
+            return true;
+          } else {
+            return false;
+          }
```
cpitclaudel, line 9986:
```suggestion
          return asIDT.EqualitySupport == IndDatatypeDecl.ES.Never
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+        }
+        if (asIDT.EqualitySupport == IndDatatypeDecl.ES.Never) {
+          return true;
+        } else if (asIDT.EqualitySupport == IndDatatypeDecl.ES.ConsultTypeArguments) {
+          return false;
+        } else {
+          if (!scc.Contains(asIDT) && CheckIfEqualityIsDefinitelyNotSupported(asIDT, dependencies)) {
+            return true;
+          }
```
cpitclaudel, line 9995:
```suggestion
        } else if (!scc.Contains(asIDT) && CheckIfEqualityIsDefinitelyNotSupported(asIDT, dependencies)) {
            return true;
        }
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+        }
+      }
+
+      // For synonym types, we need to continue the check on the supporting type
+      var asSyn = type.AsTypeSynonym;
+      if (asSyn != null) {
```
cpitclaudel, line 10001: I think you can / should use `NormalizeExpand` instead of `AsTypeSynonym`.  Would that work here?

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
-                (type.IsOpaqueType && !arg.Type.SupportsEquality)) {
-              return true;
-            }
+          // Second, we check all the parameters of the type constructor
+          foreach (var typea in arg.Type.TypeArgs) {
 
-            if (anotherIndDt_arg != null && !scc.Contains(anotherIndDt_arg)) {
-              CheckIfEqualityIsDefinitelyNotSupported(anotherIndDt_arg, dependencies);
+            // There is one subtlety: sometimes, we do not need to check the arguments
+            // It is for example the case for parameters that are not used, or class and trait types
```
cpitclaudel, line 10036: Is this entirely true (If I'm reading correctly, for classes, we don't assume that we have equality unless the class is revealed in the current context?  At least that's what I understand from reading the `SupportsEquality` code in `AST/`)

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+            var red = type.AsRedirectingType;
+            var meaningfulArg = true;
+            if (red != null) {
+              foreach (var typep in red.TypeArgs) {
+                if (!typep.NecessaryForEqualitySupportOfSurroundingInductiveDatatype) {
+                  meaningfulArg = false;
+                }
+              }
             }
```
cpitclaudel, line 10045:
```suggestion
            var meaningfulArg = red == null || red.TypeArgs.All(tp => typep.NecessaryForEqualitySupportOfSurroundingInductiveDatatype)
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
-              return true;
-            }
+          // Second, we check all the parameters of the type constructor
+          foreach (var typea in arg.Type.TypeArgs) {
 
-            if (anotherIndDt_arg != null && !scc.Contains(anotherIndDt_arg)) {
-              CheckIfEqualityIsDefinitelyNotSupported(anotherIndDt_arg, dependencies);
+            // There is one subtlety: sometimes, we do not need to check the arguments
+            // It is for example the case for parameters that are not used, or class and trait types
+            var red = type.AsRedirectingType;
```
cpitclaudel, line 10037: What does `AsRedirectingType` do?

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+        if (BasicCheckIfEqualityIsDefinitelyNotSupported(asSyn.Rhs, dependencies, scc, M)) {
+          return true;
+        }
+      }
+
+      return false;
+
+    }
 
+    bool CheckIfEqualityIsDefinitelyNotSupported(IndDatatypeDecl dt, Graph<IndDatatypeDecl/*!*/>/*!*/ dependencies) {
```
cpitclaudel, line 10011: I'm a bit confused by the polarity of the check.  Why/When do we want to check that a type definitely doesn't support equality? (As opposed to "may not support equality?")

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
 
-            if (anotherIndDt_arg != null && !scc.Contains(anotherIndDt_arg)) {
-              CheckIfEqualityIsDefinitelyNotSupported(anotherIndDt_arg, dependencies);
+            // There is one subtlety: sometimes, we do not need to check the arguments
+            // It is for example the case for parameters that are not used, or class and trait types
+            var red = type.AsRedirectingType;
+            var meaningfulArg = true;
+            if (red != null) {
+              foreach (var typep in red.TypeArgs) {
+                if (!typep.NecessaryForEqualitySupportOfSurroundingInductiveDatatype) {
```
cpitclaudel, line 10041: I don't understand what this part of the code does.  This `typep` loop is inside the `typea` loop, but `typep` only depends on `red`, which does not depend on `type`.  So, doesn't `meaningfulArg` end up being the same for each argument?

cpitclaudel (changes requested):
I've gone through the code, but I don't understand everything fully yet.  In particular:

- I'm puzzled by the loop on the type arguments of the original type inside the loop on the type arguments of the construct's argument, 
- I'm confused by the use of AsRedirectingType, which is an internal property that I'm not sure we should be using here.
- I don't fully understand why we're computing whether a type definitely can't support equality (instead of possibly can't support equality / definitely supports equality)

cpitclaudel:
Following up on my confusion, here's a program that still crashes in the current implementation:

```
class C<T> {}
datatype Pair<T, Q> = P2(T, Q)
type S<T, Q> = Pair<C<T>, Q>
datatype DS<T, Q> = DS(s: S<T, Q>)
type Final(==) = DS<int, int -> int>

function method MakeId(u: int): int -> int { x => u }

method Main() {
  var c := new C<int>;
  var f := MakeId(0);
  var g := MakeId(0);
  assert f == g;
  var ds0 := DS(P2(c, f));
  var ds1 := DS(P2(c, g));
  if (ds0 != ds1) {
    assert false;
    print 1 / 0;
  }
}
```

Source/DafnyCore/Resolver.cs
```diff
@@ -9944,43 +9944,109 @@ public CheckDividedConstructorInit_Visitor(Resolver resolver)
...
+    /// are mutually recursive and their purpose is to determine if a datatype cannot possibly support equality.
+    /// The entry point is always functoin `CheckIfEqualityIsDefinitelyNotSupported`
+    /// Function `CheckIfEqualityIsDefinitelyNotSupported` checks datatypes specifically
+    /// Function `BasicCheckIfEqualityIsDefinitelyNotSupported` cheks all other types
+    /// 
+    /// As a first approximation, a datatype definely cannot support equality if any of the following holds:
+    /// Any of its constructors is ghost
+    /// Any of the formal parameters of Any of its constructors is ghost
+    /// It is defined in terms of another type that definitely does not support equality, that is:
+    /// A datatype that does not support equality
```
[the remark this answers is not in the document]
jtristan: My understanding is that it should indeed be "definitely doesn't". The test serves as a filter, and datatypes that definitely cannot support equality should not undergo the rest of DetermineEqualitySupport. This design decision existed before I worked on this fix.

Source/DafnyCore/Resolver.cs
jtristan: I don't know. I like the way it is written now because the code almost reads as a spec, but I assume it is not idiomatic. I will leave it to you to decide what style you prefer.

Source/DafnyCore/Resolver.cs
jtristan: We only want to check when a datatype definitely doesn't support equality.

Source/DafnyCore/Resolver.cs
jtristan: I don't think it is enough, and it's easier to read that way.

Source/DafnyCore/Resolver.cs
jtristan: Thank you, you found a mistake. I need to process the type arguments and their corresponding parameters one by one.

Source/DafnyCore/Resolver.cs
jtristan: When a constructor has a formal of type A<B,C,D>, for some type A<E,F,G> we need to process both the arguments (B, C, D) and the parameters (E, F, G). AsRedirectingType is the way I have found to access the definition of type A to obtain information on its parameters. If there's a better way to do that, I would like to know.

Source/DafnyCore/Resolver.cs
jtristan: I'm not sure. I wish the reference manual laid down all the rules.

Source/DafnyCore/Resolver.cs
jtristan: This code changed a lot, suggestion not applicable anymore.

jtristan:
To give some context, the goal of this PR is not to fix all know issues with the equality support. For example, I am not trying to address #2070. The goal is only to address some issues such as #1427 or #1429, where the problem is that the existing code that filters out datatypes that are known for sure not to support equality is too weak. It is possible that it is not overall the best approach to first filter out such datatypes, but my goal has not been to rethink the equality support determination entirely, only to fix errors in the existing one.

jtristan:
> I've gone through the code, but I don't understand everything fully yet. In particular:
> 
> * I'm puzzled by the loop on the type arguments of the original type inside the loop on the type arguments of the construct's argument,

As you should, it was an error. 

> * I'm confused by the use of AsRedirectingType, which is an internal property that I'm not sure we should be using here.

I do this to access the type parameters of the underlying type definition, but I will try to see if there is a better way to do this.

> * I don't fully understand why we're computing whether a type definitely can't support equality (instead of possibly can't support equality / definitely supports equality)

CF other comment on scope of this PR

jtristan:
It was decided not to pursue with this PR and that #1427 #1429 and #2070 should all be solved at once by determining what the precise rules should be for equality support and then implementing these rules as part of the supportEquality method.

PR #2770, opened 2022-09-20
fix: Use the right comparison operators for ordinals in Go

### LINKED ISSUE #2769: Broken compilation of ordinals in Go
Input:

```dafny
method Main() {
  print 24 as ORDINAL <= 1507 as ORDINAL, "\n";
}
```

Output:

```
ordinal-go/src/ordinal.go:46:40: invalid operation: dafny.IntOfInt64(24) <= dafny.IntOfInt64(1507) (operator <= not defined on struct)
```

cpitclaudel:
Closes #2769.


cpitclaudel:
Added to #2716 instead.

PR #2787, opened 2022-09-22
Publish DafnyServer/Driver/Core to NuGet as well

### LINKED ISSUE #2776: DafnyLanguageServer NuGet package broken because DafnyServer isn't published
See https://www.nuget.org/packages/DafnyLanguageServer: clicking on the DafnyServer link under Dependencies leads to a 404, and similarly a test project that depends on DafnyLanguageServer fails to restore for the same reason.

I'm not yet sure why this isn't an issue with DafnyPipeline depending on DafnyCore, which also isn't published. It seems like the packaging process just inlined DafnyCore there instead.

We also need a mechanism similar to `release-downloads-nuget.yml` that attempts to use the other packages we are publishing, to catch issues like this ideally during the nightly build, or at the very least immediately after releasing.

robin-aws:
Fixes #2776: the root cause is that `DafnyLanguageServer` depends on `DafnyServer`, but the latter was never published. Also publishes the new `DafnyDriver` and `DafnyCore` packages I extracted out earlier, as otherwise the same problem would affect more packages on the next release.

Also replaces `version.cs` with the more automatic mechanism of a `Build.Directory.props` file, just as Boogie uses. The side effect is that the version is more consistently aligned across all the published packages. In particular it changes the current `DafnyRuntime` version from `1.2.0` to `3.8.1.40901` (or rather `3.9.0` by the time we release this).

I managed to augment the existing GHA that verifies the `dafny` dotnet tool to also do a very superficial verification that each library we publish can at least resolve all of its dependencies through NuGet.


.github/workflows/release-downloads-nuget.yml
```diff
@@ -118,3 +128,34 @@ jobs:
...
       run: |
         dafny /compile:3 /spillTargetCode:3 /compileTarget:java a.dfy
+
+  test-dafny-libraries:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        library-name: [ DafnyPipeline, DafnyLanguageServer, DafnyRuntime, DafnyCore, DafnyDriver ]
```
atomb, line 138: Should this include `DafnyServer`?

Not that we have to do it as part of this PR, but it would be great if there were a way to automatically populate this list.

atomb (changes requested):
This looks good in itself, but I just noticed one other reference to `version.cs`, in `Source/Dafny.sln`. That should probably go away, too.

.github/workflows/release-downloads-nuget.yml
robin-aws: Yup, definitely, will add.

I made an attempt at automatically populating this, but it's a bit non-trivial since `dotnet install Dafny` never works, as it complains that you're supposed to `dotnet tool install` that instead, so I'd at least need to special-case that package. Added a comment to that effect too.

atomb (approved):

PR #2789, opened 2022-09-22
Chore: Tutorial on proven fixpoints combinators

MikaelMayer:
This PR adds a tutorial on how to prove that fixpoint combinators can finish in Dafny.

### DIFF
--- a/Test/tutorial/ProvenTerminatingFixpoint.dfy
+++ b/Test/tutorial/ProvenTerminatingFixpoint.dfy
@@ -1,0 +1,259 @@
+// RUN: %dafny_0 /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+/// # Fixpoint combinators with provable termination guarantees
+///
+/// *Mikael Mayer*, 2022/09/22
+///
+/// Fixpoints are functions combinators such that, if they exist,
+/// then we don't need to define recursion or mutual recursion at all,
+/// only use fixpoints when we need recursion.
+/// Let's start by illustrating an example
+
+module {:options "-functionSyntax:4"} Fixpoints {
+
+/// We first define the usual Fibonacci function
+/// To make Dafny happy, we need to provide a (normally implicit) decreases clause,
+/// and Dafny will verify that 0 <= n-1 < n and 0 <= n-2 < n when n > 2
+
+  function Fib1(n: nat): nat
+    decreases n
+  {
+    if n <= 1 then 1 else Fib1(n-1) + Fib1(n-2)
+  }
+
+/// This is nice, but mathematically speaking, we are still defining a function in terms of itself,
+/// which is not obvious.
+/// Besides a while loop and a stack to implement recursive functions,
+/// there is also another way , using fixpoint combinators.
+///
+/// A fixpoint combinator typically takes a 'template function' that, if provided with the final function and an input,
+/// execute the body of the function we want where they replace recursive calls with the provided final functions.
+/// We can write a fibonacci function using a combinator like this:
+
+  function Fib2(n: nat): nat {
+    var comb :=
+      (f: nat -> nat, n: nat) => if n <= 1 then 1 as nat else f(n-1) + f(n-2);
+    fixpoint2(comb, n)
+  }
+
+/// The signature of this fixpoint combinator could typically be:
+
+  function fixpoint2(comb: (nat -> nat, nat) -> nat, u: nat): nat
+
+/// However, any attempt at writing the body of fixpoint will fail, because Dafny will always check for
+/// termination.
+  {
+    var callback := (u': nat) => fixpoint2(comb, u');
+    comb(callback, u)
+  }
+
+/// So let's fix that. Fix, we need to realize that, in order for fixpoint to finish, we need to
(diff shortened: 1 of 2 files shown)


Test/tutorial/ProvenTerminatingFixpoint.dfy
```diff
@@ -0,0 +1,259 @@
...
+/// However, any attempt at writing the body of fixpoint will fail, because Dafny will always check for
+/// termination.
+  {
+    var callback := (u': nat) => fixpoint2(comb, u');
+    comb(callback, u)
+  }
+
+/// So let's fix that. Fix, we need to realize that, in order for fixpoint to finish, we need to
+/// that u' is less than u. But doing so changes the type of the first argument of comb from nat -> nat
+/// to nat --> nat (see [arrow subset types](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-arrow-subset-types))
```
cpitclaudel, line 53: Missing backticks in a number of places here

Test/tutorial/ProvenTerminatingFixpoint.dfy
```diff
@@ -0,0 +1,259 @@
...
+  function fixpoint3(comb: (nat --> nat, nat) -> nat, u: nat): nat
+    decreases u
+  {
+    var callback := (u': nat) requires u' < u => fixpoint3(comb, u');
+    comb(callback, u)
+  }
+
+/// Everything looks great, but when it comes to defining our fibonacci function, things fail
+/// It says our `comb` function does not satisfy the subset constraints
+/// This is because `comb` is only taking `nat -> nat` as first argument, a total function,
```
cpitclaudel, line 65: Missing periods at the end of the entences.

Test/tutorial/ProvenTerminatingFixpoint.dfy
```diff
@@ -0,0 +1,259 @@
...
+/// So let's fix that. Fix, we need to realize that, in order for fixpoint to finish, we need to
+/// that u' is less than u. But doing so changes the type of the first argument of comb from nat -> nat
+/// to nat --> nat (see [arrow subset types](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-arrow-subset-types))
+/// Let's do the change
+
+  function fixpoint3(comb: (nat --> nat, nat) -> nat, u: nat): nat
+    decreases u
+  {
+    var callback := (u': nat) requires u' < u => fixpoint3(comb, u');
+    comb(callback, u)
```
cpitclaudel, line 60: I'm a bit confused about the names `comb` and `callback`.  Isn't `fixpoint` itself the combinator?

cpitclaudel (changes requested):
I left a few comments in the intro.  I think the tutorial starts a bit too fast, and it's not completely clear from the beginning what we're trying to do.  It would be great, before we can start looking at code, to clearly say "we would like to define a function `fix` such that for all `f, n`, `fix(f)(n) == f(fix(f))(n)`.

Beyond this, I wonder if things could be made more intuitive by starting with total functions instead of partial ones.  You could start with this, which I think is both reasonable and simple:

```dafny
function method TotalFix(f: (nat -> nat, nat) -> nat, n: nat): nat {
  var TotalFix' := (n': nat) => if n' < n then TotalFix(f, n') else 0;
  f(TotalFix', n)
}
```

By bottoming out at `0`, we avoid all termination concerns.  Then, you would ask: how do I eliminate the default case?  The natural way to do this would be:

```dafny
function method UnusablePartialFix(f: (nat --> nat, nat) -> nat, n: nat): nat {
  var PartialFix' := (n': nat) requires n' < n => UnusablePartialFix(f, n');
  f(PartialFix', n)
}
```

… but of course this doesn't let you define any interesting `f`, since `f` doesn't know when it can call the function you pass it, so you would refine it into this:

```dafny
function method UsablePartialFix(f: (nat --> nat, nat) --> nat, n: nat): nat
  requires ???
{
  var PartialFix' := (n': nat) requires n' < n => UsablePartialFix(f, n');
  f(PartialFix', n)
}
```

From there on it's mostly equational reasoning:

1. You want to be able to call `f`, so you must have `PartialFix.requires(f, n) ==> f.requires(PartialFix', n)`
2. Abstracting over `PartialFix'` to remove the recursion, you want `PartialFix.requires(f, n) ==> forall g | ??? :: f.requires(g, n)`, where `???` is a restriction on `g` (the stronger the `???` the better, because it will lead to weaker restrictions on `PartialFix`)
3. You need to be able to plug `PartialFix'` into `g`, so `???` can be as strong as the `requires` of `PartialFix'`.  In other words, the strongest `???` can be is `forall n' | PartialFix'.requires(n') ==> g.requires(n')`
4. The `requires` of `PartialFix'` is `n' => n' < n`, so the condition simplifies to `forall n' | n' < n ==> g.requires(n')`

From that you get the final definition:

```dafny
function method PartialFix(f: (nat --> nat, nat) --> nat, n: nat): nat
  requires forall g: nat --> nat, n: nat ::
    (forall n': nat | n' < n :: g.requires(n')) ==> f.requires(g, n)
{
  var PartialFix' := (n': nat) requires n' < n => PartialFix(f, n');
  f(PartialFix', n)
}
```

This version feels less magical to me.  WDYT?

PR #2790, opened 2022-09-22
chore: improve error message for seq type mismatch

### LINKED ISSUE #1887: Unclear message for sequence type mismatch
Dafny's output for the following program could be more clear:

```
function method test<T>(s: seq<T>) : seq<string>
{
  if |s| == 0 then []
  else [s[0]] + test(s[1..])
}
```

Right now it says:

```
test.dfy(4,9): Error: type does not agree with element type of seq<T> (got seq<char>)
```

What this means is that `s[0]` has type `T` but Dafny is expecting type `char`.  Maybe something like this would be better?

> Type mismatch: s[0] has type `T` instead of the expected `char` (`seq<T>` is not compatible with the expected `seq<char>` aka `string`)

alex-chew:
Fixes #1887.

This adds a wrapper to "lazily" retrieve the string representation of a sequence's element type that is unresolved while the type constraint (and its error message) is being constructed.

For simplicity, instead of the originally proposed message

> s[0] has type T instead of the expected char (seq&lt;T> is not compatible with the expected seq&lt;char> aka string)

this PR changes the sequence-type-mismatch error message to

> selected element has type T which is incompatible with expected type char (seq&lt;T> is incompatible with seq&lt;char>)

when a type mismatch arises from a single selection (e.g. `s[0]`).

When a type mismatch arises from a "multi-selection" (e.g. `s[0..2]`), the expected type is an `InferredTypeProxy` that may not resolve to a sequence type at all, in which case we can't speak of the expected sequence type's type argument. For that reason the multi-selection type mismatch error message makes no mention of the expected type argument:

> multi-selection has type seq&lt;T> which is incompatible with expected type seq&lt;char>


Source/DafnyCore/Resolver.cs
```diff
@@ -15725,6 +15725,14 @@ public ResolutionContext(ICodeContext codeContext, bool isTwoState)
       var types = new Type[] { type0, type1 };
       AllXConstraints.Add(new XConstraint(tok, constraintName, types, errMsg));
     }
+    private void AddXConstraint(IToken tok, string constraintName, Type[] types, string errMsgFormat) {
```
cpitclaudel, line 15728: Are you using this new function?

alex-chew: Oops, good catch - I ended up not needing this. I've removed it now.

Source/DafnyCore/Resolver.cs
```diff
@@ -17661,7 +17669,21 @@ class DefaultValueSubstituter : Substituter {
...
+      }
+    }
+
+    /// <summary>
+    /// Allows to lazily call ToString on the given (sequence) type's element type.
+    /// 
+    /// When adding type constraints, a sequence type's element type may not be resolved,
+    /// so attempting to eagerly call ToString on it (for use in an error message) results in a null reference.
+    /// By wrapping the sequence type in this object, the error handling logic will only call ToString
+    /// on the element type after it is resolved.
```
cpitclaudel, line 17682: Very nifty.  I would just suggest rephrasing the comment:
```suggestion
    /// When adding type constraints, a sequence type's element type (`AsSeqType.Arg`) may not be resolved,
    /// so attempting to eagerly read it (for use in an error message) results in a null reference.
    /// By wrapping the sequence type in this object, the error handling logic will only read `AsSeqType.Arg`
    /// on the element type after it is resolved.
```

cpitclaudel (changes requested):
One more small thing that I just caught; sorry!  IIUC the problem isn't ToString, it's `.AsSeqType.Arg`, right?

Source/DafnyCore/Resolver.cs
alex-chew: Yes, that's a better way of putting it. Thanks!

cpitclaudel (approved):
🎉

PR #2793, opened 2022-09-23
Inconsistencies: ref + lambda

jtristan:
Experimenting with different ways to prevent some know flaws and their impact on existing code


jtristan:
Replaced with newer PR

PR #2795, opened 2022-09-23
Build DafnyRuntime for netstandard2.0 and net452 instead of net6.0

### LINKED ISSUE #2779: DafnyRuntime not compatible with .NET frameworks earlier than .NET 6.0
Result of being too restrictive in the DafnyRuntime.csproj configuration. Blocks adoption by projects that need to support earlier versions such as https://github.com/aws/aws-encryption-sdk-dafny.

robin-aws:
Fixes #2779. 

Another attempt after the previous attempt (#2604) had to be reverted (#2710). I tried this with a fresh clone and didn't run into the Rider problem described in #2710 - I'm relatively confident the `<TargetFramework>net6.0</TargetFramework>` I've removed from `Directory.Build.props` was to blame, and/or just needing to thoroughly clean `bin` and `obj` directories.

Note that I am partially relying on post-hoc testing of the nightly build packages after this is merged, to ensure the `net452` build actually works on Windows.

Note also that the packaging warnings I refer to in one of the commit messages are an existing, orthogonal issue: https://github.com/dafny-lang/dafny/issues/3069


keyboardDrummer (dismissed):

keyboardDrummer (dismissed):

cpitclaudel (approved):

cpitclaudel (approved):
👍

Scripts/package.py
```diff
@@ -127,21 +127,23 @@ def zipify_path(fpath):
         """Zip entries always use '/' as the path separator."""
         return fpath.replace(os.path.sep, '/')
 
-    def run_publish(self, project):
+    def run_publish(self, project, framework = None):
```
cpitclaudel, line 130:
```suggestion
    def run_publish(self, project, framework=None):
```

Scripts/package.py
```diff
@@ -127,21 +127,23 @@ def zipify_path(fpath):
...
         env = dict(os.environ)
         env["RUNTIME_IDENTIFIER"] = self.target
         flush("   + Publishing " + project)
         remaining = 3
         exitStatus = 1
         while 0 < remaining and exitStatus != 0:
             remaining -= 1
+            otherArgs = []
+            if framework:
+                otherArgs += ["-f", framework]
```
cpitclaudel, line 140:
```suggestion
            framework_args = ["-f", framework] if framework else []
```

Scripts/package.py
```diff
@@ -127,21 +127,23 @@ def zipify_path(fpath):
...
             exitStatus = subprocess.call(["dotnet", "publish", path.join(SOURCE_DIRECTORY, project, project + ".csproj"),
                 "--nologo",
-                "-f", "net6.0",
                 "-o", self.buildDirectory,
                 "-r", self.target,
                 "--self-contained",
-                "-c", "Release"], env=env)
+                "-c", "Release"] + otherArgs, env=env)
```
cpitclaudel, line 146:
```suggestion
                "-c", "Release", *framework_args], env=env)
```

cpitclaudel (approved):
(Adding a few suggestions, but they are not worth delaying the merge over.)

Scripts/package.py
robin-aws: Ah thank you, I thought I recalled something like that. If the CI has to be rerun anyway I'll apply this :)

PR #2798, opened 2022-09-24
Creating structure for versioned snapshots of user documentation

davidcok:


docs/.gitignore
```diff
@@ -1,3 +1,5 @@
+**/.DS_Store
+Gemfile.lock
```
robin-aws, line 2: Nope, this must be checked in to do its job (locking down exact dependency versions so that builds are reproducible)

docs/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
robin-aws, line 1: I assume this is just here so that the snapshot has a license attached too? Could it just be at the very top level of dafny-lang.github.io instead?

docs/Snapshots.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: default
+---
+<link rel="stylesheet" href="assets/main.css">
+
+## Dafny Documentation snapshots
+
+- [Current development version](https://dafny.org/dafny)
+- [Latest release snapshot](https://dafny.org)
+- [Pre v. 3.8.2](https://dafny.org/V-prehistory)
```
robin-aws, line 10: I'm not convinced we need this snapshot, since we can have a regular 3.8.2 (or rather 3.9) snapshot very soon.

docs/index.html
```diff
@@ -15,56 +16,23 @@
...
 <body>
 
 <div>
-<h1>Dafny<br>
-<font size="-1">
-Development version
-</font>
-</h1>
+<h1>Dafny<br><p>
+<!-- VERSION INFO HERE-->Development version
```
robin-aws, line 28: Will this be the target of some pre-processing script then?

docs/HowToFAQ/ERROR_ModifiesValue1.dfy
```diff
@@ -0,0 +1,5 @@
+class A {}
```
robin-aws, line 1: What references this file?

robin-aws (changes requested):
Looks like a good clean-up!

docs/Snapshots.md
robin-aws: I also slightly prefer the pattern `v3.8.2` rather than `V-3.8.2`

RELEASE_NOTES.md
```diff
@@ -77,7 +77,7 @@
 # 3.7.0
 
 - feat: The Dafny CLI, Dafny as a library, and the C# runtime are now available on NuGet. You can install the CLI with `dotnet tool install --global Dafny`. The library is available as `DafnyPipeline` and the runtime is available as `DafnyRuntime`. (https://github.com/dafny-lang/dafny/pull/2051)
-- feat: New syntax for quantified variables, allowing per-variable domains (`x <- C`) and ranges (`x | P(x), y | Q(x, y)`). See [the quantifier domain section](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-quantifier-domains) of the Reference Manual. (https://github.com/dafny-lang/dafny/pull/2195)
```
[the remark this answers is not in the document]
davidcok: Yes, I have a task to find and fix all those links.

docs/LICENSE
davidcok: It is there also. I can delete it from the current doc version if you like, though I'd prefer to keep the current docs and the snapshots as identical as possible.

docs/Snapshots.md
davidcok: It is a placeholder as there are no actual versions yet. I'll delete the V-prehistory when we put a v3.8.2 in place. And I'll use the v3.8.2 syntax at that time.

docs/index.html
davidcok: Yes.

docs/HowToFAQ/ERROR_ModifiesValue1.dfy
davidcok: ERROR_ModifiesValue.md. For some reason ERROR_ModifiesValue1.dfy was missing and the builds of the website (though not the CI builds) were failing as a result.

docs/.gitignore
davidcok: Thanks. fixed in both PRs.

PR #2799, opened 2022-09-26
Allow writing tests with inline range and position definitions

keyboardDrummer:
### Changes

- Introduce `MarkupTestFile.cs`, a file for writing tests with inline range and position definitions.
- Use the markup test feature for some of the tests in `DefinitionTest.cs`
- Change the processing of text diffs coming from the client so the computation is linear in the size of the change, instead of linear in the size of the document + change.
- Change CodeActionHandler and DafnyCompletionHandler so they no longer do a `O(DocumentSize)` computation to convert between positions and indices
- Small change in logging so exceptions are more likely to end up in the log.

### Testing
- Performance changes are untested
- Logging changes are untested
- Other changes are a refactoring, covered by existing tests


Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs
```diff
@@ -29,200 +29,209 @@ public class DefinitionTest : ClientBasedLanguageServerTest {
...
+datatype Colors = Red | [|Green|] | Blue
 
-function Foo(value: Identity<Colors>): bool {
-  match value {
-    case Identity(Red()) => true
-    case Identity(Green) => false // Warning
+function Foo([|value|]: Identity<Colors>): bool {
+  match va$$lue {
+    case Ide$$ntity(Red()) => true
+    case Identity(Gr$$een) => false // Warning
```
MikaelMayer, line 149: I love this new test format ! I can finally understand what each position and range is about :-)
I also see that you can extract named spans in your code, but I don't see them yet in tests....
So I would like to challenge you on this second but final improvement: Rather on referring to these ranges and positions by `ranges[i]` and `positions[j]`, could you use named ranges instead?

```suggestion
function Foo({|0:value|}: Identity<Colors>): bool {
  match va$0$lue {
    case Ide$$ntity(Red()) => true
    case Identity(Gr$$een) => false // Warning
```

The point of using named ranges (and thus named positions), is that your test becomes self-contained in the source: You can just iterate over named positions, query for the definition range, and verify it corresponds to the same named range.
It makes adding new tests extremely easy.

Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
```diff
@@ -106,7 +106,6 @@ public class DafnyTextDocumentHandler : TextDocumentSyncHandlerBase {
         notificationPublisher.HideDiagnostics(documentId);
       } catch (Exception e) {
         telemetryPublisher.PublishUnhandledException(e);
-        logger.LogError(e, "error while closing the document");
```
MikaelMayer, line 109: I'm not sure about that, why did you remove this log? Is it because the logger is not available after refactoring?

Source/DafnyLanguageServer/MarkupTestFile.cs
```diff
@@ -0,0 +1,237 @@
...
+
+      var currentIndexInInput = 0;
+      var inputOutputOffset = 0;
+
+      // A stack of span starts along with their associated annotation name.  [||] spans simply
+      // have empty string for their annotation name.
+      var spanStartStack = new Stack<(int matchIndex, string name)>();
+      var namedSpanStartStack = new Stack<(int matchIndex, string name)>();
+
+      while (true) {
```
MikaelMayer, line 66: So the current algorithm is the following:
* Start with index 0
* Collect the first index at which there is a match for each of $$, [|, |], |}, or {|name:
* Find the first match. If there is an ambiguity such as [|], then try to close existing opened spans or named spans first..
* If it's a position, we add it to the position
* If we can close a span or named span, then we add it. If it's an open span or named span, we put it on a stack.
* In any case, we keep an offset to always keep the right position.

I see that you need to do something like this because you can have positions inside spans, so that makes sense.

However, could you envision the following cleaner code:
* Use `{{name:`, `[[`, `]]`, `}}` or another markup chosen by you so that you just cannot have the `[|]` or `{|]` ambiguities. I also though of `[<` and `>]` and have the positions marked like a cross like `><`, but feel free to use your own ideas.
* Make use of a regular expression that avoids running match four times and one extra time for the regular expression. I'm not concerned about performance, since these are small tests, but more about maintenance::

```
var r = new Regex(@"(?<Position>\$\$)|(?<SpanStart>\[\|)|(?<SpanEnd>\|\])|(?<NameSpanStart>\{\| ([-_.A-Za-z0-9\+]+) \:)|(?<NameSpanEnd>\|\})");

var lastIndex = 0;
foreach(var match in r.matches(input)) {
  if(match.Groups["Position"].Success) {
    ....
  } else if(match.Groups["SpanStart"].Success) {
  } ....

  // Store the string between lastIndex and match.Index
  // Reassign lastIndex
  var lastIndex = match.Index + match.Groups[0].Value.Length;
  // And other offset computations to keep track of positions
}
```

Source/DafnyLanguageServer/Workspace/DocumentTextBuffer.cs
```diff
@@ -0,0 +1,38 @@
...
+  public string Extract(Range range) {
+    return Buffer.Extract(range);
+  }
+
+  public static implicit operator TextDocumentItem(DocumentTextBuffer buffer) => buffer.TextDocumentItem;
+  public string Text => TextDocumentItem.Text;
+  public DocumentUri Uri => TextDocumentItem.Uri;
+  public int? Version => TextDocumentItem.Version;
+
+  public int NumberOfLines => Buffer.Lines.Count;
```
MikaelMayer, line 37: Oh, clever use of `Textbuffer` !

MikaelMayer (changes requested):
I love this new format for the definition tests, and the code for computing newlines is way cleaner.
I have a few suggestions to make the tests even more maintainable.

Source/DafnyLanguageServer/Handlers/DafnyTextDocumentHandler.cs
keyboardDrummer: It's moved into PublishUnhandledException so other locations would also do this logging

Source/DafnyLanguageServer.Test/Lookup/DefinitionTest.cs
keyboardDrummer: Sounds good :)

Source/DafnyLanguageServer/MarkupTestFile.cs
keyboardDrummer: > I also though of [< and >] and have the positions marked like a cross like ><, but feel free to use your own ideas.

I like that. Is the ambiguity a problem that needs to be solved though? I'm hesitant about changing the code since I took this from Roslyn so I'm confident it works and won't need much maintenance.

MikaelMayer: The fact that 15 lines of code are dedicated to solve an ambiguity which has no reason to exist other than an arbitrary design is kind-of bothering me :-) But do you think the proposed fix for the ambiguity requires too much work?

I'm also surprised that Roslyn provided such code that seems underperforming at best (O(N*M) where N is the length of the code and M is the number of spans/positions), but moreover it was very confusing to me when I reviewed it - I kept asking myself why do we do all these matches as well as one regex? Why sorting a list? Why implementing in a worse complexity what a simpler regular expression can do?

Do you think the version I suggested with regex.matches might be less clear, too long or less maintainable ?
I know you already did a lot of effort there so I don't want to ask you too much.

For me, implementing it with regex.matches will make me much more comfortable to approve it than the current code.

keyboardDrummer: Made the change. I picked `[>...<]` and `{>...<}` over `[<...>]` and `{<...>}` because I found it more visually appealing.

Source/DafnyLanguageServer/MarkupTestFile.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+      }
+
+
+      if (spanStartStack.Count > 0) {
+        throw new ArgumentException($"Saw {SpanStartString} without matching {SpanEndString}");
+      }
+
+      if (namedSpanStartStack.Count > 0) {
+        throw new ArgumentException($"Saw {NamedSpanEndString} without matching {NamedSpanEndString}");
```
MikaelMayer, line 88:
```suggestion
        throw new ArgumentException(@"Saw '{<' without matching '>}'");
```
Or you might want to create and use `NamedSpanStartString`

Source/DafnyLanguageServer/MarkupTestFile.cs
```diff
@@ -0,0 +1,183 @@
...
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using Range = OmniSharp.Extensions.LanguageServer.Protocol.Models.Range;
+
+namespace Microsoft.Dafny.LanguageServer {
+  /// <summary>
+  /// To aid with testing, we define a special type of text file that can encode additional
+  /// information in it.  This prevents a test writer from having to carry around multiple sources
+  /// of information that must be reconstituted.  For example, instead of having to keep around the
+  /// contents of a file *and* and the location of the cursor, the tester can just provide a
+  /// string with the "$" character in it.  This allows for easy creation of "FIT" tests where all
```
MikaelMayer, line 26:
```suggestion
  /// string with the "><" string in it.  This allows for easy creation of "FIT" tests where all
```

Source/DafnyLanguageServer/MarkupTestFile.cs
```diff
@@ -0,0 +1,183 @@
...
+  /// that needs to be provided are strings that encode every bit of state necessary in the string
+  /// itself.
+  /// 
+  /// The current set of encoded features we support are: 
+  /// 
+  /// $$ - The position in the file.  There can be at most one of these.
```
MikaelMayer, line 32:
```suggestion
  /// >< - The position in the file.  There can be at most one of these.
```

Source/DafnyLanguageServer/MarkupTestFile.cs
```diff
@@ -0,0 +1,183 @@
...
+  /// 
+  /// [| ... |] - A span of text in the file.  There can be many of these and they can be nested
+  /// and/or overlap the $ position.
+  /// 
+  /// {|Name: ... |} A span of text in the file annotated with an identifier.  There can be many of
```
MikaelMayer, line 37:
```suggestion
  /// {>Name: ... <} A span of text in the file annotated with an identifier.  There can be many of
```

Source/DafnyLanguageServer/MarkupTestFile.cs
```diff
@@ -0,0 +1,183 @@
...
+      var tempSpans = new Dictionary<string, List<TextSpan>>();
+
+      var outputBuilder = new StringBuilder();
+
+      // A stack of span starts along with their associated annotation name.  [||] spans simply
+      // have empty string for their annotation name.
+      var spanStartStack = new Stack<(int matchIndex, string name)>();
+      var namedSpanStartStack = new Stack<(int matchIndex, string name)>();
+
+      var r = new Regex(@"(?<Position>\>\<)|(?<SpanStart>\[\>)|(?<SpanEnd>\<\])|(?<NameSpanStart>\{\>([-_.A-Za-z0-9\+]+)\:)|(?<NameSpanEnd>\<\})");
```
MikaelMayer, line 59:
```suggestion
      var r = new Regex(@"(?<Position>><)|(?<SpanStart>\[>)|(?<SpanEnd><\])|(?<NameSpanStart>\{>([-_.A-Za-z0-9\+]+)\:)|(?<NameSpanEnd><\})");
```
My IDE makes a good job showing me what is a literal character and what is regex construct.

MikaelMayer (changes requested):
Just comments to change for review, but thanks for implementing this regex-based algorithm ! It was very easy to understand.
I'm ok dropping my ask to make tests totally self contained (by introducing a syntax like `>0<`) because as now it's already quite good.

keyboardDrummer:
> I'm ok dropping my ask to make tests totally self contained (by introducing a syntax like `>0<`) because as now it's already quite good.

Are they not self-contained right now? By ordering the ranges it also works out right?

MikaelMayer:
> Are they not self-contained right now? By ordering the ranges it also works out right?

They are almost all self-contained, except one or two tests where the ranges are not in the same order as the positions:
https://github.com/dafny-lang/dafny/pull/2799/files#diff-7be836feeb4b6b75418c6de8a5fdc026e0560d5b08faf4c9e953948efe886c3fR114

But that's ok.

MikaelMayer (approved):

keyboardDrummer:
> > Are they not self-contained right now? By ordering the ranges it also works out right?
> 
> They are almost all self-contained, except one or two tests where the ranges are not in the same order as the positions: https://github.com/dafny-lang/dafny/pull/2799/files#diff-7be836feeb4b6b75418c6de8a5fdc026e0560d5b08faf4c9e953948efe886c3fR114
> 
> But that's ok.

The problem with those tests is that some positions don't map to any range, since they're supposed to not resolve, so we would need something like `>:fail<` as a notation.

MikaelMayer (approved):

PR #2816, opened 2022-09-29
Remove support for :layerQuantifier

### LINKED ISSUE #35: Rename or remove {:layerQuantifier} attribute
Rustan has commented that, if the `{:layerQuantifier}` is to stay at all, that it should be renamed to `{:fuelQuantifier}`. In addition, if this is kept, the reference manual needs a more complete description of the purpose and effect of this attribute. If it is only used internally its name should start with a leading underscore.

davidcok:
Fixes #35


RustanLeino (approved):
I tried a simple program

``` dafny
method M(s: set<int>)
  requires forall x {:layerQuantifier} :: x in s ==> x % 2 == 0
```

which causes Dafny to emit malformed Boogie. So, with no tests and with a crash from a simple attempted use, I think it's fair to assume there are no users of this feature. Hence, I support removing it.

PR #2818, opened 2022-09-30
Use native byte/char arrays in Go

### LINKED ISSUE #2708: Go incorrectly compares array values by contents, instead of by reference
```dafny
method Main() {
  var a1 := new int[] [42];
  var a2 := new int[] [42];
  if a1 == a2 {
    print 1 / 0;
  }
}
```

Outputs `[Program halted] division by zero` when compiled and executed for Go.

Seems to be intentional and hence just a misinterpretation of Dafny semantics: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntime.go#L677-L679

(I'll likely fix this as a side-effect of #2390, but it's not guaranteed)

### LINKED ISSUE #3204: Emitted Java converts long to int and loses precision
### Dafny version

3.9.1

```dafny
newtype PrettyLong = x | 0x1F_FFFF_FFF0 <= x < 0x20_0000_0100 witness 0x1F_FFFF_FFF0 {
  predicate method True() { true }
}

method Main() {
  var s := set l: PrettyLong | 0x20_0000_0002 <= l < 0x20_0000_0008 && l.True() && l != 0x20_0000_0006;
  print s, "\n";
}
```

```code
% dafny /compile:3 /compileTarget:cs test.dfy

Running...

{137438953474, 137438953475, 137438953476, 137438953477, 137438953479}

% dafny /compile:3 /compileTarget:java test.dfy

Wrote textual form of target program to test-java/test.java
Additional target code written to test-java/_System/nat.java
Additional target code written to test-java/_System/PrettyLong.java
Additional target code written to test-java/_System/__default.java
Additional target code written to test-java/dafny/Function0.java
Running...

{}
```

### What happened?

The Java compiler emits code that converts each candidate `PrettyLong` number to a 32-bit int. This means that that none of them falls into the range mentioned by the set. Thus, it erroneously computed the empty set.

Mac

### LINKED ISSUE #3207: Array comparison causes JavaScript run-time crash
### Dafny version

3.9.1

```dafny
newtype byte = x | 0 <= x < 256

method PrintEq<T(==)>(x: T, y: T) {
  print x == y, "\n";
}

method Main() {
  var a := new byte[10];
  var b := new byte[10];

  print a == b, "\n";
  PrintEq(a, b);
}
```

```code
% dafny /compileTarget:js /compile:3 test.dfy

Running...

false
[stdin]:1009
        throw e
        ^

TypeError: a.equals is not a function
    at Object.$module.areEqual ([stdin]:22:16)
    at Function.PrintEq ([stdin]:1092:51)
    at Function.Main ([stdin]:1105:25)
    at [stdin]:1112:53
    at Object.$module.HandleHaltExceptions ([stdin]:1004:7)
    at [stdin]:1112:8
    at Script.runInThisContext (node:vm:131:12)
    at Object.runInThisContext (node:vm:308:38)
    at node:internal/process/execution:81:19
```

### What happened?

The JavaScript `areEqual` method does not properly handle arrays, so a program like the one above crashes at run time.

Mac

RustanLeino:
This PR does three things to improve performance of Go code that uses arrays:

* For nonempty arrays where the target representation of elements uses `uint8` or `rune`, the underlying `Array` uses a `[]uint8` or `[]rune` array, respectively, instead of the default `[]any`.
* For 1-dimensional arrays, the index is passed directly to the `ArrayGet`/`ArraySet` method in the Dafny runtime, rather than using a var-args version of those methods.
* When the static type of the array operations reveals that the underlying representation is `uint8` or `rune`, then special versions of `ArrayGet`/`ArraySet` are called where the argument/return is a `uint8` or `rune`, respectively, rather than the default versions, which use an `any` (that is, `interface{}`) as the argument/return.

Unlike the current compilation to Java, which introduces type descriptors for all types in order to specialize arrays, this PR uses existing type descriptors or the initialization elements provided by the program. The one case where the type cannot be recovered for specialization is for 0-length arrays, which this PR represents with an underlying `nil` array. It seems the design in this PR can be adapted to Java or JavaScript as well.

The performance improvements are as follows. The program run is included below, and the Go measurements were repeated 6 times (each varying less than 0.2s from the numbers shown here).

| target features  | time |
| ------------- | ------------- |
| C#  | 31.4s  |
| Go with specialized arrays and no boxing across Set/Get  | 27.4s  |
| Go with specialized arrays, but still using boxes across Set/Get | 31.8s  |
| Go with arrays of `any`  | 103.6s  |
| Go with arrays of `any` and using var-args even for 1-dim arrays  | 179.1s  |

``` dafny
newtype byte = x | 0 <= x < 256
newtype int32 = x | -0x8000_0000 <= x < 0x8000_0000

method MM<X>(x: X) returns (r: X) { r := x; }

method Main() {
  for i := 0 to 10_000 {
    var a := GenerateArray();
    a := CopyArray(a);
    var s := SumArray(a);
    expect s == 0;
  }
  print "done\n";
}

method GenerateArray() returns (b: array<byte>)
  ensures b.Length == 256_000
{
  b := new byte[256_000];
  for i: int32 := 0 to 256_000 {
    b[i] := (i % 256) as byte;
  }
}

method CopyArray(a: array<byte>) returns (b: array<byte>)
  requires a.Length == 256_000
  ensures b.Length == 256_000
{
  b := new byte[256_000];
  for i: int32 := 0 to 256_000 {
    b[i] := a[i];
  }
}

method SumArray(a: array<byte>) returns (sb: byte)
  requires a.Length == 256_000
{
  var s: int32 := 0;
  for i: int32 := 0 to 256_000
    invariant 0 <= s < 256
  {
    s := s + a[i] as int32;
    if 256 <= s {
      s := s - 256;
    }
  }
  sb := s as byte;
}
```

## Other fixes along the way

As part of writing tests for the new functionality, I detected and fixed various other infelicities as well:

* fix: In Java, cast to non-box type after let
* fix: In JavaScript, compare arrays by comparing their references, not their elements (issue https://github.com/dafny-lang/dafny/issues/3207)
* fix: In C#, add some necessary casts
* In Go, improved implementation of array comparisons
* In Go, fixes array comparisons (to be reference equality, not equality of elements) (https://github.com/dafny-lang/dafny/issues/2708)
* fix: In Java, fix enumerations of `long` Java values (which previously ended up truncating numbers to `int`s)
* fix: In Java, don't confuse a Dafny-defined type `Long` with Java's `java.lang.Long` (and similar for other integer types) (reported as https://github.com/dafny-lang/dafny/issues/3204)
* fix: In Python, don't share inner arrays, which was previously done when arrays have 3 or more dimensions
* Across the target languages, pass BigInteger as array sizes and use native integers for array indices
* Pretty printing: Elide the array size in `new` only when the input program does (this uses `AutoGeneratedToken`, as in other places)
* When there are several `IntBoundedPool`s, try a little harder to pick the best bounds. Alas, there are still simple cases where the target code ends up enumerating all 32-bit integers, for example. To improve the target code further, we need to make use of the partial-evaluation of expressions, which exists elsewhere in the Dafny implementation, and/or make use of run-time `Min` and `Max` routines.

Fixes #3204
Fixes #2708
Fixes #3207


Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -108,16 +108,22 @@ def __lt__(self, other):
...
 
     def __init__(self, initValue, *dims):
-        self.arr = initValue
         self.dims = list(dims)
-        for i in reversed(self.dims):
-            self.arr = Array.Box([copy.copy(self.arr) for _ in range(i)])
+        self.arr = createArrayStructure(initValue, *dims)
 
     def __dafnystr__(self) -> str:
         return '[' + ', '.join(map(string_of, self.arr)) + ']'
```
fabiomadge, line 129:
```suggestion
        return f'array{self.dims}'
```

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -108,16 +108,22 @@ def __lt__(self, other):
...
+        return [createArrayStructure(initValue, *rest) for _ in range(howmany)]
+
 class Array:
     class Box(list):
         def __dafnystr__(self) -> str:
             return '[' + ', '.join(map(string_of, self)) + ']'
 
     def __init__(self, initValue, *dims):
-        self.arr = initValue
         self.dims = list(dims)
```
fabiomadge, line 125:
```suggestion
        def create_structure(initValue, *dims):
            return [initValue if len(dims) <= 1 else create_structure(initValue, *dims[1:]) for _ in range(dims[0])]
        self.dims = list(dims)
```

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -108,16 +108,22 @@ def __lt__(self, other):
...
-        for i in reversed(self.dims):
-            self.arr = Array.Box([copy.copy(self.arr) for _ in range(i)])
+        self.arr = createArrayStructure(initValue, *dims)
```
fabiomadge, line 126:
```suggestion
        self.arr = create_structure(initValue, *dims)
```

fabiomadge (commented):
The shallow copy problem could most likely be solved by reverting `copy.copy` to `copy.deepcopy`. At some point I erroneously convinced myself that it is sufficient.
Getting rid of it is probably for the best, but please also remove the import on L11.

Test/dafny0/ArrayElementInitCompile.dfy
```diff
@@ -1,5 +1,8 @@
 // RUN: %dafny /compile:0 /print:"%t.print" /rprint:"%t.dprint" "%s" > "%t"
 // RUN: %dafny /noVerify /compile:4 /compileTarget:cs "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:java "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:js "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:go "%s" >> "%t"
```
robin-aws, line 5: Would `%testDafnyForEachCompiler` work for this case now then?

Test/comp/Comprehensions.dfy
```diff
@@ -534,3 +535,65 @@ module NewtypeRangeRegression {
     print k, "\n";
   }
 }
+
+module {:options "/functionSyntax:4"} EnumerationRegression {
+  method Test() {
+    // The Java compiler once converted each candidate Long number to a 32-bit int, which meant
+    // that none of them fell into the range mentioned by the set. Thus, it erroneously computed
+    // the empty set.
```
robin-aws, line 543: Worth logging these issues as a soundness bug even if it will be immediately fixed here.

Test/comp/Arrays.dfy
```diff
@@ -374,3 +383,494 @@ module TypeSynonym {
...
+    arr, a, arr[a + b], m, arr, c, m[a, b], b, m := arr'', 100, 120, m'', arr'', 300, c, 200, m'';
+
+    print a, " ", b, " ", c, " ", arr'[46], " ", m'[18, 28], "\n"; // 100 120 200 300 38
+  }
+}
+
+module ArrayToSeq {
+  newtype byte = x | 0 <= x < 256
+
+  trait TraitMeRite {
```
robin-aws, line 532: Open your eyes, maybe you'll see the light... :)

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -161,18 +160,14 @@ def __le__(self, other):
...
+        def create_structure(initValue, *dims):
+            return [initValue if len(dims) <= 1 else create_structure(initValue, *dims[1:]) for _ in range(dims[0])]
         self.dims = list(dims)
-        for i in reversed(self.dims):
-            self.arr = Array.Box([copy.copy(self.arr) for _ in range(i)])
+        self.arr = create_structure(initValue, *dims)
 
     def __dafnystr__(self) -> str:
-        return '[' + ', '.join(map(string_of, self.arr)) + ']'
+        return f'array{self.dims}'
```
robin-aws, line 170: Is this more consistent across backends or less?

Source/DafnyRuntime/DafnyRuntime.js
```diff
@@ -20,7 +20,7 @@ let _dafny = (function() {
       return a === b;
     } else if (BigNumber.isBigNumber(a)) {
       return a.isEqualTo(b);
-    } else if (a._tname !== undefined) {
+    } else if (a._tname !== undefined || (Array.isArray(a) && a.constructor.name == "Array")) {
```
robin-aws, line 23: Also fixing a soundness bug?

robin-aws (commented):
Just some initial comments as I work from the test cases inwards, will finish the review shortly.

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
+  }
+  totalLength := product.Int()
+  return totalLength
+}
+
+// NewArrayFromExample returns a new Array.
+// If "init" is non-nil, it is used to initialize all elements of the array.
+// "example" is used only to figure out the right kind of Array to return.
+// If "init" is non-nil, the types of "example" and "init" must agree.
+func NewArrayFromExample(example interface{}, init interface{}, dims ...Int) Array {
```
robin-aws, line 740: So I mentioned earlier this concern that determining the array element type based on an example value could make the wrong choice: if the Dafny type of the array is `array<T>` where `T` is a trait and `C1` is one class that extends it, and we provide a `C1` instance as the example, we could allocate a type-safe array of `C1`s that then later fails to accept an element of type `C2` (another class that extends `T`).

That may be a reasonable limitation of this technique and may not actually be triggerable YET, but could sneak in if we continue to apply this technique, and especially if the Dafny type system evolves such that we have more cases of subtyping that could trigger it (such as allowing value types to implement traits as well). Is this indeed safe now, and if so how do we ensure it remains safe in the future?

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
+func (_this ArrayStruct) ArrayGet1(index int) interface{} {
+  return _this.contents[index]
+}
+
+func (_this ArrayStruct) ArraySet1(value interface{}, index int) {
+  _this.contents[index] = value
+}
+
+func (_this ArrayStruct) ArrayGet1Byte(index int) byte {
+  panic("cannot specialize general ArrayGet with byte")
```
robin-aws, line 877:
```suggestion
  panic("Expected specialized array type that contains bytes, but found general-purpose array of interface{}")
```

Users should never see this but I think that message helps point debuggers towards what happened better. Similar edits to the other error messages.

An alternative would be to actually downcast to byte, but I do think it's better to panic if this should never happen.

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
+  return _this.contents[lo.Int():hi.Int()]
+}
+
+// ArrayStruct implements the EqualsGeneric interface.
+func (_this ArrayStruct) EqualsGeneric(other interface{}) bool {
+  otherArray, ok := other.(*ArrayStruct)
+  if !ok {
+    return false
   }
+  return &_this.dims[0] == &otherArray.dims[0]
```
robin-aws, line 911: Why not just `&_this == &otherArray`?

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
-func (array *Array) findIndex(ixs ...int) int {
-  i := 0
-  size := 1
-  for d := len(array.dims) - 1; d >= 0; d-- {
-    i += size * ixs[d]
-    size *= array.dims[d]
-  }
-  return i
+func (_this ArrayForByte) ArrayGet1(index int) interface{} {
+  return _this.contents[index]
```
robin-aws, line 930: Shouldn't this also be a panic? Or do we need it to support type variables?

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
+  }
+  return &_this.dims[0] == &otherArray.dims[0]
 }
 
-// Iterator iterates over the array.
-func (array *Array) Iterator() Iterator {
-  return sliceIterator(array.contents)
+/***** ArrayForChar is the Array interface specialized for char. *****/
+
+type ArrayForChar struct {
```
robin-aws, line 980: Likewise we need an ArrayForCodePoint here too

Test/dafny0/ArrayElementInitCompile.dfy
RustanLeino: No. I spent 30+ minutes trying. using `%testDafnyForEachCompiler` causes a failure from the C++ compiler, which doesn't support "Subtraction of non-native type". The new test harness does not allow this, because this particular error is generated by an explicit `throw new ...` rather than a call through the expect method `UnsupportedFeatureError`. I tried to fix this and still keeping the description string "Subtraction of non-native type", but that causes the test harness not to recognize the failure in a different way. So, I tried giving up on the additional descriptive string, but then that shows another failure in the C++ compiler where it apparently outputs an identifier named `_#Func1`.

My conclusion: The new test harness does not handle C++ failures nicely and the C++ compiler does not report everything it doesn't support. I don't have an interest in fixing either of these problems in this PR.

robin-aws (changes requested):
And here's the rest. The only major questions I have are about the future safety of the technique, and support for `--unicode-char`. Otherwise I love the improvement and I'm looking forward to applying it to more backends. I'm actually optimistic of working it into the closely-related #3167 and expressing it efficiently in Dafny itself!

Test/comp/Comprehensions.dfy
RustanLeino: Done. https://github.com/dafny-lang/dafny/issues/3204

Test/comp/Arrays.dfy
RustanLeino: I hit it with my best shot.

Test/comp/Arrays.dfy
```diff
@@ -374,3 +383,494 @@ module TypeSynonym {
...
+    print a == a; // true
+    CheckEquality(a, a); // true
+    var c: array?<byte> := null;
+    print a == c; // false
+    CheckEquality(a, c); // false
+    print c == a; // false
+    CheckEquality(c, a); // false
+  }
+
+  method TestEqualityOfSomeNonArraysToo() {
```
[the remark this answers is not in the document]
RustanLeino: Good call. It's fixed. I added a test case.

Source/DafnyRuntime/DafnyRuntime.py
RustanLeino: More. But, as the following program reveals:

``` dafny
method Main() {
  var a := new int[13];
  var m := new int[6, 28];
  print a, " ", m, "\n";
}
```

we still have ways to go for full consistency. For some targets (Go and half of JavaScript), we still print elements of the arrays. I'd rather we fix this issue in a separate PR.

Source/DafnyRuntime/DafnyRuntime.js
RustanLeino: It had led to a crash at run time for the JavaScript target. I reported it as https://github.com/dafny-lang/dafny/issues/3207.

Test/dafny0/ArrayElementInitCompile.dfy
robin-aws: I appreciate the effort in trying to make it work! The issue is that `%testDafnyForEachCompiler` allows compilers to cleanly not support a feature, but they have to declare it in a specific way so we don't just allow arbitrary exceptions or failures. It sounds like you hit such a case in the C++ compiler - I went through and refactored all the obvious cases to use the right format, but I'm sure there are messy cases remaining, especially cases where it emits bad code instead of throwing an exception (often from code that was copied from another compiler without adjusting it :)

This is fine for now, we'll definitely be doing a pass over the test suite to convert them all to `%testDafnyForEachCompiler` at which point we'll fix the misbehaving compilers.

Source/DafnyCore/Compilers/Compiler-Csharp.cs
```diff
@@ -1975,37 +1976,30 @@ protected class ClassWriter : IClassWriter {
...
-        TrExprAsInt(dimensions[i], wrs[i], inLetExprBody: false, wStmts, true, "C# arrays may not be larger than the max 32-bit integer");
+        wrs[i].Write(dimensions[i]);
       }
     }
 
     private List<ConcreteSyntaxTree> EmitNewArray(Type elmtType, IToken tok, int dimCount, bool mustInitialize, ConcreteSyntaxTree wr) {
+      ConcreteSyntaxTree EmitSizeCheckWrapper(ConcreteSyntaxTree w) {
+        w.Write($"{DafnyHelpersClass}.ToIntChecked(");
+        var wSize = w.Fork();
+        w.Write(", \"array size exceeds memory limit\")");
```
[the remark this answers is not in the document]
RustanLeino: I think we should make this change consistently for all the compilers. That is, I think we should decide on such a maximum limit for each target (either 31, 32, 63, or 64 bits, or unlimited). Then, the compilers can always use the corresponding native word size for array sizes and indices. I suggest this be done as a separate PR. I filed Issue https://github.com/dafny-lang/dafny/issues/3212 to track this.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -707,8 +710,42 @@ protected interface IClassWriter {
...
+    /// Allocates a new array with element type "elementType" and lengths "dimensions" in each dimension.
+    /// Note that "elementType" may denote a type parameter.
+    ///
+    /// Each string in "dimensions" is generated as a Dafny "int" (that is, a BigInteger).
+    ///
+    /// If "mustInitialize" is true, then fills each array element with a default value of type "elementType".
+    /// In this case, "exampleElement" must be null.
+    ///
+    /// If "mustInitialize" is false, then the array's elements are left uninitialized.
+    /// In this case, "exampleElement" may be non-null as a guide to figuring out what run-time type the array should have.
```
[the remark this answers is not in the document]
RustanLeino: Yes.

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -1500,74 +1501,34 @@ protected class ClassWriter : IClassWriter {
       return w;
     }
 
-    // TODO: Generalize the EmitArraySelectAsLvalue API to be rid of this duplication
```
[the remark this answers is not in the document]
RustanLeino: This improvement caused me some pain in merging, but it's all good now.

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
+// array-size limit has been exceeded.
+// It is expected that len(dims) is at least 1 and that each number in
+// dims is non-negative.
+func computeTotalArrayLength(dims ...Int) int {
+  product := dims[0]
+  for i := 1; i < len(dims); i++ {
+    product = product.Times(dims[i])
+  }
+  if IntOf(0x8000_0000).Cmp(product) <= 0 {
+    panic(fmt.Sprintf("array size exceeds memory limit: %v", String(product)))
```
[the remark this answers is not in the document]
RustanLeino: Correct. :(
Re 256: Good idea!

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -250,7 +250,7 @@ private struct Import {
       //   Type1 _dafny.TypeDescriptor
       // }
       //
-      // func (_this type_Class_) Default() interface{} {
+      // func (_this type_Class_) Default() any {
```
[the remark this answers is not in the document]
RustanLeino: I agree (with both you and the Go language designers) that `any` is much more readable than `interface{}`. However, for consistency, I'd rather keep it as `interface{}` even in this comment for now. So, I suggest we switch all of it over in the future (including in comments).

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
-  dims     []int
-}
-
-func newArray(dims ...Int) *Array {
+// Array is the general interface for arrays. Conceptually, it contains some
+// underlying storage (a slice) of elements, together with a record of the length
+// of each dimension. Thus, this general interface supports 1- and multi-dimensional
+// Dafny arrays.
+//
+// All array indices are given as the Go type "int". This is what the Dafny implementation
```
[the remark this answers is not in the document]
robin-aws: I'm only saying that the logic in the current Dafny-to-Go compiler selects `uint32` as the native type for the `newtype` declaration I gave above (and your PR doesn't affect that). It's not a problem for this PR since you're explicitly picking `int` for the array index type, and that does in fact line up with the type of the built-in `len()` function, for example.

I'm only asking about this because in MY PR, in order to replace some of the Go runtime with cross-compiled Dafny code, I will end up implicitly using `uint32` as the array index type, or more accurately as the sequence indexing type. I'm mentioning this in case you know of a reason that's a bad idea so I can adjust my approach. It's in no way an issue with THIS PR. Sorry for not being clearer. :)

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
+// If the produce doesn't fit in an "int", panic with the message that the
+// array-size limit has been exceeded.
+// It is expected that len(dims) is at least 1 and that each number in
+// dims is non-negative.
+func computeTotalArrayLength(dims ...Int) int {
+  product := dims[0]
+  for i := 1; i < len(dims); i++ {
+    product = product.Times(dims[i])
+  }
+  if IntOf(0x8000_0000).Cmp(product) <= 0 {
```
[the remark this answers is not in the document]
RustanLeino: Unless you push harder, I feel okay with what's there now. There's a comment above this function that says `dims` is expected to contain non-negative integers. The function is internal to the package and is called only from `NewArrayFromExample`.

I added a comment to `NewArrayFromExample` that states the assumptions and states what is checked. This function is public to the package. We could potentially add run-time checks in it (but note that we're still not checking things like: if both `example` and `init` are non-`nil`, then they are values of the same type).

Source/DafnyRuntime/DafnyRuntime.go
RustanLeino: Your proposed comparison (actually, it would be `&_this == otherArray` to have the right type) would compare the addresses of the two `ArrayStruct`s. When a Dafny array is copied, what happens in the target Go is that such an `ArrayStruct` is copied. Thus, the address of the two `ArrayStruct`s may be different and yet they might refer to the same array.

This important test case was missing from the test suite. I have added it, along with a comment next to L911.

Source/DafnyRuntime/DafnyRuntime.go
RustanLeino: I don't understand your comment. Are you proposing that `ArrayGet1` check to make sure the array has just one dimension and panic if it's not?

Source/DafnyRuntime/DafnyRuntime.go
RustanLeino: This is a good point. It's not currently a problem, because for now in Go we just need to figure out the size of the array elements. Since trait `T` and class `C1` have the same size, we're good.

As you mention, we'll have to be careful about this in the future if we allow a byte/char/codepoint to implement a trait. My guess of how it will play out is: If we're creating an `array<T>` where `T` is a trait that is implemented by, say, a Dafny `byte` type, then the example element will be a fat pointer to a `T`. From this, we'll be able to tell, at run time, that the array-element size should be that of a pointer, not that of a byte. In other words, as long as the example element is really passed in using the representation we use for a trait, then the scheme in this PR will continue to work.

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -692,183 +692,412 @@ func (seq Seq) VerbatimString(asLiteral bool) string {
...
+      for i := range arr {
+        arr[i] = x
+      }
+    }
+    return &ArrayForByte{
+      contents: arr,
+      dims:     intDims,
+    }
+  }
+  if _, ok := example.(Char); ok {
```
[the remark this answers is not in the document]
RustanLeino: It seems that `DafnyRuntime.go` defines both `Char` and `CodePoint` as type aliases for `rune`:

``` go
type Char rune
type CodePoint rune
```

Doesn't this mean that the specialization for `Char` is really a specialization for `rune`, and therefore the array optimizations already work for both `Char` and `CodePoint`?

Source/DafnyRuntime/DafnyRuntime.go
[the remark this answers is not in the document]
robin-aws: That's just fine. I would really like to harden the API for our runtimes, but I'd also even more love to provide enough functionality so that end users never have to write code that references the runtimes directly at all, so I wouldn't spend effort on it now.

Source/DafnyRuntime/DafnyRuntime.go
robin-aws: That's fair, one possibility is to provide not just a `0x0` byte value but an augmented `(0x0, MyTrait)` value of some kind. Or perhaps even just a type descriptor value if that's simpler.

Source/DafnyRuntime/DafnyRuntime.go
robin-aws: I'm saying that if `ArrayStruct.ArrayGet1Byte` panics because it's not expected (even though it could conditional work by downcasting), I would expect that `ArrayForByte.ArrayGet1` should also panic since it's also a mismatch from expectations.

I then wondered if it's possible we need this because of type variables, because we might be holding on to an optimized byte array without knowing it. But then AFAICT `array<byte>` is not a subtype of `array<T>` in Dafny unless `T` is in fact `byte`, so I think a panic here would be safe and catch bugs.

Source/DafnyRuntime/DafnyRuntime.go
robin-aws: I'm still a bit confused. I understand you're saying that Go will copy the struct as it is passed around, and we want different structs that refer to the same underlying storage to compare equal. But why compare the address of the first element of the `dims` array specifically?

Source/DafnyRuntime/DafnyRuntime.go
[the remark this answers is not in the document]
robin-aws: Nope, a [type declaration](https://go.dev/ref/spec#Type_definitions) like those in Go creates a new distinct type, so it's like `newtype` in Dafny. It means on this line you're specifically checking for `Char`, and the type test will not succeed if given a `CodePoint` or even a plain `rune`. I confirmed this with a quick test:

```go
func TestCodePointCheck(t *testing.T) {
	var cp interface{} = CodePoint('D')
	_, ok := cp.(Char)
	if (!ok) {
		t.Error("Oh shoot: ", cp)
	}
}
```

```
=== RUN   TestCodePointCheck
    DafnyRuntime_test.go:16: Oh shoot:  'D'
--- FAIL: TestCodePointCheck (0.00s)
FAIL
exit status 1
```

Note that the value prints as `'D'` rather than `D` precisely because it's a different type that has a `String()` method that adds the single quotes, which is the whole reason I introduced `CodePoint`: so it's a distinct type that behaves differently.

Source/DafnyRuntime/DafnyRuntime.go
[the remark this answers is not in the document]
RustanLeino: It seems alright to use something slightly different for the sequences. If the sequence implementation supports lazy concatenation, then it may actually support sequence lengths that exceed the maximum array length.

Source/DafnyRuntime/DafnyRuntime.go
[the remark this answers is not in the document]
RustanLeino: Thanks. I have fixed it.

Source/DafnyRuntime/DafnyRuntime.go
RustanLeino: I thought some more about this. From a correctness perspective, we should be fine. However, we may in the future want to optimize references to instances of _classes_. When such a reference is converted to an `interface{}`, I believe it goes from one machine word of storage to two. This is based on my knowledge of the _abstract_ representation of a interface reference in Go, which is a fat pointer. I don't know if the running Go actually stores such fat pointers as one or two machine words. Anyhow, if they are different and if we want to detect this at run time to (in some way I don't actually understand how) optimize the fat pointer to a single pointer, then the issue you're bringing up with traits is something we'll need to think about.

Source/DafnyRuntime/DafnyRuntime.go
RustanLeino: I argued above that `&_this == otherArray` does not work. But then, what about

    _this.contents == otherArray.contents

or

    _this.dims == otherArray.dims

This doesn't work either, because the type of `.contents` and `.dims` is a _slice_ and the only thing Go allows you to compare a slice with is `nil`. One might imagine obtaining the Go arrays underlying these slices and comparing them. However, that would require unsafe Go code.

Taking the address of an array/slice element is always okay in Go. The specific element we're taking the address does not matter, as long as it's the same for the LHS and RHS of the comparison and as long as the element exists. Using the address of `.contents[0]` is _almost_ the obvious choice, except that the contents may be empty. However, the number of dimensions is always nonempty (so `.dims[0]` exists) and the implementation currently always uses a different `.dims` for each `Array`. (If that were to change in the future, to someone share these never-mutated `.dims` arrays, then the equality comparison we use will need to change.)

Source/DafnyRuntime/DafnyRuntime.go
RustanLeino: The reason is indeed because of type variables. Here's an example where we might be holding on to an optimized byte array without knowing it:

``` dafny
newtype byte = x | 0 <= x < 256

method Main() {
  var byteArray := new byte[100];
  var intArray := new int[100];

  var _ := byteArray[5]; // this calls ArrayGet1Byte on the optimize byte array
  byteArray[5] := 64; // similarly, this calls ArraySet1Byte
  var _ := intArray[5]; // this calls ArrayGet1
  intArray[5] := 64; // similarly, this calls ArraySet1

  var _ := UpdateArrayElement(byteArray, 5, 65);
  var _ := UpdateArrayElement(intArray, 5, 65);
}

method UpdateArrayElement<X>(arr: array<X>, j: nat, x: X) returns (previous: X)
  requires j < arr.Length
  modifies arr
{
  previous := arr[j]; // this always calls ArrayGet1, even on an optimized array
  arr[j] := x; // similarly, this always calls ArraySet1
}
```

Source/DafnyRuntime/DafnyRuntime.go
RustanLeino: Done.

Source/DafnyRuntime/DafnyRuntime.go
RustanLeino: I changed `ArrayStruct`, `ArrayForByte`, `ArrayForChar`, and `ArrayForCodePoint` to be private to the package in order to contain these design decisions in the Dafny package.

robin-aws:
Just to update: I've been working on fixing the remaining issues with the new `Test/unicodechars/comp/Arrays.dfy` test case - I have them mostly fixed on a branch but there's still some non-trivial work to coerce Java functions correctly in the same style Go already does.

robin-aws:
The Java fixes are rather involved so I'm going to move them to a separate PR, and just merge this one with Java disabled in the new `Test/unicodechars/comp/Arrays.dfy` test case for now.

@fabiomadge would you mind giving my two commits (https://github.com/dafny-lang/dafny/pull/2818/commits/6835539a622389defe636e9430295d960eb07ce3 and https://github.com/dafny-lang/dafny/pull/2818/commits/bcd2f04bf556d4bf997186bea6ef27dae8af95e9) a sanity check before I merge this? Rustan asked me to add whatever was necessary to complete this while he was on vacation, but I don't want to use that to sneak code in without review. :)

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -138,7 +137,11 @@ def __dafnystr__(self) -> str:
         return '[' + ', '.join(map(string_of, self)) + ']'
 
     def __add__(self, other):
-        return Seq(super().__add__(other), isStr=self.isStr and other.isStr)
+        if self.isStr is None or other.isStr is None:
+            isStr = None
+        else:
+            isStr = self.isStr and other.isStr
+        return Seq(super().__add__(other), isStr=isStr)
```
fabiomadge, line 144: Are you sure that this changes the semantics?

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -399,6 +402,12 @@ def plus_char(a, b):
 def minus_char(a, b):
     return chr(ord(a) - ord(b))
 
+def plus_unicode_char(a, b):
+    return CodePoint(plus_char(a, b))
```
fabiomadge, line 406: I was sure to have made this remark elsewhere, but can't find it, so: Can you overload `+` and `-` instead?

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1665,7 +1661,7 @@ private class ClassWriter : IClassWriter {
         SeqType or MultiSetType => ("[", "]"),
         _ => ("{", "}")
       };
-      wr.Write(TypeHelperName(ct));
+      wr.Write(ct is SeqType ? DafnySeqMakerFunction : TypeHelperName(ct));
```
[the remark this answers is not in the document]
robin-aws: I did at first and found the code generation was a lot messier.

Source/DafnyRuntime/DafnyRuntime.py
robin-aws: Ah true, I can do that. For the record I don't care much about the usability of `CodePoint` though, since I only expect compiled Dafny code to reference it.

Source/DafnyRuntime/DafnyRuntime.py
robin-aws: Yup. Before `Seq("a", isStr=None) + Seq("b", isStr=None)` would become the same as `Seq("ab", isStr=False)`, which would still go through the inference and decide it was a string, which we don't want.

Source/DafnyRuntime/DafnyRuntime.py
robin-aws: It did end up simplifying the code generation logic a little though, which I DO care about :) , so thanks!

Source/DafnyRuntime/DafnyRuntime.py
fabiomadge:
```
>>> (None and None) == None
True
>>> (None and None) == False
False
```

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -58,6 +58,8 @@ public class PythonCompiler : SinglePassCompiler {
     const string DafnySetClass = $"{DafnyRuntimeModule}.Set";
     const string DafnyMultiSetClass = $"{DafnyRuntimeModule}.MultiSet";
     const string DafnySeqClass = $"{DafnyRuntimeModule}.Seq";
+    private string DafnySeqMakerFunction => UnicodeCharEnabled ? $"{DafnyRuntimeModule}.SeqWithoutIsStrInference" : DafnySeqClass;
+
```
fabiomadge, line 62:
```suggestion
```

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -461,6 +470,7 @@ def __iter__(self):
 class defaults:
     bool = staticmethod(lambda: False)
     char = staticmethod(lambda: 'D')
+    codepoint = staticmethod(lambda: CodePoint('D'))
```
fabiomadge, line 473:
```suggestion
    codepoint = staticmethod(lambda: CodePoint(char))
```
I _think_ that works.

Source/DafnyRuntime/DafnyRuntime.py
robin-aws: Ah you're right, `self.isStr = None` is all that needed to change. I added this at the same time and never tested it without. :) Thanks!

Source/DafnyRuntime/DafnyRuntime.py
robin-aws: Not quite, but this does:

```suggestion
    codepoint = staticmethod(lambda: CodePoint(defaults.char()))
```

fabiomadge:
The Python changes look ok to me.

robin-aws (approved):

PR #2820, opened 2022-09-30
feat: Support for arbitrary versions of z3

MikaelMayer:
Added documentation on how to make Dafny work for other versions of Z3 Added automatic overriding of prover options if Z3 version 4.8.7 or newer is detected in the language server

I tested it manually with Z3 version 4.8.7. I verified the version of Z3 was correctly sent to VSCode using the telemetry handler, and that counter-examples were working as usual (which they weren't before if the option wasn't changed)

### DIFF
--- a/Source/DafnyCore/DafnyOptions.cs
+++ b/Source/DafnyCore/DafnyOptions.cs
@@ -1332,9 +1332,9 @@ Exit code: 0 -- success; 1 -- invalid command-line; 2 -- parse or type errors;
 
 /extractCounterexample
     If verification fails, report a detailed counterexample for the
-    first failing assertion. Requires specifying the /mv option as well
-    as /proverOpt:O:model_compress=false and
-    /proverOpt:O:model.completion=true.
+    first failing assertion. Requires specifying the /mv:<file> option as well
+    as /proverOpt:O:model_compress=false (or /proverOpt:O:model.compact=false if
+    your version of z3 is >= 4.8.7) and /proverOpt:O:model.completion=true.
 
 /countVerificationErrors:<n>
     (deprecated)
--- a/Source/DafnyLanguageServer/DafnyLanguageServer.cs
+++ b/Source/DafnyLanguageServer/DafnyLanguageServer.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Text.RegularExpressions;
 using Microsoft.Dafny.LanguageServer.Handlers;
 using Microsoft.Dafny.LanguageServer.Language;
 using Microsoft.Dafny.LanguageServer.Workspace;
@@ -14,7 +15,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Boogie.SMTLib;
 using Microsoft.Extensions.Options;
-using OmniSharp.Extensions.LanguageServer.Protocol.Window;
+using Action = System.Action;
 
 namespace Microsoft.Dafny.LanguageServer {
   public static class DafnyLanguageServer {
@@ -55,6 +56,8 @@ namespace Microsoft.Dafny.LanguageServer {
       return Task.CompletedTask;
     }
 
+    private static readonly Regex Z3VersionRegex = new Regex(@"Z3 version (?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)");
+
     private static void PublishSolverPath(ILanguageServer server) {
       var telemetryPublisher = server.GetRequiredService<ITelemetryPublisher>();
       string solverPath;
@@ -62,11 +65,50 @@ namespace Microsoft.Dafny.LanguageServer {
         var proverOptions = new SMTLibSolverOptions(DafnyOptions.O);
         proverOptions.Parse(DafnyOptions.O.ProverOptions);
         solverPath = proverOptions.ExecutablePath();
+        HandleZ3Version(telemetryPublisher, proverOptions);
       } catch (Exception e) {
         solverPath = $"Error while determining solver path: {e}";
       }
 
       telemetryPublisher.PublishSolverPath(solverPath);
+    }
+
+    private static void HandleZ3Version(ITelemetryPublisher telemetryPublisher, SMTLibSolverOptions proverOptions) {
+      var z3Process = new ProcessStartInfo(proverOptions.ProverPath, "-version") {
+        CreateNoWindow = true,
+        RedirectStandardError = true,
+        RedirectStandardOutput = true,
+        RedirectStandardInput = true
+      };
+      var run = Process.Start(z3Process);
+      if (run == null) {
+        return;
+      }
+
+      var actualOutput = run.StandardOutput.ReadToEnd();
+      run.WaitForExit();
+      var versionMatch = Z3VersionRegex.Match(actualOutput);
+      if (!versionMatch.Success) {
+        // Might be another solver.
+        return;
+      }
+
+      telemetryPublisher.PublishZ3Version(versionMatch.Value);
+      var major = int.Parse(versionMatch.Groups["major"].Value);
+      var minor = int.Parse(versionMatch.Groups["minor"].Value);
+      var patch = int.Parse(versionMatch.Groups["patch"].Value);
+      if (major <= 4 && (major != 4 || minor <= 8) && (minor != 8 || patch <= 6)) {
+        return;
+      }
+
+      var toReplace = "O:model_compress=false";
+      var i = DafnyOptions.O.ProverOptions.IndexOf(toReplace);
+      if (i == -1) {
+        telemetryPublisher.PublishUnhandledException(new Exception($"Z3 version is > 4.8.6 but I did not find {toReplace} in the prover options:" + string.Join(" ", DafnyOptions.O.ProverOptions)));
+        return;
+      }
+
+      DafnyOptions.O.ProverOptions[i] = "O:model.compact=false";
     }
 
     /// <summary>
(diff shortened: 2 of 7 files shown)


Source/DafnyCore/DafnyOptions.cs
```diff
@@ -1332,9 +1332,9 @@ the program.
 
 /extractCounterexample
     If verification fails, report a detailed counterexample for the
-    first failing assertion. Requires specifying the /mv option as well
-    as /proverOpt:O:model_compress=false and
-    /proverOpt:O:model.completion=true.
+    first failing assertion. Requires specifying the /mv:<file> option as well
+    as /proverOpt:O:model_compress=false (or /proverOpt:O:model.compact=false if
+    your version of z3 is >= 4.8.7) and /proverOpt:O:model.completion=true.
```
atomb, line 1337: We'll probably want to swap what's in parentheses once we adopt a newer version by default.

atomb: Or maybe update this text to match what's in the message in `DafnyModel.cs`, which is more symmetric?

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -55,20 +56,61 @@ public static class DafnyLanguageServer {
...
+      if (!versionMatch.Success) {
+        // Might be another solver.
+        return;
+      }
+
+      telemetryPublisher.PublishZ3Version(versionMatch.Value);
+      var major = int.Parse(versionMatch.Groups["major"].Value);
+      var minor = int.Parse(versionMatch.Groups["minor"].Value);
+      var patch = int.Parse(versionMatch.Groups["patch"].Value);
+      if (major <= 4 && (major != 4 || minor <= 8) && (minor != 8 || patch <= 6)) {
```
atomb, line 100: I would find it easier to read if written like this, though it's a very minor issue.

```suggestion
      if (major <= 4 && (major < 4 || minor <= 8) && (minor < 8 || patch <= 6)) {
```

Source/DafnyLanguageServer/DafnyLanguageServer.cs
```diff
@@ -55,20 +56,61 @@ public static class DafnyLanguageServer {
...
+    private static readonly Regex Z3VersionRegex = new Regex(@"Z3 version (?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)");
+
     private static void PublishSolverPath(ILanguageServer server) {
       var telemetryPublisher = server.GetRequiredService<ITelemetryPublisher>();
       string solverPath;
       try {
         var proverOptions = new SMTLibSolverOptions(DafnyOptions.O);
         proverOptions.Parse(DafnyOptions.O.ProverOptions);
         solverPath = proverOptions.ExecutablePath();
+        HandleZ3Version(telemetryPublisher, proverOptions);
```
atomb, line 68: In the very long run, I think we may want to handle solvers other than Z3, and perhaps condition on their versions, too. But that's probably somewhat far in the future. :)

docs/DafnyRef/README.md
```diff
@@ -24,11 +24,11 @@ other aspects that are dissimilar between the pdf and online versions.
```
atomb, line 24: Did you mean to include changes to this file as part of this PR?

atomb (changes requested):
Thanks for doing this! I have a few very small changes to request, but it overall looks great.

It'd be hard to test this without installing multiple versions of Z3 in CI, which is probably a bigger step than we want to take quite yet. We probably will do that sometime before long, though, and when we do it might be worth coming back and adding a test for this feature.

Source/DafnyLanguageServer/DafnyLanguageServer.cs
MikaelMayer: Agreed, and then we can rename this method. For now, it specifically do nothing if there is no "z3 version" in the output of the process.

RELEASE_NOTES.md
```diff
@@ -4,6 +4,7 @@
 - feat: Support for parsing Basic Multilingual Plane characters from UTF-8 in code and comments (https://github.com/dafny-lang/dafny/pull/2717)
 - feat: Command-line arguments are now available from `Main` in Dafny programs, using `Main(args: seq<string>)` (https://github.com/dafny-lang/dafny/pull/2594)
 - feat: Generate warning when 'old' has no effect (https://github.com/dafny-lang/dafny/pull/2610)
+- feat: Support arbitrary version of z3 (https://github.com/dafny-lang/dafny/pull/2820)
```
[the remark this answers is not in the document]
MikaelMayer: I think Dafny already supports arbitrary version of z3 with this parameter. I was surprised but there was nothing to do there. It's just that model_compress cannot be passed for versions of z3 greater than 4.8.6
What about "Support arbitrary versions of z3 for the language server" ?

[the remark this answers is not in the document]
atomb: Yeah, I like "Support arbitrary versions of z3 for the language server", but perhaps also with the second sentence included? We could even say something about how we plan to make a newer version the default before too long.

docs/DafnyRef/Options.txt
```diff
@@ -343,9 +343,9 @@ Usage: Dafny [ option ... ] [ filename ... ]
   
   /extractCounterexample
       If verification fails, report a detailed counterexample for the
-      first failing assertion. Requires specifying the /mv option as well
-      as /proverOpt:O:model_compress=false and
-      /proverOpt:O:model.completion=true.
+      first failing assertion. Requires specifying the /mv:<file> option as well
+      as /proverOpt:O:model_compress=false (or /proverOpt:O:model.compact=false if
+      your version of z3 is >= 4.8.7) and /proverOpt:O:model.completion=true.
```
MikaelMayer, line 348:
```suggestion
      first failing assertion. Requires specifying the /mv:<file> option as well
      as /proverOpt:O:model_compress=false (for z3 version < 4.8.7)" or /proverOpt:O:model.compact=false (for z3 version >= 4.8.7), and /proverOpt:O:model.completion=true.
```

Source/DafnyCore/DafnyOptions.cs
MikaelMayer, line 1337:
```suggestion
    first failing assertion. Requires specifying the /mv:<file> option as well
    as /proverOpt:O:model_compress=false (for z3 version < 4.8.7) or /proverOpt:O:model.compact=false (for z3 version >= 4.8.7), and /proverOpt:O:model.completion=true.
```

docs/DafnyRef/Options.txt
```diff
@@ -343,9 +343,8 @@ Usage: Dafny [ option ... ] [ filename ... ]
   
   /extractCounterexample
       If verification fails, report a detailed counterexample for the
-      first failing assertion. Requires specifying the /mv option as well
-      as /proverOpt:O:model_compress=false and
-      /proverOpt:O:model.completion=true.
+      first failing assertion. Requires specifying the /mv:<file> option as well
+      as /proverOpt:O:model_compress=false (for z3 version < 4.8.7)" or /proverOpt:O:model.compact=false (for z3 version >= 4.8.7), and /proverOpt:O:model.completion=true.
```
MikaelMayer, line 347:
```suggestion
      as /proverOpt:O:model_compress=false (for z3 version < 4.8.7) or /proverOpt:O:model.compact=false (for z3 version >= 4.8.7), and /proverOpt:O:model.completion=true.
```

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -1332,9 +1332,8 @@ the program.
 
 /extractCounterexample
     If verification fails, report a detailed counterexample for the
-    first failing assertion. Requires specifying the /mv option as well
-    as /proverOpt:O:model_compress=false and
-    /proverOpt:O:model.completion=true.
+    first failing assertion. Requires specifying the /mv:<file> option as well
+    as /proverOpt:O:model_compress=false (for z3 version < 4.8.7) or /proverOpt:O:model.compact=false (for z3 version >= 4.8.7), and /proverOpt:O:model.completion=true.
```
atomb, line 1336:
```suggestion
    as /proverOpt:O:model_compress=false (for z3 version < 4.8.7) or
    /proverOpt:O:model.compact=false (for z3 version >= 4.8.7), and
    /proverOpt:O:model.completion=true.
```

atomb: I recently reformatted all of the option descriptions so they wouldn't wrap on a standard 80-column terminal, and I'd love it if we could keep them that way.

atomb (changes requested):
One tiny additional change and then I think it's good to go!

atomb (approved):
Looks great!

keyboardDrummer:
@atomb Are we planning to figure out how to configure the latest Z3 version so that it performs as well for Dafny as 4.8.5, or do we think there's no such configuration?

atomb:
> @atomb Are we planning to figure out how to configure the latest Z3 version so that it performs as well for Dafny as 4.8.5, or do we think there's no such configuration?

We don't yet know of a configuration that performs as well (and we have tried several possibilities), but this would give us a way to enable a better configuration if we discover one.

atomb (approved):

cpitclaudel:
Looks like the RELEASE_NOTES entry ended up in the wrong place.  I will fix that in #2848

hmijail:
Regarding performance of Z3, just a comment to ensure this is in the radar: version 4.8.16 was the first one available natively to MacOS on ARM, before that it's only available for Intel so it runs emulated in those computers. I'd imagine this worsens performance but haven't compared.

PR #2823, opened 2022-09-30
Enable passing program arguments when using the new CLI format

keyboardDrummer:
### Changes

Enables passing program arguments when using the `run` command of the new CLI format. There is no `--args` option or handling of `--` different from POSIX. Instead, any arguments after the first one, which is interpreted as a source file, are interpreted as arguments to the program.

This is similar to what npm and dotnet do.

We don't expect many use-cases for `run` where there is a need to pass multiple input files, but when there is, the new option `--input <file>` can be used. This option can be used multiple times to add multiple additional input files.

In the future we also expect to have a project file for Dafny that'll enable specifying all the project's source files, and then the project file will be the single passed input file, which is similar to how `dotnet` works.

### Testing

- Update an existing test to also test this feature


davidcok:
So what distinguishes arguments to the user program from options and files for dafny itself?

keyboardDrummer:
> So what distinguishes arguments to the user program from options and files for dafny itself?

The first argument is the one and only file passed to the Dafny CLI. Other Dafny options are options so they are recognised by the `-` and `--` prefix. Any arguments after the first will be arguments to the program.

atomb:
I did recently have a case where I wanted to use `/runAllTests` along with a target-language source file, which as you say this would disallow. I'd somewhat lean toward having a way to support this, such as the extra `--input` option you suggest. But I totally agree that a project file is the way to go in the long term.

davidcok:
But (a) the program being run might consist of multiple Dafny files and (b) the user program might have - or -- options itself, not to mention having .dfy files as input.

keyboardDrummer:
> I did recently have a case where I wanted to use `/runAllTests` along with a target-language source file, which as you say this would disallow. I'd somewhat lean toward having a way to support this, such as the extra `--input` option you suggest. But I totally agree that a project file is the way to go in the long term.

I see. Okay I will add it.

Test/comp/CompileWithArguments.dfy
```diff
@@ -1,11 +1,11 @@
 // RUN: %dafny /compile:0 "%s" > "%t"
-// RUN: %dafny /noVerify /compile:4 /compileTarget:cs "%s" --args csharp 1 >> "%t"
-// RUN: %dafny /noVerify /compile:4 /compileTarget:cpp "%s" --args cpp Yipee >> "%t"
-// RUN: %dafny /noVerify /compile:4 /compileTarget:java "%s" --args java heya >> "%t"
-// RUN: %dafny /noVerify /compile:4 /compileTarget:js "%s" --args javascript 2 >> "%t"
+// RUN: %baredafny run --no-verify --target:cs %args "%s" csharp 1 >> "%t"
```
MikaelMayer, line 2: What is this `%args` for ?

Test/comp/CompileWithArguments.dfy
```diff
@@ -1,11 +1,11 @@
...
+// RUN: %baredafny run --no-verify --target:cpp %args "%s" cpp Yipee >> "%t"
+// RUN: %baredafny run --no-verify --target:java %args "%s" -- java --heya >> "%t"
```
MikaelMayer, line 4: Good job on testing this.

davidcok:
So one needs to repeat the target? as in --target=java java 3 4 5
That still seems clunky to me. As is restricting the run command to just one .dfy file.
What is the objection to --args ? --- that seems clear, unambiguous, readable

Test/comp/CompileWithArguments.dfy
keyboardDrummer: It's because `run` needs to be the first argument, so we can't use %dafny because that places the default arguments for tests _before_ the verb. Alternatively, we could replace `%baredafny run %args` with `%run`

keyboardDrummer:
> So one needs to repeat the target? as in --target=java java 3 4 5 

No, one `--target <target>` is enough. The second `java` in this test has the name of a target, but it is an input to the Dafny program.

> As is restricting the run command to just one .dfy file.

Can you describe the use-case that leads you to want to pass multiple .dfy files to the run command?

My impression is that these are the two relevant use-cases:
- You have multiple dfy files in your application, one has a main method and it includes the other files, so you only reference that root file on the CLI and the includes find the other ones. If you don't want to use includes you will have to use `--input` to specify additional input files, or wait for Dafny project files.
- You are using a Dafny library that is not referenced through includes in your Dafny files. By using `--library <libraryFile>`, the library will be available for resolution and compilation, but won't be verified. If you use `--input <libraryFile>` instead then it'll be verified as well.


That still seems clunky to me.  What is the objection to --args ? --- that seems clear, unambiguous, readable

> What is the objection to --args ?

The way `--args` works is that it preprocessed the CLI arguments, using `--args` to split the arguments into a before and after list. This means that 

```dafny run sourceFile.dfy --target cs --args 3``` 

has a meaning different from

```dafny run sourceFile.dfy --args 3 --target cs```

Which I find confusing because it makes `--args` behave completely different from other options. This isn't something that similar CLIs do and I'm guessing this is also not POSIX compliant.

**Comment by robin-aws** [comment]
This looks great to me. My one question is just about documentation to make this easier: do you know of a way to populate args like `--input a.dfy --input b.dfy ...` in a shell command of some kind? It's currently common for projects to use something like `dafny $(shell find $(SOURCEDIR) -name '*.dfy') ...` in a Makefile to feed all relevant Dafny files into the CLI, and I'm not enough of a shell wizard to know how to insert `--input` before each file path (and I hit this when trying to do the same thing with the `/library` option). I'm in favour of this change but would like to document how to migrate such scripting.

**Comment by davidcok** [comment]
To respond to some of Remy's questions:

In my experience most file processing tools (on Linux/Mac systems) have CLIs that permit a mix of options and files -- ordinarily one can list as many input files as are appropriate for the command. Indeed, old-style Dafny was just that: list all the files you want -- those are the ones that will be verified.

So that is the use that is being modified significantly here. Now one has to put --input in front of all but one of them. To me, that is ugly and verbose. Making one argument by quoting or comma-separation would be better, but leaving the previous behavior is better in my opinion.

Putting includes in files makes it more difficult to select which files are verified in a given dafny run and makes it more difficult to select which version of some Dafny file (with equivalent set of modules) is being tested in a given dafny run.

Of course having non-option arguments be input files to Dafny is in tension with having non-option arguments be arguments for the user program; and the latter arguments might be options or dafny files themselves. To me I prefer having these options (only used for the run command after all) be the ones disadvantaged: identify them with --args and double quote the space separated list as the one argument to --args.

So
- I don't like changing old-Dafny behavior this much because 'dafny run' is the only situation where user arguments are needed.
- The desire to not use --args (or something equivalent) is significantly changing what dafny users are used to for a less commonly used use case.

I prefer keeping
`dafny verify A.dfy B.dfy C.dfy`
`dafny translate A.dfy B.dfy C.dfy`
`dafny build A.dfy B.dfy C.dfy`
even if that means saying
`dafny run A.dfy B.dfy C.dfy --args "1 2 3 4"`
and that all is preferable to 
 `dafny run A.dfy --input B.dfy --input C.dfy 1 2 3 4`
and certainly to having to use --input for the other subcommands

**Comment by cpitclaudel** [comment]
Just a few notes:

@davidcok 
> `dafny run A.dfy B.dfy C.dfy --args "1 2 3 4"`

This would be
> `dafny run A.dfy B.dfy C.dfy --args 1 2 3 4` 

with (no quotes), right?

@robin-aws 
> My one question is just about documentation to make this easier: do you know of a way to populate args like `--input a.dfy --input b.dfy ...` in a shell command of some kind? It's currently common for projects to use something like `dafny $(shell find $(SOURCEDIR) -name '*.dfy') ...` in a Makefile

In a shell script `find . -name '*.dfy' -printf '--input %p\n'` works, though better use `\0` and `xargs -0` to avoid issues with file names that contain blanks.

@keyboardDrummer 
> [regarding --args] This isn't something that similar CLIs do and I'm guessing this is also not POSIX compliant.

I've cited GNU parallel's `:::` before, but I could swear that I had seen `--args` somewhere else before, and I finally found it: `gdb` uses `--args`:

```
--args prog [arglist]

    Change interpretation of command line so that arguments following this option are passed as arguments to the inferior. As an example, take the following command:

    gdb ./a.out -q

    It would start GDB with -q, not printing the introductory message. On the other hand, using:

    gdb --args ./a.out -q

    starts GDB with the introductory message, and passes the option to the inferior.
```

**Review by cpitclaudel — APPROVED** [review]
Code looks good to me.  I don't have a strong opinion on `--args` vs `--` with the one-input-file constraint (I've explained in a previous thread why I wasn't a fan of just renaming `--args` to `--` while still allowing multiple Dafny input files).

**Comment by keyboardDrummer** [comment]
@davidcok I like the way you envision `--args`: as a taking a single argument, which can be in quotes to specify multiple arguments to the Dafny program. Also indeed for any commands other than `run`, we can leave the option to specify multiple input files as arguments without a downside.

> The desire to not use --args (or something equivalent) is significantly changing what dafny users are used to for a less commonly used use case.

I think it's a good callout that we should optimise for the most common use-case.

Let me try to write down the use-cases you mentioned and break them down a little more:
1. Pass arguments to the Dafny program
1. Swap out one module for another with a similar contract without having to adjust includes
1. Use-cases related to customising which files are verified:
    1. Only reverify files that you just changed
    1. Only verify files that are part of this repository, so not any library files that are somehow important.

(2) seems like a rare use-case to me, but you will be able to that in the future by using two different project files, which other project files specifies which input files are included in the program.
(3.1) is a use-case we can make better than it is now by introducing caching of verification, so when you make changes to your program and rerun it, only the invalidated verification is redone.
(3.2) is something that can be done by using either includes, the `--library` flag, or the `library` field in upcoming Dafny project files.

So, are you sure that in the future, passing arguments to the Dafny program will be a less common use-case than the others we identified?

**Comment by davidcok** [comment]
I was thinking of compiler tools in general: they allow living all sorts of input files without special options.

But the relevant ones to this discussion are, as you list, tools that have input files and options and then launch an executable with its own input files and options.

So have we come to:
dafny verify/translate/build -- list as many input files as desired (without --input)

dafny run is one of

dafny run A.dfy B.dfy C.dfy --args 2 3 4  (split command line)

dafny run A.dfy B.dfy C.dfy --args "2 3 4" (no split command line, not concerns about ordering, except that the first .dfy while is the "Main" one)

dafny run A.dfy --input B.dfy --input C.dfy 2 3 4 (bias in favor of launched program)

dafny run A.dfy --input B.dfy --input C.dfy --args 2 3 4 (bias in favor of launched program)

I prefer them in that order, and would be happy with either of the first two (which match the other subcommands and past practice), but really dislike the later two.

File: Test/comp/CompileWithArguments.dfy (hunk 1 above)
Comment (MikaelMayer) [reply Test/comp/CompileWithArguments.dfy:2]: Oh, I got it. It has nothing to do with `--args`, it's just the other options that previously were inserted in %dafny.
Would it make sense to name this option `%defaultArgs` or something similar then? At first, I thought it would be the new separators for the arguments. Or you could just move `%args` after `run`

**Comment by keyboardDrummer** [comment]
@davidcok, thanks for the comments. You’ve been clear in how you would like the CLI to work, but I want to make sure that’s not optimising for how we have worked with the Dafny CLI in the past but also for how we will work with it in the future.

Can you explain when you believe passing multiple input files will be useful in Dafny’s future?

I would find any system where you have to write down multiple input files when doing a CLI call to be cumbersome, so I expect us to remove the need for that. More specifically, I expect future Dafny users to either:
- Define a project file where you can configure what all your input sources are, after which you only have to reference the project file in the CLI call
- If they're not using a project file, have a root .dfy file that transitively imports all modules in the program, so only specifying that file on the CLI is enough to reference all others. I also expect Dafny's module system to change so module imports can directly resolve to a file (like in Python), making include directives obsolete this making it easy to define such a root file.

**Comment by MikaelMayer** [comment]
> @davidcok, thanks for the comments. You’ve been clear in how you would like the CLI to work, but I want to make sure that’s not optimising for how we have worked with the Dafny CLI in the past but also for how we will work with it in the future.
> 
> Can you explain when you believe passing multiple input files will be useful in Dafny’s future?
> 
> I would find any system where you have to write down multiple input files when doing a CLI call to be cumbersome, so I expect us to remove the need for that. More specifically, I expect future Dafny users to either:
> 
> * Define a project file where you can configure what all your input sources are, after which you only have to reference the project file in the CLI call
> * If they're not using a project file, have a root `.dfy` file that transitively imports all modules in the program, so only specifying that file on the CLI is enough to reference all others. For any executable Dafny application this will be the file containing the main method. For a library this might require an additional file, although I would recommend users to use a project file for libraries instead. I also expect Dafny's module system to change so module imports can directly resolve to a file (like in Python), making include directives obsolete so it'll be easier to define a root file.

I have to admit that I was puzzled by @keyboardDrummer  initial approach to limit the number of Dafny files to 1 and do this separations, but now, being used to the simplicity of node and python, I think it's the right thing to do. I think --input will not be used that much even if it is available, but it could serve the rare purpose of swapping a library to another one.

Here are my follow-up feedback:
* Even if there are not project files, one can always pre-build the command lines and run them using shell scripts. So the only thing we want to ensure is that things are _possible_.
* --args or -- was meant only to be used with the "run" command, not the "translate" or "build". So, when using "translate" or "build", if we encounter non-options after the first dafny file, we could still keep the old behavior, right? But that would still be cumbersome because to run, one would have to insert --input again.
* I envision that most people are going to develop in an IDE and verify either there or on the command-line,
but they will most often run the Dafny application using the `--noVerify` flag to skip verification. `dafny run` will mostly be for development, not production of course. However, I don't suggest we add this flag by default. But it's something that leave me thinking.
* --args "1 2 3 4" is the worst in my opinion.
* I like the idea of the --library flag. Down the road, we could put a dll file in the argument there and its embedded source could be used for verification.

**Comment by cpitclaudel** [comment]
>  --args: as a taking a single argument, which can be in quotes to specify multiple arguments to the Dafny program. @cpitclaudel, do you see a downside to making it work like that compared to the "split arguments" approach? 

The problem with this approach is that argument quoting in the shell works differently.  This leads to complicated and inconsistent quoting rules (I think we already have this problem with plugin arguments?)

**Comment by davidcok** [comment]
I agree that everything we want to do is possible under the various designs.

But here is how I try to think about design from a user perspective:
- How much cognitive effort is it to get a Hello World program to work? -- not much in dafny at all (for example, I'd argue that you should not have to install and learn VSCode to do this)
- Does the effort needed for each incremental step from there balance the added complexity; does the syntax stay intuitive?
   - Not just one file, but multiple files -- OK, just add them to the command-line -- still flexible in terms of files and options -- still easy to experiment
   - Now we need to construct a consistent build environment: so add include statements, or write a script or write a Makefile, or, soon I hope, write a project file listing files and options -- now we are purposely locking ourselves down (e.g., different sets of options need different project files)

Other points:
- Does the tool work similarly to other similar tools. Compiler tools (g++. java, ...) permit multiple input files on the command-line as needed: so I'd advocate dafny verify/translate/build should as well, leveraging users' experience and intuition 
- dafny run should be as close to dafny verify/translate/build as possible -- so it should allow multiple files on the command-line (both .dfy files and other files relevant to the target environment)

This to me argues for   `dafny run A.dfy B.dfy C.dfy --args 2 3 4`
which is pretty much optimal for all use cases, is to me intuitive, and easily migrates to using project files.
The design that avoids the one instance of --args requires --input multiple times.

For myself, I move back and forth between IDE (VScode) and command-lines and frequently have multiple files and versions of files that it would be a nuisance to keep having to have alternate scripts or project files for. When I do settle on some moderate sized project, then I'm all in to include directives and VSCode (and would welcome project files), except that test suites tend to be scripted command-lines.

**Comment by davidcok** [comment]
PS. I doubt anyone would ever want to mix up dafny options/files and target program options/inputs -- just for clarity I'm willing to disadvantage that use case, which to me again argues for the split command-line design.

**Comment by davidcok** [comment]
PPS. I'd also suggest that the --args in `dafny run A.dfy B.dfy C.dfy --args 2 3 4` can be omitted if none of the user-program arguments look like options or dafny-recognized file names.

**Comment by MikaelMayer** [comment]
> Does the tool work similarly to other similar tools. Compiler tools (g++. java, ...) permit multiple input files on the command-line as needed: so I'd advocate dafny verify/translate/build should as well, leveraging users' experience and intuition

There is one big difference. There is no `g++ run` or `javac run`. Compilation and running is a two-step process for these compilation tools, and the execution is made by running the executable. So it makes sense that compilation can bundle many files together.
If they had designed `g++ run` or `javac run`, I'm pretty sure they would have accepted only one entry file for each. 

So I'd advocate to simplify the interface and make Dafny look affordable like `node` or `python`, where arguments are pass-through. If someone wants to run an executable with multiple files, this is an action a bit advanced but all they need to do is, like for `g++` or `javac:`

`dafny build file1.dfy file2.dfy --out:executable.exe; ./executable.exe`
That's easy enough to justify that `dafny run` can be kept as simple as possible.

**Review by MikaelMayer — APPROVED** [review]
I'd personally approve it, but I acknowledge there is more discussion ongoing that we should sort out before merging.

File: Source/DafnyDriver/InputOptions.cs [hunk 3]
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.Boogie;
+
+namespace Microsoft.Dafny; 
+
+public class InputsOption : CommandLineOption<IEnumerable<string>> {
+  public static readonly InputsOption Instance = new();
+  public override object DefaultValue => new List<string>();
+  public override string LongName => "input";
+  public override string ArgumentName => "file";
+  public override string Description => "Specify additional input files.";
```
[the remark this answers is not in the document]
keyboardDrummer: "(use this repeatedly if you want to include multiple files)" the help already indicates this by placing a `[]` at the end of the description, although that isn't very clear. We could also customise the help to make the `[]` message more explicit.

Source/DafnyDriver/Commands/BuildCommand.cs
```diff
@@ -1,21 +1,30 @@
...
   public IEnumerable<IOptionSpec> Options => new IOptionSpec[] {
     OutputOption.Instance,
     TargetOption.Instance,
     NoVerifyOption.Instance,
     CompileVerboseOption.Instance,
   }.Concat(CommandRegistry.CommonOptions);
 
-  public void PostProcess(DafnyOptions dafnyOptions, Options options) {
+  public Command Create() {
+    var result = new Command("build", "Produce an executable binary.");
```
MikaelMayer, line 17: did I miss something or it should be "Produce an executable binary or a library"?

Source/DafnyDriver/Commands/CommandRegistry.cs
```diff
@@ -30,34 +30,45 @@ static class CommandRegistry {
...
+    FileArgument = new Argument<FileInfo>("file", "input file");
+    FileArgument.AddValidator(ValidateFileArgument());
+
+    FilesArgument = new Argument<IEnumerable<FileInfo>>("file", "input files");
+    FilesArgument.AddValidator(ValidateFileArgument());
+  }
+
+  public static Argument<FileInfo> FileArgument { get; }
+
+  private static ValidateSymbolResult<ArgumentResult> ValidateFileArgument() {
```
MikaelMayer, line 55: Where is this method used?

Source/DafnyDriver/Commands/CommandRegistry.cs
```diff
@@ -121,13 +137,8 @@ static class CommandRegistry {
...
       }
 
-      foreach (var file in context.ParseResult.GetValueForArgument(filesArgument)) {
-        dafnyOptions.AddFile(file.FullName);
-      }
-
       dafnyOptions.ApplyDefaultOptionsWithoutSettingsDefault();
-      commandSpec.PostProcess(dafnyOptions, options);
-
+      commandSpec.PostProcess(dafnyOptions, options, context);
```
MikaelMayer, line 141: A few GitHub actions warnings below

Source/DafnyDriver/InputOptions.cs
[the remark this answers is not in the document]
MikaelMayer: I'm not familiar with the [] syntax. Where is it described? In any case, if you use "files" instead of "file", I as a user would be confused because that would indicate I could do something like `--input file1.dfy,file2.dfy`, whereas if I understand `--input`, it should be one file at a time, correct?

Test/dafny0/DafnyLibClient.dfy
```diff
@@ -1,3 +1,11 @@
+// This test isn't possible with the new UI, because it wants to compile a Dafny library to a .dll
```
MikaelMayer, line 1:
```suggestion
// This test isn't possible with the new CLI commands, because it wants to compile a Dafny library to a .dll
```

Test/dafny0/DafnyLibClient.dfy
```diff
@@ -1,3 +1,11 @@
...
+// and use that as part of another Dafny project. It expects the Dafny source of the library to be included in the .dll
+// as a Source attribute. However, 'dafny build' and `dafny run` always include the Dafny runtime,
```
MikaelMayer, line 3: Isn't there an option to exclude the Dafny runtime from the build?

MikaelMayer (dismissed):
I think it's good, I have a few comments to address. Looking forward to use that very simple interface !

davidcok (changes requested):

Source/DafnyDriver/Commands/CommandRegistry.cs
keyboardDrummer: On line 47 and 50

Test/dafny0/DafnyLibClient.dfy
keyboardDrummer: The old UI has `/useRuntimeLib` yes, but I don't think it would be good to add that here.

davidcok: The translate command should have the option to include the runtime library -- I'm often annoyed/inconvenienced that it does not by default. In fact I would make it the default in all cases and drop the option.

Source/DafnyDriver/Commands/CommandRegistry.cs
MikaelMayer: Funny enough, I was looking for `ValidateSymbolResult` instead of `ValidateFileArgument`. Might be because of the forming habit of having the first identifier being the function's name, like in Dafny.

Source/DafnyDriver/Commands/BuildCommand.cs
```diff
@@ -1,21 +1,30 @@
...
+    result.AddArgument(CommandRegistry.FilesArgument);
+    return result;
+  }
+
+  public void PostProcess(DafnyOptions dafnyOptions, Options options, InvocationContext context) {
+
```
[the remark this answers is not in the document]
keyboardDrummer: Added a test for this in `git-issue-289.dfy`

Source/DafnyDriver/InputOptions.cs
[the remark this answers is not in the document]
keyboardDrummer: > I'm not familiar with the [] syntax. Where is it described?

I don't know, but you can see it if you do `dafny build --help`

Test/dafny0/DafnyLibClient.dfy
```diff
@@ -1,4 +1,4 @@
-// This test isn't possible with the new UI, because it wants to compile a Dafny library to a .dll
+// This test isn't possible with the new CLI commands, because it wants to compile a Dafny library to a .dll
 // and use that as part of another Dafny project. It expects the Dafny source of the library to be included in the .dll
 // as a Source attribute. However, 'dafny build' and `dafny run` always include the Dafny runtime,
 // so the new UI will include the runtime twice for this test causing it to fail.
```
MikaelMayer, line 4:
```suggestion
// so the new CLI will include the runtime twice for this test causing it to fail. 
```

davidcok:
Was it intentional that the default for --compile-verbose changed?

Test/dafny0/DafnyLibClient.dfy
robin-aws: The naming is definitely confusing: `/useRuntimeLib` means "assume we can link to a separate runtime library, so DON'T include that code directly in the output".

> The translate command should have the option to include the runtime library -- I'm often annoyed/inconvenienced that it does not by default. In fact I would make it the default in all cases and drop the option.

We at least need the option to not include the runtime library, because otherwise we can't support separate compilation as in this test case: 
https://github.com/dafny-lang/dafny/tree/master/Test/comp/separate-compilation

Test/comp/ExternDLL.dfy
```diff
@@ -1,4 +1,4 @@
-// RUN: %baredafny build --target=cs %args "%s" %S/ExternDLL2.cs %S/ExternHelloLibrary.dll > "%t"
+// RUN: %baredafny build --target=cs %args "%s" --input %S/ExternDLL2.cs --input %S/ExternHelloLibrary.dll > "%t"
```
robin-aws, line 1: Isn't this a breaking change?

keyboardDrummer: Reverted this change

Test/dafny0/DafnyLibClient.dfy
keyboardDrummer: I don't think discussing `/useRuntimeLib` should be in scope for this PR. I can remove the C# comment that this comment thread resulted from the PR if that makes it easier to merge.

As for the `/useRuntimeLib` discussion, the new UI has a subset of the options it'll eventually need. Whatever is missing, we'll add them when we run into a need for them.

The current state is: `dafny build` always includes the runtime, `dafny translate` does not include the runtime unless `--include-runtime` is specified, since it's meant to produce source code that is built by a native build tool, where the Dafny runtime is specified through configuration so it's only included once even if the build includes multiple Dafny projects.

keyboardDrummer:
> Was it intentional that the default for --compile-verbose changed?

Yes

davidcok (approved):

davidcok (dismissed):

Test/dafny0/DafnyLibClient.dfy
robin-aws: Oh please keep the comment, it's very useful background for the future.

I think the current state is good given `translate` supports the `--include-runtime` option. 

@davidcok Does your workflow just tend to be in two steps, where you `translate` so you can look at the target code, and then build/run it using target language commands? I can sympathize with that, but also think we should bias for end user workflows rather than Dafny contributor workflows. Ideally most end users will have no need to look at the target code and just use `build` and `run`. If they are using `translate` because they are integrating with existing code, hopefully there would be more context to help compile and run the system (e.g. a `*.csproj` file that includes `DafnyRuntime`).

davidcok:
I think I'm acting more as a Dafny end user here, in trying out and documenting the workflow to integration Dafny + language X.  In that situation, I use 'translate' to understand that the extern declarations are doing what I expect, and then 'run' to run the program -- but the 'run' requires additional .java or .class or .jar files and in addition to possible .dfy containing shims. So prefacing all these with --input is a real pain. The workaround is to do a build step and then run but that is then a two step workflow instead of one (a single --inputs with a comma-separated list, or --args for arguments would be better, but I was unconvincing in those discussions)

robin-aws:
Oh one other thing: the output of `dafny run --help` doesn't mention `--`, and I think that would be very helpful.

keyboardDrummer:
In that situation, I use 'translate' to understand that the extern declarations are doing what I expect, and then 'run' to run the program -- but the 'run' requires additional .java or .class or .jar files and in addition to possible .dfy containing shims. So prefacing all these with --input is a real pain. The workaround is to do a build step and then run but that is then a two step workflow instead of one (a single --inputs with a comma-separated list, or --args for arguments would be better, but I was unconvincing in those discussions)

> I use 'translate' to understand that the extern declarations are doing what I expect

I think that if there's a situation in which externs declarations do not correctly match with specified native code, `dafny run` _should_ already provide an error about this, and you shouldn't need to call `dafny translate`.

> the 'run' requires additional .java or .class or .jar files and in addition to possible .dfy containing shims. So prefacing all these with --input is a real pain

The solution that I have in mind for this is project files. Does that sound like a good solution to you?

>  I think I'm acting more as a Dafny end user here, in trying out and documenting the workflow to integration Dafny + language X.

I'm not sure what you mean by "integration Dafny + language X". I think about the use-cases through the following flow-chart:

- Are you working on a library or an executable?

  - Executable. Do you want to run it once or multiple times?
    - Once. Use `run`
    - Multiple times.  Use `build`

  - Library. Is your library meant to be consumed from Dafny or from another language?
      - From Dafny. Let your consumers use a submodule to import your Dafny code
      - From another language. Use `dafny translate` and a native package manager to publish a package to the native package repository

docs/dev/news/2823.feat
```diff
@@ -0,0 +1 @@
+Introduce a new Dafny CLI UI that complies with the POSIX standard and uses verbs to distinguish between use-cases.
```
robin-aws, line 1: Oh we didn't mention the new CLI already? Glad I asked then! :)

Can we say a little about how to access the new CLI? Perhaps just `Type `dafny` to view the documentation`. Otherwise it sounds like users might need to obtain and use a new binary.

robin-aws (approved):

davidcok:
We do need project files for complex projects. However, for a new user, being told that in order to have a program with more than one file they have to write a project file is to me a big impediment for new users.
So it depends a lot on how easy it is to create a new simple project file.

- David

davidcok:
You can't use include directives if your additional files are from a different language

MikaelMayer:
Right, good point !

PR #2840, opened 2022-10-04
Moving release procedures into a dev folder instead of the user-facing docs folder

davidcok:


cpitclaudel (changes requested):
Is there a reason to rename from `docs/dev/RELEASE` to `dev/RELEASE`?

davidcok:
The docs folder contains user-facing documentation (which is copied as is into the documentation snapshots). The Release procedure is development information. Robin wanted it kept in git (I had earlier proposed moving it to the wiki). So a dev folder seemed a good solution. Other ideas are welcome.

cpitclaudel:
We could exclude it from the build if we really want to not have it on the public website, but `docs/` is the first place I'd look when searching for that info, and I don't think `dev` is standard for that?

But is it really bad to have it on the website?  Having a "developer documentation" section is a good thing, I would think.

davidcok:
Abandoning this PR -- not moving the dev material - instead not copying it into the snapshots -- cf. PR 2845

Issue #1239, opened 2021-05-31
Officially recommended way to use Dafny code from other languages?


hmijail:
Probably it's a typical use case that a kernel of Dafny code needs to interact with code in other languages, like libraries or an existing codebase.

There are multiple ways to approach this. Is there an officially sanctioned / recommended way to do it, as to maximize the reliability of the final product?

* Given that Dafny compiles to C#, and has done so for longer than to any other target language, it sounds plausible that C# should be the most reliable option to use to interface with Dafny. Is there an official target language recommendation?

* The Dafny-generated C# starts by mentioning in a comment recommended compilation flags, which makes me doubt whether that's all that's needed. Does the Dafny compiler do anything worthy of mention (flags?) when it does its own compilation of the intermediate code? How to make sure that we get the same result as the Dafny toolchain would get?
    - Possible feature request: document the intermediate language compiler calls used by the Dafny compiler. If they can change between versions, make the Dafny compiler report those compiler calls.

* To avoid this doubt we could treat the Dafny compiler as a black box and dynamically load the resulting DLL, therefore avoiding introducing problems with our compilation process. But that is more laborious and errorprone than working with the intermediate source code, so is it worth it?

PR #2848, opened 2022-10-04
feat: Add a release script

cpitclaudel:
This removes most of the manual steps in the release process, except for the manual CI workflow trigger, which is left for later.

It also changes the way we do release notes: instead of updating a single file (which creates repeated PR conflicts), each release note item goes into a separate text file numbered after the corresponding PR.  The details of this are written up in `docs/news/README` and a link points to that folder in RELEASE_NOTES.md to avoid confusion.


docs/dev/RELEASE.md
```diff
@@ -2,83 +2,69 @@
...
+   prepare a release commit, and push it.
+
+2. Kick off the deep test suite by navigating to
+   <https://github.com/dafny-lang/dafny/actions/workflows/deep-tests.yml>,
+   clicking the "Run workflow" dropdown, selecting the newly created branch, and
+   clicking the "Run workflow" button. The automation for releasing below will
+   check for a run of this workflow on the exact commit to release.  (TODO:
+   Run this automatically as part of the prepare-release script.)
+
+3. Run `Scripts/release.py $VER release` from the root of the repository.  The
```
MikaelMayer, line 21: Can you explicitely say to wait till the tests are finished, and how much time it typically last?

docs/dev/news/README.md
```diff
@@ -0,0 +1,15 @@
...
+Each file in this folder describes one change in Dafny since the latest release.
+
+Files in this directory are named `<PR number>.<kind>` and each contains release notes for one merged PR.  `kind` is `feat` or `fix`.  No need to include a link to the PR in the file: it will be added automatically.
+
+## Examples
+
+`1234.fix`
+
+```
+Improved toaster settings prevents Dafny from ever burning toast. ??
```
MikaelMayer, line 12:
```suggestion
Improved toaster settings prevents Dafny from ever burning toast.
```

docs/dev/news/README.md
```diff
@@ -0,0 +1,15 @@
...
+```
+
+At release time, this becomes a bullet point in the "Bug fixes" section of `RELEASE_NOTES.md` with a link to <https://github.com/dafny-lang/dafny/pull/1234>.
```
MikaelMayer, line 15: Can you exactly say how it will be included? I assume that for the above, it will automatically add the line

- fix: Improved toaster settings prevents Dafny from ever burning  (https://github.com/dafny-lang/dafny/pull/1234)

Scripts/prepare_release.py
```diff
@@ -0,0 +1,363 @@
+#!/usr/bin/env python
```
MikaelMayer, line 1: I have a challenging question for you. How hard would it be to actually convert this code to dafny so that it can be run with `dafny -noVerify ` ?
That way, you wouldn't need to have `assert ends_with('...')` that I see in the code.
Another advantage would be that we would be able to execute tests that would greatly enhance the maintenance process. There are tons of regexps that might break that would be easy to fix if only we saw what the typical input and typical output should look like.

robin-aws (changes requested):
On a quick glance this looks awesome, very much appreciate the contribution to streamlining the release process!

I will find time soon to review properly, but my high-level strong preference is to not require doing any work on a local machine (fully scripted or not, but to find a way for it to happen in a GitHub Action workflow. There is always risk for local development state to corrupt the release process, sometimes very subtly.

Scripts/prepare_release.py
```diff
@@ -0,0 +1,363 @@
...
+        for _, paths in self.fragments.items():
+            for pth in paths:
+                pth.unlink()
+
+class Version(NamedTuple):
+    VERNUM_RE = re.compile("^(?P<prefix>[0-9]+[.][0-9]+[.][0-9]+)(?P<suffix>-.+)?$")
+
+    prefix: str # Main version number (1.2.3)
+    rdate: date # Release date
+    suffix: str # Optional marker ("alpha")
```
MikaelMayer, line 107: I would prefer very much to have complete and explicit names, so that it does not feel like a compiled code :-)

```suggestion
    version_number: str # Main version number (1.2.3)
    release_date: date # Release date
    version_number_suffix: str # Optional marker ("alpha")
```

Scripts/prepare_release.py
```diff
@@ -0,0 +1,363 @@
...
+                rendered.append(entry)
+        return "\n\n".join(rendered)
+
+    def delete(self):
+        for _, paths in self.fragments.items():
+            for pth in paths:
+                pth.unlink()
+
+class Version(NamedTuple):
+    VERNUM_RE = re.compile("^(?P<prefix>[0-9]+[.][0-9]+[.][0-9]+)(?P<suffix>-.+)?$")
```
MikaelMayer, line 103:
```suggestion
    VERSION_NUMBER_REGEX = re.compile("^(?P<prefix>[0-9]+[.][0-9]+[.][0-9]+)(?P<suffix>-.+)?$")
```

MikaelMayer (changes requested):
Yay, wonderful time-saver script, which will help us release more often I hope.

MikaelMayer:
> I will find time soon to review properly, but my high-level strong preference is to not require doing any work on a local machine (fully scripted or not, but to find a way for it to happen in a GitHub Action workflow. There is always risk for local development state to corrupt the release process, sometimes very subtly.

I think that, in a first step, mimicking what we used to do manually locally anyway is a good compromise. I tested in on my Windows machine and this script worked.
Later, we can always try to automate more the process and make it purely online, but I think that would be an improvement, not a blocker over this PR.

docs/dev/news/README.md
cpitclaudel: It actually becomes 

```markdown
## Bug fixes

- Improved toaster settings prevent Dafny from ever burning toast.
  (https://github.com/dafny-lang/dafny/pull/1234)
```

I don't think the README should say this, though.

Scripts/prepare_release.py
cpitclaudel: It would be anywhere between a huge pain and impossible, because of the range of libraries used: XML parsing, regular expressions, HTTP requests, dates and times, text manipulation, file system functions, quoting, etc (but not JSON parsing, because we do have that! ^^)

I can remove these assertions if you'd like; they'd just become requires in Dafny.  They are here to make sure that the messages are readable.

Python has better support for tests than Dafny; I can add tests if you want (but there's decreasing returns to spending days working on a script that runs once a month ^^)

Scripts/prepare_release.py
cpitclaudel: Renamed to `VERSION_NUMBER_PATTERN`, but I find it much harder to read (because the long names hide the rest of the code)

cpitclaudel:
> Fair enough, I just need the deeper review to verify we're not doing more locally than before. :)

Do you want to block this, or are you OK with Mikael's review?

docs/dev/news/README.md
MikaelMayer: I agree this is an implementation detail, but it's a detail important enough so that I seriously think it's worth mentioning it.
Otherwise, we might get developers down the road who will want to prefix their subjects with "fix:" or "feat:" or want to write something on two lines, or they will want to include a custom URL, or they won't know if they have to put a capital or not on the first letter, or if they need to include a space at the end.... all of this because the transformation is only `provide`d but not `reveal`ed :-)

And we should get rid of the need to add "??". By default, the URL should be appended automatically. I don't see a clear use case where a developer would want to include this link anywhere in the subject link.

cpitclaudel:
> I will find time soon to review properly, but my high-level strong preference is to not require doing any work on a local machine 

Agreed, and this script can run from CI as well.  The only difficulty would be creating a bot account on github that is allowed to push to the repo.

But since this sort of script invariably takes a few iterations to refine and since it's much easier to debug locally than on CI, I think it's best to start locally.

Scripts/prepare_release.py
MikaelMayer: We will probably want to run this script once a week. I'm not saying we have to convert this script to Dafny, but if it's challenging to write such a script in Dafny, then maybe it's a yellow flag as per how difficult it is to integrate with an existing environment.
Dafny compiles to python as well, so it should not be hard to have wrapper libraries.

Scripts/prepare_release.py
MikaelMayer: In the same answer spirit, it's not clear with "prefix" or "suffix" what of the version numbers are included in which :-)

Maybe the following could reach a consensus?
`prefix` => `major_minor_patch`
`suffix` stays the same.

Scripts/prepare_release.py
MikaelMayer: Thanks for the renaming, I think it's appropriate. How does the the long name hide the rest of the code? I'm not sure to understand this part.

Scripts/prepare_release.py
cpitclaudel: It's hundreds of lines of wrapper code / dozens of wrappers; I don't think Dafny is ready to be a scripting language (and that's OK)

docs/dev/news/README.md
cpitclaudel: OK, done.  I also added a test to make sure that the format was correct.

> And we should get rid of the need to add "??". By default, the URL should be appended automatically. 

`??` was a typo, you spotted it and I already accepted your patch.  Did I miss something else?

Scripts/prepare_release.py
cpitclaudel: OK, I renamed `prefix` to `main` and `suffix` to `identifier` (the name used for this in semantic versioning)

Scripts/prepare_release.py
cpitclaudel: Very long function and variable names make code harder to read, because the relevant bits of syntax / control end up further from each other, at least in my experience.

docs/dev/news/README.md
MikaelMayer: Oh I probably missed your fix. Thanks :-)

Scripts/prepare_release.py
```diff
@@ -87,56 +114,58 @@ def render(self) -> str:
             if ext not in self.fragments:
                 continue
             rendered.append(f"## {title}")
-            for pth in self.fragments[ext]:
-                text = pth.read_text(encoding="utf-8")
-                link = f"(https://github.com/dafny-lang/dafny/pull/{pth.name})"
-                entry = indent(f"- {text}\n{link}", "  ").lstrip()
+            for fr in self.fragments[ext]:
+                link = f"(https://github.com/dafny-lang/dafny/pull/{fr.pr})"
+                entry = indent(f"- {fr.contents}\n{link}", "  ").lstrip()
```
MikaelMayer, line 119: I don't think you want to add a newline between the contents and the link. At least in GitHub, where the release nodes can be seen as a formatted markdown file, even though the link seems long, it will actually look like this: #1239

Scripts/prepare_release.py
```diff
@@ -57,24 +56,52 @@ def unwrap(opt: Optional[A]) -> A:
...
+    >>> import tempfile
+    >>> with tempfile.TemporaryDirectory() as tmpdir:
+    ...   fpath = Path(tmpdir) / "1234.fix"
+    ...   _ = fpath.write_text("Improved toaster settings.\nDafny will not burn toast again.", encoding="utf-8")
+    ...   print(NewsFragments.from_directory(tmpdir).render())
+    ## Bug fixes
+    <BLANKLINE>
+    - Improved toaster settings.
+      Dafny will not burn toast again.
+      (https://github.com/dafny-lang/dafny/pull/1234)
```
MikaelMayer, line 80: Thanks for the testing code ! It helps a lot.

```suggestion
    - Improved toaster settings.
      Dafny will not burn toast again (https://github.com/dafny-lang/dafny/pull/1234)
```
```suggestion
    - Improved toaster settings.
      Dafny will not burn toast again.
      (https://github.com/dafny-lang/dafny/pull/1234)
```
See comment below for the formatting.
Besides the extra newline that is not necessarily ([have a look at here to see how release notes are currently formatted](https://github.com/dafny-lang/dafny/releases))I would also add that you would now be forcing users (and me) to do something that we don't want: Add a period before the parentheses.
So, in your example, and in your README, could you be specific about this point? Alternatively (and that would be my favorite method), you could detect the last character of the trimmed string, and if it's a period, you'd remove it.

MikaelMayer (changes requested):
Still one more cosmetic change to make it easier for everyone and ensure we are markdown-friendly.

Scripts/prepare_release.py
cpitclaudel: Done: I've removed the newline.  I find that it makes reading the release notes harder, but I won't fight about this.

The existing `RELEASE_NOTES` are inconsistent on whether to use a period,  so I've changed the code to add a period in all cases.  The version above with one period for the first sentence and no period for the second doesn't look right at all to me; do you like it better?

MikaelMayer: >  I find that it makes reading the release notes harder

I won't argue since you implemented my change, thanks :-) but I'm just curious.
Can you please clarify, is it because you read the source of release notes instead of in a browser? Or is the browser rendering itself harder to read? To me, the first image here:
image
is much nicer than the alternative that you might find more appealing based on your script:
image
I find that the second one breaks the flow, it feels like every link is important when the only thing I would like out of release notes is to read them in diagonal.

Regarding two lines: I think we should not even consider the possibility that a feat/fix line in a release notes should have two sentences. I know you added support for it, but it's not something we have done in the past and I don't see a reason why we should embrace those. So if we were to add paragraphs, I'd agree with you, we should end every sentence with a full stop, and make the link its own period-less sentence after that.
However, for a single phrase, with the comment in parentheses containing the link being part of the phrase, there should be no period between the phrase and the parentheses. If there was a period, it should come afterwards.

Now, one more principle to consider: We use bullet points to express each individual release notes, and many bullet points are not complete sentences but just fragments. In this case I think it's [conventional wisdom](https://www.instructionalsolutions.com/blog/bulleted-list-punctuation) to omit the period at the end of the sentence (after the parentheses containing the link).

docs/dev/news/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Files in this directory are named `<PR number>.<kind>` and each contains release notes for one merged PR.  `kind` is `feat` or `fix`.  No need to include a link to the PR in the file: it will be added automatically.
+
+## Examples
+
+`1234.fix`
+
+```
+Improved toaster settings prevent Dafny from ever burning toast.
+Instead, toast will always come out crispy and nice
```
MikaelMayer, line 13: If you don't like my comment above and still find that multi-line release notes are relevant, like, we want to output more user-centric release notes that not only have a title, but a small example or explanation:

```suggestion
Added support for :==, like this:
    method Test(x: int, y: int) {
      var x := 1;
      x :== 1; // Assertion succeeds.
      x :== 2; // Assertion fails.
    }
```

Then, in this case, if it's multiline, you can add a newline (no automatically inserted period) and put the link in one more link. But if someone writes only one line, then if there is a period, then you move it after the (link) that is inserted in the sentence.

Scripts/prepare_release.py
cpitclaudel: Good point.  I implemented your suggestion below: no auto punctuation fix and no newline before link, unless there's already multiple lines in the entry.

We can refine the rendering further, probably when we have more data (before the next release)

docs/dev/news/README.md
```diff
@@ -6,18 +6,35 @@ Files in this directory are named `<PR number>.<kind>` and each contains release
...
+
+  ```
+  Two new toast patterns:
+  - Dafny waterfall logo
+  - Dafny haircut logo
+  (They are the same.)
+  ```
 
-At release time, this becomes a bullet point in the "Bug fixes" section of `RELEASE_NOTES.md` with a link to <https://github.com/dafny-lang/dafny/pull/1234> and an additional period if needed, like this:
+At release time, these two files become a bullet points a new section of `RELEASE_NOTES.md` with links to the corresponding PRs, like this:
```
MikaelMayer, line 24:
```suggestion
At release time, these two files become bullet points in a new section of `RELEASE_NOTES.md` with links to the corresponding PRs, like this:
```

docs/dev/RELEASE.md
```diff
@@ -2,80 +2,64 @@
...
+   the release notes from `RELEASE_NOTES.md`.  If this is not a pre-release,
+   check the box to create a new discussion based on the release.
+
+5. Push the "Publish" button. This will trigger yet another workflow
+   that will download the published artifacts and run a smoke test
+   on multiple platforms. Again you can watch for this workflow at
+   <https://github.com/dafny-lang/dafny/actions>.
+
+6. Create a pull request to merge the newly created branch into `master` (the
+   script will give you a link to do so).  Get it approved and merged.
```
robin-aws, line 40: I don't love the idea of releasing from a branch before merging it to mainline. Could we flip the order of steps to do this first, even if it means customizing the GHA that looks for the deep tests? Are we just trying to save the extra delay on waiting for the CI?

RELEASE_NOTES.md
```diff
@@ -1,15 +1,13 @@
 # Upcoming
 
-- fix: Compiled lambdas now close only on non-ghost variables (https://github.com/dafny-lang/dafny/pull/2854)
-- fix: Crash in the LSP in some code that does not parse (https://github.com/dafny-lang/dafny/pull/2833)
+See [docs/dev/news/](docs/dev/news/).
```
robin-aws, line 3: I'd add a stronger comment that this file is no longer used as part of the release process.

docs/dev/RELEASE.md
```diff
@@ -2,80 +2,64 @@
...
-10. Update the Homebrew formula for Dafny (see below).
-    Note that it is fine to leave this for the next day,
-    and other members of the community may update the formula
-    in the meantime anyway.
-
-11. Announce the new release to the world.
+0. Ensure that you are in a clean and up-to-date repository, with the `master`
+   branch checked out and up-to-date and no uncommitted changes.
+
+1. Select a version number `$VER` (e.g., "3.0.0" or "3.0.0-alpha"), then run
```
robin-aws, line 8: Could we have the script infer the correct next version number based on whether there are any new `feat` or `feat!` news items? Or at least verify you selected the right version number.

robin-aws (changes requested):
I just reviewed the new workflow and did not look closely at the implementation, will leave that to other reviewers.

On the one hand I definitely love the new approach. On the other I do feel like this is increasing the work that has to happen on a developer's machine, since in the old way we were manually creating the release notes incrementally in each PR. Let me think a bit more on how to avoid that.

docs/dev/RELEASE.md
cpitclaudel: Releasing from master forces a freeze to avoid concurrent commits; we've been bitten by that before.  What issue do you see with releasing from a branch?  Is it the security aspect? (I think we already allow releasing by uploading a custom-packaged zip, so I haven't worried about that part)

RELEASE_NOTES.md
cpitclaudel: This is a user-facing file, though, so should we add this type of comments there?
Maybe as a comment that doesn't get rendered?

docs/dev/RELEASE.md
cpitclaudel: Sure, do you think it's worth it given that we plan to revise our version policy soon?

docs/dev/RELEASE.md
```diff
@@ -2,80 +2,64 @@
...
+   script will check that the repository is in a good state, create and check
+   out a new release branch, update `Source/version.cs` and `RELEASE_NOTES.md`,
+   prepare a release commit, and push it.
+
+2. Kick off the deep test suite by navigating to
+   <https://github.com/dafny-lang/dafny/actions/workflows/deep-tests.yml>,
+   clicking the "Run workflow" dropdown, selecting the newly created branch, and
+   clicking the "Run workflow" button. The automation for releasing below will
+   check for a run of this workflow on the exact commit to release.  (TODO:
+   Run this automatically as part of the prepare-release script.)
```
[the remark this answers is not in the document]
cpitclaudel: I would prefer to keep it here.  It's good to be reminded of the issue every time we go through these instructions.  But le me know if you disagree and I'll move it.

docs/dev/RELEASE.md
MikaelMayer: I think it's not only pretty common to release from a branch before merging it (that way, it's even possible to cherry-pick bug fixes independently without introducing new features), but in our case, as Clement said, we already encountered the case when we wanted to release after merging with master; however, there was one more commit on master since we started to prepare the release, and this commit was merged unintentionally with the release.
So although I had wished that releases are made from the history of master branches, in practice, this is dangerous.

RELEASE_NOTES.md
MikaelMayer: I don't think we should add a comment about the release process here, but I like the idea of also redirecting users to the folder where the fragments are so that they can see what to expect in the next release.

docs/dev/RELEASE.md
MikaelMayer: I'm afraid we would loose the ability to bump the major version number with such an auto-detect 😮 
My best suggestion is that:
- The script detect feat or feat! extensions
- If there are, it suggests to increment the minor version number, and if the user presses ENTER, this is it.
- If there are none (only fixes), it suggests to increment the patch version number, and if the user presses ENTER, this is it
- But if the user decides to put another version number, it takes it and continue.

MikaelMayer (commented):
Removing my "request change"

MikaelMayer (dismissed):
Commenting does not remove the "Request change". Let me approve it.

docs/dev/RELEASE.md
cpitclaudel: This is a lot of work for saving the user from typing 3 numbers ^^  I vote for leaving this for future improvement

RELEASE_NOTES.md
robin-aws: Ah I missed the fact that the script is still updating this file, I thought it was going to be frozen in this state forever more. Nevermind. :)

docs/dev/RELEASE.md
robin-aws: Okay that's a better reason, I'm happy with this.

robin-aws (dismissed):

robin-aws (dismissed):

robin-aws (approved):

PR #2849, opened 2022-10-04
feat: Rewrite the AutoExtern CLI and allow generalized name rewrites

### LINKED ISSUE #2839: AutoExtern: Allow renaming symbols
We have a workaround in compiler-bootstrap for the fact that we have two copies of `Microsoft.Dafny`: one is the handwritten one and one has the model of the AST.  If AutoExtern supported renaming some symbols, we could merge them and remove the `sed` step from the Makefile.

cpitclaudel:
This PR makes AutoExtern more flexible, removing the need for a workaround in the compiler-bootstrap project.

Description from https://github.com/dafny-lang/compiler-bootstrap/commit/a49b299fda7479d168a8c34237a890214804e161:

The Microsoft.Dafny C# namespace is very large, and we don't model every part of it¹.  Instead, we have two models:

- One, handwritten, capturing things like `BuiltIns`, `Cloner`, `ConcreteSyntaxTree`, `ErrorReporter`, etc.
- One, auto-generated, translating all of the AST.

This started causing issues with the introduction of `Dafny.IToken`, because that interface is defined in the AST but used in the `ErrorReporter`, so both of our models needed to have their own definition of `IToken`.

Dafny doesn't support merging modules defined in different files, so previously the two models were two different Dafny modules.  They both compiled to the same C# namespace using `:extern` annotations, but since Dafny checks for unicity of `:extern` module names we had to use a different name (`__AUTOGEN__`) and then run `sed` on the output.

Now we have just one module.  The main trick was extending `AutoExtern` to support renaming symbols, so that we could adjust the generated code to refer to local symbols correctly (concretely, referring to `Microsoft.Boogie.X` as `Boogie.X`).

¹ AutoExtern isn't good enough yet to translate all of Dafny automatically; for example, it doesn't handle methods.

Closes #2839


Source/AutoExtern/Program.cs
```diff
@@ -413,24 +418,102 @@ public static class Program {
...
+    return rootCommand.Invoke(args);
+  }
+
+  private static List<(string, string)> ParseRewrites(ArgumentResult result) {
+    var parsed = new List<(string, string)>();
+
+    foreach (var tok in result.Tokens) {
+      var split = tok.Value.Split(":");
+      if (split.Length != 2) {
+        result.ErrorMessage = "--rewrite takes a pair of string separated by `:`";
```
keyboardDrummer, line 502: string`s`

keyboardDrummer (commented):
Could you add a test?

keyboardDrummer:
> Now we have just one module. The main trick was extending AutoExtern to support renaming symbols, so that we could adjust the generated code to refer to local symbols correctly (concretely, referring to `Microsoft.Boogie.X` as `Boogie.X`).

Do you mean one Dafny module? How is that achieved if you still have the generated and the handwritten dafny file, and Dafny doesn't allow merging modules from different files? I think I understand the problem from your description, but not the resolution.

cpitclaudel:
Tests added :)

keyboardDrummer (approved):

PR #2854, opened 2022-10-05
Fix: Compiled lambdas now close only on non-ghost variables

### LINKED ISSUE #2690: Incorrect reference to ghost variable in compiled code
Input:

```
method M<T>(ts: seq<T>) {
  ghost var pre := t => t in ts;
  var f := t requires pre(t) => false;
}
```

Output:

```
(4832,12): error CS0103: The name '_2_pre' does not exist in the current context
```

The problem also happens in javascript, where it can cause runtime crashes:

```
    static M(ts) {
      let _0_f;
      _0_f = ((_1_pre) => function (_3_t) {
        return false;
      })(_2_pre);
      return;
    }
```

Instead of tagging this as a soundness bug, I'll just make a meta-issue for the soundness part.

The workaround is to inline the ghost var.

### LINKED ISSUE #2852: Lambdas forget not to close over ghost values at compile time
Running the following code shows that Dafny's c# compiler forgot to remove ghost variables from the closed variables.
```dafny
// Returns a function that computes the sum of n consecutive integers starting at pos
function method Sum(
  ghost remaining: nat,
  n: nat
): (p: nat -> nat)
  decreases remaining
  requires remaining == n
{
  (pos: nat) =>
    var x: nat := if n == 0 then 0 else Sum(remaining - 1, n - 1)(pos+1) + pos;
    x
}
method Main() {
  print Sum(5, 5)(10);
}
```
Run:
```
> dafny -compile:3 -spillTargetCode:2 -compileTarget:cs the_test.dfy

Errors compiling program:
(1836,15): error CS0103: The name 'remaining' does not exist in the current context
```

Might be related to https://github.com/dafny-lang/dafny/issues/531 ?

MikaelMayer:
This PR fixes the problem that lambdas were closing over ghost variables in compiled code, resulting at best in compilation errors (in C#), at worst in a soundness issue crash at run-time (in js). As you can see, the fix is very small. I added an integration test to ensure this behavior is correctly captured.

Fixes #2852
Fixes #2690 


keyboardDrummer (approved):

cpitclaudel (dismissed):
Looks good to me.  Could you check if this PR also fixes bug https://github.com/dafny-lang/dafny/issues/2690 ? If so, it would be good to copy the test from there (and if not, it would be good to add a note to that issue)

MikaelMayer:
> Looks good to me. Could you check if this PR also fixes bug #2690 ? If so, it would be good to copy the test from there (and if not, it would be good to add a note to that issue)

As a matter of fact, it does fix #2690. Let me add them as tests.

cpitclaudel (approved):

PR #2876, opened 2022-10-10
Fix: Node CLI arguments for dafny to have the right order

### LINKED ISSUE #2867: Behaviour of Dafny program command-line arguments with "dafny build" inconsistent on Javascript
```dafny
// An attempt at a CLI that behaves the same in all target languages
method Main(args: seq<string>) {
  if |args| == 0 {
    return;
  }
  print args[1..], "\n";
}
```

```
> dafny /compile:3 src/Scratch.dfy /compileTarget:js --args 1 2 3

Running...

1,2,3
> dafny build -t:js src/Scratch.dfy

> a483e79acc1e:libraries salkeldr$ node src/Scratch.js 1 2 3
/Users/salkeldr/Documents/GitHub/libraries/src/Scratch.js,1,2,3
```

MikaelMayer:
This PR fixes #2867
Tests for C#, JavaScript and C++
I tried to spend some time to make it work for Java but it seems we don't have a standard procedure for launching programs, so I leave it unspecified for now.


robin-aws (dismissed):
Nice, thanks for the quick fix! My only ask is to explicitly call this out as a `fix:` release note. This could be considered a breaking change, but I think it's reasonable to say it was underspecified and this is clearly better behaviour.

> I tried to spend some time to make it work for Java but it seems we don't have a standard procedure for launching programs, so I leave it unspecified for now.

That's fine for this PR, but FYI https://github.com/dafny-lang/dafny/blob/master/Test/comp/manualcompile/ManualCompile.dfy has the blueprint you're looking for. But I would also love to add support for command-line arguments to `%testDafnyForEachCompiler` :)

MikaelMayer:
> That's fine for this PR, but FYI https://github.com/dafny-lang/dafny/blob/master/Test/comp/manualcompile/ManualCompile.dfy has the blueprint you're looking for. But I would also love to add support for command-line arguments to `%testDafnyForEachCompiler` :)

Oh thanks, I did not know about this file ! That was very helpful.
I was able to add the tests for the remaining 3 back-ends in no time !

docs/dev/news/2876.fix
```diff
@@ -0,0 +1 @@
+Same argument order for stand-alone javascript files than for other languages
```
keyboardDrummer, line 1: than => as

keyboardDrummer (dismissed):

docs/dev/news/2876.fix
MikaelMayer, line 1:
```suggestion
Same argument order for stand-alone javascript files as for other languages
```

keyboardDrummer (approved):

cpitclaudel:
Can we refine the release notes for this?  It's hard to understand what this means without context (and reading the PR description doesn't help too much either)

PR #2877, opened 2022-10-10
fix: string-seq equality in JS runtime

### LINKED ISSUE #2581: Javascript runtime doesn't consider empty sequences equal to empty strings
```dafny
method Main() {
  assert IsEmpty("");
  expect IsEmpty("");  // Fails on Javascript only
}

predicate method IsEmpty<T>(s: seq<T>) {
  s == []
}
```

I found this because I tried to use `Seq.Reverse` from the [standard library](https://github.com/dafny-lang/libraries/blob/master/src/Collections/Sequences/Seq.dfy#L330), and it never terminated when executed because the base condition was a generic `s == []` check.

alex-chew:
Fixes #2581.


atomb (approved):
Nice, simple fix!

PR #2879, opened 2022-10-10
bug: Fixes #2733, #2559, #2882 - extra newline in Java target

### LINKED ISSUE #2559: Each compile target should ensure that output streams (stdout and stderr) are flushed before a program exits.
And there should be tests added to check that print statements without newlines do get printed eventually.

### LINKED ISSUE #2733: Code compiled to Java prints a final newline
Input:

```dafny
method Main() { print "a"; }
```

Output:

```terminal
$ Dafny /compile:4 /compileTarget:java tmp.dfy

Wrote textual form of target program to tmp-java/tmp.java
Additional target code written to tmp-java/_System/nat.java
Additional target code written to tmp-java/_System/__default.java
Running...

a
$ Dafny /compile:4 /compileTarget:cs tmp.dfy

Running...

a$
```

Notice how Java has an extraneous newline

### LINKED ISSUE #2882: bug: Java outputs expect message to stderr; other platforms to stdout
A Java build, when run standalone (not through `dafny run`) puts the failure messages from expect statements to stderr;
cs and js send them to stdout.

davidcok:
Fixes #2733. Fixes #2559. Fixes #2882.


cpitclaudel (changes requested):
Thanks! The original code looks wrong to me (and, I think, so does the modified code), because both will deadlock if a process tries to write a lot to standard error and then a lot to standard output: the standard output buffer will fill up, the program will block, and the end of the standard output stream will never be reached.

Could we just remove the two output redirects?  I'm not sure why we didn't do that when we fixed [the corresponding issue for Javascript](https://github.com/dafny-lang/dafny/pull/1824) (cc @MikaelMayer)

davidcok:
Aargh! 5 compilers with 4.5 solutions to the same problem. The D-in-D compiler is going to fix this, right?

This will take a bit more work. For the second try I emulated the solution in go/python (which does not use redirects).
That works fine for Java with a simple `dafny run` test, but in the CI environment, std-error is getting lost, so I've a bit more debugging to do.

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -3648,18 +3648,7 @@ private class SpecialNativeType : UserDefinedType {
...
-        if (process == null) {
-          return false;
-        }
-        process.WaitForExit();
-        return process.ExitCode == 0;
-      } catch (System.ComponentModel.Win32Exception e) {
-        outputWriter.WriteLine("Error: Unable to start go ({0}): {1}", psi.FileName, e.Message);
-        return false;
-      }
+      return RunProcess(psi, null, "go", outputWriter) == 0;
```
MikaelMayer, line 3651: Did you mean 
```suggestion
      return 0 == RunProcess(psi, null, "go", outputWriter);
```
?
But other than that, since you are never using the "integer" part of RunProcess, why don't you have `RunProcess` return `true` if it succeeded, and `false` otherwise? That way you don't need an extra superfluous equality.

davidcok: Because I expect that eventually we will care about the return code from the target program and RunProcess will propagate it so `dafny run` can return it (or at least report it). However, if you insist, I can push that equality check into RunProcess and out of the individual compilers.

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -15,6 +15,7 @@
 using System.ComponentModel.Design;
 using System.Diagnostics;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
```
cpitclaudel, line 18: Do you need this?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -64,6 +66,23 @@ public abstract class SinglePassCompiler : Plugins.Compiler {
...
+      try {
+        if (process == null) {
+          process = Process.Start(psi);
+          if (process == null) {
+            return -1;
+          }
+        }
+        process.WaitForExit();
+        return process.ExitCode;
+      } catch (System.ComponentModel.Win32Exception e) {
```
cpitclaudel, line 78: This looks suspicious: Win32?

Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/Helpers.java
```diff
@@ -213,7 +213,7 @@ public static void withHaltHandling(Runnable runnable) {
         try {
             runnable.run();
         } catch (DafnyHaltException e) {
-            System.err.println("[Program halted] " + e.getMessage());
+            System.out.println("[Program halted] " + e.getMessage());
```
cpitclaudel, line 216: `err` sounds right; why change it to `out`?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -64,6 +66,23 @@ public abstract class SinglePassCompiler : Plugins.Compiler {
...
     public CoverageInstrumenter Coverage;
 
+    public virtual int RunProcess(ProcessStartInfo psi, Process process, String platform, TextWriter outputWriter) {
+      try {
+        if (process == null) {
+          process = Process.Start(psi);
+          if (process == null) {
+            return -1;
+          }
+        }
```
cpitclaudel, line 75:
```suggestion
        process ?= Process.Start(psi);
        if (process == null) {
          return -1;
        }
```

Source/DafnyCore/Compilers/Compiler-java.cs
davidcok: Not any more.

Source/DafnyCore/Compilers/Compiler-go.cs
MikaelMayer: No I won't insist, I think you have a point.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -64,6 +65,23 @@ public abstract class SinglePassCompiler : Plugins.Compiler {
 
     public CoverageInstrumenter Coverage;
 
+    public virtual int RunProcess(ProcessStartInfo psi, Process process, String platform, TextWriter outputWriter) {
```
cpitclaudel, line 68: I think this API could be a bit cleaner: how about always passing in a process `p` and using `p.Start()` if it's not started yet, instead of passing both `psi` and `process`?

cpitclaudel (changes requested):

Source/DafnyCore/Compilers/SinglePassCompiler.cs
davidcok: It is what was in the code before -- do you have a suggestion?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
davidcok: Put in ??= -- if that is not what you actually meant, comment again.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
davidcok: And it is what current MS documentation says for Start() and WaitForExit()

Source/DafnyCore/Compilers/Compiler-java.cs
davidcok: Remove here and elsewhere

Source/DafnyCore/Compilers/SinglePassCompiler.cs
cpitclaudel: I was suggesting to pass just `process` and omit psi; wouldn't that work?  You can access the startinfo from the `process.StartInfo` property: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.startinfo?view=net-6.0

Source/DafnyCore/Compilers/SinglePassCompiler.cs
cpitclaudel: Thanks!  How odd.

Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/Helpers.java
cpitclaudel: ! Ok to merge this as is, but can we maybe open an issue to debate this?  @robin-aws , was it on purpose that `expect` failure are printed to `stdout`?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
davidcok: API refactored and simplified.

cpitclaudel (approved):
LGTM!  For future reference, you can create a process without starting it (https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process?view=net-6.0); maybe we could use that here to make the API even cleaner, but it's not very important.

PR #2880, opened 2022-10-10
Two more structured proof obligation descriptions

atomb:
This adds two more structured proof obligation descriptions and moves entirely away from `ErrorData` (which was removed in a boogie-org/boogie#618), allowing Dafny to work with the most recent Boogie code. These changes should have been included in #1915.

### DIFF
--- a/Source/DafnyCore/Verifier/Translator.cs
+++ b/Source/DafnyCore/Verifier/Translator.cs
@@ -7214,28 +7214,22 @@ namespace Microsoft.Dafny {
       }
     }
 
-    // TODO: update to include structured description
     Bpl.Ensures Ensures(IToken tok, bool free, Bpl.Expr condition, string errorMessage, string comment) {
       Contract.Requires(tok != null);
       Contract.Requires(condition != null);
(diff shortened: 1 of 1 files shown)


MikaelMayer (approved):
Straightforward ! However, I could never make the "errorMessage" attribute to work on assertions. Could you?

atomb:
> Straightforward ! However, I could never make the "errorMessage" attribute to work on assertions. Could you?

It looks as though the attribute that Dafny currently recognizes is `{:error "message"}`. Does that work for you? I just tried it successfully. (I don't think that changed at all as part of this refactoring.)

MikaelMayer:
> > Straightforward ! However, I could never make the "errorMessage" attribute to work on assertions. Could you?
> 
> It looks as though the attribute that Dafny currently recognizes is `{:error "message"}`. Does that work for you? I just tried it successfully. (I don't think that changed at all as part of this refactoring.)

That's probably because I was trying to write `assert {:message}` instead of `assert {:error}`. None of these attributes are documented but they would be extremely useful to design libraries, so that failing preconditions could yield to useful error messages. @davidcok  did you know this attribute.

PR #2881, opened 2022-10-11
Fixes #1174: Improved error message

### LINKED ISSUE #1174: Regression in compilation error messages (was: Dafny CLI compiler gives worse error details than VSCode's Language Server)
Dafny 3.0.0.30204 (commit d25ae2577b5d38b3b4df90a7e8062d3f6fbfd2a7)
dotnet 5.0.104
macOS 11.2.3
VSCode 1.55.1 with Dafny extension 1.4.0

I'm compiling the following bad example program with the CLI compiler. 
```
function fm1(): bool
{
    (true)
}

function method fm2():bool
{
    var v := fm1();
    (v)
}
```

Compilation fails as expected. The problem is that it emits an unhelpful/misleading error text, attributing it to a point in the source code where it makes no sense.
In contrast, VSCode shows a great message at the correct point in the code. See screenshot with both the CLI result and the IDE message.
Screen Shot 2021-04-09 at 1 28 05 pm

Is there a way to make the CLI compiler emit that same level of information? I haven't been able to find a flag to do so. If there is no such a way, please consider this a Feature Request.

davidcok:
Fixes #1174 


Source/DafnyCore/Util.cs
```diff
@@ -845,7 +845,8 @@ public class ExpressionTester {
 
       if (expr is IdentifierExpr expression) {
         if (expression.Var != null && expression.Var.IsGhost) {
-          reporter?.Error(MessageSource.Resolver, expression, "a ghost variable is allowed only in specification contexts");
+          reporter?.Error(MessageSource.Resolver, expression,
+              $"ghost variables such as {expression.Name} are allowed only in specification contexts. {expression.Name} was inferred to be ghost based on its use or declaration.");
```
cpitclaudel, line 849: Is ghostness actually inferred based on use?  I thought it was only from the body of the variable, or from it being marked "ghost" explicitly.

davidcok: By use here I did mean its initialization, but I was not sure if more of its use was considered.
Let me know if you think a wording change is needed (it will require updating a fair number of tests)

cpitclaudel: Yes, I think it would be better to not mention the use.  I would just remove the second half of the message.

cpitclaudel: Btw, for these sort of edits, I would:

1. Use `git format-patch HEAD~1` to produce a patch
2. Open the patch (called `0001-….patch`) in a text editor
3. Use search and replace, with a regexp, to apply the change to all files
4. Use `git am` to apply the patch

This way there's no need to edit all the files by hand.
Alternatively, `runTests.py` can regenerate these files automatically for you.

davidcok: I reworded, but the point of the issue was to be more explanatory.
sed did this simple change quite readily

keyboardDrummer (approved):

PR #2895, opened 2022-10-14
Fix: Add a special case for bool in the C++ backend

### LINKED ISSUE #2773: Dafny's C++ compiler prints `true` as `"1"`, not as `"true"`
Input:

```
method M() {
  print true;
}
```

Output:

```
1
```

parno:
Prints "true" and "false" instead of "1" and "0" to better match the other backends.

Fixes #2773

@cpitclaudel Given this, do you want to revert be66cdf522f267db16b? 


Source/DafnyRuntime/DafnyRuntime.h
```diff
@@ -80,6 +80,16 @@ void dafny_print(T x) {
...
 }
 
+// Special-case bool so that the C++ output matches that of other backends
+template<>
+void dafny_print<bool>(bool x) {
+  if (x) {
+    std::cout << "true";
+  } else {
+    std::cout << "false";
+  }
```
cpitclaudel, line 90:
```suggestion
  std::cout << (x ? "true" : "false");
```
?

cpitclaudel (approved):
Sweet!
Left a comment, but nothing blocking.

Source/DafnyRuntime/DafnyRuntime.h
davidcok: Or the "std' way: std::cout << std::boolalpha << x

cpitclaudel (approved):

parno:
Is there any advice on what do do when the "check-deep-tests" step fails?  I'm not sure what problem it's running into, but one of the files it appears to mention is `D:/a/dafny/dafny/dafny/Source/IntegrationTests/bin/Debug/net6.0/TestFiles/LitTests/LitTest/git-issues\Output\git-issue-Main0.dfy.tmp`, which I assume corresponds to `Test/git-issues/git-issue-Main0.dfy`, which doesn't appear to make use of C++ backend, so I'm puzzled why this fix would have affected it.

cpitclaudel:
The problem isn't from this PR, so there's no fix needed on your side (sorry for the trouble!).  This is a general health test that block further merges if the nightly build fails: as soon as that build issue is fixed, all PRs will be unblocked.

Source/DafnyRuntime/DafnyRuntime.h
parno: Good point -- I'd forgotten about that option.  That one probably requires a comment for the uninitiated.  I don't feel strongly between the two versions.

PR #2902, opened 2022-10-18
Add a "breaking changes" category to the release notes

cpitclaudel:


robin-aws (approved):
LGTM, thanks!

Only reservation is the fact that we don't have any tests for this script. I wonder if we can add something to CI (probably `singletons`), just to ensure a dry run of preparing a release doesn't hit errors.

PR #2904, opened 2022-10-18
fix(java): unqualified type cast in pattern dtor

### LINKED ISSUE #2903: Java compiler doesn't qualify casts in pattern destructors
Given the following in `Main.dfy`:

```dafny
module A {
  datatype Wrapper = Wrapper(val: int)
}

module B {
  datatype Wrapper = Wrapper
}

module Main {
  import opened A
  import B

  method Main() {
    var Wrapper(x) := Wrapper(0);
    print x;
  }
}
```

Trying to compile and run this to Java gives a `cannot find symbol` error because the `Wrapper` type cast isn't qualified:

```
$ dafny run -t java Main.dfy

/<snip>/Main-java/Main_Compile/__default.java:15: error: reference to Wrapper is ambiguous
    java.math.BigInteger _0_x = ((Wrapper)_let_tmp_rhs0)._val;
                                  ^
  both class B_Compile.Wrapper in B_Compile and class A_Compile.Wrapper in A_Compile match
1 error
```

alex-chew:
Fixes #2903.

### DIFF
--- a/Test/git-issues/git-issue-2903.dfy
+++ b/Test/git-issues/git-issue-2903.dfy
@@ -1,0 +1,25 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:cs "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:js "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:go "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:java "%s" >> "%t"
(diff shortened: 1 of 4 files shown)


Test/git-issues/git-issue-2903.dfy
```diff
@@ -0,0 +1,25 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
```
robin-aws, line 1: Could you use `%testDafnyForEachCompiler` here? The only thing you'd have to change in the test file is to avoid any unbounded `int` references, but that should be easy.

cpitclaudel (dismissed):
Fix looks good to me, nice work!
+1 to robin's comment about testing all compilers at once

cpitclaudel:
Btw, would you mind adding a changelog entry?

Test/git-issues/git-issue-2903.dfy
alex-chew: From our chat offline - I ran into some errors here so it's not worth blocking on this; we'll need to convert a bunch of tests to do this later anyway. Resolving this for now.

robin-aws (approved):

PR #2910, opened 2022-10-21
Fix: Variable declarations in match cases, formals do not trigger errors anymore

### LINKED ISSUE #2905: Name resolution error
```dafny
datatype DT = DT {
  static method m(){}
}

datatype DD = AA | BB | CC
{
  method problem() {
    match this {
      case CC =>
      case _ =>
        var a := "value";
        DT.m();
    }
  }
}
```
This minimal code, if executed with Dafny, will trigger the following error:

```
file.dfy(11,12): Error:
 more than one declaration of variable name: a#1_0
1 name resolution errors detected in file.bpl

*** Encountered internal translation error - re-running Boogie to get better deb
ug information

file.bpl(3069,6): Error: more than
one declaration of variable name: a#1_0
1 name resolution errors detected in file.bpl
```

MikaelMayer:
This PR ensures that, in the case of branch duplication in resolution, exactly one branch reuses the same variables as the original branch (local variable declarations, formals, and bound variables in patterns).

That way, we fix #2905 and moreover fix the previous issue that hovering other variables in match cases was not displaying type information

### DIFF
--- a/Source/DafnyCore/Cloner.cs
+++ b/Source/DafnyCore/Cloner.cs
@@ -195,7 +195,7 @@ namespace Microsoft.Dafny {
       }
     }
 
-    public Formal CloneFormal(Formal formal) {
+    public virtual Formal CloneFormal(Formal formal) {
       Formal f = new Formal(Tok(formal.tok), formal.Name, CloneType(formal.Type), formal.InParam, formal.IsGhost,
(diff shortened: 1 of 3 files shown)


cpitclaudel (changes requested):
LGTM!  Can you add a test case for the code that used to raise an error?

PR #2913, opened 2022-10-21
Reduce verbosity of test logs

cpitclaudel:
Our CI logs are often over 5000 lines long; do we actually need all that output?  Most of it is `Copying … to …`

### DIFF
--- a/.github/workflows/integration-tests-reusable.yml
+++ b/.github/workflows/integration-tests-reusable.yml
@@ -114,7 +114,7 @@ jobs:
         XUNIT_SHARD_COUNT: ${{ inputs.num_shards }}
         DAFNY_RELEASE: ${{ github.workspace }}\unzippedRelease\dafny
       run: |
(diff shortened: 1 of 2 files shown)


MikaelMayer (commented):
If it's only up to me, I'd approve it immediately, but I prefer to have feedback from @robin-aws .

cpitclaudel:
> Does this affect whether logger.Debug statements are shown?

It shouldn't, since this affects only the integration tests

cpitclaudel:
> the intention was to be able to see individual test names but that doesn't seem to work now anyway, and I'm not sure when that stopped working.

Ah!  In that case we need `--logger`, I've updated the docs accordingly.

robin-aws (approved):
I feel like I should leave a long and rambling comment about how much I appreciate this change instead of "LGTM", just because it's about reducing verbosity and I'm a sucker for irony.

PR #2917, opened 2022-10-24
Chore: Make release notes for PR 2876 clearer

MikaelMayer:
Fixes https://github.com/dafny-lang/dafny/pull/2876#issuecomment-1287460272


cpitclaudel (approved):
Much better, thanks!
But merging "node" and the rest is a bit confusing, no?  Why not just remove `node` entirely?
(That's what is done in Python)

Otherwise, how can one use the value of the first argument?

PR #2926, opened 2022-10-25
fix: Handle UTF-16 escapes and invalid surrogate sequences in Python and Go

### LINKED ISSUE #1890: Feature request: member methods on type aliases
One obstacle to reusing existing library types in client code is that Dafny currently does not support defining member functions on type aliases.  Here is what I would like to be able to write:

```dafny
// Library
datatype Result<+T, +Q> = Success(t: T) | Failure(q: Q)

// App
datatype AppExn = ErrorA | ErrorB(s: string)

type AppResult<+T> = Result<T, AppExn> {
  predicate method isErrorA<T>() { // HERE: A member method
    this == Failure(ErrorA)
  }
}

// Client
function method APICall() : AppResult<int>
method t() {
  var i := APICall();
  if i.isErrorA() { print "A!\n"; } // HERE: Member syntax for i.isErrorA()
  else if i.Failure? { print "!!\n"; }
  else { print i.t, "\n"; }
}
```

Here is the working Dafny code that I have to write instead:

```dafny
// Library
datatype Result<+T, +Q> = Success(t: T) | Failure(q: Q)

// App
datatype AppExn = ErrorA | ErrorB(s: string)
type AppResult<+T> = Result<T, AppExn>

predicate method isErrorA<T>(this_: AppResult<T>) { // HERE: Not a member function
  this_ == Failure(ErrorA)
}

// Client
function method APICall() : AppResult<int>
method t() {
  var i := APICall();
  if isErrorA(i) { print "A!\n"; } // HERE: Regular function code
  else if i.Failure? { print "!!\n"; }
  else { print i.t, "\n"; }
}
```

The status quo makes it harder to reuse existing types from Dafny's standard library, because newly defined methods look "second-class" (they don't get member syntax)

Could we allow member methods on type aliases?

### LINKED ISSUE #1980: Dafny string literals with surrogate characters compiled to invalid Go strings
Dafny string literals are translated literally to Go literals, but `\uXXXX` in Dafny is interpreted as a UTF-16 code point, whereas Go will interpret the same escape sequences as Unicode scalar values, which excludes surrogates.

```dafny
method Foo() {
  // Mathematical Sans-Serif Bold Italic Small Psi - represented as a surrogate of U+1D7C1
  var s := "\uD835\uDFC1";
}
```

```
$ dafny /compile:3 /compileTarget:go src/Scratch.dfy 

# command-line-arguments
src/Scratch-go/src/Scratch.go:46:37: escape is invalid Unicode code point U+D835
src/Scratch-go/src/Scratch.go:46:43: escape is invalid Unicode code point U+DFC1
```

### LINKED ISSUE #2925: Invalid UTF-16 strings cause print to crash on some backends
```dafny
method Main() {
  print("\uDFFF");
}
```

Fails to compile on Go, crashes at runtime for Python.

robin-aws:
Fixes #1980. Fixes #2925.

Should fix the root cause of #2934, but I'm not claiming the complete fix, as attempting to include unescaped, non-ASCII  characters in a Dafny source file is revealing platform-specific headaches in the integration test runner I'd like to address in a separate PR.

Printing non-ASCII characters, especially invalid UTF-16 sequences, is still inconsistent across backends, but at least should not crash after this change. Again this is difficult to test effectively across platforms in our testing architecture.


Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -17,13 +17,24 @@ def __get__(self, instance, owner):
 def print(value):
     builtins.print(string_of(value), end="")
 
+# Dafny strings are currently sequences of UTF-16 code units.
+# To make a best effort attempt at printing the right characters we attempt to decode,
+# but have to allow for invalid sequences.
+def string_from_utf_16(utf_16_code_units):
+    return b''.join([ord(c).to_bytes(2, 'little') for c in utf_16_code_units]).decode("utf-16", errors = 'replace')
```
cpitclaudel, line 24: Two small things:
```suggestion
    return b''.join(ord(c).to_bytes(2, 'little') for c in utf_16_code_units).decode("utf-16-le", errors = 'replace')
```

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -17,13 +17,24 @@ def __get__(self, instance, owner):
...
 def string_of(value) -> str:
     if hasattr(value, '__dafnystr__'):
         return value.__dafnystr__()
     elif value is None:
         return "null"
     elif isinstance(value, bool):
         return "true" if value else "false"
+    elif isinstance(value, str):
+        # This is only for Dafny char values.
+        # Dafny strings are represented as Seq's of indivdual char values,
```
cpitclaudel, line 35:
```suggestion
        # Dafny strings are represented as Seq's of individual char values,
```

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -17,13 +17,24 @@ def __get__(self, instance, owner):
...
+        # and Seq defines __dafnystr__.
+        return string_from_utf_16(value)
```
cpitclaudel, line 37: Do you need the full generality of this if it can only be the value of a Dafny char?

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -597,6 +597,8 @@ func (seq Seq) UniqueElements() Set {
 func (seq Seq) String() string {
   if seq.isString {
     s := ""
+    // Note this doesn't produce the right string in UTF-8,
+    // since it converts surrogates independently.
```
cpitclaudel, line 601: Isn't that a bug?

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -2061,12 +2061,36 @@ private class SpecialNativeType : UserDefinedType {
         wr.Write((bool)e.Value ? "true" : "false");
       } else if (e is CharLiteralExpr) {
         var v = (string)e.Value;
-        wr.Write("_dafny.Char('{0}')", TranslateEscapes(v, isChar: true));
+        wr.Write("_dafny.Char(");
+        // See comment on the StringLiteralExpr case below.
+        if (Util.Utf16Escape.IsMatch(v)) {
+          char c = Util.UnescapedCharacters(v, false).Single();
```
cpitclaudel, line 2067: Why is it always at most one?

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -2061,12 +2061,36 @@ private class SpecialNativeType : UserDefinedType {
         wr.Write((bool)e.Value ? "true" : "false");
       } else if (e is CharLiteralExpr) {
         var v = (string)e.Value;
-        wr.Write("_dafny.Char('{0}')", TranslateEscapes(v, isChar: true));
```
cpitclaudel, line 2064: Can we keep the complexity in a separate method (as it was with TranslateEscapes)?  Maybe `TranslateCharLiteral`

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -2061,12 +2061,36 @@ private class SpecialNativeType : UserDefinedType {
...
         var str = (StringLiteralExpr)e;
-        wr.Write("_dafny.SeqOfString(");
-        TrStringLiteral(str, wr);
-        wr.Write(")");
+        // It may not be possible to translate \u escapes into a valid Go string,
+        // since Go string literals have to be encodable in UTF-8,
+        // but Dafny allows invalid sequences of surrogate characters.
+        // So if any are present, just emit a sequence of the direct UTF-16 code units instead.
+        var s = (string)str.Value;
+        if (!str.IsVerbatim && Util.Utf16Escape.IsMatch(s)) {
```
cpitclaudel, line 2080: What happens if the C# string contains unpaired surrogates (not `\u` escaped, just directly, unescaped, in the string?)  Is that impossible thanks to the way we parse?

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -2061,12 +2061,36 @@ private class SpecialNativeType : UserDefinedType {
...
+          wr.Write("_dafny.SeqOfChars(");
+          var comma = "";
+          foreach (var c in Util.UnescapedCharacters(s, str.IsVerbatim)) {
+            wr.Write(comma);
+            wr.Write($"{(int)c}");
+            comma = ", ";
```
cpitclaudel, line 2086: I thought we had an API to join, but maybe not?

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -2061,12 +2061,36 @@ private class SpecialNativeType : UserDefinedType {
...
+          }
+          wr.Write(")");
```
cpitclaudel, line 2088: Can we move this to TrStringLiteral?

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -2097,6 +2121,8 @@ private class SpecialNativeType : UserDefinedType {
       }
     }
 
+
+
     protected override void EmitStringLiteral(string str, bool isVerbatim, ConcreteSyntaxTree wr) {
```
cpitclaudel, line 2126:
```suggestion
    protected override void EmitStringLiteral(string str, bool isVerbatim, ConcreteSyntaxTree wr) {
```

Source/DafnyCore/Util.cs
```diff
@@ -195,6 +196,9 @@ public static class Util {
       UnescapedCharacters(s, isVerbatimString).Iter(ch => sb.Append(ch));
       return sb.ToString();
     }
+
+    public static readonly Regex Utf16Escape = new Regex(@"(?<!\\)\\u([0-9a-fA-F]{4})");
```
cpitclaudel, line 200: This doesn't look right: how about `\\\uAAAA`?  It won't match, right?

cpitclaudel (changes requested):
Good catch and nice fixes.

Source/DafnyRuntime/DafnyRuntime.py
robin-aws: Agreed on `-le`, but I don't think we need the `[...]`. As is, `ord(c).to_bytes(2, 'little') for c in utf_16_code_units` is a generator instead of a list, and `bytes.join` only needs an `Iterable`.

Source/DafnyCore/Compilers/Compiler-go.cs
robin-aws: Yeah I did too. There's `Util.Comma` but it only produces strings rather than working with `ConcreteSyntaxTree`s.

Source/DafnyRuntime/DafnyRuntime.py
robin-aws: I still need most of it at least, since I need to convert a single character to a `bytes` of length 2 to decode from UTF-16. Given that a Python "character" is just a string with length 1, there's no real overhead to using the same function as for strings.

Source/DafnyCore/Compilers/Compiler-go.cs
robin-aws: It SHOULD be impossible because Dafny source files have to be UTF-8 encoded, and the scanner is decoding to `int` code points: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Coco/Scanner.frame#L210

That code doesn't look particularly robust in the face of invalid UTF-8 sequences mind you, but it doesn't look like it's possible to produce surrogate values.

There's also the fact that the generated scanner code includes a straight cast from `int` to `char`, which could truncate a code point to `0xFFFF`. I've got a fix for that in the unicode char branch (and I should probably cut an issue for it right away) but that still won't create surrogates at least.

Source/DafnyCore/Compilers/Compiler-go.cs
robin-aws: Nope, because we have to emit `_dafny.SeqOfChars` instead of `_dafny.SeqOfString`. I'm adding more meat on this to the comment.

Source/DafnyRuntime/DafnyRuntime.go
robin-aws: Definitely, just one I didn't want to bring in scope - I wasn't confident it was reasonable to depend on the necessary Go package to decode UTF-16. But I'll at least cut an issue (plus C++ is an even bigger mess since it uses the 8-bit `char` type).

Source/DafnyCore/Compilers/Compiler-go.cs
robin-aws: Yeah I like that, and even better to use `TrStringLiteral` to encapsulate the even bigger string case/

Source/DafnyCore/Compilers/Compiler-go.cs
robin-aws: Actually yes we can - I thought you meant to `EmitStringLiteral` at first. :)

Source/DafnyCore/Compilers/Compiler-go.cs
robin-aws: The parser should be guaranteeing it's exactly one. We do something equivalent in the translator already: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs#L306-L309

Source/DafnyCore/Compilers/Compiler-go.cs
cpitclaudel: OK, sweet.  I'll let you judge whether it's worth a Contract.Assert here — it might save our bacon down the line?

Source/DafnyCore/Compilers/Compiler-go.cs
cpitclaudel: Sweet, thanks.  `Single` will raise an exception if there's one than one element, right?  So, nice safety check actually.

Source/DafnyRuntime/DafnyRuntime.go
cpitclaudel: +1, thanks

Source/DafnyRuntime/DafnyRuntime.py
cpitclaudel: +1, thanks

Source/DafnyRuntime/DafnyRuntime.py
cpitclaudel: I think you misread my suggestion and we're in violent agreement ^^ The `[...]` was in your code, and my suggestion was to remove it :)

Source/DafnyRuntime/DafnyRuntime.go
robin-aws: Yup. And unfortunately the more I test the more I find ugly inconsistencies on corner cases. :(

Source/DafnyCore/Compilers/Compiler-go.cs
robin-aws: Actually there's nothing to assert here: I'm specifically NOT assuming surrogates are used correctly here. The only concern is accurately translating escape sequences to target language escape sequences (or something else equivalent). I'm not aware of any bugs in translating arbitrary `char` values in the string at least.

Source/DafnyCore/Util.cs
robin-aws: Note that both of those other bogus regex patterns will be removed entirely in the fix to https://github.com/dafny-lang/dafny/issues/2928, since they are trying to do entirely the wrong thing anyway.

Source/DafnyCore/Util.cs
```diff
@@ -195,54 +196,82 @@ public static class Util {
...
+        }
+      }
+    }
+
+    /// <summary>
+    /// Enumerates the sequence of regular characters and escape sequences in the given string.
+    /// For example, "ab\tuv\u12345" may be broken up as ["a", "b", "\t", "u", "v", "\u1234", "5"].
+    /// Consecutive non-escaped characters may or may not be enumerated as a single string.
+    /// </summary>
+    public static IEnumerable<string> Escapes(string p, bool isVerbatimString) {
```
robin-aws, line 243: Better name for this method?

Test/dafny0/Strings.dfy
```diff
@@ -31,12 +36,35 @@ method Main()
...
+  expect |z?| > 2;
+  expect z?[0..2] == ['\ude0e', '\ud83d'];
+  print "Weird string Z: ", z?, "\n";
+
+  var c?, d? := WeirdChars();
+  print "These characters are quite confused: ", c?, ' ', d?, "\n";
+
+  // Ensuring we're precise enough about identifying \u escapes
+  print "I'm afraid you'll find escape quite impossible, \\u007", "\n";
+  print "Luckily I have this nifty gadget from my good friend, \\\u0051", "\n";
```
robin-aws, line 67: Callback to my favourite test case ever: https://github.com/dafny-lang/dafny/blob/master/Test/expectations/Expect.dfy#L19-L22 :)

Source/DafnyRuntime/DafnyRuntime.go
robin-aws: https://github.com/dafny-lang/dafny/issues/2929

Source/DafnyCore/Util.cs
cpitclaudel: TokenizeEscapedString?

cpitclaudel (dismissed):
Looks pretty good!

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -2276,7 +2276,7 @@ private class GenericArrayElementLvalue : ILvalue {
         files.Add($"\"{Path.GetFullPath(file)}\"");
       }
       var classpath = GetClassPath(targetFilename);
-      var psi = new ProcessStartInfo("javac", string.Join(" ", files)) {
+      var psi = new ProcessStartInfo("javac", "-encoding UTF8 " + string.Join(" ", files)) {
```
cpitclaudel, line 2279: This is fine for now but we should check what encoding we use for writing.

cpitclaudel (approved):

PR #2933, opened 2022-10-27
Add an equivalent of the old -definiteAssignment option to the new CL…

### LINKED ISSUE #2915: How to call `-definiteAssignment:3` with `dafny build`
While `-help` gives the old commands
how do I reference these commands with the new command syntax?

I'm guessing that I can just append them and see what happens,
but what would `dafny build -compile:0` mean?

I would like the various options to appear under each command
and then have a clear way on how to compose them.

`definiteAssignment` would appear in the `dafny verify --help`
and `dafny build --help` would explain that any `verify` options would "pass through" if `--no-verify 0`.
Finally, in the case of `dafny build --no-verify 1` any verification options MUST fail the command :)

keyboardDrummer:
### Update

Add an equivalent of the existing `-definiteAssignment` option to the new CLI UI

Fixes #2915

### Testing

Updated two tests to use the new options

### DIFF
--- a/Source/DafnyCore/Compilers/SinglePassCompiler.cs
+++ b/Source/DafnyCore/Compilers/SinglePassCompiler.cs
@@ -1335,7 +1335,7 @@ namespace Microsoft.Dafny.Compilers {
           } else if (d is IteratorDecl) {
             var iter = (IteratorDecl)d;
             if (DafnyOptions.O.ForbidNondeterminism && iter.Outs.Count > 0) {
-              Error(iter.tok, "since yield parameters are initialized arbitrarily, iterators are forbidden by /definiteAssignment:3 option", wr);
+              Error(iter.tok, "since yield parameters are initialized arbitrarily, iterators are forbidden by enforce-determinism option", wr);
             }
 
             var wIter = CreateIterator(iter, wr);
@@ -2928,7 +2928,7 @@ namespace Microsoft.Dafny.Compilers {
               var rhs = s.Rhss[i];
               if (rhs is HavocRhs) {
                 if (DafnyOptions.O.ForbidNondeterminism) {
-                  Error(rhs.Tok, "nondeterministic assignment forbidden by /definiteAssignment:3 option", wr);
+                  Error(rhs.Tok, "nondeterministic assignment forbidden by enforce-determinism option", wr);
                 }
               } else {
                 lhss.Add(lhs);
@@ -2955,7 +2955,7 @@ namespace Microsoft.Dafny.Compilers {
         Contract.Assert(s.Lhs is not SeqSelectExpr expr || expr.SelectOne);  // multi-element array assignments are not allowed
         if (s.Rhs is HavocRhs) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s.Rhs.Tok, "nondeterministic assignment forbidden by /definiteAssignment:3 option", wr);
+            Error(s.Rhs.Tok, "nondeterministic assignment forbidden by enforce-determinism option", wr);
           }
         } else if (s.Rhs is ExprRhs eRhs && eRhs.Expr.Resolved is FunctionCallExpr fce && IsTailRecursiveByMethodCall(fce)) {
           TrTailCallStmt(s.Tok, fce.Function.ByMethodDecl, fce.Receiver, fce.Args, null, wr);
@@ -2969,7 +2969,7 @@ namespace Microsoft.Dafny.Compilers {
       } else if (stmt is AssignSuchThatStmt) {
         var s = (AssignSuchThatStmt)stmt;
         if (DafnyOptions.O.ForbidNondeterminism) {
-          Error(s.Tok, "assign-such-that statement forbidden by /definiteAssignment:3 option", wr);
+          Error(s.Tok, "assign-such-that statement forbidden by enforce-determinism option", wr);
         }
         if (s.AssumeToken != null) {
           // Note, a non-ghost AssignSuchThatStmt may contain an assume
@@ -3016,7 +3016,7 @@ namespace Microsoft.Dafny.Compilers {
         IfStmt s = (IfStmt)stmt;
         if (s.Guard == null) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s.Tok, "nondeterministic if statement forbidden by /definiteAssignment:3 option", wr);
+            Error(s.Tok, "nondeterministic if statement forbidden by enforce-determinism option", wr);
           }
           // we can compile the branch of our choice
           ConcreteSyntaxTree guardWriter;
@@ -3041,7 +3041,7 @@ namespace Microsoft.Dafny.Compilers {
           }
         } else {
           if (s.IsBindingGuard && DafnyOptions.O.ForbidNondeterminism) {
-            Error(s.Tok, "binding if statement forbidden by /definiteAssignment:3 option", wr);
+            Error(s.Tok, "binding if statement forbidden by enforce-determinism option", wr);
           }
           ConcreteSyntaxTree guardWriter;
           var coverageForElse = Coverage.IsRecording && !(s.Els is IfStmt);
@@ -3070,7 +3070,7 @@ namespace Microsoft.Dafny.Compilers {
       } else if (stmt is AlternativeStmt) {
         var s = (AlternativeStmt)stmt;
         if (DafnyOptions.O.ForbidNondeterminism && 2 <= s.Alternatives.Count) {
-          Error(s.Tok, "case-based if statement forbidden by /definiteAssignment:3 option", wr);
+          Error(s.Tok, "case-based if statement forbidden by enforce-determinism option", wr);
         }
         foreach (var alternative in s.Alternatives) {
           ConcreteSyntaxTree guardWriter;
@@ -3092,7 +3092,7 @@ namespace Microsoft.Dafny.Compilers {
         }
         if (s.Guard == null) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s.Tok, "nondeterministic loop forbidden by /definiteAssignment:3 option", wr);
+            Error(s.Tok, "nondeterministic loop forbidden by enforce-determinism option", wr);
           }
           // This loop is allowed to stop iterating at any time. We choose to never iterate, but we still
           // emit a loop structure. The structure "while (false) { }" comes to mind, but that results in
@@ -3109,7 +3109,7 @@ namespace Microsoft.Dafny.Compilers {
 
       } else if (stmt is AlternativeLoopStmt loopStmt) {
         if (DafnyOptions.O.ForbidNondeterminism) {
-          Error(loopStmt.Tok, "case-based loop forbidden by /definiteAssignment:3 option", wr);
+          Error(loopStmt.Tok, "case-based loop forbidden by enforce-determinism option", wr);
         }
         if (loopStmt.Alternatives.Count != 0) {
           ConcreteSyntaxTree whileGuardWriter;
@@ -3157,7 +3157,7 @@ namespace Microsoft.Dafny.Compilers {
         var s0 = (AssignStmt)s.S0;
         if (s0.Rhs is HavocRhs) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s0.Rhs.Tok, "nondeterministic assignment forbidden by /definiteAssignment:3 option", wr);
+            Error(s0.Rhs.Tok, "nondeterministic assignment forbidden by enforce-determinism option", wr);
           }
           // The forall statement says to havoc a bunch of things.  This can be efficiently compiled
           // into doing nothing.
@@ -3320,7 +3320,7 @@ namespace Microsoft.Dafny.Compilers {
         if (s.Body != null) {
           TrStmt(s.Body, wr);
         } else if (DafnyOptions.O.ForbidNondeterminism) {
-          Error(s.Tok, "modify statement without a body forbidden by /definiteAssignment:3 option", wr);
+          Error(s.Tok, "modify statement without a body forbidden by enforce-determinism option", wr);
         }
       } else if (stmt is TryRecoverStatement h) {
         EmitHaltRecoveryStmt(h.TryBody, h.HaltMessageVar.CompileName, h.RecoverBody, wr);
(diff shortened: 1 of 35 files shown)


Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1335,7 +1335,7 @@ public enum ResolvedUnaryOp { BoolNot, BitwiseNot, Cardinality }
           } else if (d is IteratorDecl) {
             var iter = (IteratorDecl)d;
             if (DafnyOptions.O.ForbidNondeterminism && iter.Outs.Count > 0) {
-              Error(iter.tok, "since yield parameters are initialized arbitrarily, iterators are forbidden by /definiteAssignment:3 option", wr);
+              Error(iter.tok, "since yield parameters are initialized arbitrarily, iterators are forbidden by enforce-determinism option", wr);
```
davidcok, line 1338:
```suggestion
              Error(iter.tok, "since yield parameters are initialized arbitrarily, iterators are forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -2928,7 +2928,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
               var rhs = s.Rhss[i];
               if (rhs is HavocRhs) {
                 if (DafnyOptions.O.ForbidNondeterminism) {
-                  Error(rhs.Tok, "nondeterministic assignment forbidden by /definiteAssignment:3 option", wr);
+                  Error(rhs.Tok, "nondeterministic assignment forbidden by enforce-determinism option", wr);
```
davidcok, line 2931:
```suggestion
                  Error(rhs.Tok, "nondeterministic assignment forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -2955,7 +2955,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
         Contract.Assert(s.Lhs is not SeqSelectExpr expr || expr.SelectOne);  // multi-element array assignments are not allowed
         if (s.Rhs is HavocRhs) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s.Rhs.Tok, "nondeterministic assignment forbidden by /definiteAssignment:3 option", wr);
+            Error(s.Rhs.Tok, "nondeterministic assignment forbidden by enforce-determinism option", wr);
```
davidcok, line 2958:
```suggestion
            Error(s.Rhs.Tok, "nondeterministic assignment forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -2969,7 +2969,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
       } else if (stmt is AssignSuchThatStmt) {
         var s = (AssignSuchThatStmt)stmt;
         if (DafnyOptions.O.ForbidNondeterminism) {
-          Error(s.Tok, "assign-such-that statement forbidden by /definiteAssignment:3 option", wr);
+          Error(s.Tok, "assign-such-that statement forbidden by enforce-determinism option", wr);
```
davidcok, line 2972:
```suggestion
          Error(s.Tok, "assign-such-that statement forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -3016,7 +3016,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
         IfStmt s = (IfStmt)stmt;
         if (s.Guard == null) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s.Tok, "nondeterministic if statement forbidden by /definiteAssignment:3 option", wr);
+            Error(s.Tok, "nondeterministic if statement forbidden by enforce-determinism option", wr);
```
davidcok, line 3019:
```suggestion
            Error(s.Tok, "nondeterministic if statement forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -3041,7 +3041,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
           }
         } else {
           if (s.IsBindingGuard && DafnyOptions.O.ForbidNondeterminism) {
-            Error(s.Tok, "binding if statement forbidden by /definiteAssignment:3 option", wr);
+            Error(s.Tok, "binding if statement forbidden by enforce-determinism option", wr);
```
davidcok, line 3044:
```suggestion
            Error(s.Tok, "binding if statement forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -3070,7 +3070,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
       } else if (stmt is AlternativeStmt) {
         var s = (AlternativeStmt)stmt;
         if (DafnyOptions.O.ForbidNondeterminism && 2 <= s.Alternatives.Count) {
-          Error(s.Tok, "case-based if statement forbidden by /definiteAssignment:3 option", wr);
+          Error(s.Tok, "case-based if statement forbidden by enforce-determinism option", wr);
```
davidcok, line 3073:
```suggestion
          Error(s.Tok, "case-based if statement forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -3092,7 +3092,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
         }
         if (s.Guard == null) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s.Tok, "nondeterministic loop forbidden by /definiteAssignment:3 option", wr);
+            Error(s.Tok, "nondeterministic loop forbidden by enforce-determinism option", wr);
```
davidcok, line 3095:
```suggestion
            Error(s.Tok, "nondeterministic loop forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -3109,7 +3109,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
 
       } else if (stmt is AlternativeLoopStmt loopStmt) {
         if (DafnyOptions.O.ForbidNondeterminism) {
-          Error(loopStmt.Tok, "case-based loop forbidden by /definiteAssignment:3 option", wr);
+          Error(loopStmt.Tok, "case-based loop forbidden by enforce-determinism option", wr);
```
davidcok, line 3112:
```suggestion
          Error(loopStmt.Tok, "case-based loop forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -3157,7 +3157,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
         var s0 = (AssignStmt)s.S0;
         if (s0.Rhs is HavocRhs) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s0.Rhs.Tok, "nondeterministic assignment forbidden by /definiteAssignment:3 option", wr);
+            Error(s0.Rhs.Tok, "nondeterministic assignment forbidden by enforce-determinism option", wr);
```
davidcok, line 3160:
```suggestion
            Error(s0.Rhs.Tok, "nondeterministic assignment forbidden by the --enforce-determinism option", wr);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -3320,7 +3320,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
         if (s.Body != null) {
           TrStmt(s.Body, wr);
         } else if (DafnyOptions.O.ForbidNondeterminism) {
-          Error(s.Tok, "modify statement without a body forbidden by /definiteAssignment:3 option", wr);
+          Error(s.Tok, "modify statement without a body forbidden by enforce-determinism option", wr);
```
davidcok, line 3323:
```suggestion
          Error(s.Tok, "modify statement without a body forbidden by the --enforce-determinism option", wr);
```

Test/dafny0/ForbidNondeterminismCompile.dfy.expect
```diff
@@ -1,15 +1,15 @@
...
-ForbidNondeterminismCompile.dfy(20,2): Error: nondeterministic if statement forbidden by /definiteAssignment:3 option
-ForbidNondeterminismCompile.dfy(23,2): Error: case-based if statement forbidden by /definiteAssignment:3 option
-ForbidNondeterminismCompile.dfy(32,2): Error: binding if statement forbidden by /definiteAssignment:3 option
-ForbidNondeterminismCompile.dfy(35,2): Error: nondeterministic loop forbidden by /definiteAssignment:3 option
-ForbidNondeterminismCompile.dfy(40,2): Error: case-based loop forbidden by /definiteAssignment:3 option
-ForbidNondeterminismCompile.dfy(48,12): Error: nondeterministic assignment forbidden by /definiteAssignment:3 option
-ForbidNondeterminismCompile.dfy(50,2): Error: modify statement without a body forbidden by /definiteAssignment:3 option
-ForbidNondeterminismCompile.dfy(63,16): Error: nondeterministic assignment forbidden by /definiteAssignment:3 option
+Dafny program verifier finished with 3 verified, 0 errors
+ForbidNondeterminismCompile.dfy(71,9): Error: since yield parameters are initialized arbitrarily, iterators are forbidden by enforce-determinism option
```
davidcok, line 4:
```suggestion
ForbidNondeterminismCompile.dfy(71,9): Error: since yield parameters are initialized arbitrarily, iterators are forbidden by the --enforce-determinism option
```

davidcok: And similarly for the succeeding lines

davidcok: Also corresponding changes in the UsesrGuide are needed

davidcok (changes requested):

Source/DafnyCore/Options/EnforceDeterminismOption.cs
```diff
@@ -0,0 +1,17 @@
+namespace Microsoft.Dafny;
+
+public class EnforceDeterminismOption : BooleanOption {
```
davidcok, line 3: I think you will need to run `make options` in docs/DafnyRef and then commit/push the new Options.txt file, or else the CI building of DafnyRef.pdf will fail.

davidcok (changes requested):

Source/DafnyCore/Options/EnforceDeterminismOption.cs
davidcok: It did not fail, which must mean that the tests are not checking these new options -- so that should be fixed.

davidcok (changes requested):

Source/DafnyCore/Options/EnforceDeterminismOption.cs
davidcok: And to be honest, I really don't like this split between old and new help. If I were a newish user it would be very confusing.

keyboardDrummer: The `Options.txt` file contains the old help, which is not changed by this PR, so there's no need to update it.

davidcok (approved):

seebees:
This is great, but I have 3 questions.

1. If I did not know, how would I discover that `/definiteAssignment:3 ==> build --enforce-determinism`?
2. What is `--strict-definite-assignment`? Is that the same?
3. I assume that `--enforce-determinism` is not yet in DafnyRef because we are still working things out?

robin-aws:
These options don't show up in the reference manual yet because we're only including the old `/help` contents and not the per-command `--help` contents. I think we need to fix that with some urgency though. Folks are used to being able to search for options and if they aren't there, they likely won't notice wherever we say "only the old CLI help contents are below, for new CLI help do this..."

@keyboardDrummer: it feels like `--enforce-determinism` implies `--strict-definite-assignment` but that's not clear from the documentation. What happens if I specify `--enforce-determinism:true` but `--strict-definite-assignment:false`?

keyboardDrummer:
> What happens if I specify --enforce-determinism:true but --strict-definite-assignment:false?

Then you can't use non-deterministic statements, but you can use auto-initializers of types that have them to initialise variables without assigning to them.

robin-aws:
Ah that's good if they are truly independent options. How do the four values of `/definiteAssignment` map to the four possible combinations then? I tried figuring that out but it feels like they don't actually line up.

keyboardDrummer:
> Ah that's good if they are truly independent options. How do the four values of `/definiteAssignment` map to the four possible combinations then? I tried figuring that out but it feels like they don't actually line up.

definiteAssignment  => enforce-determinism, strict-definite-assignment
0 => unmapped
1 => false, false
2 => true, false
3 => true, true

robin-aws:
Ah that's what I suspected, meaning `true, false` is a newly available option.

Can you follow up with adding that information to the reference manual? Your example above would be great to include somewhere!

keyboardDrummer:
> Ah that's what I suspected, meaning `true, false` is a newly available option.
> 
> Can you follow up with adding that information to the reference manual? Your example above would be great to include somewhere!

Do you feel those two options are logically related? Seems strange to compare two unrelated options in the reference manual.

robin-aws:
It's fine to document them independently for the most part, but we should state somewhere how `/definiteAssignment` translates to the new CLI options. Plus we should also mention `--enforce-determinism` at some point in the main body of the reference manual, and not just in the embedded `--help` text (and adding that can be a separate task).

keyboardDrummer:
> we should state somewhere how /definiteAssignment translates to the new CLI options.

Is there a good spot for that? I don't see a better one than the release notes: https://github.com/dafny-lang/dafny/pull/2948

PR #2938, opened 2022-10-27
fix: Force UTF-8 encoding in C# lit runner

cpitclaudel:


Source/XUnitExtensions/Lit/ShellLitCommand.cs
```diff
@@ -33,13 +33,68 @@ public class ShellLitCommand : ILitCommand {
...
+      // With Acute) should output a single byte, #x82.  But if the codepage is
+      // 1252 (Latin 1), then that same program should output the single
+      // byte `#xE9`.  Finally, if the codepage is 65001 (UTF8), then that same
+      // program should output `#xC3 #xA9`.
+      //
+      // Thankfully, the C# runtime takes care of this for us.  It uses
+      // `GetConsoleCP` to determine what encoding to use, then uses that when
+      // `Console.WriteLine` is called (and replaces unsupported characters by
+      // the fallback character `?`).  Here is a concrete demo, with character
+      // U+20AC (€) (which is supported by codepage 85001 but not 850):
```
robin-aws, line 62: 85001 -> 65001 (not sure why the suggestion button is gone)

Source/XUnitExtensions/Lit/ShellLitCommand.cs
```diff
@@ -33,13 +33,68 @@ public class ShellLitCommand : ILitCommand {
...
+      // output, as is done below.
+      //
+      // … but this is not enough, because dotnet run (or XUnit) also resets the
+      // codepage.  That's a bug there, and in the meantime the best hack we
+      // have is to override that codepage by setting `OutputEncoding`.
+
+      process.StartInfo.CreateNoWindow = false;
+      Console.OutputEncoding = System.Text.Encoding.UTF8;
+
+      // Finally, for Java + Ubuntu, we make sure to forward LANG:
```
robin-aws, line 95: Nit, but we're setting LANG rather than forwarding it

robin-aws (commented):
LOVE the huge comment immortalizing your immense debugging pain

Source/XUnitExtensions/Lit/ShellLitCommand.cs
cpitclaudel, line 62:
```suggestion
      // U+20AC (€) (which is supported by codepage 65001 but not 850):
```

Source/XUnitExtensions/Lit/ShellLitCommand.cs
cpitclaudel:
```suggestion
      // Finally, for Java + Ubuntu, we make sure to set LANG:
```

robin-aws (approved):

PR #2948, opened 2022-10-31
Add definiteAssignment option mapping to release notes

keyboardDrummer:
Fixes #


cpitclaudel (changes requested):
This doc change is fine, but I think we should hold off on it before we review what `--enforce-determinism` should really do; I've created a new issue at https://github.com/dafny-lang/dafny/issues/2959

PR #2952, opened 2022-10-31
feat: Support extern names containing periods for Go

### LINKED ISSUE #2045: Modules with extern names containing periods produce malformed Go code
```dafny
module {:extern "A.B"} A.B {}
```

```
$ dafny src/Scratch.dfy /compileTarget:go

Additional target code written to Scratch-go/src/A_Compile/A_Compile.go
src/Scratch-go/src/Scratch.go:7:4: expected 'STRING', found '.'
```

`Scratch.go` contains:

```go
// Dafny program Scratch.dfy compiled into Go
package Scratch

import (
  _dafny "dafny"
  _System "System_"
  A.B "A.B"
  A_Compile "A_Compile"
)
var _ _dafny.Dummy__
var _ _System.Dummy__
var _ A_Compile.Dummy__
```

Arguably this isn't a bug, because this is literally the symbol the user asked for. The intent is just to disable the name mangling that normally adds `_Compile`, though. The given external name works for other target languages, and there doesn't seem to be a way to preserve the original name that does work for Go.

robin-aws:
Fixes #2045. WIP.


cpitclaudel:
> even though we are using that in files like  …

Don't read too much in this one :) I just put the same annotation everywhere to not have to remember which is which :)

cpitclaudel (approved):
Looks good to me; @fabiomadge might want to have a look at the Python impl.

Test/comp/Extern5Nested.py
```diff
@@ -0,0 +1,13 @@
...
+"""
+
+import sys, _dafny
+
+assert "Nested.Library" == __name__
+Nested_Library = sys.modules[__name__]
+
+class default__:
+    def Foo():
+        _dafny.print(_dafny.Seq("Nested.Library.Foo\n"))
```
fabiomadge, line 13:
```suggestion
        print("Nested.Library.Foo\n")

```

Test/comp/Extern5Nested.py
```diff
@@ -0,0 +1,13 @@
+"""
+The Python compiler only supports {:extern} code on a module level, so the
+entire module must be supplied.
+"""
+
+import sys, _dafny
+
+assert "Nested.Library" == __name__
+Nested_Library = sys.modules[__name__]
```
fabiomadge, line 9:
```suggestion
Library = sys.modules[__name__]
```

Test/comp/Extern5Nested.py
```diff
@@ -0,0 +1,13 @@
+"""
+The Python compiler only supports {:extern} code on a module level, so the
+entire module must be supplied.
+"""
+
+import sys, _dafny
```
fabiomadge, line 6:
```suggestion
import sys
```

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1713,9 +1713,14 @@ private class ClassWriter : IClassWriter {
         outputWriter.WriteLine($"Unable to determine module name: {externFilename}");
         return false;
       }
+
+      var segments = moduleName.Split(".");
+      var baseName = segments.Last();
```
fabiomadge, line 1718:
```suggestion
```

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1713,9 +1713,14 @@ private class ClassWriter : IClassWriter {
...
       var mainDir = Path.GetDirectoryName(mainProgram);
       Contract.Assert(mainDir != null);
-      var tgtFilename = Path.Combine(mainDir, moduleName + ".py");
+      var tgtDir = segments[..^1].Aggregate(mainDir, Path.Combine);
+      var tgtFilename = Path.Combine(tgtDir, baseName + ".py");
```
fabiomadge, line 1722:
```suggestion
      var tgtFilename = Path.Combine(tgtDir, segments[^1] + ".py");
```

fabiomadge (commented):
The python compiler creates stub modules, even for `{:extern}` ones. If you're determined to make this change to how we treat `{:extern}` names, you should fix that too. In this case it's doing something about `Extern-py/Nested.Library.py`.

cpitclaudel (changes requested):
Fabio's comments made me realize that I didn't 100% understand the previous situation.
Currently, if we compile this:

```
module A {
  module B {
    method M() {
      print("!!");
    }
  }
}

module C {
  import A.B

  method MC() {
    B.M();
  }
}
```

… we get a module named `A_mB_Compiled`.


It's possible to give the compiled module a different name, using `{:extern "short", "long"}`, which generates a `short "long"` import.

IIUC nesting in Go is done with subfolders.  Maybe we should compile our Dafny modules to go nested packages (but maybe not).  For external modules, does it work to use the `:extern` trick above to refer to a package by it's full path?

In any case, again with Fabio's feedback above, I'm now not so sure that we actually want to peek into the `:extern` name.  One thing I misunderstood is I thought this patch provided new functionality, but in fact the user can (today) already write `{:extern "C", "A_mB_mC"}` if they want that name, right?

robin-aws:
@cpitclaudel mea culpa for not proving more about the motivation - I've added this to the PR description:

The motivation here (which I didn't initially mention enough) is defining a common extern that can be implemented in multiple target languages. It's true that Go supports `{:extern "module_name", "path/to/module"}`, but that doesn't work for other target languages (partially because the compilers generally interpret `{:extern}` as is most convenient for the target language, but more importantly because other languages don't have exactly the same distinction between a module's name and path as Go, so the parameters aren't abstract enough to make sense. The `Extern.dfy` test is the best illustration for this pattern in the test suite, as it's a single common Dafny file with target language files for each target. This PR allows the single-argument `{:extern "A.B.C"}` to be interpreted abstractly enough to make sense for all target languages.

Test/comp/Extern5Nested.py
robin-aws: Why this change, just tidier? I was just trying to be consistent but perhaps we shouldn't have this line at all?

robin-aws:
@fabiomadge Good catch, I hadn't noticed the extra `Extern-py/Nested.Library.py` file. The functionality still works but I would like to clean that up to avoid confusion. I cut a follow-up issue: https://github.com/dafny-lang/dafny/issues/2963

cpitclaudel:
Thanks for adding the motivation.  Depending on the priority of the fix, I would much prefer to have separate ":extern" tags per language, and I think we might be able to do that without much more work.  WDYT?

The reason I'm saying this is that by parsing `.` we are creating a new ad-hoc format for externs, which we will have to keep and maintain in the future.

robin-aws:
After some offline conversation, I agree that it's better not to go any further down the road of supporting the same `{:extern}` declarations across multiple target language. I'm going to keep some of the added functionality here that makes it easier to work with nested target language code, but not claim that it should be possible to use `{:extern}` consistently across languages to import it.

I'm also going to document `Extern.dfy` more clearly to NOT imply that we are committing to what it's doing as an API users can rely on, but just a convenient way of testing the support for multiple languages at once.

Users will be able to provide a compiler plugin instead, if they need to use a single Dafny declaration for multiple target languages.

keyboardDrummer:
> After some offline conversation, I agree that it's better not to go any further down the road of supporting the same {:extern} declarations across multiple target language.

Could you provide a summary of the considerations? Maybe an overview of what language specific externs would look like per language, would enable seeing how difficult it is to have a unifying `{:extern <name>}` attribute?

robin-aws:
> Could you provide a summary of the considerations? Maybe an overview of what language specific externs would look like per language, would enable seeing how difficult it is to have a unifying {:extern <name>} attribute?

Mainly what @cpitclaudel wrote above: externs are currently intended to be passed through verbatim to the target language, and the fact that `Extern.dfy` happens to work as is for the current set is something of a lucky coincidence. Interpreting the args in any way is breaking that assumption, in this case more loosely interpreting `"A.B.C"` as having some kind of `A / B / C` meaning that can be mapped "obviously" to every target language ever.

I don't have a concrete idea of how language specific externs would look yet, as that's only one possible solution to this, but @alex-chew and I will be working on a design/planning document on the more general problem shortly.

PR #2954, opened 2022-10-31
Fix: No more crash when constraining type synonyms

### LINKED ISSUE #2829: Dafny crash on a recursive definition
When evaluating this code

```
datatype Either<+L, +R> = Left(l: L) | Right(r: R)

datatype Lang<!A> = Lang(mem: (string, A) -> bool, repr: iset<A>)

function Atom<A>(s: string, value: A): Lang<A> {
  Lang((input, a) => s == input && a == value, iset{})
}

function Char(c: char): Lang<char> {
  Atom([c], c)
}

function EitherL<A, B>(l: Lang<A>, r: Lang<B>): Lang<Either<A, B>> {
  Lang((input, ab: Either<A, B>) =>
    || (exists a | a in l.repr :: ab == Left(a) && l.mem(input, a))
    || (exists b | b in r.repr :: ab == Right(b) && r.mem(input, b)),
    (iset a | a in l.repr :: Left(a)) + (iset b | b in r.repr :: Right(b))
  )
}

// Nondeterministic in general
function ConcatL<A, B> (l: Lang<A>, r: Lang<B>): Lang<(A, B)> {
  Lang((input, p) =>
    exists s1, s2, a, b | a in l.repr && b in r.repr::
      input == s1 + s2 && p == (a, b),
    iset a, b | a in l.repr && b in r.repr :: (a,b)
  )
}

function Empty<A>(): Lang<A> { Lang((_, _) => false, iset{}) }

function MapL<A, B> (l: Lang<A>, f: A -> B): Lang<B> {
  Lang((input, b) =>
    exists a: A | a in l.repr :: b == f(a) && l.mem(input, a),
    iset a | a in l.repr :: f(a)
  )
}

function Ignore<A>(l: Lang<A>): Lang<()> { MapL(l, (_) => ()) }

function Epsilon(): Lang<()> {
  Lang((input, _) => input == "", iset{})
}

function RepeatAux<A>(pred : (string, A) -> bool, repr: iset<A>, input: string, output: seq<A>): bool
  decreases |output|
{
  || (input == "" && output == [])
  || (exists init, tail, a, alist | a in repr ::
      && input == init + tail
      && output == [a] + alist
      && pred(a, init)
      && RepeatAux(pred, repr, tail, alist))
}
```

With this command:
` /home/codyroux/dotnet/dotnet Dafny.dll   myfile.dfy /verifyAllModules /compile:0 `

gives the following error:

```
Unhandled exception. System.AggregateException: One or more errors occurred. (Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index'))
 ---> System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at Microsoft.Dafny.Resolver.ImposeSubtypingConstraint(Type super, Type sub, ErrorMsg errorMsg)
   at Microsoft.Dafny.Resolver.AssignProxyAndHandleItsConstraints_aux(TypeProxy proxy, Type t, Boolean keepConstraints)
   at Microsoft.Dafny.Resolver.AssignProxyAndHandleItsConstraints_aux(TypeProxy proxy, Type t, Boolean keepConstraints)
   at Microsoft.Dafny.Resolver.AssignProxyAndHandleItsConstraints_aux(TypeProxy proxy, Type t, Boolean keepConstraints)
   at Microsoft.Dafny.Resolver.AssignKnownEndsFullstrength_SuperDirection(TypeProxy proxy)
   at Microsoft.Dafny.Resolver.ProcessFullStrength_SuperDirection(Type t, ISet`1 processed, Boolean& anyNewConstraints)
   at Microsoft.Dafny.Resolver.PartiallySolveTypeConstraints(Boolean allowDecisions)
   at Microsoft.Dafny.Resolver.SolveAllTypeConstraints()
   at Microsoft.Dafny.Resolver.ResolveFunction(Function f)
   at Microsoft.Dafny.Resolver.ResolveClassMemberBodies(TopLevelDeclWithMembers cl)
   at Microsoft.Dafny.Resolver.ResolveTopLevelDecls_Core(List`1 declarations, Graph`1 datatypeDependencies, Graph`1 codatatypeDependencies, Boolean isAnExport)
   at Microsoft.Dafny.Resolver.ResolveModuleDefinition(ModuleDefinition m, ModuleSignature sig, Boolean isAnExport)
   at Microsoft.Dafny.Resolver.ResolveProgram(Program prog)
   at Microsoft.Dafny.Main.Resolve(Program program, ErrorReporter reporter)
   at Microsoft.Dafny.DafnyDriver.ProcessFiles(IList`1 dafnyFiles, ReadOnlyCollection`1 otherFileNames, ErrorReporter reporter, Boolean lookForSnapshots, String programId)
   at Microsoft.Dafny.DafnyDriver.ThreadMain(String[] args)
   at Microsoft.Dafny.DafnyDriver.<>c__DisplayClass9_0.<Main>b__0()
```

With `/version` giving `ng.ide-vscode-2.8.2/out/resources/nightly-2022-09-28-4b97862/dafny/Dafny.dll`.

Apologies for the lack of example minimization.

MikaelMayer:
This PR fixes #2829
I added the corresponding test


Source/DafnyCore/Resolver.cs
```diff
@@ -4810,6 +4810,8 @@ public class ModuleBindings {
       Contract.Requires(sub != null && !(sub is TypeProxy));
       Contract.Requires(errorMsg != null);
       List<int> polarities = ConstrainTypeHead_Recursive(super, ref sub);
+      // Special case: strings.
+      super = super.AsSeqType is { } nonNull ? nonNull : super;
```
RustanLeino, line 4814: Although this does patch the problem for strings, it is not the right general solution. For example, the following two test cases reveal that the problem is more general:

``` dafny
type IntSeq = seq<int>
method Test0()
{
  var a;
  var b := [a];
  var s: IntSeq := a;
}

type BoolSet = set<bool>
method Test1()
{
  var a;
  var b := {a};
  var s: BoolSet := a;
}
```

The general problem is that `super` is a type synonym that fixes some type parameter of the type its based on. This means that `super` and `sub` have different numbers of type arguments. To allow `super` and `sub` to be type synonyms, the enclosing method, `ImposeSubtypingConstraint`, should start by doing:

``` c#
      super = super.NormalizeExpandKeepConstraints();
      sub = sub.NormalizeExpandKeepConstraints();
```

(Apparently, `ConstrainTypeHead_Recursive` already does this and since `sub` is a `ref` parameter to it, `sub` will have been normalized-and-expanded-keeping-constrainted after the call to `ConstrainTypeHead_Recursive`. Still, I suggest doing this for both `super` and `sub` in `ImposeSubtypingConstraint` (as I showed above), since that seems both more robust and less surprising.)

I also suggest adding Clément's two small test cases from #2829 as well as my two test cases from above. You may also add

``` dafny
datatype Option<X> = None | Some(X)
type OptionReal = Option<real>
method Test2()
{
  var a;
  var b := Some(a);
  var s: OptionReal := a;
}
```

even though this test apparently already works (I didn't look to see where, but it seems that `NormalizeExpandKeepConstraints()` has already been applied to `super` in this case).

RustanLeino (changes requested):

docs/dev/news/2829.fix
```diff
@@ -0,0 +1 @@
+No more crash when type constraining and type synonyms
```
RustanLeino, line 1:
```suggestion
No more crash when type constraining type synonyms
```

RustanLeino (changes requested):
Also, please add the additional test cases mentioned in my previous review and in the comments of the original Issue. Thanks!

RustanLeino (approved):

PR #2955, opened 2022-10-31
Fix: Returning a tuple when it should be two variables does not crash Dafny anymore

### LINKED ISSUE #2878: `ArgumentNullException` in resolver (`ProgramTraverser.Traverse`)
Input:

```
method test() returns (a: bool, b: int) {
  return (true, 0);
}
```

Output:

```
Unhandled exception. System.AggregateException: One or more errors occurred. (Value cannot be null. (Parameter 'source'))
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'source')
   at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.Dafny.ProgramTraverser.Traverse(Expression expr, String field, Object parent) in C:\dafny\Source\DafnyCore\Util.cs:line 783
   at Microsoft.Dafny.SubsetConstraintGhostChecker.Traverse(Expression expr, String field, Object parent) in C:\dafny\Source\DafnyCore\Resolver.cs:line 18218
   at Microsoft.Dafny.ProgramTraverser.<>c__DisplayClass17_0.<Traverse>b__0(Expression subExpr) in C:\dafny\Source\DafnyCore\Util.cs:line 767
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.Dafny.ProgramTraverser.Traverse(Statement stmt, String field, Object parent) in C:\dafny\Source\DafnyCore\Util.cs:line 767
   at Microsoft.Dafny.ProgramTraverser.<>c__DisplayClass17_0.<Traverse>b__2(Statement subStmt) in C:\dafny\Source\DafnyCore\Util.cs:line 769
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.Dafny.ProgramTraverser.Traverse(Statement stmt, String field, Object parent) in C:\dafny\Source\DafnyCore\Util.cs:line 767
   at Microsoft.Dafny.ProgramTraverser.Traverse(MemberDecl memberDeclaration, String field, Object parent) in C:\dafny\Source\DafnyCore\Util.cs:line 749
   at Microsoft.Dafny.ProgramTraverser.<>c__DisplayClass15_0.<Traverse>b__0(MemberDecl memberDecl) in C:\dafny\Source\DafnyCore\Util.cs:line 623
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.Dafny.ProgramTraverser.Traverse(TopLevelDecl topd) in C:\dafny\Source\DafnyCore\Util.cs:line 623
   at Microsoft.Dafny.ProgramTraverser.Traverse(List`1 topLevelDecls) in C:\dafny\Source\DafnyCore\Util.cs:line 564
   at Microsoft.Dafny.Resolver.ResolveTopLevelDecls_Core(List`1 declarations, Graph`1 datatypeDependencies, Graph`1 codatatypeDependencies, Boolean isAnExport) in C:\dafny\Source\DafnyCore\Resolver.cs:line 3546
   at Microsoft.Dafny.Resolver.ResolveModuleDefinition(ModuleDefinition m, ModuleSignature sig, Boolean isAnExport) in C:\dafny\Source\DafnyCore\Resolver.cs:line 1032
   at Microsoft.Dafny.Resolver.ResolveProgram(Program prog) in C:\dafny\Source\DafnyCore\Resolver.cs:line 504
   at Microsoft.Dafny.Main.Resolve(Program program, ErrorReporter reporter) in C:\dafny\Source\DafnyCore\DafnyMain.cs:line 210
   at Microsoft.Dafny.Main.ParseCheck(IList`1 files, String programName, ErrorReporter reporter, Program& program) in C:\dafny\Source\DafnyCore\DafnyMain.cs:line 161
   at Microsoft.Dafny.DafnyDriver.ProcessFiles(IList`1 dafnyFiles, ReadOnlyCollection`1 otherFileNames, ErrorReporter reporter, Boolean lookForSnapshots, String programId) in C:\dafny\Source\DafnyDriver\DafnyDriver.cs:line 305
   at Microsoft.Dafny.DafnyDriver.ThreadMain(String[] args) in C:\dafny\Source\DafnyDriver\DafnyDriver.cs:line 116
   at Microsoft.Dafny.DafnyDriver.<>c__DisplayClass9_0.<Main>b__0() in C:\dafny\Source\DafnyDriver\DafnyDriver.cs:line 92
```

MikaelMayer:
This PR fixes #2878


Source/DafnyCore/AST/Expressions.cs
```diff
@@ -1035,7 +1035,12 @@ public DatatypeValue(IToken tok, string datatypeName, string memberName, List<Ex
   }
 
   public override IEnumerable<Expression> SubExpressions {
-    get { return Arguments; }
+    get {
+      return Bindings.Arguments != null ? Arguments :
+        Bindings.ArgumentBindings != null ? Bindings.ArgumentBindings.Select(binding => binding.Actual) :
+        new List<Expression>() { };
+      ;
```
cpitclaudel, line 1042:
```suggestion
      return Arguments ??
        Bindings.ArgumentBindings?.Select(binding => binding.Actual) ??
        Enumerable.Empty<Expression>();
```

cpitclaudel (changes requested):
Can you clarify why we need to look at the ActualBindings and not just the Arguments?

Source/DafnyCore/AST/Expressions.cs
cpitclaudel: (I think this is correct, but I didn't test it)

MikaelMayer: It is correct. I also added a comment to justify the transformation

Source/DafnyCore/AST/Expressions.cs
```diff
@@ -1035,7 +1035,13 @@ public DatatypeValue(IToken tok, string datatypeName, string memberName, List<Ex
...
 
   public override IEnumerable<Expression> SubExpressions {
-    get { return Arguments; }
+    get {
+      // Arguments is available only when fully resolved. Otherwise,
+      // only unresolved arguments (Actual) are available.
+      return Arguments ??
+             Bindings.ArgumentBindings?.Select(binding => binding.Actual) ??
+             Enumerable.Empty<Expression>();
+    }
```
cpitclaudel, line 1044:
```suggestion
      return Arguments ?? Enumerable.Empty<Expression>();
```

cpitclaudel (changes requested):
Based on this:

```
  /// <summary>
  /// Returns the non-null subexpressions of the Expression.  To be called after the expression has been resolved; this
  /// means, for example, that any concrete syntax that resolves to some other expression will return the subexpressions
  /// of the resolved expression.
  /// </summary>
  public virtual IEnumerable<Expression> SubExpressions {
    get { yield break; }
  }
```

… I don't think we should handle the non-resolved case.  WDYT?

Source/DafnyCore/AST/Expressions.cs
MikaelMayer: There is no reason that we should return null or an empty list for sub-expressions when there _are_ sub-expressions in the AST.
Are you ok if I fix the comment that you mention?

MikaelMayer: Moreover, in the upcoming formatter PR, there will be a new field "PreResolveSubExpressions", which is mostly equal to SubExpressions. But SubExpressions' behavior is undefined before Resolution, so it's up to us to decide what its meaning is. And what more obvious to the meaning that sub-expressions to be rexolved? WDYT?

cpitclaudel: I think we shouldn't change that comment without changing the implementation of every other SubExpressions field, and I think that's beyond the scope of this PR, no?  And I don't think we should introduce new inconsistencies.

Source/DafnyCore/AST/Expressions.cs
```diff
@@ -1035,7 +1035,8 @@ public DatatypeValue(IToken tok, string datatypeName, string memberName, List<Ex
   }
 
   public override IEnumerable<Expression> SubExpressions {
-    get { return Arguments; }
+    get {
+      return Arguments ?? Enumerable.Empty<Expression>();
```
cpitclaudel, line 1039: Aren't you missing a curly brace here?

cpitclaudel (dismissed):
Fix looks good, but there's a syntax error I think?

cpitclaudel (approved):

PR #2957, opened 2022-10-31
Add README files to packages, for use with NuGet

atomb:
This PR links to the README files for the `Dafny` and `DafnyLanguageServer` packages, which will make them visible on nuget.org.

It also adds a previously-missing license identifier to `DafnyLanguageServer`, making it work better with NuGet.

It would be useful to add README files for the other packages uploaded to nuget.org, but that would require writing them.


Source/Dafny/Dafny.csproj
```diff
@@ -12,6 +12,7 @@
       <PackAsTool>true</PackAsTool>
       <ToolCommandName>dafny</ToolCommandName>
       <PackageLicenseExpression>MIT</PackageLicenseExpression>
+      <PackageReadmeFile>README.md</PackageReadmeFile>
```
MikaelMayer, line 15: Wow I did not know there was a special syntax for readme files !

MikaelMayer (approved):
That sounds good to me. What about DafnyDriver?

atomb:
> That sounds good to me. What about DafnyDriver?

The top-level `dafny` executable is now the `Dafny` package. It'd be great to have a README for `DafnyDriver` as well as all of the other library packages, but that's more work as they need to be written. :)

PR #2958, opened 2022-11-01
fix(js): areEqual(number, bignum) edge case

### LINKED ISSUE #2956: JS areEqual doesn't always compare numbers correctly
The JS runtime currently defines `_dafny.areEqual(a, b)` such that it compares as expected if:
* both `a` and `b` are just plain JS `number`s
* both `a` and `b` are `BigNumber`s
* a is a `BigNumber` and b is a `number`

But it doesn't work if `a` is a `number` and `b` is a `BigNumber`. As far as I can tell, the JS compiler correctly avoids generating comparisons between these types - but it's safer to make `areEqual` to work correctly in this case anyway.

alex-chew:
Fixes #2956.


Test/git-issues/git-issue-2956b.js
```diff
@@ -0,0 +1 @@
+global.OneOne = () => _dafny.Seq.of(1);
```
cpitclaudel, line 1: I would add a comment here, since this is a (purposeful) abuse of the API

docs/dev/news/2958.fix
```diff
@@ -0,0 +1 @@
+The JS runtime now handles pathological number-to-BigNumber equality comparisons.
```
cpitclaudel, line 1: Do we actually want to include this in the release notes?  I think of the "fixed" section as "things users might have run into / might want to know they should try again since the problem is fixed.

cpitclaudel (changes requested):

cpitclaudel:
LGTM, two small suggestions

Test/git-issues/git-issue-2956b.js
alex-chew: Good idea, will do.
```suggestion
// The JS compiler would normally convert this integer value to a BigNumber.
// Omitting the conversion tests an edge case that would not arise during normal compilation.
global.OneOne = () => _dafny.Seq.of(1);
```

docs/dev/news/2958.fix
alex-chew: That's fair, especially since this issue only comes up if writing malformed externs. I'll remove the note.

cpitclaudel (approved):

PR #2967, opened 2022-11-01
Fix: Default generic values no longer cause compilation error

### LINKED ISSUE #2885: Cannot compile default generic values in C#
If you compile the following program:
```dafny
datatype KeyValues<T> = Store(underlying: map<string, T> := map[])

function method CreateStore<T>(): (r: KeyValues<T>) {
  Store()
}
```
You'll get a compilation error:
```
Errors compiling program into csharpcompilerbug
(4886,69): error CS0246: The type or namespace name 'T' could not be found (are you missing a using directive or an assembly reference?)

(4886,36): error CS1503: Argument 1: cannot convert from 'Dafny.Map<Dafny.ISequence<char>, T>' to 'Dafny.IMap<Dafny.ISequence<char>, __T>'
```
This comes from the fact that Dafny did not infer the type parameter of the default variable to be the same T, and there is no way to provide this in the constructor as far as I know.

The workaround is to explicit the default value for now:
```dafny
datatype KeyValues<T> = Store(underlying: map<string, T> := map[])

function method CreateStore<T>(): (r: KeyValues<T>) {
  var x: map<string, T> := map[];
  Store(x)
}
```

MikaelMayer:
This PR fixes #2885
What happened was that, in the case of maps, the substitution for default value expressions did not verify if the type should be substituted.
The fix was to do the same check as for the displayExpressions just above.


atomb (approved):
Although I'm not deeply familiar with the relevant code, the change you made seems right to me, given the equivalent pattern above.

PR #2973, opened 2022-11-02
Fix: Labels are no longer compiled in the case of variable declarations

### LINKED ISSUE #2576: Miscompilation (scope issue) of labels
Input:

```
method Main() {
  label l:
    var x := 1;
  print x;
}
```

Output:

```
$ Dafny /compileTarget:java /compile:4 bug.dfy

Wrote textual form of target program to bug-java/bug.java
Additional target code written to bug-java/_System/nat.java
Additional target code written to bug-java/_System/__default.java
/home/cpitcla/git/dafny/SMT-LIB-transfromer/Src/Bench/Java.Main-java/bug-java/_System/__default.java:16: error: cannot find symbol
    System.out.print(String.valueOf(_0_x));
                                    ^
  symbol:   variable _0_x
  location: class __default
1 error
```

### LINKED ISSUE #2608: JavaScript compiler and labels issue
The following code crashes a compiled JavaScript code.

bug_js.dfy:
```dafny
method GetNext(i: int) returns (j: int, possible: bool) {
  if i == 1 {
    possible := false;
    j := 0;
  } else {
    possible := true;
    j := if i % 2 == 0 then  i / 2 else i * 3 + 1;
  }
}

method Main()
{
  var i := 100;
  var k := 27;
  while i > 0
    invariant i >= 0
  {
    label before:
    var newK, possible := GetNext(k);
    if(!possible) {
      break;
    }
    k := newK;
    print k;
    i := i - 1;
  }
}
```

```dafny

> dafny /spillTargetCode:2 /compileTarget:js /compile:3 bug_js.dfy

Running...

[stdin]:1004
        throw e
        ^

ReferenceError: _3_possible is not defined
    at Main ([stdin]:1098:13)
    at Object.$module.HandleHaltExceptions ([stdin]:999:7)
```

The generated code is ill-formed, but it's not possible to see it until it's executed:

```js
    static Main() {
      let _0_i;
      _0_i = new BigNumber(100);
      let _1_k;
      _1_k = new BigNumber(27);
      L0: {
        while ((_dafny.ZERO).isLessThan((_0_i))) {
          C0: {
            L1: {
              let _2_newK;
              let _3_possible;
              let _out0;
              let _out1;
              let _outcollector0 = _module.__default.GetNext(_1_k);
              _out0 = _outcollector0[0];
              _out1 = _outcollector0[1];
              _2_newK = _out0;
              _3_possible = _out1;
            }
            if (!(_3_possible)) { // Here _3_possible is out of scope.
              break L0;
            }
            _1_k = _2_newK;
            process.stdout.write(_dafny.toString(_1_k));
            _0_i = (_0_i).minus((_dafny.ONE));
          }
        }
      }
      return;
    }
```

MikaelMayer:
This PR fixes #2608 and fixes #2576

Here is the story. Whenever there is a label `L1` to a statement `S`, the JavaScript compiler wraps the statement with curly braces `{}` and put the label in front of it, like this:
```
L1 {
  [[S]] // Translate S to JavaScript
}
```
The problem was that, if there are labelled variable declarations, they are compiled to `let` and not available after `}`.
Looking carefully at it, we don't have goto in Dafny so labels are used only to break out of loops. But these are no loops, only variable declarations. Hence, I claim it make sense to just remove the labels when compiling variable declarations.
The test now passes.

### DIFF
--- a/Source/DafnyCore/Compilers/SinglePassCompiler.cs
+++ b/Source/DafnyCore/Compilers/SinglePassCompiler.cs
@@ -4331,7 +4331,9 @@ namespace Microsoft.Dafny.Compilers {
         //   <prelude>   // filled via copyInstrWriters -- copies out-parameters used in letexpr to local variables
         //   ss          // translation of ss has side effect of filling the top copyInstrWriters
         var w = writer;
-        if (ss.Labels != null) {
+        if (ss.Labels != null && !(ss is VarDeclPattern or VarDeclStmt)) {
+          // We are not breaking out of VarDeclPattern or VarDeclStmt, so the labels there are useless
+          // They were useful for verification
           w = CreateLabeledCode(ss.Labels.Data.AssignUniqueId(idGenerator), false, w);
         }
         var prelude = w.Fork();
--- a/docs/dev/news/2608.fix
+++ b/docs/dev/news/2608.fix
@@ -1,0 +1,1 @@
+Labels should not be compiled in the case of variable declarations
--- a/docs/dev/news/2608.fix
+++ b/docs/dev/news/2608.fix
@@ -1,1 +1,1 @@
-Labels should not be compiled in the case of variable declarations
+Labels are no longer compiled in the case of variable declarations
--- a/Test/git-issues/git-issue-2608.dfy
+++ b/Test/git-issues/git-issue-2608.dfy
@@ -1,0 +1,30 @@
+// RUN: %dafny_0 /compile:3 /compileTarget:js "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method GetNext(i: int) returns (j: int, possible: bool) {
+  if i == 1 {
+    possible := false;
+    j := 0;
+  } else {
+    possible := true;
+    j := if i % 2 == 0 then  i / 2 else i * 3 + 1;
+  }
+}
+
+method Main()
+{
+  var i := 100;
+  var k := 27;
+  while i > 0
+    invariant i >= 0
+  {
+    label before:
+    var newK, possible := GetNext(k);
+    if(!possible) {
+      break;
+    }
+    k := newK;
+    print k;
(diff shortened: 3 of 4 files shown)


Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -4331,7 +4331,9 @@ private class GetterSetterLvalueImpl : ILvalue {
         //   <prelude>   // filled via copyInstrWriters -- copies out-parameters used in letexpr to local variables
         //   ss          // translation of ss has side effect of filling the top copyInstrWriters
         var w = writer;
-        if (ss.Labels != null) {
+        if (ss.Labels != null && !(ss is VarDeclPattern or VarDeclStmt)) {
```
robin-aws, line 4334: Definitely sounds like a good change, but how confident are we that no other statement cases would similarly be broken by being wrapped in a block?

docs/dev/news/2608.fix
```diff
@@ -0,0 +1 @@
+Labels should not be compiled in the case of variable declarations
```
robin-aws, line 1:
```suggestion
Labels are no longer compiled in the case of variable declarations
```

Phrase as a description of the current behavior after the change rather than a timeless assertion about how things "should" be :)

Source/DafnyCore/Compilers/SinglePassCompiler.cs
MikaelMayer: The only problem of wrapping with block quotes (at least in javascript), is that it creates a "local scope". Other than that, you can insert {} around any statement without changing its meaning.
Hence, in our case, the only place where this matter is that we are introducing new variables for subsequent statements. All other statements are only modifying existing vars that were already in scope, so they have no reason to be broken by being wrapped into a labelled block.

robin-aws (approved):

PR #2979, opened 2022-11-03
Fix: old() cannot be inferred as a trigger alone

### LINKED ISSUE #2593: Malformed triggers
The following Dafny program has a `old` in a less-than-commendable, but legal, position:

``` dafny
predicate P(x: int)

lemma L()
  ensures forall y :: P(old(y))
```

Alas, the program causes Dafny to generate malformed Boogie code, because of the trigger Dafny generates for the quantifier.

```
% dafny test.dfy
test.dfy(5,28): Error: a matching pattern must be more than just a variable by itself: y#1
test.dfy(5,28): Error: a matching pattern must be more than just a variable by itself: y#1
2 name resolution errors detected in test__module.bpl
```

Additional trigger information:

```
Selected triggers: {old(y)}
Rejected triggers: {P(old(y))} (more specific than {old(y)})
```

MikaelMayer:
This PR fixes #2593
I added the corresponding test. Basically, previously, old(...) was considered as a potential trigger. However, the translation of Boogie of old(...) expressions is not a function, but a previous variable... so this had no meaning
I removed old(...) from the list of potential triggers.

Side notes: 
* Removing old() as a trigger altogether does not work. Even ensuring it's not a trigger killer won't do it. The test dafny1/MatrixFun.dfy fails in both of these cases, because it misses one important trigger
* If there is something like `old(f(y)+1)`, even though `old()` is marked as useless, the trigger `old(f(y))` and `f(y)` will still be created due to special traversal of the tree. This is expected. This traversal will never wrap a variable with `old()` anyway.

### DIFF
--- a/Source/DafnyCore/Triggers/TriggersCollector.cs
+++ b/Source/DafnyCore/Triggers/TriggersCollector.cs
@@ -231,7 +231,7 @@ namespace Microsoft.Dafny.Triggers {
                    expr is SeqConstructionExpr) {
           annotation = AnnotateOther(expr, false);
         } else {
-          annotation = AnnotateOther(expr, true);
+          annotation = AnnotateOther(expr, expr is not OldExpr);
         }
       }
 
@@ -245,7 +245,7 @@ namespace Microsoft.Dafny.Triggers {
           expr is SeqSelectExpr ||
           expr is MultiSelectExpr ||
           expr is MemberSelectExpr ||
-          expr is OldExpr ||
+          //expr is OldExpr ||
           expr is ApplyExpr ||
           expr is DisplayExpression ||
           expr is MapDisplayExpr ||
--- a/Source/DafnyCore/Triggers/TriggersCollector.cs
+++ b/Source/DafnyCore/Triggers/TriggersCollector.cs
@@ -231,7 +231,7 @@ namespace Microsoft.Dafny.Triggers {
                    expr is SeqConstructionExpr) {
           annotation = AnnotateOther(expr, false);
         } else {
-          annotation = AnnotateOther(expr, expr is not OldExpr);
+          annotation = AnnotateOther(expr, true);
         }
       }
 
@@ -245,7 +245,7 @@ namespace Microsoft.Dafny.Triggers {
           expr is SeqSelectExpr ||
           expr is MultiSelectExpr ||
           expr is MemberSelectExpr ||
-          //expr is OldExpr ||
+          (expr is OldExpr { Useless: false }) ||
           expr is ApplyExpr ||
           expr is DisplayExpression ||
           expr is MapDisplayExpr ||
--- a/Test/git-issues/git-issue-2593.dfy.expect
+++ b/Test/git-issues/git-issue-2593.dfy.expect
@@ -1,0 +1,4 @@
+git-issue-2593.dfy(7,24): Warning: Argument to 'old' does not dereference the mutable heap, so this use of 'old' has no effect
+git-issue-2593.dfy(7,24): Warning: Argument to 'old' does not dereference the mutable heap, so this use of 'old' has no effect
+
+Dafny program verifier finished with 0 verified, 0 errors
--- a/Test/git-issues/git-issue-2593.dfy.expect
+++ b/Test/git-issues/git-issue-2593.dfy.expect
@@ -1,4 +1,4 @@
 git-issue-2593.dfy(7,24): Warning: Argument to 'old' does not dereference the mutable heap, so this use of 'old' has no effect
-git-issue-2593.dfy(7,24): Warning: Argument to 'old' does not dereference the mutable heap, so this use of 'old' has no effect
+git-issue-2593.dfy(7,10): Warning: /!\ No terms found to trigger on.
 
 Dafny program verifier finished with 0 verified, 0 errors
--- a/Source/DafnyCore/AST/Expressions.cs
+++ b/Source/DafnyCore/AST/Expressions.cs
@@ -1831,6 +1831,7 @@ public class OldExpr : Expression {
   [Peer]
   public readonly Expression E;
   public readonly string/*?*/ At;
+  [FilledInDuringResolution] public bool Useless = false;
   [FilledInDuringResolution] public Label/*?*/ AtLabel;  // after that, At==null iff AtLabel==null
   [ContractInvariantMethod]
   void ObjectInvariant() {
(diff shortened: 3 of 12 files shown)


cpitclaudel (changes requested):
There is already some code related to `old` in https://github.com/dafny-lang/dafny/commit/885a5b7ed3a03e280c39506ef562424088374395 ; can you check that the new logic is compatible with that?

Beyond that, I think the right place to filter trigger terms is in https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Triggers/TriggersCollector.cs , not in QuantifiersCollection; WDYT?

MikaelMayer:
> There is already some code related to `old` in [885a5b7](https://github.com/dafny-lang/dafny/commit/885a5b7ed3a03e280c39506ef562424088374395) ; can you check that the new logic is compatible with that?

This commit seems weird. It seems to assume that `old(f(x))` and `old(g(x))` are valid triggers. What is true is that `f(x)` and `g(x)` are valid triggers, and old expressions in Dafny are wiped out only to ensure any referenced heap is a special heap, which does not participate in triggers as far as I can tell.
https://github.com/dafny-lang/dafny/blob/97c30648e1ac5d9cce29bdee32ec52cba7ab9fdb/Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs#L713-L715
https://github.com/dafny-lang/dafny/blob/97c30648e1ac5d9cce29bdee32ec52cba7ab9fdb/Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs#L147-L154

> Beyond that, I think the right place to filter trigger terms is in https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Triggers/TriggersCollector.cs , not in QuantifiersCollection; WDYT?

I move the filter to the place you designed. Note that not only one had to remove OldExpr, but also designate that OldExpr is not a trigger killer.

MikaelMayer:
One test is failing, and it made me realize one thing. Triggers can also depend on variables in scope, and in the case of the heap, some triggers can depend on some states of the heap.
Hence, it makes sense to have old() as part of triggers. The only way to not generate malformed boogie is to ignore old() only if they are useless. I'm going to change that.

Source/DafnyCore/Triggers/TriggersCollector.cs
```diff
@@ -245,7 +245,7 @@ internal class TriggersCollector {
           expr is SeqSelectExpr ||
           expr is MultiSelectExpr ||
           expr is MemberSelectExpr ||
-          expr is OldExpr ||
+          (expr is OldExpr { Useless: false }) ||
```
cpitclaudel, line 248: I don't understand this part.  Why is `old(x)` a trigger candidate when `old` is useful? (Shouldn't just `x` be the trigger?

Test/git-issues/git-issue-2593.dfy.expect
```diff
@@ -0,0 +1,4 @@
+git-issue-2593.dfy(7,24): Warning: Argument to 'old' does not dereference the mutable heap, so this use of 'old' has no effect
+git-issue-2593.dfy(7,10): Warning: /!\ No terms found to trigger on.
```
cpitclaudel, line 2: Shouldn't we produce `P(y)` as the trigger here?

Source/DafnyCore/Triggers/TriggersCollector.cs
cpitclaudel: I'm sorry for the piecemeal review.  I need to spend the time to read this carefully (or if you're confident we can just ship it), but I can't do that today :/

MikaelMayer: So I was basically wrong. There was one example of matrices where old(m[i, j]) was not the same trigger as m[i, j] because m was internally using a heap constant. See the error there:

https://github.com/dafny-lang/dafny/actions/runs/3395616639/jobs/5645735703

Test/git-issues/git-issue-2593.dfy.expect
cpitclaudel: But in this case old is redundant, right?  I guess maybe it's not an issue.
But is `old()` shouldn't always be a trigger killer, right?  Like `old(c.x[i])` is a reasonable trigger for `i`.

MikaelMayer: > But in this case old is redundant, right? I guess maybe it's not an issue.

I don't think it's an issue either. If there was a warning, then any expression enclosing old() would not be considered as a trigger, but after fixing this warning, all the other triggers you dreamed of become available.

> But is old() shouldn't always be a trigger killer, right? Like old(c.x[i]) is a reasonable trigger for i.

Right. `old()` become a killer trigger only when it's marked as useless, which happens inside the rewriter check. So your example would work as usual, since it would not generate any warning.

cpitclaudel: But why do we bother marking it as a trigger killer when it's useless?

MikaelMayer: Let me rephrase. This PR's fix is to prevent old(X) becoming a trigger if old(X) was previously marked as useless. The enclosing function is `IsPotentialTriggerCandidate` and will just return `false` if that is the case (previously, it returned true). That's the only change that this PR does.
https://github.com/dafny-lang/dafny/blob/4d65937ebc88c7371cb598c4005cddb82b4ff140/Source/DafnyCore/Triggers/TriggersCollector.cs#L243-L254

Now, it just happens that the calling function has a few exceptions in the case `IsPotentialTriggerCandidate` returns false, to not mark something as a trigger killer (the second argument of `AnnotateOther`)
https://github.com/dafny-lang/dafny/blob/4d65937ebc88c7371cb598c4005cddb82b4ff140/Source/DafnyCore/Triggers/TriggersCollector.cs#L212-L234
I just did not bother adding `old()` as an exception. Why would we bother adding it as an exception when it's already a warning?

cpitclaudel (approved):

cpitclaudel:
OK, I think this is ready to go.  

For the record, I didn't expect any special case for `Useless` to be needed; instead, I thought it would be enough to reject `old(…)` as a trigger in all cases.

I was hoping it would work because we have code that takes triggers found inside an `old` block and lifts them by re-wrapping them in `old`, so that in `old(f(x.a) + 1)` we select `old(f(x.a))` as the trigger, not `f(x.a)`.

That code should make it unnecessary to consider the whole `OldExpr` as a trigger, since we should *reconstruct* an `OldExpr` from whatever triggers we found inside the old expr.  An yet doing things that was (removing the `{Useless: true}` in the PR) breaks `dafny1/MatrixFun.dfy`, so clearly my mental model is missing something here.

Nonetheless, the only case where we can have `old(<variable>)` is if `old` is useless, so ruling that case out (and not all `old`s) is a reasonable fix to this issue, and we can delay updating my mental model :)

Thanks @MikaelMayer for the investigation!

Issue #2976, opened 2022-11-03
Output encoding issues on Windows


cpitclaudel:
On Windows, the following program prints (with `dafny run`) different outputs depending on the compiler that it gets executed on:

```dafny
method Main() {
  print [8364 as char];
}
```

Specifically:

### On Windows in `cmd.exe`:

```
C:\>FOR %t in (cs py java js) DO Dafny run --no-verify --target=%t encodings.dfy

C>Dafny run --no-verify --target=cs encodings.dfy

Dafny program verifier did not attempt verification
?
C:\>Dafny run --no-verify --target=py encodings.dfy

Dafny program verifier did not attempt verification

C:\>Dafny run --no-verify --target=java encodings.dfy

Dafny program verifier did not attempt verification
Ç

C:\>Dafny run --no-verify --target=js encodings.dfy

Dafny program verifier did not attempt verification
Γé¼
```

Arguably, this is the user's fault: the above happens because the default `cmd.exe` codepage doesn't even have the € character.  So, let's fix that:

### On windows with `chcp 65001` (the utf-8 codepage):

```
C:\>FOR %t in (cs py java js) DO Dafny run --no-verify --target=%t encodings.dfy

C:\>Dafny run --no-verify --target=cs encodings.dfy

Dafny program verifier did not attempt verification
€
C:\>Dafny run --no-verify --target=py encodings.dfy

Dafny program verifier did not attempt verification

C:\>Dafny run --no-verify --target=java encodings.dfy

Dafny program verifier did not attempt verification
�

C:\>Dafny run --no-verify --target=js encodings.dfy

Dafny program verifier did not attempt verification
€
```

C# and JS are good.  Python and Java are not.  

Python can be fixed with `PYTHONIOENCODING=utf8`.  
The only way to fix java is to use `JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF-8`, but unfortunately that also pollutes the output:

```
C:\>Dafny run --no-verify --target=java encodings.dfy

Dafny program verifier did not attempt verification
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
€
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
```

### On Ubuntu/WSL:

```console
$ for t in cs py java js; do Dafny run --target=$t encodings.dfy; done

Dafny program verifier finished with 1 verified, 0 errors
€
Dafny program verifier finished with 1 verified, 0 errors
€
[...]

PR #2982, opened 2022-11-03
fix: Force UTF-8 encoding in C# lit runner

cpitclaudel:
This is intended to make it possible to test Unicode output.  I skipped Java because Java needs a separate patch (since it doesn't respect the current encoding / codepage, see https://github.com/dafny-lang/dafny/issues/2976)

### DIFF
--- a/Source/XUnitExtensions/Lit/ShellLitCommand.cs
+++ b/Source/XUnitExtensions/Lit/ShellLitCommand.cs
@@ -33,12 +33,69 @@ namespace XUnitExtensions.Lit {
       process.StartInfo.RedirectStandardInput = true;
       process.StartInfo.RedirectStandardOutput = true;
       process.StartInfo.RedirectStandardError = true;
-      process.StartInfo.CreateNoWindow = true;
 
       process.StartInfo.EnvironmentVariables.Clear();
       foreach (var passthrough in passthroughEnvironmentVariables) {
         process.StartInfo.EnvironmentVariables.Add(passthrough, Environment.GetEnvironmentVariable(passthrough));
       }
+
+      // Getting console encodings right is tricky on Windows.
+      //
+      // Some background: the current console has a codepage, which is what
+      // programs that output text are supposed to use.  The command `chcp` can
+      // be used to test or set it:
+      //
+      // > chcp
+      // Active code page: 1252
+      //
+      // So, if the console has its codepage set to 850 (the old IBM codepage),
+      // then a program that wants to print `é` (U+00E9, Latin Small Letter E
+      // With Acute) should output a single byte, #x82.  But if the codepage is
+      // 1252 (Latin 1), then that same program should output the single
+      // byte `#xE9`.  Finally, if the codepage is 65001 (UTF8), then that same
+      // program should output `#xC3 #xA9`.
+      //
+      // Thankfully, the C# runtime takes care of this for us.  It uses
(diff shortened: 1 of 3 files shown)


Source/XUnitExtensions/Lit/ShellLitCommand.cs
```diff
@@ -33,13 +33,70 @@ public class ShellLitCommand : ILitCommand {
...
+      // With Acute) should output a single byte, #x82.  But if the codepage is
+      // 1252 (Latin 1), then that same program should output the single
+      // byte `#xE9`.  Finally, if the codepage is 65001 (UTF8), then that same
+      // program should output `#xC3 #xA9`.
+      //
+      // Thankfully, the C# runtime takes care of this for us.  It uses
+      // `GetConsoleCP` to determine what encoding to use, then uses that when
+      // `Console.WriteLine` is called (and replaces unsupported characters by
+      // the fallback character `?`).  Here is a concrete demo, with character
+      // U+20AC (€) (which is supported by codepage 85001 but not 850):
```
robin-aws, line 62:
```suggestion
      // U+20AC (€) (which is supported by codepage 65001 but not 850):
```

Source/XUnitExtensions/Lit/ShellLitCommand.cs
```diff
@@ -33,13 +33,70 @@ public class ShellLitCommand : ILitCommand {
...
+      // output, as is done below.
+      //
+      // … but this is not enough, because dotnet run (or XUnit) also resets the
+      // codepage.  That's a bug there, and in the meantime the best hack we
+      // have is to override that codepage by setting `OutputEncoding`.
+
+      process.StartInfo.CreateNoWindow = false;
+      Console.OutputEncoding = System.Text.Encoding.UTF8;
+
+      // Finally, for Java + Ubuntu, we make sure to forward LANG:
```
robin-aws, line 95:
```suggestion
      // Finally, for Java + Ubuntu, we make sure to set LANG:
```

Test/metatests/OutputEncoding.dfy
```diff
@@ -0,0 +1,10 @@
+// RUN: %baredafny verify %args "%s" > "%t"
+// RUN: %baredafny run --no-verify --target=cs %args "%s" >> "%t"
+// RUN: %baredafny run --no-verify --target=js %args  "%s" >> "%t"
+// RUN: %baredafny run --no-verify --target=go %args  "%s" >> "%t"
+// RUN: %baredafny run --no-verify --target=py %args  "%s" >> "%t"
+// RUN: %diff "%s.expect" "%t"
```
robin-aws, line 6: Missing the `--target=java` case.

I'm just fine with ignoring C++ for now, but could we use `%testDafnyForEachCompiler` and just add an `int` somewhere (so C++ can cleanly opt out of this test case)?

robin-aws (changes requested):
Very nice work getting this working (again) :)

Test/metatests/OutputEncoding.dfy
cpitclaudel: This is on purpose, because Java doesn't work.
The next patch will make it work (by forcing UTF-8)

robin-aws: Never mind, reading comprehension fail on the PR description. :)

PR #2983, opened 2022-11-03
Force Java output encoding to match the current encoding in `dafny run`

cpitclaudel:
Fixes https://github.com/dafny-lang/dafny/issues/2976
Depends on https://github.com/dafny-lang/dafny/pull/2982


cpitclaudel:
This works, but I'm not sure it's the right way to go, because we have no way to make Python's output consistent.  Hence, maybe we should make `dafny run` produce UTF-8 in all cases?  See the discussion in https://github.com/dafny-lang/dafny/issues/2976

cpitclaudel:
I'm closing this: we have decided in https://github.com/dafny-lang/dafny/issues/2999 to instead force all `dafny run` calls to use UTF-8

PR #2984, opened 2022-11-03
fix: Fill in missing type substitutions

### LINKED ISSUE #2947: Declaration of generic extreme predicate causes crash
The following program declarations cause the verifier to emit malformed Boogie:

``` dafny
codatatype Stream<X> = Nil | Cons(head: X, tail: Stream<X>)

least predicate Finite<X>(s: Stream<X>) {
  s == Nil || Finite(s.tail)
}
```

Note that the least predicate is defined with a type parameter.

Here is the output, where the error messages are coming from Boogie.

```
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
test.dfy(5,25): Error: undeclared identifier: _module._default.Finite$X
```

RustanLeino:
Fixes #2947

This PR fixes some missing type substitutions that affected the Boogie translation of extreme predicates.


Source/DafnyCore/AST/TopLevelDeclarations.cs
```diff
@@ -845,6 +845,21 @@ public class ModuleDefinition : IDeclarationOrUsage, INamedRegion, IAttributeBea
     }
   }
 
+  /// <summary>
+  /// Yields all functions and methods that are members of some type in the given list of
+  /// declarations, including prefix lemmas and prefix predicates.
+  /// </summary>
+  public static IEnumerable<ICallable> AllCallablesIncludingPrefixDeclarations(List<TopLevelDecl> declarations) {
```
keyboardDrummer, line 852: I'd like to have each language construct be as independent as possible. For extreme lemmas and predicates this would also mean they have mostly outwards references. Would it be possible to achieve this?

Would it be possible to translate extremes to their prefix versions and then forget about the originals? Or maybe create the prefix versions next to the originals?

keyboardDrummer (approved):

Source/DafnyCore/AST/TopLevelDeclarations.cs
RustanLeino: We need both. A program needs the extremes, since that's what you'll ultimately will use for the property you want. The prefix predicates/lemmas give a way to prove properties about the extremes, so they are also needed.

Actually, the more surprising thing (to me) is that `AllCallables` (which excludes the prefix versions) is used so many times.

Source/DafnyCore/Substituter.cs
```diff
@@ -38,13 +38,19 @@ public class Substituter {
         return new StaticReceiverExpr(e.tok, ty, e.IsImplicit) { Type = ty };
       } else if (expr is LiteralExpr literalExpr) {
         if (literalExpr.Value == null) {
-          return new LiteralExpr(literalExpr.tok) { Type = Resolver.SubstType(literalExpr.Type, typeMap) };
+          var ty = Resolver.SubstType(literalExpr.Type, typeMap);
```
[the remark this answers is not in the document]
RustanLeino: There's a (possibly misguided) optimization in the `Substitute(Expression expr)` method, which is that if the result is pointer-equal to `expr` if the substitution has no effect. An illustrative example of this is:

``` c#
      } else if (expr is MultiSetFormingExpr) {
        var e = (MultiSetFormingExpr)expr;
        var se = Substitute(e.E);
        if (se != e.E) {
          newExpr = new MultiSetFormingExpr(expr.tok, se);
        }
```

This optimization is error prone. For `LiteralExpr`, the optimization was missing, so I added it. In other places (like `DatatypeValue`, the optimization was too aggressive (meaning, "wrong") because it returned `expr` even when it should have returned a new `DatatypeValue` with a different `InferredTypeArgs` field.

The optimization is intended to save memory, but this may not make any difference, and the optimization makes the code significantly more complicated. If you suggest, I'd be happy to remove this optimization from `Substitute(Expression`.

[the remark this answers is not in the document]
RustanLeino: I built a local version with a simplified `Substitute` (that is, I removed the optimization). It passes. I'd be happy to add it to this PR, if you think, or to file a separate PR for it.

Source/DafnyCore/AST/TopLevelDeclarations.cs
keyboardDrummer: > We need both. A program needs the extremes, since that's what you'll ultimately will use for the property you want. The prefix predicates/lemmas give a way to prove properties about the extremes, so they are also needed.

Sure, so can't we place both side by side?

Source/DafnyCore/Substituter.cs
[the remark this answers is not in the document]
keyboardDrummer: > The optimization is intended to save memory, but this may not make any difference, and the optimization makes the code significantly more complicated.

Without a measurement about the performance difference I'd prefer to leave it like it is, unless you're confident the optimisation doesn't have a significant effect. I'm not sure whether comparing the difference in the test-suite runtime is a good test.

I'm hoping we can at some point generate the code for various behaviors such as Substitutor, Cloner, SubExpressions, SubStatements, and likely some others.

Source/DafnyCore/AST/TopLevelDeclarations.cs
RustanLeino: I don't understand. The `AllCallablesIncludingPrefixDeclarations` and `AllCallables` properties are declared side by side. Are you perhaps suggesting that the prefix predicates/lemmas be placed in the list of the enclosing module's top-level-decls (rather than being linked off the corresponding extreme predicate/lemma)? (And if so, do you think that will improve the organization for other parts of the implementation?)

Source/DafnyCore/Substituter.cs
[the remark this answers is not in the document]
RustanLeino: Okay, I'll leave it as is. And yes, I agree that _generating_ code for these sorts of routines would be far better than the manual upkeep we have to do today.

Source/DafnyCore/AST/TopLevelDeclarations.cs
RustanLeino: I looked into this in detail for `.PrefixLemma` and `.PrefixPredicate`. In places where the code looked like your first example, I change `AllCallables` to `AllCallablesIncludingPrefixDeclarations`. I also improved code in the various other visitors, making use of newer C# constructs. And I improved the iteration over declarations in `IsEssentiallyEmptyModuleBody()`, to be more exact than before.

In the code base, what happens with extreme lemmas/predicates and prefix lemmas/predicates is quite varied. In some places, the extreme declaration is used and the prefix version ignored. In other places, it's the other way around. And in some places, the two are handled in different ways. Therefore, I suggest that any further improvements be done separately from this PR.

keyboardDrummer (approved):
Approving but I expect fresh bugs related to this to be added in the future. How should a caller decide whether to call `AllCallablesIncludingPrefixDeclarations` or `AllCallables` without having to dive into the semantics of extreme lemma's? I feel that as long as there's these two options there's a source for bugs.

PR #2994, opened 2022-11-04
Fix: Better reporting if 'this' used in a subset type - and no crash

### LINKED ISSUE #2068: Resolver crash
The following crashes for me in 3.5.0.40314:
```
type St = s:St_ | (forall o | o in s :: o.i(this))
type St_ = map<nat,Ob>
type Ob {
    predicate i(s:St_)
}
```

MikaelMayer:
This PR fixes #2068
I added the corresponding test. The case of a "this" inside a subset type is not allowed, but there was no error mechanism to detect it. This PR solves that problem.

### DIFF
--- a/Source/DafnyCore/Resolver.cs
+++ b/Source/DafnyCore/Resolver.cs
@@ -2750,9 +2750,7 @@ namespace Microsoft.Dafny {
               if (field.Rhs != null) {
                 var ec = reporter.Count(ErrorLevel.Error);
                 scope.PushMarker();
-                if ((currentClass is ClassDecl cd && !cd.IsDefaultClass) || currentClass is DatatypeDecl ||
-                    currentClass is OpaqueTypeDecl) {
-                } else {
+                if (currentClass == null || !currentClass.AcceptThis) {
                   scope.AllowInstance = false;
                 }
                 ResolveExpression(field.Rhs, resolutionContext);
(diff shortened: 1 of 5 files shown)


Source/DafnyCore/Resolver.cs
```diff
@@ -14875,7 +14874,11 @@ public ResolutionContext(ICodeContext codeContext, bool isTwoState)
         if (currentClass is ClassDecl cd && cd.IsDefaultClass) {
```
robin-aws, line 14874: It looks like this case would also hit a crash, just later on in the pipeline?

Test/git-issues/git-issue-2068.dfy
```diff
@@ -0,0 +1,8 @@
+// RUN: %dafny_0 /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+type St = s:St_ | (forall o | o in s :: o.i(this))
```
robin-aws, line 4: As per my other comment, it would be great to show this fix working correctly in a few other illegal contexts as well, just for deeper testing coverage.

robin-aws (changes requested):

Source/DafnyCore/Resolver.cs
MikaelMayer: How would you reproduce it?

Test/git-issues/git-issue-2068.dfy
MikaelMayer: I added two other illegal context, the witness clause and a static const RHS. The static const RHS was also crashing Boogie. I was able to ensure the fixes use the same convention as for the existing code.
Of the top-level contexts, there is no other case to my knowledge where "this" should be disallowed.

Source/DafnyCore/Resolver.cs
robin-aws: I didn't have any luck, and it looks like the error immediately above should always be triggered, which seems to prevent any further problems.

robin-aws (dismissed):

robin-aws (approved):
Nice, I like the refactoring to make this a more explicitly modeled restriction.

PR #2997, opened 2022-11-04
feat: Dafny build and run for Java now use the CLASSPATH env. var.

### LINKED ISSUE #2996: [Feature request]: The dafny build and run commands for Java should use the CLASSPATH env. variable
### What is the feature you would like to see in a future version of Dafny?

In compiling and running a compiled-to-Java Dafny program, the internal invocation of javac and java should use the value of the external env. var. CLASSPATH to enable calling Java methods that are in jar files (or elsewhere on a class path).

davidcok:
Fixes #2996


Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -2334,7 +2334,10 @@ private class GenericArrayElementLvalue : ILvalue {
 
     protected string GetClassPath(string targetFilename) {
       var targetDirectory = Path.GetFullPath(Path.GetDirectoryName(targetFilename));
-      return "." + Path.PathSeparator + targetDirectory + Path.PathSeparator + Path.Combine(targetDirectory, "DafnyRuntime.jar");
+      var classpath = Environment.GetEnvironmentVariable("CLASSPATH");
+      // Note that the items in the CLASSPATH must habs absolute paths because the compilation is performed in a subfolder of where the command-line is executed
+      var cp = classpath != null && classpath.Length != 0 ? (Path.PathSeparator + classpath) : "";
+      return "." + Path.PathSeparator + targetDirectory + Path.PathSeparator + Path.Combine(targetDirectory, "DafnyRuntime.jar") + cp;
```
keyboardDrummer, line 2340: Have you considered using `string.Join(Path.PathSeparator, ".", targetDirectory, Path.Combine(targetDirectory, "DafnyRuntime.jar"), classpath)` instead?

davidcok: OK - though it is slightly more complicated because one component is optional.

keyboardDrummer: > OK - though it is slightly more complicated because one component is optional.

Woops, sorry. Maybe `string.Join(Path.PathSeparator, new [] { ".", targetDirectory, Path.Combine(targetDirectory, "DafnyRuntime.jar"), classpath }.Where(s => !string.IsNullOrEmpty(s)))`

davidcok: Well - that verges on unreadable or not-quickly-understandable to me. Can we stick with the if-then-else?

keyboardDrummer: What makes it hard to read? I feel extracting a variable is an easy way to make long expression easier to read:
```c#
var paths = new [] { ".", targetDirectory, Path.Combine(targetDirectory, "DafnyRuntime.jar"), classpath };
return string.Join(Path.PathSeparator, paths.Where(s => !string.IsNullOrEmpty(s)));
```

But you can also do an if-then if you prefer:
```c#
var paths = new List<string> { ".", targetDirectory, Path.Combine(targetDirectory, "DafnyRuntime.jar") };
if (!string.IsNullOrEmpty(classpath)) {
  paths.Add(classpath);
}
return string.Join(Path.PathSeparator, paths);
```

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -2334,7 +2334,11 @@ private class GenericArrayElementLvalue : ILvalue {
 
     protected string GetClassPath(string targetFilename) {
       var targetDirectory = Path.GetFullPath(Path.GetDirectoryName(targetFilename));
-      return "." + Path.PathSeparator + targetDirectory + Path.PathSeparator + Path.Combine(targetDirectory, "DafnyRuntime.jar");
+      var classpath = Environment.GetEnvironmentVariable("CLASSPATH");
+      // Note that the items in the CLASSPATH must have absolute paths because the compilation is performed in a subfolder of where the command-line is executed
+      return (classpath != null && classpath.Length != 0)
+        ? string.Join(Path.PathSeparator, ".", targetDirectory, Path.Combine(targetDirectory, "DafnyRuntime.jar"), classpath)
+        : string.Join(Path.PathSeparator, ".", targetDirectory, Path.Combine(targetDirectory, "DafnyRuntime.jar"));
```
MikaelMayer, line 2341: Can you please refactor `Path.Combine(targetDirectory, "DafnyRuntime.jar")` in a variable? I see why you have two cases, but since it's the same expression, it would be good to have it defined only once.

cpitclaudel: Empty items are OK in the CLASSPATH, so this code can be simplified to (`(Environment.GetEnvironmentVariable("CLASSPATH") ?? "") + "." + …`

cpitclaudel (changes requested):

keyboardDrummer:
> Empty items are OK in the CLASSPATH, so this code can be simplified to ((Environment.GetEnvironmentVariable("CLASSPATH") ?? "") + "." + …

`null` items are okay for `string.Join`, it translates them to `""`, so just `string.Join(Path.PathSeparator, ".", targetDirectory, Path.Combine(targetDirectory, "DafnyRuntime.jar"), classpath)` should work then.

MikaelMayer (approved):
Perfect ! Glad you found that empty classpaths would not cause problem.

Source/DafnyCore/Compilers/Compiler-java.cs
davidcok: I edited so that all these implicit behaviors are used and the code simplified (and comments added)

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -2334,7 +2334,10 @@ private class GenericArrayElementLvalue : ILvalue {
 
     protected string GetClassPath(string targetFilename) {
       var targetDirectory = Path.GetFullPath(Path.GetDirectoryName(targetFilename));
-      return "." + Path.PathSeparator + targetDirectory + Path.PathSeparator + Path.Combine(targetDirectory, "DafnyRuntime.jar");
+      var classpath = Environment.GetEnvironmentVariable("CLASSPATH") ?? "";
+      // Note that the items in the CLASSPATH must have absolute paths because the compilation is performed in a subfolder of where the command-line is executed
+      // Java ignores empty (oand other ill-formed) classpath entries
```
cpitclaudel, line 2339: Typo: `oand`
But I would just remove this line; it's not ill-formed, it just is the same as `.`

cpitclaudel (approved):

Source/DafnyCore/Compilers/Compiler-java.cs
davidcok: Done - though I could not find any documentation that an empty classpath *item* was interpreted as ".", just that anything not a file or directory was ignored.

PR #3000, opened 2022-11-05
`dafny run` cleanup, take 2

cpitclaudel:
This is a redo of https://github.com/dafny-lang/dafny/pull/2879, which was reverted due to failing tests on Windows.
The first commit is as in that PR; the second one is further clean ups (the commit message has all the details.)

### DIFF
--- a/Source/DafnyCore/Compilers/Compiler-js.cs
+++ b/Source/DafnyCore/Compilers/Compiler-js.cs
@@ -13,7 +13,6 @@ using System.IO;
 using System.Diagnostics.Contracts;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
-using System.Threading.Tasks;
 using JetBrains.Annotations;
 
 namespace Microsoft.Dafny.Compilers {
@@ -2481,8 +2480,8 @@ namespace Microsoft.Dafny.Compilers {
         CreateNoWindow = true,
         UseShellExecute = false,
         RedirectStandardInput = true,
-        RedirectStandardOutput = true,
-        RedirectStandardError = true,
+        RedirectStandardOutput = false,
+        RedirectStandardError = false,
       };
 
       try {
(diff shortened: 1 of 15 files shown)


Source/DafnyCore/Compilers/Compiler-js.cs
```diff
@@ -2497,7 +2491,7 @@ protected class ClassWriter : IClassWriter {
         }
         nodeProcess.StandardInput.Flush();
         nodeProcess.StandardInput.Close();
-        return 0 == RunProcess(nodeProcess, outputWriter);
+        return 0 == WaitForExit(nodeProcess, outputWriter);
```
davidcok, line 2494: Why does Javascript use WaitForExit and not the other compilers?

Test/git-issues/git-issue-2733.dfy
```diff
@@ -0,0 +1,12 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %dafny_0 /noVerify /compile:4 /compileTarget:cs "%s" >> "%t"
```
keyboardDrummer, line 2: Is it possible to use the new CLI UI here?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -64,6 +65,39 @@ public abstract class SinglePassCompiler : Plugins.Compiler {
...
+    }
+
+    public int RunProcess(ProcessStartInfo psi, TextWriter outputWriter, string errorMessage = null) {
+      try {
+        var process = Process.Start(psi);
+        if (process == null) {
+          return -1;
+        }
+        return WaitForExit(process, outputWriter, errorMessage);
+      } catch (System.ComponentModel.Win32Exception e) {
```
keyboardDrummer, line 94: What happens if the process fails to start on a non-Windows machine, does it just return -1 then? Shouldn't the branch on line 91 and on line 95 have similar behavior?

keyboardDrummer (approved):

Source/DafnyCore/Compilers/Compiler-js.cs
cpitclaudel: Because it starts the process itself, whereas for the other ones the process is started by `RunProcess`

Source/DafnyCore/Compilers/SinglePassCompiler.cs
cpitclaudel: No, it throws a Win32 exception on all platforms, which is a bit odd (at least that's what it's documented to do).

cpitclaudel: System.ComponentModel.Win32Exception

keyboardDrummer (approved):

keyboardDrummer (approved):

Test/git-issues/git-issue-2733.dfy
cpitclaudel: Oh snap, I miss this comment.
Given that CI takes multiple hours I'd prefer to do it in a separate PR; OK with you?

keyboardDrummer: That'd be great. We can also do it later in one big fell swoop

Issue #2582, opened 2022-08-11
Incorrect printing of generically-typed strings in Go (a.k.a. should `string` != `seq<char>`?)


robin-aws:
```dafny
method Main() {
  var s := "!dlroW olleH";
  PrintReversed(s);
}

method PrintReversed<T>(s: seq<T>) {
  var reversed := Reverse(s);
  print reversed, "\n";
}

function method Reverse<T>(s: seq<T>): seq<T> {
  // Not s == [] to avoid hitting https://github.com/dafny-lang/dafny/issues/2581 :)
  if |s| == 0 then [] else Reverse(s[1..]) + [s[0]]
}
```

This prints `Hello, World!` in every target language except for Go, which prints `[H, e, l, l, o,  , W, o, r, l, d, !]`.

The issue is that `string` is just an alias for `seq<char>` in Dafny, but we have a special case of implicit parametric type specialization in Dafny, where `print`-ing a string should look different than a generic sequence.

Each backend implements this in almost completely unique ways:

* C# relies on the fact that generic types are reified by checking `if (typeof(T) == typeof(char))`
* Java relies on explicitly passed `TypeDescriptor` values
* Javascript actually instantiates a native `string` instead of a `dafny.Seq` (leading to [other issues](https://github.com/dafny-lang/dafny/issues/2581))
* Go relies on compiler-inserted hints that a value is known statically to be a `seq<char>` (which is why it fails on this example)
* Python dynamically checks if every element of a sequence is a character (specifically a 1-element `str` value), which is the only reason it passes this example.
* The C++ runtime uses template specialization to special-case `operator<<` for a `DafnySequence<char>&`.

@RustanLeino has suggested that one solution would be to align on Go's approach, such that a sequence value only prints like a string if the static type of the value is `seq<char>`. This would be a step backwards in some cases, but seems reasonable given that `print` is much more of a debugging tool than a rigorous I/O feature, and hence best-effort support that is 100% consistent and simple to implement in each runtime is appealing.

Another option is to make `string` a distinct type from `seq<char>`, given that we are already planning on making breaking changes related to strings to provide better Unicode support (https://github.com/dafny-lang/rfcs/pull/13). At a minimum we could have `string` act like a `newtype` based on `seq<char>`, so that it would support all the same sequence operations but requiring `as` casting to be used as a `seq<char>`. 

[...]

PR #3016, opened 2022-11-08
feat: Unicode strings (--unicode-char)

### LINKED ISSUE #413: Revisit definition of Dafny strings (RFC #12)

### LINKED ISSUE #2928: Null character escape not encoded correctly in most backends
This program is a doozy, it ONLY runs without crashing on C#:

```dafny
method Main() {
  var s := "\03";
  assert s[0] == 0 as char;
  assert s[1] == '3';

  expect s[0] == 0 as char;
  expect s[1] == '3';
}
```

Most backends are not escaping the `\0` escape properly, and hence trying to interpret `\03` an octal escape. Go attempts to handle this by padding to `\003` which is a valid Go escape sequence, but in so doing changes the value of the string!

robin-aws:
Resolves #413. Fixes #2928.

Cutting for initial review while I address a few lingering test failures and add a bit more testing coverage. Also depends on #2976 to be fixed for the tests to pass consistently across platforms.

I've documented some of the less obvious compilation strategy decisions in `docs/Compilation/StringsAndChars.md`.

TODO:
 * Reference manual update
 * Release note
 * Address #2928 
 * Ensure characters are escaped when printed (e.g. `'\n'`)


Test/comp/UnicodeStrings.dfy
```diff
@@ -0,0 +1,166 @@
...
+newtype uint16 = x: int | 0 <= x < 0x1_0000 
+newtype uint32 = x: int | 0 <= x < 0x1_0000_0000 
+newtype uint64 = x: int | 0 <= x < 0x1_0000_0000_0000_0000 
+newtype int8 = x: int | -0x80 <= x < 0x80
+newtype int16 = x: int | -0x8000 <= x < 0x8000
+newtype int32 = x: int | -0x8000_0000 <= x < 0x8000_0000
+newtype int64 = x: int | -0x8000_0000_0000_0000 <= x < 8000_0000_0000_0000
+
+// TODO: Test cases for escapes in character literals
+// TODO: Remove token parameter in compilers if not needed
```
RustanLeino, line 13: Somewhere, also include tests that
* check that `\U10000` is parsed correctly (it may not be with the code above)
* check that `\U{E000} - 'a'` gives an error (it may not be with the code above)
* check that `\U{D000} + \U{1000}` gives an error

RustanLeino (changes requested):

Source/DafnyCore/Coco/Scanner.frame
```diff
@@ -235,6 +241,11 @@ public class UTF8Buffer: Buffer {
       int c2 = ch & 0x3F;
       ch = (c1 << 6) | c2;
     }
+    if (0x10000 < ch) {
+      ch = ch - 0x10000;
```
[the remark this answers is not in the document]
robin-aws: Good catch, and agreed with your other comment about that being an important boundary case to test in general (I want to generalize the testing to more of a matrix approach so we test every boundary case thoroughly)

Source/DafnyCore/DafnyPrelude.bpl
```diff
@@ -116,18 +116,30 @@ function {:identity} LitReal(x: real): real { x } uses {
 // -- Characters -------------------------------------------------
 // ---------------------------------------------------------------
 
+#if UNICODE_CHAR
+function char#IsChar(n: int): bool {
+  (0                  <= n && n < 55296   /* 0xD800 */) || 
+  (57344 /* 0xE000 */ <= n && n < 1114112 /* 0x11_0000 */ )
+}
+#endif
+#if !UNICODE_CHAR
```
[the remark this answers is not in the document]
robin-aws: Cool, TIL!

Source/DafnyCore/Options/UnicodeCharactersOption.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <summary>
+/// This is an example of how to add an option that works for both the legacy and the new CLI UI.
+/// </summary>
+public class UnicodeCharactersOption : BooleanOption, ILegacyOption {
+  public static readonly UnicodeCharactersOption Instance = new();
+
+  public override string LongName => "unicode-char";
+  public override string ShortName => null;
+  public string LegacyName => "unicodeChar";
+  // TODO: ???
```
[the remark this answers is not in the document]
robin-aws: This was a note to myself to figure out the right category for the option. I just figured out that this corresponds to the section headers in the older `/help` output, so "Language feature selection" is the clear winner - will update.

Source/DafnyCore/Util.cs
```diff
@@ -187,11 +187,42 @@ public static class Util {
       }
     }
 
+    public static void ValidateEscaping(IToken t, string s, bool isVerbatimString, Errors errors) {
+      if (UnicodeCharactersOption.Instance.Get(DafnyOptions.O)) {
+        foreach (var token in TokensWithEscapes(s, isVerbatimString)) {
+          if (token.StartsWith("\\u")) {
+            errors.SemErr(t, "\\u escape sequences are not permitted when Unicode chars are enabled");
```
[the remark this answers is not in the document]
robin-aws: I originally wanted to allow that, but the consistent feedback from the RFC is that it just confused the distinction between the old and new meaning of characters, so I'm convinced it's better not to: https://github.com/dafny-lang/rfcs/pull/13

Source/DafnyCore/Util.cs
```diff
@@ -201,15 +232,66 @@ public static class Util {
...
+    public static string ReplaceNullEscapesWithCharacterEscapes(string s) {
+      return ReplaceTokensWithEscapes(s, NullEscape, match => "\\u0000");
+    }
+
     /// <summary>
     /// Returns the characters of the well-parsed string p, replacing any
     /// escaped characters by the actual characters.
+    /// 
+    /// It also converts surrogate pairs to their equivalent code points
+    /// if --unicode-char is enabled - these are sythesized by the parser when
```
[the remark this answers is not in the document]
robin-aws: Thanks. BTW for typos/small obvious fixes like this, the "Add a suggestion" button is super convenient! https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request

Source/DafnyCore/Util.cs
```diff
@@ -231,7 +313,13 @@ public static class Util {
             case @"\r": yield return '\r'; break;
             case @"\t": yield return '\t'; break;
             case { } when s.StartsWith(@"\u"):
-              yield return (char)Convert.ToInt32(s[2..], 16);
+              yield return Convert.ToInt32(s[2..], 16);
+              break;
+            case { } when s.StartsWith(@"\U"):
+              yield return Convert.ToInt32(s[3..^1], 16);
```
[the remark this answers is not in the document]
robin-aws: That's right, but the precondition documented on the function by "the well-parsed string p" implies we can assume it, since the parser invokes `ValidateEscaping`. 

It's a bit tempting to introduce a mini AST-like data structure for string content with escapes, to introduce a bit more type safety into this approach...

Source/DafnyCore/Verifier/Translator.ExpressionWellformed.cs
```diff
@@ -844,7 +844,9 @@ private class WFOptions {
               var e0 = FunctionCall(expr.tok, "char#ToInt", Bpl.Type.Int, etran.TrExpr(e.E0));
               var e1 = FunctionCall(expr.tok, "char#ToInt", Bpl.Type.Int, etran.TrExpr(e.E1));
               if (e.ResolvedOp == BinaryExpr.ResolvedOpcode.Add) {
-                builder.Add(Assert(GetToken(expr), Bpl.Expr.Lt(Bpl.Expr.Binary(BinaryOperator.Opcode.Add, e0, e1), Bpl.Expr.Literal(65536)), new PODesc.CharOverflow()));
+                builder.Add(Assert(GetToken(expr),
+                  FunctionCall(Token.NoToken, BuiltinFunction.IsChar, null,
+                    Bpl.Expr.Binary(BinaryOperator.Opcode.Add, e0, e1)), new PODesc.CharOverflow()));
```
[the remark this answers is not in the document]
robin-aws: Excellent catch!

Source/DafnyRuntime/DafnyRuntime.go
```diff
@@ -459,7 +508,16 @@ func SeqOfChars(values ...Char) Seq {
   return Seq{arr, true}
 }
 
+func SeqOfCodePoints(values ...CodePoint) Seq {
```
cpitclaudel, line 511: Isn't this the same as `SeqOf` and `SeqOfChars` above?

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -3491,7 +3505,7 @@ private class SpecialNativeType : UserDefinedType {
       } else {
         Contract.Assert(ct is SeqType);  // follows from precondition
         if (ct.Arg.IsCharType) {
-          wr.Write("_dafny.SeqOfChars");
+          wr.Write(UnicodeChars ? "_dafny.SeqOfCodePoints" : "_dafny.SeqOfChars");
```
cpitclaudel, line 3508: What does this branch correspond to?  Do we really need the `SeqOfCodePoints` or could we just fall through to `SeqOf` below in the unicode case?

cpitclaudel, line 3508: Also, with literals, shouldn't we be able to write `[]rune(<literal>)`?

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -658,6 +658,14 @@ protected class ClassWriter : IClassWriter {
       return TypeName(type, wr, tok, boxed, false, member);
     }
 
+    private string CharTypeName(bool boxed) {
+      if (UnicodeChars) {
+        return boxed ? "dafny.CodePoint" : "int";
```
cpitclaudel, line 663: Why not `rune` or `int32`?

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -1006,9 +1022,14 @@ protected class ClassWriter : IClassWriter {
       } else if (e.Value is bool value) {
         wr.Write(value ? "true" : "false");
       } else if (e is CharLiteralExpr) {
-        wr.Write($"'{(string)e.Value}'");
+        var v = (string)e.Value;
+        if (UnicodeChars && Util.MightContainNonAsciiCharacters(v, false)) {
+          wr.Write($"{Util.UnescapedCharacters(v, false).Single()}");
```
cpitclaudel, line 1027: Even it it contains non-ascii, can't we still use `'…'` in Go?

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -1150,18 +1171,18 @@ protected class ClassWriter : IClassWriter {
       if (type != null && type.AsArrayType != null) {
         arrays.Add(type.AsArrayType.Dims);
       }
-      if (type.IsDatatype && type.AsDatatype is TupleTypeDecl tupleDecl) {
+      if (type != null && type.IsDatatype && type.AsDatatype is TupleTypeDecl tupleDecl) {
```
cpitclaudel, line 1174:
```suggestion
      if (type?.AsDatatype is TupleTypeDecl tupleDecl) {
```

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -1150,18 +1171,18 @@ protected class ClassWriter : IClassWriter {
...
+      if (type != null && type.IsTypeParameter) {
         EmitSuppression(wr);
       }
       wr.Write("{0} {1}", type != null ? TypeName(type, wr, tok) : "Object", name);
       if (leaveRoomForRhs) {
         Contract.Assert(rhs == null); // follows from precondition
       } else if (rhs != null) {
         wr.WriteLine($" = {rhs};");
-      } else if (type.IsIntegerType) {
+      } else if (type != null && type.IsIntegerType) {
```
cpitclaudel, line 1185: When is type null here?

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -2136,15 +2168,19 @@ private class GenericArrayElementLvalue : ILvalue {
...
-                        arg.Type.AsSeqType.Arg.IsCharType;
         bool isGeneric = arg.Type.AsSeqType != null &&
                          arg.Type.AsSeqType.Arg.IsTypeParameter;
-        if (isString) {
+        if (arg.Type.IsStringType) {
           TrParenExpr(arg, wr, false, wStmts);
           wr.Write(".verbatimString()");
-        } else if (isGeneric) {
+        } else if (arg.Type.IsCharType) {
+          if (UnicodeChars) {
```
cpitclaudel, line 2177: Is the condition here that the Dafny char type be boxed? (Which happens in the unicode case)?

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -3679,28 +3716,73 @@ private class GenericArrayElementLvalue : ILvalue {
...
+      if (from.IsArrayType && to.IsArrayType) {
+        var dims = from.AsArrayType.Dims;
+        Contract.Assert(dims == to.AsArrayType.Dims);
+        if (dims > 1) {
+          return false;
+        }
+
+        var udtFrom = (UserDefinedType)from.NormalizeExpand();
+        var udtTo = (UserDefinedType)to.NormalizeExpand();
+        return udtFrom.TypeArgs[0].IsTypeParameter && !udtTo.TypeArgs[0].IsTypeParameter;
```
cpitclaudel, line 3736: I don't know this part of the code well.  I think @fabiomadge understands the downcast stuff better than I do

Source/DafnyCore/Compilers/Compiler-js.cs
```diff
@@ -1317,12 +1341,21 @@ protected class ClassWriter : IClassWriter {
       } else if (e.Value is bool) {
         wr.Write((bool)e.Value ? "true" : "false");
       } else if (e is CharLiteralExpr) {
-        var v = (string)e.Value;
-        wr.Write("'{0}'", v == "\\0" ? "\\u0000" : v);  // JavaScript doesn't have a \0
+        var escaped = TranslateEscapes((string)e.Value);
+        if (UnicodeChars) {
+          wr.Write($"new _dafny.CodePoint('{escaped}'.codePointAt(0))");
```
cpitclaudel, line 1346: It's pretty magical that this works in JS

Source/DafnyCore/Compilers/Compiler-js.cs
```diff
@@ -1317,12 +1341,21 @@ protected class ClassWriter : IClassWriter {
...
+        } else {
+          wr.Write($"'{escaped}'");
+        }
       } else if (e is StringLiteralExpr) {
         var str = (StringLiteralExpr)e;
-        // TODO: the string should be converted to a Dafny seq<char>
-        TrStringLiteral(str, wr);
+        if (UnicodeChars) {
+          wr.Write($"_dafny.Seq.UnicodeFromString(");
```
cpitclaudel, line 1353: Let's try to give the this the same name in every backend?

Source/DafnyCore/Compilers/Compiler-js.cs
```diff
@@ -1919,7 +1952,7 @@ protected class ClassWriter : IClassWriter {
 
     bool IsDirectlyComparable(Type t) {
       Contract.Requires(t != null);
-      return t.IsBoolType || t.IsCharType || AsNativeType(t) != null || t.IsRefType;
+      return t.IsBoolType || (t.IsCharType && !UnicodeChars) || AsNativeType(t) != null || t.IsRefType;
```
cpitclaudel, line 1955: This is because we translate to a CodePoint object, right?
But then I wonder: why don't we translate to a native int instead of an object?

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -853,14 +865,21 @@ private class ClassWriter : IClassWriter {
...
+      wr.Write($"\"{ErrorReporter.TokenToString(tok)}: \" + ");
+
+      if (UnicodeChars && messageExpr.Type.IsStringType) {
+        TrParenExpr(messageExpr, wr, false, wStmts);
+        wr.Write(".VerbatimString(False)");
+      } else {
+        wr.Write($"{DafnyRuntimeModule}.string_of(");
+        TrExpr(messageExpr, wr, false, wStmts);
+        wr.WriteLine(")");
       }
```
cpitclaudel, line 880: Let's share this code with `EmitPrintStmt` above

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1557,7 +1599,11 @@ private class ClassWriter : IClassWriter {
         if (e.ToType.IsNumericBased(Type.NumericPersuasion.Real)) {
           (pre, post) = ($"{DafnyRuntimeModule}.BigRational(", ", 1)");
         } else if (e.ToType.IsCharType) {
-          (pre, post) = ("chr(", ")");
+          if (UnicodeChars) {
+            (pre, post) = ($"{DafnyRuntimeModule}.CodePoint(chr(", "))");
```
cpitclaudel, line 1603: Why do we need a special codepoint type in Python?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1872,9 +1881,9 @@ protected class NullClassWriter : IClassWriter {
 
       foreach (MemberDecl memberx in c.Members) {
         var member = (memberx as Function)?.ByMethodDecl ?? memberx;
-        if (!member.IsStatic) {
-          thisContext = c;
-        }
+        // if (!member.IsStatic) {
+        thisContext = c;
+        // }
```
cpitclaudel, line 1886: ?

cpitclaudel (commented):
I've read through the compilers, and the approach looks reasonable to me.  My only concern is whether we need a (reference) type to hold codepoints rather than just using a native `int`.  IOW, why can't we define `char` to be a `newtype` with 16 bits in the non-unicode mode and 32 bits in the `unicode` type?  Is it to make it easier to track which sequence is a string and which is not?

Source/DafnyCore/AST/Types.cs
```diff
@@ -269,6 +269,7 @@ public abstract class Type {
   public bool IsRealType { get { return NormalizeExpand() is RealType; } }
   public bool IsBigOrdinalType { get { return NormalizeExpand() is BigOrdinalType; } }
   public bool IsBitVectorType { get { return AsBitVectorType != null; } }
+  public bool IsStringType => AsSeqType != null && AsSeqType.Arg.IsCharType;
```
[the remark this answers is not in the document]
robin-aws: Agreed actually, this is just what Rider suggested I change it to :)

Source/DafnyCore/Compilers/Compiler-cpp.cs
```diff
@@ -96,6 +97,11 @@ public class CppCompiler : SinglePassCompiler {
     protected override string ClassAccessor => "->";
 
     protected override void EmitHeader(Program program, ConcreteSyntaxTree wr) {
+      // This seems to be a good place to check for unsupported options
+      if (UnicodeChars) {
+        throw new UnsupportedFeatureException(program.GetFirstTopLevelToken(), Feature.UnicodeChars);
```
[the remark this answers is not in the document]
robin-aws: There's `--unicode-char:false` instead. We could independently consider supporting the `--no-...` idiom for all boolean flags though.

Source/DafnyCore/Compilers/Compiler-cpp.cs
```diff
@@ -2279,9 +2298,8 @@ protected class ClassWriter : IClassWriter {
         if (e.ToType.IsNumericBased(Type.NumericPersuasion.Real)) {
           throw new UnsupportedFeatureException(e.tok, Feature.RealNumbers);
         } else if (e.ToType.IsCharType) {
-          wr.Write("_dafny.Char(");
+          wr.Write("(char)");
```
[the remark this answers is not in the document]
robin-aws: The C++ compiler is completely rejecting `--unicode-char` for now, so there's no such support implemented yet. These changes are just me fixing other incidental bugs I found before deciding it didn't make sense to support it yet (many like this one, where the code was just copied from another compiler and it's always been emitting bad code). :)

Source/DafnyCore/Compilers/Compiler-java.cs
robin-aws: Java doesn't have either, or any other such specialized type. :(

Source/DafnyCore/Compilers/Compiler-java.cs
robin-aws: This is the Java compiler :)

Source/DafnyCore/Compilers/Compiler-java.cs
robin-aws: It's used to represent the native java.lang.Object type. There's a big comment on `NativeObjectType`, which is what I should be using here instead of `null` :)

Source/DafnyCore/Compilers/SinglePassCompiler.cs
robin-aws: Ah part of a different attempt to fix the hackery around coercing the result from a generic method, will revert.

Source/DafnyCore/Compilers/Compiler-go.cs
```diff
@@ -2821,7 +2835,7 @@ private class SpecialNativeType : UserDefinedType {
       var initWr = EmitCoercionIfNecessary(fromType, toType, expr.tok, wr);
       TrExpr(expr.Initializer, initWr, inLetExprBody, wStmts);
       wr.Write(")");
-      if (fromType.Result.IsCharType) {
+      if (fromType.Result.IsCharType && !UnicodeChars) {
         wr.Write(".SetString()");
```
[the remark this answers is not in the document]
robin-aws: Tags the sequence as actually being a string at runtime, which we don't want in --unicode-char mode. Added a comment.

Source/DafnyCore/Compilers/Compiler-go.cs
robin-aws: Oh good call, `SeqOfCodePoints` is indeed pointless. :) Removed from the runtime too.

robin-aws: We do something like that for literals, yes, but this is a collection display, so at best something like `['H', 'e', 'l', 'l', 'o']`. I don't think it's worth special-casing a sequence of characters into a string literal.

Source/DafnyCore/Compilers/Compiler-java.cs
robin-aws: Yup, it's boxing the raw `int` value as a `CodePoint`, which will print the right thing in its `toString()` method. But I realize now I should have put an explicit call to `toString()`, instead of assuming the calling context will.

But there's an even better and more efficient way: I can just call `Helpers.ToCharLiteral(value)` instead :)

Source/DafnyCore/Compilers/Compiler-java.cs
robin-aws: FYI it's not really related to the DowncastClone concept, it's a Java-specific problem involving having to manually box/unbox between an `int` and a `CodePoint` when interfacing with generics.

Source/DafnyCore/Compilers/Compiler-js.cs
robin-aws: Yeah it's the thing that both JS and Python do where there's no character type, only strings of length one. It looks a little weird but I don't know of a better way to structure this.

Source/DafnyCore/Compilers/Compiler-js.cs
robin-aws: I'm going to leave that out of scope for now since the backends are already really inconsistent. But integrating the dafny-runtime-in-dafny implementation of `seq<T>` will force more consistency by necessity at least. :)

Source/DafnyCore/Compilers/Compiler-python.cs
robin-aws: Good call, it's interesting that only the Java compiler has `EmitToString` when both `print` and `expect` implicitly depend on that operation. I might even define that for all compilers.

robin-aws: No that's too aggressive, there's too much of the "best effort to track strings at runtime" logic in the other compilers.

Source/DafnyRuntime/DafnyRuntime.go
robin-aws: Yup, as per the other comment I nuked this

Source/DafnyCore/Coco/Scanner.frame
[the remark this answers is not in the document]
robin-aws: Turns out Coco was also using `0x1_0000` as an EOF marker too, so extra glad it's an explicit test case now. :)

robin-aws:
> I've read through the compilers, and the approach looks reasonable to me. My only concern is whether we need a (reference) type to hold codepoints rather than just using a native `int`. IOW, why can't we define `char` to be a `newtype` with 16 bits in the non-unicode mode and 32 bits in the `unicode` type? Is it to make it easier to track which sequence is a string and which is not?

Just to close the loop, this is justified in `docs/Compilation/StringsAndChars.md` (which it turns out Clement had accidentally missed :) ), with more background in https://github.com/dafny-lang/dafny/issues/2582.

Test/comp/UnicodeStrings.dfy
robin-aws: * `\U1_0000` is now a boundary case (along with several others) in UnicodeStrings.dfy
* `Test/unicodechars/dafny0/Char.dfy` now includes the missing verification cases for hitting the surrogate range

Source/DafnyCore/Coco/Scanner.frame
```diff
@@ -205,10 +205,16 @@ public class Buffer {
...
   public override int Read() {
     int ch;
+    if (pendingLowSurrogate != -1) {
+      ch = pendingLowSurrogate;
+      pendingLowSurrogate = -1;
+      return ch;
+    }
     do {
       ch = base.Read();
       // until we find a utf8 start (0xxxxxxx or 11xxxxxx)
```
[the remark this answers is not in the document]
robin-aws: Yeah, we didn't add the UTF8 parsing logic, that was always part of the default `.frame` file, but it never handled non-BMP characters and hence the need for the patch.

I'm not sure if using a `Stream` will work but I'd definitely be happy skipping all this non-standard implementation if possible, I will experiment.

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -2136,15 +2187,18 @@ private class GenericArrayElementLvalue : ILvalue {
...
+        if (arg.Type.IsStringType) {
           TrParenExpr(arg, wr, false, wStmts);
           wr.Write(".verbatimString()");
-        } else if (isGeneric) {
+        } else if (arg.Type.IsCharType && UnicodeChars) {
+          wr.Write($"{DafnyHelpersClass}.ToCharLiteral(");
+          TrExpr(arg, wr, false, wStmts);
+          wr.Write(")");
+        } else if (isGeneric && !UnicodeChars) {
+          // This happens to not work when --unicode-char is true anyway,
```
RustanLeino, line 2200: What do you mean "to not work"? Is the code buggy?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -129,6 +129,10 @@ public abstract class SinglePassCompiler : Plugins.Compiler {
     protected string IntSelect = ",int";
     protected string LambdaExecute = "";
 
+    protected static bool UnicodeChars => UnicodeCharactersOption.Instance.Get(DafnyOptions.O);
```
RustanLeino, line 132: Rename this variable. `UnicodeChars` suggests the variable holds a set/array of characters. But it's actually a boolean. How about something like `UnicodeMode` or `InUnicodeMode` or `FullUnicodeSupport` or `UnicodeSupport``?

Source/DafnyCore/Dafny.atg
```diff
@@ -856,15 +858,27 @@ TOKENS
...
+        // as needed after reading the UTF-8 bytes.
+        // But that means character literals can appear to have more than one "character"
+        // in them. Normally we would just allow more than one charChar here and reject
+        // invalid sequences of them later, but because we allow single quotes in the
+        // "ident" rule above, that would then overlap with identifiers.
+        // Instead we only allow a surrogate pair here.
+        | highSurrogate lowSurrogate
         | "\\\'" | "\\\"" | "\\\\" | "\\0" | "\\n" | "\\r" | "\\t"
         | "\\u" hexdigit hexdigit hexdigit hexdigit
+        | "\\U{" hexdigit { ['_'] hexdigit } "}"
```
RustanLeino, line 873: Sweet!

Test/comp/UnicodeStrings.dfy
```diff
@@ -0,0 +1,246 @@
...
+  print trickyString2, "\n";
+
+  // Testing that runtimes don't confuse a seq<uint32> for a string
+  // (which would be a problem if we used Int32 in C# instead of Rune, for example)
+  var s := "Ceci n'est pas une string";
+  var notAString := MapToInt32(s);
+  print notAString, "\n";
+
+  // Ensuring character arithmetic can be compiled
+  var sarcastic := "Oh UNicOdE, tHaT's a REaL usEFuL FEaTuRe!";
```
RustanLeino, line 181: :)

Test/unicodechars/dafny0/Char.dfy
```diff
@@ -0,0 +1,117 @@
+// RUN: %dafny_0 /compile:0 /unicodeChar:1 /print:"%t.print" /dprint:"%t.dprint" "%s" > "%t"
```
RustanLeino, line 1: Should this file (and possibly some of the others) be in `Test/comp` instead? Or, if there are many of them, I suppose they could also go into a new directory like `Test/unicode`. (These are points to consider, not hard suggestions from me.)

docs/DafnyRef/Grammar.md
```diff
@@ -358,13 +350,24 @@ In this section the "\\" characters are literal.
...
 null, new line, carriage return, tab, or a
 Unicode character with given hexadecimal representation.
+Which Unicode escape form is allowed depends on the value of the `--unicode-char` option.
+
+If `--unicode-char:false` is provided,
+`\uXXXX` escapes can be used to specify any UTF-16 code unit.
+
+If `--unicode-char:true` option is provided,
+`\U{X..X}` escapes can be used to specify any Unicode scalar value.
+There must be at least one hex digit in between the braces, and at most six.
```
RustanLeino, line 368: Mention that consecutive digits are optionally separated by one `_`.

docs/DafnyRef/Types.md
```diff
@@ -475,18 +484,26 @@ supported escape sequences are the following:
...
  `\"`               | the character `"`
  `\\`               | the character `\`
- `\0`               | the null character, same as `\u0000`
+ `\0`               | the null character, same as `\u0000` or `\U{0}`
  `\n`               | line feed
  `\r`               | carriage return
  `\t`               | horizontal tab
- `\u`_xxxx_         | [universal (unicode) character](https://en.wikipedia.org/wiki/Universal_Character_Set_characters) whose hexadecimal code is _xxxx_,  where each _x_ is a hexadecimal digit
+ `\u`_xxxx_         | [UTF-16 code unit](https://en.wikipedia.org/wiki/UTF-16) whose hexadecimal code is _xxxx_,  where each _x_ is a hexadecimal digit
+ `\U{`_x..x_`}`     | [Unicode scalar value](https://unicode.org/glossary/#unicode_scalar_value) whose hexadecimal code is _x..x_,  where each _x_ is a hexadecimal digit
```
RustanLeino, line 492: Mention underscores.

RustanLeino (changes requested):

Test/unicodechars/dafny0/Char.dfy
robin-aws: I intended `Test/unicodechars` to imitate the `allocated1` directory that includes several test files from elsewhere under `Test` and reproduces the folder layout to a degree. For some, like this one, I couldn't just include the original because it uses `\u` escapes that are forbidden in `--unicode-char` mode, and I needed to add more test cases specific to that mode.

The real question might be whether I should still consider the main `UnicodeStrings.dfy` test file the `--unicode-char` version of `Test/dafny0/Strings.dfy, and therefore move it to `Test/unicodechars/dafny0/Strings.dfy`. It's obviously fairly different now, but in spirit I think that makes sense.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
robin-aws: Completely agree. Perhaps `UnicodeCharEnabled`, to align as closely as possible with the phrasing of the option?

I'm wary of the fact that we're already abusing the term "Unicode" by saying things like "Full unicode support". UTF-16 is already a completely reasonable Unicode encoding that can cover all code points, and we're miles away from "supporting" all concepts from the entire standard, like collation and extended grapheme clusters.

"`--unicode-char`" should only be interpreted to mean "define the `char` type to be more correct and interoperable with other Unicode implementations".

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -24,6 +24,86 @@ namespace Dafny {
   using System.Collections.Immutable;
   using System.Linq;
 
+  // Similar to Rune, which would be perfect to use
```
RustanLeino, line 27:
```suggestion
  // Similar to System.Text.Rune, which would be perfect to use
```

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -24,6 +24,86 @@ namespace Dafny {
...
+
+    public static bool operator >(Rune left, Rune right) => left._value > right._value;
+
+    public static bool operator >=(Rune left, Rune right) => left._value >= right._value;
+
+    public static explicit operator Rune(int value) => new Rune(value);
+    public static explicit operator Rune(BigInteger value) => new Rune((uint)value);
+
+    public override string ToString() {
+      return char.ConvertFromUtf32(Value);
```
RustanLeino, line 84: Will this produce a Dafny-like literal? Will this `ToString()` method be used in a way that expects that?

Source/DafnyRuntime/DafnyRuntime.py
```diff
@@ -72,16 +72,37 @@ def c_label(name: str = None):
...
     def __init__(self, __iterable = None, isStr = False):
         if __iterable is None:
             __iterable = []
-        self.isStr = isStr \
-                     or isinstance(__iterable, str) \
-                     or (isinstance(__iterable, Seq) and __iterable.isStr) \
-                     or (not isinstance(__iterable, GeneratorType)
-                         and all(isinstance(e, str) and len(e) == 1 for e in __iterable)
-                         and len(__iterable) > 0)
+        if isStr is None:
```
RustanLeino, line 97: Why would `isStr` ever be `None`. Do we not expect it to be passed in as either `False` or `True`? Or is the idea to support this third value in order to give different behavior?

Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/CodePoint.java
```diff
@@ -0,0 +1,62 @@
...
+ * specifically, which corresponds to the Dafny `char` type
+ * when using --unicode-char.
+ */
+public final class CodePoint {
+
+    // Caching a subset of values just like the built-in box types like Character.
+    private static class CodePointCache {
+        private CodePointCache(){}
+
+        static final CodePoint cache[] = new CodePoint[127 + 1];
```
RustanLeino, line 15: I assume there's some reason to think of this size as `127 + 1` (instead of just `128`), but I don't understand what it is. I suggest you add a comment.

RustanLeino (changes requested):

Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/CodePoint.java
robin-aws: Not a good reason :) It's copied almost verbatim from `java.lang.Character`, and the built-in boxing types are a little more general where the limit is configurable for some types, and some handle values less than 0. I've cleaned this up now.

Source/DafnyRuntime/DafnyRuntime.py
robin-aws: The latter, and I added a big chunky docstring to explain what's going on here.

Source/DafnyRuntime/DafnyRuntime.cs
robin-aws: Ah good point, added this comment to clarify:

    // Defined this way to be consistent with System.Text.Rune,
    // but note that Dafny will use Helpers.ToString(rune) instead,
    // which will print in the style of a Dafny character literal.

Source/DafnyCore/Compilers/Compiler-java.cs
robin-aws: Out of date comment, will remove.

Source/DafnyCore/Util.cs
```diff
@@ -201,15 +232,66 @@ public static class Util {
...
-      UnescapedCharacters(s, isVerbatimString).Iter(ch => sb.Append(ch));
+      if (UnicodeCharactersOption.Instance.Get(DafnyOptions.O)) {
+        UnescapedCharacters(s, isVerbatimString).Iter(ch => sb.Append(new Rune(ch)));
+      } else {
+        UnescapedCharacters(s, isVerbatimString).Iter(ch => sb.Append((char)ch));
+      }
       return sb.ToString();
     }
 
+    private static readonly Regex UnicodeEscape = new Regex(@"\\U\{([0-9a-fA-F]+)\}");
```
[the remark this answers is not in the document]
robin-aws: Went ahead with supporting underscores, which makes this moot. I still have the same test after removing underscores.

RustanLeino (approved):
Great job! This is a good feature and a clear stand on the issue of what Dafny strings are.

keyboardDrummer:
Awesome to see this process of careful design followed by implementation!

PR #3033, opened 2022-11-09
Adding arm64-osx architecture to  publish script

### LINKED ISSUE #2889: Update release process to include building a Dafny that's usable on M* Macs.
Many Dafny users have trouble installing Dafny on their Mac:

- https://github.com/dafny-lang/dafny/issues/2869
- https://github.com/dafny-lang/ide-vscode/issues/283
- https://github.com/dafny-lang/ide-vscode/issues/246

Instead of making it easier for these users to build Dafny from source on their M* Mac, which so far has been an error prone process, we should ensure new versions of Dafny include a M* Mac build.

### Implementation A

Use dotnet cross-compilation to let our GitHub actions MacOS machine build for both x64 and arm64. This can be done using the `--arch` option of `dotnet`: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-build#options

### Implementation B

Build a platform independent Dafny, which requires `dotnet` to be installed on the machine, and which does not include a Z3 binary. The Dafny VSCode extension will have to be updated so that it installs `dotnet` and downloads Z3.

### Implementation C

Build a platform independent Dafny, which requires `dotnet` to be installed on the machine, and which includes the MacOS M* Z3 binary. The Dafny VSCode extension will have to be updated so that it installs `dotnet`.

prvshah51:
Fixes #2889 

The packaging script is using Z3 version names to identify and publish Dafny Binaries for various OS and Architecture but as Z3 4.8.5 doesn’t have a version for osx-arm64 , here the changes are copying info from osx-64 and changing it to publish for M* macbooks.


Scripts/package.py
```diff
@@ -215,6 +218,12 @@ def discover(args):
 
         for release_js in js["assets"]:
             release = Release(release_js, args.version, args.out)
+            # Copying assets of osx and chnaging value of "name" to include "arm64" as architecture and then add it to same list of releases. 
+            if release.os_name == "osx":
+                tmp_release_js = release_js
+                tmp_release_js.update({'name': 'z3-4.8.5-arm64-osx-11.0.zip'})
+                tmp_release = Release(tmp_release_js, args.version, args.out)
+                yield tmp_release
```
keyboardDrummer, line 226: I don't get it. Does this release something with arm64 in the name but with the same assets as x64 ?

keyboardDrummer (commented):
Is the aim of this PR to support publishing an arm64 release of Dafny?

If we want to support publishing for arm64, shouldn't we also ask dotnet to build Dafny for arm64, instead of only enabling using an arm64 build of Z3? I'm surprised there are no changes to the dotnet calls in this PR.

Scripts/package.py
prvshah51: The whole release process is based on available Z3 versions, even the publish process.  dotnet publish command is called by run_publish function where value passed to choose architecture is self.platform:
https://github.com/prvshah51/dafny/blob/a38fcfd8ab9d69e40adbef0c784c8aaf7e57a642/Scripts/package.py#L142
 which is determined from the `name` value of these dictionaries:
https://github.com/prvshah51/dafny/blob/a38fcfd8ab9d69e40adbef0c784c8aaf7e57a642/Scripts/package.py#L90

keyboardDrummer: Okay. So will this change will make any `osx` release, in additional to doing an x64 release, also do an arm64 release, both using `z3-4.8.5-x64` ?

prvshah51: Yes, This change adds one more release data to the list of releases which will do an `arm64-osx` release and both will be using `z3-4.8.5-x64`.

Scripts/package.py
```diff
@@ -171,6 +171,9 @@ def pack(self):
         missing = []
         with zipfile.ZipFile(self.dafny_zip, 'w',  zipfile.ZIP_DEFLATED) as archive:
             with zipfile.ZipFile(self.z3_zip) as Z3_archive:
+                #changing directory value for osx-arm64 to copyover files from z3-4.8.5-x64-osx-10.14.2.zip
+                if self.target == "osx-arm64":
+                    self.directory = "z3-4.8.5-x64-osx-10.14.2"
```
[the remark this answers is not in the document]
prvshah51: That's a good point. z3 v4.8.16 is the earliest version when they start releasing for `arm64-osx`. We can aim for that.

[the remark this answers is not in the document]
keyboardDrummer: @atomb do you know if we could move to Z3 4.8.16 as the recommended version?

Also, @prvshah51, I'm confused about what is required for a Dafny release to run on an M* machine. Does it need an arm64 Dafny build, an arm64 Z3 build, both, or neither?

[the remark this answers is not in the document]
prvshah51: I am sure that an arm64 z3 build is not needed, as far as I know a cross-compiled arm64 Dafny build will have the correct version of .NET runtime and will eliminate Dotnet related issues, as till now the `x64` Dafny build was getting installed with `arm64` version of Dotnet runtime.

prvshah51:
M1 machines were able to run x64 binaries sometimes but that also depended on version of .Net runtime, and that setup was significantly slower. Dotnet was quick to launch .Net runtime for arm machines and thus everyone had installed those but were installing Dafny published for X64-OSx platform, So publishing for Arm64-OSx will ensure tha users don't face issues related to Dotnet.

robin-aws (approved):

PR #3042, opened 2022-11-10
Fix: Ghost diagnostics are always up to date

### LINKED ISSUE #3041: Ghost highlighting highlights more than it should
Failing code
------------

```dafny
method Test() {
  ghost var x := 1;
  print 2;
}
```

Steps to reproduce the issue
----------------------------

* Dafny version: 3.9.1.41027
* Dafny VSCode extension version: latest master
* Write the body of the method by hand.

Expected behavior
-----------------
image
This is what you get when you open a file containing this.

Actual behavior
---------------
image
This is what you get if you type the method's body content

MikaelMayer:
Fixes #3041 

What happened was that, if the ghost highlights did not change, no notification was sent to VSCode.
This is problematic, because VSCode believes that text inserted at the end of decorations is part of the decoration itself,
and ghost diagnostics are a decoration. That means that, if you type a non-ghost sentence after a ghost sentence, the non-ghost sentence will be highlighted as ghost.

### DIFF
--- a/Source/DafnyLanguageServer.Test/Synchronization/GhostDiagnosticsTest.cs
+++ b/Source/DafnyLanguageServer.Test/Synchronization/GhostDiagnosticsTest.cs
@@ -16,6 +16,31 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization {
   public class GhostDiagnosticsTest : ClientBasedLanguageServerTest {
 
     [TestMethod]
+    public async Task GhostDiagnosticsShouldBeRepublished() {
+      var source = @"
+method Test() {
+  ghost var x := 2;
+}
+".TrimStart();
+      await SetUp(new Dictionary<string, string>() {
+        { $"{GhostOptions.Section}:{nameof(GhostOptions.MarkStatements)}", "true" }
+      });
+      var documentItem = CreateTestDocument(source);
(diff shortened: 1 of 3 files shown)


keyboardDrummer (commented):
How about solving this on the VSCode side? Changing it on the server side to send ghost diagnostics even if they haven't changed consumes more bandwidth. VSCode can also reset the decoration when text is added after it.

MikaelMayer:
> How about solving this on the VSCode side? Changing it on the server side to send ghost diagnostics even if they haven't changed consumes more bandwidth. VSCode can also reset the decoration when text is added after it or the document is changed.

To solve it on the VSCode side, one would need to update ghost diagnostics to prevent VScode from inferring new bounds.
When would we do this? If the ghost diagnostics did change, we might have a race condition between applying previous ghost diagnostics and accepting the newest ones.

How about a mixed approach, where we send one more type of lightweight notification saying "Hey, reapply the ghost diagnostics because they haven't changed".?

keyboardDrummer:
> "Hey, reapply the ghost diagnostics because they haven't changed".?

How would you explain that API choice without referring to our current VSCode implementation?

MikaelMayer:
Closed in favor of
https://github.com/dafny-lang/ide-vscode/pull/304

PR #3047, opened 2022-11-11
feat: /runAllTests:1 MUST set the exit code on failed tests

### LINKED ISSUE #3043: [Bug]: runAllTests:1 MUST return an exit code
### Dafny version

3.9.0.41003

Any Dafny file with more than one test in it should suffice.

There are two.
First:

```code
dafny \
		-vcsCores:$(CORES) \
		-compileTarget:cs \
		-spillTargetCode:3 \
		-runAllTests:1 \
		-compile:0 \
		-useRuntimeLib \
		-out runtimes/net/tests/TestsFromDafny \
		`find ./test -name '*.dfy'` \
		-library:src/Index.dfy
```

Which will get you `TestsFromDafny.cs`.
Next, we will tell dotnet to execute those tests:
```
dotnet run \
		--project runtimes/net/tests/ \
		--framework net6.0
```

The result will be that every test is run and its result is reported as PASSED or FAILED. 

However, the generated `TestsFromDafny.cs` will exit 0 regardless of the results of the tests. 

This is very bad.

### What happened?

When Dafny is told to:
- `spillTargetCode:3`
- `runAllTests:1`

It generates something like:
```
bool _367_success;
_367_success = true;
...<All the tests in try catch blocks, on catch, set _367_success False>...
if (!(_367_success)) {
  Dafny.Helpers.Print(Dafny.Sequence<char>.FromString(@"Test failures occurred: see above."));
}
```

Which is very groovy.
But, it really needs to also add an exit or return code based on the value of `success`. 
Otherwise, you are running the tests, 
and printing the result of them,
but not exposing the result in a use-able form for CI.

Linux, Windows, Mac

robin-aws:
Fixes #3043.

Replaces just printing a "Test failures occurred: see above." message with an `expect` statement instead, such that if there are test failures the overall program will halt instead of terminate normally.

In addition, ensure that if a program halts when executed through a main method (as opposed to invoked from external code) ensure the exit code of the process is not zero. This could be considered a breaking change, but I think it's very reasonable to call this a strict improvement in behavior if not an outright bug fix instead, as anyone with failing expect statements they are not aware of will almost certainly be glad to discover them ASAP.

In combination, this ensures that when using `/runAllTests:1`, a test failure will cause the exit code to be non-zero.

Also added support for prefixing a LIT command line with `!` to succeed iff the exit code is non-zero. This is already a fairly portable shell syntax and supported by the original LLVM LIT runner.


Source/DafnyCore/Resolver/RunAllTestsMainMethod.cs
```diff
@@ -194,17 +191,14 @@ public class RunAllTestsMainMethod : IRewriter {
     }
 
     // For now just print a footer to call attention to any failed tests.
```
robin-aws, line 193:
```suggestion
    // Use an expect statement to halt if there are any failed tests.
```

cpitclaudel (approved):
👏

PR #3049, opened 2022-11-12
feat: Make sure that `dafny run` always requests UTF-8 encoded output

### LINKED ISSUE #2999: Change `dafny run` to always force UTF-8 output
The `dafny run` command currently makes no effort to control the encoding of its output (what bytes get generated when the user writes `print "foo"`.

Concretely, strings get printed using the default printing of the target language (`System.out.print` in Java, `Console.Write` in C#, etc.).  This leads to inconsistencies in printing, because different language runtimes use different encodings by default: some use UTF-8, and others obey the system’s codepage (`GetConsoleCP` on Windows) or locale (`LANG` / `LC_*` on Linux).  Often the behavior changes depending on whether the code is attached to a console or to a pipe (this is the case for Python and C#, for example)

The problem mostly shows up on Windows, because most Linux distributions default to UTF-8, and so we get the same output whether a language defaults to UTF-8 or to the system’s locale.

Examples of broken output are given in https://github.com/dafny-lang/dafny/issues/2976

Such discrepancies are not necessarily an issue for `dafny build`: its task is to produce code in each of its target languages, and then it’s the user’s responsibility to configure each runtime (maybe?).

But for `dafny run` we own the whole experience, and so we should strive to print consistent output.

There are three ways to do this:

1. Always use UTF-8, on all platforms.
2. Always use UTF-8 on Windows and always respect the system locale on Linux and macOS.  
3. Always use the current codepage or encoding.  This is the “right” thing to do, but it’s also inconsistent with a lot of languages.

Option 1 is “wrong” on all platforms (it ignores the user’s locale / codepage), but it’s very simple to explain to users.  Arguably, it’s also more important for Dafny to be completely predictable than for other languages (?).  The implementation is simple.

Option 2 is “wrong” on Windows, but it’s what some other languages do, so at least we’d be in good company.  The implementation is somewhat complicated, because we’d have to override what some of our languages do (for example C# seems to ignore LANG and use UTF-8 on Linux).  Testing is also hard, because different languages behave differently even for the same locale (Java and Python look at LANG but behave differently for LANG=C, etc.)

Option 3 is the “right” one, but it’s also not what most languages do. As far as I understand, the reason they don’t do this is that users on Windows don’t know about `chcp` and codepages, so they get confused by the non-utf-8-default, and blame application vendors instead of blaming their own configuration and/or their OS.  Even then it would be my preference, it’s tricky to implement, especially in Python (Python 3.7 moved to default-UTF-8 on Windows and added an environment variable to revert to using the current codepage instead, but then Python 3.8 broke that flag; it’s a [known issue](https://github.com/python/cpython/issues/86427)).

So, we will change `dafny run` to always force UTF-8 output (option 1).  There will be no change for now to `dafny build`.

cpitclaudel:
Closes #2999 (but beware of encoding issues in Go and C++).


Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -632,6 +632,7 @@ private record TargetPaths(string Directory, string Filename) {
       Contract.Requires(dafnyProgram != null);
       Contract.Assert(dafnyProgramName != null);
 
+      // TODO: The outputWriter parameter is unused.  Remove it?
```
robin-aws, line 635: Really? I see `outputWriter.WriteLine("Program compiled successfully")` below for e.g., just guarded by `--compile-verbose`

docs/DafnyRef/Statements.md
```diff
@@ -1713,14 +1713,28 @@ explicitly invoke this conversion.
...
+to `print` into standard-output byte strings is left to the runtime of the
+language that the Dafny code is compiled to (some language runtimes use UTF-8 in
+all cases; others obey the current locale or console encoding).
+
+In most cases, the standard-output encoding can be set before running the
+compiled program using language-specific flags or environment variables
+(e.g. `-Dfile.encoding=` for Java).  This is in fact how `dafny run` operates:
+it uses language-specific flags and variables to enforce UTF-8 output regardless
+of the target language (but note that the C++ and Go backends currently have
+limited support for UTF-16 surrogates).
```
robin-aws, line 1736: This is great but I feel it should also be in documentation specifically for the `dafny run` command and not just for the `print` statement. Perhaps http://dafny.org/dafny/DafnyRef/DafnyRef#256-using-dafny-from-the-command-line?

robin-aws (commented):
Whew, nice. I assume Go didn't need any changes because it already enforces UTF-8 output?

Source/DafnyDriver/DafnyDriver.cs
cpitclaudel: Really: specifically, it's never used by *callers*; it always get its default value `null`, which then routes to Console.Out

docs/DafnyRef/Statements.md
cpitclaudel: Will do.

cpitclaudel:
> I assume Go didn't need any changes because it already enforces UTF-8 output?

Well, not really; it's more that Go passes strings through without modification, so as long as we construct UTF-8 we're fine (which we do for since UTF-16 code units, but not for paired ones, hence the commented-out part of the test.

Source/DafnyDriver/DafnyDriver.cs
robin-aws: Oh I see. If you end up having to push any more commits, it would be great to clarify that in the comment. But I wouldn't spend another CI run on it. :)

docs/DafnyRef/UserGuide.md
```diff
@@ -142,7 +142,8 @@ then runs it with the three command-line arguments `1 2 3`
 If building a `.dfy` file containing the `Main` method requires additional
 `.dfy` files, those files can be mentioned in `include` directives in the one `.dfy` file listed in the `dafny run` command. If other kinds of files (e.g., `.dll`, `.jar`) are needed, then use 
 the `--input` option for each extra file or use `dafny build` to build the executable and then run it in an 
-additional step.
+additional step.  
+**Note:** `dafny run` will typically produce the same results as the executables produced by `dafny build`.  The only expected differences are performance (`dafny run` may not optimize as much as `dafny build`), and target-language-specific configuration issues (e.g. encoding issues: `dafny run` sets language-specific flags to request UTF-8 output for the [`print`](#print-encoding) statement in all languages, whereas `dafny build` leaves language-specific runtime configuration to the user).
```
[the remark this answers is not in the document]
cpitclaudel: There are two spaces at the end of the additional step line, so there should be a line break

Source/DafnyDriver/DafnyDriver.cs
cpitclaudel:
```suggestion
      // TODO: outputWriter seems to always be passed in as `null`.  Remove it?
```

robin-aws (approved):
0xF0 0x9F 0x9A 0x80

:)

PR #3051, opened 2022-11-14
Add --verify-included-files and --boogie-filter options

keyboardDrummer:
### Changes

- To all commands that verify, add an option `--verify-included-files` that causes included files to also be verified.
- To the `verify` command, add an option `---boogie-filter=<pattern>` which can be used to verify a subset of proofs. This is useful when working on a proof without using the Dafny IDE to do verification.


cpitclaudel:
Do you have the motivation for this PR written up somewhere?  I'm not sure I follow, and I don't think performance is the reason we currently don't verify everything by default; is it?  Some other reasons might be that different files require different options (this happens if you use the arithmetic stdlib), that you are using the CLI instead of the IDE for verification (we have multiple customers that work this way currently), that you have code written against the libraries and you don't want to have to pass `--library` explicitly, etc.

I think all of these use cases are valid and would suffer from a change of default, so I'm nto convinced it's good to change the default.  `--no-verify-includes` would be good, but you talk about removing it once we have parallel parsing and typechecking.  That's confusing too, because includes currently have to be parsed, so running multiple instances of Dafny side by side does not actually speed things up versus a single instance, right?

There's also the question of what happens when we don't have explicit `include`s anymore: will `dafny verify <file.dfy>` do the same for every file in the project?

As an alternative to consider, maybe we could (once we have project files) have:

- `dafny verify`, equivalent to `dafny verify <project-file.dafnyproj>`: verify all files in the project, but not the libraries; and an option `--verify-all` or `--verify-libraries` to also verify the libraries (I'm flexible on the default there)
- `dafny verify foo.dfy`: Only verify file `foo`.

robin-aws:
I agree that performance should not be the motivation here. The biggest issue is that users are often surprised that files that are only included but not passed explicitly on the CLI are not verified.

We should only remove `--no-verify-includes` if and when we're convinced users have had enough time to migrate away from needing it.

keyboardDrummer:
> Do you have the motivation for this PR written up somewhere?

To make the CLI easier to use by having intuitive defaults (/verifyAllModules=true) and no unneeded options  (remove --no-verify-includes) when it is no longer needed.

> Some other reasons might be that different files require different options

OK. Shouldn't we use module level options for that? I guess `noNLarith` currently can't be used at the module level, but we can change that.

> Some other reasons might be that different files require different options (this happens if you use the arithmetic stdlib)

Can you sketch the setup when using that? Does the consuming project not turn off non-linear arithmetic for its all its files?

> that you are using the CLI instead of the IDE for verification (we have multiple customers that work this way currently)

That use-case is covered by this PR, see the `--filter` option, is it not?

> that you have code written against the libraries and you don't want to have to pass --library explicitly

In the previous scenario you have to pass all your own source files on the CLI, and reference library files X times using includes. In the new one you only have to pass your root files, and the library root files on the CLI, which seems strictly less work.

> you talk about removing it once we have parallel parsing and typechecking. That's confusing too, because includes currently have to be parsed, so running multiple instances of Dafny side by side does not actually speed things up versus a single instance, right?

I think that's true only if there's only a single root file for the Dafny project, one file that includes all other files, but if you have multiple root files then the non-shared dependency trees of those can be parsed in parallel by invoking a separate Dafny instance for each.

cpitclaudel:
> OK. Shouldn't we use module level options for that?

Yes

> Does the consuming project not turn off non-linear arithmetic for its all its files?

The stdlib doesn't, for example.  But, module-level options would be good here too

> That use-case is covered by this PR, see the --filter option, is it not?

Not sure; can do I use --filter to reverify one whole file?

> In the previous scenario you have to pass all your own source files on the CLI. In the new one you only have to pass your root files, and the library root files on the CLI, which seems strictly less work.

I don't think it's easy to determine what root files are (but I'm sure that we can improve this anyway, since we could do exclusions based on e.g. folder or other project files.

> if you have multiple root files then the non-shared dependency trees of those can be parsed in parallel by invoking a separate Dafny instance for each.

Ha, good point!  I hadn't considered that. Though you still reparse the shared dependencies, if any.

> Part of the discussion in this PR is about removing --no-verify-includes but this PR doesn't do that. Shall we have that discussion when we're actually removing it? Conditions might be different then.

It's about the whole design and default change, I think.   You haven't commented on the alternate proposal that I posted above; do you have an opinion about it?

As another alternative, we could have a `--scope` flag to `dafny verify`:

- `dafny verify --scope=project` would check the whole project but not its dependencies
- `dafny verify --scope=all` would check the whole project  and its dependencies
- `dafny verify --scope=file foo.dfy` would verify just `foo.dfy`
- `dafny verify --scope=project foo.dfy` would verify `foo.dfy` and its transitive dependencies restricted to the current project
- `dafny verify --scope=all foo.dfy` would verify `foo.dfy` and all of its transitive dependencies

And then the default could be `--scope=all` if you want.

cpitclaudel:
I'm concerned that we're making the easy case harder.  There are two examples:

- No-IDE verification (verifying a file on the CLI without the IDE).  Dafny has no persistent caching, so re-verifying everything is costly.  Project files will help.
- No-project-file verification.  I often write small one-off scripts that import the libraries; these used to be verified with `dafny foo.dfy`; now it's `dafny verify foo.dfy`, which is longer but OK; and with this PR it will be `dafny verify --no-verify-includes foo.dfy`, which gets unpleasantly long for a simple, common use case.

Once we have project files, I think it could be a good default to reverify everything up to library boundaries.

keyboardDrummer:
Thinking about this more, the choice whether to verify includes or not is relevant only in the situation where no project file is used, because when using project files, one doesn't need to use includes. Doing `dafny verify <projectFile>` will already verify all files in a project regardless of whether it verifies includes or not.

When no project file is used, it seems sensible to only verify the files that are actually passed on the CLI. The verify includes option seems mostly used for Dafny test-cases.

I will updated the PR and description accordingly.

Source/DafnyDriver/Commands/CommandRegistry.cs
```diff
@@ -81,6 +81,7 @@ static class CommandRegistry {
     var wasInvoked = false;
     string optionFailure = null;
     var dafnyOptions = new DafnyOptions();
+    dafnyOptions.VerifyAllModules = true;
```
cpitclaudel, line 84: Why is this here?

Source/DafnyDriver/Commands/FilterOption.cs
```diff
@@ -0,0 +1,24 @@
...
+
+public class FilterOption : CommandLineOption<IEnumerable<string>> {
+  public static readonly FilterOption Instance = new();
+
+  public override object DefaultValue => Enumerable.Empty<string>();
+  public override string LongName => "filter";
+  public override string ArgumentName => "pattern";
+
+  public override string Description => @"
+(experimental) Only check proofs matched by pattern <p>. This option may be specified multiple times to match multiple patterns. The pattern <p> may contain * wildcards which match any character zero or more times.".TrimStart();
```
cpitclaudel, line 15: This is a bit unclear.  Even if its unsatisfactory, we might want to say explicitly that this matches Boogie names?

Source/DafnyDriver/Commands/FilterOption.cs
```diff
@@ -0,0 +1,24 @@
...
+  public override void Parse(CommandLineParseState ps, DafnyOptions options) {
+    throw new System.NotImplementedException();
+  }
+
+  public override string PostProcess(DafnyOptions options) {
+    options.ProcsToCheck.AddRange(Get(options).Select(s => "*" + s + "*"));
```
cpitclaudel, line 21: Do we really want to always add `*`?

cpitclaudel (dismissed):
Thanks!  I have a few questions / suggestions.
I also wonder whether `--filter` is the right name.  Maybe `--verify-only`?

keyboardDrummer:
> Thanks! I have a few questions / suggestions. I also wonder whether `--filter` is the right name. Maybe `--verify-only`?

The option sits in the context of the `verify` command, so I don't think the word `verify` needs to be part of its name.

Source/DafnyDriver/Commands/FilterOption.cs
keyboardDrummer: I'm not sure why the * at the end is necessary, but I can't get anything to pass the filter without it. The * at the start is because Dafny adds internally defined prefixes to the proof names when translating them to Boogie, so I don't think we can do without it.

Source/DafnyDriver/Commands/FilterOption.cs
keyboardDrummer: Why would we want to mention Boogie? The user should be able to filter in terms of their Dafny proofs.

Source/DafnyDriver/Commands/CommandRegistry.cs
keyboardDrummer: Should be removed :) thanks.

Source/DafnyDriver/Commands/FilterOption.cs
keyboardDrummer: I think ideally, the configured pattern is used to do a symbol search at the Dafny level, and then Dafny's Boogie translation machinery is invoked to get the Boogie name for this symbol, which is passed to Boogie's `ProcsToCheck`. But building that doesn't seem like it's worth the effort run now, since this works well enough.

Source/DafnyDriver/Commands/FilterOption.cs
cpitclaudel: I was suggesting a documentation improvement: what I meant was that the option currently takes Boogie names (doesn't it?), and that while this isn't ideal, at least we should document it so that users know what to pass (it's not clear to me what it means that a proof is matched by a pattern).

Source/DafnyDriver/Commands/FilterOption.cs
cpitclaudel: But it's convenient to be able to match on specific prefixes, so always adding the `*` isn't ideal; could we just recommend adding them in the doc?

keyboardDrummer: Are you suggesting to remove the automatic prefix and postfix wildcards and document with: "Currently the pattern matching is on the Boogie names that are generated by Dafny. If you're not sure how those are generated, please add * and * at the start and end of your pattern" ?

> But it's convenient to be able to match on specific prefixes

The situation in which you want to match everything with a particular prefix, but that prefix is also the substring of a symbol you don't want to match, seems rare to me. I would guess that supporting that is less important than making this feature more user friendly through the automatic pre and post wildcards, but I'm OK with both. Let me know which one you prefer.

cpitclaudel: I prefer the version where we don't introduce the wildcards; it's relatively common to have methods an modules named the same, which causes collisions in `/proc`

keyboardDrummer: > it's relatively common to have methods an modules named the same

If there's a method `Y` with fully qualified name `X.Y` and a module `Y`, then there's no problem right? `--filter=X.Y` still works. What's the common situation you're worried about?

Test/verification/filter.dfy
```diff
@@ -0,0 +1,11 @@
+// RUN: %baredafny verify %args --bprint:"%t.bpl" --filter="Succeeds" %s > %t
+// RUN: %baredafny verify %args %s >> %t || true
```
robin-aws, line 2:
```suggestion
// RUN: ! %baredafny verify %args %s >> %t
```

Test/verification/filter.dfy.expect
```diff
@@ -0,0 +1,6 @@
+
+Dafny program verifier finished with 1 verified, 0 errors
+filter.dfy(10,16): Error: A postcondition might not hold on this return path.
+filter.dfy(10,10): Related location: This is the postcondition that might not hold.
+
+Dafny program verifier finished with 1 verified, 1 error
```
robin-aws, line 6: Isn't this result backwards? The first call filters to `Success` so it should succeed AFAICT

Source/DafnyCore/Options/VerifyIncludesOption.cs
```diff
@@ -0,0 +1,13 @@
+namespace Microsoft.Dafny; 
+
+public class VerifyIncludesOption : BooleanOption {
+  public static readonly VerifyIncludesOption Instance = new();
+  public override string LongName => "verify-includes";
```
robin-aws, line 5: Name feels a little off to me, like we're verifying the include statements themselves rather than the included content. How about `verify-included-files`?

Source/DafnyDriver/Commands/FilterOption.cs
robin-aws: I would actually prefer we name this `--boogie-filter`, to be clearer about its semantics and also to let us define `--filter` later on, if and when it's reasonable to implement in terms of Dafny concepts. Then we can document the latter as filtering on names of assertion batches.

Source/DafnyDriver/Commands/FilterOption.cs
robin-aws: As per the above I think we should not add the `*`s as long as we're clearer we're filtering to Boogie names.

Test/dafny0/Include.dfy
```diff
@@ -1,4 +1,5 @@
-// RUN: %dafny_0 /compile:0 /print:"%t.print" /dprint:"%t.dprint" "%s" > "%t"
+// RUN: %baredafny verify %args "%s" > "%t" || true
+// RUN: %baredafny verify %args --verify-included-files "%s" >> "%t" || true
```
robin-aws, line 2:
```suggestion
// RUN: ! %baredafny verify %args "%s" > "%t"
// RUN: ! %baredafny verify %args --verify-included-files "%s" >> "%t"
```

Test/verification/filter.dfy.expect
keyboardDrummer: The double quotes made our custom lit parser fail and ignore the argument.

robin-aws: Never mind, I was confused by the spacing of the output. The first two lines are from the successful run from filtering to the `Success`, and the following errors are from the second run that doesn't filter.

I do see that the custom lit parser is dropping the single quotes in the argument, but that doesn't seem to affect correctness here. Might be worth fixing for maximum consistency in the future though.

PR #3054, opened 2022-11-14
Add missing options to new CLI

keyboardDrummer:
### Changes
- Add `test` command that replaces the `/runAllTests` option. This command has similar options as the run command, except it takes multiple input files as arguments.
- Add `generate-tests` command, which replaces `/generateTestMode` and related options.
- Add `find-dead-code` command, which replaces `/warnDeadCode` and related options.
- Ported over the existing options: quantifier-syntax, function-syntax options, warn-shadowing, warnings-as-errors, warn-missing-constructor-parenthesis

### Tests
Updated tests to use new command and options


Source/DafnyCore/Options/DisableNonLinearArithmeticOption.cs
```diff
@@ -0,0 +1,15 @@
...
+class DisableNonLinearArithmeticOption : BooleanOption {
+  public static readonly DisableNonLinearArithmeticOption Instance = new();
+  public override string LongName => "disable-nonlinear-arithmetic";
+
+  public override string Description => @"      
+Reduce Dafny's knowledge of non-linear arithmetic (*,/,%).
+  
+Results in more manual work, but also produces more predictable behavior.".TrimStart();
+  public override string PostProcess(DafnyOptions options) {
+    options.DisableNLarith = Get(options);
```
cpitclaudel, line 12: Isn't this option deprecated?  I thought we wanted to remove it rather than port it over.

Source/DafnyCore/Options/FunctionSyntaxOption.cs
```diff
@@ -0,0 +1,63 @@
...
+      options.FunctionSyntax = DafnyOptions.FunctionSyntaxOptions.Migration3To4;
+    } else if (value == "experimentalDefaultGhost") {
+      options.FunctionSyntax = DafnyOptions.FunctionSyntaxOptions.ExperimentalTreatUnspecifiedAsGhost;
+    } else if (value == "experimentalDefaultCompiled") {
+      options.FunctionSyntax = DafnyOptions.FunctionSyntaxOptions.ExperimentalTreatUnspecifiedAsCompiled;
+    } else if (value == "experimentalPredicateAlwaysGhost") {
+      options.FunctionSyntax = DafnyOptions.FunctionSyntaxOptions.ExperimentalPredicateAlwaysGhost;
+    } else {
+      return $"Invalid argument to option {LongName}";
+    }
```
cpitclaudel, line 59: Could a switch work here?

Source/DafnyCore/Options/ICommandSpec.cs
```diff
@@ -0,0 +1,54 @@
...
+
+public interface ICommandSpec {
+
+  public static Argument<FileInfo> FileArgument { get; }
+
+  private static ValidateSymbolResult<ArgumentResult> ValidateFileArgument() {
+    return r => {
+      var value = r.Tokens[0].Value;
+      if (value.StartsWith("--")) {
+        r.ErrorMessage = $"{value} is not a valid argument";
```
cpitclaudel, line 17: Why?

Source/DafnyCore/Options/QuantifierSyntaxOption.cs
```diff
@@ -0,0 +1,36 @@
...
+
+  public override string PostProcess(DafnyOptions options) {
+    var value = Get(options);
+    if (value == "3") {
+      options.QuantifierSyntax = DafnyOptions.QuantifierSyntaxOptions.Version3;
+    } else if (value == "4") {
+      options.QuantifierSyntax = DafnyOptions.QuantifierSyntaxOptions.Version4;
+    } else {
+      return $"Invalid argument to option {LongName}";
+    }
```
cpitclaudel, line 32: Same here?

Source/DafnyCore/Options/WarnMissingConstructorParenthesisOption.cs
```diff
@@ -0,0 +1,14 @@
+namespace Microsoft.Dafny;
+
+class WarnMissingConstructorParenthesisOption : BooleanOption {
+  public static readonly WarnMissingConstructorParenthesisOption Instance = new();
+  public override string LongName => "warn-missing-constructor-parentheses";
+  public override string Description => @"
+    Emits a warning when a constructor name in a case pattern is not
+    followed by parentheses.".TrimStart();
```
cpitclaudel, line 8: I would prefer to not port the individual warning options to the new CLI.  Instead we should have a single option `--warn` taking a list of warning identifiers.  WDYT?

Source/DafnyCore/Options/WarnShadowingOption.cs
```diff
@@ -0,0 +1,14 @@
+namespace Microsoft.Dafny;
+
+class WarnShadowingOption : BooleanOption {
+  public static readonly WarnShadowingOption Instance = new();
+  public override string LongName => "warn-shadowing";
+  public override string Description => @"
+    Emits a warning if the name of a declared variable caused another
+    variable to be shadowed.".TrimStart();
```
cpitclaudel, line 8: Same here: `--warn=+shadowing`

Source/DafnyDriver/Commands/TestCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  public void PostProcess(DafnyOptions dafnyOptions, Options options, InvocationContext context) {
+    dafnyOptions.Compile = true;
+    dafnyOptions.RunAfterCompile = true;
+    dafnyOptions.RunAllTests = true;
+    dafnyOptions.ForceCompile = NoVerifyOption.Instance.Get(options);
+    dafnyOptions.CompileVerbose = false;
+  }
+}
```
cpitclaudel, line 30: Does this return the right error code when a test fails?

Source/DafnyTestGeneration/DeadCodeCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+public class DeadCodeCommand : ICommandSpec {
+  public IEnumerable<IOptionSpec> Options =>
+    new IOptionSpec[] {
+      LoopUnrollOption.Instance,
+      SequenceLengthLimitOption.Instance,
+      VerificationTimeLimitOption.Instance,
+    }.Concat(ICommandSpec.CommonOptions);
+
+  public Command Create() {
+    var result = new Command("find-dead-code", "(Experimental) Use block-coverage tests to warn about potential dead code.");
```
cpitclaudel, line 17: Is this documented somewhere?

Test/dafny0/ForallCompilationNewSyntax.dfy.expect
```diff
@@ -1,2 +1,6 @@
+ForallCompilationNewSyntax.dfy(26,4): Warning: /!\ No terms found to trigger on.
+ForallCompilationNewSyntax.dfy(35,4): Warning: /!\ No terms found to trigger on.
+ForallCompilationNewSyntax.dfy(44,4): Warning: /!\ No terms found to trigger on.
+ForallCompilationNewSyntax.dfy(64,4): Warning: /!\ No trigger covering all quantified variables found.
```
cpitclaudel, line 4: I'd prefer to add triggers by hand instead of this, to avoid unrelated messages.  But this is fine.

cpitclaudel (changes requested):
I left a few comments.  The most important bit I guess is the warnings; I think it would be too bad to propagate our current warning system to the new CLI.

Source/DafnyCore/Options/DisableNonLinearArithmeticOption.cs
keyboardDrummer: Judging by `https://github.com/dafny-lang/libraries/tree/master/src/NonlinearArithmetic` I thought it was in active use. Is it not?

Source/DafnyCore/Options/ICommandSpec.cs
keyboardDrummer: Removed it. I think it was an ill-conceived attempt at warning the user against accidentally passing options as files.

Source/DafnyCore/Options/WarnMissingConstructorParenthesisOption.cs
keyboardDrummer: So instead of `--warn-X --warn-Y --warn-Z`, users would write `--warn=+X+Y+Z`, or are you suggesting something else? It doesn't seem particularly easier to use and it adds a new concept to the CLI UI. Having separate options for separate warnings also makes it easier to define a help description for each of them.

Source/DafnyDriver/Commands/TestCommand.cs
keyboardDrummer: I don't think it does. The behavior is determined by the code in `/Users/rwillems/Documents/SourceCode/dafny/Source/DafnyCore/Resolver/RunAllTestsMainMethod.cs`. I'm guessing Dafny currently doesn't have a way of letting its programs return a non-zero exit code. Letting `dafny test` return the right exit code seems like a significant amount of work, delaying that seems worth the breakage of having to modify the `dafny test` exit code later on.

Test/dafny0/ForallCompilationNewSyntax.dfy.expect
keyboardDrummer: How do I do that? Or find documentation for how to do that?

cpitclaudel: I think this:

```
  forall x: int <- {1, 2} {:trigger false} {    
  }
```

should work

Test/comp/ForallNewSyntax.dfy
```diff
@@ -1,9 +1,9 @@
...
-// RUN: %dafny /quantifierSyntax:4 /noVerify /compile:4 /spillTargetCode:2 /compileTarget:js "%s" >> "%t"
-// RUN: %dafny /quantifierSyntax:4 /noVerify /compile:4 /spillTargetCode:2 /compileTarget:go "%s" >> "%t"
-// RUN: %dafny /quantifierSyntax:4 /noVerify /compile:4 /spillTargetCode:2 /compileTarget:java "%s" >> "%t"
-// RUN: %dafny /quantifierSyntax:4 /noVerify /compile:4 /spillTargetCode:2 /compileTarget:py "%s" >> "%t"
+// RUN: %baredafny verify %args --quantifier-syntax:4 "%s" > "%t"
+// RUN: %baredafny run %args --quantifier-syntax:4 --no-verify --target:cs "%s" >> "%t"
+// RUN: %baredafny run %args --quantifier-syntax:4 --no-verify --target:js "%s" >> "%t"
+// RUN: %baredafny run %args --quantifier-syntax:4 --no-verify --target:go "%s" >> "%t"
+// RUN: %baredafny run %args --quantifier-syntax:4 --no-verify --target:java "%s" >> "%t"
+// RUN: %baredafny run %args --quantifier-syntax:4 --no-verify --target:py "%s" >> "%t"
```
cpitclaudel, line 6: Does `%baredafny %args run` work?  In tools like `git` you can pass common options between `git` and the command, e.g. `git --git-dir=… pull …`

keyboardDrummer: It might after we drop support for the old CLI UI, but right now we detect whether the first argument is a command to turn on the new UI

Test/dafny0/ForallCompilationNewSyntax.dfy.expect
keyboardDrummer: That gives a Boogie error
`ForallCompilationNewSyntax.dfy(26,4): Error: trigger must mention all quantified variables, but does not mention: k#0_2`

Source/DafnyCore/Options/DisableNonLinearArithmeticOption.cs
cpitclaudel: It's in use, but we're not super happy about that:

```

  /noNLarith
      Reduce Z3's knowledge of non-linear arithmetic (*,/,%).
  
      Results in more manual work, but also produces more predictable
      behavior. (This switch will perhaps be replaced by /arith in the
      future. For now, it takes precedence of /arith.)
```

Source/DafnyCore/Options/WarnMissingConstructorParenthesisOption.cs
cpitclaudel: That's right, some syntax like this.  It's fine if the syntax and perhaps the implementation ends up looking like `--warn-xyz`, as long as there are easy ways to say:

- Turn on all warnings, except `X`
- Treat all warnings as errors, except `X` (treat it as a warning)
- Treat `X` as an error

etc (there may be some combinations I'm forgetting right now).  All of these are pretty standard features, I think (e.g. https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings).  Additionally, having identifiers for warnings and errors will make it easy to have docs (and explainers, down the line) for each of them (e.g. https://eslint.org/docs/latest/rules/getter-return or https://pylint.pycqa.org/en/latest/user_guide/messages/error/bad-format-character.html)

Source/DafnyDriver/Commands/TestCommand.cs
cpitclaudel: I think we recently merged a fix for this in `/runAllTests` (relevant issue is https://github.com/dafny-lang/dafny/issues/3043); let's just check that the fix also works with the new `dafny test`.

Test/dafny0/ForallCompilationNewSyntax.dfy.expect
cpitclaudel: Hmm, that's a regression.  Let's stick with what you have for now then.

Test/comp/ForallNewSyntax.dfy
cpitclaudel: !! I forgot.  Of course.

Source/DafnyCore/Options/WarnMissingConstructorParenthesisOption.cs
keyboardDrummer: > Treat all warnings as errors, except X
> Treat X as an error

What's the use-case for either of those options? To me a warning is a problem that must to be resolved before before merging to main, but not before development testing. We have the option `--warnings-as-errors` which should always be turned on during CI. Why would you want some warnings as errors and others not? Would you use that during development or CI ? During CI seems wrong because then you enable checking in warnings and your project may grow in amount of warnings leading you to ignore them. During development I can imagine but it seems too far fetched. It would be you differentiate between three categories of warnings:

1. Never useful, turn off
1. Too dangerous to be checked in, error in CI. Not too dangerous for exploratory testing, warning during development.
1. So dangerous you don't even want to do exploratory testing with this, since it's likely you have a bug.

Is there really a user that will differentiate between 2 and 3 ?

> Treat all warnings as errors, except X

I don't think anyone would use that option. Warnings that are off by default are the ones for which we think the extra safety is likely not worth the effort, and there will be an arbitrarily large bucket of such warnings. Is there any chance you would want approximately everything in the bucket of not recommended things? Also, if you use this option, upgrading to a newer Dafny version may prevent your project from compiling because a new warning has been added.

> Additionally, having identifiers for warnings and errors will make it easy to have docs

Why? It'll be harder to have CLI documentation for it.

---

How about we introduce the `--warn-X` syntax now, and we leave the option open to have `--error-X` in the future, and `--all-warnings` together with `--warn-X=false` ?

Source/DafnyDriver/Commands/TestCommand.cs
keyboardDrummer: Tests now check both the zero and non-zero exit code scenario.

Source/DafnyCore/Options/DisableNonLinearArithmeticOption.cs
keyboardDrummer: I've marked it as experimental

Source/DafnyTestGeneration/DeadCodeCommand.cs
robin-aws: It's documented quite well in https://github.com/dafny-lang/dafny/tree/master/Source/DafnyTestGeneration, but we don't mention it in the reference manual or user guide or anything like that.

robin-aws: I'd reword the description of the feature slightly though, because it doesn't actually involve block-coverage tests, it just uses the same common technique of trying to find a counterexample for an `assert false;`. How about:

```suggestion
    var result = new Command("find-dead-code", "(Experimental) Use counterexample generation to warn about potential dead code.");
```

Source/DafnyTestGeneration/GenerateTestsCommand.cs
```diff
@@ -0,0 +1,102 @@
...
+    }.Concat(ICommandSpec.CommonOptions);
+
+  enum Mode {
+    Path,
+    Block
+  }
+
+  readonly Argument<Mode> modeArgument = new("mode", @"
+Block - Prints block-coverage tests for the given program.
+Path - Prints path-coverage tests for the given program.");
```
robin-aws, line 27: Let's make the modes lowercase to match the style of the rest of the new CLI.

Source/DafnyTestGeneration/GenerateTestsCommand.cs
```diff
@@ -0,0 +1,102 @@
...
+      Mode.Path => TestGenerationOptions.Modes.Path,
+      Mode.Block => TestGenerationOptions.Modes.Block,
+      _ => throw new ArgumentOutOfRangeException()
+    };
+  }
+}
+
+class TargetMethod : StringOption {
+  public static readonly TargetMethod Instance = new();
+  public override object DefaultValue => null!;
```
robin-aws, line 55: Doesn't `null!` always lead to an exception, or am I misreading this?

Test/dafny0/ForbidNondeterminism.dfy
```diff
@@ -1,4 +1,4 @@
-// RUN: %baredafny verify %args --strict-definite-assignment --enforce-determinism "%s" > "%t" || true
+// RUN: %baredafny verify %args --strict-definite-assignment "%s" > "%t" || true
```
robin-aws, line 1: Why this change? Wasn't that the point of the test? :)

Test/dafny0/PrintEffects.dfy
```diff
@@ -1,5 +1,5 @@
-// RUN: %dafny /compile:3 "%s" > "%t"
-// RUN: %dafny_0 /compile:3 /trackPrintEffects:1 "%s" >> "%t"
+// RUN: %baredafny run %args "%s" > "%t"
+// RUN: %baredafny run %args --track-print-effects "%s" >> "%t" || true
```
robin-aws, line 2:
```suggestion
// RUN: ! %baredafny run %args --track-print-effects "%s" >> "%t"
```

Not just tidier, strictly more precise since it will fail if the exit code is 0.

Test/git-issues/git-issue-1564-3.dfy
```diff
@@ -1,6 +1,6 @@
-// RUN: %dafny /compile:0 /functionSyntax:3 /env:0 /dprint:- "%s" > "%t"
```
robin-aws, line 1: Are we dropping `/env`?

Source/DafnyCore/Options/WarnMissingConstructorParenthesisOption.cs
robin-aws: I think in the context of this PR, just supporting `--warn-X` for now is fine.

In a wider scope, though, I think having these warnings enabled by default and being able to disable individual warnings (probably even specific instances of warnings on specific bits of code) is really important for them to actually be useful.

Should we be considering enabling warnings by default in Dafny 4?

robin-aws (changes requested):
So happy to finally get my `dafny test`! :)

Source/DafnyDriver/Commands/TestCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+    new IOptionSpec[] {
+      TargetOption.Instance,
+      NoVerifyOption.Instance,
+      EnforceDeterminismOption.Instance,
+      VerificationTimeLimitOption.Instance,
+    }.Concat(ICommandSpec.CommonOptions);
+
+  public Command Create() {
+    var result = new Command("test", "(Experimental) Execute every method in the program that's annotated with the {:test} attribute.");
+    result.AddArgument(ICommandSpec.FilesArgument);
```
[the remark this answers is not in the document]
davidcok: And also differentiate 'test' and 'generate-tests'

Test/dafny0/PrintEffects.dfy
keyboardDrummer: Applied this across all tests

Test/git-issues/git-issue-1564-3.dfy
keyboardDrummer: I've made `env:0` the default and there's no `--env` yet

keyboardDrummer:
> In a wider scope, though, I think having these warnings enabled by default and being able to disable individual warnings (probably even specific instances of warnings on specific bits of code) is really important for them to actually be useful.

I think a warning not enabled by default means it's not recommended to use, and then I wonder why we are maintaining it, so I agree with you.

Test/dafny0/ForbidNondeterminism.dfy
keyboardDrummer: `--enforce-determinism` was testing that this flag has no effect when only verifying. This PR removing the option `--enforce-determinism` from the `verify` command so it's not possible to keep it here.

Source/DafnyTestGeneration/GenerateTestsCommand.cs
keyboardDrummer: `!` just tells it not to produce a warning.

Source/DafnyDriver/Commands/TestCommand.cs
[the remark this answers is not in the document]
keyboardDrummer: > Is it possible to tweak FilesArgument so that we get verify [<file>...] and so on, except for run <file>?

I'm pretty sure it is although it involves tweaking the help generation which I haven't played around with yet.

cpitclaudel (dismissed):
Looks good to me overall.  We can worry about the syntax of warnings later.

As for enabling all the warnings we currently have by default: strong -1 for me.  Some are much too sensitive and flag too many false positives.

Source/DafnyCore/Options/WarnMissingConstructorParenthesisOption.cs
cpitclaudel: > What's the use-case for either of those options?

Same for other languages: precise control about warnings.
In our case, we have warnings that are informational with high false-positive rates, and others that are true issues.  Different teams will want to handle them differently.
For example, the non-determinism flags should be warnings, not a separate option.  That warning could be on by default.  Some teams would turn it off; other would turn it (selectively) into an error.

> Why? It'll be harder to have CLI documentation for it.

For the reason I listed above: we can have one webpage per warning identifier.
I don't understand how having unique identifiers for warnings and errors will make CLI documentation harder (we can still have individual CLI flags, though I'm not sold on that being useful).

> Also, if you use this option, upgrading to a newer Dafny version may prevent your project from compiling because a new warning has been added.

So does `--warnings-as-errors`, which you say should always be turned on during CI, right?
In fact, having a way to treat all warnings as errors except for a few specific exception makes this problem easier to handle: if you have a codebase with `warnings-as-errors` and we add a warning in a new Dafny release, your code will break (assuming you trigger the warning). If you have a flag that allows to turn off warnings as errors for that specific warning, you can push that and still benefit from the update. With the current situation, you have to turn off warnings-as-error for *all* warnings.

keyboardDrummer: > if you have a codebase with warnings-as-errors and we add a warning [that is on by default] in a new Dafny release, your code will break (assuming you trigger the warning)

If we add a check that's on by default, then we need to add an option to turn that check off to be backwards compatible.

> Same for other languages: precise control about warnings.
In our case, we have warnings that are informational with high false-positive rates, and others that are true issues. Different teams will want to handle them differently.

My argument was that you only need to be able to turn the check fully on or fully off, where fully on would mean it's a warning during development and an error during CI.

I think it's generally considered bad practice to have something that doesn't prevent you from running the program, to be an error during development. It's important to be able to code freely, run and test, and only then go into polishing your program so it passes CI.

> For example, the non-determinism flags should be warnings, not a separate option. 

I agree that the difference between language feature flags and checks doesn't seem so crisp!

Source/DafnyTestGeneration/GenerateTestsCommand.cs
robin-aws: Ah it's not a "non-null cast" then. It is difficult to make `DefaultValue` an `object?` instead in `IOptionSpec`?

robin-aws (dismissed):
Added a couple of comments and replies but nothing blocking.

robin-aws:
> As for enabling all the warnings we currently have by default: strong -1 for me. Some are much too sensitive and flag too many false positives.

I agree just flipping on all possible warnings by default is not a good idea, but I think we do need to settle on a philosophy for warnings that makes at least some of them on by default, and that may mean taking the Dafny 4 opportunity to do that as a breaking change. @keyboardDrummer perhaps we should cut a separate issue for that?

jtristan (dismissed):
Superficial changes resulting from merge since previous approval.

alex-chew (approved):
Approving based on minimal changes from prior approvals.

PR #3056, opened 2022-11-15
feat: Compiler optimization: Omit wrappers around single other type

RustanLeino:
This PR introduces a compiler optimization for datatypes. In particular,

* A type
    ```` dafny
    datatype Record = Record(x: int)
    ````
    is compiled to the target type `int`. More generally, if an inductive `datatype` has exactly one non-ghost constructor and that constructor has exactly one non-ghost member (and some other conditions apply, see fine print below), then values of the datatype are compiled as that member, without the enclosing datatype-constructor "wrapper".
* A tuple type with ghost components is compiled to the smaller tuple without the ghost components. For example, the 4-tuple type `(ghost int, real, ghost bool, bv3)` is compiled to the 2-tuple type `(real, bv3)`. (This optimization is not entirely new with this PR, but this PR makes the treatment of this more uniform.)
* As a special case of the above, if a ghost tuple is trimmed down to a non-ghost 1-tuple, then (on behalf of the first bullet above), the target type is further changed to be just that non-ghost component. For example, `(int, ghost int)` is compiled as just `int`.

## Benefits

The optimization has several benefits. As one example, a function

``` dafny
function F(x: X): (Result, ghost AuxiliaryData)
```

is compiled to just

``` dafny
function F(x: X): Result
```

That is, there is no run-time cost of having also returned a ghost component in the Dafny source code.

As a second example, types that use datatype constructors as "tags" get optimized in the compiled code. For instance,

``` dafny
datatype Box<X> = Box(value: X)
```

is compiled to just `X` (whatever `X` happens to be).

As a third example, types with uninitialized values (using ghost constructors) have any run-time cost. For instance,

``` dafny
datatype MaybeInitialized<X> =
  | ghost Uninitialized
  | Initialized(value: X)
```

is compiled to just `X` (and the uninitialized value is typically compiled to a zero-equivalent value of the target type `X`).

## The fine print

The optimization is _not_ performed when one of the following applies:

* the datatype is declared with `{:extern}`
* the datatype is a `codatatype`
* the datatype has a `const` member
* the datatype declaration is involved in a cycle; for example,
    ``` dafny
    datatype MyselfForever = ghost Nothing | Something(MyselfForever)
    ```
    is not optimized
* the command-line option `/optimizeErasableDatatypeWrapper:0` is used.

## Supporting optimizations

The PR provides full compilation support for C#, Java, JavaScript, Go, and Python (but not C++, since the Dafny-to-C++ compiler lacks many features). (In the implementation, the compiler getter `OptimizesErasableDatatypeWrappers` lets a compiler, like C++, opt out of the optimization altogether.)

Destructors (and update expressions) are optimized accordingly. For example, if `r` has type `Record` from above, then `r.x` is compiled into (the target code for) `r` and `r.(x := E)` is compiled as just `E`. As another example, if `v` has type `(ghost int, int, int)`, then `v.1` is compiled into `v.0`.

For a datatype `T` with exactly one non-ghost constructor, say `C`, a discriminator expression `t.C?` is compiled as just `true`.

The PR also optimizes an expression `d.(g := E)` into just `d` if `g` is a ghost field.

Other things that the optimization requires (and this PR handles) are:

* Equality. For example, if `r` and `s` are of type `Record` above, then the `Record` comparison `r == s` is compiled into the integer comparison `r.x == s.x`.
* `match` constructs.
* Default values.
* Type descriptors.
* For C#, `DowncastClone`.
* Instance members of optimized types turn into custom-receiver static members.

## Compatibility

For code written entirely in Dafny, this PR should have no impact (other than optimized behavior at run time). Code that links to the Dafny code directly will notice the missing wrappers. Such code can add the `{:extern}` attribute to the `datatype` declaration, since that will disable the optimization. To opt out of the optimization entirely, use `/optimizeErasableDatatypeWrapper:0`.


Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1325,8 +1339,14 @@ private class ClassWriter : IClassWriter {
...
-      wr.Write(source);
-      wr.Write(ctor.EnclosingDatatype is TupleTypeDecl ? $"[{dtor.NameForCompilation}]" : $".{IdProtect(dtor.CompileName)}");
+      if (IsInvisibleWrapper(ctor.EnclosingDatatype, out var coreDtor)) {
+        Contract.Assert(coreDtor.CorrespondingFormals.Count == 1);
+        Contract.Assert(dtor == coreDtor.CorrespondingFormals[0]); // any other destructor is a ghost
+        wr.Write(source);
+      } else {
+        wr.Write(source);
+        wr.Write(ctor.EnclosingDatatype is TupleTypeDecl ? $"[{dtor.NameForCompilation}]" : $".{IdProtect(dtor.CompileName)}");
+      }
```
fabiomadge, line 1349:
```suggestion
      wr.Write(source);
      if (IsInvisibleWrapper(ctor.EnclosingDatatype, out var coreDtor)) {
        Contract.Assert(coreDtor.CorrespondingFormals.Count == 1);
        Contract.Assert(dtor == coreDtor.CorrespondingFormals[0]); // any other destructor is a ghost
      } else {
        wr.Write(ctor.EnclosingDatatype is TupleTypeDecl ? $"[{dtor.NameForCompilation}]" : $".{IdProtect(dtor.CompileName)}");
      }
```

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -751,7 +754,12 @@ private class ClassWriter : IClassWriter {
 
     protected override string TypeName_Companion(Type type, ConcreteSyntaxTree wr, IToken tok, MemberDecl member) {
       type = UserDefinedType.UpcastToMemberEnclosingType(type, member);
```
fabiomadge, line 756: Can we move this inside the else block?

Source/DafnyCore/Resolver.cs
```diff
@@ -14691,6 +14691,39 @@ class LazyString_OnTypeEquals {
       }
     }
 
+    /// <summary>
+    /// Return a type that is like "type", but with type arguments "arguments".
+    /// </summary>
+    public static Type ReplaceTypeArguments(Type type, List<Type> arguments) {
```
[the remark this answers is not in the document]
RustanLeino: Good point. I had placed it there to be next to the related `SubstType` method, but it is equally out of place.

I create PR https://github.com/dafny-lang/dafny/pull/3057 to move two previous type-related utility methods into a new place: a class `TypeUtil`. I will merge with that PR and will move `ReplaceTypeArguments` there as well.

RustanLeino:
> Could this optimisation be done in a pass that runs before SinglePassCompiler?
> 
> I think we agreed as a group that transformations which can be implemented as separate passes, should be. The benefit in this case is that we wouldn't add more code to a transformation that's already complex.

@keyboardDrummer Yes, that would be better and cleaner! (I had started this PR over a year ago and had returned to it now to finish it up.) To do this, we need some new machinery for cloning/substituting resolved expressions/statements/declarations. Once we have that machinery, I imagine we'll use it a lot.

Should we put this PR on hold (indefinitely) and reimplement it as a transformation once we (separately) build that machinery?

Source/DafnyCore/Compilers/Compiler-python.cs
RustanLeino: No, the updated `type` is also needed in the `if` guard.

fabiomadge: Are you sure?

keyboardDrummer:
> @keyboardDrummer Yes, that would be better and cleaner! (I had started this PR over a year ago and had returned to it now to finish it up.) To do this, we need some new machinery for cloning/substituting resolved expressions/statements/declarations. Once we have that machinery, I imagine we'll use it a lot.

Note that we're adding the ability to clone resolved AST nodes in https://github.com/dafny-lang/dafny/pull/2734

> Should we put this PR on hold (indefinitely) and reimplement it as a transformation once we (separately) build that machinery?

I'm OK with both. I think this PR adds a small complexity increase to the compilation back-ends. If there's no rush in merging this then I think waiting until https://github.com/dafny-lang/dafny/pull/2734 lands is better, but if a user would greatly benefit from this now then let's merge it.

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -271,7 +272,9 @@ public class PythonCompiler : SinglePassCompiler {
       var wDefault = btw.NewBlockPy($"def default(cls, {UsedTypeParameters(dt).Comma(FormatDefaultTypeParameterValue)}):");
       var groundingCtor = dt.GetGroundingCtor();
       if (groundingCtor.IsGhost) {
-        wDefault.WriteLine($"return {ForcePlaceboValue(UserDefinedType.FromTopLevelDecl(dt.tok, dt), wDefault, dt.tok)}");
+        wDefault.WriteLine($"return lambda: {ForcePlaceboValue(simplifiedType, wDefault, dt.tok)}");
```
fabiomadge, line 275: We only use `simplifiedType` here, but we don't even need it, because `ForcePlaceboValue` does the same step for us.
```suggestion
        wDefault.WriteLine($"return lambda: {ForcePlaceboValue(UserDefinedType.FromTopLevelDecl(dt.tok, dt), wDefault, dt.tok)}");
```

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -751,7 +754,12 @@ private class ClassWriter : IClassWriter {
 
     protected override string TypeName_Companion(Type type, ConcreteSyntaxTree wr, IToken tok, MemberDecl member) {
       type = UserDefinedType.UpcastToMemberEnclosingType(type, member);
-      return TypeName(type, wr, tok, member);
+      if (type.NormalizeExpandKeepConstraints() is UserDefinedType udt && udt.ResolvedClass is DatatypeDecl dt && IsInvisibleWrapper(dt, out _)) {
```
fabiomadge, line 757:
```suggestion
      if (type.NormalizeExpandKeepConstraints() is UserDefinedType {ResolvedClass: DatatypeDecl dt} udt && IsInvisibleWrapper(dt, out _)) {
```

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1167,6 +1175,12 @@ private class ClassWriter : IClassWriter {
     protected override ILvalue EmitMemberSelect(Action<ConcreteSyntaxTree> obj, Type objType, MemberDecl member,
       List<TypeArgumentInstantiation> typeArgs, Dictionary<TypeParameter, Type> typeMap, Type expectedType,
       string additionalCustomParameter = null, bool internalAccess = false) {
+      var memberStatus = GetMemberStatus(member);
+      if (memberStatus == MemberCompileStatus.Identity) {
+        return SimpleLvalue(obj);
+      } else if (memberStatus == MemberCompileStatus.AlwaysTrue) {
+        return SimpleLvalue(w => w.Write("True"));
+      }
```
fabiomadge, line 1183:
```suggestion
      switch (GetMemberStatus(member)) {
        case MemberCompileStatus.Identity:
          return SimpleLvalue(obj);
        case MemberCompileStatus.AlwaysTrue:
          return StringLvalue("True");
```

fabiomadge (commented):
Sorry for revisiting.

RustanLeino:
> > Should we put this PR on hold (indefinitely) and reimplement it as a transformation once we (separately) build that machinery?
>
> I'm OK with both. I think this PR adds a small complexity increase to the code-generators. If there's no rush in merging this then I think waiting until https://github.com/dafny-lang/dafny/pull/2734 lands is better, but if a user would greatly benefit from this now then let's merge it.

@keyboardDrummer Thinking about this some more, I suggest we then work toward merging this PR as is. This will add the feature and also includes some other fix, and will remove the need to constantly keep this PR is limbo and up-to-date with the latest sources. Then, when we're ready to start writing transforms for our compiler, we can use this as an early application thereof. The `IsInvisibleWrapper` and `SimplifyType` methods in this PR will be used by the transformation, and so will the test cases added in this PR.

keyboardDrummer:
> > > Should we put this PR on hold (indefinitely) and reimplement it as a transformation once we (separately) build that machinery?
> > 
> > 
> > I'm OK with both. I think this PR adds a small complexity increase to the code-generators. If there's no rush in merging this then I think waiting until #2734 lands is better, but if a user would greatly benefit from this now then let's merge it.
> 
> @keyboardDrummer Thinking about this some more, I suggest we then work toward merging this PR as is. This will add the feature and also includes some other fix, and will remove the need to constantly keep this PR is limbo and up-to-date with the latest sources. Then, when we're ready to start writing transforms for our compiler, we can use this as an early application thereof. The `IsInvisibleWrapper` and `SimplifyType` methods in this PR will be used by the transformation, and so will the test cases added in this PR.

Sounds good

Source/DafnyCore/Compilers/Compiler-python.cs
```diff
@@ -1167,6 +1174,14 @@ private class ClassWriter : IClassWriter {
     protected override ILvalue EmitMemberSelect(Action<ConcreteSyntaxTree> obj, Type objType, MemberDecl member,
       List<TypeArgumentInstantiation> typeArgs, Dictionary<TypeParameter, Type> typeMap, Type expectedType,
       string additionalCustomParameter = null, bool internalAccess = false) {
+      switch (GetMemberStatus(member)) {
+        case MemberCompileStatus.Identity:
+          return SimpleLvalue(obj);
+        case MemberCompileStatus.AlwaysTrue:
+          return SimpleLvalue(w => w.Write("True"));
```
fabiomadge, line 1181:
```suggestion
          return StringLvalue(True);
```

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1047,9 +1049,134 @@ public class TypeArgumentInstantiation {
...
+            // there is exactly one non-ghost constructor
+            var ctor = dt.Ctors[i];
+            var j = ctor.Destructors.FindIndex(dtor => !dtor.IsGhost);
+            if (0 <= j) {
+              if (ctor.Destructors.FindIndex(j + 1, dtor => !dtor.IsGhost) == -1) {
+                // there is exactly one non-ghost parameter to "ctor"
+                var candidateCoreDestructor = ctor.Destructors[j];
+                var properlyContainedTypeDecls = new HashSet<TopLevelDecl>();
+                CompiledTypes(candidateCoreDestructor.Type, properlyContainedTypeDecls);
+                if (!properlyContainedTypeDecls.Contains(dt)) {
```
keyboardDrummer, line 1088: I'm not sure I understand the purpose of this check. Why can we not always optimise away the constructor at this point?

If I have

```dafny
datatype Wrapper = Wrapper(ints: Ints)
datatype Ints = Ints(x: int, y: int) | AnotherWrapper(wrapper: Wrapper)
```

where it seems `properlyContainedTypeDecls.Contains(dt) == true`, then wouldn't we still be able to write that as

```dafny
datatype Ints = Ints(x: int, y: int) | AnotherWrapper(wrapper: Ints)
```

?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1047,9 +1049,134 @@ public class TypeArgumentInstantiation {
...
+        } else if (member is DatatypeDiscriminator) {
+          // a discriminator of an inductive or coinductive datatype
+          return dt.Ctors.Count(c => !c.IsGhost) == 1 ? MemberCompileStatus.AlwaysTrue : MemberCompileStatus.Ordinary;
+        }
+      }
+
+      return MemberCompileStatus.Ordinary;
     }
+
+    protected bool IsInvisibleWrapper(DatatypeDecl dt, out DatatypeDestructor coreDestructor) {
```
keyboardDrummer, line 1068: What does `invisible` in the name mean? Could we rename this to `ShouldDatatypeBeErased` ?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -426,6 +427,14 @@ public enum IncludesModes {
           WarnShadowing = true;
           return true;
 
+        case "optimizeInvisibleDatatypeWrappers": {
```
keyboardDrummer, line 430: If you want this option to be available in the new CLI UI (and you do), then you have to add a value of `IOptionSpec` to `ICommandSpec.ExecutionOptions`. You can add the same value to `DafnyOptions.AvailableNewStyleOptions` to also make it available in the old UI.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1047,9 +1049,134 @@ public class TypeArgumentInstantiation {
...
+        if (IsInvisibleWrapper(dt, out var dtor)) {
+          var typeSubst = TypeUtil.TypeSubstitutionMap(dt.TypeArgs, udt.TypeArgs);
+          return CanBeLeftUninitialized(TypeUtil.SubstType(dtor.Type, typeSubst));
+        }
+      }
+      return false;
+    }
+
+    /// <summary>
+    /// Remove any invisible type wrappers and simplify ghost typle types.
```
keyboardDrummer, line 1133: `typle`

keyboardDrummer, line 1133: I'm quite looking forward to seeing this be implemented as a separate pass. I think it'll improve the readability a lot.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1047,9 +1049,134 @@ public class TypeArgumentInstantiation {
     protected abstract ILvalue EmitMemberSelect(Action<ConcreteSyntaxTree> obj, Type objType, MemberDecl member, List<TypeArgumentInstantiation> typeArgs, Dictionary<TypeParameter, Type> typeMap,
       Type expectedType, string/*?*/ additionalCustomParameter = null, bool internalAccess = false);
 
-    protected void EmitArraySelect(string index, Type elmtType, ConcreteSyntaxTree wr) {
-      EmitArraySelect(new List<string>() { index }, elmtType, wr);
+    protected enum MemberCompileStatus { Ordinary, Identity, AlwaysTrue }
+
+    protected MemberCompileStatus GetMemberStatus(MemberDecl member) {
```
keyboardDrummer, line 1054: I would move the new members of `SinglePassCompiler`, GetMemberStatus, IsInvisibleWrapper, CompiledTypes and SimplifyType to a separate class with a name like `WrapperTypeEraser`.

Source/DafnyCore/AST/TypeUtil.cs
```diff
@@ -0,0 +1,173 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Linq;
+
+namespace Microsoft.Dafny;
+
+public class TypeUtil {
+  public static Dictionary<TypeParameter, Type> TypeSubstitutionMap(List<TypeParameter> formals, List<Type> actuals) {
```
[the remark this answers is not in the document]
RustanLeino: Oh my! Yes, this is the result of a bad merge. (I had merged the other PR into this one before the other one was merged into master.)

Source/DafnyCore/AST/TypeUtil.cs
```diff
@@ -0,0 +1,173 @@
...
+    for (int i = 0; i < formals.Count; i++) {
+      subst.Add(formals[i], actuals[i]);
+    }
+    return subst;
+  }
+
+  /// <summary>
+  /// If the substitution has no effect, the return value is pointer-equal to 'type'
+  /// </summary>
+  public static Type SubstType(Type type, Dictionary<TypeParameter, Type> subst) {
```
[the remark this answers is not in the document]
RustanLeino: Yep, sorry. Bad merge.

Source/DafnyCore/AST/TupleTypeDecl.cs
```diff
@@ -12,10 +13,12 @@ public class TupleTypeDecl : IndDatatypeDecl {
 
   public int NonGhostDims => ArgumentGhostness.Count(x => !x);
 
+  [CanBeNull] public readonly TupleTypeDecl NonGhostTupleTypeDecl;
```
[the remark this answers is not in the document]
RustanLeino: I made `NonGhostTupleTypeDecl` into a property getter.

I don't understand what you mean to say with the `RuntimeVersion` code snippet.

Source/DafnyCore/AST/BuiltIns.cs
```diff
@@ -233,7 +233,15 @@ public class BuiltIns {
     argumentGhostness ??= new bool[dims].Select(_ => false).ToList();
     if (!tupleTypeDecls.TryGetValue(argumentGhostness, out var tt)) {
       Contract.Assume(allowCreationOfNewType);  // the parser should ensure that all needed tuple types exist by the time of resolution
-      tt = new TupleTypeDecl(argumentGhostness, SystemModule, DontCompile());
+
+      var nonGhostDims = argumentGhostness.Count(x => !x);
+      TupleTypeDecl nonGhostTupleTypeDecl = null;
+      if (nonGhostDims != dims && (nonGhostDims != 1 || !DafnyOptions.O.Compiler.OptimizesInvisibleDatatypeWrappers)) {
```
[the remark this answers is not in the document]
RustanLeino: It will just generate a `Tuple1` class as well. That seems alright. I'll change the test to just `nonGhostDims != dims`.

Source/DafnyCore/Compilers/Compiler-Csharp.cs
```diff
@@ -304,7 +304,7 @@ public class CsharpCompiler : SinglePassCompiler {
       Contract.Requires(wr != null);
 
       var type = UserDefinedType.FromTopLevelDecl(enclosingTypeDecl.tok, enclosingTypeDecl);
-      var initializer = TypeInitializationValue(type, wr, enclosingTypeDecl.tok, false, true);
+      var initializer = DefaultValue(type, wr, enclosingTypeDecl.tok, true);
```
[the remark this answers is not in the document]
RustanLeino: I added a comment, referring to `docs/Compilation/AutoInitialization.md`.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1047,9 +1049,134 @@ public class TypeArgumentInstantiation {
...
+          // a discriminator of an inductive or coinductive datatype
+          return dt.Ctors.Count(c => !c.IsGhost) == 1 ? MemberCompileStatus.AlwaysTrue : MemberCompileStatus.Ordinary;
+        }
+      }
+
+      return MemberCompileStatus.Ordinary;
     }
+
+    protected bool IsInvisibleWrapper(DatatypeDecl dt, out DatatypeDestructor coreDestructor) {
+      if (!DafnyOptions.O.OptimizeInvisibleDatatypeWrappers ||
```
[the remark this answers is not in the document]
RustanLeino: I added some comments in the doc-header of the method.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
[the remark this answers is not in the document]
RustanLeino: That would be a possibility, but I chose the bool-return/out-parameter idiom instead.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1047,9 +1049,134 @@ public class TypeArgumentInstantiation {
...
+          !OptimizesInvisibleDatatypeWrappers ||
+          !(dt is IndDatatypeDecl) ||
+          dt.IsExtern(out _, out _) ||
+          dt.Members.Any(member => member is Field)) {
+        // don't optimize
+      } else {
+        var i = dt.Ctors.FindIndex(ctor => !ctor.IsGhost);
+        if (0 <= i) {
+          if (dt.Ctors.FindIndex(i + 1, ctor => !ctor.IsGhost) == -1) {
+            // there is exactly one non-ghost constructor
```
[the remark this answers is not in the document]
RustanLeino: My new doc-comment about the method address this now.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1047,9 +1049,134 @@ public class TypeArgumentInstantiation {
...
+            var ctor = dt.Ctors[i];
+            var j = ctor.Destructors.FindIndex(dtor => !dtor.IsGhost);
+            if (0 <= j) {
+              if (ctor.Destructors.FindIndex(j + 1, dtor => !dtor.IsGhost) == -1) {
+                // there is exactly one non-ghost parameter to "ctor"
+                var candidateCoreDestructor = ctor.Destructors[j];
+                var properlyContainedTypeDecls = new HashSet<TopLevelDecl>();
+                CompiledTypes(candidateCoreDestructor.Type, properlyContainedTypeDecls);
```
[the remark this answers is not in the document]
RustanLeino: I renamed it to `GetCompiledComponentTypes` and improve its doc-comment.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
RustanLeino: Very good catch! The current algorithm is too strict and misses the opportunity to optimize your `Wrapper` type. I'll improve the algorithm.

As for the question about the purpose of the check, here is an answer: If optimizing away the constructor gives back the type itself, then there would be a problem. To illustrate, it may help to imagine a type like

``` dafny
datatype Cycle = Ctor(Cycle)
```

It would not be possible to remove the wrapper `Ctor`, because doing so would give another type whose wrapper `Ctor` we'd remove, and it is also a wrapper `Ctor` that we'd remove, etc. However, Dafny doesn't actually allow an `datatype` like `Cycle`, so to illustrate the real problem, I have to show a more complicated type. Here is one:

``` dafny
datatype GList =
  | ghost End
  | More(GList)
```

Dafny does allow this type. We don't represent `End` at run time, since it's ghost. The other constructor has one argument, so it looks like it could be optimized. But this would then run into the infinite-cycle problem I gave a taste of with the `Cycle` type above.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
RustanLeino: With "invisible wrapper", I meant that the single compiled constructor is "erasable". I considered renaming it `IsEraseableDatatype`, but is a bit misleading, since it's only the "wrapper" part that's erased--not the entire datatype value. I'll rename it `IsEraseableDatatypeWrapper` and will do the same renaming for the test file and command-line option.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
RustanLeino: I moved them. The way I did it, the qualified names now contain some redundancy.

Source/DafnyCore/DafnyOptions.cs
RustanLeino: Indeed, I do. I changed it. Well, I don't understand if the `Description` should talk about arguments `0` and `1` or arguments `false` and `true`.

Source/DafnyCore/AST/BuiltIns.cs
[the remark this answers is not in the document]
RustanLeino: The suggested change worked. However, in the end, I decided to revert it to the code above. That is, the test does include `nonGhostDims != 1 || !DafnyOptions.O.Compiler.OptimizesInvisibleDatatypeWrappers` again, so that we will _not_ generate a `Tuple1` class. It would seem odd to generate a class that we know is never going to be used.

Source/DafnyCore/DafnyOptions.cs
keyboardDrummer: > Well, I don't understand if the Description should talk about arguments 0 and 1 or arguments false and true

Yes that's a problem. I'm adding the option of having two descriptions here: https://github.com/dafny-lang/dafny/pull/3154

keyboardDrummer (approved):

Source/DafnyCore/Compilers/DatatypeWrapperEraser.cs
```diff
@@ -0,0 +1,193 @@
...
+      }
+      coreDtor = null;
+      return false;
+    }
+
+    /// <summary>
+    /// Return "true" if a traversal into the components of "type" finds "lookingFor" before passing through any type in "visited".
+    /// "lookingFor" is expected not to be a subset type, and "visited" is expected not to contain any subset types.
+    /// </summary>
+    private static bool CompiledTypeContains(Type type, TopLevelDecl lookingFor, ISet<TopLevelDecl> visited) {
```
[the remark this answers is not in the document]
RustanLeino: Nice suggestion. I'll be sure to do that when I have ISet's in the future.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -1047,9 +1049,134 @@ public class TypeArgumentInstantiation {
...
+        var i = dt.Ctors.FindIndex(ctor => !ctor.IsGhost);
+        if (0 <= i) {
+          if (dt.Ctors.FindIndex(i + 1, ctor => !ctor.IsGhost) == -1) {
+            // there is exactly one non-ghost constructor
+            var ctor = dt.Ctors[i];
+            var j = ctor.Destructors.FindIndex(dtor => !dtor.IsGhost);
+            if (0 <= j) {
+              if (ctor.Destructors.FindIndex(j + 1, dtor => !dtor.IsGhost) == -1) {
+                // there is exactly one non-ghost parameter to "ctor"
+                var candidateCoreDestructor = ctor.Destructors[j];
```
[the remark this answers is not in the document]
RustanLeino: I thought, in my old-school brain, that the indices were clear enough for this task. Your suggested names would be much better than my `i` and `j`.

Yes, it was the creation of a new `List` that I was avoiding. Then again, I imagine the creation of the various lambda expressions (like `d => !d.IsGhost`) will take as much time and space as an additional `List`, so not much is saved. Because I used `firstDestructorIndex + 1` as the starting point, I believe the list would be traversed only once, but unless the C# compiler is good at optimizing, it would create a lambda expression for each invocation of `FindIndex`.

Regarding copying the entire AST and omitting ghost things: I'm very much in favor of this. A lot of the code that is duplicated across the compilers has to do with avoiding ghosts, so that would be so nice to do once and for all.

PR #3057, opened 2022-11-15
refactor: Move Resolver.{TypeSubstitutionMap, SubstType}

RustanLeino:
This PR just moves two type-related utility methods from `Resolver`:
* `Resolver.SubstType` -> abstract instance method `Type.Subst`
* `Resolver.TypeSubstitutionMap` -> `TypeParameter.SubstitutionMap`


Source/DafnyCore/AST/TypeUtil.cs
```diff
@@ -0,0 +1,140 @@
...
+    for (int i = 0; i < formals.Count; i++) {
+      subst.Add(formals[i], actuals[i]);
+    }
+    return subst;
+  }
+
+  /// <summary>
+  /// If the substitution has no effect, the return value is pointer-equal to 'type'
+  /// </summary>
+  public static Type SubstType(Type type, Dictionary<TypeParameter, Type> subst) {
```
keyboardDrummer, line 23: Consider making this an extension method, so you can write `someType.Substitute(substitutionMap)`

Even better would be to have this be an abstract method defined on `Type`, so the code is more local and there's a static check against forgetting to add cases for this behavior.

Source/DafnyCore/AST/TypeUtil.cs
```diff
@@ -0,0 +1,140 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Linq;
+
+namespace Microsoft.Dafny;
+
+public class TypeUtil {
+  public static Dictionary<TypeParameter, Type> TypeSubstitutionMap(List<TypeParameter> formals, List<Type> actuals) {
```
keyboardDrummer, line 9: Consider removing `Type` from the name since that's already part of `TypeUtil`. Also consider moving this to `TypeParameter` since it strongly related to it.

keyboardDrummer (dismissed):

keyboardDrummer (approved):

Source/DafnyCore/AST/Expressions.cs
```diff
@@ -1183,6 +1183,9 @@ public abstract class ResolverType : Type {
     public override bool ComputeMayInvolveReferences(ISet<DatatypeDecl>/*?*/ visitedDatatypes) {
       return false;
     }
+    public override Type Subst(Dictionary<TypeParameter, Type> subst) {
+      throw new NotImplementedException();
```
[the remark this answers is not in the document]
RustanLeino: Good point. Since this PR was already auto-merged, I will incorporate this change in PR https://github.com/dafny-lang/dafny/pull/3056, which had triggered this one.

PR #2059, opened 2022-04-23
fix: Use correct type for let expression generated by datatype update

cpitclaudel:
Closes GH-1479.

Co-authored-by: @RustanLeino

PR #3060, opened 2022-11-15
Fix: Use correct type for map update expression

### LINKED ISSUE #3059: [Bug]: subset type for maps fails to verify overriden key/value pairs
### Dafny version

3.9.1

```dafny
type foo = m: map<int, int> | forall n <- m.Keys :: m[n] < 5

function addToFoo(m: foo): foo
  ensures false
{
  m[1 := 7]
}
```

```code
dafny verify problem

Dafny produces 1 verified and 0 error.
```

### What happened?

It should have verified that `m[1 := 7]` does not have necessarily the same type as `m` itself.

Linux, Windows, Mac, WSL

MikaelMayer:
This PR fixes #3059
The solution was similar to the one implemented in https://github.com/dafny-lang/dafny/pull/2059
It consists in not setting the type of the expression `m[k := v]` as the same type of `m`, but as the partially resolved type of `m` (whatever this means, I just took the code in other branches).
I added a corresponding test.

This PR also introduces the command-line argument `--nonzero-exit-code-if-errors` set by default to `true` so that we can use the new Dafny CLI with `%baredafny verify %args_0` for files that are supposed to fail.


Test/git-issues/git-issue-3059.dfy
```diff
@@ -0,0 +1,10 @@
+// RUN: %baredafny verify %args_0 "%s" > "%t"
```
MikaelMayer, line 1: @keyboardDrummer  yay ! we can use the new CLI for tests now.

Source/DafnyCore/Resolver.cs
```diff
@@ -15074,7 +15074,8 @@ public ResolutionContext(ICodeContext codeContext, bool isTwoState)
         ResolveExpression(e.Index, resolutionContext);
         ResolveExpression(e.Value, resolutionContext);
         AddXConstraint(expr.tok, "SeqUpdatable", e.Seq.Type, e.Index, e.Value, "update requires a sequence, map, or multiset (got {0})");
-        expr.Type = e.Seq.Type;
+        var ty = PartiallyResolveTypeForMemberSelection(expr.tok, e.Seq.Type);
+        expr.Type = ty;
```
fabiomadge, line 15078:
```suggestion
        expr.Type = PartiallyResolveTypeForMemberSelection(expr.tok, e.Seq.Type);
```

cpitclaudel, line 15078: This does not look right to me — there's no member selection here.  Instead, shouldn't we just revert to the base type? (`e.Seq.Type` with the constraints removed?)

Source/DafnyDriver/BoogieAffectingOptions/NonZeroExitCodeIfErrorsOption.cs
```diff
@@ -0,0 +1,18 @@
...
+public class NonZeroExitCodeIfErrorsOption : BooleanOption {
+  public static readonly NonZeroExitCodeIfErrorsOption Instance = new();
+  public override object DefaultValue => true;
+  public override string LongName => "nonzero-exit-code-if-errors";
+  public override string ShortName => null;
+  public override string ArgumentName => "boolean";
+
+  public override string Description => @"
+if false then always exit with a 0 exit code, regardless of whether errors are found.
+If true (default) then use the appropriate exit code.".TrimStart();
```
cpitclaudel, line 12: It's a bit uncommon to have such a flag.  Shouldn't we find another solution in the test suite itself?

cpitclaudel (changes requested):

Source/DafnyDriver/BoogieAffectingOptions/NonZeroExitCodeIfErrorsOption.cs
MikaelMayer: We already have this option in a substantial portion of our 1400+ tests, whenever you have %dafny_0, it means you have this flag.
Since `%baredafny verify` only takes new arguments, we need to expose this option to make it work.
I don't know of another simple way.

What we could do is hide this option to users, but I don't know how to do that. @keyboardDrummer ?

cpitclaudel:
> I think the way to fix this problem is to create two new type proxies (call them `resultKeyType` and `resultValueType`) and set `expr.Type` to a new `MapType` using those two type proxies as type arguments. Then, add a constraint that `expr.Type` is a subtype of that new `MapType`.

That would be nice, but it's not that simple (I think?): we don't know yet that `m` is a map, either.  At that point in the program all we know is that is should satisfy `SeqUpdatable`.  But I think `InferredTypeProxy`s drop constraints by default, so wouldn't that do?

Source/DafnyDriver/BoogieAffectingOptions/NonZeroExitCodeIfErrorsOption.cs
cpitclaudel: But the description of that flag says "deprecated" in the old CLI:

```
  /countVerificationErrors:<n>
      (deprecated)
      0 - Set exit code to 0 regardless of the presence of any other errors.
      1 (default) - Emit usual exit code (cf. beginning of the help message).
```

cpitclaudel: I think for this test we should just use the old CLI.  And we can then (in a separate PR) enhance LitTests.cs to accept `|| true` at the end of a command line, which is the "right" way to ignore an exit code.  WDYT?

MikaelMayer:
> The proposed fix does not eliminate the problem. Here is an example where it remains:

I just tested the proposed fix against your example and it worked. Proof:

image


```

RustanLeino (approved):

cpitclaudel (approved):

PR #3065, opened 2022-11-17
Remove one instance of static state in the compilation pipeline

### LINKED ISSUE #2945: Post-mortem/investigation on #2828
https://github.com/dafny-lang/dafny/pull/2790 introduced a case ([issue ref](https://github.com/dafny-lang/dafny/issues/2828)), where the resolver crashed on an invalid Dafny program. This kind of regression is especially disruptive for those using the VSCode extension, since it crashes the extension completely, and it often isn't obvious to users what part of their code is triggering the bug so it looks like the extension is 100% broken.

Let's look into a few questions:

1. Are there general mechanisms for making the extension and/or language server more robust on a program that can't be resolved without an exception?
2. Are there general testing strategies for catching such bugs in the first place? Perhaps we could leverage https://github.com/dafny-lang/xdsmith to fuzz test by generating invalid programs.
3. Could we make the extension automatically offer to revert back to a previous version of Dafny when this happens?

keyboardDrummer:
Fixes #2945

### Changes

Remove an instance of static state in the compilation pipeline. Such static state is bad because the pipeline may be run multiple times, causing subsequent runs to be affected by previous ones.

### Testing

- TODO, but out of scope of this PR: change our integration tests to run all the tests in a single Dafny process.


MikaelMayer (commented):
Really !this variable was static?! Well spotted.
Approved, happy to merge, but could you please add a release note about this bug fix?

MikaelMayer (approved):
Well done

PR #3067, opened 2022-11-17
Chore: Support for --nonzero-exit-code

MikaelMayer:
Until today, all our integration tests are using the old Dafny CLI syntax
This has the following disadvantages:
- We want to fully replace old Dafny CLI syntax, yet, all our tests files are written using this command.
- Tests files are often used as documentation for some features.
- The new Dafny CLI is not as much tested compared to the previous one, so might lack components we might not be aware of.

In this PR, I propose that we move the needle by enabling all the integration tests to run using the new Dafny CLI syntax.
One of the major blocker is the existence of tests that rely on Dafny to generate an error while returning a zero exit code. These can't currently be reproduced in the new CLI because it does not have the necessary option.

My contribution in this PR is to:
- Add an option to replicate in the new CLI an option that was available in the old CLI that we need for tests
- Add a lightweight syntax to refer to this option without mentioning it in the lit format.

Until now, we were using the current option `-countVerificationErrors:0` for these tests. This option is embedded when we write `%dafny_0`. The doc says this option is "deprecated",  but we haven't found a correct workaround to this, have we? 
Some tests are expecting failures, some don't, so it's not like we can change the XUnit test to ignore nonzero exit codes.
To my sense this option should only be hidden to the regular user, not deprecated.

This PR thus does not introduce anything new but simply exposes the equivalent option to the new CLI, named explicitly `--nonzero-exit-code-if-errors`, with a boolean value equals to true by default. We really want users who include this option to completely write `--nonzero-exit-code-if-errors=false`, as this is a behavior that no one would expect - except for the test cases we are running.
I would like to know how to hide it from the regular user @keyboardDrummer as well, if there is such a mechanism, because as of today, this option shows up when running `dafny verify --help`

For the tests themselves, I modify the lit engine so that if we use `%args_0` instead of `%args`, it will include `--nonzero-exit-code-if-error=false`.
Thus, we can now transform

| Previously | Now |
| ------------- | ------------- |
| `%dafny /compile:0 "%s" > "%t"`  | `%baredafny verify %args "%s" > "%t"` |
| `%dafny_0 /compile:0 "%s" > "%t"`  | `%baredafny verify %args_0 "%s" > "%t"` **Missing until this PR**|
|  `%dafny /compile:3 "%s" > "%t"`  | `%baredafny run %args "%s" > "%t"`  |
|  `%dafny /compile:4 "%s" > "%t"`  | `%baredafny run %args --no-verify "%s" > "%t"`  |

### DIFF
--- a/Source/DafnyDriver/BoogieAffectingOptions/NonZeroExitCodeIfErrorsOption.cs
+++ b/Source/DafnyDriver/BoogieAffectingOptions/NonZeroExitCodeIfErrorsOption.cs
@@ -1,0 +1,18 @@
+namespace Microsoft.Dafny;
+
+public class NonZeroExitCodeIfErrorsOption : BooleanOption {
+  public static readonly NonZeroExitCodeIfErrorsOption Instance = new();
+  public override object DefaultValue => true;
+  public override string LongName => "nonzero-exit-code-if-errors";
+  public override string ShortName => null;
+  public override string ArgumentName => "boolean";
+
+  public override string Description => @"
+if false then always exit with a 0 exit code, regardless of whether errors are found.
+If true (default) then use the appropriate exit code.".TrimStart();
+
+  public override string PostProcess(DafnyOptions options) {
+    options.CountVerificationErrors = Get(options);
+    return null;
+  }
+}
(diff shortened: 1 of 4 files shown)


keyboardDrummer:
I think I prefer having to add `|| true` after the test line if you want to halt on a non-zero exit code, since it's more explicit than replacing `%args` with `%args_0`

MikaelMayer:
Closing in favor of || true or !

PR #3072, opened 2022-11-18
feat: support for traits as type arguments with variance on datatypes in Java

### LINKED ISSUE #2013: Feature request: support for traits as type arguments with variance on datatypes in Java
Follow-up from #1499, which looks like it was resolved by #1578 (which only added support in C#) accidentally.

RustanLeino:
This PR removes that pesky "not supported" error for Java where it wouldn't allow a `trait` to be passed in as a variant type parameter of a (co)datatype.

Fixes #2013

The solution is pretty simple. Basically, it just emits casts when needed. To be accepted by Java's static type checker, these casts first upcast to `Object` and then downcast to the desired type. Since the JVM doesn't keep track of type parameters, it'll never know what hit it. (In particular, this means that we don't need the expensive `DowncastClone` mechanism that we're using for C# to satisfy .NET's more informed runtime.)

To make it work, this PR also makes more precise the `IsTargetSupertype` method for Java. In particular, to check if `A<X, Y>` is a supertype of `B<U, V, W>`, first keep converting the latter up the `trait` parents of `B` until it gets to `A` (if ever). This is already done for all the compilers. Call the result `A<X', Y'>`. For the other compilers, one would now start to see if the parameters `X, Y` are target supertypes of `X', Y'` with variance in mind (for example, if the first type parameter is contravariant, we would actually check if `X'` is a supertype of `X`). For Java, however, we continue with variance in mind only if `A` is a collection type (because those already have run-time support for variance, using Java's bounded existential types). For non-collection types, we instead check if `X, Y` are target-type equal to `X', Y'`. If they are not, the compiler calls `EmitDowncast`. (To read these in the code, look at `EmitDowncastIfNecessary` in `SinglePassCompiler.cs`.)

We could have used simple casts like this for collection types, too. But since I already have those, I left those alone.


Test/git-issues/git-issue-2013.dfy
```diff
@@ -0,0 +1,191 @@
...
+
+    var S: set<Z> := {c};
+    var T: set<X> := S;
+    var U: set<Y> := S;
+    T := U;
+    U := T;
+    print S, " ", T, " ", U, "\n"; // {_module.C} {_module.C} {_module.C}
+
+    // The assignments in the following example don't need covariance of Dt, because
+    // Dafny checks that the value with its payment is suitable for the target type.
```
fabiomadge, line 135:
```suggestion
    // Dafny checks that the value with its payload is suitable for the target type.
```
?

Test/git-issues/git-issue-2013.dfy
```diff
@@ -0,0 +1,191 @@
...
+  class C extends X, Y { }
+
+  method Test() {
+    var c := new C;
+
+    var s: set<int> := {100};
+    var t: set<nat> := s;
+    var u: set<Even> := s;
+    t := u;
+    u := t;
```
fabiomadge, line 124:
```suggestion
    s := t;
```
Both seem like a decent idea.

docs/dev/news/3072.feat
```diff
@@ -0,0 +1 @@
+feat: support for traits as type arguments with variance on datatypes in Java
```
fabiomadge, line 1:
```suggestion
feat: support for traits as type arguments by allowing variance on datatypes in Java
```

fabiomadge (dismissed):
I did not expect this to be as simple. Good for us!

docs/dev/news/3072.feat
RustanLeino: I also add "fully allowing"

Test/git-issues/git-issue-2013.dfy
RustanLeino: For further testing, I prefer `u`, since the types `Even` and `nat` are not subtypes of each other.

RustanLeino: I added your test, too (here and in a few more places).

Source/DafnyCore/Compilers/Compiler-java.cs
```diff
@@ -2623,17 +2628,17 @@ private class GenericArrayElementLvalue : ILvalue {
...
+      var typeParams = typeArgs.Count == 0 ? "" : $"<{BoxedTypeNames(typeArgs, wr, dt.tok)}>";
       if (!isCoCall) {
         // For an ordinary constructor (that is, one that does not guard any co-recursive calls), generate:
         //   Dt.<T>create_Cons( args )
         wr.Write($"{dtName}.{typeParams}{DtCreateName(ctor)}({arguments})");
       } else {
         wr.Write($"new {dt.CompileName}__Lazy(");
         wr.Write("() -> { return ");
-        wr.Write($"new {DtCtorName(ctor)}({arguments})");
+        wr.Write($"new {DtCtorName(ctor)}{typeParams}({arguments})");
```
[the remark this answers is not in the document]
RustanLeino: Good find! It was handled incorrectly in 3 compilers. I fixed it.

fabiomadge (approved):

PR #3081, opened 2022-11-20
Small adjustments to discussion of array types

### LINKED ISSUE #2863: Documentation error re arrays
@Cok, David [C], there is a problem in the documentation that can be easily fixed for that. The first meaningful mention of the array type array<T> is in an obscure caveat section of 15.1, well after the section where arrays are initially introduced. Do you think it would make sense to affix the variable `res` with `array<int>`, so that users have a first impression of what the array type should look like?

davidcok:
Fixes #2863.


prvshah51 (approved):
LGTM!

PR #3087, opened 2022-11-21
doc: #2864 - adding explanation of RHS expressions

### LINKED ISSUE #2864: Documentation error
9/30 (John Tristan)

I think I might have spotted an error in the reference manual, but I'm concerned that I might have missed something so I would like to ask about it before creating an issue. In the reference manual, [array allocation](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-array-allocation) and [object allocation](https://dafny.org/dafny/DafnyRef/DafnyRef#2117-object-allocation) are listed as being expressions. However, it seems wrong because:

    Viewing them as expressions doesn't match the rest of the semantics of expressions
    Using them as an expression in an specification clause is an error (with a message saying "invalid LogicalExpression" which is confusing BTW, because LogicalExpression is a specific entry in the BNF)
    I looked at the parser implementation which seems to indeed treat them as statements

Am I missing something? (edited) 

--------
I think the intent of the manual is to say that they are only "Right-Hand-Side Expressions", a special kind of expressions that can only be used to the right of a := or :- and in a return or a yield.  I agree that it's a touch confusing that Expressions are a kind of Right-hand-side expression, rather than the reverse, so maybe we could rename "21.15. Right-Hand-Side Expressions" to "21.15. Right-Hand-Sides" to clarify?
------
Yes, I think it would be reasonable
-----
Two clarifications:

    They can be used as RHS of := when it is a variable assignment, but not a let-binding
    The RHS BNF entry where these allocations appear in the parser implementation produce statements, not expressions

-----
Although interestingly, for the array it does produce an expression
-----
Isn't the problem really that there are two kinds of expressions, the ones that are logical formulas, and the terms that appear in statements (including references)?  They share a lot of the same BNF, but they're not completely identical either. For example, a return cannot return an old expression. (edited) 
-----
To me, the manual would be a lot clearer if it carefully distinguished between those.
-----
Even if we don't want to change the presentation of the syntax in the reference manual, it would help to present some typing rules to specify which subset of expressions are allowed in which context. What do you think?

-----
Although one argument against this last idea is that it is actually Dafny's parser that segregates these expressions.
-----
Anything that makes it clearer is what I like !
-----
BTW, Mikael, I initially assumed that this is what allowlemma was intended to do, but I can't quite make sense of it
-----
For this user-facing documentation, the grammar should not even be visible, but on a separate file. Or at least at the end of each section. Only examples should be provided with explanations, at the beginning. (edited) 
-----
Changing the presentation of the grammar in the RM is an outstanding task of mine.But to the original question: is it correct that object/array allocation is permitted syntactically just where a method invocation is permitted? Or should it be?
-----
There's a useful comment [here](https://dafny.org/dafny/DafnyRef/DafnyRef#2115-right-hand-side-expressions) : Right-hand-side expressions appear in the following constructs: ReturnStmt, YieldStmt, UpdateStmt, UpdateFailureStmt, or VarDeclStatement. These are the only contexts in which arrays or objects may be allocated, or in which havoc may be produced.
-----
 To me, the bottom line is this: I think it would be highly valuable to make it as easy as possible for the reader to understand what is Dafny's language of logical formulas (but also what is the language of expressions that makes sense in the context of writing code)
-----
I would even go as far as layering the specification of logical formulas so that we first describe the subset that correspond to higher-order logic, and then the full language
-----
I think that layering is already there, but maybe not clear enough?  There are statements, RHS expressions, and expressions.  Expressions are a subset of RHS expressions.
-----
I don't understand the last claim. old(3) is an expression, but it is not an RHS expression, right?

davidcok:
Fixes #2864


MikaelMayer (commented):
One worry to address please. Otherwise, good writing !

MikaelMayer (dismissed):
One minor thing you don't need to address. Good job !

PR #3093, opened 2022-11-22
Ensure non-parseable SeqIndexs do not throw on parse failure

### LINKED ISSUE #3048: [Bug]: Non-integer Seq#Indexes lead to LSP crash
### Dafny version

3.9.1.41027

https://gist.github.com/dijkstracula/b2863891f53650fcf4607abd2b35991f

```code
Trying to extract a counterexample leads to a BigInteger parsing exception.
```

### What happened?

The issue is that not every Seq#index appears to be a `Model.Integer`: in particular, the above code generates a Seq such that one of the "indices" is in fact a `DatatypeValue`:  here's such an example

image

With the BigInteger parsing code removed, the counterexample that crashes is:

```
At "{" (file:///Users/ntaylor/school/smrt/src/Disks/CowZonedDisk.dfy:173):
...
    @1:seq<int> = (Length := 2, [2] := @10)
    @2:ZonedDisk.State = State(zones := @11)
    @11:seq<Zone.State> = (Length := 3, [(- 1)] := @15, [0] := @16, [2] := @17)
    @15:? = ()
```

This makes sense: the string `(- 1)` can't be correctly parsed as a BigInteger, so the exception being thrown is what we expect.  

I'm not sure that removing the BigInteger parsing code is completely the correct solution, in that I as a user don't really know how one is meant to read the meaning of the above counterexample: is the non-integral index morally index `1`?  If so, that value should be patched up before handing it up towards the user.  An index of `-1` doesn't make sense, so that's unlikely to be the intended value.  And, if it's an indirect reference to something else in the counterexample context, it too should be patched up.

I understand that `-` and `/` refer to constructors at some level of the Dafny stack, but don't really understand their meaning here.  If the "fix" is to simply check the type of the Seq#index and return `Arguments[1]` if it's a `DatatypeValue` then I can put that patch out; however, it feels slightly hacky and I get the sense that there is a more principled solution that probably lies farther down the stack (or potentially even in Boogie?).

Mac

dijkstracula:
Occasionally the language server is passed a seq index that is not parseable as an integer.  While the root cause has not yet been determined, it's still causing an LSP crash.  This patch ensures that non-parseable indices are handled separately from those that are parseable.

After this patch, any non-integral indices are appended to the end of the sorted output: [the crashing test](https://gist.github.com/dijkstracula/b2863891f53650fcf4607abd2b35991f) in #3048 now produces,

```
@4:seq<seq<seq<_module.uint8>>> = (Length := 9899, [0] := @14, [8100] := @15, [9898] := @16, [(- 1)] := @17)
```

Fixes #3048

### DIFF
--- a/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
+++ b/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
@@ -1106,9 +1106,9 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
         /* Then, extract the number of non-integral indexed sequences from the repro case... */
         .Count(IsNegativeIndexedSeq);
 
-      Assert.IsTrue(nonIntegralIndexedSeqs > 0,
-"If we do not see at least one non-integral index in this test case, then the backend changed " +
-        "The indices being returned to the Language Server.");
+      Assert.IsTrue(nonIntegralIndexedSeqs > 0, "If we do not see at least one non-integral index in " +
+                                                "this test case, then the backend changed " +
+                                                "The indices being returned to the Language Server.");
     }
 
     /* Helper for the NonIntegerSeqIndices test. */
--- a/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
+++ b/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
@@ -1,13 +1,12 @@
 ﻿using System.Collections.Generic;
+using System.IO;
 using Microsoft.Dafny.LanguageServer.Handlers.Custom;
 using Microsoft.Dafny.LanguageServer.IntegrationTest.Extensions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using OmniSharp.Extensions.LanguageServer.Protocol;
-using OmniSharp.Extensions.LanguageServer.Protocol.Client;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
-using DafnyServer.CounterexampleGeneration;
 using Microsoft.Dafny.LanguageServer.IntegrationTest.Util;
 using Microsoft.Dafny.LanguageServer.Workspace;
 
@@ -1086,6 +1085,47 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
       Assert.IsTrue(counterExamples[0].Variables["c:M.C?"] is
         "(c1 := '\\u1023', c2 := '\\u1023')" or
         "(c2 := '\\u1023', c1 := '\\u1023')");
+    }
+
+    /// <summary>
+    /// Tests that a Dafny program where a sequence with non-integral indices is generated as part of
+    /// a counterexample.  This would previously crash the LSP before #3093.
+    /// For more details, see https://github.com/dafny-lang/dafny/issues/3048 .
+    /// </summary>
+    [TestMethod]
+    public async Task NonIntegerSeqIndices() {
+      string fp = Path.Combine(Directory.GetCurrentDirectory(), "Various", "TestFiles", "3048.dfy");
+      var source = await File.ReadAllTextAsync(fp, CancellationToken);
+      var documentItem = CreateTestDocument(source);
+
+      await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+
+      /* First, ensure we can correctly extract counterexamples without crashing... */
+      var nonIntegralIndexedSeqs = (await RequestCounterExamples(documentItem.Uri))
+        .SelectMany(cet => cet.Variables.ToList())
+        /* Then, extract the number of non-integral indexed sequences from the repro case... */
+        .Count(IsNegativeIndexedSeq);
+
+      /* If we do not see at least one, then the backend changed the indices being returned to the Language Server. */
+      Assert.IsTrue(nonIntegralIndexedSeqs > 0);
+    }
+
+    /* Helper for the NonIntegerSeqIndices test. */
+    private static bool IsNegativeIndexedSeq(KeyValuePair<string, string> kvp) {
+      var r = new Regex(@"\[\(- \d+\)\]");
+      return kvp.Key.Contains("seq<_module.uint8>") && r.IsMatch(kvp.Value);
+    }
+
+    [TestMethod]
+    public void TestIsNegativeIndexedSeq() {
+      Assert.IsFalse(
+        IsNegativeIndexedSeq(new KeyValuePair<string, string>("uint8", "42")));
+      Assert.IsFalse(
+        IsNegativeIndexedSeq(new KeyValuePair<string, string>("seq<_module.uint8>", "(Length := 42, [0] := 42)")));
+      Assert.IsTrue(
+        IsNegativeIndexedSeq(new KeyValuePair<string, string>("seq<_module.uint8>", "(Length := 9899, [(- 1)] := 42)")));
+      Assert.IsTrue(
+        IsNegativeIndexedSeq(new KeyValuePair<string, string>("seq<seq<_module.uint8>>", "(Length := 1123, [(- 12345)] := @12)")));
     }
   }
 }
--- a/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
+++ b/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
@@ -1106,7 +1106,6 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
         /* Then, extract the number of non-integral indexed sequences from the repro case... */
         .Count(IsNegativeIndexedSeq);
 
-      /* If we do not see at least one, then the backend changed the indices being returned to the Language Server. */
       Assert.IsTrue(nonIntegralIndexedSeqs > 0,
 "If we do not see at least one non-integral index in this test case, then the backend changed " +
         "The indices being returned to the Language Server.");
--- a/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
+++ b/Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
@@ -1107,7 +1107,9 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Various {
         .Count(IsNegativeIndexedSeq);
 
       /* If we do not see at least one, then the backend changed the indices being returned to the Language Server. */
-      Assert.IsTrue(nonIntegralIndexedSeqs > 0);
+      Assert.IsTrue(nonIntegralIndexedSeqs > 0,
+"If we do not see at least one non-integral index in this test case, then the backend changed " +
+        "The indices being returned to the Language Server.");
     }
 
     /* Helper for the NonIntegerSeqIndices test. */
--- a/Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
+++ b/Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
@@ -18,79 +18,75 @@ type Zone = seq<Block>
 
 /* a zone map is a sequence of contiguous half-open intervals. */
 
-predicate zone_map_well_formed(n_blocks: uint64, zone_map: seq<(uint64, uint64)>)
-{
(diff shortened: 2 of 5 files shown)


docs/dev/news/3048.fix
```diff
@@ -0,0 +1 @@
+Do not sort SeqIndices that are not parseable as integers
```
MikaelMayer, line 1:
```suggestion
Counter-examples with non-integer sequence indices do not crash Dafny anymore.
```

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -772,18 +772,29 @@ public class DafnyModel {
...
-        IEnumerable<(Model.Element, BigInteger)> tpls =
-          fSeqIndex.AppsWithArg(0, var.Element)
-          .Select(tpl => (Unbox(tpl.Result), BigInteger.Parse(tpl.Args[1].ToString())))
-          .OrderBy(tpl => tpl.Item2);
-
-        foreach (var (res, idx) in tpls) {
-          var idxstr = idx.ToString();
-          var e = DafnyModelVariableFactory.Get(state, res, "[" + idxstr + "]", var);
+        // Otherwise, sequences can be reconstructed index by index, ensuring indices are in ascending order.
+        // NB: per https://github.com/dafny-lang/dafny/issues/3048 , not all indices may be parsed as a BigInteger,
```
MikaelMayer, line 776: If your file with the reproducing example is MIT-licensed, would you mind adding it as a test case in DafnyLanguageServer.Text/Various/TestFiles and ensure this PR is tested? I think it's an excellent file that would be good to keep around, even if it is big.

MikaelMayer (changes requested):
Wonderful ! Feels good to have more safety guard. I have two suggestions, let me know if you need clarification or have better ideas.

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
dijkstracula: Sure, I was going to do that but wondered if it was too long to do so, so I never followed through - will do.

dijkstracula: Hi @MikaelMayer , the latest commit includes the repro case as a unit test; let me know if isn't quite wired up in the way you'd like.

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+type Zone = seq<Block>
+
+/* a zone map is a sequence of contiguous half-open intervals. */
+
+predicate zone_map_well_formed(n_blocks: uint64, zone_map: seq<(uint64, uint64)>)
+{
+    && 0 < |zone_map| as int
+    && zone_map[0].0 == 0
+    && zone_map[|zone_map| - 1].1 == n_blocks
+    && zone_map_ordered_total(zone_map)
```
MikaelMayer, line 26:
```suggestion
predicate zone_map_well_formed(n_blocks: uint64, zone_map: seq<(uint64, uint64)>) {
  && 0 < |zone_map| as int
  && zone_map[0].0 == 0
  && zone_map[|zone_map| - 1].1 == n_blocks
  && zone_map_ordered_total(zone_map)
```

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+}
+
+predicate zone_map_ordered_total(zone_map: seq<(uint64, uint64)>)
+{
+    && 0 < |zone_map| as int
+    && (forall i :: 0 <  i < |zone_map| ==> zone_map[i-1].1 == zone_map[i].0)
+    && (forall i :: 0 <= i < |zone_map| ==> zone_map[i].0 < zone_map[i].1)
+    && (forall i :: 0 <= i < |zone_map| ==> (zone_map[i].1 - zone_map[i].0) as int < UINT64_MAX)
```
MikaelMayer, line 34:
```suggestion
predicate zone_map_ordered_total(zone_map: seq<(uint64, uint64)>) {
  && 0 < |zone_map| as int
  && (forall i :: 0 <  i < |zone_map| ==> zone_map[i-1].1 == zone_map[i].0)
  && (forall i :: 0 <= i < |zone_map| ==> zone_map[i].0 < zone_map[i].1)
  && (forall i :: 0 <= i < |zone_map| ==> (zone_map[i].1 - zone_map[i].0) as int < UINT64_MAX)
```

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/* State transitions take Constants as, well, constant. */
+
+datatype Constants = Constants(
+    n_blocks: uint64,
+    zone_map: seq<(uint64, uint64)> // [b, e)
```
MikaelMayer, line 41:
```suggestion
  n_blocks: uint64,
  zone_map: seq<(uint64, uint64)> // [b, e)
```

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+datatype State = State(
+        /* A zone is a contiguous region of a disk. */
+        zones: seq<Zone>,
+
+        /* Which buffer will we fault the next write's zone over into? */
+        buffer_zone: int,
+
+        /* Logical zone -> physical zone remapping (to account for the
+         * buffer zone, which will float around as writes happen) */
+        buffer_map: seq<int>
```
MikaelMayer, line 55:
```suggestion
  /* A zone is a contiguous region of a disk. */
  zones: seq<Zone>,

  /* Which buffer will we fault the next write's zone over into? */
  buffer_zone: int,

  /* Logical zone -> physical zone remapping (to account for the
   * buffer zone, which will float around as writes happen) */
  buffer_map: seq<int>
```

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+)
+
+predicate Valid(c: Constants, s: State)
+{
+    && zone_map_well_formed(c.n_blocks, c.zone_map)
+    && |s.zones| == |c.zone_map|
+    && (forall z :: 0 <= z < |s.zones| ==>
+        |s.zones[z]| == (c.zone_map[z].1 - c.zone_map[z].0) as int)
+}
```
MikaelMayer, line 64:
```suggestion
predicate Valid(c: Constants, s: State) {
  && zone_map_well_formed(c.n_blocks, c.zone_map)
  && |s.zones| == |c.zone_map|
  && (forall z :: 0 <= z < |s.zones| ==>
      |s.zones[z]| == (c.zone_map[z].1 - c.zone_map[z].0) as int)
}
```

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+
+
+predicate Write(c: Constants, s: State, s': State, lba: uint64, val: Block)
+    requires Valid(c, s)
+    requires 0 <= lba < c.n_blocks
+{
```
MikaelMayer, line 70:
```suggestion
predicate Write(c: Constants, s: State, s': State, lba: uint64, val: Block)
  requires Valid(c, s)
  requires 0 <= lba < c.n_blocks
{
```

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+    //var p :| resolve_lba(c, lba, p);
```
MikaelMayer, line 71:
```suggestion
```

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+    // The zone containing the block we want to destructively write
+    // is now the new buffer zone
+
+    // Our former buffer zone now contains the zone with the
+    // destructive update
+    && s'.zones[s.buffer_zone] == dst_zone
+
+    // No other zones are touched
+    && (forall i :: 0 <= i < |s'.zones| ==>
+        (i != pzone && i != s.buffer_zone ==> s.zones[i] == s'.zones[i]))
```
MikaelMayer, line 89:
```suggestion
  var (lzone, off) := (0,0);
  var pzone := s.buffer_map[lzone];

  var src_zone := s.zones[pzone];
  var dst_zone := src_zone[ off := val ];

  && |s.zones| == |s'.zones|

  // The zone containing the block we want to destructively write
  // is now the new buffer zone

  // Our former buffer zone now contains the zone with the
  // destructive update
  && s'.zones[s.buffer_zone] == dst_zone

  // No other zones are touched
  && (forall i :: 0 <= i < |s'.zones| ==>
      (i != pzone && i != s.buffer_zone ==> s.zones[i] == s'.zones[i]))
```

Note that, for cosmetic reasons, you might also want to choose to prefix every "var" with "&&", like this:

```
&& var x == 2;
&& var y == 3;
&& x < y
```
but that's obviously completely optional and a matter of taste. Feel free to ignore this comment.

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
```diff
@@ -0,0 +1,96 @@
...
+}
+
+lemma WritePreservesValid(c: Constants, s: State, s': State, lba: uint64, val: Block)
+    requires Valid(c, s)
+    requires Write(c, s, s', lba, val);
+    ensures Valid(c, s')
+{}
```
MikaelMayer, line 96:
```suggestion
lemma WritePreservesValid(c: Constants, s: State, s': State, lba: uint64, val: Block)
  requires Valid(c, s)
  requires Write(c, s, s', lba, val);
  ensures Valid(c, s')
{}

```

Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
```diff
@@ -1087,5 +1086,46 @@ class C {
...
+      await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+
+      /* First, ensure we can correctly extract counterexamples without crashing... */
+      var nonIntegralIndexedSeqs = (await RequestCounterExamples(documentItem.Uri))
+        .SelectMany(cet => cet.Variables.ToList())
+        /* Then, extract the number of non-integral indexed sequences from the repro case... */
+        .Count(IsNegativeIndexedSeq);
+
+      /* If we do not see at least one, then the backend changed the indices being returned to the Language Server. */
+      Assert.IsTrue(nonIntegralIndexedSeqs > 0);
```
MikaelMayer, line 1110:
```suggestion
      Assert.IsTrue(nonIntegralIndexedSeqs > 0, "If we do not see at least one, then the backend changed the indices being returned to the Language Server.");
```

MikaelMayer (changes requested):
Last batch of requests I believe. This PR is great ! Thanks for putting in all the work.

Source/DafnyLanguageServer.Test/Various/TestFiles/3048.dfy
dijkstracula: Interesting, I didn't realise that declarations could be part of conjunctions, but that's useful to know (often I've wanted to introduce, morally, a let-binding in the middle of a larger expression).  Thanks!

MikaelMayer: Well, technically they still aren't, the code above is just parsed as
```
&& (var x == 2;
&& (var y == 3;
(&& x < y)))
```
which simply means Dafny support single-operand conjunctions :-)

dijkstracula:
> Last batch of requests I believe. This PR is great ! Thanks for putting in all the work.

Done - I just went ahead and squashed your formatting changes, so I think we should be good to go!

MikaelMayer (approved):
Great work !

Issue #3095, opened 2022-11-22
[Feature request]: `{:only}` on assertions


MikaelMayer:
Dafny developers writing long methods often face long development cycles while developing the proofs of individual assertions. Even if they add `{:vcs_split_on_every_assert}`, the verifier uses wastes cores trying to verify things that are not relevant yet for the user.

As a workaround, It became almost idiomatic for Dafny developers to add `assume false;` to branches not in focus to speed-up verification.  This is because the verifier can use `false` to prove anything, so branches starting with `assume false;` are verified almost instantly, giving time for Dafny to focus on branches not having it. This mechanism is used both at the beginning of a proof (so that Dafny does not complain about multiple times the same unproven result) and during the proof (to "comment out" portions of the code that were already proven and that are not supposed to be changed)

However, consider the following example:

```
method Test(inputs...) {
  BigPreComputations(); // [1]
  match X {
  case 1 => assume false; // Proof already done, "commented out" by assume false;
  case 2 => // code in focus
  case 3 => assume false; // Proof to do...
  ...
  case n=> assume false; ...
 
  }
}
```

This emergent mechanism has major shortcomings:
- It does not scale with the number of branches. Focusing on one branches among n requires adding n-1 `assume false`
- All the assertions before the case split (in `[1]`) can't be marked with "assume false" easily; explicit assertions could using `by { assume false; }`, but implicit assertions would need to be made explicit, which, although easy in the case of preconditions, is not trivial for other kinds of assertions such as allocations or existentials.
- Forgetting an `assume false` somewhere could lead to a disastrous experience down the road when the compiler steps in. This could be solved by #3076 though.
- Sometimes there are so many facts available that I have personally seen a case where "assume false" did not speed up the verifier that much because it did not use that fact.

What developers really want is a way to tell the verifier to temporarily _focus_ on a part of the proof.
The good news, we already have experience of focusing. It works already for methods and functions.
* One can use `/proc:<functionName>` on the command line
* When verifying one function or method, all the others are accessed only by their axioms, and not their implementations.

Therefore, there is a straightforward mechanism to focus on a proof: we can temporarily transform assertions in assumptions, like we did when verifying a function, when we transform all the assertions of other functions into assumptions (axioms). This has the following advantages:
- Easy scaling: No more need to `assume false` in other branches, only one command or annotation per thing in focus.
[...]

PR #3116, opened 2022-11-25
Feat: conditional verification, detection of trivially false assumptions

### LINKED ISSUE #3089: [Feature request]: Alert users to obviously contradictory preconditions
### What is the feature you would like to see in a future version of Dafny?

Pulling out the discussion in the comments section of #3076 into its own issue:

I was recently bitten by a super-silly precondition of mine, essentially of the form `requires x < x` for some `x: int`, and only much later did I realise I had, somewhere, tied myself in a knot[1]. By this point, the contradiction was in a totally different module that I had paged out of my mind, so it took awhile to find.  It would be nice if Dafny could expose up impossible preconditions so the user knows that they're not inadvertently introducing an `exfalso` style hole in their proofs.  It'd be fine if this was only a warning, indicating the presence of such a precondition or that a proof transitively involves such a precondition (via a `require`d predicate or lemma, for instance.)

@MikaelMayer replied thus:

> This is a very interesting idea that indeed deserves to be studied on its own. It has its own challenges:  Trying to prove "false" when it's not possible can be extremely expensive for a verifier. How to ensure we can do both? Maybe caching down the road?

This seems like a problem, indeed - most of the type proving false would not be possible, presumably, so in the typical case we'd be giving the solver an unnecessary workout.

To this end, Mikael proposed a subset of the feature, where straightforward contradictory syntactic patterns could be caught cheaply:

> ```
> x < x    where x is a variable
> x > x
> x != x
> x == x.field  where x is a datatype.
> exists x :: E where can be easily inferred to be always false
> !E where E can be easily inferred to be always true
> E && B   if E or B are always false.
> E || B   if E and B are always false.
> ```
 
I suspect something like this would have caught my particular issue (technically my "reflexive inequality" involved an array access, akin to `a[i] < a[i]`, but presumably it's no more complex to pull in an array axiom or two?)

I might suggest two points of discussion:

1) Another potential pitfall that would be nice to have a story around would be for "obvious contradictions" _between_ preconditions: for instance, if I `require a[i] == b[i]`, I better not also `require a != b`, or some such thing.  In terms of implementation this could well be orthogonal to a localised syntax-driven check.  By the same token, I could also imagine at some level all preconditions get ANDed together into some (in this case contradictory) formula.  This could still abut complexity issues, of course.

2) There could be a concern that users might expect the precondition contradiction checker to be smarter than it is, and rely on it for more complicated cases than it knows about.  For instance, suppose that array indexing equality was _not_ part of the hardcoded syntactic checks (as it isn't listed in the above set of expressions): I might conclude that the issue in my project was elsewhere from the `requires a[i] < a[i]` case, since I might assume that Dafny catches precondition contradictions in general.  I don't know how often users trip over things like this.

Thanks!
Nathan

[1]: power dafny user lesson 1: if your code is consistently passing the verifier and you're starting to think, "boy, am I ever good at writing correct programs", be wary ;-)

MikaelMayer:
Fixes #3089

Dafny's proofs are meant to be sound, but what happens if one assumes falsehood? Verification passes without a single warning.
While this is the desired behavior when developing proofs and "commenting out" certain branches for the verifier, not being shown even just a warning about them is concerning

Currently, the only safety net Dafny provides is that the compiler will not compile Dafny programs containing `assume XXXX` not marked as `{:axiom}`. However, the following reasons makes it hard to swallow that Dafny does not react at all when encountering `assume XXXX` for every assumption.

- After verifying a branch, the VSCode extension displays a green gutter, even if there are more branches to verify. So yes, it's sound, but it's not useful for a user.
- It can be even misleading: For proofs that take days, users reported they would forget about an assume false somewhere, and be bitten by it much later.
- It can be disastrous: Some users reported that they wrote `ensures` clause on externs that trivially assumes false, but they did not see it and it caused all proofs calling this external code to pass. Very dangerous. There is no safety net even for a compiler.

While there might be a place to insert warnings (and we can discuss that below, see future work), this PR's first aim is to introduce a more accurate gutter icon feedback when encountering assumptions.
The design of the gutter icons is not yet ready, but I approached several users to validate the user experience already. Here is what will happen, which is what this PR implements:

# User experience.

Here are some principles I validated
* If a method/function/declaration has some errors, there is no need to report any assumption to the user. That would be extra information that is not yet useful. They have a more important problem to solve, which is to make verification pass given their assumption. Thus, icons are unchanged when there is at least one error in the context.
* Only if the method/function/declaration has no verification error, then it is time to show the user the remaining assumptions.
* Since assumptions can be out of view, it's imperative that the normally "green and vertical" line for the entire method/function/declaration is changed to something meaning "There are still assumptions remaining"
* In this context of "Context verified with assumptions", assumptions should be clearly visible.
* Are marked as "assumptions" the following
  * Any `assume` that is not _trivially true_ and that is not marked as `{:axiom}`. I have seen users using  `assume {:split_here} true;` for example, they should not be marked as assumptions
  * Any `requires` that is _trivially false_.
  * Any `ensures` of a method or a function without body (`{:extern}` or not) whose condition is trivially false
* _trivially_ means, in the context of this PR, the static function `Translator.IsExpressionAlways`:
  * `false` is trivially false
  * `true` is trivially true
  * `!X` is trivially true if `X` is trivially false, and trivially false if `X` is trivially true
  * `X && Y` is trivially true if both `X` and `Y` are trivially true, and trivially false if `X` or `Y` are trivially false.
  * `X || Y` is trivially true if either `X` or `Y` is trivially true, and trivially false if both `X` and`Y` are trivially false.
  * `X ==> Y` is trivially true if either `!X` or `Y` is trivially true, and trivially false if both `!X` and`Y` are trivially false.
  * `X <==> Y` is trivially true if both `X` and `Y` are trivially true or both false, and trivially false if both `!X` and`Y` have opposite trivial boolean values.
  * `X == Y` is trivially true if `X` and `Y` are booleans and `X <==> Y` is trivially true, or if rendering the expression X and Y results in the same string. It's trivially false if `Y = X.field` for some field. (or reverse)
  * `X < Y` is trivially false if the string representation of X and the string representation of Y are the same
     It also contains a trivial test when X and Y are literals
  * `X != Y` is trivially false if the string representation of X and the string representation of Y are the same
     It also contains a trivial test when X and Y are literals  It's trivially true if `Y = X.field` for some field (or reverse)
  * `X > Y` is trivially false if the string representation of X and the string representation of Y are the same
     It also contains a trivial test when X and Y are literals
  * `forall ... | Z :: X` is trivially true if `Z ==> X` is trivially true, but never trivially false (we would first need to show that the collection is non-empty, we could do this in some case, not sure it's useful)
  * `exists... | Z :: X` is trivially false if `Z && X` is trivially false, but never trivially true(we would first need to show that the collection is non-empty, we could do this in some case, not sure it's useful)

I've left out any closure `requires` as it's not harmful to define a closure without calling it in any context, so the requires false would be easily detected.

# Benefits that this PR brings:

* Once a branch is verified, a user will not be misled into thinking that nothing left needs to be verified, if they left some `assume false`.
* `requires` that are trivially `false` are flagged
* Assumptions become visible when the method is verified
* No warnings on `assume false` in case it's desired (and it often is)
* Trivially false `ensures` of bodiless methods are flagged and become visible.

# Caveats of this PR.

I would like to echo @dijkstracula 's comment

> There could be a concern that users might expect the precondition contradiction checker to be smarter than it is, and rely on it for more complicated cases than it knows about. For instance, suppose that array indexing equality was not part of the hardcoded syntactic checks (as it isn't listed in the above set of expressions): I might conclude that the issue in my project was elsewhere from the requires a[i] < a[i] case, since I might assume that Dafny catches precondition contradictions in general. I don't know how often users trip over things like this.

# How this PR was tested

I created tests in the language server that illustrate all of the conditions above exhaustively.

I also created a draft visual in VSCode to view the behavior, and ask users questions about their usage of assumptions, and about what they thought of the user experience. Most were supportive. Those who did not know if they should support this or not were ones who were not used to using assumptions in the code, so they wouldn't be harmed anyway.

Note that this PR is a draft until there is proper icon support for VSCode. Otherwise, such contexts would appear as if no verification was made (which is not bad, but not a great user experience)

# Future visual considerations when designing gutter icons for this PR.

* This visual, let's call it "Context verified with assumptions", identified as ` |?` in the tests (compared to ` | ` for verified context) should be closer to "verified" than to "error with context". Indeed, since `assume false` is part of the development cycle, the cycle is usually: "Error with context" (yellow lines) => "Context verified with assumptions" => Navigate to other assumptions => Remove or weaken the assumptions => "Error with context" (yellow lines) => Proof => etc. Hence, every time users finish the branch of a proof, they should be at least "rewarded" with the design "Context verified with assumptions".
  Hence, this visual should be closer to green, possibly green-yellowish.
  Moreover, it's not helpful to dim the green or split the green in two, because it reduces visibility like when verification is "stale". We need another visual
* The visual of remaining assumptions should be obvious, prominent and for example yellow (to indicate a warning)

# Future work

* If we agree, we could also turn some of this feedback as real warnings. This can be a separate PR though, I'm happy to do it.
* In the future, one might also use these gutter icons for other purposes, such as [identifying localized proofs](https://github.com/dafny-lang/dafny/issues/3095) 

### DIFF
--- a/Source/DafnyCore/Verifier/Translator.cs
+++ b/Source/DafnyCore/Verifier/Translator.cs
@@ -3864,6 +3864,10 @@ namespace Microsoft.Dafny {
       if (q is LiteralExpr { Value: bool value } && value == truth) {
         return true;
       }
+
+      if (q is UnaryOpExpr unaryOpExpr && unaryOpExpr.Op == UnaryOpExpr.Opcode.Not) {
+        return IsExpressionAlways(unaryOpExpr.E, !truth);
+      }
       if (q is BinaryExpr binaryExpr) {
         var e0 = binaryExpr.E0;
         var e1 = binaryExpr.E1;
--- a/Source/DafnyCore/Verifier/Translator.cs
+++ b/Source/DafnyCore/Verifier/Translator.cs
@@ -3851,6 +3851,106 @@ namespace Microsoft.Dafny {
       MakeAssert(tok, q, desc, kv);
     }
 
+    public static bool IsExpressionAlways(Expression q, bool truth) {
+      q = q.WasResolved() ? q.Resolved : q;
+      if (q is ForallExpr forallExpr) {
+        return truth && IsExpressionAlways(Expression.CreateImplies(
+          forallExpr.Range ?? Expression.CreateBoolLiteral(Token.NoToken, true), forallExpr.Term, true), true);
+      }
+      if (q is ExistsExpr existsExpr) {
+        return !truth && IsExpressionAlways(Expression.CreateAnd(
+          existsExpr.Range ?? Expression.CreateBoolLiteral(Token.NoToken, true), existsExpr.Term, true), false);
+      }
+      if (q is LiteralExpr { Value: bool value } && value == truth) {
+        return true;
+      }
+      if (q is BinaryExpr binaryExpr) {
+        var e0 = binaryExpr.E0;
+        var e1 = binaryExpr.E1;
+        switch (binaryExpr.Op) {
+          case BinaryExpr.Opcode.And:
+            return truth
+              ? IsExpressionAlways(e0, true) && IsExpressionAlways(e1, true)
+              : IsExpressionAlways(e0, false) || IsExpressionAlways(e1, false);
+          case BinaryExpr.Opcode.Or:
+            return truth
+              ? IsExpressionAlways(e0, true) || IsExpressionAlways(e1, true)
+              : IsExpressionAlways(e0, false) && IsExpressionAlways(e1, false);
+          case BinaryExpr.Opcode.Imp:
+            return truth
+              ? IsExpressionAlways(e0, false) || IsExpressionAlways(e1, true)
+              : IsExpressionAlways(e0, true) && IsExpressionAlways(e1, false);
+          case BinaryExpr.Opcode.Iff:
+            return truth
+              ? (IsExpressionAlways(e0, false) && IsExpressionAlways(e1, false) ||
+                 IsExpressionAlways(e0, true) && IsExpressionAlways(e1, true))
+              : (IsExpressionAlways(e0, true) && IsExpressionAlways(e1, false) ||
+                 IsExpressionAlways(e0, false) && IsExpressionAlways(e1, true));
+          case BinaryExpr.Opcode.Eq:
+            if (truth) { // Obvious case when A == B is true, when the two representations are the same.
+              return Printer.ExprToString(e0) == Printer.ExprToString(e1);
+            } else {
+              // Cases in which we absolutely know that an equality can't hold
+              // - One expression is a datatype and the other is destructing this datatype
+              // - Two literals that are different
+              if (e0 is LiteralExpr { Value: BigInteger b1 } && e1 is LiteralExpr { Value: BigInteger b2 }) {
+                return b1.CompareTo(b2) != 0;
+              }
+              if (e0 is LiteralExpr { Value: bool bb1 } && e1 is LiteralExpr { Value: bool bb2 }) {
+                return bb1 != bb2;
+              }
+              if (e0.Type.NormalizeExpand().AsIndDatatype is IndDatatypeDecl
+                  && e1.Type.NormalizeExpand().AsIndDatatype is IndDatatypeDecl
+                  && Printer.ExprToString(e0) is var e0str
+                  && Printer.ExprToString(e1) is var e1str
+                  && (e1str.StartsWith(e0str + ".") || (e0str.StartsWith(e1str + ".")))) {
+                return true;
+              }
+              return false;
+            }
+          case BinaryExpr.Opcode.Lt: {
+              if (!truth && Printer.ExprToString(e0) == Printer.ExprToString(e1)) { // Obvious case when A < B is wrong: B == A
+                return true;
+              }
+              // Obvious case when A < B is always true, when A and B are bigintegers
+              if (e0 is LiteralExpr { Value: BigInteger b1 } && e1 is LiteralExpr { Value: BigInteger b2 }) {
+                return truth == (b1.CompareTo(b2) < 0);
+              }
+              return false;
+            }
+          case BinaryExpr.Opcode.Gt: {
+              // Obvious case when A > B is wrong: B == A
+              if (!truth && Printer.ExprToString(e0) == Printer.ExprToString(e1)) {
+                return true;
+              }
+              // Obvious case when A > B is always true, when A and B are bigintegers
+              if (e0 is LiteralExpr { Value: BigInteger b1 } lit1 && e1 is LiteralExpr { Value: BigInteger b2 }) {
+                return truth == (b1.CompareTo(b2) > 0);
+              }
+              return false;
+            }
+          case BinaryExpr.Opcode.Neq: {
+              // Obvious case when A != B is wrong: B == A
+              if (!truth && Printer.ExprToString(e0) == Printer.ExprToString(e1)) {
+                return true;
+              }
+              // Obvious case when A > B is always true, when A and B are bigintegers
+              if (e0 is LiteralExpr { Value: var value1 } && e1 is LiteralExpr { Value: var value2 }) {
+                if (value1 is BigInteger b1 && value2 is BigInteger b2) {
+                  return truth == (b1.CompareTo(b2) != 0);
+                } else if (value1 is bool bb1 && value2 is bool bb2) {
+                  return truth == (bb1 != bb2);
+                }
+              }
+
+              return false;
+            }
+        }
+      }
+
+      return false;
+    }
+
     /// <summary>
     /// Returns true if it can statically determine that the expression q always evaluates to truth
     /// </summary>
--- a/Source/DafnyCore/AST/Expressions.cs
+++ b/Source/DafnyCore/AST/Expressions.cs
@@ -3167,6 +3167,7 @@ public class StmtExpr : Expression {
       Contract.Assert(false); throw new cce.UnreachableException();  // unexpected statement
     }
   }
+  public override IEnumerable<INode> Children => new List<INode> { S, E };
 }
 
 public class ITEExpr : Expression {
--- a/Source/DafnyLanguageServer.Test/GutterStatus/LinearRenderingTest.cs
+++ b/Source/DafnyLanguageServer.Test/GutterStatus/LinearRenderingTest.cs
@@ -51,6 +51,22 @@ public class LinearRenderingTest {
             : LineVerificationStatus.VerifiedVerifying,
         _ => throw new ArgumentOutOfRangeException()
       },
+      // If a node is marked with VerifiedWithAssumption
+      GutterVerificationStatus.VerifiedWithAssumption => currentStatus switch {
+        CurrentStatus.Current => contextHasErrors && isSingleLine
+                                 || !contextHasErrors && contextIsPending && !isSingleLine
+            ? LineVerificationStatus.Nothing
+            : LineVerificationStatus.VerifiedWithAssumption,
+        CurrentStatus.Obsolete => contextHasErrors && isSingleLine
+         || !contextHasErrors && contextIsPending && !isSingleLine
+          ? LineVerificationStatus.Scheduled
+            : LineVerificationStatus.VerifiedWithAssumptionObsolete,
+        CurrentStatus.Verifying => contextHasErrors && isSingleLine ||
+            !contextHasErrors && (contextIsPending && !isSingleLine)
+            ? LineVerificationStatus.Verifying
+            : LineVerificationStatus.VerifiedWithAssumptionVerifying,
+        _ => throw new ArgumentOutOfRangeException()
+      },
       // We don't display inconclusive on the gutter (user should focus on errors),
       // We display an error range instead
       GutterVerificationStatus.Inconclusive => currentStatus switch {
@@ -83,6 +99,8 @@ public class LinearRenderingTest {
               Assert.AreEqual(
                 RenderLineVerificationStatusOriginal(isSingleLine, contextHasError, contextIsPending, currentStatus, verificationStatus),
                 VerificationTree.RenderLineVerificationStatus(isSingleLine, contextHasError, contextIsPending, currentStatus, verificationStatus)
+                , "verificationStatus={0}, currentStatus={1} isSingleLine={2}, contextHasError={3}, contextIsPending={4}",
+                  verificationStatus, currentStatus, isSingleLine, contextHasError, contextIsPending
                 );
               contextIsPending = !contextIsPending;
             } while (!contextIsPending);
--- a/Source/DafnyLanguageServer.Test/GutterStatus/LinearVerificationGutterStatusTester.cs
+++ b/Source/DafnyLanguageServer.Test/GutterStatus/LinearVerificationGutterStatusTester.cs
@@ -46,7 +46,13 @@ public abstract class LinearVerificationGutterStatusTester : ClientBasedLanguage
     { LineVerificationStatus.AssertionVerifiedInErrorContextObsolete, "[o]" },
     { LineVerificationStatus.AssertionVerifiedInErrorContextVerifying, "[Q]" },
     { LineVerificationStatus.AssertionVerifiedInErrorContext, "[O]" },
-    { LineVerificationStatus.ResolutionError, @"/!\" }
+    { LineVerificationStatus.ResolutionError, @"/!\" },
+    { LineVerificationStatus.VerifiedWithAssumption, " |?" },
+    { LineVerificationStatus.VerifiedWithAssumptionObsolete, " I?" },
+    { LineVerificationStatus.VerifiedWithAssumptionVerifying, " $?" },
+    { LineVerificationStatus.AssumptionInVerifiedContext, " =?" },
+    { LineVerificationStatus.AssumptionInVerifiedContextObsolete, " -?" },
+    { LineVerificationStatus.AssumptionInVerifiedContextVerifying, " ~?" },
   };
 
   private static bool IsNotIndicatingProgress(LineVerificationStatus status) {
@@ -59,7 +65,9 @@ public abstract class LinearVerificationGutterStatusTester : ClientBasedLanguage
            status != LineVerificationStatus.ErrorContextObsolete &&
            status != LineVerificationStatus.ErrorContextVerifying &&
            status != LineVerificationStatus.AssertionVerifiedInErrorContextObsolete &&
-           status != LineVerificationStatus.AssertionVerifiedInErrorContextVerifying;
+           status != LineVerificationStatus.AssertionVerifiedInErrorContextVerifying &&
+           status != LineVerificationStatus.VerifiedWithAssumptionObsolete &&
+           status != LineVerificationStatus.VerifiedWithAssumptionVerifying;
   }
   public static string RenderTrace(List<LineVerificationStatus[]> statusesTrace, string code) {
     var codeLines = new Regex("\r?\n").Split(code);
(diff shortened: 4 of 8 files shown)


keyboardDrummer:
> If we agree, we could also turn some of this feedback as real warnings. This can be a separate PR though, I'm happy to do it.

Since warnings are part of the standard LSP API, why not implement those first? I think it would also simplify the scope of this PR since then there's no API design consideration.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -3851,6 +3851,122 @@ public enum StmtType { NONE, ASSERT, ASSUME };
       MakeAssert(tok, q, desc, kv);
     }
 
+    public static bool IsExpressionAlways(Expression q, bool truth) {
```
keyboardDrummer, line 3854: Could you move this to VerificationDiagnosticsParams? This file is already so large and this method isn't used here.

MikaelMayer: > Could you move this another file? This file is already so large and this method isn't used here. I think I would prefer it close to the other gutter code, or if you think this is useful for other scenarios maybe in its own file in DafnyCore.

I think it will be interesting to other functions in DafnyCore down the road, so I moved it to its own file `QuickVerifier`.

> Should you ignore assume statements that have {:axiom} on them?

Yes, the same way the compiler ignores them.

> Is there a reason to ignore trivially true assume statements? Is there a use-case for them? It seems simpler to 'warn' about all these assume statements as well.

I've seen the use of `assume {:split_here} true;` so that it triggers Boogie splits of verification without introducing new assertions to verify. The assumptions don't have any reason to be warned about, even if it is simpler. Of course, this is was not for our own Dafny compiler.

> I think @atomb is planning to add an analysis that detects which requires/ensures are always false, using the verifier. Would that be a more powerful version of this?

Yes that will be a more powerful version of this. This is why I renamed the file `QuickVerifier`, because it can say if something is `unsat` or `unknown` for a very small subset of all propositions, but which were relevant for many of our customers. This heuristic feature executes linearly with the size of the conditions and only detect patterns, so it won't slow down verification like @atomb might do - but that will be needed anyway probably at least for CI builds.

keyboardDrummer: > I've seen the use of assume {:split_here} true; so that it triggers Boogie splits of verification without introducing new assertions to verify. The assumptions don't have any reason to be warned about, even if it is simpler. Of course, this is was not for our own Dafny compiler.

I'd still be in favor of not adding that edge-case check. I think pushing users to consistently use `assert {:split_here} true;` instead of `assume {:split_here} true` will make the language easier to understand.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -3851,6 +3851,122 @@ public enum StmtType { NONE, ASSERT, ASSUME };
...
+                } else if (value1 is bool bb1 && value2 is bool bb2) {
+                  return truth == (bb1 != bb2);
+                }
+              }
+              if (truth
+                  && e0.Type.NormalizeExpand().AsIndDatatype is IndDatatypeDecl
+                  && e1.Type.NormalizeExpand().AsIndDatatype is IndDatatypeDecl
+                  && Printer.ExprToString(e0) is var e0str
+                  && Printer.ExprToString(e1) is var e1str
+                  && (e1str.StartsWith(e0str + ".") || (e0str.StartsWith(e1str + ".")))) {
```
cpitclaudel, line 3958: Why do you have to invoke the printer as part of this static analysis?

MikaelMayer: It makes it easy to reason about expressions equality. It is slower than extracting the original string, but it's more robust. It's slower than if we had an up-to-date and well maintained Equals() function on expressions, but this was never implemented. This could be done in the future, and I keep the scope of this PR as small as possible,
Printers are linear and cost a linear amount of memory so I don't think this is much of a concern.

Source/DafnyCore/Verifier/Translator.cs
MikaelMayer: > I'd still be in favor of not adding that edge-case check. I think pushing users to consistently use assert {:split_here} true; instead of assume {:split_here} true will make the language easier to understand.

Ok another case then. Let's say a user writes this:

`assume a[i] == a[i]` instead of `assume a[i] == a[j];`

If we signal all assumptions, then the first case will be silently ignored as the user will expect the assumption to be underlined as it would be for the second case. It might not help them to spot an issue in the assumption, in that the assumption does not help at all.

So, checking if the assumption is trivially true will give the feedback that, given a user would expect assumptions to be marked, if it's not marked, they will know this assumption is useless.

Source/DafnyCore/Verifier/Translator.cs
cpitclaudel: I don't know :/ I'd rather just remove this feature than depending on the printer for static analysis.

Source/DafnyCore/Verifier/Translator.cs
keyboardDrummer: I think Dafny should have the following warnings:
- A warning for every `assume` without {:axiom}. We can incorporate this one in the gutter icons.
- A warning for expressions with free variables that always have the same value, such as `a[i] == a[i]`. For reasons of performance we might only do this check for expressions passed to `assert` and `assume` statements. I suggest we don't incorporate this one in the gutter icons, but just leave it as a warning.

Given the above:
`assume a[i] == a[i]` <-- two warnings
`assume {:axiom} a[i] == a[i]` <-- one warning
`assume a[i] == a[j]` <-- one warning
`assume {:axiom} a[i] == a[j]` <-- zero warnings

It's not yet clear to me what information you want to incorporate into gutter icons. Do you want any warnings that relate to verification to cause the verification context to show up as 'not green' in the gutter?

I can also imagine two other options:
- Any warning, whether it relates to verification or not, causes the gutter icons of the context to show as non-green
- The gutter icons are green as long as no verification errors occur in the context, so all warnings, including those coming from `assume` statements, are ignored.

What I like above the above two options is that I find them simple to understand.

Source/DafnyCore/Verifier/Translator.cs
cpitclaudel: I meant removing this specific part, not the whole feature.  It's an approximation anyway, so making it a bit less precise isn't an issue, is it?

cpitclaudel: Besides, isn't the check wrong?

```
datatype T = A | B {
  const ths := this
}

method M(t: T) {
  assert t.ths == t;
  assume t.ths != t; // Should be flagged
}
```

I really don't think relying on pretty-printing is the right way to go.

Source/DafnyCore/Verifier/Translator.cs
atomb: Yeah, even though I am planning a deeper analysis for contradictory assumptions I think having a really fast check for obvious cases is worthwhile. Detecting those as early as possible is valuable.

Source/DafnyCore/Verifier/Translator.cs
MikaelMayer: > I meant removing this specific part, not the whole feature. It's an approximation anyway, so making it a bit less precise isn't an issue, is it?

Making it a bit less precise just wouldn't have caught multiple development issues I have witness in the past year, such as `i < i`, `SomeFunction(a, b) != SomeFunction(a, b)`, or recently @dijkstracula 's  `a[i] != a[i]`.
I would feel very bad after this PR if someone had such an issue of this type again, and I told him, "I'm sorry, your code could have been flagged if we just hadn't removed this small part from the PR"

> I really don't think relying on pretty-printing is the right way to go.

I get your point that we don't want to ship half-baked features. Every time we issue a warning, users will start relying on these warnings in their everyday's development process, and might get mad if some warnings don't show up where they think they should.

Now, I never claimed relying on pretty printing is the "right" way to go. I claim I claim it's a cheap, brutal and efficient way to get the job to start.
Plus, since it's not clever, users are quickly going to realize it - and won't be surprised in your example that the warning isn't issued.
Take as example that no one entirely on the null-pointer analysis of C#, or its warning that a condition is always true/false because it's never complete, there are always ways to create a condition that is trivially false and that is not flagged, for example.
So, especially because we are generating warnings and not errors, we don't have any reason to believe that users will have high expectations of these warnings

Defining a scope for quick static analysis can be done later - not by me, I'm working on proof locality, not static analysis - and will be another topic. I'm not sure who will want to work on it, but if that means we skip these already useful warnings for the next 12 months, I would just strongly oppose against removing this feature in this PR.

cpitclaudel: But just to be clear: that test is wrong, right?

Source/DafnyCore/Verifier/Translator.cs
MikaelMayer: Thanks for your feedback, always appreciated. Let's see if we can converge to a practical roadmap.

> I think Dafny should have the following warnings:

Please, consider that the scope of this PR is not about warnings, it's about indicating contextual proof status. You interpreted it as Dafny should add warnings, but I believe this is a different (and useful) point that I want to keep out of scope of this PR. If you want I can work on warnings after this PR. I agree that, for now, we should consider how this PR can interact with warnings reported in the gutter in the future, so that we can choose the "icons" wisely. See the discussion below.

> * A warning for every assume without {:axiom}.

When it will come to warnings, I think you have the right approach. We could just issue two warnings. For this PR, I have no plans of making "double warning gutter icons".

> * A warning for expressions with free variables that always have the same value, such as a[i] == a[i]. For reasons of performance we might only do this check for expressions passed to assert and assume statements.

What you probably had in mind was not "expressions with free variables that always have the same value", but "expressions that are always trivially true", right? Otherwise, you are assuming equality assumptions, which this PR does not.

> 1. Any warning, whether it relates to verification or not, causes the gutter icons of the context to show as non-green

I like the idea of displaying warnings in the gutter in any case in principle - the same way we already display diagnostic errors with triangles - but in this case I would like the gutter to switch from green to the conditional verification symbol **if and only if there is at least one warning that concerns potential global verification issues when the code is verified**.
For example, the warnings that affect verification are nontrivial assumptions not in requires, a forall statement without body, etc.
Warnings that don't affect global verification outcomes, to my opinion, should not trigger the contextual "verified" icons to be changed, such as comparing null to a non-null type, importing an empty module, requires clause not compilable, unnecessary (==) if type parameter is ghost, the conclusion of a forall statement will not be known...

> 2. The gutter icons are green as long as no verification errors occur in the context, so all warnings, including those coming from assume statements, are ignored.

As described in this PR's description, that would defeat the whole purpose of indicating to the user that there are still some assume statements in the scope of the proof. So this is a big no for this one, even if it's simple.

Source/DafnyCore/Verifier/Translator.cs
MikaelMayer: Oh I completely oversaw the test itself. My head too much trying to describe the benefits of implementing this PR cheaply :-)

Yes, you're right. I did not think of inner fields destructions. I made the test robust by ensuring that the indirection is a datatype destructor. I also added a test to capture your behavior.

Thank you very much for this contribution !

cpitclaudel (commented):
I won't veto anything, but I'm uncomfortable with relying on the printer for this.  The printer is poorly tested, has known bugs, and generally isn't something that (I think) we should be relying on for equality checks between nodes.

cpitclaudel (changes requested):
Actually, I do want to veto the use of the printer, because we *do* have a generic way to check expression equality.  Could you use this?

https://github.com/dafny-lang/dafny/blob/5946299cc578a13d4d6aa7e325d47d8d2200e5bc/Source/DafnyCore/Triggers/TriggerExtensions.cs#L102

It would probably be best to move it into the `AST/` directory somewhere, now that it's needed by more than just triggers.

MikaelMayer:
> Actually, I do want to veto the use of the printer, because we _do_ have a generic way to check expression equality. Could you use this?
> 
> https://github.com/dafny-lang/dafny/blob/5946299cc578a13d4d6aa7e325d47d8d2200e5bc/Source/DafnyCore/Triggers/TriggerExtensions.cs#L102
> 
> It would probably be best to move it into the `AST/` directory somewhere, now that it's needed by more than just triggers.

That was a very clever remark. Thanks for pointing this code out. After using it, the functionality works the same, and I feel better too. It even does the checks of bound variables.
I did move the code using your great script git-cp, but that code was also referring to triggers. I guess it's ok.

Source/DafnyCore/Verifier/Translator.cs
keyboardDrummer: > Please, consider that the scope of this PR is not about warnings, it's about indicating contextual proof status. You interpreted it as Dafny should add warnings, but I believe this is a different (and useful) point that I want to keep out of scope of this PR. If you want I can work on warnings after this PR. I agree that, for now, we should consider how this PR can interact with warnings reported in the gutter in the future, so that we can choose the "icons" wisely. See the discussion below.

Actually I'm just talking about warnings to better understand this PR, since they're so related. That's also why I'm arguing to let any `assume` without axiom trigger the warning gutter context, since I'm planning to let any assume without axiom trigger a warning and the two should be consistent. (Ideally we would achieve the consistency by computing (part of) the gutter icons on the client based on the warnings, but now let's do it using the APIs.

> Warnings that don't affect global verification outcomes, to my opinion, should not trigger the contextual "verified" icons to be changed

Maybe that's the right thing to do but why? Why not all warnings? Why would the gutter only try to help you with verification problems instead of anything that requires more actions in that piece of code?

> What you probably had in mind was not "expressions with free variables that always have the same value", but "expressions that are always trivially true", right? Otherwise, you are assuming equality assumptions, which this PR does not.

I was imagining which warnings we'd ideally like to have in the future.

Source/DafnyCore/Verifier/QuickSyntaxBasedVerifier.cs
```diff
@@ -0,0 +1,138 @@
...
+using Microsoft.Dafny.Triggers;
+
+namespace Microsoft.Dafny.ExprExtensions;
+
+/// <summary>
+/// This class's code contains helpers that help verifying assertions without
+/// the use of Boogie
+/// See Translator.IsExprAlways for a smaller version for Boogie expressions
+/// </summary>
+public static class QuickSyntaxBasedVerifier {
```
keyboardDrummer, line 11: You've described this verifier by saying it's "quick" "trivial" and "obvious", but I'd like to be more precise about what the scope is. And once it's clear what the scope is, is it complete or do you expect it to grow more?

Source/DafnyCore/Verifier/Translator.cs
MikaelMayer: > Maybe that's the right thing to do but why? Why not all warnings? Why would the gutter only try to help you with verification problems instead of anything that requires more actions in that piece of code?

I think the primary reason was to make them not too complex, but I would agree it's not a good reason. Let's investigate some more points and see whether we can arrive at something we both like.

Until today, the gutter icons have been used almost exclusively to display information related to code verification, with the only exception of resolution errors. But we display resolution errors using sharp triangles on top of greyed previous verification. We don't use a specific "context" saying there are resolution errors in scope, it can be seen on the entire file on the scroll bar, and locally on the gutter if needed. And I think that's ok.

So while I also want to possibly display warnings on gutter icons (especially when there are no errors or context errors), I would like to not always change the "context" value. I would like to keep true to the original verification gutter icons principle: If the context is one line green, Dafny did not identify something to worry about the relationship between the spec and the code.

Now, I think it's going to be not easy to distinguish warnings that should affect the verification outcome if acted on, and those which don't, but I think we could make an effort. Can we mark the source of such warnings like "Verifier" and "Resolver" ? The first one would impact verification, the second ones wouldn't.
For example, I strongly believe that warnings about dead code, about no trigger for quantifiers found, comparing null to non-null type, should not trigger the feeling that there is something to fix between the spec and the implementation. That would be annoying. They are just there to help optimize code, not to actually warn of something serious that could impact verification. Such warnings could be displayed as a simple icon (like resolution error icons) on top of any verification context that is not an error, not affecting the context itself like verification warnings would.

One the other side, a warning about requires false, assume anything (nontrivial, I would still argue for that) without `{:axiom}`, or while loops without a body, should for sure trigger the verification warning in context.

Source/DafnyCore/Verifier/QuickSyntaxBasedVerifier.cs
MikaelMayer: What is unclear exactly? Can you please be more precise about what do you mean by "scope"? I precisely defined the notion of trivialness in this PR's description. In heuristics, there is nothing complete. Would you like me to put the PR's description in the comments of the class? Or rename the class?

PR #3117, opened 2022-11-28
Add refined AST cloner and fix bugs in AST traversal code

keyboardDrummer:
Part of https://github.com/dafny-lang/dafny/pull/2734

### Changes
- Enable `Cloner` to also cloned AST nodes including the members set during refinement
- Move AST Node specific cloning code from `Cloner` to the respective AST nodes.


Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -38,6 +40,36 @@ public abstract class INode {
...
+        var current = toVisit.Pop();
+        if (!visited.Add(current)) {
+          continue;
+        }
+
+        if (!beforeChildren(current)) {
+          continue;
+        }
+
+        foreach (var child in current.Children.Reverse()) {
```
cpitclaudel, line 60: Most nodes have iterators (not lists) as children, so `Reverse` is probably not the most efficient.  Does the traversal order matter?  If so maybe you could use a dequeue instead of a stack with `Reverse`?

keyboardDrummer: Removed the Reverse for now. Actually seems this is only used in the follow-up PR, woops!

Source/DafnyCore/Cloner.cs
```diff
@@ -294,19 +349,9 @@ public class Cloner {
...
-      } else if (expr is IdentifierExpr) {
-        var e = (IdentifierExpr)expr;
-        return new IdentifierExpr(Tok(e.tok), e.Name);
-
-      } else if (expr is DatatypeValue) {
-        var e = (DatatypeValue)expr;
-        return new DatatypeValue(Tok(e.tok), e.DatatypeName, e.MemberName, e.Bindings.ArgumentBindings.ConvertAll(CloneActualBinding));
-
+        return new AutoGhostIdentifierExpr(this, e);
       } else if (expr is DisplayExpression) {
```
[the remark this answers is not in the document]
keyboardDrummer: They should. Do you want that to be done in this PR? I can also do it in a follow-up, then I can do the rest of the compile pattern matches first.

atomb (dismissed):
Looks good!

atomb (approved):

atomb (approved):

PR #3122, opened 2022-11-29
Feature: Test Generation Redesign

Dargones:
This PR focuses almost exclusively on the `DafnyTestGeneration` module and updates it with new features that include, among other things:

1. Support for function methods
2. Support for datatypes
4. Extraction of pre- and post- conditions to be used as oracles at runtime.

The new features allow applying test generation on a larger scale, e.g. on the standard library of AWS Encryption SDK.

This is a large pull request, so I would be grateful for any guidance as to how to move it forward.


Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -769,7 +769,7 @@ public class ActualBindings {
       ArgumentBindings = actuals.ConvertAll(actual => new ActualBinding(null, actual));
     }
 
-    private List<Expression> arguments; // set by ResolveActualParameters during resolution
+    public List<Expression> arguments; // set by ResolveActualParameters during resolution
```
keyboardDrummer, line 772: `ActualBindings` tries hard to keep this private, so I'm guessing it should stay that way.

Source/DafnyLanguageServer/Workspace/DocumentOptions.cs
```diff
@@ -25,11 +25,10 @@ public class DocumentOptions {
       // ShowSnippets == true enable boogie assertion's token to contain the range of expressions, not their single token
       ShowSnippetsOption.Instance.Set(DafnyOptions.O, true);
     }
-
     public List<string> AugmentedProverOptions =>
       DafnyOptions.O.ProverOptions.Concat(new List<string>() {
         "O:model_compress=false", // Replaced by "O:model.compact=false" if z3's version is > 4.8.6
-        "O:model.completion=true",
+        // "O:model.completion=true", TODO: determine if this is really necessary
```
keyboardDrummer, line 31: Please resolve the TODO

keyboardDrummer (commented):
I think we should discuss whether test generation is a core Dafny feature or an extension.

At some point we will add enough plugin API to the Dafny compiler so that code for the test generation feature _could_ be moved to a separate repository.

If test generation is an extension that would mean it should live in another repository, where it would not release in tandem with new Dafny releases, be allowed to lag behind in which Dafny version it supports and get a different priority for maintenance.

I would argue that right now it's better to consider it an extension, because to me it's not clear there's enough Dafny language developer capacity to maintain it to the bar of a core feature. In the future it could still be promoted to being a core feature, depending on Dafny user needs and the capacity of Dafny language developers.

If we decide to consider test generation an extension for now, then I think we don't need to be too strict about reviewing the code in `DafnyTestGeneration`. Of course if there are parts where you would like feedback on, I'd be happy to give them regardless.

@robin-aws @cpitclaudel could you share your thoughts on this?

Source/DafnyTestGeneration/ProgramModifier.cs
```diff
@@ -16,57 +24,79 @@ namespace DafnyTestGeneration {
...
   /// condition is met (such as when a block is visited or a path is taken)
   /// </summary>
-  public abstract class ProgramModifier : ReadOnlyVisitor {
+  public abstract class ProgramModifier {
 
     // The implementation to test.
     // If null, all implementations will be tested.
     // If not null, other implementations can be inlined.
-    protected Implementation? ImplementationToTarget;
+    protected Implementation/*?*/ ImplementationToTarget;
```
codyroux, line 32: Probably good to remove this comment.

cpitclaudel:
> I would argue that right now it's better to consider it an extension, because to me it's not clear there's enough Dafny language developer capacity to maintain it to the bar of a core feature.

Agreed (and it would let development proceed at a faster pace, without requiring core team reviews for every change).

atomb:
I like the idea of considering it an extension, but I also wanted to say that I'd already set aside time to take a look at the original draft PR that could now go to this, so I'm happy to provide some comments soon.

Source/DafnyTestGeneration/ProgramModifier.cs
Dargones: Done

Source/DafnyLanguageServer/Workspace/DocumentOptions.cs
Dargones: Have figured it out

Source/DafnyCore/AST/DafnyAst.cs
Dargones: Done. I now make use of the publicly available `AcceptArgumentExpressionsAsExactParameterList()` instead.

Dargones:
Thank you, @keyboardDrummer @cpitclaudel @atomb @codyroux for the immediate feedback on the PR. Thinking of this as an extension makes sense to me from a software engineering perspective but I am not sure how that is best translated to maintenance\review decisions. My understanding is that there is interest in using test generation for existing Dafny projects and in that case continuing to package this functionality together with Dafny could be helpful. I am also not sure yet how difficult \ possible it would be to convert this to use the plugin API - will take a look. @robin-aws  @rakamaric do you have any thought on this?

keyboardDrummer:
> My understanding is that there is interest in using test generation for existing Dafny projects and in that case continuing to package this functionality together with Dafny could be helpful.

Definitely, given that interest I'd like to have test generation be part of the dafny CLI and repo. However, we do need to be realistic about how much Dafny language developer capacity we have. If we don't have enough capacity I'd prefer to consider it an "extension", meaning it wouldn't be part of dafny-lang/dafny.

If we structure test generation as an extension, it would be released as a separate tool, so you would do `dotnet tool install dafny-generate-tests` and then it would be available as a separate CLI, or possibly even automatically add a new command to the Dafny CLI.

> I am also not sure yet how difficult \ possible it would be to convert this to use the plugin API - will take a look. @robin-aws @rakamaric do you have any thought on this?

I think if we give the SinglePassCompiler a field 
```c#
delegate CompileMemberDeclaration(MemberDecl decl, Program program, TopLevelDeclWithMembers c, IClassWriter classWriter);

IDictionary<Type, CompileMemberDeclaration> memberCompilers;
```

Then the test generation plugin could update `memberCompilers` for typeof(Method), so that it adds the `if (Attributes.Contains(m.Attributes, "synthesize")) {` check.

Similarly, for the imports and the multimatcher, we could adds fields:
```c#
delegate ImportGenerator(ConcreteSyntaxTree wr);
IList<ImportGenerator> additionaImportGenerators
```

```c#
delegate BuiltinGenerator(ConcreteSyntaxTree wr);
IList<BuiltinGenerator> additionaIBuiltinGenerators;
```

robin-aws:
Apologies for not getting to this after the earlier ping, having a look right now and will help unblock ASAP.

robin-aws:
I've not yet reviewed the PR in detail, just looked over what's changing and put some thought into the high-level questions. I'm hoping to actually review the PR as well, but there's a good chance I won't get to it until the new year.

I fully agree that it would be much better to have this functionality independent from the core Dafny implementation, so that it doesn't have to be reviewed by Dafny maintainers. In fact I would go one step further and state that it would be much better for the testing framework itself to be independent: I don't like the fact that it's baked into the language itself, especially since the approach of translating to annotations supported by target language frameworks like xUnit or JUnit means the language itself has to be opinionated on which framework is the "correct" one to use, as prior reviewers have rightfully pointed out.

The compiler plugin feature technically makes it possible to move the test generation and the raw `{:test}` support out of the repo, but it's not a small ask. I'm also not satisfied with the idea that any utilities that do this kind of kind of metaprogramming can only be implemented by writing low-level compiler logic, and I'm intending to propose some additional language features in the new year that would address that problem.

The current state is that the testing framework and the existing test generation is already part of the Dafny implementation, so this is a reasonable contribution to improve on what's already in scope for Dafny maintainers. I think it's valuable enough that we should review and hopefully accept the contribution, but also actively put time into factoring the whole testing framework out of Dafny.

I'm absolutely personally willing to spend some of my capacity on both activities. I'll try to leave at least a high-level review ASAP.

Source/DafnyCore/TestGenerationOptions.cs
```diff
@@ -10,8 +10,13 @@ public class TestGenerationOptions {
     public enum Modes { None, Block, Path };
     public Modes Mode = Modes.None;
     [CanBeNull] public string TargetMethod = null;
-    public uint? SeqLengthLimit = null;
+    public int SeqLengthLimit = -1;
```
robin-aws, line 13: This is a step backwards in terms of clarity IMO, what's the motivation?

Source/DafnyTestGeneration/DafnyInfo.cs
```diff
@@ -63,37 +377,88 @@ private class DafnyInfoExtractor : BottomUpVisitor {
           Visit(classDecl);
         } else if (d is IndDatatypeDecl datatypeDecl) {
           Visit(datatypeDecl);
+        } else if (d is NewtypeDecl newTypeDecl) {
```
robin-aws, line 380: I'd love to move `Visit(TopLevelDecl d)` up to the base `BottomUpVisitor` while we're at it. I've been annoyed previously that the code visitor implementation doesn't cover the whole AST. :)

Source/DafnyTestGeneration/DafnyInfo.cs
```diff
@@ -63,37 +377,88 @@ private class DafnyInfoExtractor : BottomUpVisitor {
...
+        Type baseType, Expression/*?*/ witness, List<TypeParameter> typeArgs) {
+        if (witness != null) {
+          info.witnessForType[newTypeName] = witness;
+          if (DafnyOptions.O.TestGenOptions.Verbose) {
+            Console.Out.WriteLine($"// Values of type {newTypeName} will be " +
+                                  $"assigned the default value of " +
+                                  $"{Printer.ExprToString(info.witnessForType[newTypeName])}");
+          }
+        } else if (DafnyOptions.O.TestGenOptions.Verbose) {
+          Console.Out.WriteLine($"// Warning: Values of type {newTypeName} " +
```
robin-aws, line 399: This sounds unsound to me, and I don't think just printing a warning is strong enough. The previous compromise was that a user could use the synthesized code even if the exit code was non-zero, so that it was a "you're voiding the warranty" kind of situation. That's the only way I'm okay with this.

Source/DafnyTestGeneration/DafnyInfo.cs
```diff
@@ -105,32 +470,116 @@ private class DafnyInfoExtractor : BottomUpVisitor {
...
 
       private new void Visit(Function f) {
-        var methodName = f.Name;
-        if (path.Count != 0) {
-          methodName = $"{string.Join(".", path)}.{methodName}";
+        info.functions[f.FullDafnyName] = f;
+      }
+    }
+
+    private class ClonerWithSubstitution : Cloner {
```
robin-aws, line 481: Does `Substituter` not work for this use case?

Source/DafnyTestGeneration/Utils.cs
```diff
@@ -1,34 +1,165 @@
...
+        typ => replacements.ContainsKey(typ.Name) ?
+          replacements[typ.Name] :
+          new UserDefinedType(typ.tok, typ.Name, typ.TypeArgs));
+    }
+
     /// <summary>
     /// Parse a string read (from a certain file) to a Dafny Program
     /// </summary>
-    public static Program? Parse(string source, string fileName = "") {
+    public static Program/*?*/ Parse(string source, string fileName = "") {
```
robin-aws, line 58: Again a step backwards?

docs/DafnyRef/Options.txt
```diff
@@ -360,30 +360,30 @@ Usage: Dafny [ option ... ] [ filename ... ]
   
   ---- Test generation options -----------------------------------------------
   
+  /warnDeadCode
```
robin-aws, line 363: FYI we've got a new `dafny` CLI that supports top-level verbs now, and the preferred way to invoke this feature is now `dafny generate-tests <options...>`. You'll have to refactor a bit when you merge from the latest, but the nice things is that you can avoid all the `/generate` prefixes. :)

robin-aws (commented):
Just did a quick high-level skim and tried to provide useful comments. That's probably the best I can do until the new year.

@atomb you mentioned earlier you had intended to review this, are you still interested in taking that on?

atomb:
> @atomb you mentioned earlier you had intended to review this, are you still interested in taking that on?

Yes. I'm now thinking I'll have a chance Monday.

Source/DafnyCore/TestGenerationOptions.cs
Dargones: I changed `null` to `-1` here so that is would be consistent with the [new CLI](https://github.com/dafny-lang/dafny/blob/72bf43a506aea1a0e7cacf63c18b53f41167bbb7/Source/DafnyTestGeneration/GenerateTestsCommand.cs#L62-L74), which uses `Option<int>`. Not sure what's the best way around this is.

Source/DafnyTestGeneration/DafnyInfo.cs
Dargones: This makes sense! Will move this bit and will see if I can draft a separate PR for any other missing branches I might find)

Source/DafnyTestGeneration/DafnyInfo.cs
Dargones: Got it, thank you for catching this - will redo.

Source/DafnyTestGeneration/DafnyInfo.cs
Dargones: The difference with `Substituter` is, I think, that I need to work with a resolved AST.

Source/DafnyTestGeneration/Utils.cs
Dargones: I removed nullability annotations mostly because they would cause warning-as-errors when interacting with code that does not use them, e.g. one in DafnyLanguageServer. Will see if I can bring them back!

docs/DafnyRef/Options.txt
Dargones: Yes, this feature is amazing! I already updated that part for the PR but will sync again in the next few days

Dargones:
Thank you @robin-aws and @keyboardDrummer, for all the feedback! I will address all the notes above in the next few days to try to help the PR and will explore ways of separating into a new repository for the next iteration of the tool.

Source/DafnyTestGeneration/TestMethod.cs
```diff
@@ -16,26 +18,34 @@ public class TestMethod {
...
     // default type to replace any type variable with
     private readonly Type defaultType = Type.Int;
     // the DafnyModel that describes the inputs to this test method
     private readonly DafnyModel dafnyModel;
-
     // Set of all types for which a {:synthesize} - annotated method is needed
     // These methods are used to get fresh instances of the corresponding types
-    private static readonly HashSet<string> TypesToSynthesize = new();
+    private static readonly Dictionary<string, UserDefinedType> TypesToSynthesize = new();
+    // is set to true whenever the tool encounters something it does not support
```
atomb, line 41: This comment seems like it refers to an older version of this field.

Source/DafnyTestGeneration/TestMethod.cs
```diff
@@ -64,12 +76,39 @@ public class TestMethod {
...
+        var returnName = "o";
+        if (dafnyInfo.IsTrait(TypesToSynthesize[typ])) {
+          var types = dafnyInfo.GetTypesForTrait(TypesToSynthesize[typ]);
+          int id = 0;
+          var typeToNameDict = new Dictionary<string, string>();
+          foreach (var resultTyp in types) {
+            typeToNameDict[resultTyp.ToString()] = "o_res_" + id++;
+          }
+          typeToNameDict[typ] = "o";
+          result += $"\nmethod {{:synthesize}} {methodName}(" +
```
atomb, line 92: It would be lovely to eventually use a pretty-printer for Dafny source, and some nice combinators to construct common patters, and emit this from an AST. Maybe it could even type check the result! But that's future work, I think.

Source/DafnyTestGeneration/TestMethod.cs
```diff
@@ -122,114 +181,437 @@ public class TestMethod {
...
+          ValueCreation.Add((mapName, asType ?? variableType, $"map[{string.Join(", ", mappingStrings)}]"));
+          return mapName;
+        case UserDefinedType tupleType when tupleType.Name.StartsWith("_System.Tuple") || tupleType.Name.StartsWith("_System._tuple"):
+          errorMessages.Add("// Failed - temporary disable datatype support");
+          var tupleName = "d" + nextValueId++;
+          ValueCreation.Add((tupleName, DafnyModel.UnknownType, "(" +
+            string.Join(",", variable.Children.Values
+            .Select(v => ExtractVariable(v.First(), null))) + ")"));
+          return tupleName;
+        case DafnyModelTypeUtils.DatatypeType dataType:
```
atomb, line 336: Yay!

atomb (changes requested):
I requested a few small-ish changes, and had some other comments that are more likely relevant to future work. I think this is almost ready!

As to treating it as an extension, I'm in favor of the idea but expect we have a bit of work to do to make that a smooth experience. Once we've made a few improvements to the command execution and plugin infrastructure, this might be a nice first component to migrate to using it.

atomb:
Oh, one other comment for future work: we should eventually compare how runtime evaluation of contracts works between this PR and the implementation of the `/testContracts` flag ([here](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Resolver/ExpectContracts.cs)). It might be that we could merge the two implementations.

Source/DafnyTestGeneration/TestMethod.cs
Dargones: Yes, outputting an AST rather than a string is high on my priority list because it would also allows people to optionally modify the AST before printing to file.

Source/DafnyTestGeneration/TestMethod.cs
Dargones: Thank you for catching this! I have updated the code to use a list instead and made sure the documentation is consistent (a set would make most sense but `UserDefinedType` is not hashable, so I was trying to get around that with a dictionary that maps a string representation to the object...)

Source/DafnyTestGeneration/README.md
```diff
@@ -35,12 +35,13 @@ We currently provide no oracle against which to compare the result. In the futur
 
 ## How to Generate Tests?
```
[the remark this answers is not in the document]
Dargones: Done!

Source/DafnyTestGeneration/GenerateTestsCommand.cs
```diff
@@ -100,3 +102,52 @@ class LoopUnrollOption : IntegerOption {
     return null!;
   }
 }
+
+internal class VerboseOption : BooleanOption {
```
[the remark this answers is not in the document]
Dargones: Have updated the CLI with the new interface - this is indeed much more compact!

Source/DafnyLanguageServer/Workspace/DocumentOptions.cs
```diff
@@ -25,7 +25,6 @@ public class DocumentOptions {
       // ShowSnippets == true enable boogie assertion's token to contain the range of expressions, not their single token
       ShowSnippetsOption.Instance.Set(DafnyOptions.O, true);
     }
-
```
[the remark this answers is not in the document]
Dargones: That was accidental - have fixed it now!

Source/DafnyCore/TestGenerationOptions.cs
```diff
@@ -53,36 +58,50 @@ public enum Modes { None, Block, Path };
...
-/warnDeadCode
-    Use block-coverage tests to warn about potential dead code.
-
 /generateTestSeqLengthLimit:<n>
-    If /testMode is not None, using this argument adds an axiom that
-    sets the length of all sequences to be no greater than <n>. This is
-    useful in conjunction with loop unrolling.
-
+    Add an axiom that sets the length of all sequences to be no greater 
+    than <n>. Negative value indicates no limit.
```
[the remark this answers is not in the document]
Dargones: I think using 0 as no limit makes sense - have updated both the new and old CLI accordingly.

Source/DafnyCore/TestGenerationOptions.cs
```diff
@@ -29,15 +34,15 @@ public enum Modes { None, Block, Path };
...
+              _ => throw new Exception("Invalid value for generateTestMode")
             };
           }
           return true;
 
         case "generateTestSeqLengthLimit":
           var limit = 0;
           if (ps.GetIntArgument(ref limit)) {
-            SeqLengthLimit = (uint)limit;
+            SeqLengthLimit = limit;
```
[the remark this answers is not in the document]
Dargones: I decided to use 0 for default as you suggest above. Have updated both the new and old CLI accordingly.

Source/DafnyCore/TestGenerationOptions.cs
Dargones: I have updated this option to be of type `uint` and to use `0` for default as per @atomb's suggestion - I think this should be better on all counts.

Source/DafnyTestGeneration/DafnyInfo.cs
Dargones: Done - Dafny will throw an error and eventually exit with a non-zero exit code (but will not stop generating tests so the user can decide what to do with the result)

Dargones:
Thank you, @atomb and @robin-aws for reviewing this! I have updated the code with your suggestions. There is one failing test in the CI but I don't think it is related to the PR (`EnsuresManyDocumentsCanBeVerifiedAtOnce` on OSX). I have more detailed comments above but here are some of the key points:

> The previous compromise was that a user could use the synthesized code even if the exit code was non-zero, so that it was a "you're voiding the warranty" kind of situation. That's the only way I'm okay with this.

Done. In the absence of a witness or another problem, Dafny will throw an error and eventually exit with a non-zero exit code (but will not stop generating tests so the user can decide what to do with the result)

> Oh, one other comment for future work: we should eventually compare how runtime evaluation of contracts works between this PR and the implementation of the /testContracts flag 

The use cases are somewhat different (I need to rename some identifiers when adding the runtime check) but I think it makes sense to eventually merge these two implementations. I will see if I can do a separate PR for that.

> Re: `/warnDeadCode` It's perhaps also worth mentioning here that we may have multiple techniques for detecting dead code in the long run. Having this as an option to dafny generate-tests would make it clear what technique it's using, whereas it's less clear here (even though having it in the "Test generation options" section helps).

The new CLI has a dedicated `find-dead-code` command, so I imagine in the future one could make the command accept an argument that specifies the method by which dead code is identified? I can also merge `generate-tests` and `find-dead-code` commands if that makes more sense. Let me know either way!

keyboardDrummer:
>> Re: /warnDeadCode It's perhaps also worth mentioning here that we may have multiple techniques for detecting dead code in the long run. Having this as an option to dafny generate-tests would make it clear what technique it's using, whereas it's less clear here (even though having it in the "Test generation options" section helps).

> The new CLI has a dedicated find-dead-code command, so I imagine in the future one could make the command accept an argument that specifies the method by which dead code is identified? I can also merge generate-tests and find-dead-code commands if that makes more sense. Let me know either way!

I would prefer the commands to focus on the use-case instead of the implementation strategy. If there are two different techniques for finding dead code, I would consider placing both of them under the `find-dead-code` command.

atomb (approved):
This looks great to me now. Sorry for taking a while to get back to you!

atomb (approved):

PR #3126, opened 2022-11-29
Issue1427 type equality

### LINKED ISSUE #1427: datatype doesn't correctly compute equality-support

keyboardDrummer:
Fixes #1427

### DIFF
--- a/Source/DafnyCore/AST/Types.cs
+++ b/Source/DafnyCore/AST/Types.cs
@@ -826,7 +826,8 @@ public abstract class Type {
   /// <summary>
   /// Returns whether or not any values of the type can be checked for equality in compiled contexts
   /// </summary>
-  public virtual bool SupportsEquality => true;
+  public virtual bool SupportsEquality =>
+    true;
 
   /// <summary>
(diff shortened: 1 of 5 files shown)


cpitclaudel:
I would be careful about this; have you looked at the following two PRs?

- https://github.com/dafny-lang/dafny/pull/2066
- https://github.com/dafny-lang/dafny/pull/2768

keyboardDrummer:
> I would be careful about this; have you looked at the following two PRs?
> 
> * [Fix computation of equality support for datatypes (fixes #1427) #2066](https://github.com/dafny-lang/dafny/pull/2066)
> * [Debug Equality #2768](https://github.com/dafny-lang/dafny/pull/2768)

Thanks! I'll close this.

PR #3130, opened 2022-11-29
Do not refer to an implicit assignment in error messages on return stmts

### LINKED ISSUE #3125: Error messages on return statements may refer to an implicit assignment
```dafny
method Foo() returns (i : int) {
  return "hello";
}
```

fails with:

```
Error: RHS (of type string) not assignable to LHS (of type int)
```

It would be better if the error message referred to the return statement instead of a 'LHS'  and an assignment.

dijkstracula:
I had a bit of time this morning, and since #3125 was derived from an essentially-not-an-issue issue that I filed, made sense that I should take a swing at resolving it!

Previously, a type error in a method `return` statement refers to an assignment, because the Resolver treats a return as an implicit update to the named return variable.  This patch checks whether an assignment is updating a formal outparameter and produces a more specific error message if so.

Given the following Dafny program,

```dafny
method Foo(s: string) returns (i : int) {
    i := "hello";
    return "hi";
}

function foo(): int {
    "hello"
}
```

The errors produced are:

```
git-issue-3125.dfy(5,6): Error: RHS (of type string) not assignable to LHS (of type int)
git-issue-3125.dfy(6,4): Error: Method return value mismatch (expected int, got string)
git-issue-3125.dfy(9,9): Error: Function body type mismatch (expected int, got string)
```

This also updates existing unit tests that expected the old error message - I think I've caught them all but CI will let me know if I'm wrong :-)

Fixes #3125 


dijkstracula:
Note that a consequence of this patch is that direct assignments to the named return variable talk about the return type of the method, rather than the assignment:

```
+++ b/Test/dafny0/ConstantErrors.dfy
@@ -41,7 +41,7 @@ module D {  // const's with inferred types

   newtype SmallInt = x | 0 <= x < 100
   method M() returns (sm: SmallInt) {
-    sm := ten;  // error: "int" is not assignable to "Smallint"
+    sm := ten;  // Method return value mismatch (expecting SmallInt, got int)
   }
```

I don't know if folks have strong opinions about this; given that during treewalking, at the point of reporting the type error we've forgotten whether the statement is an UpdateStmt or a ReturnStmt, I'm not sure there's an ideal solution here that isn't super-invasive...

Source/DafnyCore/Resolver/Resolver.cs
```diff
@@ -9742,7 +9742,22 @@ public class ResolveTypeReturn {
...
           ResolveExpression(rr.Expr, resolutionContext);
           Contract.Assert(rr.Expr.Type != null);  // follows from postcondition of ResolveExpression
-          AddAssignableConstraint(stmt.Tok, lhsType, rr.Expr.Type, "RHS (of type {1}) not assignable to LHS (of type {0})");
+
+          bool isHiddenUpdateToOutVar = false;
+          if (s.Lhs is IdentifierExpr) {
+            IdentifierExpr ie = (IdentifierExpr)s.Lhs;
+            if (ie.Var is Formal) {
+              Formal f = ((Formal)ie.Var);
+              if (!f.InParam) {
```
MikaelMayer, line 9751: Can you please make these tests in a single if?
E.g.
`if (s.Lhs is IdentifierExpr {Var: Formal {InParam: false}}) {`

Source/DafnyCore/Resolver/Resolver.cs
```diff
@@ -9742,7 +9742,22 @@ public class ResolveTypeReturn {
...
+                isHiddenUpdateToOutVar = true;
```
MikaelMayer, line 9752: Based on your PR description, the update is not necessarily hidden, it might be in plain sight.

But I would love to see your new custom message only for implicit updates. We already have a partial mechanism for that in place. Here is how I suggest you could do this:
* Leverage IsImplicit from expressions, which is currently applied only for "this" and static receivers. 
* Make this property writable on IdentifierExpr
* Make Implicit set to true when building the implicit assignment:
https://github.com/dafny-lang/dafny/blob/50c2c99822f7ebe75ce23bab13d207dce5f57633/Source/DafnyCore/Resolver/Resolver.cs#L9559-L9565
* Use this implicit to refine your error message to only the case where the identifier is implicit.

MikaelMayer (changes requested):
That sounds great ! I have an idea to mitigate the effect and use this new error message more precisely. See my comments.

Source/DafnyCore/Resolver/Resolver.cs
dijkstracula: Nice!  I wasn't aware that you could do this in C#.

Source/DafnyCore/Resolver/Resolver.cs
dijkstracula: Nice!  I went looking for something like `IsImplicit` but missed it, and this is certainly nicer than the best idea I could come up, which was to extend a subclass for `IdentifierExpr`.  Will do so now, thanks for the pointers!

dijkstracula: Done - as of 59991658c64ea9348b2fed19e2505a773e02ef16 the resolver distingushes implicit assignments (the unit test is updated to make this distinction explicit)

Test/dafny0/ResolutionErrors.dfy.expect
```diff
@@ -75,6 +75,8 @@ ResolutionErrors.dfy(491,8): Error: when allocating an object of type 'Luci', on
 ResolutionErrors.dfy(493,15): Error: class Lamb does not have an anonymous constructor
 ResolutionErrors.dfy(546,7): Error: RHS (of type List<A>) not assignable to LHS (of type List<B>) (covariant type parameter would require A <: B)
 ResolutionErrors.dfy(551,7): Error: RHS (of type List<A>) not assignable to LHS (of type List<B>) (covariant type parameter would require A <: B)
+ResolutionErrors.dfy(551,7): Error: RHS (of type List<A>) not assignable to LHS (of type List<B>) (covar
```
MikaelMayer, line 78: Why do we have this error in double now?

dijkstracula: mea culpa; I reverted the non-changing unit test changes by hand and I must have made a typo during so.  (I wonder why running tests locally didn't catch this?  Something to investigate on my end.)

MikaelMayer (approved):

PR #3145, opened 2022-12-01
fix: Add `{:inline}` to char#IsChar function

robin-aws:
Without this, even with `--unicode-char` disabled, factoring out the `0 <= n && n < 65536` expression into a separate function can still perturb verification performance unnecessarily.

I can't think of a quick way to test for this unfortunately, but it's a safe enough change I argue it's better to merge now, and revisit testing more thoroughly for such performance regressions in the integration suite.

### DIFF
--- a/Source/DafnyCore/DafnyPrelude.bpl
+++ b/Source/DafnyCore/DafnyPrelude.bpl
@@ -117,12 +117,12 @@ function {:identity} LitReal(x: real): real { x } uses {
 // ---------------------------------------------------------------
 
 #if UNICODE_CHAR
-function char#IsChar(n: int): bool {
+function {:inline} char#IsChar(n: int): bool {
   (0                  <= n && n < 55296   /* 0xD800 */) || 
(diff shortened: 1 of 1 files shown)


atomb (approved):

Source/DafnyCore/DafnyPrelude.bpl
```diff
@@ -117,12 +117,12 @@ function {:identity} LitReal(x: real): real { x } uses {
...
 
 #if UNICODE_CHAR
-function char#IsChar(n: int): bool {
+function {:inline} char#IsChar(n: int): bool {
   (0                  <= n && n < 55296   /* 0xD800 */) || 
   (57344 /* 0xE000 */ <= n && n < 1114112 /* 0x11_0000 */ )
 }
 #else
-function char#IsChar(n: int): bool {
+function {:inline} char#IsChar(n: int): bool {
```
MikaelMayer, line 125: I wish we could have this for the compiler too :-)

PR #3153, opened 2022-12-04
Revising presentation of grammar in RM

### LINKED ISSUE #2542: Extract Dafny grammar definition from reference manual into a separate reference
I would argue that the low-level BNF-style grammar definition snippets in the current reference manual are a bit misplaced, since most readers will be trying to understand Dafny features at a higher-level and will prefer to read user-facing descriptions and examples. However, I do find the simplified (compared to reading `Dafny.atg` at least :) declarations of the language grammar very handy at times.

Related to this issue, a feature request for extracting railroad diagrams from the Coco/R grammar: https://github.com/dafny-lang/dafny/issues/1244

davidcok:
Fixes #2542


robin-aws:
Definitely an improvement already! I like the pattern of opening sections with examples but also linking to the relevant grammar production.

I would also go further and not necessarily let the grammar dictate how the material is divided into sections. Expressions are a perfect example: the grammar introduces a lot of intermediate production labels like "Primary" or "Constant or Atomic" that are only necessary to get precedence right, but don't actually matter to end users. "Top-level Expressions" are really just "Expressions" to users. I'd expect mostly one section that just discusses all the available operators, with a single precedence table.

That doesn't have to be addressed all in the same PR, but it's worth keeping in mind for times when it's easier to address at the same time as we move the grammar snippets out of each section.

davidcok:
This PR was accomplished in pieces through other PRs

PR #3154, opened 2022-12-06
Improve option interface

keyboardDrummer:
### Changes

- The `Function` class is moved into its own file
- Remove `IOptionSpec`, this was an abstraction that supported both the new and the legacy option UI and model, but having this makes it more difficult to migrate away from the legacy parts. The new interface for defining options has one clean new part that is simply `System.CommandLine.Option` with no abstraction over it, and uses separate `RegisterLegacyX` calls to register legacy UI and model bindings.
- Remove the options `show-snippets, useBaseFileName, print, rprint, bprint, warn-as-errors` from `dafny server`.
- Removing `ICommandSpec` is left for another PR

### Testing
None needed since this is a refactoring


MikaelMayer:
We actually need the option show-snippets for the language server, because it determines if we display errors on full ranges or only on single tokens (the latter can sometimes be useful). Currently, this option is set to full ranges, but it might change.

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -15,28 +19,87 @@
...
   public class DafnyOptions : Bpl.CommandLineOptions {
+
+    public void ApplyBinding(Option option) {
+      if (bindings.ContainsKey(option)) {
+        bindings[option](this, Get(option));
+      }
+    }
+
+    public T Get<T>(Option<T> option) {
+      return (T)Options.OptionArguments.GetOrCreate(option, () => default(T));
```
robin-aws, line 47: Wouldn't it be better to use `() => option.GetDefaultValue()` instead?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -15,28 +19,87 @@
...
+    }
+
+    public object Get(Option option) {
+      return Options.OptionArguments[option];
+    }
+
+    public void Set(Option option, object value) {
```
robin-aws, line 54: Do we ever need this version given the more type-safe version below?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -15,28 +19,87 @@
...
-        PrintModeOption.Instance,
-        ResolvedPrintOption.Instance,
-        OldCompileVerboseOption.Instance,
-        OutputOption.Instance,
-        PluginOption.Instance,
-        UseRuntimeLibOption.Instance,
-        UnicodeCharactersOption.Instance,
-        OptimizeErasableDatatypeWrapperOption.Instance,
-      });
+    private static Dictionary<Option, Action<DafnyOptions, object>> bindings = new();
```
robin-aws, line 64:
```suggestion
    private static Dictionary<Option, Action<DafnyOptions, object>> legacyBindings = new();
```

Or are these used for anything other than legacy bindings?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -270,6 +350,15 @@ public enum IncludesModes {
             return true;
           }
 
+        case "compileVerbose": {
```
robin-aws, line 353: Why aren't this and "useRuntimeLib" handled with legacy bindings?

docs/DafnyRef/Options.txt
```diff
@@ -487,12 +483,9 @@ Usage: Dafny [ option ... ] [ filename ... ]
           target.
       2 (default) - As in 1, but only resolve method bodies in
           non-included Dafny sources.
-  /library:<file>
```
robin-aws, line 490: Interesting, what was the root cause of having `/library` duplicated?

Source/DafnyDriver/Commands/Options.cs
```diff
@@ -0,0 +1,7 @@
+using System.Collections.Generic;
+using System.CommandLine;
+
+namespace Microsoft.Dafny;
+
+public class OptionsBag {
```
robin-aws, line 6: Unused?

robin-aws (changes requested):
Very nice refactoring, great to use more of the existing `System.CommandLine.Option` functionality.

Source/DafnyCore/AST/Function.cs
```diff
@@ -0,0 +1,267 @@
+using System.Collections.Generic;
```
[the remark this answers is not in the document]
keyboardDrummer: I added quite option related code to the class so it was a sensible time to put it in its own file.

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -729,4 +745,34 @@ private record TargetPaths(string Directory, string Filename) {
     #endregion
 
   }
+
+  class NoCompiler : Compiler {
```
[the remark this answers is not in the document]
keyboardDrummer: When a command doesn't use a compiler, there's still some pointless option processing code that depends on a compiler being defined. Previously the C# compiler was implicitly used in this case, but now I had to either use it explicitly or add this `NoCompiler`.

When I do another refactoring that replaces ICommandSpec with direct `System.CommandLine.Command` usage, I will also remove NoCompiler, since then the commands that don't use a compiler won't run the option processing code that currently uses the configured compiler.

Source/DafnyCore/DafnyOptions.cs
keyboardDrummer: Yes but that doesn't exist. You can only get the default value by doing an actual parse of command line options. Maybe I can remove the default case here though.

Source/DafnyCore/DafnyOptions.cs
keyboardDrummer: They're not.

Source/DafnyCore/DafnyOptions.cs
keyboardDrummer: My thought was that we should use RegisterLegacyUi for new options that never existed in the legacy Ui, but for existing legacy options we can just keep the existing code to ensure we don't break anything.

docs/DafnyRef/Options.txt
keyboardDrummer: A coding mistake? I don't think I understand the question.

Source/DafnyCore/DafnyOptions.cs
keyboardDrummer: It is used.

Source/DafnyCore/AST/Printer.cs
```diff
@@ -9,15 +9,71 @@
...
+  public enum PrintModes {
+    Everything,
+    DllEmbed,
+    NoIncludes,
+    NoGhost
+  }
+
   public class Printer {
+
+    static Printer() {
```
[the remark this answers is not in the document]
keyboardDrummer: I will move all of the RegisterLegacyUi calls to DafnyOptions to ensure the right order and that they're not forgotten.

keyboardDrummer:
> > Remove the options show-snippets, useBaseFileName, print, rprint, bprint, warn-as-errors from dafny server.
> 
> Were all of these broken and/or 100% useless? Otherwise we're playing pretty fast and loose with backwards compatibility for a command we haven't marked as experimental.

I am playing fast and loose because the command was just released and I can't imagine anyone is using it right now.

keyboardDrummer:
> We actually need the option show-snippets for the language server, because it determines if we display errors on full ranges or only on single tokens (the latter can sometimes be useful). Currently, this option is set to full ranges, but it might change.

It's on by default now, even though there's no option for it.

keyboardDrummer:
> Very nice refactoring, great to use more of the existing `System.CommandLine.Option` functionality.

Thanks. It's silly that we didn't have this new approach to begin with. The interface this PR is replacing was what I came up with before I looked at `System.CommandLine`

Source/DafnyCore/DafnyOptions.cs
robin-aws: Ah it's quite possible I misread the `System.CommandLine` API.

Source/DafnyCore/DafnyOptions.cs
robin-aws: I get it now, you're reverting back to the old code from before you extracted `CompileVerboseOption`. Makes sense!

docs/DafnyRef/Options.txt
robin-aws: I'm just not seeing where the mistake was that the PR seems to fix. Just trying to improve my own understanding. :)

robin-aws (approved):

docs/DafnyRef/Options.txt
keyboardDrummer: It declared the documentation both in the old way and in the new but support legacy way.

PR #3157, opened 2022-12-07
Implementing %exit in the test infrasstructure

davidcok:
Adds %exit to the test infrastructure.
A lit command like // RUN: %exit n %dafny ...
where n is a number runs the command after n and tests that its exit code is `n`.

This makes obsolete the need for /countVerificationErrors:0 and dafny_0
and improves on the ! operator to allow tests for specific error codes


Test/errortests/InvalidArgument1A.dfy
```diff
@@ -0,0 +1,2 @@
+// RUN: ! %baredafny verify %args --function-syntax:2 null.dfy > "%t"
```
MikaelMayer, line 1: Can't you use %exit here as well?

MikaelMayer (commented):
This is great !

Could you also make /countVerificationError parameter explicitly deprecated in the documentation? Also, did you check that there is no more %dafny_0 in the tests?

Test/errortests/InvalidArgument1A.dfy
davidcok: Yes - once this PR is approved, a sequel will roll it out to all the tests.

davidcok:
So far this is just the implementation of %exit -- once approved I will roll it out to all the tests and get rid of /countVerificationErrors:0

MikaelMayer (approved):
I just fixed a typo in the title for the auto-merge commit. Thanks a lot for making the tests better !

PR #3160, opened 2022-12-08
Changing tests to not use dafny_0 and /countVerificationErrors:0

davidcok:


Test/errortests/InvalidArgument1.dfy.expect
```diff
@@ -1,8 +1,62 @@
-Argument '2' not recognized. Must be one of:
-	'3'
-	'4'
-	'migration3to4'
-	'experimentalDefaultGhost'
-	'experimentalDefaultCompiled'
-	'experimentalPredicateAlwaysGhost'
+Description:
```
MikaelMayer, line 1: Why is this output changing so much? Seems like something is wrong here.

Test/git-issues/github-issue-2563.dfy
```diff
@@ -1,2 +1,2 @@
-// RUN: %dafny /compile:0 /countVerificationErrors:0 '/verificationLogger:csv;log.csv' "%s" > "%t
-// RUN: %diff %s.expect %t
+// RUN: %exits-with 1 %dafny /compile:0 '/verificationLogger:csv;log.csv' "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
```
MikaelMayer, line 2: Wow well spotted !

Test/lit.site.cfg
```diff
@@ -178,7 +174,8 @@ config.substitutions.append( ('%refmanexamples', os.path.join(repositoryRoot, 'd
 config.substitutions.append( ('%os', os.name) )
 config.substitutions.append( ('%ver', ver ) )
 config.substitutions.append( ('%sed', 'sed -E' ) )
-config.substitutions.append( ('%exit', "python " + os.path.join(repositoryRoot, 'Scripts/test-exit.py') ) )
+config.substitutions.append( ('%exits-with', "python " + os.path.join(repositoryRoot, 'Scripts/test-exit.py') ) )
+config.substitutions.append( ('!', "python " + os.path.join(repositoryRoot, 'Scripts/test-exit.py') + " -z" ) )
```
MikaelMayer, line 178: Interesting substitution!  How did it work before?

MikaelMayer (commented):
I think one test's output is problematic, otherwise, great addition to make tests stricter !

Scripts/test-exit.py
```diff
@@ -1,9 +1,33 @@
+## This small script takes as first argument a test indicator
+## and the rest of the arguments are a command to run, with its
+## command-line arguments.
+## The script runs the command and then if the first argument is
+## (1) -ok -- returns success, ignoring the exit code of the command
```
robin-aws, line 5: Perhaps `-any` would be more accurate? If `-ok` is more idiomatic for scripts ignore this. :)

Scripts/test-exit.py
```diff
@@ -1,9 +1,33 @@
...
+if arg == "-ok":
+    exit(0)
+if arg == "-z":
+    if p.returncode == 0 :
+        print( 'Expected non-zero exit code' )
+        exit(1)
+    else:
+        exit(0)
+if p.returncode != int(arg) :
+    print( 'Expected exit code', arg, ', actual result is', p.returncode )
```
robin-aws, line 30:
```suggestion
    print( 'Expected exit code', arg, 'actual result is', p.returncode )
```

Since `print` includes its own commas.

Test/errortests/InvalidArgument1.dfy.expect
robin-aws: I think it's because the `2>` was replaced with `>`

Test/lit.site.cfg
davidcok: the ! command did not work before on original lit (just in the Unit version)

Scripts/test-exit.py
davidcok: Python print does not include commas in the output (by default), so this change should not happen

Test/errortests/InvalidArgument1.dfy.expect
davidcok: Oops - that change of 2> to > was inadvertent. I changed it back so that the expected output is not dependent on the help text (and contains the error message!).

This does raise a bigger policy question about what dafny output should go to stdout and what to stderr and which information should be checked by the tests. But not a question for this PR.

Test/lit.site.cfg
robin-aws: Weird, ! was working for me on the original lit at least (e.g. `// RUN: ! false` passed). It's possible it wasn't portable enough to work on all OS' for some reason though. This is a reasonable implementation as well.

robin-aws (approved):
I especially like that some tests had `%dafny_0` directly replaced by `%dafny`, meaning they didn't need the former in the first place. :)

Test/lit.site.cfg
davidcok: It did not work for me (so I fixed it) -- note that %runDafnyForEachTest does not (for me) either. I'd like to investigate that, but it is not needed for this PR

PR #3162, opened 2022-12-08
Allow --no-verify when using the verify command

keyboardDrummer:
Enables checking only if a Dafny project resolves, without also verifying it.


davidcok (approved):

MikaelMayer:
It makes sense to enable resolving without verification ! Thanks for thinking about it.

I am not a fan of `dafny verify --no-verify`. Seems like we will get pushback from the community. I'll post this here before anyone else :-)
image

I would feel much more comfortable if we had "dafny resolve" for that purpose and don't enable `--no-verify` for `dafny verify`

PR #3166, opened 2022-12-09
Make a single page of the How to FAQs

### LINKED ISSUE #3163: HowToFAQ single page view

I love https://dafny.org/dafny/HowToFAQ/ but sometimes I want to search all answers for a particular keyword. Can we get a view that puts all Q/A pairs on a single page that is searchable. 

Personally, I think this is a better solution than trying to build a keyword index or introduce tags etc, although those solutions would also be helpful for my lookup problem.

davidcok:
Fixes #3163


atomb (approved):
This looks good to me.

Unrelated to this change, the last entry in `index.md` seems to be formatted incorrectly (at least based on looking at the rendered version). Also, will the final version have a different style sheet? I have a feeling it could look a little nicer with something other than the default HTML rendering.

PR #3168, opened 2022-12-09
Add resolve command in favour of verify --no-verify

keyboardDrummer:
### Changes

Add resolve command in favour of verify --no-verify

### Testing

None, but upcoming @davidcok ?

### DIFF
--- a/Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
+++ b/Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
@@ -22,6 +22,27 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization {
   public class DiagnosticsTest : ClientBasedLanguageServerTest {
 
     [TestMethod]
+    public async Task GitIssue3155ItemWithSameKeyAlreadyBeenAdded() {
+      var source = @"
+datatype Test =
+    | A(field: int)
+    | B(field: int)
+
+
+predicate updateTest(test: Test, test': Test)
+{
+    test' == test.(field := 1)
+}
+".TrimStart();
+      var documentItem = CreateTestDocument(source);
+      await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+
+      var diagnostics = await GetLastDiagnostics(documentItem, CancellationToken);
+      Assert.AreEqual(0, diagnostics.Length);
+      await AssertNoDiagnosticsAreComing(CancellationToken);
+    }
+
+    [TestMethod]
     public async Task GitIssue3062CrashOfLanguageServer() {
       var source = @"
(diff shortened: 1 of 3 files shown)


Source/DafnyDriver/Commands/VerifyCommand.cs
```diff
@@ -22,3 +22,19 @@ class VerifyCommand : ICommandSpec {
     dafnyOptions.Compile = false;
   }
 }
+
+class ResolveCommand : ICommandSpec {
+  public IEnumerable<Option> Options => ICommandSpec.ConsoleOutputOptions.
+    Concat(ICommandSpec.CommonOptions);
+
+  public Command Create() {
+    var result = new Command("resolve", "Resolve the program.");
```
davidcok, line 31:
```suggestion
    var result = new Command("resolve", "Only check for parse and resolution errors.");
```

Source/DafnyDriver/Commands/TranslateCommand.cs
```diff
@@ -11,8 +11,7 @@ class TranslateCommand : ICommandSpec {
       CommonOptionBag.Output,
       CommonOptionBag.CompileVerbose,
       CommonOptionBag.IncludeRuntime,
-    }.Concat(ICommandSpec.VerificationOptions).
```
davidcok, line 14: I would think that the translate command would still allow --no-verify

davidcok (changes requested):
Thanks for this - I did agree with Mikael's comment.

Source/DafnyDriver/Commands/BuildCommand.cs
```diff
@@ -9,8 +9,7 @@ class BuildCommand : ICommandSpec {
   public IEnumerable<Option> Options => new Option[] {
     CommonOptionBag.Output,
     CommonOptionBag.CompileVerbose,
-  }.Concat(ICommandSpec.VerificationOptions).
-    Concat(ICommandSpec.ExecutionOptions).
+  }.Concat(ICommandSpec.ExecutionOptions).
```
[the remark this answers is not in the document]
keyboardDrummer: Maybe we could tweak the formatter to enforce your preference? I'm ambivalent about both options.

davidcok (approved):

Source/DafnyDriver/Commands/VerifyCommand.cs
MikaelMayer: Great suggestion !

MikaelMayer (approved):
Great job at separating the two concepts. Now my only (ultra-small) worry is that if a user makes dafny --help they might be presented the option to "resolve" first instead of "verify". But I'm not a user who uses the command line anyway so I don't have a strong opinion about what to do.
Let's keep this, this is an improvement already.

PR #3175, opened 2022-12-09
Initial C# port of auditor

atomb:
This PR integrates the functionality of the [auditor plugin](https://github.com/dafny-lang/compiler-bootstrap/tree/main/src/Tools/Auditor) as built-in functionality exposed through the `dafny audit` command.


Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -176,5 +186,15 @@ public class CommonOptionBag {
       options.DefiniteAssignmentLevel = value ? 1 : 2;
     });
 
+    DafnyOptions.RegisterLegacyBinding(ReportFileOption, (options, value) => {
+      options.AuditorReportFile = value;
```
keyboardDrummer, line 190: Instead of the field `dafnyOptions.AuditorReportFile`, use `dafnyOptions.Get(ReportFileOption)`, and then you don't need to call `RegisterLegacyBinding` either.

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -119,6 +119,16 @@ public class CommonOptionBag {
   public static readonly Option<bool> IncludeRuntime = new("--include-runtime",
     "Include the Dafny runtime as source in the target language.");
 
+  public static readonly Option<string> ReportFileOption = new("--report-file",
```
keyboardDrummer, line 122: Could you move these options to AuditCommand.cs?

Source/DafnyCore/Auditor/AuditReport.cs
```diff
@@ -0,0 +1,154 @@
...
+      "|----|--------|-------------------|------|-----|----------|\n";
+    var rows = AllAssumptions().Select(RenderAssumptionRowsMarkdown);
+    return header + String.Join("\n", rows);
+  }
+
+  public string RenderMarkdownSections() {
+    StringBuilder text = new StringBuilder();
+
+    foreach (var module in modulesWithEntries) {
+      text.AppendLine($"# Module `{module.Name}`");
```
MikaelMayer, line 102: Would it make sense to make the audit smaller by not reporting types and members if there is no audit warning in that module anyway? To achieve that, you could store the temporary strings and not emit sub-sections if they are empty. That way, the report would only contain the assumptions.

I'm not using the auditor myself so I don't know what is really useful for Dafny users. You probably know better.

Test/auditor/TestAuditor.dfy.html.expect
```diff
@@ -0,0 +1,47 @@
...
+    div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
+    ul.task-list{list-style: none;}
+    ul.task-list li input[type="checkbox"] {
+      width: 0.8em;
+      margin: 0 0.8em 0.2em -1.6em;
+      vertical-align: middle;
+    }
+    .display.math{display: block; text-align: center; margin: 0.5rem auto;}
+    table tr:nth-child(odd) td {
+      background-color: #BBBBBB;
```
MikaelMayer, line 21:
```suggestion
      background-color: #EEE;
```
It looks better if it's lighter.

Test/auditor/TestAuditor.dfy.html.expect
```diff
@@ -0,0 +1,47 @@
...
+<tr><td>GenerateBytes</td><td>True</td><td>False</td><td>True</td><td>Declaration with `{:extern}` has a requires clause.</td><td>Test external caller (maybe with `/testContracts`) or justify.</td></tr>
+<tr><td>GenerateBytes</td><td>True</td><td>False</td><td>True</td><td>Declaration with `{:extern}` has a ensures clause.</td><td>Test external callee (maybe with `/testContracts`) or justify.</td></tr>
+<tr><td>GenerateBytesWithModel</td><td>True</td><td>False</td><td>True</td><td>Declaration with `{:extern}` has a requires clause.</td><td>Test external caller (maybe with `/testContracts`) or justify.</td></tr>
+<tr><td>GenerateBytesWithModel</td><td>True</td><td>False</td><td>True</td><td>Declaration with `{:extern}` has a ensures clause.</td><td>Test external callee (maybe with `/testContracts`) or justify.</td></tr>
+<tr><td>GenerateBytesWrapper</td><td>True</td><td>False</td><td>False</td><td>Definition has `assume` statement in body.</td><td>Replace with `assert` and prove or add `{:axiom}`.</td></tr>
```
MikaelMayer, line 40:
```suggestion
<tr><td>GenerateBytesWrapper</td><td>True</td><td>False</td><td>False</td><td>Definition has `assume` statement in body.</td><td>Replace with <code>assert</code> and prove or add `{:axiom}`.</td></tr>
```
in HTML, could you please replace all backticks by code tags?

Test/auditor/TestAuditor.dfy.html.expect
```diff
@@ -0,0 +1,47 @@
...
+    div.column{flex: auto; overflow-x: auto;}
+    div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
+    ul.task-list{list-style: none;}
+    ul.task-list li input[type="checkbox"] {
+      width: 0.8em;
+      margin: 0 0.8em 0.2em -1.6em;
+      vertical-align: middle;
+    }
+    .display.math{display: block; text-align: center; margin: 0.5rem auto;}
+    table tr:nth-child(odd) td {
```
MikaelMayer, line 20:
```suggestion
    thead {
          position: sticky;
          top: 0px;
          background: white;
    }
    table tr:nth-child(odd) td {
```
Looks much nicer for bigger tables if the header is sticky :-)

Test/auditor/TestAuditor.dfy.html.expect
```diff
@@ -0,0 +1,47 @@
...
+  <link rel="stylesheet" href="assets/style.css" />
+</head>
+<body>
+<script src="https://code.jquery.com/jquery-3.6.1.min.js"></script>
+<script src="https://unpkg.com/sorttable@1.0.2/sorttable.js" crossorigin></script>
+<table>
+<tr><th>Name</th><th>Compiled</th><th>Explicit Assumption</th><th>Extern</th><th>Issue</th><th>Mitigation</th></tr>
+<tr><td>T</td><td>True</td><td>False</td><td>False</td><td>Trait method calls may not terminate (uses `{:termination false}`).</td><td>Remove or justify.</td></tr>
+<tr><td>MinusBv8NoBody</td><td>False</td><td>False</td><td>False</td><td>Ghost declaration has no body and has an ensures clause.</td><td>Provide a body or add `{:axiom}`.</td></tr>
+<tr><td>LeftShiftBV128</td><td>False</td><td>True</td><td>False</td><td>Declaration has explicit `{:axiom}` attribute.</td><td>Provide a proof, test, or other justification.</td></tr>
```
MikaelMayer, line 33:
```suggestion
<tr><td>LeftShiftBV128</td><td>False</td><td class="true">True</td><td>False</td><td>Declaration has explicit `{:axiom}` attribute.</td><td>Provide a proof, test, or other justification.</td></tr>
```
I suggest that for the second and third column of booleans, you add a class if it's true, and a slightly different color (or bold) like this in CSS:

```
true {
  font-weight: bold;
}
```

MikaelMayer (changes requested):
This looks great! A few questions and requests there and there.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,211 @@
...
+using System.Linq;
+
+namespace Microsoft.Dafny.Auditor;
+
+/// <summary>
+/// A set of assumption-related properties of a Dafny entity. Certain
+/// combinations of these properties classify the entity as containing
+/// an assumption.
+/// </summary>
+public struct AssumptionProperties {
```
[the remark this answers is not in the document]
atomb: An alternative to what's there right now but perhaps compatible with what @keyboardDrummer is suggesting would be to have a separate traversal (not a set of instance methods on the nodes) that goes directly from the tree to a set of (issue, mitigation) pairs. That is, `GetAssumptionTags` could be called `GetAssumptionDescriptions` and return a list of `(string, string) ` pairs (or a record) directly.

I'd be quite happy to move in that direction. Early in the design I wanted to avoid doing many traversals over the whole tree, but the way things worked out that wouldn't happen even with the direct implementation.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,211 @@
...
+         "Provide a valid `decreases` clause or justify the absence."));
+    }
+    if (IsTrait && MayNotTerminate) {
+      descs.Add(
+        ("Trait method calls may not terminate (uses `{:termination false}`).",
+         "Remove or justify."));
+    }
+    if (HasVerifyFalseAttribute) {
+      descs.Add(
+        ("Definition has `{:verify false}` attribute.",
```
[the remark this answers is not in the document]
atomb: Every assumption detected here is emitted as a warning by default (but can be written to a file in one of several formats, instead, as an option). So it seems to me the question is more one of which of the assumption types identified here should be warned about when not using the auditor. I could imagine that some of them might always be on, and some might be turned on by something like a `--warn-assumptions` flag. Do either of you have thoughts on that?

The fact that it may make sense to always emit some of these warnings argues even more strongly for including this functionality in the core implementation of Dafny rather than as an optional plugin.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,211 @@
...
+      descs.Add(
+        ((IsGhost ? "Ghost" : "Compiled") +
+          " declaration has no body" +
+          (EnsuresClauses is null ? "." : " and has an ensures clause."),
+        "Provide a body or add `{:axiom}`."));
+    }
+    if (IsCallable && HasExternAttribute && RequiresClauses is not null) {
+      descs.Add(
+        ("Declaration with `{:extern}` has a requires clause.",
+         "Test external caller (maybe with `/testContracts`) or justify."));
```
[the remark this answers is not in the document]
atomb: I might remove the `justify` text, or enable it optionally. The reason for it is that output produced in the `markdown-ietf` format can be used by [`Duvet`](https://github.com/awslabs/duvet), which can then keep track of whether there are comments in the source code explaining why a particular assumption has been made. In this particular case, one of those comments would probably appear on the code that's responsible for testing the contract on the extern. However, it doesn't make much sense when not using that output format.

As for `requires` clauses, it depends on how the `{:extern}` is being used. It can be used to mark code as being called by an external client (and possibly to change the compiled name in that case). If it's being used that way, then any `requires` clause is an unchecked assumption. Ideally, we would distinguish between external "imports" and "exports", and warn about `ensures` clauses in the former case and `requires` clauses in the latter case. That'll have to wait for the FFI redesign that @robin-aws is working on though.

Source/DafnyCore/Options/CommonOptionBag.cs
atomb: Will do!

Source/DafnyCore/Options/CommonOptionBag.cs
atomb: Will do!

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,211 @@
...
+  /// </summary>
+  /// <returns>whether this entity contains assumptions</returns>
+  public bool HasAssumptions() {
+    return HasAxiomAttribute ||
+           (IsCallable &&
+            (((EnsuresClauses is not null || RequiresClauses is not null) &&
+              (MissingBody || HasExternAttribute)) ||
+             HasAssumeInBody)) ||
+           (MayNotTerminate && (IsCallable || IsTrait));
+  }
```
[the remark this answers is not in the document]
atomb: I put an explanation of this in one of my previous comment responses. :) It's an assumption for code that'll be _called by_ external code.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,211 @@
...
+
+  /// <summary>
+  /// Describe all of the assumptions this entity makes.
+  /// </summary>
+  /// <returns>a list of (assumption, mitigation) pairs</returns>
+  public IEnumerable<(string, string)> Description() {
```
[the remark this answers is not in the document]
atomb: Yeah, good call. I'll do that.

Source/DafnyCore/Auditor/AuditReport.cs
atomb: If I understand what you're saying, that's already what it's doing. The `modulesWithEntries` collection includes only the modules that have audit warnings.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,211 @@
...
+           s.SubStatements.Any(StmtContainsAssume) |
+           s.SubExpressions.Any(ExprContainsAssume);
+  }
+
+  private static bool ExprContainsAssume(Expression e) {
+    if (e is null) {
+      return false;
+    }
+
+    return (e is StmtExpr es && StmtContainsAssume(es.S)) ||
```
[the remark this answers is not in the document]
atomb: Ah, good catch. I'll do that.

Test/auditor/TestAuditor.dfy.html.expect
atomb: Sounds good!

Test/auditor/TestAuditor.dfy.html.expect
atomb: I'll audit the "explicit assumption" logic, but the idea behind it is that it includes things that are clearly intended as assumptions, as opposed to those that could be accidental. So `assume {:axiom}` would be included but `assume` on its own would not. A different term might be warranted. Ideas welcome!

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
keyboardDrummer: > And I think it's appropriate to encode this information in a separate struct than trying to embed it deeply in AST nodes.

Why? I don't see a downside to putting this information into the AST nodes. I would say that whatever is mandatory for Dafny, such as parsing, printing and resolution, should be in the AST nodes. I think this auditor is similar to warnings generated during resolution.

> which in effect has the same effect as having a struct that contains assumptions.

It's not the same, because it's easy to see what methods you can override on a class, but it's not easy to see every place where all ASTs nodes are handled by a switch.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
keyboardDrummer: > To warn or not to warn, the question is more about when it is appropriate to warn users and how much. It's not debatable to warn the user during an audit. However, I'm not sure I want to clutter my diagnostics with warnings when all I want is the result of verifying a portion of code. So I wouldn't mind these warnings if they were the only thing left (meaning I'm ready to move forward, I fixed all my errors on unrelated code), but while I'm having errors in my IDE, having such warnings for annotations I added manually myself to bypass the verifier would be extremely annoying.

Could you explain what a warning means to you? 

To me a warning means I need to resolve this before I create a commit, so I feel free to ignore them or hide them until I'm happy with my code and I'm ready to polish it for a commit.

It sounds like for you a warning is something that you already want to resolve while you're still getting verification to pass. Why is that?

> Every assumption detected here is emitted as a warning by default

What do you mean? With a warning I mean that if you run `dafny verify` on a source file that includes `{:verify false}`, it returns a warning for that. Right now, that's not the case.

> I could imagine that some of them might always be on

I think I said this before, but any warning that is _not_ on by default is a UX flaw to me. We're telling users that probably they don't want to use this (because the default is OFF), but they should consider using it (that's why we added it). It's better not to overload the user with these small choices and instead just give a recommendation.

`--warn-assumptions` should be on by default.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,211 @@
...
+         "Test external callee (maybe with `/testContracts`) or justify."));
+    }
+    if (IsCallable && HasAssumeInBody) {
+      descs.Add(
+        ("Definition has `assume` statement in body.",
+         "Replace with `assert` and prove or add `{:axiom}`."));
+    }
+    if (IsCallable && MayNotTerminate) {
+      descs.Add(
+        ("Method may not terminate (uses `decreases *`).",
```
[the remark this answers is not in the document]
keyboardDrummer: But doesn't `decreases *` allow you to assert false?

```dafny
method Foo() 
  decreases * 
  ensures false {
  Foo();
}
```

> which is useful for REPL or infinitely running apps.

For apps like that, which have to consume IO in order to progress, I would suggest we introduce some "end of time" counter that decreases every time IO is consumed.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
keyboardDrummer: Can you explain what actions users can take to remove flagged code from the report? If there are situations where that is not possible, then I don't think the report will be effective.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
MikaelMayer: > Could you explain what a warning means to you?
To me a warning means I need to resolve this before I create a commit, so I feel free to ignore them or hide them until I'm happy with my code and I'm ready to polish it for a commit.
It sounds like for you a warning is something that you already want to resolve while you're still getting verification to pass. Why is that?

I don't place the bar of resolving warnings before creating a commit. 
Moving forward one sub-proof is enough for me when writing Dafny to realize I should commit my result, even if there are other assume false. These kinds of warnings should be resolved when creating a PR to the main branch however, as they should be detected by the CI.

I'm not sure of your description of a warning, but it supposes that there is a single notion of warning. But in a verification-aware language like Dafny, I started recently to strongly believe there should be two kinds of warnings: verifier warnings and compiler/resolver warnings. To me, the only reason to trigger a verifier warning is when Dafny was able to prove that the code is correct, but it found unproven assumptions, or obviously false requirements, that the user should be aware of - so that they don't commit their code knowingly.
Users writing Dafny regularly understand why, as they are using proof scaffolding techniques where warnings would not be relevant until they actually prove something. Adding warnings on top of their regular proof authoring experience where they already struggle to prove a goal would be suboptimal.
Of course, these warnings should be reported by the auditor as well.

The second notion of warnings are regular warnings triggered by any mainstream language, such as a redundant case, or a variable that looks like a existing constructor in a pattern matching. These should be reported in any case.

We don't have mechanisms to hide warnings as of today, so I'm not counting on hiding warnings as part of a workflow - and I'm not sure it would be even desirable to force people hiding verification warnings. This looks to me like a UX flaw.

On the side, I do agree that every warning should be turned on by default. But only shown when some conditions are met, as explained above.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
MikaelMayer: > As for requires clauses, it depends on how the {:extern} is being used. It can be used to mark code as being called by an external client (and possibly to change the compiled name in that case). If it's being used that way, then any requires clause is an unchecked assumption

Oh yeah indeed that makes sense. I hadn't thought of that. It's a way to fail more gracefully. How would could you provide the information that the external caller's interface is tested?

Source/DafnyCore/Auditor/AuditReport.cs
MikaelMayer: Does the filtering works as well on top-level declarations, and members? That's also something I had in mind

Test/auditor/TestAuditor.dfy.html.expect
MikaelMayer: Oh I get it. Maybe "Axiomatized assumptions", "Forced assumptions"? Or perhaps keep "Explicit assumption" but just change the error message "Definition has `assume` statement  with no `{:axiom}` in body. Either put {:axiom} or replace the assume by an assert"?

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
MikaelMayer: > Why? I don't see a downside to putting this information into the AST nodes. I would say that whatever is mandatory for Dafny, such as parsing, printing and resolution, should be in the AST nodes. I think this auditor is similar to warnings generated during resolution.

So you want to make the auditor mandatory? It looks like that's the single point of debate. If it's not mandatory, then it makes sense to keep it separate. If it's mandatory, then yes of course we should just put all its information into the AST warnings and get rid of the auditor itself.
I believe it's not mandatory as it could reduce productivity, as explained in another post.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
MikaelMayer: > Are you saying users want to allow this in their committed code?

Yes, that's actually the only useful case I know of ensuring false ! By ensuring false, you actually proved that the method never terminates and loops forever. You can do this only on methods, not functions of course.
Moreover, a method calling this other method must also have `decreases *`, and if your called method ensures false, then it guarantees that the code after the call is never executed. We could even emit a verification warning about that.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
keyboardDrummer: > I don't place the bar of resolving warnings before creating a commit.

Sorry, I should have said before pushing to CI.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
atomb: If you use Duvet, you can add a comment to the code doing the testing that declares that it tests that property. There isn't immediately a way to do that with the other report formats.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
atomb: Having thought about this some more, I think I'm convinced that embedding the assumption descriptions in the AST _does_ make sense, in part because it'll make sure that there's a clear place to put any description of assumptions that may be present in any language construct as the language evolves. That's just a question of where assumption descriptions exist, though, not a question of when or how they should be reported.

I firmly believe that there are multiple use cases, each of which warrants different "warnings", or at least different output. The key long-term goal of the auditor, in my mind, is to give a complete view of the possible caveats to the verification of a particular program. Some of those caveats may be completely unavoidable (i.e., "we assume that Z3 v4.8.5 is correct", which is not currently in the auditor report but will ultimately be included), so they don't make sense as warnings during development but do make sense as part of a complete assurance summary.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
keyboardDrummer: > By ensuring false, you actually proved that the method never terminates and loops forever.

I see. I agree then it shouldn't be a warning. Should methods with `decreases *` be allowed to have an ensures clause, or maybe those give warnings?

Source/DafnyDriver/Commands/AuditCommand.cs
```diff
@@ -21,5 +57,9 @@ class AuditCommand : ICommandSpec {
   public void PostProcess(DafnyOptions dafnyOptions, Options options, InvocationContext context) {
     dafnyOptions.Compile = false;
     dafnyOptions.Verify = false;
+    dafnyOptions.AuditProgram = true;
+    dafnyOptions.AuditorReportFile = dafnyOptions.Get(ReportFileOption);
```
keyboardDrummer, line 61: I would prefer not having the field `DafnyOptions.AuditorReportFile` and instead using `dafnyOptions.Get(ReportFileOption)` everywhere. That way, `DafnyOptions` doesn't contain code related to many different things, and the option code can be located purely where the option is used.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,79 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.Dafny.Auditor;
+
+public record AssumptionDescription(string issue, string mitigation, bool isExplicit) {
+  public static AssumptionDescription AxiomAttributeAssumption = new(
```
keyboardDrummer, line 8: I would inline these statics.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,79 @@
...
+      yield return decl.Name + ": " + tickIssue + " Possible mitigation: " + tickMitigation;
+    }
+  }
+}
+
+/// <summary>
+/// Combines AssumptionExtractor with AuditReport to generate a full
+/// report from a program.
+/// </summary>
+public class ReportBuilder {
```
keyboardDrummer, line 60: I feel like `ReportBuilder` doesn't carry it's own weight. I would move `BuildReport` into `AuditReport`.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -0,0 +1,79 @@
...
+    return new(
+      issue: (isGhost ? "Ghost" : "Compiled") +
+             " declaration has no body and has an ensures clause.",
+      mitigation: "Provide a body or add [{:axiom}].",
+      isExplicit: false);
+  }
+}
+
+public record Assumption(Declaration decl, IEnumerable<AssumptionDescription> assumptions) {
+  public IEnumerable<string> Warnings() {
```
keyboardDrummer, line 47: Warnings is an existing concept, a message with a source location that shows up in the IDE and in the CLI when using `dafny verify`, and this doesn't seem to refer to those warnings. Can you use a different term?

Source/DafnyCore/AST/MemberDecls.cs
```diff
@@ -466,6 +467,34 @@ public class Method : MemberDecl, TypeParameter.ParentType, IMethodCodeContext {
...
+
+      if (Attributes.Contains(Attributes, "extern") && Ens.Count > 0) {
+        yield return AssumptionDescription.ExternWithEnsures;
+      }
+
+      if (Attributes.Contains(Attributes, "extern") && Req.Count > 0) {
+        yield return AssumptionDescription.ExternWithRequires;
+      }
+
+      if (AllowsNontermination) {
```
keyboardDrummer, line 492: Should this show up in the report even if the method has no ensures clauses? Many programs can potentially run forever by design, and in Dafny you need to use `decreases *` to write those programs.

```dafny
// This method looks fine. Why would you report it?
method LoopForever() decreases * {
  while(true) decreases * {
    DoUsefulThing();
  }
}
```

```dafny
// The ensures clause is misleading, because you can't actually make use of it since a call to PointlessEnsures never finishes. I think this should be a warning.
method PointlessEnsures() 
  decreases * 
  ensures 1 == 2 {
  PointlessEnsures();
}
```

Source/DafnyCore/AST/MemberDecls.cs
```diff
@@ -466,6 +467,34 @@ public class Method : MemberDecl, TypeParameter.ParentType, IMethodCodeContext {
...
+
+      if (Children.Any(n => n is AssumeStmt)) {
+        yield return AssumptionDescription.AssumeInBody;
+      }
+
+      if (Body is null && Ens.Count > 0) {
+        yield return AssumptionDescription.NoBody(IsGhost);
+      }
+
+      if (Attributes.Contains(Attributes, "extern") && Ens.Count > 0) {
```
keyboardDrummer, line 484: What about implicit clauses coming from subtypes in the arguments? What about the modifies clause? Shouldn't anything with `{:extern}` on it show up in the report since there's always an (implicit) modifies clause. Explicitly specifying `modifies *` is not allowed by the language.

Source/DafnyDriver/Commands/AuditCommand.cs
atomb: I believe that would require moving the option definition back into `DafnyCore` instead of having it be a member of the `AuditCommand` class, since it would need to be checked from `DafnyCore` code (i.e., from the auditor code). Are you happy with having the option move back?

Source/DafnyCore/Auditor/Assumption.cs
atomb: I really like having them all in one place, so they can easily be reviewed for wording. Large string constants embedded in definitions seem like code smells to me.

Source/DafnyCore/AST/MemberDecls.cs
atomb: Yeah, this should include any potential subtypes on the return type, as well. For functions, I think anything that doesn't have a postcondition (including a constraint on the return type) is safe. But I agree that we should always include methods. Thanks for pointing that out.

atomb:
> Dafny programs currently emit errors and warnings, two types of alerts where warnings have lower priority than errors, but both should be resolved before merging code. I see these assumption descriptions as a third type of alert that has a priority lower than errors and warnings, since they may persist in the codebase. I think some of the assumption descriptions in this PR are important enough that they should be actual warnings instead. These warnings could then also be shown in the report, by collecting all warnings in the Dafny program and filtering on verification as a source.

I like this way of thinking about it. I'll mull over some refactoring with that in mind.

Source/DafnyCore/AST/MemberDecls.cs
atomb: I'm somewhat undecided about this, myself. Clearly a method with `decreases *` _and_ an `ensures` clause is an issue. Even without one, though, I think they're worth calling out, though I don't feel super strongly about that. They are necessary for certain programs, but they should also be rare. The key situation I can see occurring is that you put in `decreases *` at first because termination is hard to prove, but with the intention of removing it, and then forget to do so.

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
MikaelMayer: Yes postconditions should still be allowed. For example, if the method exits (there could be a break in the infinite loop), then the postconditions are still proved (and you cannot ensure "false" anymore). In general, it's not possible to determine if a method is going to halt or not, so we cannot issue a warning that would be relevant.

Source/DafnyCore/AST/MemberDecls.cs
MikaelMayer: It could also be that you are writing an interactive program for an arbitrary number of human commands before you exit the loop. In this case, there is no choice but use `decreases *`. However, you still want to ensure postconditions.
It would make sense down the road to generate a warning if we can prove statically that a method marked with `decreases *` can never terminate and has postconditions. For example, if the main execution path goes through a while loop and there is no break or return statement. But I think that would be a separate feature.

Source/DafnyCore/AST/MemberDecls.cs
MikaelMayer: Right, if there is a subtype in the return value, it's like having an ensures.

Functions could be implemented by methods that modify the state, so it's never possible to be sure of anything. But I guess we can leave it out for later.

Source/DafnyDriver/Commands/AuditCommand.cs
keyboardDrummer: I would move AuditCommand and options to DafnyCore then

Source/DafnyCore/Auditor/Assumption.cs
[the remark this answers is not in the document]
keyboardDrummer: I see, thanks

Source/DafnyDriver/Commands/AuditCommand.cs
atomb: I've moved the `audit` options into the `Auditor` class, so the rest of the codebase (except `AuditCommand`) doesn't need to know about them.

I'm not entirely sure I understand what goal you have in mind for the plugin mechanism here, though.

Source/DafnyCore/Auditor/Assumption.cs
atomb: Re-reading this, I think it makes sense to leave it as-is. The purpose of this method is to produce warnings of exactly the type you mention: they'll be displayed in yellow at the CLI and would show up as warnings in the IDE if it ever invoked the auditor code. See [here](https://github.com/dafny-lang/dafny/pull/3175/files#diff-40edc58c676412b086903a0ed18fe28bfaeef19dc7d758876683ac8454dc5139R79-R81).

Source/DafnyCore/AST/MemberDecls.cs
atomb: I think I'm now leaning toward not even including it in the auditor, but I'm not 100% convinced. Yes, there are definitely cases where you can't avoid using `decreases *`, so I don't think it should be a default warning. But the case where you use it as a way to temporarily avoid needing to prove termination and then forget to remove it does seem like a legitimate mistake that could undermine the guarantees that you believe you've achieved.

atomb:
After thinking some more about it (and an offline conversation), I'm thinking that the following may be candidates for default warnings:
* Anything marked with `{:verify false}`
* Anything marked with `{:termination false}`
* (maybe) Any `assume` statements

The others are clearly assumptions of some sort (arguably not `decreases *`) but very often done intentionally, and I think belong more in the auditor, which I see as a tool to be used later in development to see where you stand with respect to assurance, but that doesn't bug you with initially-irrelevant stuff up front. Any thoughts @keyboardDrummer or @MikaelMayer?

Source/DafnyCore/Auditor/AuditReport.cs
atomb: Yes, see lines 121 and 135.

Source/DafnyDriver/Commands/AuditCommand.cs
keyboardDrummer: > I'm not entirely sure I understand what goal you have in mind for the plugin mechanism here, though.

Then `Auditor/Auditor.cs`, `Auditor/AuditReport.cs` could move to DafnyDriver, and `Resolver.cs` wouldn't need changes.

Source/DafnyCore/Auditor/Auditor.cs
```diff
@@ -0,0 +1,135 @@
...
+  public enum ReportFormat { HTML, MarkdownTable, MarkdownIETF, Text }
+
+  public static readonly Option<string> ReportFileOption = new("--report-file",
+    "Specify a path to store the audit report file. Without this, the report will take the form of standard warnings.");
+  public static readonly Option<bool> CompareReportOption = new("--compare-report",
+    "Compare the report that would have been generated with the existing file given by --report-file, and fail if they differ.");
+
+  public static string SupportedReportFormats = "plain text in the format of warnings ('txt', 'text'); standalone HTML ('html'); a Markdown table ('md', 'markdown', 'md-table', 'markdown-table'); or an IETF-language document in Markdown format ('md-ietf', 'markdown-ietf')";
+
+  public static readonly Option<ReportFormat?> ReportFormatOption = new("--report-format",
```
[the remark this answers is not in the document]
atomb: For the format, I guess I could argue either way. For the file, I think just using `--file` would be confusing, because it could refer to an input (`.dfy`) file, too.

Source/DafnyCore/Auditor/Auditor.cs
```diff
@@ -0,0 +1,135 @@
...
+
+/// <summary>
+/// A rewriter pass that produces an AuditReport and writes it to either
+/// standard output or a file, in one of several formats. If neither a file
+/// or a format is provided, it reports all assumptions as warnings.
+/// </summary>
+public class Auditor : IRewriter {
+  public enum ReportFormat { HTML, MarkdownTable, MarkdownIETF, Text }
+
+  public static readonly Option<string> ReportFileOption = new("--report-file",
```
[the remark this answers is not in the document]
atomb: See above.

atomb:
I'd be open to not requiring changes to `Resolver.cs` and putting the auditor code into the list of plugins, instead. I'm not sure I see the benefit to moving it into `DafnyDriver`, though (as opposed to having it be an entirely external plugin).

keyboardDrummer:
> After thinking some more about it (and an offline conversation), I'm thinking that the following may be candidates for default warnings:
> 
> * Anything marked with `{:verify false}`
> * Anything marked with `{:termination false}`
> * (maybe) Any `assume` statements
> 
> The others are clearly assumptions of some sort (arguably not `decreases *`) but very often done intentionally, and I think belong more in the auditor, which I see as a tool to be used later in development to see where you stand with respect to assurance, but that doesn't bug you with initially-irrelevant stuff up front. Any thoughts @keyboardDrummer or @MikaelMayer?

> Anything marked with `{:verify false}`

Sounds good

> Any `assume` statements

You mean those without {:axiom} ?

> Anything marked with `{:termination false}`

`{:termination false}` is something that Dafny users currently need to put in their codebases IIUC, so having a warning on it could be OK but then I'd turn it off by default.

> I'm not sure I see the benefit to moving it into DafnyDriver

I think I'd prefer having code that can be in DafnyDriver instead of in DafnyCore, live in Driver.

atomb:
> > Any `assume` statements
> 
> You mean those without {:axiom} ?

Ah, yes, sorry for the ambiguity.
 
> > Anything marked with `{:termination false}`
> 
> `{:termination false}` is something that Dafny users currently need to put in their codebases IIUC, so having a warning on it could be OK but then I'd turn it off by default.

Sounds good to me.
 
> > I'm not sure I see the benefit to moving it into DafnyDriver
> 
> I think I'd prefer having code that can be in DafnyDriver instead of in DafnyCore, live in Driver.

Could you say anything about why? Doing so would preclude using it from the IDE, or any other library-based interface to Dafny .

Source/DafnyCore/AST/MemberDecls.cs
MikaelMayer: That's a legitimate concern. Let me alleviate it.
If in a random method, if you use decreases * to avoid proving termination, you'll actually introduce more problems than you'll solve. This is because Dafny will require you to put decreases * in the method signature as well and in all the calling methods.
This is never worth the hassle in development, so to avoid proving termination, one would usually write the assume of the termination (or assume false), which the auditor would check.

However, as part of a comprehensive report, it would still be worth knowing if the application is suppose to terminate or not. So reporting the name of all methods that use decreases * makes sense, so that someone auditing the code would verify that there are only a very small amount of them.
You could also emit a warning if a method name with decreases * does not contain Loop(), Main() or Run() for example.
But for the current auditor, I'm in favor of not taking into account decreases * because of their constraints that prevents them from being used as scaffolding.

MikaelMayer:
> > > Any `assume` statements
> > 
> > 
> > You mean those without {:axiom} ?
> 
> Ah, yes, sorry for the ambiguity. My inclination would be to make these be warnings in a separate PR, and make them available even when not running `dafny audit`. Any objection to that?

Provided that:
* Assumptions are warnings and displayed after errors in VSCode
* They are not rendered in the gutter icons until they are verification
* They are reported as "verification" warnings and not "resolution warnings".

I'm ok marking these as warnings. It will also make me easier to move forward with https://github.com/dafny-lang/dafny/pull/3116 so thanks if you volunteer to do it !

atomb:
>  My inclination would be to make these be warnings in a separate PR, and make them available even when not running `dafny audit`. Any objection to that?
> 
> Provided that:
> 
>     * Assumptions are warnings and displayed after errors in VSCode

Yep, that would be my plan (and the warnings emitted by the auditor satisfy that).
 
>     * They are not rendered in the gutter icons until they are verification

I'm not sure how to ensure this, but I'll coordinate with you to make sure it happens.

>     * They are reported as "verification" warnings and not "resolution warnings".

I just updated the auditor to do this, and any default warnings should follow that pattern.

keyboardDrummer:
> Could you say anything about why?

To prevent code unrelated to the auditor from accidentally depending on something in the auditor, so it's closer to a third-party plugin.

> Doing so would preclude using it from the IDE

I think you could still use the auditor from the IDE since the IDE can invoke the CLI from the terminal. In what other way would you like to use the auditor from the IDE? 

> , or any other library-based interface to Dafny .

What would the use-case be?

Source/DafnyCore/Auditor/Auditor.cs
```diff
@@ -0,0 +1,135 @@
...
+    "Specify a path to store the audit report file. Without this, the report will take the form of standard warnings.");
+  public static readonly Option<bool> CompareReportOption = new("--compare-report",
```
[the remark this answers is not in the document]
seebees: This makes sense. This does mean that the report needs to produce deterministic output.

seebees:
I would want both `{:axiom}` and `{:termination false}` to appear in the report.
Since both are an edge case.

If you are just reading the report for a one-off,
it is not too hard to just skip over.
On the other hand if you are using the report as part of CI,
then you can pretty easily work around these cases.
It is MUCH better to have a handle on all the existing cases of  `{:termination false}`
that you expect to exist in your code base and are there because they are needed
as opposed to some random inserted `{:termination false}` that is added in error.

atomb:
I think I see three remaining open questions with this PR:
* Should any of the issues reported by `dafny audit` _also_ be warnings when using `dafny verify` or `dafny resolve`? I think several should be, but in a separate PR.
* Should the auditor code be part of the `DafnyDriver` package? I don't particularly like this solution, though I do like the idea of making the code use the plugin interface and be capable of being separated out eventually. If we were to move the code in the `Auditor` directory into `DafnyDriver`, I'd argue that `RunAllTestsMainMethod` and `ExpectContracts` should go there, too, if not others.
* Should we remove the `--compare-report` option? It's easy enough to do, but the current behavior also seems convenient to me. I'm on the fence.

Am I missing anything else?

davidcok:
On (1) - I say yes, though I'd have to review a list of audit warnings to say which should be dafny warnings.
On (3) - I'd keep it for now. I can imagine a project that has some audit findings it is willing to live with but wanting CI to check that those findings have not changed.

keyboardDrummer:
First off, we've been discussing this PR for quite a long time, and I wonder whether it wouldn't be better to just merge it so we get more data on the feature and the implementation. We can always change things again later. I will approve this PR as is, but with comments.

> * Should the auditor code be part of the `DafnyDriver` package? I don't particularly like this solution, though I do like the idea of making the code use the plugin interface and be capable of being separated out eventually. 

In a previous comment (https://github.com/dafny-lang/dafny/pull/3175#discussion_r1045581054) I asked whether we could put the code that determines the assumptions in the AST nodes that contain those assumptions. We've now switched to that but only partially, since the code in `Method` still refers to other AST node types like `ForallStmt` and `OneBodyLoopStmt`, as opposed to have an interface `IHasAssumption : INode` that is implemented by `ForallStmt` and `OneBodyLoopStmt`, and that new AST types could implement. There's also some duplication between `Function` and `Method` that could be avoided if we had a single AST traversal that looks for `IHasAssumption` in all nodes.

However, since I'm a little unsure what this feature will evolve towards, I've changed my mind on the above and would prefer to have all the code in DafnyDriver, so the code for the feature is all in the same project and isolated from the rest.

> If we were to move the code in the `Auditor` directory into `DafnyDriver`, I'd argue that `RunAllTestsMainMethod` and `ExpectContracts` should go there, too, if not others.

Agreed. 

> Should any of the issues reported by dafny audit also be warnings when using dafny verify or dafny resolve? I think several should be, but in a separate PR.

If we keep the code from this feature isolated in DafnyDriver, then I think that question can be out of scope for this PR.

> * Should we remove the `--compare-report` option? It's easy enough to do, but the current behavior also seems convenient to me. I'm on the fence.

Let's leave it in.

keyboardDrummer (dismissed):

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -41,6 +41,10 @@ public abstract class INode {
     /// </summary>
     public abstract IEnumerable<INode> Children { get; }
 
+    public IEnumerable<INode> DeepChildren() {
```
keyboardDrummer, line 44: This is extremely subjective, but I'm used to calling a property like this `Descendants`

Source/DafnyCore/AST/MemberDecls.cs
```diff
@@ -467,6 +468,46 @@ public class Method : MemberDecl, TypeParameter.ParentType, IMethodCodeContext {
...
+
+    if (DeepChildren().Any(n => n is AssumeStmt)) {
+      yield return AssumptionDescription.AssumeInBody;
+    }
+
+    if (DeepChildren().Any(n => n is ForallStmt forallStmt && forallStmt.Body is null)) {
+      yield return AssumptionDescription.ForallWithoutBody;
+    }
+
+    if (DeepChildren().Any(n => n is OneBodyLoopStmt loopStmt && loopStmt.Body is null)) {
```
keyboardDrummer, line 490: Traversing `DeepChildren()` three times, while in practice it probably doesn't matter much, doesn't look great. I think changing this into a single `foreach` would be better.

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -195,7 +195,7 @@ public class DafnyOptions : Bpl.CommandLineOptions {
     }
 
     public DafnyOptions()
-      : base("Dafny", "Dafny program verifier", new DafnyConsolePrinter()) {
+      : base("dafny", "Dafny program verifier", new DafnyConsolePrinter()) {
```
MikaelMayer, line 198: Why this change?

MikaelMayer (dismissed):
I'll approve it as this as we discussed about it. I'm fine leaving the code for the audit as part of DafnyCore because it feels like it belongs there. It will make it easier as well to ship Dafny as a library with audit integrated, and when we will add warnings for the things that are being audited, being able to detect duplicates feels more like the code should be at the same place.
For the command to do the diff, I'm good leaving it there so that one can do a diff without having to check the OS (diff does not exist on Windows)
Let's merge it and see how it evolves.

Source/DafnyCore/DafnyOptions.cs
atomb: It makes the output of `dafny audit --help` (and other command `--help` invocations) more correct. Previously, the help text made it appear that the name of the executable was `Dafny`.

Source/DafnyCore/AST/DafnyAst.cs
atomb: Yeah, I like that name better, too. :)

Source/DafnyCore/AST/MemberDecls.cs
atomb: Even better would probably be to adopt your suggestion above to make this be an assumption on the loop itself, rather than on the method or function, and have member declarations yield all assumptions from their descendants. I'll make that change.

MikaelMayer (approved):

PR #3177, opened 2022-12-12
Fix a bug in dafny server and turn on the DafnyDriver tests in CI

keyboardDrummer:
Fix a bug in dafny server and turn on the DafnyDriver tests in CI


MikaelMayer (approved):
Is this changed covered by a test? I don't expect a regression since we want to move static to non-static code, but I haven't run the DafnyDriver.Test so I don't know if it is related.
Good job fixing this though.

keyboardDrummer:
> Is this changed covered by a test? I don't expect a regression since we want to move static to non-static code, but I haven't run the DafnyDriver.Test so I don't know if it is related. Good job fixing this though.

It's covered by the test that this PR is turning on in CI.

PR #3179, opened 2022-12-12
Undo the extra expressions added by PR #2783

atomb:
This uses `FormatTokens` to keep track of the beginning of a sequence of expressions prefixed by `&&` or `||` instead of a synthetic identity expression.

Fixes a performance regression caused by #2783.


Source/DafnyCore/Dafny.atg
```diff
@@ -3820,44 +3820,49 @@ ImpliesExpression<out Expression e0, bool allowLemma, bool allowLambda, bool all
...
-    // We conjoin "true" to the left, for two reasons. One reason is to keep track of the source
-    // location of the initial AndOp operator. The other reason is to make, when there is only
-    // one conjunct, that the type checker still checks it to be a boolean.
-    (. e0 = new BinaryExpr(x, BinaryExpr.Opcode.And, new LiteralExpr(new AutoGeneratedToken(x), true), e0); .)
+  ( AndOp                                                       (. x = t; prefixOp = t; .)
+    RelationalExpression<out e0, allowLemma, allowLambda, allowBitwiseOps> (. first = e0; .)
     { IF(IsAndOp())  /* read a conjunction as far as possible */
       AndOp                                                     (. x = t; .)
-      RelationalExpression<out e1, allowLemma, allowLambda, allowBitwiseOps>      (. e0 = new BinaryExpr(x, BinaryExpr.Opcode.And, e0, e1); .)
+      RelationalExpression<out e1, allowLemma, allowLambda, allowBitwiseOps>      (. e0 = new BinaryExpr(x, BinaryExpr.Opcode.And, e0, e1, prefixOp); .)
```
MikaelMayer, line 3830: Yes right, it's the outermost && that gets the prefix, as you guessed &&1 ((a &&2 b) &&1 c)

Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -1928,21 +1929,25 @@ public enum AccumulationOperand { None, Left, Right }
     Contract.Invariant(E1 != null);
   }
 
-  public BinaryExpr(IToken tok, Opcode op, Expression e0, Expression e1)
+  public BinaryExpr(IToken tok, Opcode op, Expression e0, Expression e1, IToken prefixOp)
```
fabiomadge, line 1932:
```suggestion
  public BinaryExpr(IToken tok, Opcode op, Expression e0, Expression e1, IToken prefixOp = null)
```
Wouldn't this be less invasive?

atomb: It would be, and that's how I originally wrote it, but @MikaelMayer pointed out that it would make it easier to make mistakes during cloning by not cloning the `prefixOp`. Does avoiding those mistakes outweigh the extra change complexity? I'm undecided.

Source/DafnyCore/PrecedenceLinter.cs
```diff
@@ -92,6 +92,8 @@ class PrecedenceLinterVisitor : TopDownVisitor<LeftMargin> {
           // a SeqSelectExpr is the open-bracket). To avoid looking at that internal structure, we instead
           // use the .StartToken for these expressions.
           column = expr.StartToken.col;
+        } else if (expr is BinaryExpr binaryExpr && binaryExpr.PrefixOp is not null) {
```
keyboardDrummer, line 95: Are you not able to do something like `binaryExpr.E0.RangeToken.EndToken.Next` so you don't need `binaryExpr.PrefixOp` and all the related changes?

atomb: I think that you're right that this entire branch of the condition isn't necessary.

The `prefixOp` parameter is still necessary, I think, to ensure that it can go into the `FormatTokens` property (which is used to compute the `StartToken` property used in the previous branch).

fabiomadge (approved):

Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -1928,21 +1929,25 @@ public enum AccumulationOperand { None, Left, Right }
...
+  public BinaryExpr(IToken tok, Opcode op, Expression e0, Expression e1, IToken prefixOp = null)
     : base(tok) {
     Contract.Requires(tok != null);
     Contract.Requires(e0 != null);
     Contract.Requires(e1 != null);
     this.Op = op;
     this.E0 = e0;
     this.E1 = e1;
+    this.PrefixOp = prefixOp;
+    if (prefixOp != null) {
```
keyboardDrummer, line 1941: Can't you compute `prefixOp` using `e0.RangeToken.End.Next` ? Not having prefixOp as an extra parameter seems quite valuable.

Source/DafnyCore/PrecedenceLinter.cs
keyboardDrummer: Then this comment applies to where FormatTokens is updated 😄

Source/DafnyCore/AST/Expressions/Expressions.cs
atomb: Ah, that may be possible. That's an artifact of doing most of this work before merging with PR #3070. Previously, `RangeToken` wasn't updated to include the leading `||` or `&&`, but now it is. I'll investigate.

atomb: It is in fact possible to remove `prefixOp`, and it's much nicer that way. Thanks for pointing it out!

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -69,6 +69,10 @@ public abstract class INode {
             }
 
             void updateStartEndTokRecursive(INode node) {
+              if (node is null) {
```
MikaelMayer, line 72: I purposedly not included the check because it should never be called on null node. Well I guess we can now put a breakpoint in this if for later!

atomb: Yeah, I suspect one of the elements of `Children` winds up being `null`, even though I agree it should be left out in that case.

I'm not too worried about this, because, if we decided to allow `null` children, this is clearly the right thing to do with them. :)

PR #3189, opened 2022-12-14
Add manual-lemma-induction option

keyboardDrummer:
### Changes

When `--manual-lemma-induction` is specified, induction is no longer applied automatically for lemmas. For boolean options it's nice if the default is false because then you don't have to type `--<option-name>=false` to turn it off, but can type `--<option-name>` to turn it on.

### Testing
Updated existing test to use the new option

---

@RustanLeino is there a place in the documentation where I can see which inductive calls are generated by the automatic lemma induction?


RustanLeino (approved):

RustanLeino:
@keyboardDrummer The best place to learn about what's going on for lemmas is this [Dafny Power User note](http://leino.science/papers/krml269.html).

PR #3192, opened 2022-12-14
No verify on resolve

keyboardDrummer:
Fix resolve command so that it does not verify.

### Testing

Added a test for the resolve command


robin-aws (approved):
:)

PR #3193, opened 2022-12-15
Move call graph building

RustanLeino:
This PR moves the building of the call graph out from the general name resolution + type inference phase of the Resolver. This call-graph building also sets the following fields (which were previously set in the resolution/inference phase, too):

- the `.IsRecursive`, `.AllCalls`, and `.ContainsQuantifier` fields of `Function`,
- the `.Uses` field of `ExtremePredicate`, and
- the .`IsRecursive` and `.AssignedAssumptionVariables` fields of `Method`.

In addition, this PR fixes some minor issues, like
- some missing/confused visitor cases,
- parsing attributes after the first RHS of `:-`,
- some missing resolution and/or printing of attribtues.

Please note: It is not the purpose of this PR to make everything beautiful in the resolver. It aspires only to move the call-graph (and related) functionality into its own resolution phase, while not making anything less beautiful.

This PR is a first step in preparing for an upcoming PR that refreshes Dafny's type inference. By itself, this PR does not change type inference; it merely moves some non-type-inference stuff out of the current type-inference phase.


Source/DafnyCore/AST/Statements/Statements.cs
```diff
@@ -245,7 +245,19 @@ public ProduceStmt(IToken tok, IToken endTok, List<AssignmentRhs> rhss)
...
+          }
+        }
+      }
+    }
+  }
+  public override IEnumerable<Expression> SpecificationSubExpressions {
+    get {
+      foreach (var e in base.SpecificationSubExpressions) { yield return e; }
+      if (Rhss != null) {
+        foreach (var rhs in Rhss) {
```
cpitclaudel, line 259: I like to write this sort of code like that:
```suggestion
        foreach (var rhs in Rhss ?? Enumerable.Empty<…>) {
```

Source/DafnyCore/AST/Statements/Statements.cs
```diff
@@ -293,16 +305,42 @@ public abstract class AssignmentRhs : INode {
...
-  public virtual IEnumerable<Expression> SubExpressions {
+  public IEnumerable<Expression> SubExpressions {
+    get {
+      foreach (var e in SpecificationSubExpressions) {
+        yield return e;
+      }
+
+      foreach (var e in NonSpecificationSubExpressions) {
+        yield return e;
+      }
```
cpitclaudel, line 320: …and this sort of code like this:

```suggestion
      foreach (var e in SpecificationSubExpressions.Concat(NonSpecificationSubExpressions)) {
        yield return e;
      }
```

Source/DafnyCore/AST/Statements/Statements.cs
RustanLeino: Changed.

keyboardDrummer (dismissed):

keyboardDrummer:
Looks great :) One alternative implementation to consider, possibly for the future, is to move the AST node specific code into the AST nodes, by declaring an interface like `ICaller` with a method `AddCallEdges(CallGraphBuilder builder, CallGraphBuilderContext context)`, and defining a type based generic traversal method like

```c#
void Traverse<ToVisit>(Program program, Action<Path<ToVisit>> onVisit);

interface Path<T> {
  T Value { get; }
  Path<INode> Parent { get; }
}

void BuildCallGraph(Program program) {
  Traverse(program, callerPath => callerPath.Value.AddCallEdges(builder, GetContextFromPath(callerPath));
}
```

And then calling it like

Source/DafnyCore/Resolver/CallGraphBuilder.cs
```diff
@@ -0,0 +1,604 @@
...
+    /// <summary>
+    /// Add edges to the callgraph.
+    /// See comment about AddCallGraphEdgeForField.
+    /// </summary>
+    private void AddTypeDependencyEdges(ICodeContext context, Type type) {
+      Contract.Requires(type != null);
+      Contract.Requires(context != null);
+      if (context is ICallable caller && type is NonProxyType) {
+        type.ForeachTypeComponent(ty => {
+          if ((ty as UserDefinedType)?.ResolvedClass is ICallable cl && caller.EnclosingModule == cl.EnclosingModule) {
```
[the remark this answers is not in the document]
RustanLeino: Declarations of newtypes, subset types, type synonyms, and datatype declarations are in the intersection. I agree it's odd to say these are "callables". I think the reason for this is that the resolver looks for cycles among such definitions (for example, a subset type whose constraint calls a function that's defined in terms of a `const` whose type is that same subset type), and that cycle detection piggy-backs on the other edges of the call graph.

I did not try to change that organization in this PR.

RustanLeino:
Thanks for the comments. I'll leave the `BuildCallGraph` suggestion as a possible future improvement. (There may also be other changes to cycle detection coming up in my other name-resolution PRs.)

keyboardDrummer (approved):

Source/DafnyCore/Resolver/CallGraphBuilder.cs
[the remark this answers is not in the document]
keyboardDrummer: Makes sense. What would such a cycle with a datatype declaration look like?

PR #3199, opened 2022-12-15
Simplify fix from PR #3179

atomb:
Simplifies the fix from PR #3179 significantly, removing the need for an extra parameter to the constructor for `BinaryExpr`.


Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -43,7 +43,7 @@ public abstract class INode {
     // Contains tokens that did not make it in the AST but are part of the expression,
     // Enables ranges to be correct.
     // TODO: Re-add format tokens where needed until we put all the formatting to replace the tok of every expression
-    protected IToken[] FormatTokens = null;
+    internal IToken[] FormatTokens = null;
```
MikaelMayer, line 46: Oh yes indeed, since we will use OwnedTokens. Makes sense.

Source/DafnyCore/Dafny.atg
```diff
@@ -3934,16 +3934,21 @@ LogicalExpression<out Expression e0, bool allowLemma, bool allowLambda, bool all
       AndOp                                                     (. x = t; .)
       RelationalExpression<out e1, allowLemma, allowLambda, allowBitwiseOps>
          (. startToken = e0.StartToken;
-            e0 = new BinaryExpr(x, BinaryExpr.Opcode.And, e0, e1, firstOp);
-            e0.RangeToken = new RangeToken(startToken, t); .)
+            var fmtToken = e0.RangeToken.EndToken.Next;
+            e0 = new BinaryExpr(x, BinaryExpr.Opcode.And, e0, e1);
+            e0.RangeToken = new RangeToken(startToken, t);
+            e0.FormatTokens = new[] { fmtToken };
```
MikaelMayer, line 3940: Isn't it true that fmtToken == x ?
In this case, i might be useless to store the format token anyway.

Source/DafnyCore/Dafny.atg
```diff
@@ -3952,15 +3957,20 @@ LogicalExpression<out Expression e0, bool allowLemma, bool allowLambda, bool all
...
     [ IF(IsAndOp()) AndOp (. SemErr(t, "Ambiguous use of && and ||. Use parentheses to disambiguate."); .) ]
     (. e0.RangeToken = new RangeToken(firstOp, t);
        if (e0 == first) {
          // There was only one disjunct. To make sure that the type checker still checks it to
          // be a boolean, we disjoin [sic] "false" to its left.
-         e0 = new BinaryExpr(x, BinaryExpr.Opcode.Or, new LiteralExpr(new AutoGeneratedToken(x), false), e0, firstOp);
+         var fmtToken = e0.RangeToken.EndToken.Next;
+         e0 = new BinaryExpr(x, BinaryExpr.Opcode.Or, new LiteralExpr(new AutoGeneratedToken(x), false), e0);
          e0.RangeToken = new RangeToken(firstOp, t);
+         e0.FormatTokens = new[] { fmtToken };
```
MikaelMayer, line 3973: That seems wrong. If we only have

`&& A`
then e0  is A and e0.RangeToken.EndToken is the token after A. This is not what we want, right?

atomb: The use of `RangeToken.EndToken.Next` was originally a suggestion from @keyboardDrummer, and it seems to have the correct behavior, even though it does surprise me. I'm curious whether he can say anything about the reasoning he had.

Source/DafnyCore/Dafny.atg
atomb: You may be right. It's possible that this can be simplified even further.

atomb: Oh, of course. Instead of `fmtToken` it should be `firstOp`, as it was when `firstOp` was passed into the constructor.

Source/DafnyCore/Dafny.atg
keyboardDrummer: It was just a guess. What does it mean to have only `&& A` @MikaelMayer ?

keyboardDrummer (approved):

PR #3211, opened 2022-12-16
Fix debugging

keyboardDrummer:
### Changes

- C# properties should be constant time to compute. This PR changes `INode.RangeToken` and `INode.OwnedTokens` from properties to method since they may traverse the AST.
- Turn Token into a `class` instead of a `record`, because the generated `ToString` method was overflowing because Tokens form cycles through the `Prev` and `Next` properties.


stefan-aws (approved):
Looks good to me!

MikaelMayer:
> C# properties should be constant time to compute. This PR changes INode.RangeToken and INode.OwnedTokens from properties to method since they may traverse the AST.

I'm sorry, we have to revert this PR, or at least warn our users about this breaking change in an element in docs/dev/. Half of this PR's change is not even necessary if we ensured RangeToken was never computed but always assigned, as discussed previously, see below.

First, for the debugger, there is an option to prevent properties from being automatically evaluated, and that include conversion to string, so you wouldn't have the problem with the generated ToString either. Please look into it. You should disable aggressive evaluation of properties. I don't agree that properties should be computed in constant time.

Even before this PR, it was possible to rely on StartToken and EndToken in many nodes to access the trivia, and I know some users already started to rely on it, before this PR. Their code will immediately break with this renaming. It's not just a "fix". You are modifying a public interface.
Moreover, I don't know if you remember, but it's only temporary that RangeToken is computed from the code itself. As you saw in Dafny.atg, it's assigned and not computed. We kept the previous computing code because the alternative was to move all the RangeToken, StartToken, OwnedToken and EndToken in a format section that would replace the "Tok", so that it's never lost. But this refactoring alone was not done in the OwnedToken PR because it was too long to preview it.
The thing is, now, once we will do this necessary change anyway, we will be left with methods that are computed in constant time, defeating the purpose of your "fix".
I'd be happy to review a PR of a code that would replace the first "Tok" in constructor with a decent format holder.
Sorry for putting trouble after this PR has been merge, but I think there hasn't been enough discussion about the fix and I feel a bit worry both for our users (for the unnecessary change of API) and for the future development (we are solving a problem one way that has better proper fix).

keyboardDrummer:
> First, for the debugger, there is an option to prevent properties from being automatically evaluated, and that include conversion to string, so you wouldn't have the problem with the generated ToString either. Please look into it. You should disable aggressive evaluation of properties.

I have turned off such options but I didn't get to a good debugging experience.

> I don't agree that properties should be computed in constant time.

Why not? A property implies that it can be computed quickly. Microsoft documentation advises using a method instead of a property when "The operation is orders of magnitude slower than a field set would be." : https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)?redirectedfrom=MSDN

> You are modifying a public interface.

I don't think we have any C# APIs for which we promise to even try to maintain consistency, or a mechanism to prevent us from doing so. If we want to have stable C# APIs, then we should introduce tests that verify certain APIs don't change, and that'll trigger an API design process.

> Moreover, I don't know if you remember, but it's only temporary that RangeToken is computed from the code itself. As you saw in Dafny.atg, it's assigned and not computed. We kept the previous computing code because the alternative was to move all the RangeToken, StartToken, OwnedToken and EndToken in a format section that would replace the "Tok", so that it's never lost. But this refactoring alone was not done in the OwnedToken PR because it was too long to preview it.
The thing is, now, once we will do this necessary change anyway, we will be left with methods that are computed in constant time, defeating the purpose of your "fix".

Once `StartToken` and `EndToken` are constant time, we should change them back into properties instead of methods.

> Their code will immediately break with this renaming.

Seems like a trivial thing to patch and not something unexpected since we've let any API user know that the APIs are subject to change, haven't we?

MikaelMayer:
> I have turned off such options but I didn't get to a good debugging experience.

Yeah I hear you. I also prefer too to view as much properties as I can, that's always helpful when debugging. So I did some research to see if we could get something not as drastic as deactivating all properties at once. And I found one: For properties that you are still concerned about, you can play around with the [DebugBrowsable attribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debuggerbrowsableattribute?redirectedfrom=MSDN&view=net-7.0). I'm fine to use this attribute in the Dafny codebase.
And since you replaced the record by a class, we don't have the problem for ToString, right?

> Why not? A property implies that it can be computed quickly. Microsoft documentation advises using a method instead of a property when "The operation is orders of magnitude slower than a field set would be." : https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)?redirectedfrom=MSDN

To me, the use of "orders of magnitude" to justify this PR does not get along with the actual implementation.
To illustrate "orders of magnitude" in the doc you quote, they mention I/O file reading, asynchronous blocks. Accessing the file system is order of magnitudes slower than iterating over a few tokens, which traverses at most each token once for every tree. For most nodes that are relevant for the implementation, it's typically dozen of tokens for the property OwnedTokens, and for RangeToken.
Would you call a for loop with 12 elements "orders of magnitude more complex than a field set"? Note the "s" to "orders".
All the remaining of the docs points to the fair use of a property in this case. RangeToken is stored in the process memory, does not result in observable side effects, always returns the same result and not a copy of an internal state.

Again, if you are concerned about this little extra computation when debugging, it's a good sign you want to use the [DebugBrowsable attribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debuggerbrowsableattribute?redirectedfrom=MSDN&view=net-7.0) that I found.

>  If we want to have stable C# APIs, then we should introduce tests that verify certain APIs don't change, and that'll trigger an API design process.

Wait, there was a test in DafnyPipeline.Test about range tokens and start tokens, and you just did the renaming there without probably considering that this was a true API test. That should have been a red flag that would have triggered the discussion as you mention it. I'm sure you meant no harm but the very existence of this test shows that this was tested public API.

> Seems like a trivial thing to patch and not something unexpected since we've let any API user know that the APIs are subject to significant change, haven't we?

If there is a thing that would bother me a lot,  as a user, is that a public API changes _back and forth_. I can understand moving in one direction, but planning to go backwards in the future just don't make sense to me, seems like poor planning.
I think we agreed before that any change to public API deserves more team discussion than just one review that did not even saw this important point (sorry @stefan-aws  you probably did not know about that potential breaking change ^^)

Given all these elements, I would like to ask you if you please to:
* Revert the methods to properties so that we don't change this tested public API
* Use `[DebugBrowsable]` attribute to prevent some properties you think might be concerning in the debugger.

Feel free to ask around if my requests are not appropriate. If you obtain the approval from a majority of the team, then my only ask would be to add a towncrier element in docs/dev/....break to indicate at least the breaking change in the release notes.

stefan-aws:
Thank you @MikaelMayer for the additional feedback. I haven't been aware of the potential implications to public API that you mentioned. Reading both of your answers, it seems to me that there is no settled consensus yet on how much consistency should generally be guaranteed. With this in mind, I think it is a good idea to broaden the discussion and obtain the feedback of some other team members, before making a discussion on how to proceed with this PR.

keyboardDrummer:
> To me, the use of "orders of magnitude" to justify this PR does not get along with the actual implementation.
To illustrate "orders of magnitude" in the doc you quote, they mention I/O file reading, asynchronous blocks. Accessing the file system is order of magnitudes slower than iterating over a few tokens, which traverses at most each token once for every tree. For most nodes that are relevant for the implementation, it's typically dozen of tokens for the property OwnedTokens, and for RangeToken.
Would you call a for loop with 12 elements "orders of magnitude more complex than a field set"? Note the "s" to "orders".
All the remaining of the docs points to the fair use of a property in this case. RangeToken is stored in the process memory, does not result in observable side effects, always returns the same result and not a copy of an internal state.

Without understanding the computation of RangeToken too well, my impression was it has the potential for traversing the entire AST. If you're saying that in practice the computation is limited to traversing <100 elements, then sure let's call it constant time and make it a property.

> Wait, there was a test in DafnyPipeline.Test about range tokens and start tokens, and you just did the renaming there without probably considering that this was a true API test. That should have been a red flag that would have triggered the discussion as you mention it. I'm sure you meant no harm but the very existence of this test shows that this was tested public API.

I consider that to be just a test, not something to prevent us from changing APIs. As I see it, we have no contract on our C# APIs.

> Use `[DebugBrowsable]` attribute to prevent some properties you think might be concerning in the debugger.

Good to know about that attribute, but it still seems better to prevent slow properties.

PR #3223, opened 2022-12-20
fix: Cross-modular override check

fabiomadge:
Fixes 2500.

To achieve this, we first give the override axiom the visibility of the overriding function. To ensure we only use the axiom in case we pass the check, we introduce an intermediate visibility level. The ensures clause of the overridden function can reference the underlying function, so we need to substitute the overridden function by the overriding function in the override check.


Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -3486,65 +3485,41 @@ public enum StmtType { NONE, ASSERT, ASSUME };
...
+        if (AlwaysUseHeap || f.ReadsHeap) {
+          argsC.Add(etran.HeapExpr);
+        }
+
+        argsC.AddRange(implInParams.Select(var => new Bpl.IdentifierExpr(f.tok, var)));
+
+        Expr funcExpC = new Bpl.NAryExpr(f.tok, funcIdC, argsC);
         var resultVar = new Bpl.IdentifierExpr(resultVariable.tok, resultVariable);
-        builder.Add(TrAssumeCmd(f.tok, Bpl.Expr.Eq(funcExpC, resultVar)));
+        builder.Add(TrAssumeCmd(f.tok, Expr.Eq(funcExpC, resultVar)));
```
RustanLeino, line 3513: I know Rider keeps saying the `Bpl.` is not necessary. I guess it's fine to get rid of it, but I've been fine with seeing the prefix, since it reminds me that `Bpl.Expr` is not `Expression`.

RustanLeino (commented):
The fix looks good. Please add some test cases.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -946,8 +944,9 @@ public partial class Translator {
       //adding assert W <= Frame’
       AddFunctionOverrideSubsetChk(f, builder, etran, localVariables, substMap, typeMap);
 
-      //adding assume Q; assert Post’;
+      //adding assume Q;
       //adding assume J.F(ins) == C.F(ins);
+      //assert Post’;
```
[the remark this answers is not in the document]
fabiomadge: The comment is outdated. The no assume anymore.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -3486,65 +3485,41 @@ public enum StmtType { NONE, ASSERT, ASSUME };
...
       foreach (var en in f.OverriddenFunction.Ens) {
-        Expression postcond = Substitute(en.E, null, substMap, typeMap);
-        bool splitHappened;  // we don't actually care
-        foreach (var s in TrSplitExpr(postcond, etran, false, out splitHappened)) {
-          if (s.IsChecked) {
-            builder.Add(Assert(f.tok, s.E, new PODesc.FunctionContractOverride(true)));
-          }
+        var receiver = new ImplicitThisExpr(f.tok) { Type = Resolver.GetThisType(f.tok, (TopLevelDeclWithMembers)f.EnclosingClass) };
+        var sub = new FunctionCallSubstituter(receiver, substMap, typeMap, f.OverriddenFunction, f);
+        Expression postcond = sub.Substitute(en.E);
```
RustanLeino, line 3518: I think this replaces too many occurrences of the function. For example, I suspect it will make the following override verify, but it mustn't:

``` dafny
trait Tr {
  function F(x: nat, other: Tr): int
    ensures x != 0 && other.F(x - 1, other) < 20 ==> F(x, other) % 2 == other.F(x - 1, other) % 2
}

class Cl extends Tr {
  function F(x: nat, other: Tr): int
    ensures 50 <= F(x, other)
}
```

Note that the calls `other.F(x - 1, other)` should not be changed by the substitution.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -2157,8 +2157,8 @@ class Specialization {
         pre = BplAnd(pre, etran.TrExpr(Substitute(req.E, null, substMap)));
       }
       // useViaContext: fh < FunctionContextHeight
-      var lowerBound = (f.EnclosingClass.EnclosingModuleDefinition.CallGraph.GetSCCRepresentativePredecessorCount(f) + 1) * 2;
-      Expr useViaContext = !InVerificationScope(f) ? Bpl.Expr.True : Expr.Le(Expr.Literal(lowerBound), etran.FunctionContextHeight());
+      var visibilityLevel = f.EnclosingClass.EnclosingModuleDefinition.CallGraph.GetSCCRepresentativePredecessorCount(f);
+      Expr useViaContext = !InVerificationScope(f) ? Bpl.Expr.True : Expr.Lt(MkFunctionHeight(visibilityLevel), etran.FunctionContextHeight());
```
RustanLeino, line 2161: Good. I like this `... * 2 + 1 < ...` better than the `(... + 1) * 2 <= ...`.

Test/git-issues/git-issue-2500.dfy
```diff
@@ -0,0 +1,22 @@
+// RUN: %dafny_0 /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+module A {
+  trait {:termination false} Trait {
+    function TotallyNotZero() : (ret: int)
```
RustanLeino, line 6: Also add some tests that use the function itself (rather than the named result value `ret`), since those cases may be tricky.

robin-aws (commented):
Great simple solution, and appreciate the helpful comments and code cleanup along the way!

I assume we don't need the same solution for pre-conditions as well, since the equality axiom comes after the assertion in that case?

Test/git-issues/git-issue-2500.dfy
```diff
@@ -0,0 +1,24 @@
...
+  function True(): (ret: bool)
+    // Missing: ensures ret != 0
+  { false }
+  function True'(): (ret: bool)
+    // Missing: ensures True'()
+  { false }
+  function Other(x: nat, free: M.Tr): bool
+    // Different receiver
+    ensures x != 0 && Other(x-1, free) ==> Other(x-1, free)
+  { false }
```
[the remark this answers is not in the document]
fabiomadge: This is a positive test, as the ensures is a tautology. If we weren't careful when lowering `Other` the, the override check could pass. I did add a version of your suggestion nevertheless and it uncovered an issue.

Source/DafnyCore/Verifier/FunctionCallSubstituter.cs
```diff
@@ -3,26 +3,26 @@
...
-        List<Expression> newArgs = SubstituteExprList(e.Args);
-        FunctionCallExpr newFce = new FunctionCallExpr(expr.tok, e.Name, receiver, e.OpenParen, e.CloseParen, newArgs, e.AtLabel);
-        if (e.Function == A) {
+      if (expr is FunctionCallExpr e) {
+        var receiver = Substitute(e.Receiver);
+        var newArgs = SubstituteExprList(e.Args);
+        var newFce = new FunctionCallExpr(e.tok, e.Name, receiver, e.OpenParen, e.CloseParen, newArgs, e.AtLabel);
+        if (e.Function == A && e.Receiver is ThisExpr && receiver is ThisExpr) {
           newFce.Function = B;
           newFce.Type = e.Type; // TODO: this may not work with type parameters.
```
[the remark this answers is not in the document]
fabiomadge: Neither? I couldn't find a situation where it affects the way I use it, Boogie has more shallow types, but there may be others, where getting this right is crucial.

robin-aws (approved):

PR #3226, opened 2022-12-20
Attempt to fix release-brew job in CI

### LINKED ISSUE #3225: Fix the running of release-brew in CI
### Dafny version

3.10.0

```dafny
brew install dafny in release-brew.yml appears to repeatedly fail on the brew link step. Will try without that, as it should not be needed now that installing dafny has a dependency on dotnet.
```

```code
Just run the CI -- it often fails for the release-brew job
```

### What happened?

As above.

Mac

davidcok:
Fixes #3225


keyboardDrummer:
Should we not remove support for installing through brew?

keyboardDrummer (approved):

.github/workflows/release-brew.yml
```diff
@@ -14,16 +14,15 @@ jobs:
     steps:
     - name: Install dafny
       run: |
-        brew install dafny || echo DONE
-        brew link --overwrite dotnet
```
keyboardDrummer, line 18: Was the intention here to overwrite a dotnet Dafny installation? Seems surprising since this actions job doesn't install Dafny through dotnet.

davidcok: It does seem surprising, which is why I removed it.

davidcok:
I think we should keep the option to install through brew. It is a very common and familiar package manager for Mac users. It certainly is for me.

Issue #3190, opened 2022-12-14
Tab completion support on `dafny` cli


seebees:
### Summary

Dafny has a lot of command line options,
I don't always remember them.
It would be great to be able to tab complete.

### Background and Motivation

When I try and run a command sometimes I forget
exactly what the command is called.
So I have to open a terminal, run `dafny /help`
search for my command and then I can run.

### Proposed Feature

`dafny path/to/file.dfy -timeLimit 30 -proc "*foo*" -ve`<tab>

```
/verifySeparately
/verifySnapshots
/verificationLogger
/verifyAllModules
/version
```

### Alternatives

This would make my life easier.
Workarounds abound :)

keyboardDrummer:
Autocomplete is a feature of the library we used to make the new Dafny CLI UI, `dafny verify` etc, but I think it currently doesn't work because we also still support the old UI.

PR #3228, opened 2022-12-20
Enable command line completion

keyboardDrummer:
Enable the [parse] and [suggest] directives, the latter which should enable command line completion to work.

Hopefully fixes https://github.com/dafny-lang/dafny/issues/3190

https://learn.microsoft.com/en-us/dotnet/standard/commandline/syntax#directives


MikaelMayer:
Can you please explain how this works?

keyboardDrummer:
> Can you please explain how this works?

For CLI completions, the completion engine will call `dafny [suggest] ..normal arguments..` and use the output of that to provide completions to the CLI. Without this PR, `dafny [suggest]` doesn't work.

Once we merge this PR and publish a new CLI version, users still need to follow the steps here to get CLI completion working: https://learn.microsoft.com/en-us/dotnet/standard/commandline/tab-completion#enable-tab-completion

MikaelMayer:
Can you please add documentation about how to enable it in our docs? After that I'll approve it.

keyboardDrummer:
> Can you please add documentation about how to enable it in our docs? After that I'll approve it.

I would first like to see it actually work, but not sure how to test that without this change

MikaelMayer (approved):
Sounds good to me.

PR #3229, opened 2022-12-20
Fix help for --quantifier-syntax option

### LINKED ISSUE #3172: Incorrect error message for --quantifier-syntax
### Dafny version

current dev

```dafny
dafny verify --quantifier-syntax:Version3 null.dfy
```

```code
dafny verify --quantifier-syntax:Version3 null.dfy
Version3 is not a valid argument to quantifier-syntax Did you mean one of the following?
Version3
Version4
```

### What happened?

The error message suggests using 'Version3' or 'Version4' when in fact the correct options are '3' and '4'

Mac

keyboardDrummer:
Fixes #3172


keyboardDrummer:
> Please add a test corresponding to the code in the issue?

A test for the help? That seems close to duplication since the definitions of the CLI are already quite high level, almost like a specification. We also don't have tests for the descriptions etc. I think what I would prefer to do is introduce a standard mechanism for mapping enum values to different names and write tests for that.

davidcok:
A test? Well yes-
a) A common practice for bug fixing is to make a (failing) test that repeats the bug, fix the bug, then keep the test that now succeeds so the bug fix does not break with some maintenance in the future
b) Any custom error path ought to have a test.

But no worries - I found this bug in the process of making an error message catalog, so there will be a test there when I merge that PR.

PR #3230, opened 2022-12-20
Accept --.dfy filename

### LINKED ISSUE #3173: Improper behavior of --
### Dafny version

3.9.1 + dev

```dafny
dafny verify -- --.dfy
```

```code
dafny verify -- --.dfy

--.dfy is not a valid argument

Description:
  Verify the program.

Usage:
...
```

### What happened?

After `--`, `--.dfy` ought to be interpreted as a legitimate file.

In fact, lines like 
`dafny verify -- -function-syntax:4 null.dfy  `
behave correctly, but
`dafny verify -- --function-syntax:4 null.dfy  `
does not

Mac

keyboardDrummer:
Fixes #3173

### DIFF
--- a/Source/DafnyCore/DafnyOptions.cs
+++ b/Source/DafnyCore/DafnyOptions.cs
@@ -67,8 +67,8 @@ features like traits or co-inductive types.".TrimStart(), "cs");
       RegisterLegacyUi(CommonOptionBag.Prelude, ParseFileInfo, "Input configuration", "dprelude");
 
       RegisterLegacyUi(CommonOptionBag.Libraries, ParseStringElement, "Compilation options", defaultValue: new List<string>());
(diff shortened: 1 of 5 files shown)


davidcok (commented):
What is the goal of the change from stain to FileInfo, ParseString to parseFileInfo? Is that related to the problem with --.dfy (for which I thought just the change in ICommandSpec.cs was needed)?

keyboardDrummer:
> What is the goal of the change from string to FileInfo, ParseString to parseFileInfo? Is that related to the problem with --.dfy (for which I thought just the change in ICommandSpec.cs was needed)?

It's unrelated. It might improve CLI completion, so it'll correctly expand filenames, once we get that working.

davidcok (approved):

Issue #1914, opened 2022-03-16
Language server not reporting error when z3 is not installed


MikaelMayer:
To reproduce this problem, for example
1) Install Dafny locally, but forget to install z3 (on an existing installation, rename the `dafny/binaries/z3` folder)
2) Set up the Dafny VSCode extension to point to the generated `DafnyLanguageServer.dll`
3) Try to verify a file.

It will say "verification started" but will never finish. It should instead warn the user about this problem.

Issue #1477, opened 2021-10-01
Running DafnyTestGeneration locally fails


keyboardDrummer:
The test generation tests fail on my machine. The solver log contains:

```
Prover error: line 6 column 28: unknown parameter 'model_compress'
Legal parameters are:
  auto_config (bool) (default: true)
  debug_ref_count (bool) (default: false)
  dot_proof_file (string) (default: proof.dot)
  dump_models (bool) (default: false)
  memory_high_watermark (unsigned int) (default: 0)
  memory_max_alloc_count (unsigned int) (default: 0)
  memory_max_size (unsigned int) (default: 0)
  model (bool) (default: true)
  model_validate (bool) (default: false)
  proof (bool) (default: false)
  rlimit (unsigned int) (default: 0)
  smtlib2_compliant (bool) (default: false)
  stats (bool) (default: false)
  timeout (unsigned int) (default: 4294967295)
  trace (bool) (default: false)
  trace_file_name (string) (default: z3.log)
  type_check (bool) (default: true)
  unsat_core (bool) (default: false)
  verbose (unsigned int) (default: 0)
  warning (bool) (default: true)
  well_sorted_check (bool) (default: false)

Fatal Error: ProverException: Prover died with no further output, perhaps it ran out of memory or was killed.
```

@Dargones any idea?

keyboardDrummer:
@Dargones any ideas?

keyboardDrummer:
Caused by using the wrong version of Z3, maybe we should add an explicit check for that: https://github.com/dafny-lang/dafny/issues/538

Dargones:
I believe `model_compress` is one of the default prover settings meaning that using 4.8.7 would probably cause the Dafny verifier to fail as well, so this is not necessarily specific to test generation. Do let me know if you would like me to change anything in counterexamples\tests!

keyboardDrummer:
FYI: @gauravpartha also ran into this problem but it revealed itself by letting the Dafny IDE crash when it started verification, showing the error "Header must provide Content-Length property". https://github.com/dafny-lang/dafny/issues/1941

atomb:
PR #3233 adds more support for detecting the version of Z3 that Dafny is using. I think the remaining work for this issue is to adapt the test generator to check that version and adapt the Z3 options it sets accordingly.

Dargones:
https://github.com/dafny-lang/dafny/pull/3711 should close this issue by making test generation check the version of Z3 it is using and selecting prover options accordingly

PR #3233, opened 2022-12-20
Improve Z3 detection

### LINKED ISSUE #2370: If Z3 can't be run, Dafny sometimes hangs instead of reporting an error
In some cases in which Dafny can't find Z3, Dafny hangs instead of reporting the expected error "Cannot find any prover executable".  This can make it confusing for the user to troubleshoot the problem.  Failure to find Z3 isn't likely to happen when using a binary release of Dafny that has Z3 bundled at the right path, but it can happen when building Dafny from source or using a custom build system.

**Steps to reproduce** (tested on my Linux machine):
1. Download and extract https://github.com/dafny-lang/dafny/releases/download/v3.7.1/dafny-3.7.1-x64-ubuntu-16.04.zip.
2. Delete the `dafny/z3` directory.
3. Save the example code below to `hang-test.dfy`.
4. Run: `./dafny/dafny /compile:0 hang-test.dfy`

Example code:
```dafny
lemma Dummy1(x: int)
  requires x == 1
  ensures x + 1 == 2 {}

lemma Dummy2(x: int)
  requires x == 1
  ensures x + 1 == 2 {}

lemma Dummy3(x: int)
  requires x == 1
  ensures x + 1 == 2 {}
```

**Expected behavior:** Dafny exits with an error message, "Cannot find any prover executable".

**Actual behavior:** Dafny hangs.

Some further testing I did suggests that the hang happens when the number of methods to be verified is greater than the value of the `/vcsCores` option, which defaults to 1.  With the example code above (which has 3 methods), I observe:
- `./dafny/dafny /compile:0 /vcsCores:3 hang-test.dfy` -> `Cannot find any prover executable` (within a few seconds)
- `./dafny/dafny /compile:0 /vcsCores:2 hang-test.dfy` -> hang

So this may be some kind of deadlock in a thread pool.

atomb:
* Be more explicit about the path of the executable to invoke.
* Identify the version and set different Z3 options accordingly.

Fixes #2370
Enables a fix for #1477
Interacts with #1914 (which may already be fixed)


Source/DafnyCore/DafnyOptions.cs
```diff
@@ -1055,35 +1058,85 @@ public enum IncludesModes {
...
-      } else {
-        var platform = (int)System.Environment.OSVersion.Platform;
 
-        var isUnix = platform == 4 || platform == 6 || platform == 128;
+        confirmedProverPath = proverPath;
+      }
 
-        var z3binName = isUnix ? "z3" : "z3.exe";
+      var platform = System.Environment.OSVersion.Platform;
+      var isUnix = platform == PlatformID.Unix || platform == PlatformID.MacOSX || (int)platform == 128;
```
MikaelMayer, line 1083: Can you please add a comment about this 128?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -1103,10 +1156,14 @@ public enum IncludesModes {
...
       SetZ3Option("type_check", "true");
-      SetZ3Option("smt.case_split", "3"); // TODO: try removing
       SetZ3Option("smt.qi.eager_threshold", "100"); // TODO: try lowering
       SetZ3Option("smt.delay_units", "true");
-      SetZ3Option("smt.arith.solver", "2");
+
+      if (z3Version is not null && z3Version.Major == 4 && z3Version.Minor == 8 && z3Version.Build == 5) {
+        // These options tend to help with Z3 4.8.5 but hurt with newer versions of Z3.
+        SetZ3Option("smt.case_split", "3");
+        SetZ3Option("smt.arith.solver", "2");
```
MikaelMayer, line 1165: How much are they hurting? Can you please describe the problem in a separate issue?

 Also consider describing a range of versions in the conditional, rather than an exact version. Perhaps refactor this:
https://github.com/dafny-lang/dafny/pull/3233/files#diff-b3d55f76ae78d57880964c42ed16a37e9e6a8a3a7ca4022cb96367bccda8d667R80

Source/DafnyCore/DafnyMain.cs
```diff
@@ -319,6 +319,12 @@ private class IncludeComparer : IComparer<Include> {
       Microsoft.Boogie.Program boogieProgram, string programId) {
       var moduleId = (programId ?? "main_program_id") + "_" + moduleName;
 
+      var proverPath = DafnyOptions.O.ProverOptions.Find(o => o.StartsWith("PROVER_PATH="));
+      if (proverPath is null && DafnyOptions.O.Verify) {
+        Console.WriteLine("Z3 not found in explicit --solver-path option, Dafny distribution, or PATH environment variable.");
```
MikaelMayer, line 324:
```suggestion
        Console.WriteLine("Z3 not found. Please either provide a path to the z3 executable using the `--solver-path <path>` option, or manually place the `z3` folder next to the Dafny executable you are using (this folder should contain `bin/z3` or `bin/z3.exe`), or set the PATH environment variable to also target the folder containing the z3 executable.");
```
Errors with explicit action items are always appreciated.

MikaelMayer (changes requested):
Some slight changes, but otherwise, thanks for migrating z3 detection from DafnyLanguageServer to Dafny and making it even more robust!

Source/DafnyCore/DafnyOptions.cs
atomb: I honestly don't know what it is. The code previously compared with three constant integers that weren't explained. I saw that two of them were equal to named constants in the `PlatformID` class, so I changed the code to refer to those, but the third one didn't correspond to any documented value. I suspect it actually shows up in practice, but I don't know when. This version seems better than the old code, even if not perfect.

Source/DafnyCore/DafnyOptions.cs
atomb: My experience is that any version newer than 4.8.5 works better with those options removed. I considered adding code to produce an error if it detects a version _older_ than 4.8.5. If I do that, then the conditional above really should just be for a single version. What do you think?

Source/DafnyCore/DafnyMain.cs
atomb: I updated the text to a slightly reworded version of your suggestion.

Source/DafnyCore/DafnyOptions.cs
MikaelMayer: If you add this error if we detect a version older than 4.8.5, then yes that's fine. Where would you add this error?
If you do not add it in this PR, then I would prefer to see a version "<=" instead of just "=".

Source/DafnyCore/DafnyOptions.cs
MikaelMayer: I would feel comfortable removing it as, according to the documentation, 128 is not a valid value:
https://learn.microsoft.com/en-us/dotnet/api/system.platformid?view=net-7.0

atomb: Sounds good to me.

Source/DafnyCore/DafnyOptions.cs
atomb: I went for changing it to `<=` for now, since it's a little awkward to a signal an error from that location and I haven't actually seen evidence that versions < 4.8.5 _don't_ work. Before we release 4.0 we may want to test those early versions and have a more complete strategy for how to handle the entire range of available versions.

MikaelMayer (approved):

PR #3240, opened 2022-12-21
Stop publishing the language server

keyboardDrummer:
Stop publishing the language server as a NuGet package, since the VSCode extension now uses the dafny CLI to start the language server.

Might helps with the nightly publish failing on the language server cleanup step. It seems related to the custom output directory `..\..\Binaries`


atomb (approved):
Looks good. I'm pretty sure anyone using the language server DLL before can use `dafny server` for newer versions of Dafny (and old versions of the NuGet package for older versions of Dafny).

PR #3241, opened 2022-12-21
Enable skip-deep-test label

keyboardDrummer:
Fixes #

### DIFF
--- a/.github/workflows/check-deep-tests-reusable.yml
+++ b/.github/workflows/check-deep-tests-reusable.yml
@@ -18,6 +18,7 @@ jobs:
         path: dafny
         submodules: true
     - uses: actions/github-script@v6
+      if: !contains(github.event.pull_request.labels.*.name, 'skip-deep-tests')
       with:
         script: |
           const script = require('${{ github.workspace }}/dafny/.github/workflows/check-for-workflow-run.js')


fabiomadge:
I prefer the current situation. If we know master is broken, we shouldn't compound the problem. I would prefer moving the time we run the deep tests instead.

keyboardDrummer:
> If we know master is broken, we shouldn't compound the problem.

That makes sense. However, I think the situation is more complex because sometimes we know master is not broken but we still have a failing nightly.

fabiomadge:
In that case the work and wall-clock time required to fix it should be negligible, right?

keyboardDrummer:
> In that case the work and wall-clock time required to fix it should be negligible, right?

What do you mean by negligible? This often prevents me from merging for some hours and requires me to push empty commits to several PRs to retrigger CI when the nightly build passes again.

fabiomadge:
Less then 5 minutes of your time and about an hour of wall-clock time. I agree that it's bad that you end up being the one fixing it disproportionately often, hence the proposal to move the time we run those tests.

keyboardDrummer:
> Less then 5 minutes of your time and about an hour of wall-clock time, if it's really just restarting the tests. I agree that it's bad that you end up being the one fixing it disproportionately often, hence the proposal to move the time we run those tests.

Yesterday it took about the entire day before we fixed the nightly, and in the mean time it's not just not possible to merge other PRs, but also not possible to trigger the integration tests on them to discover problems, so it can be a full day of lost wall-clock time.

Maybe we could change CI so it does always run the integration tests even if nightly is failing, but it only allows you to merge if nightly passed.

fabiomadge:
True, but yesterday wasn't just about restarting the tests. You can also still run test tests locally. I think that we should reserve the CI resources we have to get it running again, given that they are in fact constrained.

PR #3244, opened 2022-12-21
fix: Count proof obligations of loop headers

### LINKED ISSUE #3243: Loop invariant check omitted in absence of other proof obligations
### Dafny version

3.10.0

If the only proof obligation in a method body comes is the loop-invariant-holds-on-entry check, then apparently that check is omitted, too. The following program should report 2 errors, but it reports 0.

```dafny

method LoopWithBody(x: int)
  decreases *
{
  var i := 0;
  while i < x
    invariant i <= x // error: loop invariant does not hold on entry
    decreases *
  {
  }
}

method BodylessLoop(x: int)
{
  var i := 0;
  while i < x
    invariant i <= x // error: loop invariant does not hold on entry
}
```

Note, this is a soundness bug because it says a program is correct when it actually violates a loop invariant. However, you can't actually prove `false` from this bug, because if you added the proof obligation `false` (or even a proof obligation `true`), then the bug no longer manifests.

```code
% dafny test.dfy
test.dfy(15,2): Warning: note, this loop has no body (loop frame: i)

test.dfy(15,2): Error: a loop without a body cannot be compiled
```

### What happened?

See above.

Mac

RustanLeino:
This PR changes the verifier to increment the assertion count when it encounters a loop with non-free invariants. This increment was previously missing, which caused certain "loop invariant on entry" checks to be omitted (see #3243).

Fixes #3243


MikaelMayer (approved):
This solves this soundness bug, but now I realize that the mechanism for computing the number of assertions (and emitting nothing if there is no assertion) is stateful, which makes me wonder whether we are missing other cases as well. Isn't there a way that we wouldn't need to manually increment a variable?

RustanLeino:
Yes, it could be stateless. One way to do that would be to generate the entire Boogie procedure implementation and then apply a visitor that checks if there are any proof obligations. Indeed, Boogie could even do that instead of Dafny. It is alarming that the bug that this PR fixes has been around unnoticed for several years.

cpitclaudel:
It appears that the bug was introduced by https://github.com/dafny-lang/dafny/commit/2d4176b1 , but this PR isn't reverting the relevant part of that commit (instead it uses a more localized fix).  @RustanLeino could you have a second look to check if we might have introduced any other issue not covered by this PR with the original faulty commit?

PR #3253, opened 2022-12-22
Add release note for target->language change

keyboardDrummer:
Add release note for target->language change

### DIFF
--- a/docs/dev/news/3239.feat
+++ b/docs/dev/news/3239.feat
@@ -1,0 +1,1 @@
+For the command `translate`, rename the option `--target` into `language` and turn it into a mandatory argument instead of an optional option.


docs/dev/news/3239.feat
```diff
@@ -0,0 +1 @@
+For the command `translate`, rename the option `--target` into `language` and turn it into a mandatory argument instead of an optional option.
```
MikaelMayer, line 1:
```suggestion
For the command `translate`, removed the option `--target` and turned it into a mandatory first argument instead.
```

keyboardDrummer: What's the reason for this suggestion?

MikaelMayer: It is common but if you look carefully at the discussion, consistency with the rest of the project  is more important than a particular style. Moreover, this is not a commit message, this is a release notes. See previous release notes for inspiration to be consistent:
https://github.com/dafny-lang/dafny/releases/tag/v3.10.0
The typical release notes:
- Use nominal phrases whenever possible
- Describe in the present if it's a new capability
- Use the past tense if something was modified or fixed

We never use the imperative sense for release notes.
Any other question about my suggestion or it was the only point?

keyboardDrummer: How about we just make it past tense then?

PR #3254, opened 2022-12-22
Fixing formatting in markdown documents with html comments

davidcok:


docs/check-examples
```diff
@@ -62,7 +62,7 @@ defaultCommand=
 defaultExit=
 defaultLang=
 
-while getopts 'd:x:l:' opt; do
+while getopts 'md:x:l:' opt; do
```
MikaelMayer, line 65: What does this md means? Can you add a comment about this line?

MikaelMayer (commented):
Thanks for fixing the doc. One minor request

docs/check-examples
davidcok: That is from another PR that adds -m as an option. If you don't mind I'll comment it there so I don't have to rerun all the tests again -- the m d x and l are options the : means the preceding option letter takes an argument.

RustanLeino (approved):

PR #3255, opened 2022-12-22
refactor: Introduce ASTVisitor for call-graph computation

RustanLeino:
This PR introduces an `ASTVisitor` class that visits all expressions, statements, and user-provided types in a given list of top-level declarations. This visitation structure supports the visitation done by the recently introduced `CallGraphBuilder`, so `CallGraphBuilder` now plugs into the new `ASTVisitor`. This makes it more clear what the `CallGraphBuilder` does. Also, it prepares the stage for further refactorings (in upcoming PRs) that disentangle early stages of name resolution/type inference.


Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -977,6 +977,321 @@ public class Specification<T> : INode, IAttributeBearingDeclaration
     public override IEnumerable<INode> Children => Expressions;
   }
 
+  public abstract class ASTVisitor<VisitorContext> where VisitorContext : IASTVisitorContext {
```
keyboardDrummer, line 980: I would put this in a separate file, together with `IASTVisitorContext`.

keyboardDrummer (commented):
Nice separation of concerns : ) I'm hopeful we will generate classes such as ASTVisitor in the future to make it even more maintainable.

Source/DafnyCore/AST/DafnyAst.cs
RustanLeino: Done.

Source/DafnyCore/AST/AstVisitor.cs
```diff
@@ -0,0 +1,333 @@
...
+            foreach (BoundVar v in mc.Arguments) {
+              VisitUserProvidedType(v.Type, context);
+            }
+          }
+        }
+
+        // Visit subexpressions
+        expr.SubExpressions.Iter(ee => VisitExpression(ee, context));
+
+        // Visit substatements
```
MikaelMayer, line 259: Can you please visit substatements before subexpressions? Since all our substatements are before expressions such as lemma calls and assertions, I would prefer the visit to be in chronological order.

MikaelMayer (approved):
Hopefully we can make it as the visitor that will rule them all :-) thanks.

PR #3261, opened 2022-12-23
Adding a FAQ on loop modifies

davidcok:


docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+
+In Dafny's way of reasoning about a loop (which is typical of verifier systems), 
+the loop invariant is the only thing you can assume at the top of every iteration. 
+In other words, at the top of each loop iteration, it is as if all variables have 
+arbitrary values, with the condition that the loop invariant is satisfied. 
+Sometimes, the word _havoc_ is used to describe this: the verifier havocs all variables, 
+that is, gives all variables arbitrary values, and then assumes the loop invariant.
+
+If that’s all the verifier did, life would be a pain, because you’d have to write a l
+oop invariant about all variables in your program, even those that have nothing to do with the loop.
```
RustanLeino, line 23: l + oop == loop

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+This default behavior turns out to be right in almost all cases in practice, 
+which is why you normally don’t have to think about this issue.
+
+But when it is not, you need to be explicit about the `modifies` clause and the loop invariants. 
+In particular
+- write a `modifies` clause for the loop that includes all objects whose fields might be 
+assigned in the loop body (or listed in the modifies clauses of anything the loop body calls)
+- then write a loop invariant that includes assertions about any variable that is listed in the modifies clause
+or is an assignment target in the loop body. 
+Very typically, the invariant will give a value for a havoced variable showing its relationship to the loop index.
```
RustanLeino, line 57: I think "each havoced variable" is better than "a havoced variable" here.

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+
+For example, a loop that sets the elements of an array to some initial value might look like this:
+```dafny
+method init(a: array<int>) 
+  modifies a
+  ensures forall j | 0 <= j < a.Length :: a[j] == j;
```
RustanLeino, line 63: No semi-colon at the end of this line

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+{
+  var i: int := 0;
```
RustanLeino, line 65:
```suggestion
  var i := 0;
```

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+
+  while (i < a.Length)
```
RustanLeino, line 67:
```suggestion
  while i < a.Length
```

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+```dafny
+method init(a: array<int>) 
+  modifies a
+  ensures forall j | 0 <= j < a.Length :: a[j] == j;
+{
+  var i: int := 0;
+
+  while (i < a.Length) 
+    modifies a
+    invariant 0 <= i <= a.Length && forall j | 0 <= j < i :: a[j] == j;
```
RustanLeino, line 69:
```suggestion
    invariant 0 <= i <= a.Length && forall j | 0 <= j < i :: a[j] == j
```

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+  {
+    a[i] := i;
+    i := i+1;
```
RustanLeino, line 72:
```suggestion
    i := i + 1;
```

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+```
+
+Note the following points:
+- The method specifies that it might modify the elements of the array `a`.
+- The loop says that it modifies the same array. 
+In fact that modifies clause could be omitted
+because it would be inherited from the enclosing context
+- The loop also modifies `i`, but as `i` is a local variable of non-reference type, it need not be listed in the modifies clause.
```
RustanLeino, line 82: Omit "of non-reference type". What is said holds for local variables of any type. (I think the comment is trying to get to that the _state_ referenced by a reference-type local variable may or may not be modified. But I think that's confusing to bring up here.)

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+- The loop also has an invariant, which has two conjuncts
+   - One conjunct states what may be assumed about `i`. Even though `i` is not in the modifies clause 
+    it is an assignment target, so we need to say what may be assumed about it (prior to the loop test)
+   - The other conjunct says what is known about the elements of `a`, which depends on `i`,
```
RustanLeino, line 86: "what may be assumed" and "what is known about" are confusing. They sound like there's no associated proof obligation. These conjuncts of the loop invariant state "what is intended to hold on loop boundaries (more precisely, what is intended to hold at the top of each iteration)". One way to modify these bullets would be to say
- One conjunct talks about the local variable `i`. ...
- The other conjunct talks about the elements of `a`. ...

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+- The loop says that it modifies the same array. 
+In fact that modifies clause could be omitted
+because it would be inherited from the enclosing context
+- The loop also modifies `i`, but as `i` is a local variable of non-reference type, it need not be listed in the modifies clause.
+- The loop also has an invariant, which has two conjuncts
+   - One conjunct states what may be assumed about `i`. Even though `i` is not in the modifies clause 
+    it is an assignment target, so we need to say what may be assumed about it (prior to the loop test)
+   - The other conjunct says what is known about the elements of `a`, which depends on `i`, 
+    that is, on how many iterations of the loop have been executed
+- And then from that loop invariant and the fact that the loop has now terminated (which implies that `i == a.Length`), Dafny can prove the method's postcondition.
```
RustanLeino, line 88: "has now terminated" can be improved. I suggest

- After the loop, Dafny uses the loop invariant and the negation of the loop guard to conclude `i == a.Length`, and from that and the invariant, Dafny can prove the method's postcondition.

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+because it would be inherited from the enclosing context
+- The loop also modifies `i`, but as `i` is a local variable of non-reference type, it need not be listed in the modifies clause.
+- The loop also has an invariant, which has two conjuncts
+   - One conjunct states what may be assumed about `i`. Even though `i` is not in the modifies clause 
+    it is an assignment target, so we need to say what may be assumed about it (prior to the loop test)
+   - The other conjunct says what is known about the elements of `a`, which depends on `i`, 
+    that is, on how many iterations of the loop have been executed
+- And then from that loop invariant and the fact that the loop has now terminated (which implies that `i == a.Length`), Dafny can prove the method's postcondition. 
+
+Even when Dafny can infer an appropriate modifies clause, it cannot infer loop invariants. The user always needs to supply those (until verification technology improves).
```
RustanLeino, line 90: What would you think of:

```suggestion
Even when Dafny can infer an appropriate modifies clause, it does not infer loop invariants, so the user always needs to supply those. 
```

RustanLeino (changes requested):
Optionally, if you want to add an example where the loop modifies the heap in an "innocent" way, here's one:

``` dafny
class Counter {
  var n: nat
}

// print "hello" c.n times and then increment c.n
method PrintAndIncrement(c: Counter)
  modifies c
  ensures c.n == old(c.n) + 1
{
  for _ := 0 to c.n
    // To verify this method, the loop needs one of the following two lines:
    invariant c.n == old(c.n)
    modifies {} // empty modifies clause
  {
    PrintHello();
  }
  c.n := c.n + 1;
}

method PrintHello() {
  print "hello\n";
}
```

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
+---
+title: If I have an assertion about a class and a loop that doesn't mention (read, modify) the class, why does dafny fail to establish the assertion after the loop?
```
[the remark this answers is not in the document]
davidcok: Changed it - though the title is a quote from the original user question

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,90 @@
...
+---
+
+## Question
+
+If I have an assertion about a class and a loop that doesn't mention (read, modify) the class, 
+why does dafny fail to establish the assertion after the loop?
+
+## Answer
+
+The short answer is that you need an appropriate combination of modifies clause and
```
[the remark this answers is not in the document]
davidcok: Fixed

docs/HowToFAQ/FAQLoopModifies.md
davidcok: Fixed

docs/HowToFAQ/FAQLoopModifies.md
davidcok: Agreed

docs/HowToFAQ/FAQLoopModifies.md
davidcok: Right

docs/HowToFAQ/FAQLoopModifies.md
davidcok: Fine

docs/HowToFAQ/FAQLoopModifies.md
davidcok: Edited as you suggest. Though I think I would be confused. Are you saying that a local variable of reference type nevoid not be in a modifies clause even if its state is changed?

docs/HowToFAQ/FAQLoopModifies.md
davidcok: OK. (glossing over break statements in both formulations...)

docs/HowToFAQ/FAQLoopModifies.md
davidcok: Revised

docs/HowToFAQ/FAQLoopModifies.md
RustanLeino: All `modifies` clauses, for both methods and loops, always only talk about which parts of the heap are modified. If you want to modify a local variable in a loop, you just go ahead and do so--no specification needed, because Dafny will peer into the loop body and figure out which local variables may be modified.

docs/HowToFAQ/FAQLoopModifies.md
```diff
@@ -0,0 +1,126 @@
...
+    PrintHello();
+  }
+  c.n := c.n + 1;
+}
+
+method PrintHello() {
+  print "hello\n";
+}
+```
+The inner for-loop can be specified and the whole method verified in two different ways.
```
RustanLeino, line 117: "inner"? The next paragraphs also speak of "inner".

RustanLeino (changes requested):

RustanLeino (approved):

PR #3263, opened 2022-12-23
Documentation: Added text about body-less loops.

davidcok:
This resolves a TODO in the text of the RM.


docs/DafnyRef/Statements.md
```diff
@@ -1244,6 +1247,13 @@ The directions `to` or `downto` are contextual keywords. That is, these two
 words are part of the syntax of the `for` loop, but they are not reserved
 keywords elsewhere.
 
+Just like for while loops, the body of a for-loop may be omitted during
+verification. This suppresses attempts to check assertions (like invariants)
+that would occur at the end of the loop. Eventually, however a body must
+be provided; the compiler will not compile a method containing a body-less
+for-loop.
```
stefan-aws, line 1254: Overall this looks good to me. The only comment I have is that I find the wording here maybe slightly too strong, compared to how it is given for the while-loop. It is true, that the compiler will not compile a method with a body-less for-loop. But this doesn't mean that a body _must_ be provided (eventually). Sometimes one might only want to use the Dafny verifier.

stefan-aws (approved):

docs/DafnyRef/Statements.md
davidcok: A body-less function is equivalent to an unproved assumption. So it should be caught by any review (e.g. the auditor) that looks for threats to soundness. So in that sense, yes , you do have to eventually supply a body that shows the specifications you have been using are implementable.

PR #3275, opened 2022-12-27
Allow the :test attribute when using run and build

### LINKED ISSUE #3260: Compiling with {:test} on .NET fails
### Dafny version

3.10.0+dev

```dafny
method {:test} Main() {
  print "Hi!\n";
}
```

```code
dafny run -t:cs test.dfy


Errors compiling program into tmp2
(4995,6): error CS0246: The type or namespace name 'Xunit' could not be found (are you missing a using directive or an assembly reference?)
```

### What happened?

Should compile successfully (and run)

Mac

keyboardDrummer:
Fixes #3260

Also adds a hidden `--spill-translation` attribute for the `run`,`build` and `test` commands.

### DIFF
--- a/Source/DafnyCore/Compilers/Compiler-Csharp.cs
+++ b/Source/DafnyCore/Compilers/Compiler-Csharp.cs
@@ -3390,7 +3390,7 @@ namespace Microsoft.Dafny.Compilers {
     }
 
     private void AddTestCheckerIfNeeded(string name, Declaration decl, ConcreteSyntaxTree wr) {
-      if (DafnyOptions.O.RunAllTests || !Attributes.Contains(decl.Attributes, "test")) {
+      if (DafnyOptions.O.Compile || DafnyOptions.O.RunAllTests || !Attributes.Contains(decl.Attributes, "test")) {
         return;
       }
 
--- a/Source/DafnyCore/Options/DeveloperOptionBag.cs
+++ b/Source/DafnyCore/Options/DeveloperOptionBag.cs
@@ -9,7 +9,7 @@ public class DeveloperOptionBag {
     @"In case the Dafny source code is translated to another language, emit that translation.") {
     IsHidden = true
   };
-  
+
   public static readonly Option<bool> UseBaseFileName = new("--use-basename-for-filename",
     "When parsing use basename of file for tokens instead of the path supplied on the command line") {
     IsHidden = true
--- a/Source/DafnyCore/Options/DeveloperOptionBag.cs
+++ b/Source/DafnyCore/Options/DeveloperOptionBag.cs
@@ -5,13 +5,17 @@ namespace Microsoft.Dafny;
 
 public class DeveloperOptionBag {
 
+  public static readonly Option<bool> SpillTranslation = new("--spill-translation") {
+    IsHidden = true
+  };
+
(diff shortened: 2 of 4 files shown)


Source/DafnyCore/Compilers/Compiler-Csharp.cs
```diff
@@ -3390,7 +3390,7 @@ private class CSharpCompilationResult {
     }
 
     private void AddTestCheckerIfNeeded(string name, Declaration decl, ConcreteSyntaxTree wr) {
-      if (DafnyOptions.O.RunAllTests || !Attributes.Contains(decl.Attributes, "test")) {
+      if (DafnyOptions.O.Compile || DafnyOptions.O.RunAllTests || !Attributes.Contains(decl.Attributes, "test")) {
```
davidcok, line 3393: Just checking that the intended behavior is that methods marked {:test} are only compiled when testing

keyboardDrummer: They're always compiled when compiling, but the XUnit attribute is only added when using `dafny translate`.

Source/DafnyCore/Options/DeveloperOptionBag.cs
```diff
@@ -5,13 +5,18 @@ namespace Microsoft.Dafny;
 
 public class DeveloperOptionBag {
 
+  public static readonly Option<bool> SpillTranslation = new("--spill-translation",
+    @"In case the Dafny source code is translated to another language, emit that translation.") {
+    IsHidden = true
+  };
+  
   public static readonly Option<bool> UseBaseFileName = new("--use-basename-for-filename",
```
davidcok, line 13: There are instances of the old spelling that ought to be changed concurrently (even if the old spelling still works):
grep -r -e '--useBaseNameForFileName' ../.. | grep -v Binary
../../Test/lit.site.cfg:'--useBaseNameForFileName',
../../Source/DafnyDriver/DafnyDriver.cs:      "--useBaseNameForFileName",
../../Source/DafnyCore/Options/DeveloperOptionBag.cs:  public static readonly Option<bool> UseBaseFileName = new("--useBaseNameForFileName",
../../Source/IntegrationTests/bin/Debug/net6.0/TestFiles/LitTests/LitTest/lit.site.cfg:'--useBaseNameForFileName',
../../Source/IntegrationTests/bin/Debug/net6.0/TestFiles/LitTests/LitTest/lit.site.cfg:config.substitutions.append( ('%resolveargs', "--useBaseNameForFileName" ) )
../../docs/check-examples:##              "dafny verify --useBaseNameForFileName $F" and checks 
../../docs/check-examples:##              "dafny resolve --useBaseNameForFileName $F" and checks
../../docs/check-examples:##              "dafny run --useBaseNameForFileName $F" and, 
../../docs/check-examples:          "$dafny" $com --useBaseNameForFileName $options $text > actual

keyboardDrummer: Seems like this grep was not done on this PR

davidcok (approved):

Source/DafnyCore/Options/DeveloperOptionBag.cs
davidcok: My mistake, and I was also looking a a stale list of changes...
Thanks for making this option consistently spelled.

PR #3276, opened 2022-12-27
Setting cores to 0 will use half of the available cores

keyboardDrummer:
Setting cores to 0 will use half of the available cores.

### DIFF
--- a/Source/DafnyCore/Options/BoogieOptionBag.cs
+++ b/Source/DafnyCore/Options/BoogieOptionBag.cs
@@ -17,8 +17,8 @@ public static class BoogieOptionBag {
     ArgumentHelpName = "arguments",
   };
 
-  public static readonly Option<int> Cores = new("--cores", () => 1,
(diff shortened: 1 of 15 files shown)


Source/DafnyCore/Options/BoogieOptionBag.cs
```diff
@@ -40,7 +40,8 @@ public static class BoogieOptionBag {
         o.Parse(SplitArguments(boogieOptions).ToArray());
       }
     });
-    DafnyOptions.RegisterLegacyBinding(Cores, (o, f) => o.VcsCores = f);
+    DafnyOptions.RegisterLegacyBinding(Cores,
+      (o, f) => o.VcsCores = f == 0 ? System.Environment.ProcessorCount / 2 : (int)f);
```
cpitclaudel, line 44: Shouldn't this round up?  What happens on a machine with one core?

cpitclaudel (changes requested):

cpitclaudel (approved):

PR #3282, opened 2022-12-27
feat: More hovering information on the LSP

### LINKED ISSUE #3281: More relevant information on hovering failing assertions
### Dafny version

3.10.0

```dafny
predicate P(i: int) {
  i <= 0
}

predicate Q(i: int, j: int) {
  i == j || -i == j
}

function method Toast(i: int): int
  requires P(i)

method Test(i: int) returns (j: nat)
  ensures Q(i, j)
{
  if i < 0 {
    return -i;
  } else {
    return Toast(i);
  }
}
```

```code
Paste this code in VSCode.

If you hover over Q(i, j) (the failing postcondition), it does not tell you exactly which of the conjuncts of Q is failing
If you hover over the last "return" keyword, it just tell you that a postcondition is failing, but not which one.
If you hover over the call to Toast, it tells you that a function precondition might not hold, but it does not tell you which one.
```

### What happened?

Currently, when we hover in the IDE an error message, we don't get all the related locations nor the related code.
Even worse, when we introduce a resolution error, hovering over it hides it under a potential ton of obsolete verification error. They should be removed from hovering.
All of this is extremely frustrating for a advanced Dafny developer like me, when using nested predicates, etc.
It could be marked as a "feature" but it's so frustrating that it was not taken account in the original design that I mark it as an error reporting bug.

Windows

MikaelMayer:
Fixes #3281 

Example code:
```dafny
predicate P(i: int) {
  i <= 0
}

predicate Q(i: int, j: int) {
  i == j || -i == j
}

function method Toast(i: int): int
  requires P(i)

method Test(i: int) returns (j: nat)
  ensures Q(i, j)
{
  if i < 0 {
    return -i;
  } else {
    return Toast(i);
  }
}
```

| Before: | After |
| --------|-------|
| Hovering Q(i,j) image | image |
| Hovering the second return image | image |
| Hovering the call to Toast() image | image |
| The resolver error appears after all obsolete verification error<br>image | image |

I added language server tests and tested it in VSCode, and it works as expected.

### DIFF
--- a/Source/DafnyCore/AST/DafnyAst.cs
+++ b/Source/DafnyCore/AST/DafnyAst.cs
@@ -116,9 +116,8 @@ namespace Microsoft.Dafny {
               UpdateStartEndToken(node.GetEndToken());
             } else {
               UpdateStartEndToken(node.tok);
+              node.Children.Iter(UpdateStartEndTokRecursive);
             }
-
-            node.Children.Iter(UpdateStartEndTokRecursive);
           }
 
           UpdateStartEndTokRecursive(this);
--- a/Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
+++ b/Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
@@ -2,7 +2,9 @@
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using JetBrains.Annotations;
 using Microsoft.Dafny.LanguageServer.IntegrationTest.Extensions;
+using Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization;
 using Microsoft.Dafny.LanguageServer.IntegrationTest.Util;
 using Microsoft.Dafny.LanguageServer.Workspace.Notifications;
 using Microsoft.Extensions.Configuration;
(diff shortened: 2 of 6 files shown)


MikaelMayer:
Found bugs if hovering over the "return" in this program:

```
module ProblemModule {
  datatype X =
    | Cons(head: int, tail: X)
    | Nil
  {
    predicate method Valid() {
      this.Cons? && tail.Valid()
    }
  }
}

method Test() returns (j: int)
  ensures j == 1
{
  return 2;
}
```

keyboardDrummer (approved):

MikaelMayer:
Disabling auto-merge because the following simple example has garbage hovering message:

```dafny
datatype Test = Test(i: int)
{
  predicate Valid() {
    i > 0
  }
  predicate CanAct() requires Valid() {
    i > 1
  }
  method Tester(other: Test) {
    assert other.Valid();
    assert other.CanAct();
  }
}
```
If you hover over `other.Valid();`, you get
image


and if you hover over `other.CanAct()`, you get
image
image
Clearly, some ranges are wrong.

MikaelMayer:
One more failing test case

```dafny

datatype ValidTester = Tester(next: ValidTester2) | Tester2(next: ValidTester2) | Test3(next: ValidTester2)
{
  predicate Valid() {
    ((this.Tester? || this.Tester2?) && this.next.Valid()) || (this.Test3? && !this.next.Valid())
  }

  function method apply(): int requires Valid() {
    2
  }
  static method Test(c: ValidTester) {
    var x := c.apply();
  }
}

datatype ValidTester2 = MoreTest(i: int, next: ValidTester2) | End {
  predicate Valid(defaultValue: int := 0) {
    0 <= defaultValue &&
    (this.End? || (this.MoreTest? && this.next.Valid() && i > 0))
  }
}
```

If you hover over the failing c.apply(), it will quote random things. I guess it's because there is a defaultValue whose token is outside of the regular range.

keyboardDrummer (approved):

PR #3284, opened 2022-12-28
refactor: Clean up resolution passes 0 and 1

### LINKED ISSUE #3273: Equality support inferred for nullable iterator but not non-null iterator
### Dafny version

3.10.0

The iterator declaration

``` dafny
iterator Iter<Y>(u: set<Y>) // this infers Y to be (==)
```

gives rise to two types, the nullable type `Iter?` and the non-null type `Iter`. From the signature of the iterator declaration, Dafny infers the `(==)` characteristic for type parameter `Y` (since the signature includes the type `set<Y>`).

Therefore, if `D` is a type that does _not_ support equality, then neither the type `Iter?<D>` nor the type `Iter<D>`. Alas, Dafny reports an error only for `Iter?<D>`, not for `Iter<D>`.

Diagnosis: While Dafny infers `(==)` for `Y` for the type `Iter?`, it does not apply the same learning to type `Iter`.

Repro:

``` dafny
iterator Iter<Y>(u: set<Y>) // this infers Y to be (==)

method Repro<K>() {
  var iter: Iter<K>; // should be an error, because K is not (==)
}
```

```
$ dafny test.dfy

```

### What happened?

This is what is expected for the 8 cases of inferred/declared `(==)`, passing in equality-support type or not, and using the nullable/non-null version of the iterator type:

``` dafny
iterator IterA<X(==)>() // this declares X to be (==)

iterator IterB<Y>(u: set<Y>) // this infers Y to be (==)

method Test<K, L(==)>() {
  var a0: IterA<K>; // error: K is not (==)
  var a1: IterA<L>;
  var b0: IterB<K>; // error: K is not (==)
  var b1: IterB<L>;

  var c0: IterA?<K>; // error: K is not (==)
  var c1: IterA?<L>;
  var d0: IterB?<K>; // error: K is not (==)
  var d1: IterB?<L>;
}
```

The issue does not appear to affect soundness. The following example is close, but no cigar:

``` dafny
iterator EqIter<Y>(u: set<Y>, x: Y, y: Y) yields (eq: bool)
  yield ensures eq == (x == y)
  ensures false
{
  while true {
    yield x == y;
  }
}

datatype Dt = Dt(a: int, ghost b: int)

method Main() {
  var d0 := Dt(0, 0);
  var d1 := Dt(0, 1);
  assert d0 != d1; // clearly

  // The following ought to be an error, because EqIter should take a (==) type argument.
  var iter: EqIter<Dt>;
  // The following does give an error, since the constructor lives in class EqIter whose
  // type parameer has been inferred to be (==).
  iter := new EqIter({}, d0, d1); // error:
  var more := iter.MoveNext();
  assert more;
  assert iter.eq == (d0 == d1) == false; // according to the yield-ensures clause
  print iter.eq, "\n";
  if iter.eq {
    var ugh := 1 / 0; // according to the verifier, we never get here
  }
}
```

Mac

RustanLeino:
This PR is the final clean-up before I can start to merge in the new type inference. It cleans up the AST traversals that occur during "pass 0" and "pass 1" of the resolver. Previously, it wasn't so clear what exactly these passes did, but now it's more clear:

* Pass 0
    - resolve names and infer types
    - check that type inference did not leave any types underspecified
    - perform substitution for DefaultValueExpression, compute .ResolvedOp, and some similar checks and AST updates as necessary to make the AST coherent
* Pass 1
    - discover bounds
    - compute call graph
    - compute and check ghost things
    - determine tail recursion
* Passes 2 and 3
    - many more passes through the AST to perform checks and compute of some additional information

Each "pass" really consists of several traversals of the AST. So, "phase" or something like that would be a better name.

With this PR, the dependencies between pass 0 and pass 1 have been sorted out, and so have the dependencies among the various things that are happening in pass 1.

Pass 0 is the most delicate one, because it starts with an AST about which very little is known. The general direction of the work in this PR is to try to make later traversals be as independent of each other as possible.

Previously, there were many special cases that have added up over time. This PR tries to use the AST Visitor to get a more uniform treatment of things in the AST. There were several places where top-level declarations include, for example, a `class`, whereas its automatically declared non-null (subset type) declaration was accessible only via the class. And there were, and still are, several places where by-method declarations or prefix predicates/lemmas are not available from the standard list of `.Members` but is instead accessible only via the the function or lemma declaration that created these. The PR improves on this situation, but only as needed to disentangle passes 0 and 1. There's plenty of room for further beautification and simplification, especially in passes 2 and 3, but that's out of scope for this PR.

This PR does not try to improve the tasks and traversals performed by "pass 2" and "pass 3". Indeed, it's not clear what the distinction between 2 and 3 is, but I think it's just historical that various things fell into one of these two.

A few awkward dependencies remain between the passes. The one that comes to mind is that the body of a prefix predicate/lemma cannot be filled until the call-graph has been built. For this reason, `CheckTypeInference` is run on these prefix bodies once they have been constructed. It's possible that this could be improved by making sure that the prefix bodies are created without a need to call `CheckTypeInference` on them at all.

This PR also fixes some minor things:
* pretty-printing of `newtype` declarations
* scope and two-state checks for attributes
* inferred equality support for non-null iterators (issue #3273)

Fixes #3273

Breaking changes: This PR impacted the test suite primarily in these ways:
* The call graph printed by `/rprint` may show things in a different order than before and may show some vertices without edges that were not shown before. I have tried to mitigate this (for now and for the future) by applying further sorting criteria when printing out the call graph information.
* Resolution error messages may be ordered differently before. This is because the new traversals are done in a more uniform way.
* Some resolution errors may shadow other resolution errors. This is because some checks have moved from one pass to another.
Users who depend on the exact nature of the resolution errors or on `/rprint` may be affected by this PR.


Source/DafnyCore/AST/AstVisitor.cs
```diff
@@ -260,18 +262,26 @@ public abstract class ASTVisitor<VisitorContext> where VisitorContext : IASTVisi
 
         // Visit subexpressions
         expr.SubExpressions.Iter(ee => VisitExpression(ee, context));
+
+        // Finish by calling the post-visitor method
```
keyboardDrummer, line 266: Comment seems obsolete given the name `PostVisitOneExpression`

Source/DafnyCore/AST/AstVisitor.cs
```diff
@@ -317,17 +327,25 @@ public abstract class ASTVisitor<VisitorContext> where VisitorContext : IASTVisi
 
         // Visit substatements
         stmt.SubStatements.Iter(ss => VisitStatement(ss, context));
+
+        // Finish by calling the post-visitor method
```
keyboardDrummer, line 331: Comment seems redundant given the name `PostVisitOneExpression`

Source/DafnyCore/ExpressionTester.cs
```diff
@@ -230,33 +230,30 @@ public class ExpressionTester {
       return isCompilable;
     } else if (expr is LambdaExpr lambdaExpr) {
       return CheckIsCompilable(lambdaExpr.Body, codeContext);
-    } else if (expr is ComprehensionExpr comprehensionExpr) {
-      var uncompilableBoundVars = comprehensionExpr.UncompilableBoundVars();
+    } else if (expr is ComprehensionExpr) {
+      var e = (ComprehensionExpr)expr;
```
keyboardDrummer, line 234: What's the advantage of the pattern

```c#
else if (expr is ComprehensionExpr) {
  var e = (ComprehensionExpr)expr;
```

over

```c#
if (expr is ComprehensionExpr comprehensionExpr) 
```

?

Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
```diff
@@ -21,6 +21,120 @@
...
+        Contract.Assert(currentClass == null);
+
+        ResolveNamesAndInferTypesForOneDeclaration(topd, initialRound);
+
+        Contract.Assert(AllTypeConstraints.Count == 0);
+        Contract.Assert(currentClass == null);
+      }
+    }
+
+    void ResolveNamesAndInferTypesForOneDeclaration(TopLevelDecl topd, bool initialRound) {
```
keyboardDrummer, line 41: Seems like two methods would be better than one method with the `initialRound` parameter. The shared code, which seems to be very littles, can be put into methods.

Source/DafnyCore/Resolver/Resolver.cs
```diff
@@ -2670,270 +2675,104 @@ public class ModuleBindings {
...
-            var dd = (TopLevelDeclWithMembers)d;
-            ResolveClassMemberBodies(dd);
-          }
-        }
-        allTypeParameters.PopMarker();
+        var boundsDiscoveryVisitor = new BoundsDiscoveryVisitor(reporter);
+        boundsDiscoveryVisitor.VisitDeclarations(declarations);
       }
 
+      // Build call graph.
```
keyboardDrummer, line 2721: Comment seems redundant based on the code `CallGraphBuilder.Build(declarations, reporter);`

keyboardDrummer (approved):

PR #3287, opened 2022-12-28
Revert "Fix debugging (#3211)" except the record => class

MikaelMayer:
This PR reverts #3211 
* except the change that transforms the record Token into a class, because that is necessary for avoiding ToString recursion in record (apparently, I really thought dotnet would have been more clever and I did not encounter the issue but @keyboardDrummer  did)
* and adds attributes `[DebuggerBrowsable(DebuggerBrowsableState.Never)]` to avoid surcharging the debugger.

The problem with #3211 are that they unnecessarily break client code, and that RangeToken, StartToken, EndToken and OwnedTokens are better viewed as (lazily computed) properties than methods by definition of C#, and their computation do not take more time than previously thought anyway.

@keyboardDrummer  could you please test it and verify it works for you adequatedly?


Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -27,6 +27,7 @@ public class FilledInDuringResolutionAttribute : System.Attribute { }
   public abstract class INode {
 
     public IToken tok = Token.NoToken;
+    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
```
keyboardDrummer, line 30: Why do we need this attribute?

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -77,108 +78,114 @@ public abstract class INode {
     // TODO: Re-add format tokens where needed until we put all the formatting to replace the tok of every expression
     internal IToken[] FormatTokens = null;
 
-    public RangeToken RangeToken {
-      set => rangeToken = value;
-    }
+    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
```
keyboardDrummer, line 81: Why do we need this attribute?

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -77,108 +78,114 @@ public abstract class INode {
...
-            if (token1.pos < startTok.pos) {
-              startTok = token1;
+              if (token1.pos + token1.val.Length > endTok.pos + endTok.val.Length) {
+                endTok = token1;
+              }
             }
 
-            if (token1.pos + token1.val.Length > endTok.pos + endTok.val.Length) {
-              endTok = token1;
+            void updateStartEndTokRecursive(INode node) {
```
keyboardDrummer, line 105: Could you leave in the uppercase rename?

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -77,108 +78,114 @@ public abstract class INode {
...
+        return rangeToken;
       }
-
-      return rangeToken;
+      set => rangeToken = value;
     }
 
-    public IToken GetStartToken() => GetRangeToken()?.StartToken;
-    public IToken GetEndToken() => GetRangeToken()?.EndToken;
+    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
```
keyboardDrummer, line 144: Why do we need this attribute?

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -77,108 +78,114 @@ public abstract class INode {
...
+    public IToken StartToken => RangeToken?.StartToken;
+
+    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
```
keyboardDrummer, line 147: Why do we need this attribute?

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -77,108 +78,114 @@ public abstract class INode {
...
     /// <summary>
     /// A token is owned by a node if it was used to parse this node,
     /// but is not owned by any of this Node's children
     /// </summary>
-    public IEnumerable<IToken> GetOwnedTokens() {
-      if (OwnedTokensCache != null) {
-        return OwnedTokensCache;
-      }
+    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
+    public IEnumerable<IToken> OwnedTokens {
```
keyboardDrummer, line 157: I don't think `OwnedTokens` should be a property since it does a computation. I would like to follow the MS guidelines that I linked to before, and not have any properties that are orders of magnitude slower than a field access.

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -77,108 +78,114 @@ public abstract class INode {
     // TODO: Re-add format tokens where needed until we put all the formatting to replace the tok of every expression
     internal IToken[] FormatTokens = null;
 
-    public RangeToken RangeToken {
-      set => rangeToken = value;
-    }
+    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
+    public virtual RangeToken RangeToken {
```
keyboardDrummer, line 82: I think this should only be a property if we plan to remove the computation. Could you link from here to an issue that describes the change you have in mind?

Source/DafnyCore/AST/DafnyAst.cs
MikaelMayer: Good question, indeed it's not relevant to the problem to be solved directly.
I had read when learning about this attribute that it can be used to avoid displaying duplicate field and wrapping properties (such as tok and Tok) in the debugger. I think it is clever, so I'm enabling it. One might counter-argue that they will look for "Tok" and not think of looking at "tok", if that ever happens, we can revert that. What do you think?

Source/DafnyCore/AST/DafnyAst.cs
MikaelMayer: Right, that's exactly the plan that, following your advice, I described here:
https://github.com/dafny-lang/dafny/issues/3290

Source/DafnyCore/AST/DafnyAst.cs
MikaelMayer: Because there is a field `rangeToken` so `RangeToken` is a duplicate normally.

Source/DafnyCore/AST/DafnyAst.cs
MikaelMayer: For now, to avoid recomputing StartToken / RangeToken in the debugger, which you said could cause problems.

Source/DafnyCore/AST/DafnyAst.cs
MikaelMayer: For now, to avoid recomputing StartToken / RangeToken in the debugger view, which you said could cause problems.

Source/DafnyCore/AST/DafnyAst.cs
MikaelMayer: `OwnedTokens` is as fast as a field access after the first time, and the code called only once does not rely on I/O or asynchronous programming, and its total execution (over all the nodes) is linear in terms of the length of the code.

I sometimes call them "lazy properties" because I'm used in Scala and in Scala this is a standard practice.
In C#, we call such properties "Cached Properties", and it's acceptable that they are doing computation the first time, e.g. I don't see a comment in any of the answers and threads there that complains that any of these properties should be a method instead of a property because it does computation the first time https://stackoverflow.com/questions/2579329/cached-property-easier-way

Note that we could make it even more concise using techniques such as described here
https://stackoverflow.com/a/63085705/1287856

keyboardDrummer: > I don't see a comment in any of the answers and threads there that complains that any of these properties should be a method instead of a property because it does computation the first time

This one does: https://stackoverflow.com/a/2579369/93197

I'm OK with having cached computation properties. I think it's safer to use methods for anything that might trigger a computation, so the programmer is made aware of that, but I guess if you use a cached property in a hot loop, the caching will limit the amount of computation done and the hotness of the loop is irrelevant.

Source/DafnyCore/AST/DafnyAst.cs
keyboardDrummer: Can't we make `tok` private? Then it doesn't show up in the standard list.

Source/DafnyCore/AST/DafnyAst.cs
keyboardDrummer: When I had trouble with it the computation of Token.ToString() was still broken, so let's try not adding this attribute.

Source/DafnyCore/AST/DafnyAst.cs
keyboardDrummer: When I had trouble with it the computation of Token.ToString() was still broken, so let's try not adding this attribute.

Source/DafnyCore/AST/DafnyAst.cs
keyboardDrummer: `rangeToken` is not shown in the debugger because it's non-public, so I think we can leave this attribute out.

Source/DafnyCore/AST/DafnyAst.cs
MikaelMayer: I don't know how you configured your debugger, but private fields do show up my debugger. I would be very annoyed if they weren't, after all, what's the point of debugging if you can't figure out the internal state of the object?

Source/DafnyCore/AST/DafnyAst.cs
MikaelMayer: > This one does: https://stackoverflow.com/a/2579369/93197

Good finding. It's a least upvoted answer, but it's an answer.

> Who's we?

I and those C# developers who are not bothered by a property having an extra computation the first time :-)

> I think we should use C# source generators ([link](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview)) to write a [Cached] attribute that you can slap on any property (or method even), to make it cached.

We can do that if we really feel it's needed. Or use the more idiomatic Lazy<>

keyboardDrummer (approved):

Source/DafnyCore/AST/DafnyAst.cs
keyboardDrummer: In my debugger they're hidden under a separate menu, so they don't create noise unless you explicitly look for them.

PR #3299, opened 2022-12-30
Implementing errorIDs

### LINKED ISSUE #3182: Consider associating unique IDs with errors
### Summary

Should Dafny errors have unique IDs? And what form and internal API is needed to support them?

Somewhat related: are dafny error messages ever going to be localized?

### Background and Motivation

Some motivations:
- A unique error ID can be shown to the user and associated with documentation about the problem; it might also provide helpful information in a problem report. Con: Dafny error messages themselves are unique strings or string templates.
- A unique ID (not necessarily exposed to the user) could be used internally. For example the IDE, upon encountering an error with ID, could automatically link to information that provided more context for the user.

Note that some error messages are templates -- that is strings with holes that are filled in differently in different contexts. For example: Error: a _decl_ cannot be declared 'ghost'.
So there is not a unique string literal for each possible message

Implementation questions:
- is the error code an int or a string? Edit: I see that LSP error codes can be ints or strings. And to clarify here: this issue is concerned with diagnostics related to the Dafny source code, not communication errors between the LS and its client.
- is it simply embedded in the error message in a structured way or do we modify all the error propagation methods in the dafny tool to propagate the error code in parallel with the error message  (like location information is)?

### Proposed Feature

No proposal yet.

### Alternatives

_No response_

davidcok:
Fixes #3182

Implementing an error ID in order to enable showing the user more detailed error explanations and quick fixes.


Source/DafnyCore/Coco/Parser.frame
```diff
@@ -178,20 +191,52 @@ public class Errors {
...
+    var y = new Token(x.line,x.col);
+    y.pos = x.pos;
+    y.kind = x.kind;
+    y.Filename = x.Filename;
+    y.val = len <= x.val.Length ? x.val.Substring(0, len) : x.val.PadRight(len, ' ');
+    return new CodeActionToken(t, y);
+  }
+
+  // Nested range is up to, not including, the next token
+  public static IToken NestCodeAction(IToken t, IToken next) {
```
MikaelMayer, line 222:
```suggestion
  public static IToken NestCodeAction(IToken t, IToken next = null) {
    next = next ?? t.Next;
```
That way, when you know that the second token is the next, you don't need to repeat the token in the arguments.

Source/DafnyCore/Coco/Parser.frame
```diff
@@ -178,20 +191,52 @@ public class Errors {
...
+    return new NestedToken(t, NewCodeAction(t, next.pos - t.pos));
+  }
+
+  public static IToken NestCodeAction(IToken t, int len) {
+    return new NestedToken(t, NewCodeAction(t, len));
+  }
+
+  public static IToken FindCodeActionRange(IToken toptok) {
```
MikaelMayer, line 230:
```suggestion
  public static CodeActionToken FindCodeActionRange(IToken toptok) {
```

Source/DafnyCore/Dafny.atg
```diff
@@ -94,13 +94,13 @@ void CheckDeclModifiers(ref DeclModifierData dmod, string declCaption, AllowedDe
 {
   declCaption = (declCaption.StartsWith("i") || declCaption.StartsWith("o") ? "an " : "a ") + declCaption;
   if (dmod.IsAbstract && ((allowed & AllowedDeclModifiers.Abstract) == 0)) {
-    SemErr(dmod.AbstractToken, $"{declCaption} cannot be declared 'abstract'");
+    SemErr(ParserErrorDetail.p_abstract_not_allowed, Errors.NestCodeAction(dmod.AbstractToken, dmod.AbstractToken.Next), $"{declCaption} cannot be declared 'abstract'");
```
MikaelMayer, line 97:
```suggestion
    SemErr(ParserErrorDetail.p_abstract_not_allowed, Errors.NestCodeAction(dmod.AbstractToken), $"{declCaption} cannot be declared 'abstract'");
```
If you implement my change, you can remove the second argument

Source/DafnyCore/DafnyCore.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <Target Name="RunCoco" BeforeTargets="PreBuildEvent" Outputs="$(ProjectDir)Parser.cs;$(ProjectDir)Scanner.cs" Inputs="$(ProjectDir)Dafny.atg">
+  <Target Name="RunCoco" BeforeTargets="PreBuildEvent" Outputs="$(ProjectDir)Parser.cs;$(ProjectDir)Scanner.cs" Inputs="$(ProjectDir)Dafny.atg;$(ProjectDir)/CoCo/Parser.frame;$(ProjectDir)/CoCo/Scanner.frame">
```
MikaelMayer, line 3: Ahhhhh that's why it was not reconstructing when I was changing Scanner.frame ! Can you please also add Parser.frame in inputs and Parser.cs in outputs?

Source/DafnyCore/ErrorDetail.cs
```diff
@@ -0,0 +1,38 @@
...
+  // Adds information into the dictionary
+  public static void Add(string errorID, string detail) {
+    errorDetailMap.Add(errorID, detail);
+  }
+
+  // Returns the error detail for the given error ID;
+  // result could be null if there is no such detail
+#nullable enable
+  public static string? GetDetail(string errorID) {
+    return errorDetailMap.ContainsKey(errorID) ? errorDetailMap[errorID] : null;
```
MikaelMayer, line 33: To avoid two hash calls:
```suggestion
    if(errorDetailMap.TryGetValue(errorID, out var result)) {
      return result;
    }
    return null;
```

Source/DafnyCore/ParserErrorDetail.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using System;
+using System.Numerics;
+using System.Globalization;
+
+namespace Microsoft.Dafny;
+
+public class ParserErrorDetail {
+
+  public static string p_bad_const_initialize_op = "p_bad_const_initialize_op";
```
MikaelMayer, line 12: To avoid having to repeat the name in the strong, you can consider using an enum
https://stackoverflow.com/a/71090888/1287856

Source/DafnyCore/Reporting.cs
```diff
@@ -43,11 +44,26 @@ public abstract class ErrorReporter {
...
         if (!include.ErrorReported) {
-          Message(source, ErrorLevel.Error, include.tok, "the included file " + tok.Filename + " contains error(s)");
+          Message(source, ErrorLevel.Error, "", include.tok, "the included file " + tok.Filename + " contains error(s)");
           include.ErrorReported = true;
         }
       }
-      Message(source, ErrorLevel.Error, tok, msg);
+      Message(source, ErrorLevel.Error, "", tok, msg);
+    }
+    public void Error(MessageSource source, string errorID, IToken tok, string msg) {
```
MikaelMayer, line 53: Can the method on line 38 just call this method with errorID = "" ?

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -71,18 +71,59 @@ public record DafnyCodeActionWithId(DafnyCodeAction DafnyCodeAction, int Id);
...
+    }
+  }
+
   private DafnyCodeActionProvider[] GetDafnyCodeActionProviders() {
     return new List<DafnyCodeActionProvider>() {
       new VerificationDafnyCodeActionProvider()
-    }.Concat(
+    }
+    .Concat(new List<DafnyCodeActionProvider> { new ErrorMessageCodeActionProvider() })
+    .Concat(
```
MikaelMayer, line 126:
```suggestion
    , new ErrorMessageCodeActionProvider()
    }
    .Concat(
```

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
```diff
@@ -71,18 +71,59 @@ public record DafnyCodeActionWithId(DafnyCodeAction DafnyCodeAction, int Id);
...
+        Kind = CodeActionKind.QuickFix
       };
       return t;
     }
     ).ToArray();
     return new CommandOrCodeActionContainer(codeActions);
   }
 
+
+  public class ErrorMessageCodeActionProvider : DiagnosticDafnyCodeActionProvider {
```
MikaelMayer, line 84: Can you move this class to a new file in the same folder as the class `VerificationDafnyCodeActionProvider`? I think it's the language/ folder
Also, please consider migrating all of the lone class DafnyCodeActions inside this handler.

MikaelMayer (commented):
This is great ! It will bring a lot of value, especially because we will be able to use ALT+F8 to trigger quick fixes.

Source/DafnyCore/Coco/Parser.frame
davidcok: No. If there is just one token or no CodeActionTokens found, then the top token is returned, which is not necessarily a CodeActionToken.

Source/DafnyCore/DafnyCore.csproj
davidcok: They are there already. But yes, I was very pleased when I found this was the only change needed fo this problem. The have been may times when I altered a .frame file and wondered why nothing changed in the program behavior.

Source/DafnyCore/Reporting.cs
davidcok: Done. In fact the method on line 38 will eventually go away. I just have not cleaned up all the various methods that eventually report a Message until I was sure of the overall structure.

Source/DafnyLanguageServer/Handlers/DafnyCodeActionHandler.cs
MikaelMayer: This as well.

.github/workflows/xunit-tests.yml
```diff
@@ -65,6 +65,7 @@ jobs:
       run: dotnet build -warnaserror --no-restore ${{env.solutionPath}}
     - name: Run DafnyLanguageServer Tests
       run: |
+        ## Run twice to catch unstable code (Issue #2673)
```
MikaelMayer, line 68: Hmm that leaves me with an aftertaste. Suppose checking in an unstable tests crash 1/4 of the time (3/4 chances of success). After testing twice, we wlll have a success rate of 1/2
If someone then commits the tests when the two runs succeed, confident that a crash would have been caught from the two consecutive runs,
every other subsequent PR will now have 1/2 chances of failing, more than one fourth like the original.
So this is a two-sided coin. By raising the bar to entry, if anything goes wrong randomly, we will affect every one new PR out there.
I'm just saying it out loud as a motivation to claim that
1) We should run the tests only once.
2) If a test fails on any PR, unrelated to that PR, it should be immediately reported as an issue and deactivated so that future PR don't have the same issue that would otherwise accumulate.

Source/DafnyCore/Reporting.cs
```diff
@@ -242,7 +236,7 @@ public class ConsoleErrorReporter : BatchErrorReporter {
           errorLine += $" {msg} {TokenToString(tok)}";
         }
 
-        if (DafnyOptions.O.CompileVerbose) {
+        if (DafnyOptions.O.CompileVerbose && false) { // Need to control tests better before we enable this
```
MikaelMayer, line 239: Did you intend to keep this in production? looks like a leftover 🍕

Source/DafnyLanguageServer/Language/ErrorMessageDafnyCodeActionProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+    if (data is null) {
+      return def;
+    }
+    try {
+      String s = data.ToString();
+      int k1 = s.IndexOf(" ");
+      int k2 = s.IndexOf(" ", k1 + 1);
+      int line = Int32.Parse(s.Substring(0, k1));
+      int column = Int32.Parse(s.Substring(k1 + 1, k2 - k1 - 1));
+      int length = Int32.Parse(s.Substring(k2 + 1));
```
MikaelMayer, line 23:
```suggestion
      var nums = s.split(" ");
      var line = Int32.Parse(s[0]);
      var column = Int32.Parse(s[1]);
      var length = Int32.Parse(s[2]);
```
might be more maintainable.

Also, I know it's the first version so no need to take action now, but have you considered that Ranges there could span multiple lines? It's probably not useful in the parser errors, but for the resolver errors it might be, and for the verifier errors it's surely.

Test/git-issues/git-issue-3187a.dfy.expect
```diff
@@ -1,4 +1,4 @@
-git-issue-3187a.dfy(5,12): Error: a const field should be initialized using ':=', not '='
-git-issue-3187a.dfy(7,12): Error: a local variable should be initialized using ':=', ':-', or ':|', not '='
-git-issue-3187a.dfy(8,22): Error: a variable in a let expression should be initialized using ':=', ':-', or ':|', not '='
+git-issue-3187a.dfy(5,10): Error: a const field should be initialized using ':=', not '='
```
MikaelMayer, line 1: That error is so much more accurate, thanks !

docs/check-examples
```diff
@@ -118,7 +118,7 @@ do
   ## For header lines, extract the error message from
   ## the header text (everything but the initial ##[ ]* )
   ( echo "$line" | grep -E -e '[#][#] ' > /dev/null ) \
-   && msg=`echo "$line" | sed -E -e 's/^##[ ]*[*]*//' -e 's/[*]*$//' -e 's/\\*/\\\\*/' -e 's/_[a-z]+_/.*/g'`
+   && msg=`echo "$line" | sed -E -e 's/ {#.*}//' | sed -E -e 's/^##[ ]*[*]*//' -e 's/[*]*$//' -e 's/\\*/\\\\*/' -e 's/_[a-z]+_/.*/g'`
```
MikaelMayer, line 121: Can you explain me what this change does? At least I would love if there was a change in the comment that reflects what the change on this obscure regex is ^^

MikaelMayer (changes requested):
Minor requests, overall, looks good to merge next time ! This is the start of the next great IDE feature for Dafny I believe.

.github/workflows/xunit-tests.yml
davidcok: Up to you (and Remy). I just put in a comment to explain why the line is replicated.

Source/DafnyCore/Reporting.cs
davidcok: I wanted to put out error detail when --verbose was true. But many tests have --verbose (equivalently /compileVerbose) true. But you are correct that I forgot about this.

Source/DafnyLanguageServer/Language/ErrorMessageDafnyCodeActionProvider.cs
davidcok: Yes - and might consist of multiple segments. But I thought I'd implement that when I came to actual cases that needed it.

Test/git-issues/git-issue-3187a.dfy.expect
davidcok: Thanks

docs/check-examples
davidcok: I added a comment in the script. It converts the error message in the header into a regex to match against the eror message produced by running dafny against the example, to be sure the example does what I say.

Source/DafnyCore/Reporting.cs
davidcok: Took out the block for now.

MikaelMayer (approved):
Final approval, great job, let's have it !

PR #3311, opened 2023-01-03
Def assignment update

### LINKED ISSUE #3092: Always allow havoc assignments unless enforcing determinism in compiled code
Dafny supports havoc assignment: http://dafny.org/dafny/DafnyRef/DafnyRef#2118-havoc-right-hand-side

Currently certain definite assignment checks will prevent using this assignment. Those checks should be relaxed so that the assignment is allowed, unless Dafny has been configured to enforce determinism and the assignment is in compiled code.

Also, consider marking ` --relax-definite-assignment` as deprecated.

Much more context is available in the discussion in this PR: https://github.com/dafny-lang/dafny/pull/3052

### LINKED ISSUE #3232: Overly strict behavior of definite assignment
### Dafny version

3.10.0+dev

```dafny
method Enqueue(n: nat)
  {
    var a := new int[n];
    forall i | 0 <= i < n {
      a[i] := i;
    }
  }
```

```code
daffy verify test.dfy
```

### What happened?

/Users/davidcok/projects/dafny/dafny/docs/DafnyRef/test.dfy(3,13): Error: unless an initializer is provided for the array elements, a new array of 'int' must have empty size

It appears that the default setting of definite assignment requires array elements to be set on allocation even if they aren't used. This is a real nuisance and requires either globally using --relax-definite-assignment or putting in an unneeded initializer, which is unnecessary computation.

Mac

keyboardDrummer:
### Changes

- In new CLI UI, turn off array initialisation def assignment check unless --enforce-determinism is used. Fixes #3232
-  Allow havoc assignment to initialise a variable, but only if it has a non-empty type. Fixes #3092

@RustanLeino should havocing a variable of a non-empty type even be allowed?

### Testing

- Havoc assignment to a field count as an assignment
- Havoc to a local variable counts as an assignment
- Arrays created with new CLI UI and without `--enforce-determinism`, don't need to initialise their values or be empty.


Test/dafny0/DefiniteAssignment.dfy
```diff
@@ -77,12 +82,12 @@ method Main()
...
+method DontForgetHavoc<G(00)>(a: G, h: int) returns (k: G) {
   var x:G, y:G, z:G := a, *, a;
   if h < 10 {
     k := x;  // fine
   } else if h < 20 {
-    k := y;  // error: y has not yet been assigned
+    k := y;
   } else {
     z := *;
     return z;  // this is fine, since z was assigned before the havoc
```
davidcok, line 93: Why does 'z was assigned before the havoc' matter? It doesn't for 'y', right?

davidcok (changes requested):

Test/dafny0/DefiniteAssignment.dfy
keyboardDrummer: I think this comment can go

davidcok (approved):

RustanLeino:
> @RustanLeino should havocing a variable of a non-empty type even be allowed?

(I'm assuming you meant "possibly empty", not "non-empty".)

Since we allow `x :| true;` where the type of `x` is possible empty, it makes sense to also allow `x := *;`.

keyboardDrummer:
> > @RustanLeino should havocing a variable of a non-empty type even be allowed?
> 
> (I'm assuming you meant "possibly empty", not "non-empty".)
> 
> Since we allow `x :| true;` where the type of `x` is possible empty, it makes sense to also allow `x := *;`.

I see :-)

kjx:
Don't you effectively define `x := *;`  as `x :| true;`.

In fact: given you have  `x :| true;` why do you need `x := *;` as well?

PR #3312, opened 2023-01-03
Non null type null check

keyboardDrummer:
Requires https://github.com/dafny-lang/dafny/pull/3311 to be merged to let tests pass

### Changes
Do not report null dereference errors on not-nullable types

### Testing

Dereference non-nullable types that were assigned using havoc

### DIFF
--- a/Test/comp/NestedArrays.dfy
+++ b/Test/comp/NestedArrays.dfy
@@ -1,4 +1,4 @@
-// RUN: %baredafny verify --relax-definite-assignment %args "%s" > "%t"
+// RUN: %baredafny verify %args "%s" > "%t"
 // RUN: %baredafny run --no-verify --target=cs %args "%s" >> "%t"
 // RUN: %baredafny run --no-verify --target=js %args  "%s" >> "%t"
 // RUN: %baredafny run --no-verify --target=go %args  "%s" >> "%t"
@@ -10,11 +10,7 @@
  * fixed, Java is omitted from this file.
  *
  * Also, note, these tests mostly check that auto-initialized variables of array
- * types get proper values. Without --relax-definite-assignment (which is
- * our new, planned default), there will still be some way to ask the compiler
- * to generate these values (a plan is to change ":= *" to be evidence for a
- * definite assignments). At that time, the functionality tested in this
- * file will still be needed in the compilers.
+ * types get proper values.
  */
 
 method Main() {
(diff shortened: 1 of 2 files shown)


Test/comp/NestedArrays.dfy
```diff
@@ -28,31 +24,31 @@ datatype GenericSingletonRecord<X> = GenericSingletonRecord(x: X)
 class MyClass { }
 
 method TestDefaultArrayValues<X>() {
-  var arr: array<array<int>>;
-  var srr: array<SingletonRecord>;
-  var grr: array<GenericSingletonRecord<array<int>>>;
-  var xrr: array<array<X>>;
+  var arr: array<array<int>> := *;
```
MikaelMayer, line 27: Do you have a test about what happens if someone does not use ":= *"? If not, I'd suggesting another test file to capture that behavior.

Test/comp/NestedArrays.dfy
```diff
@@ -10,11 +10,7 @@
  * fixed, Java is omitted from this file.
  *
  * Also, note, these tests mostly check that auto-initialized variables of array
- * types get proper values. Without --relax-definite-assignment (which is
- * our new, planned default), there will still be some way to ask the compiler
- * to generate these values (a plan is to change ":= *" to be evidence for a
- * definite assignments). At that time, the functionality tested in this
- * file will still be needed in the compilers.
+ * types get proper values.
```
MikaelMayer, line 13: I can't believe we have no documentation for `:= *`
Otherwise, I would have asked you to add something to the doc about that. If not you, @davidcok , would you be able to add documentation on `:= *`?

MikaelMayer (changes requested):
Minor requests, otherwise thanks to deal with this issue !

Test/comp/NestedArrays.dfy
keyboardDrummer: There's still a lot of .expect files that contain "variable/field X, which is subject to definite-assignment rules, might be uninitialized here", which is what you get if you don't use the havoc.

Test/comp/NestedArrays.dfy
keyboardDrummer: Good point!

davidcok: There is (at least now) documentation on havoc assignments. It presumes that any new values satisfies whatever non-null or subset constraints the type of the LHS has.

Test/VSComp2010/Problem1-SumMax.dfy.expect
```diff
@@ -1,2 +1,2 @@
 
-Dafny program verifier finished with 3 verified, 0 errors
```
MikaelMayer, line 2: I think you have a separate PR for that change right? And I would like to discuss it as a team because I would rather prefer to have imprecise positive feedback (which for example indicates that it's not vacuously verified because it was the wrong file) rather than no positive feedback at all.

keyboardDrummer: I agree that the feedback is important. I'm putting this PR on hold until I find a way to get precise positive feedback.

PR #3315, opened 2023-01-04
Adding a test case that checks the error for two-character char literals

davidcok:
Adds a missing test case as described in the title.


robin-aws:
Thanks for adding the test case I thought I missed. Turns out I just searched in the wrong repo when trying to find it though. :) It's already here, so no need for this PR after all: https://github.com/dafny-lang/dafny/blob/master/Test/dafny0/CharLiteralParseErrors.dfy

PR #3316, opened 2023-01-04
Improve the style of dafny.org/dafny

davidcok:


docs/index.html
```diff
@@ -23,7 +23,7 @@
 <body>
 <header>
 <h1>The Dafny Programming and Verification Language</h1>
-<!-- VERSION INFO HERE-->Development version
+<!-- VERSION INFO HERE--><center>Development version</center>
```
fabiomadge, line 26: I think you [should](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center) use `text-align` here.

fabiomadge (changes requested):
Thanks!

docs/index.html
davidcok: Done

fabiomadge (approved):

PR #3324, opened 2023-01-04
Fix: Wording of assertion failure to match what happens

### LINKED ISSUE #3216: Improve assertion failure wording
### Dafny version

3.10.0

```dafny
assert 1 == 2;
```

```code
Open in VSCode.
```

### What happened?

I've done some formal verification of SystemVerilog using SVA and Synopsys's VC_Static tool. With that tool you write properties, click "Prove" and then wait. The possible states of a property are:

* Proven
* Vacuously proven (the whole property is always true but only because the antecedent of an implication can never happen)
* Falsified
* Inconclusive (not enough time was allowed to prove or falsify it).

A counter-example is only available in the Falsified case.

This all makes intuitive sense to me.

Dafny doesn't seem to do it in the same way. Proven is the same, but all other three states result in the error `assertion might not hold` which by any reasonable interpretation means it has been falsified.

Even worse, you can press F7 to generate a counter-example even for the inconclusive case where the example actually satisfies the assertion!

Could Dafny not distinguish Falsified and Inconclusive? If not I think the error should at least be reworded to `could not prove assertion` which I think is what it is actually trying to say. (Similarly for postconditions etc.)

For the counter-examples it should be trivial to determine if the generated counter-example passes or fails the assertion and give a message like `this example disproves the assertion` or `this example passes the assertion but Dafny cannot reason about it` (or whatever a passing counter-example is supposed to mean).

Hope this isn't too negative. My overall experience of Dafny so far has been very positive, mainly because of the excellent VSCode extension and amazing documentation.

Windows

MikaelMayer:
This PR fixes #3216
It completely replaces the wording ".... might not hold" by "could not prove..."

There are many reasons for that: We already use the wording "could not prove..." in other places and, funny enough, we (I) previously justified the introduction of "might not hold" by

> - fix: The error "assertion violation" is replaced by the better wording "assertion might not hold". This indicates better that the verifier is unable to prove the assertion.

If we replace "might not hold" by "could not prove", then "might not hold" does not requires explanation anymore.
I got also several complaints from the wording that, although softer than the previous "assertion violation", still indicates a point where Dafny is unsure what to do.

Moreover, the implication of "assertion might not hold" places the doubt on the programmer, which can be frustrating: If the programmer wants to find Dafny's counter-examples, he would be upset to discover they are incomplete, and that their assertion is actually valid. Saying "Could not prove" shifts the doubt from the programmer to Dafny by finally recognizing that the programmer might hit a limitation of Dafny and perhaps just need to introduce things to help Dafny prove its result. It also emphasizes the need to "prove" rather than the "make the assertion hold".

Here are the messages that were replaced: (before/after)
```
This postcondition might not hold on a return path
Could not prove this postcondition on a return path

A postcondition might not hold on this return path
Could not prove a postcondition on this return path

the calculation step between the previous line and this line might not hold
Could not prove that the calculation step between the previous line and this line hold

This loop invariant might not hold on entry
Could not prove this loop invariant on entry

A precondition for this call might not hold
Could not prove a precondition for this call

This is the precondition that might not hold
This is the precondition that could not be proven

Error: assertion might not hold
Error: Could not prove assertion

Error: function precondition might not hold
Error: Could not prove function precondition
```

I also changed the UserGuide.md to reflect this wording, section verification debugging.


Source/DafnyCore/Verifier/ProofObligationDescription.cs
```diff
@@ -297,7 +297,7 @@ public class AssertStatement : ProofObligationDescription {
       : $"error is impossible: {customErrMsg}";
 
   public override string FailureDescription =>
-    customErrMsg ?? "assertion might not hold";
+    customErrMsg ?? "Could not prove assertion";
```
atomb, line 300:
```suggestion
    customErrMsg ?? "could not prove assertion";
```

The convention in the rest of the messages is for them to start with a lower-case letter.

Source/DafnyCore/Verifier/ProofObligationDescription.cs
```diff
@@ -279,7 +279,7 @@ public class PreconditionSatisfied : ProofObligationDescription {
       : $"error is impossible: {customErrMsg}";
 
   public override string FailureDescription =>
-    customErrMsg ?? "function precondition might not hold";
+    customErrMsg ?? "Could not prove function precondition";
```
atomb, line 282:
```suggestion
    customErrMsg ?? "could not prove function precondition";
```

The convention in the rest of the messages is for them to start with a lower-case letter.

docs/DafnyRef/UserGuide.md
```diff
@@ -464,11 +464,12 @@ Most output from `dafny` is directed to the standard output of the shell invokin
...
+In this section, we suggest a methodology to figure out [why Dafny cannot prove a single assertion](#sec-verification-debugging), we propose techniques to deal with [assertions that slow a proof down](#sec-verification-debugging-slow), we explain how to [verify assertions in parallel or in a focused way](#sec-assertion-batches), and we also give some more examples of [useful options and attributes to control verification](#sec-command-line-options-and-attributes-for-verification).
 
 ### 25.6.1. Verification debugging when verification fails {#sec-verification-debugging}
 
-Let's assume one assertion is failing ("assertion might not hold" or "postcondition might not hold"). What should you do next?
+Let's assume one assertion is failing ("Could not prove assertion" or "Could not prove postcondition"). What should you do next?
```
atomb, line 471: Same case distinction here.

docs/HowToFAQ/ERROR_SeqComp.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Error: function precondition might not hold"
+title: "Error: Cannot prove function precondition"
```
atomb, line 2: Ditto.

docs/HowToFAQ/onepage.md
```diff
@@ -2213,7 +2213,7 @@ To declare `formula` as _contravariant_ use `formula<-T>`. Then `formula<U>` is
 
 Type parameter characteristics are discussed in [the reference manual](../DafnyRef/DafnyRef.html#sec-type-parameter-variance)
 
-# "Error: function precondition might not hold"
+# "Error: Cannot prove function precondition"
```
atomb, line 2216: And here.

atomb (changes requested):

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -7233,7 +7233,7 @@ public ForceCheckToken(IToken tok)
       Contract.Ensures(Contract.Result<Bpl.Ensures>() != null);
 
       Bpl.Ensures ens = new Bpl.Ensures(ForceCheckToken.Unwrap(tok), free, condition, comment);
-      ens.Description = new PODesc.AssertStatement(errorMessage ?? "This is the postcondition that could not be proven.");
+      ens.Description = new PODesc.AssertStatement(errorMessage ?? "this is the postcondition that could not be proven.");
```
davidcok, line 7236: If you are going with the lower case style, then remove the ending periods as well

davidcok (changes requested):

Source/DafnyCore/Verifier/Translator.cs
atomb: It is usually in a full sentence, though, of the form "Error: here is something in lowercase."

davidcok: Yes, but the style in dafny elsewhere is predominantly to start with lower-case and not have a period, except when there are clearly multiple sentences. I'm not wedded to this style, and am open to a team discussion, but I'd like it to be consistent.

atomb: Ah, good point. I'm fine with leaving out the periods.

atomb (dismissed):
Looks good!

MikaelMayer:
Requires 
https://github.com/boogie-org/boogie/pull/676
to be merged

Source/IntegrationTests/TestFiles/LitTests/LitTest/proof-obligation-desc/modifiable.dfy.expect
```diff
@@ -1,14 +1,14 @@
...
+modifiable.dfy(10,4): Error: modified field could not be proved to be in the current modifies clause
  Asserted expression: this in {}
-modifiable.dfy(33,7): Error: assignment might update an object not in the enclosing context's modifies clause
+modifiable.dfy(33,7): Error: modified field could not be proved to be in the current modifies clause
  Asserted expression: s[0] == s[1] || s[0] == s[1] as C || s[0] in {s[1]} || s[0] in {s[1] as C}
-modifiable.dfy(42,6): Error: assignment might update an object field not in the enclosing context's modifies clause
+modifiable.dfy(42,6): Error: modified field could not be proved to be in the current modifies clause
  Asserted expression: c == s[1]
-modifiable.dfy(50,5): Error: assignment might update an array element not in the enclosing context's modifies clause
+modifiable.dfy(50,5): Error: modified field could not be proved to be in the current modifies clause
```
atomb: The new message talks about a field, but the location being updated is an array element, as the old description correctly stated.

atomb (changes requested):
Your `review.sh` script from #6318 was very helpful!

It pointed out one inconsistency in the error messages, exhibited by `modified.dfy`, which I commented on below.

It also pointed out some slightly out-of-scope changes related to command-line handling in `docs/DafnyRef/UserGuide.md`, but I think they all seem reasonable and I don't object to including them.

atomb (dismissed):

atomb (approved):

PR #3330, opened 2023-01-05
Fix: Errors on refined modules displayed in refined modules, not abstracts ones

### LINKED ISSUE #3329: Module refinement breaks outline, error reporting and gutter icons
There is strange behavior with the gutter icons when verifying refining an abstract modules.

```dafny
abstract module Test {
  method Encrypt()
    ensures false
}

module Ok refines Test {
  function test(): int {
    2
  }
  // Might not work as expected
  method Encrypt() {
  }
}
```
See the result here
image
* The error on the refined module is expressed in the abstract module, which completely breaks not only the gutter icons, but also the outline (there is no method Encrypt in the refined module)

We can use this behavior to pretend we prove false, at least on the gutter icons:
image

It's even worse when the abstract module is included instead of being in the same file, because its errors are reported on the file containing the refined module, which is out of offset. Look at the postcondition error underlined.... in the comment.
image

MikaelMayer:
This PR fixes #3329

I had to fix several things for that. Here were the problems I fixed:
* Refined declarations were using the same token as their abstract counterpart. This prevented gutter errors to be recorded at all for these declarations.
* Refined tokens' filenames are the filename plus the [moduleName], which interfered with the ability of the gutter icons rendered to consider that refining modules were actually part of the file.

### DIFF
--- a/Source/DafnyCore/Rewriters/RefinementTransformer.cs
+++ b/Source/DafnyCore/Rewriters/RefinementTransformer.cs
@@ -596,51 +596,76 @@ namespace Microsoft.Dafny {
       }
     }
 
-    Method CloneMethod(Method m, List<AttributedExpression> moreEnsures, Specification<Expression> decreases, BlockStmt newBody, bool checkPreviousPostconditions, Attributes moreAttributes) {
+    Method CloneMethod(Method m, Method newDeclaration, Specification<Expression> decreases, BlockStmt newBody, bool checkPreviousPostconditions, Attributes moreAttributes) {
       Contract.Requires(m != null);
       Contract.Requires(!(m is Constructor) || newBody == null || newBody is DividedBlockStmt);
       Contract.Requires(decreases != null);
+      List<AttributedExpression> moreEnsures = newDeclaration.Ens;
+      IToken newMethodTok = new NestedToken(newDeclaration.tok, m.tok, "it refines the following abstract declaration");
+      var result = refinementCloner.WithNewDeclarationTokenWrapping(newDeclaration.tok, () => {
+        var tps = m.TypeArgs.ConvertAll(refinementCloner.CloneTypeParam);
+        var ins = m.Ins.ConvertAll(p => refinementCloner.CloneFormal(p, false));
+        var req = m.Req.ConvertAll(refinementCloner.CloneAttributedExpr);
+        var mod = refinementCloner.CloneSpecFrameExpr(m.Mod);
 
-      var tps = m.TypeArgs.ConvertAll(refinementCloner.CloneTypeParam);
-      var ins = m.Ins.ConvertAll(p => refinementCloner.CloneFormal(p, false));
-      var req = m.Req.ConvertAll(refinementCloner.CloneAttributedExpr);
-      var mod = refinementCloner.CloneSpecFrameExpr(m.Mod);
+        var ens = refinementCloner.WithRefinementTokenWrapping(
+          () => m.Ens.ConvertAll(refinementCloner.CloneAttributedExpr), !checkPreviousPostconditions);
 
-      var ens = refinementCloner.WithRefinementTokenWrapping(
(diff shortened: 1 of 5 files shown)


Source/DafnyCore/Rewriters/RefinementTransformer.cs
```diff
@@ -596,51 +596,76 @@ public RefinementTransformer(Program p)
       }
     }
 
-    Method CloneMethod(Method m, List<AttributedExpression> moreEnsures, Specification<Expression> decreases, BlockStmt newBody, bool checkPreviousPostconditions, Attributes moreAttributes) {
+    Method CloneMethod(Method m, Method newDeclaration, Specification<Expression> decreases, BlockStmt newBody, bool checkPreviousPostconditions, Attributes moreAttributes) {
       Contract.Requires(m != null);
       Contract.Requires(!(m is Constructor) || newBody == null || newBody is DividedBlockStmt);
       Contract.Requires(decreases != null);
+      List<AttributedExpression> moreEnsures = newDeclaration.Ens;
+      IToken newMethodTok = new NestedToken(newDeclaration.tok, m.tok, "it refines the following abstract declaration");
```
MikaelMayer, line 604: Here I just indented the code and wrapped it with the call `refinementCloner.WithNewDeclarationTokenWrapping`
and I changed the use of "tok" with the `newMethodTok` above

Source/DafnyCore/Rewriters/RefinementTransformer.cs
```diff
@@ -881,7 +906,7 @@ public RefinementTransformer(Program p)
                   replacementBody = MergeBlockStmt(replacementBody, prevMethod.BodyForRefinement);
                 }
               }
-              var newM = CloneMethod(prevMethod, m.Ens, decreases, replacementBody, prevMethod.BodyForRefinement == null, m.Attributes);
+              var newM = CloneMethod(prevMethod, m, decreases, replacementBody, prevMethod.BodyForRefinement == null, m.Attributes);
```
MikaelMayer, line 909: Here I needed the entire new method, not just its ensures clause, so that I recover correctly the method's token.

Source/DafnyLanguageServer.Test/GutterStatus/LinearVerificationGutterStatusTester.cs
```diff
@@ -114,7 +114,8 @@ public abstract class LinearVerificationGutterStatusTester : ClientBasedLanguage
 
   protected async Task<List<LineVerificationStatus[]>> GetAllLineVerificationStatuses(
       TextDocumentItem documentItem,
-      TestNotificationReceiver<VerificationStatusGutter> verificationStatusGutterReceiver
+      TestNotificationReceiver<VerificationStatusGutter> verificationStatusGutterReceiver,
+      bool intermediates = true
```
MikaelMayer, line 118: This mechanism was introduced in the yet unmerged PR
https://github.com/dafny-lang/dafny/pull/3116
and enables only testing the final gutter icons, not the intermediates, so that it's both easier and supports concurrent verification.

Source/DafnyLanguageServer.Test/GutterStatus/SimpleLinearVerificationGutterStatusTester.cs
```diff
@@ -14,6 +14,24 @@ public class SimpleLinearVerificationGutterStatusTester : LinearVerificationGutt
...
+ | :}
+ | :
+ | :module Ok refines Test {
+ | :  function test(): int {
+ | :    2
+ | :  }
+ | :  // Might not work as expected
+[=]:  method Encrypt() {
+[ ]:  }
+   :}", intermediates: false);
```
MikaelMayer, line 32: Previously, this test was entirely green, and only fixing the refinement token would put the postcondition error on line 22, which should not be the case. The real error is now a nested one.

PR #3331, opened 2023-01-05
Fix: Other CLI files that are not found do not trigger exceptions

### LINKED ISSUE #2719: Passing a DLL path that doesn't exist raises an exception
`dafny foobar.dfy` gives you a relatively clean `Unable to open included file` error, but `dafny foobar.dll` leads to a `FileNotFoundException` (because it tries to extract the `DafnySourceAttribute` text if it exists, without handling this case).

```dafny
method Test() {}
```

MikaelMayer:
This PR fixes #2719

### DIFF
--- a/Source/DafnyDriver/DafnyDriver.cs
+++ b/Source/DafnyDriver/DafnyDriver.cs
@@ -259,7 +259,13 @@ namespace Microsoft.Dafny {
 
         var supportedExtensions = options.Compiler.SupportedExtensions;
         if (supportedExtensions.Contains(extension)) {
-          otherFiles.Add(file);
+          if (File.Exists(file)) {
+            otherFiles.Add(file);
+          } else {
+            options.Printer.ErrorWriteLine(Console.Out,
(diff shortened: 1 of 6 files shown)


Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -259,7 +259,14 @@ public enum CommandLineArgumentsResult {
 
         var supportedExtensions = options.Compiler.SupportedExtensions;
         if (supportedExtensions.Contains(extension)) {
-          otherFiles.Add(file);
+          // .h files are not part of the build, they are just emitted as includes
+          if (File.Exists(file) || extension == ".h") {
+            otherFiles.Add(file);
+          } else {
+            options.Printer.ErrorWriteLine(Console.Out,
+              "*** Error: file " + file + " not found");
```
fabiomadge:
```suggestion
            options.Printer.ErrorWriteLine(Console.Out, $"*** Error: file {file} not found");
```

fabiomadge (approved):

PR #3334, opened 2023-01-06
Add test case and release notes for 3333 fix

### LINKED ISSUE #3333: Match expression type inference is incorrect
### Dafny version

3.10.0

```dafny
newtype uint16 = x: int | 0 <= x < 0x1_0000
datatype Datatype = A | B
{
  function method Foo(): uint16 {
    match this
    case A => 1
    case B => 2
  }
}
```

```code
dafny verify filename.dfy

```

### What happened?

I expect to get an error, since newtypes cannot be automatically converted to their basetype.

```
filename.dfy(5,18): Error: Function body type mismatch (expected uint16, got int)
1 resolution/type errors detected in filename.dfy
```

Mac

keyboardDrummer:
Fixes #3333

The actual fix was already accidentally made in https://github.com/dafny-lang/dafny/pull/2734, this PR is just adding a test case and a release note, and removing some obsolete code.

### DIFF
--- a/Source/DafnyCore/AST/Expressions/NestedMatchExpr.cs
+++ b/Source/DafnyCore/AST/Expressions/NestedMatchExpr.cs
@@ -67,12 +67,6 @@ public class NestedMatchExpr : Expression, ICloneable<NestedMatchExpr> {
       return;
     }
 
-    var dtd = sourceType.AsDatatype;
-    var subst = new Dictionary<TypeParameter, Type>();
-    if (dtd != null) {
-      subst = TypeParameter.SubstitutionMap(dtd.TypeArgs, sourceType.TypeArgs);
-    }
-
     Type = new InferredTypeProxy();
     foreach (var kase in Cases) {
       resolver.scope.PushMarker();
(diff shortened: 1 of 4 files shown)


Test/git-issues/git-issue-3333.dfy.expect
```diff
@@ -0,0 +1,2 @@
+git-issue-3333.dfy(8,18): Error: Function body type mismatch (expected uint16, got int)
```
MikaelMayer, line 1: At some point it will be good to actually say in this case "To declare a constant integer as a uint16, write "1 as uint16x"", but I guess it's out of this scope.

Source/DafnyCore/AST/Expressions/NestedMatchExpr.cs
```diff
@@ -67,12 +67,6 @@ public class NestedMatchExpr : Expression, ICloneable<NestedMatchExpr> {
       return;
     }
 
-    var dtd = sourceType.AsDatatype;
```
MikaelMayer, line 70: I guess this is dead code, right?

MikaelMayer (approved):
Thanks for adding a test.

Source/DafnyCore/AST/Expressions/NestedMatchExpr.cs
keyboardDrummer: Jep

PR #3336, opened 2023-01-06
Fix: Modifies clauses with set operations can now be proven

### LINKED ISSUE #3265: Can't prove modifies clause
### Dafny version

3.10.0

```dafny
datatype K = K(ghost MainSet: set<object>, Element: object)

datatype KUp = KUp(ghost MainSet: set<object>, Element: object, k: K) {
  method DoIt()
    modifies MainSet - {Element}
  {
    var p := MainSet - {Element};
    var c := k.MainSet - {k.Element};
    assume c <= p;
    assert forall o: object | o in c ::
             o in p;  // Verified
    assert c == k.MainSet - {k.Element}; // Verified
    modify k.MainSet - {k.Element}; // Modify error
  }
}
```

```code
Just open this file in VSCode.
```

### What happened?

The modify statement fails, although if you replace `k.MainSet - {k.Element}` by `c`; it suddenly verifies. But these two expressions are equal !
The workaround thus is to use `c` but this error is disturbing, as I found no way for Dafny to verify the modify statement without having a single variable.

Windows

MikaelMayer:
This PR fixes #3265

### DIFF
--- a/Source/DafnyCore/Triggers/TriggersCollector.cs
+++ b/Source/DafnyCore/Triggers/TriggersCollector.cs
@@ -384,13 +384,44 @@ namespace Microsoft.Dafny.Triggers {
 
         if (!children_contain_killers) {
           // Add only if the children are not killers; the head has been cleaned up into non-killer form
-          collected_terms.Add(new_term);
+          collected_terms.AddRange(CollectInSetOperations(new_expr, expr));
         }
       }
       Contract.Assert(new_term != null);  // this checks our assumption that "new_exprs" contains at least one value.
 
       // This new node is a killer if its children were killers, or if it's non-cleaned-up head is a killer
       return new TriggerAnnotation(children_contain_killers || expr_is_killer, new_term.Variables, collected_terms);
+    }
+
+    private IEnumerable<TriggerTerm> CollectInSetOperations(Expression new_expr, Expression original_expr) {
+      yield return new TriggerTerm { Expr = new_expr, OriginalExpr = original_expr, Variables = CollectVariables(original_expr) };
+      if (new_expr is BinaryExpr
+        {
+          ResolvedOp: BinaryExpr.ResolvedOpcode.InSet,
+          E0: var e0, E1: BinaryExpr
+          {
+            ResolvedOp: BinaryExpr.ResolvedOpcode.SetDifference or
+                          BinaryExpr.ResolvedOpcode.Union or
+                          BinaryExpr.ResolvedOpcode.Intersection,
+            E0: var e10,
+            E1: var e11
+          }
+        }) {
+        var newTriggerTerm1 = new BinaryExpr(new_expr.tok, BinaryExpr.Opcode.In, e0, e10) {
+          ResolvedOp = BinaryExpr.ResolvedOpcode.InSet,
+          Type = Type.Bool
+        };
+        foreach (var triggerTerm in CollectInSetOperations(newTriggerTerm1, original_expr)) {
+          yield return triggerTerm;
+        }
+        var newTriggerTerm2 = new BinaryExpr(new_expr.tok, BinaryExpr.Opcode.In, e0, e11) {
+          ResolvedOp = BinaryExpr.ResolvedOpcode.InSet,
+          Type = Type.Bool
+        };
+        foreach (var triggerTerm in CollectInSetOperations(newTriggerTerm2, original_expr)) {
+          yield return triggerTerm;
+        }
+      }
     }
 
     private TriggerAnnotation AnnotateApplySuffix(ApplySuffix expr) {
--- a/Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
+++ b/Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
@@ -1772,6 +1772,7 @@ BplBoundVar(varNameGen.FreshId(string.Format("#{0}#", bv.Name)), predef.BoxType,
         bool pr;
         if (s is BinaryExpr && aggressive) {
           BinaryExpr bin = (BinaryExpr)s;
+          /*
           switch (bin.ResolvedOp) {
             case BinaryExpr.ResolvedOpcode.Union:
               return Boogie.Expr.Or(TrInSet_Aux(tok, elmt, elmtBox, bin.E0, aggressive, out pr), TrInSet_Aux(tok, elmt, elmtBox, bin.E1, aggressive, out pr));
@@ -1781,7 +1782,7 @@ BplBoundVar(varNameGen.FreshId(string.Format("#{0}#", bv.Name)), predef.BoxType,
               return Boogie.Expr.And(TrInSet_Aux(tok, elmt, elmtBox, bin.E0, aggressive, out pr), Boogie.Expr.Not(TrInSet_Aux(tok, elmt, elmtBox, bin.E1, aggressive, out pr)));
             default:
               break;
-          }
+          }*/
         } else if (s is SetDisplayExpr) {
           SetDisplayExpr disp = (SetDisplayExpr)s;
           Boogie.Expr disjunction = null;
@@ -1895,9 +1896,6 @@ BplBoundVar(varNameGen.FreshId(string.Format("#{0}#", bv.Name)), predef.BoxType,
         if (s is BinaryExpr && aggressive) {
           BinaryExpr bin = (BinaryExpr)s;
           switch (bin.ResolvedOp) {
-            case BinaryExpr.ResolvedOpcode.Union:
-            case BinaryExpr.ResolvedOpcode.Intersection:
-            case BinaryExpr.ResolvedOpcode.SetDifference:
             case BinaryExpr.ResolvedOpcode.MultiSetUnion:
             case BinaryExpr.ResolvedOpcode.MultiSetIntersection:
               return true;
--- a/Test/git-issues/git-issue-3265.dfy
+++ b/Test/git-issues/git-issue-3265.dfy
@@ -1,0 +1,3 @@
+// RUN: %baredafny verify %args_0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
--- a/Test/git-issues/git-issue-3265.dfy
+++ b/Test/git-issues/git-issue-3265.dfy
@@ -1,3 +1,103 @@
-// RUN: %baredafny verify %args_0 "%s" > "%t"
+// RUN: %baredafny verify %args "%s" > "%t"
 // RUN: %diff "%s.expect" "%t"
 
+
+method SetDifferenceTrigger(
+    ghost MainSet: set<object>, Elements: set<object>,
+    ghost MainSet2: set<object>, Elements2: set<object>)
+  modifies MainSet - Elements
+{
+  assume forall o: object | o in MainSet2 - Elements2 ::
+            o in MainSet - Elements;  // Verified
+  forall o <- MainSet2 ensures true {
+    if o !in Elements2 {
+      assert o in MainSet;
+      assert o !in Elements;
+    }
+  }
+}
+
+
+method SetUnionTrigger(
+    ghost MainSet: set<object>, Elements: set<object>,
+    ghost MainSet2: set<object>, Elements2: set<object>)
+  modifies MainSet - Elements
+{
+  var p := MainSet - Elements;
+  var c := MainSet2 - Elements2;
+  assume forall o: object | o in MainSet2 + Elements2 ::
+            o in MainSet + Elements;  // Verified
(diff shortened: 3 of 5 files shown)


Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
```diff
@@ -1772,6 +1772,7 @@ public ExpressionTranslator(ExpressionTranslator etran, string modifiesFrame)
         bool pr;
         if (s is BinaryExpr && aggressive) {
           BinaryExpr bin = (BinaryExpr)s;
+          /*
```
MikaelMayer, line 1775: Is it ok to remove this code?

Test/git-issues/git-issue-3265.dfy.expect
```diff
@@ -0,0 +1,2 @@
+
+Dafny program verifier finished with 7 verified, 0 errors
```
MikaelMayer, line 2: Previously all tests were failing.

Test/git-issues/git-issue-3265.dfy
```diff
@@ -0,0 +1,103 @@
...
+    assert o in MainSet || o in Elements;
+  }
+}
+
+method SetIntersectionTrigger(
+    ghost MainSet: set<object>, Elements: set<object>,
+    ghost MainSet2: set<object>, Elements2: set<object>)
+  modifies MainSet - Elements
+{
+  assume forall o: object | o in (MainSet2 * Elements2) ::
```
RustanLeino, line 40: Parentheses not needed

Test/git-issues/git-issue-3265.dfy
```diff
@@ -0,0 +1,103 @@
...
+
+
+method SetDifferenceTrigger(
+    ghost MainSet: set<object>, Elements: set<object>,
+    ghost MainSet2: set<object>, Elements2: set<object>)
+  modifies MainSet - Elements
+{
+  assume forall o: object | o in MainSet2 - Elements2 ::
+            o in MainSet - Elements;  // Verified
+  forall o <- MainSet2 ensures true {
```
RustanLeino, line 12: I suggest formatting `forall` statements (here and below) like

``` dafny
forall o <- MainSet2
  ensures true
{
```

Test/git-issues/git-issue-3265.dfy
```diff
@@ -0,0 +1,103 @@
...
+    ghost MainSet2: set<object>, Element2: object)
+  modifies MainSet - {Element}
+{
+  var p := MainSet - {Element};
+  var c := MainSet2 - {Element2};
+  assume c <= p;
+  assert forall o: object | o in c ::
+            o in p;  // Verified
+  assert c == MainSet2 - {Element2}; // Verified
+  modify MainSet2 - {Element2}; // Modify error
```
RustanLeino, line 58: Change the comment, since this is no longer an error. You could drop it altogether, or (probably better) you could say "this used to give a modify error".

Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
```diff
@@ -1899,9 +1900,6 @@ public ExpressionTranslator(ExpressionTranslator etran, string modifiesFrame)
         if (s is BinaryExpr && aggressive) {
           BinaryExpr bin = (BinaryExpr)s;
           switch (bin.ResolvedOp) {
-            case BinaryExpr.ResolvedOpcode.Union:
-            case BinaryExpr.ResolvedOpcode.Intersection:
-            case BinaryExpr.ResolvedOpcode.SetDifference:
             case BinaryExpr.ResolvedOpcode.MultiSetUnion:
             case BinaryExpr.ResolvedOpcode.MultiSetIntersection:
```
RustanLeino, line 1904: I think it would be good to do the analogous change also for multisets. You could do that in this same PR, in a separate PR, or, if nothing else, file an Issue that reminds us that this would be a good change to make in the future.

Source/DafnyCore/Triggers/TriggersCollector.cs
```diff
@@ -390,6 +390,37 @@ internal class TriggersCollector {
...
+        foreach (var triggerTerm in CollectInSetOperations(newTriggerTerm1, original_expr)) {
+          yield return triggerTerm;
+        }
+        var newTriggerTerm2 = new BinaryExpr(new_expr.tok, BinaryExpr.Opcode.In, e0, e11) {
+          ResolvedOp = BinaryExpr.ResolvedOpcode.InSet,
+          Type = Type.Bool
+        };
+        foreach (var triggerTerm in CollectInSetOperations(newTriggerTerm2, original_expr)) {
+          yield return triggerTerm;
+        }
```
RustanLeino, line 420: What happens for some input like

``` dafny
function F(x: int): set<int>

method Test(x: int)
  requires forall x: int, y: int, z: int :: x in F(y) * F(z)
{
}
```

Previously, this would generate `{:trigger x in F(y) * F(z)}`. Will this now additionally generate

    {:trigger x in F(y)}
    {:trigger x in F(z)}

If so, then note that these are not valid triggers, since neither one by itself covers all bound variables.

Source/DafnyCore/Triggers/TriggersCollector.cs
```diff
@@ -390,6 +390,37 @@ internal class TriggersCollector {
       return new TriggerAnnotation(children_contain_killers || expr_is_killer, new_term.Variables, collected_terms);
     }
 
+    private IEnumerable<TriggerTerm> CollectInSetOperations(Expression new_expr, Expression original_expr) {
+      yield return new TriggerTerm { Expr = new_expr, OriginalExpr = original_expr, Variables = CollectVariables(original_expr) };
```
RustanLeino, line 394: I don't understand the role of `OriginalExpr`, but I note that the previous L384  added just `new_expr`, not a `TriggerTerm` wrapped around `new_expr`. Is that as intended?

RustanLeino (changes requested):

Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
RustanLeino: Yes, unless the test suite shows us otherwise.
When you do remove it, please also remove the unused variable `pr` on L1772.

Source/DafnyCore/Triggers/TriggersCollector.cs
MikaelMayer: Indeed, it now generates these additional triggers. Let me have a look.

Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
MikaelMayer: I'll do this in this PR if it's not too hard

MikaelMayer: > I think it would be good to do the analogous change also for multisets

Done ! That wasn't too hard. The only difference was that the trigger expressions were of the form (A + B)[x] and so this is what I split into two new triggers A[x] and B[x] (if the bound variables are the same of course!)

Source/DafnyCore/Triggers/TriggersCollector.cs
MikaelMayer: > If so, then note that these are not valid triggers, since neither one by itself covers all bound variables.

Fixed by checking that the number of bound variables remains the same. Good catch !

Source/DafnyCore/Triggers/TriggersCollector.cs
MikaelMayer: > I note that the previous L384 added just new_expr, not a TriggerTerm wrapped around new_expr. Is that as intended?

I'm not sure of what you see but here is the diff I see on my side.


```diff
         if (!children_contain_killers) {
           // Add only if the children are not killers; the head has been cleaned up into non-killer form
-          collected_terms.Add(new_term);
+          collected_terms.AddRange(CollectInSetOperations(new_expr, expr));
         }
```

So it looks like we were previously adding the same thing that I add now, except that now I generate more triggers. Could you elaborate a bit more?

Source/DafnyCore/Triggers/TriggersCollector.cs
```diff
@@ -390,6 +390,73 @@ internal class TriggersCollector {
...
+            BinaryExpr.ResolvedOpcode.MultiSetIntersection or
+            BinaryExpr.ResolvedOpcode.MultiSetDifference,
+            E0: var seq0,
+            E1: var seq1
+          },
+          E0: var seqArg
+        } newSeqSelectExpr) {
+        // Multiset triggers are of the form expression[index]
+        var newTriggerTerm1 = new SeqSelectExpr(newExpr.tok, true, seq0, seqArg, null, newSeqSelectExpr.CloseParen) {
+          Type = Type.Int
```
RustanLeino, line 417: Shouldn't this type be `seq0.Type`?

Source/DafnyCore/Triggers/TriggersCollector.cs
```diff
@@ -390,6 +390,73 @@ internal class TriggersCollector {
...
+        };
+        foreach (var triggerTerm in CollectInSetOperations(newTriggerTerm1, originalExpr, newBoundVariables)) {
+          yield return triggerTerm;
+        }
+        var newTriggerTerm2 = new SeqSelectExpr(newExpr.tok, true, seq1, seqArg, null, newSeqSelectExpr.CloseParen) {
+          Type = Type.Int
```
RustanLeino, line 423: Shouldn't this type be `seq1.Type`?

Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
```diff
@@ -1915,8 +1905,8 @@ public ExpressionTranslator(ExpressionTranslator etran, string modifiesFrame)
             case BinaryExpr.ResolvedOpcode.Union:
             case BinaryExpr.ResolvedOpcode.Intersection:
             case BinaryExpr.ResolvedOpcode.SetDifference:
-            case BinaryExpr.ResolvedOpcode.MultiSetUnion:
-            case BinaryExpr.ResolvedOpcode.MultiSetIntersection:
+              //case BinaryExpr.ResolvedOpcode.MultiSetUnion:
+              //case BinaryExpr.ResolvedOpcode.MultiSetIntersection:
```
RustanLeino, line 1909: If these are no longer needed, delete the lines. But why are the multiset operations excluded and the set operations still included?

RustanLeino (changes requested):

Source/DafnyCore/Triggers/TriggersCollector.cs
MikaelMayer: Good catch, it should actually be `seq1.Type.AsMultiSetType.Arg`

Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
MikaelMayer: The set operations are still included because there is a flag that can be turned on for sets (aggressive) which is used for frame conditions.
Without these, the tests compiling the Dafny runtime failed, see here
https://github.com/dafny-lang/dafny/actions/runs/4734312552/jobs/8403056164
I found that by reverting this code to its original form, the dafny runtime would verify again. Do you have an idea why?

Source/DafnyCore/Triggers/TriggersCollector.cs
RustanLeino: My point was that, previously, `collected_terms.Add(new_term);` adds `new_term`, where the new `collected_terms.AddRange(CollectInSetOperations(new_expr, expr));` adds

* nothing, or
* `new TriggerTerm { Expr = newExpr, OriginalExpr = originalExpr, Variables = CollectVariables(newExpr) }` (which is the thing that has a `TriggerTerm` wrapped around the `newExpr`), or
* that `new TriggerTerm...` plus additional triggers.

This is probably fine. I'm just clarifying what I had tried to comment on.

Source/DafnyCore/Verifier/Translator.ExpressionTranslator.cs
RustanLeino: I was unable to reproduce the failure. However, I think it is correct to _omit_ the multiset operations here, because `RewriteInExpr` is supposed to be in synch with `TrInMultiSet_Aux`, which only rewrites `MultiSetDisplayExpr`.

RustanLeino (dismissed):
Good work!

MikaelMayer:
Variability tests show that this PR does not incur any meaningful difference, so I'm setting up auto-merge.

Issue #3290, opened 2022-12-29
Refactoring needed: Replace "Tok" with a format


MikaelMayer:
# Context

Currently, every AST node stores a "Tok", which is used to report errors.
We recently introduced "RangeToken" on nodes so that we would get also the first and last token of every node, and that RangeToken is assigned in the parser

Some nodes produced by the resolver are actually clones of existing nodes, or entirely new nodes, both of which only reuse "Tok" and not "RangeToken"
This is why, for these nodes, there remains an old computation for RangeToken which tries to infer the Range from sub-nodes's Tok, format tokens, and range tokens:
https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/AST/DafnyAst.cs#L89

# Problems:

* The manual computation of RangeToken is not reliable. RangeTokens should be computed on the pre-resolved AST (which they are in Parser.cs), but the manual computation usually is executed post-resolution on resolved nodes, which sometimes creates weird ranges that heuristics can't rule out. For example, there are instances where the computed range takes into account tokens that should be marked as implicit and which are not, which makes unreasonably large ranges https://github.com/dafny-lang/dafny/pull/3282#issuecomment-1367066779
* We recently introduced "OwnedTokens", which should entirely replace "FormatTokens", but because of this old computation, we have still to keep "FormatTokens". `OwnedTokens` is never called except for the formatter. which works on pre-resolution phase.
* Moreover, because of this old computation, there is a debate on whether "RangeToken" should be made a method instead of a property. This debate shouldn't exist, because this old computation should never need to be executed normally.
* We can't however go in every place where tokens are transferred from one node to a new node and also transfer the RangeToken, it would be extremely hard to find all such cases, but also to maintain.

# Suggested solution

Since we use "Tok" in nodes only for reporting, I suggest we do the following:
* We replace "Tok" in every constructor by a new "Format" field, stored in `INode`
* We remove the old computation altogether, so that `RangeToken` can only be assigned (in the Parser), no debate on whether it should be a property or a method.
* `class Format` contains `Tok`, the `RangeToken`, its quick shortcuts `StartToken` and `EndToken`, but not the `OwnedTokens` that is lazily computed on demand and requires access to the children of the node. `OwnedTokens` stays in `INode`. However, its cache `OwnedTokensCache` can surely be stored in the `Format`
* We remove the field `FormatTokens` on expressions and all the arguments that were used to depend on it. 
* For compatibility:
  *  `Tok => Format.Tok`
[...]

PR #3338, opened 2023-01-07
Make a clone of the parsed document available in compilation

keyboardDrummer:


Source/DafnyCore/AST/TopLevelDeclarations.cs
```diff
@@ -938,8 +938,8 @@ public class ModuleDefinition : INode, IDeclarationOrUsage, INamedRegion, IAttri
   public override IEnumerable<INode> Children => (Attributes != null ? new List<INode> { Attributes } : Enumerable.Empty<INode>()).Concat(TopLevelDecls);
 }
 
-public class DefaultModuleDecl : ModuleDefinition {
-  public DefaultModuleDecl()
+public class DefaultModuleDef : ModuleDefinition {
```
MikaelMayer, line 941: Ahhhhhh I was wondering why I was always confused by module declarations and module definitions.... you nailed it !

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -71,30 +71,29 @@ INotificationPublisher notificationPublisher
...
-      if (errorReporter.HasErrors) {
-        statusPublisher.SendStatusNotification(textDocument, CompilationStatus.ParsingFailed);
-        return documentAfterParsing;
-      }
 
-      var compilationUnit = symbolResolver.ResolveSymbols(textDocument, program, out _, cancellationToken);
-      var symbolTable = symbolTableFactory.CreateFrom(program, compilationUnit, cancellationToken);
+      var textDocument = documentAfterParsing.TextDocumentItem;
+      var parsedProgram = documentAfterParsing.Program;
+      var clonedProgram = new Program(parsedProgram.Name,
```
MikaelMayer, line 88: Oops the cloning removes all the ranges and owned tokens, that means that the parsed document will be useless for formatting... I don't think we can get around solving https://github.com/dafny-lang/dafny/issues/3290 now

keyboardDrummer: Can we not just clone those as well? Not cloning them seems like a bug

MikaelMayer: Yes, but since they are not part of the constructor, we might forget about them, either in current cloners, in future cloners, or in future language constructs. Anyway, we can add the range manually for now.

PR #3341, opened 2023-01-08
Documenting compilation errors

davidcok:
More work on the error catalog

### DIFF
--- a/Source/IntegrationTests/LitTests.cs
+++ b/Source/IntegrationTests/LitTests.cs
@@ -124,8 +124,7 @@ namespace IntegrationTests {
 
     [FileTheory]
     [FileData(Includes = new[] { "**/*.dfy", "**/*.transcript" },
-              Excludes = new[] { "**/Inputs/**/*", "**/Output/**/*",
-                "examples/induction-principle-code/*"
+              Excludes = new[] { "**/Inputs/**/*", "**/Output/**/*"
               })]
     public void LitTest(string path) {
       LitTestCase.Run(path, Config, output);


Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -3220,7 +3220,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
         var s0 = (AssignStmt)s.S0;
         if (s0.Rhs is HavocRhs) {
           if (DafnyOptions.O.ForbidNondeterminism) {
-            Error(s0.Rhs.Tok, "nondeterministic assignment forbidden by the --enforce-determinism option", wr);
+            Error(s0.Rhs.Tok, "nondeterministic assignment forbidden by the --enforce-determinism", wr);
```
stefan-aws, line 3223:
```suggestion
            Error(s0.Rhs.Tok, "nondeterministic assignment forbidden by --enforce-determinism", wr);
```

docs/HowToFAQ/Errors.md
```diff
@@ -10,8 +10,11 @@ layout: default
 
 <script src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML" type="text/javascript"></script>
 
+This page contains a catalog of the dafny tool's Error and Warning messages,
```
stefan-aws, line 13:
```suggestion
This page contains a catalog of Dafny's error and warning messages,
```

docs/check-examples
```diff
@@ -61,7 +61,7 @@
 ##                  the actual output matches the content of the file <expect>
 ##   %check-run <expect> : runs the command
 ##              "dafny run --use-basename-for-filename $F" and,
-##              if <expect> is present, then the output matches the
+##              if <expect> is present, then checks that the output matches the
```
stefan-aws, line 64:
```suggestion
##              if <expect> is present, checks that the output matches the
```

docs/DafnyRef/Statements.md
```diff
@@ -928,7 +928,7 @@ has the same meaning as
 
 <!-- %no-check -->
 ```dafny
-if exists x :| P { var x :| P; S } else { T }
+if exists x :: P { var x :| P; S } else { T }
```
stefan-aws, line 931: Good catch!

docs/HowToFAQ/Errors-Parser.md
```diff
@@ -1022,13 +1023,11 @@ are grouped. The above example should be written as either `(5 | 6) & 7` or `5 |
 
 ## **Error: too many characters in character literal**
 
+<!-- %check-resolve --unicode-char:false -->
 ```dafny
-const c := '\U{103456}';
+const c := '🚀';
```
stefan-aws, line 1028: How does this illustrate the error type _too many characters in character literal_? What is your issue with the current example? Please check that the links to more detail below this part are correct.

stefan-aws (commented):
Looks good overall. I found a few typos (imo) and have one open question.

docs/HowToFAQ/Errors-Compiler.md
```diff
@@ -1,104 +1,529 @@
...
+  var x: int;
+  x :| x < 5;
+}
+```
 
-## nondeterministic loop forbidden by the --enforce-determinism option
+The `--enforce-determinism` option requires a target program to be deterministic and predictable:
+there may be no program statements that have an arbitrary, even if deterministic, result.
+Hence this 'assign any value that satisfies the predicate' (`:|`) statement is not permitted with `--enforce-determinism`,
+even if there is only one such possible value.
```
atomb, line 320: I could imagine someone saying "why?" about "even if there is only one such possible value". Is it worth explaining here?

docs/HowToFAQ/Errors-Compiler.md
```diff
@@ -1,104 +1,529 @@
...
+}
+```
+
+The `--enforce-determinism` option requires a target program to be deterministic and predictable:
+there may be no program statements that have an arbitrary, even if deterministic, result.
+The [`if` with a binding guard](../DafnyRef/DafnyRef#sec-binding-guards) (`if ... :| `) checks that the 
+given predicate is satisfiable by some value. If not, then the 'else' branch is executed;
+but if so, the 'then' branch is executed with an arbitrary value that satisifies the predicate.
+Because of this arbitrary selection, the if-with-binding-guard is not permitted with `--enforce-determinism`,
+even if there is exactly one value that satisfies the predicate.
```
atomb, line 360: Again, it could be worth having a sentence about why.

docs/HowToFAQ/Errors-Compiler.md
```diff
@@ -1,104 +1,529 @@
...
+The `--enforce-determinism` option requires a target program to be deterministic and predictable:
+there may be no program statements that have an arbitrary, even if deterministic, result.
+The case-based while statement allows the case guards to be evaluated in an arbitrary order and
+an arbitrary one of those found to be true is chosen to be executed.
+Hence the case-based loop is not permitted with `--enforce-determinism`.
+
+To enforce a deterministic order to the evaluation, use a a chain of if-then-else statements
+or series of `if` statements in which the then-branch ends in a continue statement.
+
+## **Error: compiler currently does not support assignments to more-than-6-dimensional arrays in forall statements**
```
atomb, line 424: 6 dimensions ought to be enough for anybody. ;)

docs/HowToFAQ/Errors-Compiler.md
```diff
@@ -1,104 +1,529 @@
...
+TODO
+
+## **Error: Comparison of a handle can only be with another handle**
+
+<!-- TODO -->
+_The documentation of the {:handle} attribute is in progress._
+
+<!-- DafnyCore/Compilers/Synthesizer-Csharp.cs -->
+
+## **Error: '_feature_' is not an element of the _target_ compiler's UnsupportedFeatures set**
```
atomb, line 464: This looks like a duplicate.

atomb (changes requested):

docs/HowToFAQ/Errors.md
davidcok: Edited, but left 'dafny' as that is the tool name.

docs/DafnyRef/Statements.md
davidcok: That's why as much code in docs as possible should be automatically checked...

docs/HowToFAQ/Errors-Parser.md
davidcok: This kind of character is one that is not supported by in --unicode-char:false. I've asked Robin for text to put in as explanation.

docs/HowToFAQ/Errors-Compiler.md
```diff
@@ -1,104 +1,529 @@
...
+The dafny compiler prepares a representation of the Dafny program in the desired target programming
+language and then invokes the target language's compiler on those files as a subsidiary process.
+This error message indicates that this sub-process would not start;
+it gives some additional information about what the invoking process was told about the failure to start.
+
+Typically this error happens when the installation of the underlying compiler is not present or is incomplete,
+such as the proper files not having the correct permissions.
+
+
+## **Error: Error: '_feature_' is not an element of the _target_ compiler's UnsupportedFeatures set**
```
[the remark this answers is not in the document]
davidcok: It indicates a (minor) bug -- the unsupported feature is being used in the program, but the in-tool documentation of unsupported features does not include it.

docs/HowToFAQ/Errors-Compiler.md
```diff
@@ -1,104 +1,529 @@
...
+<!-- %check-legacy %options -compile:2 -Main:mmm -->
+```dafny
+method mmm(i: int) {}
+```
+
+To create an executable program, the compiler needs to know which is the method 
+to call to start the program, typically known as the 'main' method or main entry point. 
+`dafny` chooses such a method based on a few rules (cf. [details in the Dafny User Guide](#sec-user-guide-main)).
+
+But there are restrictions on which methods may be used as a main entry point.
```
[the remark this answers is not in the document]
davidcok: It does -- right after the text excerpt listed here.

docs/HowToFAQ/Errors-Compiler.md
```diff
@@ -1,104 +1,529 @@
...
+--- the compiler does not know what content to give it.
+Even ghost functions must have bodies because body-less ghost functions are a form of unchecked assumptions.
 
-## Error: Function _name_ must be compiled to use the {:test} attribute
+## **Error: Function _name_ must be compiled to use the {:test} attribute**
 
-```dafny <!-- %check-translate -->
+<!-- %check-run -->
+```dafny
 function {:test} f(): int { 42 }
```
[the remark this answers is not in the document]
davidcok: Revising everything for 4.0 syntax is a (large) upcoming task. I did not do it here because users are still using Dafny 3

docs/HowToFAQ/Errors-Compiler.md
davidcok: I added an explanatory sentence -- essentially it might be too hard to do and it might be too hard to determine whether it is too hard to do.

docs/HowToFAQ/Errors-Compiler.md
davidcok: Same explanation

docs/HowToFAQ/Errors-Compiler.md
davidcok: And you can do more than 6 if the indices have obvious ranges.

davidcok: I haven't come up with a failing example yet.

docs/HowToFAQ/Errors-Compiler.md
davidcok: Yes. Removed. (Though in many cases of replicates the same error message is produced I more than one place and I made sure each one was reachable, even if there is just one example.)

docs/HowToFAQ/Errors-Compiler.md
```diff
@@ -1,104 +1,543 @@
...
+  x :| x < 5;
+}
+```
+
+The `--enforce-determinism` option requires a target program to be deterministic and predictable:
+there may be no program statements that have an arbitrary, even if deterministic, result.
+Hence this 'assign any value that satisfies the predicate' (`:|`) statement is not permitted with `--enforce-determinism`,
+even if there is only one such possible value.
+(The tool does not try to determine whether there is just one value and
+whether there is a reasonably way to compute it.)
```
atomb, line 326:
```suggestion
whether there is a reasonable way to compute it.)
```

atomb (changes requested):
A couple of tiny typos, and it looks like the formatter check is failing, but otherwise looks great.

docs/HowToFAQ/Errors-Parser.md
```diff
@@ -1022,18 +1023,19 @@ are grouped. The above example should be written as either `(5 | 6) & 7` or `5 |
...
+iwith a single such unit, i.e. characters in the Basic Multilingual Plane. 
+The rocket ship emoji above, however, is encoded with two surrogate code points.
 
-
-
-A character literal consists of two `'` characters enclosing either
-- a single ASCII character (that is not a backslash)
-- a backslash-escaped character
-- a sequence of characters designating a unicode character
+This can be fixed by enabling the --unicode-char mode, as that defines char as any i
```
stefan-aws, line 1037:
```suggestion
This can be fixed by enabling the --unicode-char mode, as that defines char as any
```

docs/HowToFAQ/Errors-Parser.md
stefan-aws: Thanks for clarifying

davidcok:
Me and my too-quick vi -- thanks Stefan

docs/HowToFAQ/Errors.md
stefan-aws: That's fine with me

stefan-aws (approved):

atomb (approved):

PR #3342, opened 2023-01-08
chore: Remove final `set-output`

fabiomadge:
Fixes #


MikaelMayer (approved):
If it makes the CI to check, then everything else is the same, so I'll approve it tentatively.

PR #3348, opened 2023-01-10
Fix auditor test on Windows

atomb:
This hopefully fixes the test for the auditor on Windows by using `%diff` instead of `diff` in the test file.


atomb:
Closing as the branch is not in the right state.

PR #3351, opened 2023-01-10
chore: Delete outdated assets for the nightly builds

fabiomadge:


robin-aws (changes requested):
Don't do this, we have customers using older nightly builds and haven't warned that we'll be deleting them

fabiomadge:
I agree, let's wait for some pushback before me automate it.

PR #3352, opened 2023-01-11
chore: Only check the nightly tests on `dafny-lang`

fabiomadge:

### DIFF
--- a/.github/workflows/check-deep-tests-reusable.yml
+++ b/.github/workflows/check-deep-tests-reusable.yml
@@ -18,6 +18,7 @@ jobs:
         path: dafny
         submodules: true
     - uses: actions/github-script@v6
+      if: github.repository_owner == 'dafny-lang'
       with:
         script: |
           const script = require('${{ github.workspace }}/dafny/.github/workflows/check-for-workflow-run.js')


davidcok:
Don't do this -- there is no dependency of the doctests on the deep tests. It is useful to be able to check them even if Depp tests fail, even though they will have to wait until the deep tests pass to be merged.

fabiomadge:
This condition should now always pass when not running on `dafny-lang/dafny`. The rational for not running them is keeping nonessential load off CI while it's broken.

davidcok:
Let me check my understanding. When pushes are made to a PR on a fork, do those resulting CI runs happen in a user's own account (and therefore will run anyway), despite a recent failure of the nightly tests?

fabiomadge:
It depends on the target of the PR. If a PR is specified to merge into `dafny-lang/dafny`, it will run on the `dafny-lang/dafny` infrastructure, but if it will end up in the fork, the fork is footing the bill.

keyboardDrummer:
Is there data to indicate we need to reduce load when the nightly is failing? It's costly to prevent other tests from running while nightly is broken since they block people from developing.

fabiomadge:
What data are you looking for? This change doesn't block development in additional circumstances, but does unblocks us faster.

keyboardDrummer:
> What data are you looking for? This change doesn't block development in additional circumstances, but does unblocks us faster.

Doesn't this PR add additional deep-test checks for different workflows? There might be some confusion because this PR also changes the deep-test so it does not run on forks.

fabiomadge:
Yes, in case a block is in place, it will apply to more tests, but it doesn't make the blocking more frequent. I don't think the marginal value of additional tests is constant. If they all run, then you're happy, as soon as a single required test is missing, you can't merge anyway. So there is limited value, while still consuming the scarce resources better used for fixing the broken tests.

keyboardDrummer:
> Yes, in case a block is in place, it will apply to more tests, but it doesn't make the blocking more frequent. I don't think the marginal value of additional tests is constant. If they all run, then you're happy, as soon as a single required test is missing, you can't merge anyway. So there is limited value, while still consuming the scarce resources better used for fixing the broken tests.

I'm trying to settling the discussion of whether we should block tests on the nightly failing or not, and for that it's useful to know whether there is data to suggest not doing so causes a resource scarcity.

fabiomadge:
I personally don't think gathering this data is worth the effort, but you can use your time how you see fit.

keyboardDrummer:
> I personally don't think gathering this data is worth the effort, but you can use your time how you see fit.

I was asking whether you have existing data that leads you to say it's better not to consume build resources while nightly is broken because they're too scarce.

If we don't have that, why don't we try allowing the build to run when nightly is broken? If that turns out to cause a resource scarcity, we can disallow it again.

keyboardDrummer:
> I did experience other builds delaying my CI fixes. Otherwise I wouldn't propose this change.

Okay, I didn't realise that. I haven't noticed it, but maybe I didn't pay enough attention.

I'll leave it up to you do decide, but note that if we don't block the build we get to know better if it causes scarcity, and if we do block the build then we don't.

PR #3353, opened 2023-01-11
Enable testing on personal forks of dafny repo, on push

davidcok:
Triggers testing workflows to run on personal forks, when commits are pushed. The tests will run whether or not nightly tests have succeeded.

For incremental work, delay creating a PR until the work is ready for review and use these local tests to check that all is well.
Look at the Actions page of your forked repo to see the status.


robin-aws (commented):
I like the general idea, but I'm also worried about wasting resources with the way I use my fork, at least. I push to my fork regularly just to back up work, even if the code is nowhere near ready for the full test suite.

Could we find a way using `workflow_dispatch` to trigger all jobs with a single explicit button push instead?

PR #3357, opened 2023-01-12
Enable passing a percentage as a value for --cores

### LINKED ISSUE #3328: Relative load option in new CLI
### Summary

The legacy CLI had the `/vcsLoad` argument, which allowed to specify a multiplier on the number of detected cores. This made things rather self-adjusting across systems.

The new CLI seems to only have `--cores` to specify the absolute number of cores to use. This requires more hand-holding.

### Background and Motivation

Dev machines typically have very different specs to cloud CI runners. It's important to accommodate the verifying/build process to each case.

### Proposed Feature

Keep some way to automatically adjust the load depending on the system. `/vcsLoad` was coarse, but might be good enough.

### Alternatives

_No response_

keyboardDrummer:
Fixes #3328

```
  --cores <count>                         Run the Dafny CLI using <n> cores, or using 
                                          <XX%> of the machine's logical cores. 
                                          [default: 6]
```

Default is 50% of logical cores. Sadly the help shows that number already being resolved. Changing that is complicated.


Source/DafnyCore/Options/BoogieOptionBag.cs
```diff
@@ -17,8 +19,34 @@ public static class BoogieOptionBag {
...
+        return number;
+      }
+
+      result.ErrorMessage = $"Number of cores to use must be greater than 0";
+      return 1;
+    }
+    result.ErrorMessage = $"Could not parse number {value}";
+    return 1;
+  }, true,
+    "Run the Dafny CLI using <n> cores, or using <XX%> of the machine's logical cores.") {
```
MikaelMayer, line 49:
```suggestion
    "Run the Dafny verifier using <n> cores, or using <XX%> of the machine's logical cores.") {
```
Is this text tested somewhere?

Source/XUnitExtensions/Lit/LitCommandWithRedirection.cs
```diff
@@ -38,6 +38,12 @@ public class LitCommandWithRedirection : ILitCommand {
         errorFile = config.ApplySubstitutions(argumentsList[redirectErrorIndex + 1]).Single();
         argumentsList.RemoveRange(redirectErrorIndex, 2);
       }
+      var redirectErrorAppendIndex = argumentsList.IndexOf("2>>");
+      if (redirectErrorAppendIndex >= 0) {
+        errorFile = config.ApplySubstitutions(argumentsList[redirectErrorAppendIndex + 1]).Single();
```
MikaelMayer, line 43: I don't understand this code part, can you please explain it to me?

keyboardDrummer: > I get that you want to redirect the standard error to the file as well, but I'm surprised that you have to introduce a new rule for that.

How would it know what `2>>` means without a new rule? It's a from scratch parser for lit commands so it doesn't understand anything by default. Maybe you could make it smart so if it understand `2` and `>>` separately it can also combine them, but it can't do that right now.

Source/DafnyCore/Options/BoogieOptionBag.cs
keyboardDrummer: I'm not sure I understand your question, the_text itself is not repeated somewhere else where it is diffed against.

Source/XUnitExtensions/Lit/LitCommandWithRedirection.cs
MikaelMayer: > How would it know what 2>> means without a new rule? I

I mean, I'm surprised we hadn't had the need tor redirect error output before this PR, that's what I meant.

My comprehension problem was simpler. I just looked at the code above to see what it was doing, and I understand you are generalizing "2>", which I remembered was used in test cases. So that's good for me.

Source/DafnyCore/Options/BoogieOptionBag.cs
MikaelMayer: I mean, normally, when you change CLI options, the help at --help should reflect these changes. Aren't we testing the output of /help? Maybe we should in a future PR.

MikaelMayer (approved):
Still a few tests to fix, but the functionality is working.

MikaelMayer (approved):

PR #3363, opened 2023-01-13
Do not fail the deep test check on in progress runs

keyboardDrummer:
If there is a deep test check, and multiple runs qualify for that check, but the most recent run in still in progress, do not fail the check on that run, but instead look at the first not in progress run.

Also add a missing return statement so we don't get don't confusing error reporting like this:
image


fabiomadge:
Doesn't this mean that the test always run as long as there is a deep test in progress? You should still make sure that the previous run passed.

keyboardDrummer:
> Doesn't this mean that the tests always run as long as there is a deep test in progress? You should still make sure that the previous run passed.

You're right, thanks! Updated the code.

.github/workflows/check-for-workflow-run.js
```diff
@@ -15,9 +15,10 @@ module.exports = async ({github, context, core, workflow_id, sha, ...config}) =>
   // run for this SHA we see.
   const runFilterDesc = sha ? `${workflow_id} on ${sha}` : workflow_id
   for (const run of result.data.workflow_runs) {
-    if (!sha || run.head_sha == sha) {
-      if (run.conclusion != "success") {
+    if ((!sha || run.head_sha === sha) && run.conclusion !== "in_progress") {
+      if (run.conclusion !== "success") {
         core.setFailed(`Last run of ${runFilterDesc} did not succeed: ${run.html_url}`)
+        return
```
fabiomadge, line 21: Can you pull that out?

fabiomadge (approved):

.github/workflows/check-for-workflow-run.js
keyboardDrummer: I will, but I wonder whether it's worth the extra build trip ^,^

PR #3375, opened 2023-01-18
Remove quantified tokens

keyboardDrummer:
### Changes
Remove QuantifiedVariableRangeToken and QuantifiedVariableDomainToken

### Testing
Refactoring so no testing is needed

### DIFF
--- a/Source/DafnyCore/AST/Expressions/Expressions.cs
+++ b/Source/DafnyCore/AST/Expressions/Expressions.cs
@@ -1711,6 +1711,8 @@ public class TypeTestExpr : TypeUnaryExpr {
 }
 
 public class BinaryExpr : Expression, ICloneable<BinaryExpr> {
+  public bool SecondArgumentIsDomainOfQuantifiedVariable { get; set; }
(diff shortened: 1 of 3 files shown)


Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -1711,6 +1711,8 @@ public TypeTestExpr(IToken tok, Expression expr, Type toType)
 }
 
 public class BinaryExpr : Expression, ICloneable<BinaryExpr> {
+  public bool SecondArgumentIsDomainOfQuantifiedVariable { get; set; }
+  public bool SecondArgumentIsQuantifiedVariableRange { get; set; }
```
MikaelMayer, line 1715: I'm all in to reduce the number of tokens, but now this information is not cloned if the BinaryExpr is clone. How do you ensure that the behavior is the same?

keyboardDrummer: Fixed

MikaelMayer (approved):
Great job, two less token types !

PR #3379, opened 2023-01-19
Remove IToken constructor parameter from statements

keyboardDrummer:
### Changes
- Update all constructors of statements so that they only take a RangeToken and no IToken as an argument. Token based behavior is kept the same by overriding the `Node.Tok` property for certain AST nodes.
- Small behavioural change in the reporting of parallel assignments. The reporting token for assignments on than the first is not the `:=` operator as it was, but the RHS on the assignment.

### Testing
- Updated tests that had reporting on parallel assignments.


Test/dafny0/GhostAutoInit.dfy.expect
```diff
@@ -14,7 +14,7 @@ GhostAutoInit.dfy(125,21): Error: variable 'a', which is subject to definite-ass
 GhostAutoInit.dfy(138,21): Error: variable 'a', which is subject to definite-assignment rules, might be uninitialized here
 GhostAutoInit.dfy(140,21): Error: variable 'b', which is subject to definite-assignment rules, might be uninitialized here
 GhostAutoInit.dfy(153,2): Error: out-parameter 'e', which is subject to definite-assignment rules, might be uninitialized at this return point
-GhostAutoInit.dfy(161,21): Error: cannot establish the existence of LHS values that satisfy the such-that predicate
+GhostAutoInit.dfy(161,4): Error: cannot establish the existence of LHS values that satisfy the such-that predicate
```
MikaelMayer, line 17: This changes the error reporting from

```
    var d: EmptyType :| true; 
                     ^
```
to 
```
    var d: EmptyType :| true; 
    ^
```

which is less meaningful. Could we stick to reporting on `:|`?

keyboardDrummer: Ok!

Source/DafnyCore/AST/Node.cs
```diff
@@ -134,6 +77,78 @@ public abstract class Node : INode {
     }
     set => rangeToken = value;
   }
+}
+
+public abstract class RangeNode : Node {
+  public override RangeToken RangeToken { get; set; } // TODO remove set when TokenNode is gone.
```
MikaelMayer, line 83:
```suggestion
  public override RangeToken RangeToken { get; set; } // TODO remove set when TokenNode is gone.
```
Will TokenNode ever go away?

Source/DafnyCore/AST/Node.cs
```diff
@@ -134,6 +77,78 @@ public abstract class Node : INode {
...
+    RangeToken = cloner.Tok(original.RangeToken);
+  }
+
+  protected RangeNode(RangeToken rangeToken) {
+    RangeToken = rangeToken;
+  }
+}
+
+
+public abstract class Node : INode {
```
MikaelMayer, line 95: Can you please move this class above its children? I would like it better.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -646,7 +646,7 @@ public partial class Translator {
         // translate the body
         TrStmt(m.Body, builder, localVariables, etran);
         m.Outs.Iter(p => CheckDefiniteAssignmentReturn(m.Body.RangeToken.EndToken, p, builder));
-        stmts = builder.Collect(m.Body.Tok);
+        stmts = builder.Collect(m.Body.RangeToken.StartToken); // TODO should this be EndToken?  the parameter name suggests it should be the closing curly
```
MikaelMayer, line 649: Which parameter name? Of Collect?

Test/contract-wrappers/AllExterns.dfy.expect
```diff
@@ -3,9 +3,9 @@ CheckExtern.dfy(22,12): Warning: The ensures clause at this location cannot be c
 
 Dafny program verifier did not attempt verification
 FooTest: FAILED
-	CheckExtern.dfy(3,12): Runtime failure of ensures clause from CheckExtern.dfy:3:13
+	CheckExtern.dfy(3,10): Runtime failure of ensures clause from CheckExtern.dfy:3:13
```
MikaelMayer, line 6: Previously, the error was pointing correclty on the token that was the most representative of the clause:
```
  ensures y >= 10
            ^^
```
With your change, it now points to the start.
```
  ensures y >= 10
          ^
```

which is much more ambiguous (consider a conjunction `a && b`, one might think that the error comes from a() and not from the entire conjunction). Could you put back the previous behavior please?

Test/dafny0/AssumptionVariables0.dfy.expect
```diff
@@ -1,5 +1,5 @@
 AssumptionVariables0.dfy(7,31): Error: there may be at most one assignment to an assumption variable, the RHS of which must match the expression "a0 && <boolean expression>"
-AssumptionVariables0.dfy(8,35): Error: there may be at most one assignment to an assumption variable, the RHS of which must match the expression "a2 && <boolean expression>"
+AssumptionVariables0.dfy(8,42): Error: there may be at most one assignment to an assumption variable, the RHS of which must match the expression "a2 && <boolean expression>"
```
MikaelMayer, line 2: The error was here.
```
    ghost var a1, {:assumption} a2 := true, false;  // error
                                   ^^
```
Now it is here
```
    ghost var a1, {:assumption} a2 := true, false;  // error
                                          ^
```
I'm not very familiar with assumptions, but I just read it was meant to cache verification at the Boogie level.
https://www.microsoft.com/en-us/research/wp-content/uploads/2016/12/krml245.pdf
That said, I think I the error should have been on neither of those places above but on the variable "a2" itself, as it must be assigned "true" and not "false".
If you are unsure about what is best, could you please at least revert to the previous behavior? I don't think highlighting a comma is meaningful anyway.

Test/dafny0/MapMergeSubtraction.dfy.expect
```diff
@@ -6,5 +6,5 @@ MapMergeSubtraction.dfy(49,19): Error: type of right argument to + (map<bool, re
 MapMergeSubtraction.dfy(51,15): Error: map subtraction expects right-hand operand to have type set<bool> (instead got set<int>) (covariant type parameter would require int <: bool)
 MapMergeSubtraction.dfy(65,15): Error: map subtraction expects right-hand operand to have type set<int> (instead got iset<int>)
 MapMergeSubtraction.dfy(67,15): Error: map subtraction expects right-hand operand to have type set<int> (instead got iset<int>)
-MapMergeSubtraction.dfy(84,7): Error: RHS (of type set<int>) not assignable to LHS (of type map<int, real>)
+MapMergeSubtraction.dfy(84,11): Error: RHS (of type set<int>) not assignable to LHS (of type map<int, real>)
```
MikaelMayer, line 9: Same error here, reporting on the "," instead of on the ":=". I think the ":=" is the best of both, but highlighting the first RHS would have been even better.

Test/dafny0/TypeConstraints.dfy.expect
```diff
@@ -1,6 +1,6 @@
 TypeConstraints.dfy(42,6): Error: RHS (of type int) not assignable to LHS (of type bool)
 TypeConstraints.dfy(49,6): Error: RHS (of type int) not assignable to LHS (of type bool)
-TypeConstraints.dfy(56,9): Error: RHS (of type int) not assignable to LHS (of type bool)
+TypeConstraints.dfy(56,13): Error: RHS (of type int) not assignable to LHS (of type bool)
```
MikaelMayer, line 3: Same reporting problem here. Reporting on ":=" (before) is preferable than on "," but reporting on one of the RHS would have been even better.

MikaelMayer (changes requested):
Looks good ! I think this transformation will make it easier to maintain the code.
A few items to fix IMHO but otherwise let's ship it !

Source/DafnyCore/AST/Node.cs
keyboardDrummer: Jep. Next PRs ^,^

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
keyboardDrummer: Yes, it reads `public StmtList Collect(IToken endCurlyBrace)`

Source/DafnyCore/AST/Node.cs
keyboardDrummer: Of course, no problem :)

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
MikaelMayer: Oh I see. This is weird indeed. I think it should be RangeToken.EndToken, but it's true that StartToken probably matches the previous semantics - but these semantics could be wrong.
As a follow-up quick PR, could you experiment replacing StartToken by EndToken? I suspect that it's the token of the "return path", which I would have always thought should be the end curly brace, not the first one which I think is confusing. Perhaps this will be the way to fix it? A low-hanging fruit there.

MikaelMayer (approved):
Thanks for enhancing error reporting as well !

PR #3383, opened 2023-01-20
Error explanations for Rewriter errors

davidcok:
Part of 3181


docs/HowToFAQ/Errors-Rewriter.md
```diff
@@ -0,0 +1,343 @@
+
+<!-- %check-resolve %default %useHeadings -->
+
+<!-- ./DafnyCore/Rewriters/TimeLimitRewriter.cs -->
+
+## **Warning: timeLimitMultiplier annotation overrides " + name + " annotation**
+
+```ddafny
```
alex-chew, line 8:
```suggestion
```dafny
```

docs/HowToFAQ/Errors-Rewriter.md
```diff
@@ -0,0 +1,343 @@
...
+method {:timeLimitMultiplier 10} {:timeLimit 5} m() {}
+```
+
+The {:timeLimitMultiplier n} attribute tells the prover to use a time limit that is the given number
+times the time limit otherwise specified by the options. The {:timeLimit n} attribute simply sets a new
+value for the time limit. It does not make sense to have both attributes on the same declaration. 
+One or the other should be removed.
```
alex-chew, line 15: If both attributes are present, what is the actual behavior? The warning text suggests to me that `timeLImitMultiplier` takes precedence, but the explanation doesn't say one way or the other. Since this is marked as a warning and not an error, I think we should be clear about the actual behavior in case the user chooses to ignore the warning.

docs/HowToFAQ/Errors-Rewriter.md
```diff
@@ -0,0 +1,343 @@
...
+  b ==> b && 
+  b          // this parses as b ==> (b && b), but reads as (b ==> b) && b
+}
+```
+
+Long expressions are best split up across multiple lines; readers often use indentation as a guide
+to the nesting of subexpressions. For example, parallel conjuncts or disjuncts may be written with 
+the same indentation. This warning occurs when the observed indentation is likely to cause
+a misunderstanding of the code, particularly of the precedence of operations. It is suggested
+to use parentheses or to redo the indentation to make the structure of the expression clearer.
```
alex-chew, line 140: Is it necessary to write these two cases as separate explanations?

docs/HowToFAQ/Errors-Rewriter.md
```diff
@@ -0,0 +1,343 @@
...
+```dafny
+method {:test} m(i: int) returns (j: int, k: int) 
+{
+  j, k := 0,0;
+}
+```
+
+This error only occurs when using `dafny test`. That command executes all methods attributed
+with `{:test}`, but such test methods are limited to methods with only one output parameter.
+This is purely an implementation limitation.
```
alex-chew, line 167: If a user is trying to do this, can/should we recommend a workaround of returning a single tuple-typed value? Alternatively, if we plan to fix the implementation limitation, can we include a link to an issue so users can check its status?

docs/HowToFAQ/Errors-Rewriter.md
```diff
@@ -0,0 +1,343 @@
...
+
+The `--test-assumptions` option inserts tests of various contracts and implicit assumptions 
+in a Dafny method, such as the requires and ensures clauses.
+However, because these inserted tests are compiled into the target
+program as runtime checks, they cannot refer to any ghost variables.
+(This mechanism implements Dafny's runtime-assertion checking.)
+ 
+## **Warning: Internal: no wrapper for _name_**
+
+This message indicates a bug within dafny. Please report the program that
```
alex-chew, line 188:
```suggestion
This message indicates a bug within Dafny. Please report the program that
```

docs/HowToFAQ/Errors-Rewriter.md
```diff
@@ -0,0 +1,343 @@
...
+triggered the message to [the Github issues site](https://www.github.com/dafny-lang/dafny/issues).
+
+## **Warning: No :test code calls _name_**
+
+This warning indicates that some method marked with {:extern} is not called
+by any method marked with {:test}. The intention is to check coverage of
+the programmed unit tests. The warning only appears when using
+`/testContracts:TestedExterns` with the legacy CLI.
+It is likely to be removed in a future version of dafny.
```
alex-chew, line 197:
```suggestion
It is likely to be removed in a future version of Dafny.
```

alex-chew (changes requested):

docs/HowToFAQ/Errors-Rewriter.md
davidcok: OK - done

docs/HowToFAQ/Errors-Rewriter.md
davidcok: I don't expect anyone to read this document from top to bottom. I expect they will either (1) search the text using some part of the error message or (2) use a URL link (those anchors are not in place yet) to a specific error explanation from the hover information in the IDE. So I have not worried about repetition. Rather I try to make each explanation standalone.

docs/HowToFAQ/Errors-Rewriter.md
davidcok: Made an issue, and a link here.

docs/HowToFAQ/Errors-Rewriter.md
davidcok: Dafny is the language; dafny is the tool. Reworded to be clearer.

docs/HowToFAQ/Errors-Rewriter.md
davidcok: Again, dafny is the tool. Put in back quotes to emphasize.

docs/HowToFAQ/Errors-Rewriter.md
alex-chew: Ah, I only now see that the error messages differ in the "which starts at" versus "which ends at" - I agree that separate searchable entries are a good idea.

alex-chew (approved):

PR #3384, opened 2023-01-20
Clone the type of FreshExpr when resolved

atomb:
Previously, the when cloning a resolved FreshExpr, the type was not transferred to the clone.

I don't yet have a small test case for this. The original was large and I have yet to reduce it.


Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -1666,6 +1666,7 @@ public FreshExpr(IToken tok, Expression e, string at = null)
     var result = new FreshExpr(cloner.Tok(tok), cloner.CloneExpr(E), At);
     if (cloner.CloneResolvedFields) {
       result.AtLabel = AtLabel;
+      result.Type = Type;
```
keyboardDrummer, line 1669: We should add a cloning constructor and use that instead:

```
public FreshExpr(Cloner cloner, FreshExpr original) : base(cloner, original) {
  At = original.At;
  if (cloner.CloneResolvedFields) {
    AtLabel = original.AtLabel;
  }
}

public FreshExpr Clone(Cloner cloner) { return new FreshExpr(cloner, this); }
```

MikaelMayer (approved):
I wish we had a test for that but I couldn't reproduce an error that this would fix, however, it is better because it clones the type. Thanks for fixing it !

PR #3399, opened 2023-01-24
Deprecate :dllimport and :handle

### LINKED ISSUE #2778: Remove :dllimport attribute (and :handle)
The `:dllimport` attribute was once introduced to interface between Dafny and C/C++ code under .NET. It uses a tiny sliver of a .NET interface to native code. However, I don't think this has ever been used and it introduces some special cases in the Dafny implementation. Plus also, I think there are probably better ways to encode an extern pointer (called "handle" by the current `:dllimport` mechanism). So, I suggest we remove this attribute.

davidcok:
Fixes #2778 and recent discussion: deprecates :dllimport and :handle


RustanLeino (approved):
Thanks for implementing this deprecation message.

PR #3400, opened 2023-01-24
Update tests to work with Z3 4.12.1

atomb:
The updates the default version of Z3 to 4.12.1 and updates superficial aspects of the tests (error order, counterexample values, occasional command-line option tweaks) to work with it.

Those test changes have some minor caveats:
* Some of the tests in `DafnyTestGeneration.Test` check very specific things about the solver results that I don't think are required of valid results. I've commented out what I think are the overly-restrictive checks, but they should really be replaced with similar but less restrictive ones. @Dargones, do you have any thoughts?
* Some tests require specific settings of the Z3 parameter `smt.qi.eager_threshold` to succeed, but there isn't a single value that allows them all to succeed. There's even one test I had to split into two files to give two values of this parameter. Let's investigate whether the current value is a good one, in general. Tracked as #3558.
* Dafny is now able to prove an assertion in `Fuel.dfy` that it shouldn't be able to, according to comments in the file. I don't think this is technically either a soundness or incompleteness issue, but it's a strange behavior. Edit: it turns out that the latest nightly of Z3 goes back to being unable to prove that assertion.
* One test (`github-issues/github-issue-2174.dfy`) requires a special setting of the arithmetic solver. This is something we should investigate further on other code bases. Issue #3501 tracks that investigation.
* The `BitvectorsMore` test was missing errors since there were too many in one method. It had `/errorLimit:10`, but that wasn't even high enough. A newer solver version resulted in the order of errors changing, causing known errors to be hidden and previously-untracked errors to be revealed. A higher error limit solves this.


davidcok (dismissed):
Just a few changes on the PR itself. Also
- anything that 'needs investigating" should be turned into an Issue (whose number is given in the PR description).
But the big question is this: I'm very wary of requiring users to update to the very latest OS. Can we not support whatever Z3 is needed to go back at least one OS version? And automatically make the recommended options for Z3. As a user, I need to be concerned with proving my program -- I would not want to have to be a Z3 expert (or even know that Z3 is there).  Getting or building 4.12.1 for older OSes is likely the easier course -- but should be taken before we spring this on users.

Source/DafnyTestGeneration.Test/Various.cs
```diff
@@ -153,7 +153,7 @@ static method eightPaths (i:int)
 ".TrimStart();
       var program = Utils.Parse(source);
       var methods = await Main.GetTestMethodsForProgram(program).ToListAsync();
-      Assert.IsTrue(methods.Count is >= 4 and <= 6);
+      Assert.IsTrue(methods.Count is >= 3 and <= 6);
```
Dargones, line 156: I think changing this to `methods.Count is >= 2 and <= 6` would make sure the assertion accounts for all possible solutions.

Dargones:
Thank you, @atomb for pinging about test generation tests. The changes make sense to me - I will look over other tests too to see if there are any potential problems.

Test/allocated1/dafny0/BitvectorsMore.dfy
```diff
@@ -1,3 +1,3 @@
-// RUN: %exits-with 4 %dafny /verifyAllModules /allocated:1 /print:"%t.print" /env:0 "%s" > "%t"
+// RUN: %exits-with 4 %dafny /errorLimit:10 /verifyAllModules /allocated:1 /print:"%t.print" /env:0 "%s" > "%t"
```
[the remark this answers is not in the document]
atomb: Just to be clear, this _increases_ the error limit beyond the default of 5, so that we collect all errors that happen. I just looked, though, and using `0` would tell it to find as many as possible, so that'd be a better choice here, and everywhere in the test suite.

I generally agree that tests with a large number of errors should be broken up, but the particular case where this was necessary feels, to me, better as a unit than multiple pieces.

Test/dafny0/FunctionSpecifications.dfy
```diff
@@ -49,7 +49,7 @@ function FibWithExtraPost(n: int): int
 function DivergentPost(n: int): int
   requires 0 <= n;
   ensures 1 <= n ==> DivergentPost(n-1) == DivergentPost(n-1);
-  ensures DivergentPost(2*n - n) == DivergentPost(2*(n+5) - 10 - n);  // these are legal ways to denote the result value of the function
+  //ensures DivergentPost(2*n - n) == DivergentPost(2*(n+5) - 10 - n);  // these are legal ways to denote the result value of the function
```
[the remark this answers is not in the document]
atomb: Ah, I didn't mean to leave that commented out. Good catch!

Source/DafnyTestGeneration.Test/Various.cs
atomb: Thanks!

atomb:
> Thank you, @atomb for pinging about test generation tests. The changes make sense to me - I will look over other tests too to see if there are any potential problems.

Thanks for confirming!

atomb:
> Just a few changes on the PR itself. Also
> 
>     * anything that 'needs investigating" should be turned into an Issue (whose number is given in the PR description).
>       But the big question is this: I'm very wary of requiring users to update to the very latest OS. Can we not support whatever Z3 is needed to go back at least one OS version? And automatically make the recommended options for Z3. As a user, I need to be concerned with proving my program -- I would not want to have to be a Z3 expert (or even know that Z3 is there).  Getting or building 4.12.1 for older OSes is likely the easier course -- but should be taken before we spring this on users.

I wasn't clear in the original description, but most of the "needs investigating" I plan to resolve before merging this. But, for any that remain, I'll definitely create an issue.

As to requiring recent Linux distributions, I totally agree. My open question is how best to resolve it, and I was hoping reviewers may have thoughts. The primary options I see are 1) building it ourselves as part of CI (which would slow down our builds quite a lot), 2) building a binary like the one on the Z3 GitHub releases page but compatible with older Linux distributions and putting it somewhere generally accessible (where?), or 3) waiting a little while to see if the Z3 developers respond to the issue I opened about their Linux binaries (but we don't want to wait very long).

atomb:
Update on the Z3 binary issue: they've already updated the nightly builds to go back to ubuntu-20.04, and are looking into how they could update the release.

Test/allocated1/dafny0/BitvectorsMore.dfy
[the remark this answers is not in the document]
davidcok: Yes - we should make errorLimit=0 the default in the tests for everything -- both in lit and in the CI tests. I'll add that to my Dafny 4 run and see if we catch anything that way.

atomb:
Several updates:
* This now uses the nightly 4.12.2 pre-release build as a test of how well those builds work for us. This causes some tests to fail again. Shortly, I'm going to update this to use the self-built Z3 binaries from https://github.com/dafny-lang/solver-builds, which should resolve those test failures.
* The Windows DafnyLanguageServer.Test runs are running out of memory. I'm trying to diagnose why.
* I think I've addressed all other comments on this so far.

atomb:
The Z3 builds used by this PR are now custom-built from the release commits of Z3 4.12.1, hosted [here](https://github.com/dafny-lang/solver-builds). (Binaries for 4.8.5 are also available in that form.)

The only remaining issue I'm aware of is that of the language server unit tests running out of memory.

robin-aws (changes requested):
Conservatively blocking this until it retargets `main-4.x` instead (and most of the content will be merged in separate PRs first)

Test/allocated1/Allocated1.dfy
```diff
@@ -1,4 +1,4 @@
 
-// RUN: %exits-with 4 %dafny /verifyAllModules /allocated:1 /errorLimit:99 /compile:0 /dprint:"%t.dprint" /autoTriggers:0 "%s" > "%t"
+// RUN: %exits-with 4 %dafny /verifyAllModules /allocated:1 /errorLimit:0 /compile:0 /dprint:"%t.dprint" /autoTriggers:0 "%s" > "%t"
```
atomb, line 2: This might make sense to include with 3.x, too. Same with other instances of `/errorLimit:0`.

Test/allocated1/dafny0/BitvectorsMore.dfy.expect
```diff
@@ -9,6 +9,7 @@ BitvectorsMore.dfy(105,35): Error: shift amount must not exceed the width of the
 BitvectorsMore.dfy(105,38): Error: when converting shift amount to a bit vector, the value to be converted might not fit in bv7
 BitvectorsMore.dfy(107,34): Error: shift amount must not exceed the width of the result (67)
 BitvectorsMore.dfy(107,37): Error: when converting shift amount to a bit vector, the value to be converted might not fit in bv7
+BitvectorsMore.dfy(108,34): Error: shift amount must not exceed the width of the result (67)
```
atomb, line 12: The extra errors are an artifact of `/errorLimit:0`.

Test/cli/errorLimit.dfy
```diff
@@ -1,4 +1,4 @@
-// RUN: %exits-with 4 %baredafny verify %args --error-limit=2 "%s" > "%t"
+// RUN: %exits-with 4 %baredafny verify %args --error-limit=0 "%s" > "%t"
```
atomb, line 1: This change might make sense to include for 3.x, as well. Any error limit other than 0 can lead to non-determinism.

Test/dafny0/Termination.dfy
```diff
@@ -487,7 +487,7 @@ datatype Tree = Empty | Node(root: int, left: Tree, right: Tree)
       }
   }
 
-  lemma {:induction this} EvensSumToEven()  // explicitly use "this" as quantified over by induction hypothesis
+  lemma {:induction this} {:vcs_split_on_every_assert} EvensSumToEven()  // explicitly use "this" as quantified over by induction hypothesis
```
atomb, line 490: I discovered these during the creation of this branch, but it couldn't hurt to include them in 3.x, as well.

Test/dafny2/SnapshotableTrees.dfy
```diff
@@ -1,4 +1,4 @@
-// RUN: %exits-with 4 %dafny /compile:4 /dprint:"%t.dprint" "%s" > "%t"
```
atomb, line 1: I should probably switch this back to 4 and add back in the execution output.

Test/dafny4/SoftwareFoundations-Basics.dfy
```diff
@@ -269,7 +269,7 @@ lemma plus_lemma(a: Nat, b: Nat, c: Nat)
 {}
 
 // This lemma expresses:  m*(2*n) == (2*m)*n
-lemma mult_lemma(m: Nat, n: Nat)
+lemma {:vcs_split_on_every_assert} mult_lemma(m: Nat, n: Nat)
```
atomb, line 272: This could also be included for 3.x

atomb:
The things that should be included in 3.x are now in #3628. We can merge that into`master` and update `main-4.0` to include it, as well.

Test/dafny0/BitvectorsMore.dfy.expect
```diff
@@ -471,12 +473,14 @@ BitvectorsMore.dfy(157,26): Error: shift amount must not exceed the width of the
...
+BitvectorsMore.dfy(160,26): Error: shift amount must not exceed the width of the result (2)
 BitvectorsMore.dfy(168,33): Error: when converting shift amount to a bit vector, the value to be converted might not fit in bv0
 BitvectorsMore.dfy(169,33): Error: when converting shift amount to a bit vector, the value to be converted might not fit in bv0
 BitvectorsMore.dfy(170,33): Error: when converting shift amount to a bit vector, the value to be converted might not fit in bv0
 BitvectorsMore.dfy(171,33): Error: when converting shift amount to a bit vector, the value to be converted might not fit in bv0
 BitvectorsMore.dfy(193,26): Error: shift amount must not exceed the width of the result (5)
 BitvectorsMore.dfy(194,26): Error: shift amount must not exceed the width of the result (5)
 
-Dafny program verifier finished with 9 verified, 37 errors
+Dafny program verifier finished with 9 verified, 41 errors
```
atomb, line 486: The larger number of errors here is due to `/errorLimit:0`.

.github/workflows/doc-tests.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       run: |
         sudo apt-get install -qq libarchive-tools
         mkdir -p dafny/Binaries/z3/bin
-        wget -qO- https://github.com/dafny-lang/solver-builds/releases/download/snapshot-2023-02-17/z3-4.8.5-ubuntu-20.04-bin.zip | bsdtar -xf -
+        wget -qO- https://github.com/dafny-lang/solver-builds/releases/download/snapshot-2023-02-17/z3-4.12.1-ubuntu-20.04-bin.zip | bsdtar -xf -
```
robin-aws, line 39: FYI (to fix before the next time we change the version number :), it's not too hard to extract out the value from source code to avoid duplicating it all over the place in these files. For example: https://github.com/dafny-lang/dafny/blob/master/.github/workflows/msbuild.yml#L35

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
```diff
@@ -60,13 +60,13 @@ method Abs(x: int) returns (y: int)
...
-This is assertion #1 of 4 in method Abs  
+This is assertion #2 of 4 in method Abs  
 Resource usage: ??? RU"
       );
       await AssertHoverMatches(documentItem, (7, 11),
         @"[**Error:**](???) assertion might not hold  
-This is assertion #2 of 4 in method Abs  
-Resource usage: 9K RU"
+This is assertion #1 of 4 in method Abs  
+Resource usage: ??? RU"
```
robin-aws, line 69: Do you know why we lost this?

Test/allocated1/dafny0/Computations.dfy
```diff
@@ -1,3 +1,3 @@
-// RUN: %dafny /verifyAllModules /allocated:1 /compile:0 /print:"%t.print" /dprint:"%t.dprint" "%s" > "%t"
+// RUN: %dafny /verifyAllModules /allocated:1 /proverOpt:O:smt.qi.eager_threshold=25 /compile:0 /print:"%t.print" /dprint:"%t.dprint" "%s" > "%t"
```
robin-aws, line 1: Is this an option we should (right now) suggest as a verification technique in the manual and (slightly later) add a Dafny-level option for of some kind?

Test/dafny0/AutoContracts.dfy.expect
```diff
@@ -581,11 +581,11 @@ AutoContracts.dfy(50,4): Error: A postcondition might not hold on this return pa
 AutoContracts.dfy(49,24): Related location: This is the postcondition that might not hold.
 AutoContracts.dfy(79,21): Error: A postcondition might not hold on this return path.
 AutoContracts.dfy(60,16): Related location: This is the postcondition that might not hold.
-AutoContracts.dfy[N1](65,20): Related location
 AutoContracts.dfy(79,21): Error: A postcondition might not hold on this return path.
 AutoContracts.dfy(60,16): Related location: This is the postcondition that might not hold.
 AutoContracts.dfy[N1](65,20): Related location
 AutoContracts.dfy(79,21): Error: A postcondition might not hold on this return path.
 AutoContracts.dfy(60,16): Related location: This is the postcondition that might not hold.
+AutoContracts.dfy[N1](65,20): Related location
```
robin-aws, line 589: Not related to this PR, but why do we have "N1" as a line number?

Test/dafny2/SnapshotableTrees.dfy
robin-aws: Yup :)

Test/git-issues/git-issue-2301.dfy
```diff
@@ -48,7 +48,7 @@ class Twostate {
   }
 
   twostate predicate GoodVariations(c: Twostate, d: Twostate, e: Twostate, f: Twostate)
-    reads [this], c, {d, e}, multiset{f}
+    reads this, c, {d, e}, multiset{f}
```
robin-aws, line 51: Why did this have to change?

Test/vstte2012/BreadthFirstSearch.dfy
```diff
@@ -111,10 +111,7 @@ class BreadthFirstSearch<Vertex(==)>
         C, N, d := N, {}, d+1;
       }
 
-      assert dest in R(source, d, AllVertices) ==> dest in C by { reveal R(); }
-      assert d != 0 ==> dest !in R(source, d-1, AllVertices) by { reveal R(); }
```
robin-aws, line 115: Why is it okay to remove these?

robin-aws (approved):
WHEW. I do want to discuss my questions but merging this to get a RC for Dafny 4.0 earlier. I cut #3651 to make sure we come back to them before releasing.

.github/workflows/doc-tests.yml
atomb: Yeah, that would be a nice improvement.

Source/DafnyLanguageServer.Test/Lookup/HoverVerificationTest.cs
atomb: The `???` matches anything, so that change just makes the test less fragile as the solver evolves.

Test/allocated1/dafny0/Computations.dfy
atomb: I've considered that. I'd like to experiment with it a bit more on production code and see if it makes any difference there.

Test/dafny0/AutoContracts.dfy.expect
atomb: No idea! :)

Test/vstte2012/BreadthFirstSearch.dfy
atomb: They were intermediate assertions introduced to make it easier to prove the loop invariant. But it turned out that only one of them was necessary. The rest just made Z3 work harder.

robin-aws: Thanks, I suspected as much. It's hard to tell on a quick glance which is the "important" assertion. Perhaps specification coverage mechanisms would help here. :)

Test/dafny0/AutoContracts.dfy.expect
robin-aws: Oh I actually missed that it was in addition to the actual source location. It's the first time I've encountered that.

PR #3407, opened 2023-01-26
Separate execution and compilation code of back-ends

keyboardDrummer:
This is a refactoring that separates translation from Dafny to a string in a particular language, from other language specific tasks, such as running a program in that language or building a native executable artifact for that language.

While I think this refactoring is good on its own, the main driver for it is to enable another refactoring. I'd like to extract logic that doesn't relate to string manipulation out of `SinglePassCompiler` by moving it to a base class, so that this logic can be reused when compiling Dafny to an AST (as opposed to compiling to a string or concrete syntax tree). This PR removes the base class from `SinglePassCompiler`, freeing it up to have another base class to which logic can be extracted.


RustanLeino (approved):
Looks good. In doing this review, I assumed that most of the deletions from compiler and additions to backend are code moves. In other words, I did not try to check that the cut-and-paste was done correctly.

PR #3412, opened 2023-01-26
Test if this is a typo in prelude

### LINKED ISSUE #3411: Higher-order functions problem
### Discussed in https://github.com/dafny-lang/dafny/discussions/3405

<div type='discussions-op-text'>

<sup>Originally posted by **Nangos** January 26, 2023</sup>
Hi All,

I recently met a problem when verifying with higher-order functions. It boils down to the following minimal example:

```dafny
var f: int -> int;
assert seq(1, x => f(x))[0] == f(0);  // No problem
assert seq(1, f)[0] == f(0);  // Error: assertion might not hold
```

In the general case (especially when writing _class methods_), one would have to write `x reads f.reads(x) requires f.requires(x) => f(x)` in the place of `f`. This will be quite a boilerplate.

After some random attempts, I found an acceptable workaround:
```dafny
// Seems like a mere identity function. What's going on here?
function magic<S, T>(f: S ~> T): (S ~> T) {
  x reads f.reads(x) requires f.requires(x) => f(x)
}

assert seq(1, magic(f))[0] == f(0);  // No problem now
```

I am still curious about what exactly is happening under the hood and whether there are more elegant ways to deal with this.  Any suggestion is welcome.</div>

MikaelMayer:
This PR fixes #3411 
Seems like one of Boogie's axiom was wrong, but it was unnoticed because another application axiom was ignoring the wrong argument.
I added the test case of the issue.

### DIFF
--- a/Test/git-issues/git-issue-3411.dfy
+++ b/Test/git-issues/git-issue-3411.dfy
@@ -1,0 +1,7 @@
+// RUN: %baredafny verify %args "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method Test(f: int -> int) {
+  assert seq(1, x => f(x))[0] == f(0);  // No problem
+  assert seq(1, f)[0] == f(0);  // Error: assertion might not hold
+}
--- a/Test/git-issues/git-issue-3411.dfy
+++ b/Test/git-issues/git-issue-3411.dfy
@@ -3,5 +3,5 @@
 
 method Test(f: int -> int) {
   assert seq(1, x => f(x))[0] == f(0);  // No problem
-  assert seq(1, f)[0] == f(0);  // Error: assertion might not hold
+  assert seq(1, f)[0] == f(0);  // No problem anymore
 }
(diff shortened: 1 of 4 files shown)


Test/git-issues/git-issue-3411.dfy
```diff
@@ -0,0 +1,7 @@
+// RUN: %baredafny verify %args "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method Test(f: int -> int) {
+  assert seq(1, x => f(x))[0] == f(0);  // No problem
+  assert seq(1, f)[0] == f(0);  // Error: assertion might not hold
```
MikaelMayer, line 6:
```suggestion
  assert seq(1, f)[0] == f(0);  // No problem anymore
```

RustanLeino (approved):
Great sleuthing! Thanks for fixing this bug! (I'm surprised anything at all could have worked with the bug.)

MikaelMayer:
The reason why it was working before was because of this axiom

```boogie
axiom (forall t0: Ty, 
    t1: Ty, 
    heap: Heap, 
    h: [Heap,Box]Box, 
    r: [Heap,Box]bool, 
    rd: [Heap,Box]Set Box, 
    bx0: Box :: 
  { Apply1(t0, t1, heap, Handle1(h, r, rd), bx0) } 
  Apply1(t0, t1, heap, Handle1(h, r, rd), bx0) == h[heap, bx0]);
```
As you can see, `t1` is not used in this beta reduction.

PR #3414, opened 2023-01-26
Support for dfy files sources in Dafny itself.

MikaelMayer:
This is a first integration of a sample DFY file in our codebase. This DFY file will be completed by #2399 .

### DIFF
--- a/Source/DafnyCore/AST/Formatting.dfy
+++ b/Source/DafnyCore/AST/Formatting.dfy
@@ -1,0 +1,97 @@
+/// Microsoft.Dafny.Formatting.ReindentProgramFromFirstToken() takes
+/// - A first token, from which an entire program is available
+/// - A IIndentationFormatter, that knows how to indent every token
+/// It then prints every token (proven) in order, reindenting its leading trivia and trailing trivia
+
+module {:extern "System"} {:compile false} {:options "-functionSyntax:4"} System {
+  class {:extern "Text.StringBuilder"} {:compile false} CsStringBuilder {
+    ghost var built: CsString
+    constructor {:extern} ()
+    method {:extern "Append"} Append(s: CsString)
+      modifies this
+      ensures built == String.Concat(old(built), s)
+    // This hack is necessary because ToString is replaced by ToString_
+    // in the C# compiler, even if it's declared extern...
+    method {:extern @"ToString().ToString"} ToString() returns (r: CsString)
+      ensures r == built
+  }
+  type {:extern "String"} CsString(!new,==) {
+    ghost predicate {:extern} Contains(needle: CsString)
+    lemma {:axiom} ContainsTransitive(other: CsString, needle: CsString)
+      requires Contains(other) && other.Contains(needle)
+      ensures Contains(needle)
+  }
+  lemma ContainsTransitive()
+    ensures forall s1: CsString, s2: CsString, s3: CsString
+              | s1.Contains(s2) && s2.Contains(s3) :: s1.Contains(s3)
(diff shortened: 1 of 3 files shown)


MikaelMayer:
@keyboardDrummer  looking forward your feedback on this one.

Source/DafnyCore/AST/Formatting.dfy
```diff
@@ -0,0 +1,97 @@
+/// Microsoft.Dafny.Formatting.ReindentProgramFromFirstToken() takes
+/// - A first token, from which an entire program is available
+/// - A IIndentationFormatter, that knows how to indent every token
+/// It then prints every token (proven) in order, reindenting its leading trivia and trailing trivia
+
+module {:extern "System"} {:compile false} {:options "-functionSyntax:4"} System {
```
keyboardDrummer, line 6: Should this external module be in a separate file?

keyboardDrummer (commented):
I'm a bit confused because the Dafny code doesn't seem to be used. It's just externals right?

Source/DafnyCore/AST/Formatting.dfy
MikaelMayer: > I'm a bit confused because the Dafny code doesn't seem to be used. It's just externals right?

Yikes ! Let me know what happened during this refactoring.

This external module acts as an "include" with an equivalent of "use", so I think it's relevant to keep it in this file, until we actually need to reuse it.

MikaelMayer:
I fixed this PR and the PR description as well, thanks @keyboardDrummer  for spotting the issue.

robin-aws (changes requested):
Great, good to fix this gap. Make sure you add some kind of validation in CI that the generated code actually matches what the Dafny code compiles to though. You might be able to imitate what I did for the runtime code: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntimeDafny/Makefile#L17-L18

MikaelMayer:
> Great, good to fix this gap. Make sure you add some kind of validation in CI that the generated code actually matches what the Dafny code compiles to though. You might be able to imitate what I did for the runtime code: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntimeDafny/Makefile#L17-L18

That was a great suggestion and a good hint to get me started on that. I also added a check like you that Dafny files should be formatted :-)

Source/DafnyCore/AST/Formatting.dfy
```diff
@@ -0,0 +1,183 @@
...
+        // returns the trailing trivia but with its indentation corrected.
+        function GetNewTrailingTrivia(token: IToken): CsString
+
+        // Given a token (including leading and trailing trivia), the current indentation, and whether the last trivia
+        // was ending with a newline, returns the two new leading and trailing trivia of the token,
+        // the new current indentation and whether the last trailing trivia ended with a newline
+        method GetNewLeadingTrailingTrivia(token: IToken)
+          returns (newLeadingTrivia: CsString,
+                   newTrailingTrivia: CsString)
+          requires token.Valid()
```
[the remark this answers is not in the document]
MikaelMayer: > Are any of the requires or ensures clauses needed? This method doesn't have a modifies clause so I would assume token.allTokens == old(token.allTokens) holds by default.

Good catch. I removed all pre/postconditions as it works without.

keyboardDrummer (approved):
Awesome :-) thanks

keyboardDrummer (approved):

PR #3418, opened 2023-01-30
Dafny to Dafny compiler sketch

jtristan:
Skeleton for the Dafny to Dafny compiler.


MikaelMayer:
I would love to merge it incrementally to master, but I see that some tests labelled "for all compilers" actually are preventing us from doing so. They should be deactivated for in-development compilers and reactivated later.
@robin-aws  what's your idea for that? I think you were pushing to test all compilers in many files.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -347,6 +347,7 @@ protected interface IClassWriter {
     protected virtual string True { get => "true"; }
     protected virtual string False { get => "false"; }
     protected virtual string Conj { get => "&&"; }
+    protected virtual string AssignmentSymbol { get => " = "; }
```
MikaelMayer, line 350: Ahah that's a fun fact that Dafny is the first language we support to override that AssignmentSymbol ^^. I had advocated to change that but I guess it's too late.

Source/DafnyCore/Compilers/Dafny/Compiler-dafny.cs
```diff
@@ -0,0 +1,750 @@
...
+    }
+
+    protected override ConcreteSyntaxTree EmitForStmt(IToken tok, IVariable loopIndex, bool goingUp, string endVarName,
+      List<Statement> body, LList<Label> labels, ConcreteSyntaxTree wr) {
+      // var direction = goingUp ? "to" : "downto";
+      // var lowWr = new ConcreteSyntaxTree();
+      // wr.Format($"for {IdName(loopIndex)} := {lowWr} {direction} {endVarName}");
+      // var bodyWr = wr.NewBlock();
+      // TrStmtList(body, bodyWr);
+      // return lowWr;
```
MikaelMayer, line 324: This is commented out because it's not working well enough? You can keep it, just asking.

Source/DafnyCore/Compilers/Dafny/Compiler-dafny.cs
```diff
@@ -0,0 +1,750 @@
...
+      bool mustInitialize, [CanBeNull] string exampleElement, ConcreteSyntaxTree wr, ConcreteSyntaxTree wStmts) {
+      throw new UnsupportedFeatureException(Token.NoToken, Feature.RunAllTests);
+    }
+
+    protected override void EmitLiteralExpr(ConcreteSyntaxTree wr, LiteralExpr e) {
+      wr.Write(e.ToString());
+    }
+
+    protected override void EmitStringLiteral(string str, bool isVerbatim, ConcreteSyntaxTree wr) {
+      wr.Write($"\"{str}\"");
```
MikaelMayer, line 381: I think you need an escaping mechanism there down the road, for now I suggest you say it's unsupported.
```suggestion
      if(str.Contains("\"") || str.Contains("\\") || !str.All(char.IsAscii)) {
        throw new UnsupportedFeatureException(Token.NoToken, Feature.UnicodeChars);
      } else {
        wr.Write($"\"{str}\"");
      }
```

Source/DafnyCore/Compilers/Dafny/DafnyBackend.cs
```diff
@@ -0,0 +1,97 @@
...
+     * In order to work for the continuous integration, we need to call the Dafny compiler using dotnet
+     * because dafny is not currently in the path
+     */
+
+    var where = System.Reflection.Assembly.GetExecutingAssembly().Location;
+    where = System.IO.Path.GetDirectoryName(where);
+    var dafny = where + "/Dafny.dll";
+
+    var opt = DafnyOptions.O;
+    var psi = PrepareProcessStartInfo("dotnet", opt.MainArgs.Prepend("/compileTarget:cs").Prepend("/compile:4").Prepend("/compileVerbose:0").Prepend("/printVerifiedProceduresCount:0").Prepend("/noVerify").Prepend(targetFilename).Prepend(dafny));
```
MikaelMayer, line 57:
```suggestion
    var psi = PrepareProcessStartInfo("dotnet", opt.MainArgs
      .Prepend("/compileTarget:cs")
      .Prepend("/compile:4")
      .Prepend("/compileVerbose:0")
      .Prepend("/printVerifiedProceduresCount:0")
      .Prepend("/noVerify")
      .Prepend(targetFilename)
      .Prepend(dafny));
```

MikaelMayer (changes requested):
Looks promising ! I suggest two minor improvements.

Source/DafnyCore/Compilers/Dafny/Compiler-dafny.cs
jtristan: Correct, and I would like to keep it because it is going in the right direction and is an example, for me, on how to use the "write head" properly

Source/DafnyCore/Compilers/Dafny/Compiler-dafny.cs
jtristan: I'm confused, I did test it on programs that use unicode and it worked.

MikaelMayer (approved):
Good to go ! Let's ship it !

PR #3426, opened 2023-01-31
Cok 2542 slice - Specifications

davidcok:


docs/DafnyRef/Specifications.md
```diff
@@ -424,21 +443,28 @@ class A {
...
+function f()
+  reads *
+function g()
+  reads o, oo
+function h()
+  reads { o }
+```
 
 Functions are not allowed to have side effects; they may also be restricted in
 what they can read. The _reading frame_ of a function (or predicate) is all
```
stefan-aws, line 466:
```suggestion
what they can read. The _reading frame_ of a function (or predicate) consists of all
```

docs/DafnyRef/Specifications.md
```diff
@@ -503,6 +529,11 @@ a function from `int` to `set<object>`, then `reads F` has the meaning
 set x: int, o: object | o in F(x) :: o
 ```
 
+For each function value `f`, Dafny defines the function `f.reads`,
+which takes the same arguments as `f` and returns that set of objects
+that `f` reads (according to its reads clause) with those arguments.
+`f.reads` has type `int ~> set<object>`.
```
stefan-aws, line 535: Are you referring to a particular function `f` here?

docs/DafnyRef/Specifications.md
```diff
@@ -601,61 +635,80 @@ interprets JML's `assigns/modifies` in sense (b).
...
+{
+  var i := 10;
+  while 0 < i
+    invariant 0 <= i < 10
+}
+```
 
 An `invariant` clause is used to specify an invariant
 for a loop. If more than one `invariant` clause is given for
 a loop the effective invariant is the conjunction of
```
stefan-aws, line 653:
```suggestion
a loop, the effective invariant is the conjunction of
```

docs/DafnyRef/Specifications.md
```diff
@@ -601,61 +635,80 @@ interprets JML's `assigns/modifies` in sense (b).
...
-the conditions specified.
+the conditions specified, in their textual order.
```
stefan-aws, line 654: What is a textual order?

docs/DafnyRef/Specifications.md
```diff
@@ -601,61 +635,80 @@ interprets JML's `assigns/modifies` in sense (b).
...
+    decreases i
+    ensures r >= 0
+  { 
+    ... 
   }
-````
+}
+```
 
 A method specification is zero or more `modifies`, `requires`,
```
stefan-aws, line 681:
```suggestion
A method specification consists of zero or more `modifies`, `requires`,
```

stefan-aws: This sounds quite _formal_ to me. Same elsewhere. Is this a standard way of expressing these sort of things?

docs/DafnyRef/Specifications.md
```diff
@@ -667,48 +720,39 @@ clauses because lambda expressions do not have names and thus cannot be recursiv
...
 A loop specification provides the information Dafny needs to
-prove properties of a loop. The ``InvariantClause_`` clause
+prove properties of a loop. It contains `invariant`,
+`decreases`, and `modifies` clauses.
+
+The `invariant` clause
 is effectively a precondition and it along with the
 negation of the loop test condition provides the postcondition.
-The ``DecreasesClause`` clause is used to prove termination.
+The `decreaases` clause is used to prove termination.
```
stefan-aws, line 753:
```suggestion
The `decreases` clause is used to prove termination.
```

docs/DafnyRef/Specifications.md
```diff
@@ -531,14 +561,21 @@ reads on the values from `lo` to `hi`, but the larger set denoted by
...
+```dafny
+class A { var f: int }
+const o: object?
+const p: A?
+method M()
+  modifies { o, p }
+method N()
+  modifies { }
+method Q()
+  modifies o,p`f
```
stefan-aws, line 577:
```suggestion
  modifies o, p`f
```

docs/DafnyRef/Refinement.md
```diff
@@ -61,7 +61,7 @@ Typically this is the presence of a `...` token.
 
 ## 22.1. Export set declarations
 
-A refining export set declaration begins with the syntax
+A refining export set declaration begins with [the syntax](#g-module-export}
```
stefan-aws, line 64: This link doesn't work yet.

stefan-aws (changes requested):
Great job, thank you.

docs/DafnyRef/Specifications.md
davidcok: No, its 'forall' functions f

docs/DafnyRef/Specifications.md
davidcok: The order in the source text, in case previous ones are needed to establish later ones or to ensure the later ones are well-defined. Same as for preconditions and for postconditions

docs/DafnyRef/Specifications.md
davidcok: I'd say so. This document is a combination of a reference manual (for which formality is typical), a user guide and a tutorial. So the language style is a bit of a mix.

docs/DafnyRef/Specifications.md
stefan-aws: Then, why do you write `f.reads` has type `int ~> set<object>`. Shouldn't it be, say, `T` instead of `int`?

docs/DafnyRef/Refinement.md
davidcok: Fixed the typo.

docs/DafnyRef/Specifications.md
stefan-aws: Yes, thank you.

docs/DafnyRef/Specifications.md
davidcok: f is a function,. f.reads is also a function. It takes the same arguments as f, but returns the set of objects read. Similarly f.requires takes the same arguments as f but returns a boolean (wither the precondition is satisfied or not).

The use case is to be able to include, say f.requires in a precondition of another funciton/method that has f as an argument.  So map(f: T->bool, s: seq<T>): seq<bool> might have forall t: t in s :: f.requires(t) as a precondition.

davidcok: Change the int to T

stefan-aws (dismissed):

stefan-aws (dismissed):

docs/DafnyRef/GrammarDetails.md
```diff
@@ -93,7 +93,7 @@ These definitions use
 * curly braces to indicate 0-or-more repetitions
 * parentheses to indicate grouping
 * a `-` sign to indicate set difference: any character sequence matched by the left operand except character sequences matched by the right operand
-* a sequence of any of the above indicates concatentation (without whitespace)
+* a sequence of any of the above to indicate concatentation without whitespace
```
stefan-aws, line 96:
```suggestion
* a sequence of any of the above to indicate concatenation without whitespace
```

stefan-aws (approved):

PR #3429, opened 2023-01-31
Chore: re-added the --stdin option

MikaelMayer:
In preparation for the `dafny format` command to support `--stdin`, I re-added this option that was previously not ported.


davidcok (dismissed):

davidcok:
After I approved, I realized that there ought to be a test of --stdin that puts out an error message, to be sure that the filename in the error message is sensible (does not crash).

davidcok (approved):

PR #3444, opened 2023-02-02
Report PropagateFailure/Extract methods as being deprecated

### LINKED ISSUE #2693: Unexpected PropagateFailure/Extract side effects
Dafny's error mechanism `:-` makes use of three user-defined members: `IsFailure`, `PropagateFailure`, and `Extract`. I think we made a design error when we allowed the latter two to be methods. I suggest we change the design to insist that these 3 members (or 2, in the case where `Extract` is not present) be functions.

The problem with the current design is that it allows one statement to have more than one round of side effects. For example, if `M()` is a method with side effects, then the statement

``` dafny
x :- M();
```

will (first round) have the side effects of `M` and and then (second round) the side effects of either `PropagateFailure()` or `Extract()`. Having more than one round of side effects is more taxing for a human, and it's especially insidious because the `PropagateFailure()` and `Extract()` calls are not directly visible in the program text.

## Example

To further illustrate the issue, here is a complete example. Two lines in method `Test()` are marked with `// surprise`. If you read only method `Test()`, you'd expect the conditions on those two lines to hold. To figure out what's going wrong, you need to figure out that `PropagateFailure()` and `Extract()` can have side effects. Along the way, you may also be further confused, thinking that the side effects of `IncreaseAndReturn` have something to do with the two surprises (which they do not).

``` dafny
method Test() returns (result: Result)
  ensures result.Bad? ==> result.c.IsFailure()
{
  var c := new EvenGood_OddBad;
  c.a := 0;
  var x :- IncreaseAndReturn(c);
  assert !c.IsFailure();
  return Good(x);
}

class EvenGood_OddBad {
  var a: int

  function method IsFailure(): bool
    reads this
  {
    a % 2 == 1
  }

  method PropagateFailure() returns (r: Result)
    requires IsFailure()
    modifies this
  {
    a := 0;
    return Bad(this);
  }

  method Extract() returns (x: int)
    requires !IsFailure()
    modifies this
  {
    x := a;
    a := 19;
  }
}

datatype Result = Good(x: int) | Bad(c: EvenGood_OddBad)

method IncreaseAndReturn(c: EvenGood_OddBad) returns (r: EvenGood_OddBad)
  modifies c
  ensures old(c.a) <= c.a && r == c
{
  var n: nat;
  c.a := c.a + n;
  return c;
}
```

RustanLeino:
Fixes #2693

This PR generates a warning if the `PropagateFailure` or `Extract` member involved in a supposed failure-compatible type is a method. A future PR (at or after Dafny version 4) will remove the support.

So, with this PR, the `IsFailure`, `PropagateFailure`, and `Extract` members must all be functions.

The PR also removes the previous limitation that `IsFailure` or `PropagateFailure` be ghost. Not only was this check done quite early during resolution (before other ghost checking), but (more to the point) it is fine to allow these behaviors (for example, in ghost methods). The cost is that the errors generated may not be as direct, since they will be complaining about automatically generated code.


Source/DafnyCore/Resolver/Resolver.cs
```diff
@@ -6426,37 +6426,47 @@ public class ResolveTypeOption {
...
         }
       }
 
+      void checkIsFunction([CanBeNull] MemberDecl memberDecl, bool allowMethod) {
+        if (memberDecl == null || memberDecl is Function) {
+          // fine
+        } else if (allowMethod && memberDecl is Method) {
+          // give a deprecation warning, so we will remove this language feature around the Dafny 4 time frame
+          origReporter.Deprecated(MessageSource.Resolver, tok,
+            $"Support for member '{memberDecl.Name}' in type '{tp}' (used indirectly via a :- statement) being a method is being deprecated;" +
```
davidcok, line 6455:
```suggestion
            $"Support for member '{memberDecl.Name}' in type '{tp}' (used indirectly via a :- statement) being a method is deprecated;" +
```

Test/git-issues/git-issue-2693.dfy.expect
```diff
@@ -0,0 +1,8 @@
+git-issue-2693.dfy(10,10): Warning: Support for member 'PropagateFailure' in type 'EvenGood_OddBad?' (used indirectly via a :- statement) being a method is being deprecated; declare it to be a function instead
+git-issue-2693.dfy(10,10): Warning: Support for member 'Extract' in type 'EvenGood_OddBad?' (used indirectly via a :- statement) being a method is being deprecated; declare it to be a function instead
```
davidcok, line 2:
```suggestion
git-issue-2693.dfy(10,10): Warning: Support for member 'Extract' in type 'EvenGood_OddBad?' (used indirectly via a :- statement) being a method is deprecated; declare it to be a function instead
```

Test/git-issues/git-issue-859a.dfy.expect
```diff
@@ -1,7 +1,7 @@
...
-git-issue-859a.dfy(33,2): Error: The Extract member may not be ghost unless the initial LHS is ghost
-git-issue-859a.dfy(44,9): Error: The IsFailure member may not be ghost (type FailureCompatible used in :- statement)
-git-issue-859a.dfy(44,9): Error: The PropagateFailure member may not be ghost (type FailureCompatible used in :- statement)
+git-issue-859a.dfy(34,4): Error: The Extract member may not be ghost unless the initial LHS is ghost
+git-issue-859a.dfy(44,7): Error: a call to a ghost function is allowed only in specification contexts (consider declaring the function with 'function method')
+git-issue-859a.dfy(55,4): Error: return statement is not allowed in this context (because it is guarded by a specification-only expression)
+git-issue-859a.dfy(82,10): Error: Member 'IsFailure' in type 'NotFunctionsAtAll' (used indirectly via a :- statement) is expected to be a function
+git-issue-859a.dfy(82,10): Error: Member 'PropagateFailure' in type 'NotFunctionsAtAll' (used indirectly via a :- statement) is expected to be a function
+git-issue-859a.dfy(82,10): Error: Member 'Extract' in type 'NotFunctionsAtAll' (used indirectly via a :- statement) is expected to be a function
```
davidcok, line 6: The error messages are so different -- are we actually testing the same functionality or error conditions?

davidcok (changes requested):
A few moderately minor matters.

Test/git-issues/git-issue-2693.dfy
```diff
@@ -0,0 +1,80 @@
...
+    c.a := 0;
+    var x :- IncreaseAndReturn(c); // error: postcondition failure; deprecation warnings: PropagateFailure/Extract are methods
+    assert !c.IsFailure(); // error: assertion failure
+    return Good(x);
+  }
+
+  class EvenGood_OddBad {
+    var a: int
+
+    function method IsFailure(): bool
```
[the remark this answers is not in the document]
RustanLeino: Done

Test/git-issues/git-issue-859a.dfy.expect
RustanLeino: I had rewritten this test file substantially, and hence the expected output is also quite different.

Test/git-issues/git-issue-1094.dfy
```diff
@@ -98,11 +98,10 @@ datatype MStatus = Okay | Error(code: int) {
   predicate method IsFailure() {
     Error?
   }
-  method PropagateFailure() returns (m: MStatus)
+  function method PropagateFailure(): MStatus
```
davidcok, line 101: Please make this Dafny 4 ready. Perhaps change the syntax and putting an option in the RUN command

Test/git-issues/git-issue-1094.dfy
```diff
@@ -132,15 +131,14 @@ datatype MResult<X> = Success(x: X) | Failure(code: int) {
   predicate method IsFailure() {
     Failure?
   }
-  method PropagateFailure<U>() returns (result: MResult<U>)
+  function method PropagateFailure<U>(): MResult<U>
```
davidcok, line 134: Here too

Test/git-issues/git-issue-1094.dfy
```diff
@@ -132,15 +131,14 @@ datatype MResult<X> = Success(x: X) | Failure(code: int) {
...
     requires Failure?
-    ensures result.Failure? && result.code == code
   {
-    return MResult.Failure(code);
+    MResult.Failure(code)
   }
-  method Extract() returns (result: X)
+  function method Extract(): X
```
davidcok, line 139: And here

Test/git-issues/git-issue-817.dfy
```diff
@@ -3,7 +3,7 @@
 datatype Result<T> = Failure(msg: string) | Success(value: T) {
   predicate method IsFailure() { Failure? }
   function method PropagateFailure(): Result<T> requires IsFailure() { this }
-  method Extract() returns (t: T) requires !IsFailure() ensures t == this.value { return this.value; }
+  function method Extract(): T requires !IsFailure() { this.value }
```
davidcok, line 6: And here

Test/git-issues/git-issue-817a.dfy
```diff
@@ -3,7 +3,7 @@
 datatype Result<T> = Failure(msg: string) | Success(value: T) {
   predicate method IsFailure() { Failure? }
   function method PropagateFailure(): Result<T> requires IsFailure() { this }
-  method Extract() returns (t: T) requires !IsFailure() ensures t == this.value { return this.value; }
+  function method Extract(): T requires !IsFailure() { this.value }
```
davidcok, line 6: And here, etc.

Test/git-issues/git-issue-817b.dfy
```diff
@@ -4,7 +4,7 @@
 datatype Result<T> = Failure(msg: string) | Success(value: T) {
   predicate method IsFailure() { Failure? }
   function method PropagateFailure(): Result<T> requires IsFailure() { this }
-  method Extract() returns (t: T) requires !IsFailure() ensures t == this.value { return this.value; }
+  function method Extract(): T requires !IsFailure() { this.value }
```
davidcok, line 7: And here

Test/git-issues/git-issue-859a.dfy.expect
davidcok: OK

davidcok (changes requested):
I still think you should make everything dafny4 ready by adding options to the commadline and changing function method to function. (and function to ghost function if that is part of the test)

RustanLeino:
I changed the test files I touch into Dafny 4 syntax, except the three tests that use `RUN %testDafnyForEachCompiler`, because that command doesn't seem to want any additional parameters.

davidcok (approved):
OK - we'll have to deal with the %testDafnyForEachCompiler all at once at some future time.

Not in this PR, but I'd also suggest that in altering tests in the future, move toward using the new command syntax.

Source/DafnyCore/Resolver/Resolver.cs
```diff
@@ -6427,37 +6427,47 @@ public class ResolveTypeOption {
...
             tp, expectExtract ? "and" : "but not");
         }
       }
 
+      void checkIsFunction([CanBeNull] MemberDecl memberDecl, bool allowMethod) {
+        if (memberDecl == null || memberDecl is Function) {
+          // fine
+        } else if (allowMethod && memberDecl is Method) {
+          // give a deprecation warning, so we will remove this language feature around the Dafny 4 time frame
+          origReporter.Deprecated(MessageSource.Resolver, tok,
```
davidcok, line 6455: Youi will need to sprinkle 'ErrorID.None' into these Deprecated calls, here and elsewhere

davidcok (changes requested):

davidcok (changes requested):
I spotted three files that have function methods in them, If you want to leave them alone as Dafny 3 syntax, thats fine -- I'll catch them anyway in doing the Dafny 4 test dry runs. But if they are just an oversight, please change them, as the tests have to run again anyway.

davidcok (approved):
If the changes I suggest aren't made before this auto-merges, they will be part of the Dafny4 modifications of the test suite anyway.

PR #3450, opened 2023-02-02
fix: Exclude type information for “new object” allocations

### LINKED ISSUE #3449: new object should give object of arbitrary type
### Dafny version

3.11

```dafny
class Basic { }

method Test()
{
  var obj: object := new object;
  if * {
    assert !(obj is Basic); // should fail, but currently verifies
  } else {
    assert !(obj is array<int>); // should fail, but currently verifies
  }
}
```

```code
> dafny test.dfy
```

### What happened?

The reference manual says that `new object` "allocates a new object (of an unspecified class type)". Alas, the verifier assigns it a very specific type, which is not equal to declared classes in the program. The proper behavior is for the verifier not to assume any information about the dynamic type of the object being allocated.

So, the program above should give an error for each of the assertions.

Mac

RustanLeino:
Fixes #3449

This PR fixes the verifier so that it does not assume any specific dynamic type of the object allocated by `new object`. This is a special allocation that is supposed to allocate an object of an arbitrary type. (See section 14.1 of the reference manual.)

The PR also improves the source location reported when a program attempts `new X` where `X` is a non-`class`, non-`object` type.

### DIFF
--- a/Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
+++ b/Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
@@ -5386,7 +5386,7 @@ namespace Microsoft.Dafny {
             if (cl != null && !(rr.EType.IsTraitType && !rr.EType.NormalizeExpand().IsObjectQ)) {
               // life is good
             } else {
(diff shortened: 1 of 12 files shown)


davidcok (approved):
I might have put the error message on the type rather than on the new, but its fine on the new.

jtristan (approved):

PR #3457, opened 2023-02-04
FAQs from user assistance discussions

davidcok:


docs/HowToFAQ/FAQThisFrameField.md
```diff
@@ -0,0 +1,25 @@
...
+trait Test {
+  var y: int
+  var z: int
+  predicate readsY() reads this`y { true }
+  predicate readsZ() reads this`z { true }
+  function {:opaque} MyFunc(x: bool): int
+    reads if x then readsY.reads() else readsZ.reads() {
+    if x then y else z
+  }
+}
```
MikaelMayer, line 23: @RustanLeino told me that actually the function reads() itself overapproximate in this case :-(
Fortunately, there is another workaround.

```suggestion
trait Test {
  var y: int
  var z: int
  function {:opaque} MyFunc(x: bool): int
    reads (if x then {this} else {})`y, (if x then {} else {this})`z {
    if x then y else z
  }
}
```

MikaelMayer (changes requested):
Fantastic ! It's a super well formatted version of our discussions. I just tested the second workaround I suggested, and it works, so you can use it.

MikaelMayer (approved):

PR #3467, opened 2023-02-06
Chore: Fixed flacky test that had concurrent issues

MikaelMayer:
Fixes a CI issue where a test was sometimes failing randomly.

The reason was, there was a shared variable that was mutated by three different threads. I converted it to a concurrent dictionary, and everything works smoothly for even 10 threads.

### DIFF
--- a/Source/DafnyLanguageServer.Test/GutterStatus/LinearVerificationGutterStatusTester.cs
+++ b/Source/DafnyLanguageServer.Test/GutterStatus/LinearVerificationGutterStatusTester.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


davidcok (approved):

Source/DafnyLanguageServer.Test/GutterStatus/LinearVerificationGutterStatusTester.cs
```diff
@@ -141,7 +143,10 @@ var previousPerLineDiagnostics
       previousPerLineDiagnostics = newPerLineDiagnostics;
     }
 
-    previousTraces = traces;
+    previousTraces.AddOrUpdate(verificationStatusGutterReceiver,
+      oldKey => traces,
+      (oldKey, oldTraces) => traces
```
robin-aws, line 148:
```suggestion
      (oldKey, _) => traces
```

robin-aws (commented):
Looks like at least a necessary change, but you also hit a failure in CI still, so perhaps not yet sufficient?

PR #3477, opened 2023-02-07
Chore: Link from dafny.org to blogs

MikaelMayer:
This PR adds a link for the website to display the blog.

On `https://dafny.org/`, the iframe will be replaced by the different blog links dynamically.
On every other website (http -only, localhost), the iframe will be displayed, because of the CORS policy.


davidcok (approved):
Looks OK, but I'm no so familair with web programming, so we'll see if it works once pmerged.

PR #3479, opened 2023-02-08
fix: Override checks with `canCall` precondition 

### LINKED ISSUE #2500: {:termination false} also allows incorrect implementations of trait functions

fabiomadge:
We strengthen this check by performing it on the visibility level of the implementing function, but restricting the enabled axiom to the callers of the trait. To achieve this, we add the canCall precondition to the axiom. Additionally, we consider imported trait functions when computing the visibility levels.

Fixes #2500.


RustanLeino:
So far, I have inspected the Boogie being generated, and it looks good!

RustanLeino (approved):
Nice work!

Issue #3066, opened 2022-11-17
Run the integration tests in a single process to detect usages of static state


keyboardDrummer:
The language server runs compilation multiple times using a single process. To test that such repeated executions in a single process work well together, it would be good to also do this while running integration tests.

Related issue: https://github.com/dafny-lang/dafny/issues/2945

Issue #632, opened 2020-05-26
Run all applicable tests for all target languages


robin-aws:
Tests that execute programs currently specify some subset of the target languages in their lit headers. It's very difficult to determine what features are supported for each language. Even if some tests don't currently pass for all languages, there should be metadata in the project somewhere that explicitly documents they need to be skipped.

I'm not sure how well lit will support this, so we may want to consider other testing frameworks for at least this subset of tests.

robin-aws:
This should also hopefully provide a better way to test lots of source files with and without `/allocated1`. This is currently handled by duplicating many test files with a single include statement.

RustanLeino:
The way to accomplish this with the current test configuration has the following form:

    lit -D dafny_params=/compileTarget:go dafny0/Compilation.dfy

To run the whole test suite (parsing tests, resolution tests, verification tests, compilation tests for all targets in the `comp` directory, and compilation tests for C# everywhere else), do:

    lit .

as usual. Then, to repeat the compilation tests for Java, JavaScript, and Go (except the ones in the `comp` directory that were already done above), do:

    for target in java js go ; do lit -D dafny_params=/compileTarget:$target `find . -iname '*.dfy' | grep -v "/comp/" | xargs grep -l "RUN:.*compile:[1234]"` ; done

davidcok:
An additional improvement to the test system would be a means of knowing which language is failing -- currently not easy to determine.  Also to run a subset of target languages for situations in which not all targets are supported on a development platform.

robin-aws:
Agreed. My plan is for tests that are valid Dafny programs to generate an independent test case for each target language. Each can pass, fail, or create an expected failure/skip result as appropriate, and the third case can be used to track gaps in target language support.

I have a POC working that uses xUnit, which has good support for such parameterized tests and great UI in IDEs like Rider. It looks like I'll have to do a bit of plugin work to match the parallelism we're currently getting from lit though (both across threads and across multiple hosts via sharding).

robin-aws:
We now have the `%testDafnyForEachCompiler` lit command designed for this, the remaining work is just to work through the test suite and use it. We may first want to rethink the syntax for this command(s) to align better with the new CLI and the `%run`/`%verify`/etc. lit commands we have now.

[...]

PR #3486, opened 2023-02-08
Automated test coverage analysis

atomb:
Run unit and integration tests with coverage analysis enabled and upload the results in the test results artifact, both in Cobertura and marked-up HTML source. Also prints a summary of those coverage results in the logs.

There's one key improvement I'd like to make, to add the coverage summary summary as a comment on PRs, as shown [here](https://github.com/marketplace/actions/code-coverage-summary). Doing that requires figuring out some permission changes.

However, the results are useful, as-is, so I think it makes sense to merge sooner and implement those enhancements later.

To fully view the results, download the `test-coverage-results` artifact from the `test-coverage-analysis` job and look in the `coverage-html` directory. Or, if you want to re-generate the HTML, or another coverage summary format, use the `reportgenerator` tool on the files in `coverage-cobertura`.

Note that, for some reason, the `DafnyLanguageServer.Test` package only gets good results when run with `dotnet coverage` as opposed to the `dotnet test --collect`, and these results are in a different format. This output is automatically converted to the Cobertura format as part of the CI job, though, and merged with the rest.


.github/workflows/msbuild.yml
```diff
@@ -66,3 +71,65 @@ jobs:
...
+        uses: actions/download-artifact@v3
+        with:
+          name: integration-test-results-ubuntu-20.04
+
+      - name: Download unit test artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: unit-test-results-ubuntu-20.04
+
+      - name: Show files
```
davidcok, line 105: This looks like debugging code. Do you want it long term?

.github/workflows/xunit-tests.yml
```diff
@@ -16,15 +10,12 @@ concurrency:
 ##  chmod - whether the Z3 download needs execute permissions added
 ##  coverage -- unused?
```
davidcok, line 11: Is this still unused?

davidcok: Actually I see it is now set in the matrix. but I can't see where it is used. It appears that coverage testing is run on every platform.

davidcok (changes requested):
Some general questions and comments.

.github/workflows/xunit-tests.yml
atomb: Yeah, that was there before, and unused, and when I added coverage testing I made it unconditional (which doesn't seem to be a problem). I'll remove that parameter.

.github/workflows/msbuild.yml
atomb: Probably not, now that you mention it. :)

robin-aws (changes requested):
Love that we're finally adding this! Some top-level thoughts:

* I'm a bit concerned about how much this seems to slow down the tests. Based on a quick comparison to a PR from a few days ago, the slowest shard on ubuntu went from 20m to 26m, for example. That's not a reason to block this PR, but I think we'd better prioritize optimizing how we run the integration tests via things like https://github.com/dafny-lang/dafny/issues/632#issuecomment-1442115909 and/or https://github.com/dafny-lang/dafny/issues/3066.
* It looks like the CI on this PR marked some scopes as below the threshold, but the checks were still green?
* What the heck is going on with DafnyRuntime? :) It's coverage seems way lower than it should be, but it isn't 0% so it's half-working somehow?

robin-aws:
Another related top-level thought: https://github.com/dafny-lang/dafny/issues/3696

.github/workflows/integration-tests-reusable.yml
```diff
@@ -131,18 +131,19 @@ jobs:
         DAFNY_RELEASE: ${{ github.workspace }}\unzippedRelease\dafny
       run: |
         cmd /c mklink D:\a\dafny\dafny\unzippedRelease\dafny\z3\bin\z3-4.8.5 D:\a\dafny\dafny\unzippedRelease\dafny\z3\bin\z3-4.8.5.exe
-        dotnet test --logger trx --logger "console;verbosity=normal" dafny/Source/IntegrationTests
+        dotnet test --logger trx --collect:"XPlat Code Coverage" dafny/Source/IntegrationTests/IntegrationTests.csproj
```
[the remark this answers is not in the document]
atomb: Ah, yes, I think so.

.github/workflows/msbuild.yml
```diff
@@ -66,3 +71,65 @@ jobs:
...
+        run: ls -lR
+
+      - name: Merge reports
+        run: |
+          dotnet coverage merge DafnyLanguageServer.Test.coverage -o DafnyLanguageServer.Test.cobertura.xml -f cobertura
+          reportgenerator \
+            -reports:"./**/coverage.cobertura.xml;*.cobertura.xml" \
+            -assemblyfilters:"+AutoExtern;+dafnylanguageserver.dll;+Dafny;+DafnyCore;+DafnyDriver;+DafnyRuntime;+DafnyTestGeneration" \
```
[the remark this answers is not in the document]
atomb: I think it, or something like it, is necessary. Otherwise, it includes various system assemblies in the report, dropping coverage significantly and cluttering up the report.

.github/workflows/msbuild.yml
```diff
@@ -66,3 +71,65 @@ jobs:
...
+        with:
+          filename: ./coverage-cobertura/Cobertura.xml
+          badge: true
+          fail_below_min: true
+          format: markdown
+          hide_branch_rate: false
+          hide_complexity: true
+          indicators: true
+          output: both
+          thresholds: '86 90'
```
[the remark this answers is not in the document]
atomb: Those are the bars for "yellow" and "green" badges. I'll add a comment to that effect. Anything below 86 will be rejected. Anything below 90 (though we'll probably want to adjust that) will get a yellow badge (if I get badges on PRs working!).

atomb:
> Another related top-level thought: #3696

Yeah, I like that idea.

atomb:
Strangely, when I added `run-deep-tests`, the coverage went from 87% to 55%. I'm not sure what's going on there.

atomb:
>     * I'm a bit concerned about how much this seems to slow down the tests. Based on a quick comparison to a PR from a few days ago, the slowest shard on ubuntu went from 20m to 26m, for example. That's not a reason to block this PR, but I think we'd better prioritize optimizing how we run the integration tests via things like [#632 (comment)](https://github.com/dafny-lang/dafny/issues/632#issuecomment-1442115909) and/or #3066.

Yeah, I'd love to put either of those in place, though probably as part of a separate PR.

>     * It looks like the CI on this PR marked some scopes as below the threshold, but the checks were still green?

I think the threshold applies only to the overall total. We could probably generate a separate report for each package if we wanted to constrain them, as well.

>     * What the heck is going on with DafnyRuntime? :) It's coverage seems way lower than it should be, but it isn't 0% so it's half-working somehow?

There's only one test in `DafnyRuntime.Tests`, for the `BigRational`/`double` conversions I added a while back. I think that's what it's capturing. Are there `DafnyRuntime` tests elsewhere? I don't think the .NET infrastructure is smart enough to trace Dafny programs that are compiled and then run by the test suite, unfortunately (although it does trace `dafny` sub-processes created immediately by the test suite).

.github/workflows/integration-tests-reusable.yml
```diff
@@ -131,18 +131,19 @@ jobs:
         DAFNY_RELEASE: ${{ github.workspace }}\unzippedRelease\dafny
       run: |
         cmd /c mklink D:\a\dafny\dafny\unzippedRelease\dafny\z3\bin\z3-4.8.5 D:\a\dafny\dafny\unzippedRelease\dafny\z3\bin\z3-4.8.5.exe
-        dotnet test --logger trx --logger "console;verbosity=normal" dafny/Source/IntegrationTests
+        dotnet test --logger trx --logger "console;verbosity=normal" --collect:"XPlat Code Coverage" dafny/Source/IntegrationTests/IntegrationTests.csproj
```
keyboardDrummer, line 134: Does this actually measure coverage of the integration tests, given that most of the code is run in separate processes and not the main testing process?

keyboardDrummer (dismissed):
Awesome, thanks :)

keyboardDrummer:
> To fully view the results, download the `test-coverage-results` artifact from the `test-coverage-analysis` job and look in the `coverage-html` directory. Or, if you want to re-generate the HTML, or another coverage summary format, use the `reportgenerator` tool on the files in `coverage-cobertura`.

I found the summary in the logs, but I wasn't able to figure out how to find or download these artifacts, despite trying to follow GH documentation: https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts

.github/workflows/integration-tests-reusable.yml
atomb: I was impressed to find that, yes, it does. We get over 80% coverage of DafnyCore just from this one `dotnet test` command.

atomb:
> > To fully view the results, download the `test-coverage-results` artifact from the `test-coverage-analysis` job and look in the `coverage-html` directory. Or, if you want to re-generate the HTML, or another coverage summary format, use the `reportgenerator` tool on the files in `coverage-cobertura`.
> 
> I found the summary in the logs, but I wasn't able to figure out how to find or download these artifacts, despite trying to follow GH documentation: https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts

I think that's because the run failed, for two reasons: 1) due to the failing compilation on Windows, 2) due to the fact that `enable-deep-tests` caused the coverage to drop below the allowed level. I could probably set the upload step to always run.

atomb:
I've discovered why `run-deep-tests` results in less coverage: the integration tests run in a different way when that label is set, and that results in `dotnet test` not being able to track coverage. So, I've decided to make this job run only on normal PRs (without the `run-deep-tests` label) for now. That will allow it to serve its key purpose of gating PRs on test coverage. Later we may be able to refactor how we run integration tests to avoid the problem.

davidcok (dismissed):
Looking forward to evaluating these results -- would like a clearer guide to how to view them

.github/workflows/msbuild.yml
```diff
@@ -67,3 +72,109 @@ jobs:
       # This is the best way to fix an issue in master that was only caught by the nightly build.
       all_platforms: ${{ contains(github.event.pull_request.labels.*.name, 'run-deep-tests') || contains(github.event.push.labels.*.name, 'run-deep-tests')}}
       num_shards: 5
+
+  test-coverage-analysis:
+    runs-on: ubuntu-20.04
+    # Don't include in run-deep-tests, because the way it currently runs
+    # messes with integration test coverage tracking. Hopefully running
+    # integration tests as xUnit tests will help avoid this issue.
```
robin-aws, line 80: `run-deep-tests` still runs them as xUnit tests FYI, the issue is probably related to running from a packaged release of Dafny instead of directly on the source projects.

atomb: Yes, I think you're right, that the issue is related to running from a packaged release. And I misstated my comment: I was really referring to the change of running them all within the same process. I'll update the comment as part of a merge with `master`.

robin-aws (approved):

davidcok (approved):

.github/workflows/xunit-tests-reusable.yml
```diff
@@ -34,29 +24,29 @@ jobs:
...
         submodules: recursive
     - name: Setup .NET
       uses: actions/setup-dotnet@v3
       with:
         dotnet-version: 6.0.x
     - name: Install dependencies
-      run: dotnet restore ${{env.solutionPath}}
+      run: |
+        dotnet restore ${{env.solutionPath}}
+        dotnet tool install dotnet-coverage
```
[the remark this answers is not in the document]
atomb: Perhaps so, yes. That'd make it more seamless to run locally.

PR #3488, opened 2023-02-08
Fix: Gutter icons not in sync with the diagnostics

### LINKED ISSUE #3487: Gutter icons and status bar not in sync
Failing code
------------

Cannot reproduce on a simple file

Steps to reproduce the issue
----------------------------

* Dafny version: 3.10.0.41215
* Dafny VSCode extension version: (local) https://github.com/dafny-lang/ide-vscode/commit/624e4c948c30b3eab4d1e0620a27f17f6dcfb093

Expected behavior
-----------------

When verification finishes, lines with an error are always marked in red, and the status bar should show "Verification failed"

Actual behavior
---------------
I added and removed a newline 32 times for these two lines, and the issue appeared 15 times, roughly one time over 2.
image

The following always happens, no matter what (EDIT: It's because I saved this into a file. File has no space no special character, it's on Windows)
image

MikaelMayer:
This PR makes the test `EnsuresManyDocumentsCanBeVerifiedAtOnce` stable and fixes #3487 at the same time. This is not something I think I can readily write a test for, because it's a race issue. I tested locally that the error was gone when it was there before, when clicking "run this test until failure" I explain the solution in the code as well because of the lack of tests


davidcok (approved):
I'll trust you (with verification to come in CI:-)

PR #3491, opened 2023-02-08
Fix canonical paths for paths with spaces

MikaelMayer:
After testing today's version of Dafny in VSCode, I realized that a path with name triggered an error when attempting to resolve includes. Hence, I added a unit test to catch this regression for next time and fixed the method that transforms file paths (which can be gross URI) to actual file locations

### DIFF
--- a/Source/DafnyCore/DafnyMain.cs
+++ b/Source/DafnyCore/DafnyMain.cs
@@ -43,23 +43,19 @@ namespace Microsoft.Dafny {
         return (new Uri(filePath)).LocalPath;
       }
 
-      if (filePath.StartsWith("file:\\")) {
-        var withoutPrefix = filePath.Substring("file:\\".Length);
-        var tentativeURI = "file:///" + withoutPrefix.Replace("\\", "/");
-        if (Uri.IsWellFormedUriString(tentativeURI, UriKind.RelativeOrAbsolute)) {
-          return (new Uri(tentativeURI)).LocalPath;
+      var potentialPrefixes = new List<string>() { "file:\\", "file:/", "file:" };
+      foreach (var potentialPrefix in potentialPrefixes) {
+        if (filePath.StartsWith(potentialPrefix)) {
+          var withoutPrefix = filePath.Substring(potentialPrefix.Length);
+          var tentativeURI = "file:///" + withoutPrefix.Replace("\\", "/");
+          if (Uri.IsWellFormedUriString(tentativeURI, UriKind.RelativeOrAbsolute)) {
+            return (new Uri(tentativeURI)).LocalPath;
+          }
+          // Recovery mechanisms for the language server
+          return filePath.Substring(potentialPrefix.Length);
         }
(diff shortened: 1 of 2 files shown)


Source/DafnyPipeline.Test/RelativePaths.cs
```diff
@@ -10,5 +10,14 @@ public class RelativePaths {
     public void Test() {
       Assert.Equal(0, DafnyDriver.ThreadMain(new string[] { "/spillTargetCode:3", "warnings-as-errors.dfy" }));
     }
+
+    [Fact]
+    public void TestFileConversion() {
```
alex-chew, line 15:
```suggestion
    public void TestFileCanonicalization() {
```

Source/DafnyPipeline.Test/RelativePaths.cs
```diff
@@ -10,5 +10,14 @@ public class RelativePaths {
...
+      var filePath = @"file:\c:\Users\xxx\Documents\with%20space\test.dfy";
+      var filePath1 = @"file:c:\Users\xxx\Documents\with%20space\test.dfy";
+      var expected = @"c:\Users\xxx\Documents\with space\test.dfy";
```
alex-chew, line 18: Please add corresponding tests for [POSIX paths](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_266) too.

alex-chew (changes requested):

alex-chew (approved):

Source/DafnyCore/DafnyMain.cs
```diff
@@ -43,11 +43,18 @@ public class DafnyFile {
         return (new Uri(filePath)).LocalPath;
       }
 
-      if (filePath.StartsWith("file:\\")) {
-        // Recovery mechanisms for the language server
-        return filePath.Substring("file:\\".Length);
+      var potentialPrefixes = new List<string>() { "file:\\", "file:/", "file:" };
```
keyboardDrummer, line 46: I'm surprised we need custom code to handle file paths. Why can't `Uri` help us with all our path parsing and canonicalisation needs?

MikaelMayer: I'm surprised too.
There is some history about that. In the formatter, I had to switch to plain file URIs for token's filenames, because otherwise it would not be consistent with VSCode URIs. It looks like VSCode's URIs are slightly different than LSP URI, for example, a period in a disk name is encoded on one and not the other.

I wish there is a better solution but for now it's urgent we have a fix. I remember having spent a lot of time previously to ensure everything was working acceptably on the CLI and the language server (including unsaved files, and stdin), both on Windows and Linux.
Yesterday was the first time I tested the code for paths with spaces, and that's probably why it wasn't caught before. After this fix is pushed, feel free to experiment with different alternatives.

keyboardDrummer (approved):

PR #3492, opened 2023-02-08
Fixes #3489 - apparently incorrect code

### LINKED ISSUE #3489: incorrect call of cloning in Cloner.cs
### Dafny version

3.11.0+dev

```dafny
This code looks incorrect to Cok & Leino

      if (cf != null && cf.Rhs != null && !RevealedInScope(f)) {
        // We erase the RHS value. While we do that, we must also make sure the declaration does have a type, so instead of
        // cloning cf.Type, we assume "f" has been resolved and clone cf.Type.NormalizeExpandKeepConstraints().
        return new ConstantField(Tok(cf.tok), cf.Name, null, cf.IsStatic, cf.IsGhost, CloneType(cf.Type.NormalizeExpandKeepConstraints()), CloneAttributes(cf.Attributes));
      }
```

_No response_

### What happened?

The cf.IsStatic should be cf.HasStaticKeyword

though I haven't been able to create a test case that is incorrect as a result.

Mac

davidcok:
Fixes #3489


Source/DafnyCore/AST/Cloner.cs
```diff
@@ -815,7 +815,7 @@ class ScopeCloner : DeepModuleSignatureCloner {
       if (cf != null && cf.Rhs != null && !RevealedInScope(f)) {
         // We erase the RHS value. While we do that, we must also make sure the declaration does have a type, so instead of
         // cloning cf.Type, we assume "f" has been resolved and clone cf.Type.NormalizeExpandKeepConstraints().
-        return new ConstantField(Tok(cf.tok), cf.Name, null, cf.IsStatic, cf.IsGhost, cf.IsOpaque, CloneType(cf.Type.NormalizeExpandKeepConstraints()), CloneAttributes(cf.Attributes));
+        return new ConstantField(Tok(cf.tok), cf.Name, null, cf.HasStaticKeyword, cf.IsGhost, cf.IsOpaque, CloneType(cf.Type.NormalizeExpandKeepConstraints()), CloneAttributes(cf.Attributes));
```
MikaelMayer, line 818: Is there a test that could catch this? I'm worrying that these two fields are too similar.

davidcok: I have not thought of or found one. IsStatic is computed as HasStaticKeyword or "is static by default".
So a problem would come when a const  declaration was not declared static, but is static by default, and
then the AST is cloned, and the HasStaticKeyword filed is used for something directly.
(This is a reason I'm not fond of C#'s computed properties)

davidcok: Then please approve it. But if you can tell me under what conditions such a cloine is made and for what purpose, I'll keep thinking about a test case.

RustanLeino (approved):

PR #3493, opened 2023-02-08
Fix: Hovering with information about foreign docs

MikaelMayer:
This PR fixes a LSP crash when hovering an assertion which refers to an expression outside of the current doc

If this PR fails, it's possible that it requires https://github.com/dafny-lang/dafny/pull/3491 to be merged first


Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -258,7 +258,10 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
           // It's not necessary to restate the postcondition itself if the user is already hovering it
           // however, nested postconditions should be displayed
           if (errorToken is BoogieRangeToken rangeToken && !hoveringPostcondition) {
-            deltaInformation += "  \n" + CouldProveOrNotPrefix + ideState.TextDocumentItem.Text.Substring(rangeToken.StartToken.pos,
+            var originalText =
+              ideState.Uri == rangeToken.Filename ? ideState.TextDocumentItem.Text :
+              File.ReadAllText(rangeToken.ActualFilename);
```
atomb, line 263: It's probably worth waiting for a future PR, so that this issue can be fixed quickly, but it would be nice in the long run to cache the text of files already looked at so that the server doesn't need to re-read the file every time someone hovers.

atomb (approved):
The test is still failing, since it can't find the relevant Dafny file. Besides fixing that, and one note about a performance concern we may want to resolve in the future, I think this looks good.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: I agree. looks like it's going to be more complex than I thought though..

keyboardDrummer, line 263: I'd rather not add `File.ReadAllText` than being able to show the deltaInformation when the hover needs information from a file that's not `ideState.TextDocumentItem.Text`

I think the way this should work is: `deltaInformation += " 
  \n" + CouldProveOrNotPrefix + ((Program)program).FindNodeFromToken(rangeToken.StartToken).PrintOriginal()`

MikaelMayer: Why would you look for a node in the tree when the rangeToken contains everything?

However, you have a point: since the tokens contain everything, we could just go from start to end and print tokens. I'll implement this instead.

Source/DafnyLanguageServer.Test/Lookup/TestFiles/foreign-verify.dfy
```diff
@@ -0,0 +1,3 @@
+predicate P(i: int) {
```
atomb, line 1: Oh, yeah, the tests wouldn't pass without the file available. ;)

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -258,8 +258,8 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
           // It's not necessary to restate the postcondition itself if the user is already hovering it
           // however, nested postconditions should be displayed
           if (errorToken is BoogieRangeToken rangeToken && !hoveringPostcondition) {
-            deltaInformation += "  \n" + CouldProveOrNotPrefix + ideState.TextDocumentItem.Text.Substring(rangeToken.StartToken.pos,
-              rangeToken.EndToken.pos + rangeToken.EndToken.val.Length - rangeToken.StartToken.pos);
+            var originalText = rangeToken.PrintOriginal();
+            deltaInformation += "  \n" + CouldProveOrNotPrefix + originalText;
```
atomb, line 262: This is much nicer than reading the whole file again. :)

atomb (dismissed):

Source/DafnyCore/AST/Tokens.cs
```diff
@@ -211,6 +211,19 @@ public class BoogieRangeToken : TokenWrapper {
   public override IToken WithVal(string newVal) {
     return this;
   }
+
+  public string PrintOriginal() {
+    var tok = WrappedToken;
+    var s = tok.val;
```
keyboardDrummer, line 217: Could you use a `StringBuilder`, and longer variable names?

keyboardDrummer (commented):
Could you address this comment? https://github.com/dafny-lang/dafny/pull/3493/files#r1102689236

MikaelMayer:
> Could you address this comment? https://github.com/dafny-lang/dafny/pull/3493/files#r1102689236

Oh yeah thanks for letting me know again, I did not see this comment. Implemented.

keyboardDrummer (approved):

PR #3495, opened 2023-02-08
chore: Prepare to fix allocated 4

RustanLeino:
This PR contains some simple code improvements. The bigger context is that I'm working on a fix for `/allocated:4` and these changes were done on the way. I separated these changes into its own PR to make reviewing (of both this PR and the next one) easier. This PR does not change the behavior of Dafny (or at least that's my intention).


Source/DafnyCore/AST/Types/ArrowType.cs
```diff
@@ -43,6 +41,21 @@ public class ArrowType : UserDefinedType {
     }
   }
 
+  public static void ComputeResolvedOpcodeAndBoundedPool(Expression collection, CollectionType collectionType,
```
keyboardDrummer, line 44: Should the computation of bounded pools be tied more closely to types? Should we have `CollectionType.GetBoundedPool` or `Type.GetBoundedPool` ? I'm not sure why this code relates mostly to `ArrowType` and I see it's also used outside of `ArrowType`

Source/DafnyCore/AST/Types/ArrowType.cs
```diff
@@ -43,6 +41,21 @@ public class ArrowType : UserDefinedType {
...
+    out BinaryExpr.ResolvedOpcode resolvedOpcode, out ComprehensionExpr.BoundedPool boundedPool) {
+    if (collectionType is SetType setType) {
+      resolvedOpcode = BinaryExpr.ResolvedOpcode.InSet;
```
keyboardDrummer, line 47: Maybe this opCode can be tied to the BoundedPool ? `CollectionBoundedPool.ResolvedOpCode`.

Source/DafnyCore/Verifier/Translator.DataTypes.cs
```diff
@@ -673,6 +670,25 @@ partial class Translator {
...
+    /// Return list of variables
+    ///     d: DatatypeValue, T0,T1,...: Ty
+    /// expression
+    ///     d
+    /// expression
+    ///     T(T0,T1,...)
+    /// and predicate
+    ///     $Is(d, T(T0,T1,...))
+    /// </summary>
+    private void MkIsPredicateForDatatype(DatatypeDecl dt, out List<Bpl.Variable> boundVariables,
```
keyboardDrummer, line 683: Please use longer names.

Source/DafnyCore/Verifier/Translator.DataTypes.cs
```diff
@@ -673,6 +670,25 @@ partial class Translator {
       }
     }
 
+    /// <summary>
+    /// Return list of variables
+    ///     d: DatatypeValue, T0,T1,...: Ty
```
keyboardDrummer, line 675: Should `Ty` be `T` here?

RustanLeino: These comments (through the verifier) are both sketches (that is, they don't highlight every corner case) and (unfortunately) sometimes cryptic. Here, `Ty` is correct, because that's the name of the type we use in Boogie to denote Dafny type terms. The `T` that occurs a few lines down (in `T(T0,T1,...)`) is meant to denote the type to which `T0,T1,...` are type parameters.

Source/DafnyCore/Verifier/Translator.DataTypes.cs
RustanLeino: Done

Source/DafnyCore/AST/Types/ArrowType.cs
RustanLeino: In general, there may be other bounded pools associated with an `in` expression, too. So, I just kept them together like they were.

Source/DafnyCore/AST/Types/ArrowType.cs
RustanLeino: Eagle eyes! :)

I had been a bit unhappy about the placement of this method. Reflecting on it some more, I now think putting it near the various `Expression.Create...` methods, since it produces information that you need when creating an expression of the form `x in C`. So, I moved it there.

Btw, the `boundedPool` out-parameter of this method is not used by any caller in this PR, but it is used in my soon-to-come follow-up PR.

Source/DafnyCore/AST/Types/ArrowType.cs
keyboardDrummer: Sure, but you can still let the `BoundedPool` say which opcode should be used with it right?

Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -783,6 +783,26 @@ public abstract class Expression : TokenNode {
     return e;
   }
 
+  /// <summary>
+  /// For a given collection type "collectionType", return the ResolvedOpcode and appropriate BoundedPool
+  /// used in creating a resolved expression of the form "x in collection", where "collection" is given as
+  /// a parameter.
+  /// </summary>
+  public static void ComputeResolvedOpcodeAndBoundedPool(Expression collection, CollectionType collectionType,
```
keyboardDrummer, line 791: I would still change this signature to `CollectionBoundedPool CollectionType.ComputeBoundedPool(Expression collection)`, and add a property `ResolveOpCode CollectionBoundedPool.OpCode`. If you add a new CollectionType, don't you want the implementor to think about how to create a BoundedPool for it?

In general, I think up to and including the resolution phase, so that includes parsing printing and resolution, all code specific to an AST type should live inside its definition.

Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -783,6 +783,26 @@ public abstract class Expression : TokenNode {
...
+    out BinaryExpr.ResolvedOpcode resolvedOpcode, out ComprehensionExpr.BoundedPool boundedPool) {
+    if (collectionType is SetType setType) {
+      resolvedOpcode = BinaryExpr.ResolvedOpcode.InSet;
+      boundedPool = new ComprehensionExpr.SetBoundedPool(collection, collectionType.Arg, collectionType.Arg, setType.Finite);
+    } else if (collectionType is SeqType) {
+      resolvedOpcode = BinaryExpr.ResolvedOpcode.InSeq;
+      boundedPool = new ComprehensionExpr.SeqBoundedPool(collection, collectionType.Arg, collectionType.Arg);
+    } else {
```
keyboardDrummer, line 799: What about MapType? I see there is a `MapBoundedPool`

keyboardDrummer (approved):

Source/DafnyCore/AST/Expressions/Expressions.cs
RustanLeino: Okay, I created `CollectionType.GetBoundedPool` as you suggested. I also created a `ResolvedOpcodeForIn` property, but I place it in `CollectionType` as well (rather than in `CollectionBoundedPool`), since it could be useful outside the context of bounded pools, too.

keyboardDrummer (approved):

PR #3498, opened 2023-02-09
Adding more quick fixes

davidcok:


Test/README.md
```diff
@@ -25,6 +25,20 @@ dotnet test --logger "console;verbosity=normal" Source/IntegrationTests --filter
 [See here](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test) for more information about
 the `dotnet test` command and other supported options.
 
+## Writing tests
+
+Lit tests use a very restricted form of command-line: just commands, little piping, just > and >> redirection.
```
MikaelMayer, line 30:
```suggestion
Lit tests use a very restricted form of command-line: executables and their arguments piping, `>` and `>>` redirections (for respectively writing and appending stdout to a file) and `2>` and `2>>` redirections (for respectively writing and appending both stdout and stderr to a file).
```

Source/DafnyCore/AST/Tokens.cs
```diff
@@ -224,7 +224,6 @@ public class CodeActionToken : TokenWrapper {
   public String StartLength() {
     return (line - 1) + " " + (column - 1) + " " + length;
   }
-
   public override IToken WithVal(string newVal) {
```
MikaelMayer, line 227: I think the newline should be kept here
```suggestion

  public override IToken WithVal(string newVal) {
```

Source/DafnyCore/Coco/Parser.frame
```diff
@@ -223,12 +229,18 @@ public class Errors {
 
   // Code action range is up to, not including, the next token
   public static CodeActionToken CodeAction(IToken sourceTok, IToken next) {
-    return new CodeActionToken(sourceTok, sourceTok.col, next.pos - sourceTok.pos);
+    return CodeAction(sourceTok, sourceTok.col, next.pos - sourceTok.pos);
+  }
+
+  // Code action range is up to, not including, the given token's Next token 
+  // so the range includes following white space
+  public static CodeActionToken CodeActionToNext(IToken sourceTok) {
```
MikaelMayer, line 237: Could you rename this just `CodeAction`? I think it's pretty intuitive that code actions on a single token will consider this token until the next.

Source/DafnyCore/Dafny.atg
```diff
@@ -125,7 +126,8 @@ void CheckDeclModifiers(ref DeclModifierData dmod, string declCaption, AllowedDe
 
 void CheckAttribute(Errors errors, IToken attr) {
   if (attr.val == "dllimport" || attr.val == "handle") {
-    errors.Deprecated(attr, $"Attribute :{attr.val} is deprecated and will be removed in Dafny 4.0");
+    errors.Deprecated(ErrorID.p_deprecated_attribute, Errors.CodeAction(attr, attr.Prev.col, attr.Next.Next.col - attr.Prev.col),
```
MikaelMayer, line 129: I wish developers don't need to perform arithmetic at the calling site. Could this be a call like `(attr, attr.Prev, attr.Next.Next)` ?
Also, did these attributes have arguments? If so, we might want to replace the argument IToken attr by the range of the captured attribute so that you don't have an unintuitive `Next..Next` there

Source/DafnyCore/Dafny.atg
```diff
@@ -1099,7 +1102,8 @@ SubModuleDecl<DeclModifierData dmod, ModuleDefinition parent, out ModuleDecl sub
   ( ModuleDefinition<dmod, parent, out submodule>
   | ModuleImport<parent, out submodule>
   | ModuleExport<parent, out submodule>
-    (. if (isTopLevel) SemErr(submodule.tok, "There is no point to an export declaration at the top level");
+    (. if (isTopLevel) SemErr(ErrorID.p_superfluous_export, submodule.tok.Prev,
```
MikaelMayer, line 1105: I'm wondering if this is always the "export" identifier. I hope so?

Source/DafnyCore/ParserErrorDetail.cs
```diff
@@ -35,27 +34,339 @@ public class ParserErrorDetail {
...
+    Add(ErrorID.p_unexpected_type_characteristic,
+    @"
+[Type characteristics](../DafnyRef/DafnyRef#sec-type-parameters), 
+indicated in parentheses after the type name, state properties of the otherwise uninterpreted or opaque type.
+The currently defined type characteristics are designated by `==` (equality-supporting), `0` (auto-initializable), `00` (non-empty), and `!new` (non-reference).
+");
+
+    Add(ErrorID.p_missing_type_characteristic,
+    @"
+[Type characteristics](.. / DafnyRef / DafnyRef#sec-type-parameters),
```
MikaelMayer, line 237:
```suggestion
[Type characteristics](.. /DafnyRef/DafnyRef#sec-type-parameters), 
```

Source/DafnyLanguageServer/DafnyCodeActions.cs
```diff
@@ -33,22 +33,105 @@ public class DafnyCodeActions {
...
+    // p_bitvector_too_large -- no code action
+    // p_array_dimension_too_large -- no code action
+    // p_superfluous_exoirt -- TODO - need a code action
+    // p_bad_module_decl -- no code action
+    AddRemoveAction(ErrorID.p_extraneous_comma_in_export, "remove comma");
+    AddReplaceAction(ErrorID.p_top_level_field, "replace 'var' by 'const'", "const"); // also remove entire declaration?
+    // p_bad_datatype_refinement -- what code action
+    AddReplaceAction(ErrorID.p_no_mutable_fields_in_value_types, "replace 'var' by 'const'", "const");
+    AddReplaceAction(ErrorID.p_bad_const_initialize_op, "replace = with :=", ":=");
+    // p_const_is_missing_type_or_init -- no code action
```
MikaelMayer, line 88: Could you perhaps add a code action named "show me an example" that inserts `: int := 2` ?

Source/DafnyLanguageServer/DafnyCodeActions.cs
```diff
@@ -33,22 +33,105 @@ public class DafnyCodeActions {
...
+    AddRemoveAction(ErrorID.p_output_of_function_not_ghost, "remove 'ghost'");
```
MikaelMayer, line 89: Is there a way you could remove all these `remove 'XXX'` at the argument site, and at the parameter side, if this label is not provided, compute it automatically as `"remove '"+token.val.trim()+"'"`
That way, we could simplify the code while keeping the flexibility of having a custom remove message.
Same for "replace".

Source/DafnyLanguageServer/DafnyCodeActions.cs
```diff
@@ -33,22 +33,105 @@ public class DafnyCodeActions {
...
+    AddRemoveAction(ErrorID.p_no_older_on_output_formals, "remove 'older'");
+    // p_var_decl_must_have_type -- no code action
+    // p_no_init_for_var_field -- remove entire initiallizer
+    // p_datatype_formal_is_not_id -- no code action - perhaps remove the name and the colon
+    AddRemoveAction(ErrorID.p_nameonly_must_have_parameter_name, "remove 'nameonly'");
+    AddReplaceAction(ErrorID.p_should_be_yields_instead_of_returns, "replace 'returns' by 'yields'", "yields");
+    AddRemoveAction(ErrorID.p_type_parameter_variance_forbidden, "remove type parameter variance");
+
+    AddReplacementActions(ErrorID.p_unexpected_type_characteristic, new Tuple<String, String>[]{
+      Tuple.Create("replace with '=='", "=="),
```
MikaelMayer, line 102: Don't you want to insert instead of replacing?

MikaelMayer (changes requested):
These code actions are a great foundation for novice Dafny developers, great job implementing them (and adding detailed error messages !)
I wish there is more links to the documentation in the extended error messages, like at least once per error message. For example, for the constants, having a link to the section https://dafny.org/latest/DafnyRef/DafnyRef#224-const-declarations would be great. That would make it easy to discover related documentation for users starting with Dafny.

Keep up this excellent work!

Source/DafnyCore/Dafny.atg
davidcok: These did not have attributes. Once Remy has ranges for AST nodes, I can grab the range for the whole Attribute, not just for the id. So I'll let this refactoring wait until the next PR, if you don't object.

Source/DafnyCore/Dafny.atg
davidcok: It is. But I do worry that edits to the token or the grammar can break these.

Source/DafnyCore/ParserErrorDetail.cs
```diff
@@ -35,27 +34,339 @@ public class ParserErrorDetail {
...
-    );
+");
 
-    ErrorDetail.Add(ErrorID.p_deprecated_semicolon,
-  @"
-Semicolons are required after statements and declarations in method bodies,  
-but are deprecated after declarations within modules and types.
+    ErrorDetail.Add(ErrorID.p_deprecated_attribute,
+    @"
+The `{:handle}` and `{:dllimport}` attributes are obsolete and unmaintained. They will be removed.
```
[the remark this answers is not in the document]
davidcok: OK. (Did you mean for a change?)

Source/DafnyLanguageServer/DafnyCodeActions.cs
davidcok: In this case there is some token there that is illegal as a type characteristic -- so the action is to replace it.

Source/DafnyLanguageServer/DafnyCodeActions.cs
davidcok: Done - and figured out how to get text from the Document to do it

Source/DafnyCore/Dafny.atg
```diff
@@ -1989,13 +1999,18 @@ MethodDecl<DeclModifierData dmod, bool allowConstructor, out Method/*!*/ m>
...
+         SemErr(ErrorID.p_method_missing_name, la, "a method must be given a name (expecting identifier)");
        }
      }
   .)
   (
     [ GenericParameters<typeArgs, false> ]
-    [ KType<ref kType>              (. if (!(isGreatestLemma || isLeastLemma)) { SemErr(t, "type of _k can only be specified for least and greatest lemmas"); } .)
+    [ KType<ref kType>              (. if (!(isGreatestLemma || isLeastLemma)) { 
+                                         SemErr(ErrorID.p_extraneous_k, 
+                                                t.Prev == null || t.Prev.Prev == null || t.Next == null ? t : Errors.CodeAction(t.Prev.Prev,t.Next),
```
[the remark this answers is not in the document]
davidcok: Rewritten.

Source/DafnyLanguageServer/DafnyCodeActions.cs
MikaelMayer: Oh yeah, I see it's triggered only when given digits which are not 0 or 00, makes sense

Source/DafnyCore/Coco/Parser.frame
MikaelMayer: Whatever is the most frequent for single token should be just CodeAction(). The most frequent for removal is token to the next, and the most frequent for replacement is token to its final position.

I think what I am not comfortable with now is that we call this "CodeAction", whereas it has nothing to do with a code action, it's actually a *range* used by a code action.

So it might make sense to actually rename the class to CodeActionRange.

In any case, I think the case is perfect for an option, so I think for tokens the signature should be like this.

`CodeActionRange(IToken tok, bool IncludesTrailingSpace = false)`

By the way, now that I can think of it, you might want to only remove `token.TrailingTrivia` and not up to the next token. And even more, you might only want to remove the first consecutive whitespaces of `token.TrailingTrivia` that are not newlines. Here is why:

```
class X {
  static
  static /* comment about static */
  method Test() {
  }
}
```
In the example above, your code action would also remove the comment, not just the keyword. I think removing comments, in the unlikely event there are some, is not desired in any case.

MikaelMayer (approved):
Looks good to me ! I did not see tests on the language server side though. There exist already a bunch of such tests (look for CodeAction). Feel free to reuse the same syntax parser (which makes writing such tests easy)

Source/DafnyCore/ErrorDetail.cs
```diff
@@ -10,14 +10,61 @@ public static class ErrorDetail {
   public enum ErrorID {
```
keyboardDrummer, line 10: Why do we need this enum? How about using strings for the error codes wherever we create the error message? A very large enum with a value for each error message doesn't seem right to me. How does this scale to plugins with error messages in different assemblies that cannot be reached from `DafnyCore` ?

Source/DafnyCore/ParserErrorDetail.cs
```diff
@@ -12,50 +12,361 @@ public class ParserErrorDetail {
 
   public static void init() {
 
-    ErrorDetail.Add(ErrorID.p_bad_const_initialize_op,
+    Add(ErrorID.p_duplicate_modifier,
```
keyboardDrummer, line 15: I think it's better to have the explanations be located where the error messages are also created, so the code remains local.

Source/DafnyCore/Reporting.cs
```diff
@@ -235,6 +235,13 @@ public class ConsoleErrorReporter : BatchErrorReporter {
           msg = nestedToken.Message ?? "[Related location]";
           errorLine += $" {msg} {TokenToString(tok)}";
         }
+
+        if (DafnyOptions.O.CompileVerbose && false) { // Need to control tests better before we enable this
```
keyboardDrummer, line 239: Should this be removed?

Source/DafnyLanguageServer/DafnyCodeActions.cs
```diff
@@ -33,22 +33,145 @@ public class DafnyCodeActions {
...
   private static bool initialized = false;
   public static void init() {
     if (initialized) {
       return;
     }
     initialized = true;
-    codeActionMap.Add(ErrorID.p_bad_const_initialize_op, (Diagnostic diagnostic, Range range) => ReplacementAction("replace = with :=", diagnostic, range, ":="));
-    codeActionMap.Add(ErrorID.p_abstract_not_allowed, (Diagnostic diagnostic, Range range) => RemoveAction("remove 'abstract'", diagnostic, range));
-    codeActionMap.Add(ErrorID.p_no_leading_underscore, (Diagnostic diagnostic, Range range) => RemoveAction("remove underscore", diagnostic, range));
+    AddRemoveAction(ErrorID.p_duplicate_modifier, "remove duplicate modifier'");
```
keyboardDrummer, line 110: Please register the actions where the error message is created.

Source/DafnyCore/ErrorDetail.cs
davidcok: In early discussion with Mikhael on this design we debated among:
1) using just constant string literals everywhere
2) using 'const string abc = "abc";' declarations and then using the name everywhere
3) using an enum

The first seemed too prone to easily missed misspellings
The second seemed verbose.
The third seemed best.

Source/DafnyCore/ParserErrorDetail.cs
davidcok: We debated this too. At least for the batch that are parser error explanations, it seemed to make the parser code cluttered to insert explanations in the code for all the productions. And the other extreme, one location for all explanations seemed too non-local. So we compromised by having one file that is in close association with the code whose errors it explains.

Source/DafnyCore/Reporting.cs
davidcok: It was intended to be. Although it will be added back when we figure out which option to control showing explanations on the command-line.

Source/DafnyLanguageServer/DafnyCodeActions.cs
davidcok: The problem is that registering and using the actions is done in the language server. The error message is created in the DafnyCore code. We did not want to drage LS code into the DafnyCore

keyboardDrummer: I think refactorings can be part of DafnyCore, and it's not something that should only be available to do language server. You can imagine running `dafny format --include-code-actions` and also applying these code actions.

Source/DafnyCore/ParserErrorDetail.cs
keyboardDrummer: What do you mean by cluttered?

`Dafny.atg` is too large and centralised as is, and ideally we break it up into pieces, but that's something for the future. For now I think adding the extra information in `Dafny.atg` is OK, but I'm also OK with having a separate file for error information if it only relates to the parser.

Source/DafnyCore/ErrorDetail.cs
MikaelMayer: I wanted to use the enum just to avoid repeating a variable name and a constant with the same value as the variable name.
If you have a way to do that @keyboardDrummer so that it's compatible with any other user-provided strings, then I think the would be the best

Source/DafnyCore/ParserErrorDetail.cs
MikaelMayer: I agree with David. Because explanations can be heavy, it's better if they are in a separate file so that they don't distract a developer who thinks about the code.
It's the same situation as for the ProofObligationDescription, we move the wording in separate classes so that, even though we don't have the wording locally when creating the assertion, it's still easily accessible, and some elements can also be factored together.

PR #3500, opened 2023-02-09
Misc error message improvements and bug fixes: #1637, #3496, #3497, #2200, #2477, #1618, #2830

### LINKED ISSUE #1618: Bad error message when forgetting the parens on a reveal statement
Example:

```
function {:opaque} mul(a: int, b: int): int
{
  a * b
}

method M()
{
  reveal mul;
}
```

Dafny reports:

```
tmp.dfy(8,9): Error: expression is not allowed to invoke a lemma (reveal_mul)
1 resolution/type errors detected in tmp.dfy
```

It is not obvious to the user that they forgot the parens.

### LINKED ISSUE #1637: Error messages involving subset types
Sometimes I get an error message when using a subset type.
I can't seem to get a simple repo, so this is just an example, not a full repo.

```dafny
datatype Thing = Thing
type Foo = s: seq<uint8> | |s| < 10 witness *

function Bar(t: Thing): Foo

predicate Baz(f: Thing -> seq<uint8>)

predicate compose() {
  Baz(Bar)
}
```

I would get an error at `Baz(Bar)` that says something like "Bar does not match Thing -> seq<uint8>" and something about covariance...
I propose that we include what Bar does provide so that it is more clear what Dafny is confused about.
Sometimes Dafny can work things out,
but other time it can not.
If the error message was more like "Bar provides Thing -> Foo which does not match Thing -> seq<uint8>"
it would be clearer what is going on.

### LINKED ISSUE #2200: Show expected argument types for incorrect number of arguments
Currently (v 3.6.0) when a function gets an incorrect number of arguments, the following message is displayed:

```
(7,5): Error: wrong number of arguments (function 'foo' expects 2, got 1)
```

In my development flow, I often know which function to call, but need to figure out which arguments to call it with; having the argument types in the error message would greatly help with that.

E.g.
```
(7,5): Error: wrong number of arguments (function 'foo' expects 2 parameters 'a' of type int, 'b' of type string, got 1)
```

or maybe

```
(7,5): Error: wrong number of arguments (function 'foo' expects 2 parameters (a: int, b: string), got 1)
```
but that is less in tune with the error message on mismatched types (but correct number of args).

An obvious workaround is to guess the number of arguments, and give them "dummy" values, e.g. `42`, but this seems like a pretty easy fix.

### LINKED ISSUE #2477: Confusing message about automatically generated decrease clauses
Forked from #2384 :

```
trait T {
  predicate P()
    reads {this}
}

class C extends T {
  predicate P() // predicate's decreases clause must be below or equal to that in the trait
  {
    true
  }
}
```

Quoth @RustanLeino : the problem stems from the automatically generated `decreases` clause. It would be nice to change `decreases clause` to `(automatically generated) decrease clause` in the error message in this case, to try to reduce confusion.

### LINKED ISSUE #2830: Poor location of error message printed by /functionSyntax flag
With /functionSyntax:4 The error 
Error: the phrase 'function method' is not allowed; to declare a compiled function, use just 'function'
gives a location at the end of the function instead of near the beginning where the problem is.
```
predicate method f() {













  true
}
```
/Users/davidcok/projects/mydafny/Line.dfy(16,0): Error: the phrase 'predicate method' is not allowed; to declare a compiled predicate, use just 'predicate'

### LINKED ISSUE #3496: Error message for assume statement is misleading

If an `assume` statement is left in the source code, Dafny complains:

    Error: an assume statement without an {:axiom} attribute cannot be compiled

This message seems to suggest that an `assume` statement _can_ be compiled if you only add the `{:axiom}` attribute to it. Here's a replacement error message that I think is more to the point:

    Error: an assume statement cannot be compiled (use the {:axiom} attribute to let the compiler ignore the statement)

### LINKED ISSUE #3497: Deprecate forall statement with no bound variables
### Summary

Dafny allows a `forall` statement to take no bound variables. For example,

``` dafny
forall
  ensures F(x) == 100
{
  // proof why F(x) == 100...
}
```

This degenerate form of `forall` was once useful for stating and proving properties in the middle of code. However, Dafny has now for years had an `assert by` statement that does the same thing:

``` dafny
assert F(x) == 100 by {
  // proof why F(x) == 100...
}
```

The `assert by` keywords immediately suggest the meaning of the statement, whereas the degenerate form of the `forall` statement looks mysterious or ill-formed. Clearly, the `assert by` statement is to be preferred.

I therefore suggest we deprecate the 0-variable form of the `forall` statement. This will promote the clearer statement and will reduce code paths in the Dafny implementation.

### Background and Motivation

.

### Proposed Feature

.

### Alternatives

_No response_

davidcok:
Fixes #3497. Fixes #3496. Fixes #2200. Fixes #1618. Fixes #1637. Fixes #2477. Fixes #2830.


docs/check-examples
```diff
@@ -200,6 +200,7 @@ do
         capture=1
         dOptions=$defOptions
         if [ "$command" == "%check-verify" ]; then
+          ##cat $text
```
atomb, line 203: Is this a change you meant to include in this PR?

atomb (approved):
This looks good, with one low-priority question.

docs/check-examples
davidcok: It is a debugging line -- I'll take it out in the next PR of this sort.

atomb (approved):

PR #3503, opened 2023-02-10
Fix: Implementations to start verifying after completed assertion batches

### LINKED ISSUE #3502: Flacky test because of race condition
I found another race condition that led to the LSP test `EnsuresManyDocumentsCanBeVerifiedAtOnce ` failing, which could happen on any other tests, it's just that this one is the most stressful for the system.

Here is the diff between expected and actual
```
  .  |  |  |  I  |  | :predicate F(i: int) {
  .  |  |  |  I  |  | :  false // Should not be highlighted in gutter.
  .  |  |  |  I  |  | :}
     |  |  |  I  |  | :
- .  S [S][ ][I][S][ ]:method H()
- .  S [=][=][-][~][O]:  ensures F(1)
- .  S [=][=][-][~][=]:{//Next: { assert false;
- .  S [S][ ][I][S][ ]:}
+ . [I][S][ ][I][S][ ]:method H()
+ . [=][~][=][-][~][O]:  ensures F(1)
+ . [=][~][=][-][~][=]:{//Next: { assert false;
+ . [I][S][ ][I][S][ ]:}
```
Looking at it, it means that the status of the method was set to "running" _after_ the first assertion batch was verified. I can witness I have seen this behavior in real instances of VSCode, where it was saying it verified, and then it pretended it started verifying only to provide the same result afterwards.

This comes from line 258 of Compilation.cs:
https://github.com/dafny-lang/dafny/blob/f224f9a15f2a225883570eea8a20f1b69bd74dbd/Source/DafnyLanguageServer/Workspace/Compilation.cs#L253-L260

Basically, it's a place where we tell the gutter icons that the implementation has started, whereas the boogie engine had already started, send the notification that the implementation had started but it was ignored, and sent the indication about the result of verifying the assertion batch, which was taken into account. Here is the code that indicates that the original implementation verification started message was ignored.

https://github.com/dafny-lang/dafny/blob/f224f9a15f2a225883570eea8a20f1b69bd74dbd/Source/DafnyLanguageServer/Language/AssertionBatchCompletedObserver.cs#L51-L52

Originally (back in July), this code was the only ground truth to tell if an implementation had started verifying.

MikaelMayer:
This PR fixes #3502 according to the observations
The test `EnsuresManyDocumentsCanBeVerifiedAtOnce` is already showing this behavior and should be able to capture with a relatively decent probability if this happens in the future.


Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
```diff
@@ -302,6 +307,12 @@ INotificationPublisher notificationPublisher
           return;
         }
 
+        if (!implementationNode.Started) {
+          // Race condition identified in https://github.com/dafny-lang/dafny/issues/3502
+          // We first need to mark the implementation as it started to verify
+          ReportVerifyImplementationRunning(implementation);
```
keyboardDrummer, line 313: Based on the Boogie code, I see how "Running" can be reporter after assertion batch results.

Looking at https://github.com/boogie-org/boogie/blob/master/Source/ExecutionEngine/ImplementationTask.cs#L163

```c#
  private async Task<VerificationResult> RunInternal(CancellationToken cancellationToken, Action<IVerificationStatus> notifyStatusChange) {

    var enqueueTask = engine.EnqueueVerifyImplementation(ProcessedProgram, new PipelineStatistics(),
      null, null, Implementation, cancellationToken, TextWriter.Null);

    // From this point on, assertion batch results can be reported, even though status has not yet been set to running (or queued even)

    var afterEnqueueStatus = enqueueTask.IsCompleted ? (IVerificationStatus)new Running() : new Queued();
    notifyStatusChange(afterEnqueueStatus);

    var verifyTask = await enqueueTask;
    if (afterEnqueueStatus is not Running) {
      notifyStatusChange(new Running());
    }

    var result = await verifyTask;

    // Assertion batch results can no longer be reported.

    CacheStatus = new Completed(result);
    notifyStatusChange(CacheStatus);
    return result;
  }
```

keyboardDrummer (approved):
Nice fix on the Dafny side. I will fix it on the Boogie side so it won't be as confusing for feature Boogie customers (and this fix'll be obsolete then)

keyboardDrummer (approved):
I'm surprised by the output of the failed tests.

PR #3509, opened 2023-02-10
chore: Revert likely culprits for CI failures

fabiomadge:
Fixes #


atomb (approved):
This does seem to fix the issue, so let's put it in place for now. I was hoping to determine why the new-style commands were failing to find Z3, but haven't figured that out yet. It'll be easy enough to un-revert things once we fix that issue.

PR #3510, opened 2023-02-10
Redo: Add new macros to tests for new CLI

davidcok:


keyboardDrummer (dismissed):

Source/IntegrationTests/LitTests.cs
```diff
@@ -38,10 +38,15 @@ public class LitTests {
 
       var repositoryRoot = Path.GetFullPath("../../../../../"); // Up from Source/IntegrationTests/bin/Debug/net6.0/
 
+      string[] defaultResolveArgs = new[] { "resolve", "--use-basename-for-filename" };
+      string[] defaultVerifyArgs = new[] { "verify", "--use-basename-for-filename", "--cores:2", "--verification-time-limit:300" };
+      //string[] defaultTranslateArgs = new[] { "translate", "--use-basename-for-filename", "--cores:2", "--verification-time-limit:300" };
```
keyboardDrummer, line 43: Why is this commented out?

keyboardDrummer (approved):

Source/IntegrationTests/LitTests.cs
davidcok: dafny translate takes an extra subcommand that needs to be after the 'translate'. The style of defining %translate to be "dafny translate <default options>' does not work. Instead I just use %baredafny translate <platform> <options> for the few tests that use translate.

PR #3512, opened 2023-02-11
Defining macros for using new CLI in tests

davidcok:
This just defines new macros without tests that use them.
- Using them seemed to cause some nightly test failures
- Defining the macros allows them to be used lcoally and for debugging the nightly tests problem.


keyboardDrummer (commented):
I think there should be tests using these macros, otherwise they aren't tested.

davidcok:
They are in followon PRs, which are waiting on this one. I broke them up so that it was clearer what was changing when builds failed.

PR #3534, opened 2023-02-14
Pre-Dafny 4 edits - statements chapter

davidcok:


docs/DafnyRef/Statements.md
```diff
@@ -390,64 +390,78 @@ method m() {
...
+### 8.6.2. Method call with out-parameters
+
+This form uses `:=` to denote the assignment of the out-parameters of the method to the 
+corresponding number of LHS values.
+<!--%no-check -->
+```dafny
+a, b.e().f := m() {:attr};
+```
+
+In this case, the right-hand-side must be emthod call and the number of
```
atomb, line 422:
```suggestion
In this case, the right-hand-side must be a method call and the number of
```

docs/DafnyRef/Statements.md
```diff
@@ -1891,10 +1907,26 @@ depend on any testing framework in the target language.
...
+to see if there are some conditions when the assert is not actually true. So one might have
+paired assert/expect statements:
+<!-- %no-check -->
+```dafny
+assert _P_;
+expect _P_;
+```
+Once the program is debugged, both statements can be removed.
+Note that it is important that the `assert` come before the `expect`, because
+to the verifier, the `expect` is an `assume`, which would automatically make
```
atomb, line 1923:
```suggestion
by the verifier, the `expect` is interpreted as an `assume`, which would automatically make
```

I'm not set on this, but it might be an improvement.

atomb (changes requested):
One tiny (pre-existing?) typo, and one possible wording suggestion that might be an improvement (or you might come up with something even better).

atomb (approved):

PR #3537, opened 2023-02-15
Fix docs default for /normalizeNames, /errorTrace, and /emitDebugInformation

tchajed:
The default changed in 3.4 to become true rather than false.

The option is set to true here: https://github.com/dafny-lang/dafny/blob/e7611742e812f9da7862a02f6d3503857212e981/Source/DafnyCore/DafnyOptions.cs#L199-L205

### DIFF
--- a/docs/DafnyRef/Options.txt
+++ b/docs/DafnyRef/Options.txt
@@ -561,12 +561,12 @@ Usage: dafny [ option ... ] [ filename ... ]
                             of the path supplied on the command line
 
   /emitDebugInformation:<n>
-                0 - do not emit debug information
-                1 (default) - emit the debug information :qid, :skolemid and set-info :boogie-vc-id
+                0 (default) - do not emit debug information
+                1 - emit the debug information :qid, :skolemid and set-info :boogie-vc-id
 
   /normalizeNames:<n>
-                0 (default) - Keep Boogie program names when generating SMT commands
-                1 - Normalize Boogie program names when generating SMT commands. 
+                0 - Keep Boogie program names when generating SMT commands
+                1 (default) - Normalize Boogie program names when generating SMT commands.
                   This keeps SMT solver input, and thus output, 
                   constant when renaming declarations in the input program.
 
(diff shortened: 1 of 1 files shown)


tchajed:
I seem to have missed something - it seems like Options.txt is actually auto-generated, and its the code that produces `dafny /help` that's actually wrong. I could use some help understanding where the fix should go.

davidcok:
I'm currently editing documentation. So I'll look into this and your fix.

But tell me what you are trying to accomplish (Options.txt is indeed autogenerated).
Make the documentation in line with actual behavior on these three options?

davidcok:
Those are boogie options. You will have to work with the boogie team to fix those. Dafny does not control them; it just passes on the help information boogie provides.

atomb:
Ah, that's an unfortunate issue where the help text is generated from Boogie, where the defaults are as they're listed, but Dafny overrides those defaults without updating the help text.

Given that we're moving toward the new version of the CLI with top-level verbs that doesn't show the Boogie help at all, I'm inclined to leave the Boogie help text as-is and, to the extent that those options are useful to change at the Dafny level, provide new, Dafny-specific (and Dafny-named) options with the correct help text generated from Dafny.

davidcok:
Actually - the help text comes from boogie and they are the correct boogie defaults. But Dafny sets different values for its use of boogie. As we are changing the way options are implemented in Dafny (and hiding those from boogie), we are not planning to fix this now.

tchajed:
Ok, it sounds like this is a "won't fix" bug and eventually there will be a different interface for setting Boogie options, so I'll close this PR (which anyway isn't the right fix for the underlying issue).

The option that actually came up in our use case was `/normalizeOptions:0`, which makes the generated SMT files much easier to debug. Hopefully in the new CLI there remains a way to do this.

keyboardDrummer:
> The option that actually came up in our use case was /normalizeOptions:0, which makes the generated SMT files much easier to debug. Hopefully in the new CLI there remains a way to do this.

Could you comment on why you're debugging the SMT files?

PR #3540, opened 2023-02-15
Issue#3513 -- allow value of library option to be  list

### LINKED ISSUE #3513: Allow a directory as the value of --libraries
### Summary

Currently Dafny requires all library files to be listed individually as .dfy files. This requires users to know just what is needed from a possibly large collection of files.

### Background and Motivation

A user should be able to use a library without having to worry too much about how it is internally packaged -- which files are present, which depend on which etc. While include files are helpful, the usual way to use a library is for the user to be able to say: here is a library, use what you need.

### Proposed Feature

That --libraries take a directory as its value, meaning all the .dfy files recursively in that directory

### Alternatives

_No response_

davidcok:
Fixes #3513

Allows the value of the --library option to be a comma-separated sequence of files or directories.


Test/git-issues/git-issue-3513/git-issue-3513.dfy
```diff
@@ -0,0 +1,8 @@
+// RUN: %exits-with 2 %baredafny resolve --use-basename-for-filename "%s" > "%t"
+// RUN: %exits-with 2 %baredafny resolve --use-basename-for-filename --library:"%S/src/A.dfy" "%s" >> "%t"
```
keyboardDrummer, line 2: The comma splitting behavior is not tested

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -169,6 +169,23 @@ public enum CommandLineArgumentsResult {
       OK_EXIT_EARLY
     }
 
+    static HashSet<string> SplitOptionValueIntoFiles(HashSet<string> inputs) {
+      var output = new HashSet<string>();
+      foreach (var input in inputs) {
+        var values = input.Split(',');
+        foreach (var slice in values) {
+          var name = slice.Trim();
+          if (System.IO.Directory.Exists(name)) {
```
keyboardDrummer, line 178: I would leave out the `System.IO` in front of Directory.

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -169,6 +169,23 @@ public enum CommandLineArgumentsResult {
       OK_EXIT_EARLY
     }
 
+    static HashSet<string> SplitOptionValueIntoFiles(HashSet<string> inputs) {
+      var output = new HashSet<string>();
```
keyboardDrummer, line 173: I would name this `result` instead of `output`. Output sounds like the CLI will print it.

keyboardDrummer (approved):

Source/DafnyDriver/DafnyDriver.cs
davidcok: Edited in a followup PR.

Test/git-issues/git-issue-3513/git-issue-3513.dfy
davidcok: Added a test in a followup PR.

PR #3542, opened 2023-02-15
Fixes #3267 -- improved command-line error messages

### LINKED ISSUE #3267: Better error message for misspelled or incorrect command
### Dafny version

3.10.0+dev

_No response_

```code
dafny format tmp.dfy
*** Error: 'format': Filename extension '' is not supported. Input files must be Dafny programs (.dfy) or supported auxiliary files (.cs, .dll)
```

### What happened?

In a command-focussed CLI, the first arg is supposed to be a command -- so this message will be confusing. Something like

*** Error: the first argument 'format', is neither a command, an option, a dafny file (*.dfy), nor a supported auxiliary file (_list_)

Mac

davidcok:
Makes file names in command-line error messages respect --use-basename-for-filename and generally improves their appearance. This also makes such errors testable in lit.

Fixes #3267


keyboardDrummer (approved):
I don't fully understand the problem this change is solving, and the code itself looks quite subtle, but it also looks harmless so I'll approve.

I wonder whether we should change the option parser so it will parse using the new mode by default, and only if that has errors, attempt the old mode, and then only if that has no errors, show the output of the old mode. And then for the new mode we should rely on configuring `System.CommandLine` as much as possible to generate the error messages. Would this parsing approach make the changes in this PR obsolete?

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -275,26 +275,34 @@ public enum CommandLineArgumentsResult {
           // Fall through and try to handle the file as an "other file"
         }
 
+        var relative = System.IO.Path.GetFileName(file);
+        bool useRelative = options.UseBaseNameForFileName || relative.StartsWith("-");
```
keyboardDrummer, line 279: What is the significance of the `-` here?

davidcok: It means it is likely the start of an option. And even if its not, one just gets the relative name of the file, which is a more readable error message

davidcok:
In working with the new CLI quite a bit in writing and debugging tests and examples, (a) the fact that --=use-basename-for-filename was not respected made it not possible to write tests and (b) the error messages for simple typos, e.g. in a verb, were messy and confusing. This change fixes that and includes tests.

PR #3547, opened 2023-02-15
Change order the OSes are run to put longest first

davidcok:
When the CI queue is congested, some jobs are in the queue a while before they even start.
The total elapsed time can be reduced by putting the longer jobs first. This PR does that for the XUnit tests 
(Mac first, then Windows, then Ubuntu)


keyboardDrummer (approved):
It does give you feedback later in case there is congestion, but alright ^,^

keyboardDrummer:
> Which feedback are you referring to?

If you have a failing unit test, you may now be alerted to that only after the slow tests have run, instead of after the fast ones have.

PR #3553, opened 2023-02-16
Axiom warnings

### LINKED ISSUE #2961: Warn on axioms without {:axiom}

keyboardDrummer:
Fixes #2961.

### Changes
- In situations where we believe the developer is intentionally bypassing verification as a _temporary_ development measure, always produce a warning. Some previous cases produced no warning and now do, and some produces an error and now produce a warning.
- In situations where we believe the developer is intentionally bypassing verification for production purposes, do not produce a warning or an error.

Specifically:
- Warn on any bodyless function or methods without an `{:axiom}` attribute, regardless of whether we're generating code or not.
- Warn on any `assume` statement, or an `:| assume <E>`. without an `{:axiom}` attribute
- Do not error on bodyless ghost functions, bodyless ghost methods and assume statements when generating code, even if they don't have an `{:axiom}` attribute

Table overview old situation:
|    | Verification | CodeGen |
| ----- | ----- | ----- |
| body-less ghost function or method without `{:axiom}` or `{:extern}` | **nothing** | **error** |
| body-less compiled function or method without `{:axiom}` or `{:extern}` | **nothing** | error |
| body-less ghost function or method with `{:axiom}` | nothing | nothing |
| body-less ghost function or method with `{:extern}` | nothing | nothing |
| body-less compiled function or method with `{:axiom}` | nothing | error |
| body-less compiled function or method with `{:extern}` | nothing | nothing |
| body-less loop | warning | **warning** |
| `assume` statement without `{:axiom}` | **nothing** | **error** |
| `assume` statement with `{:axiom}` | nothing | nothing |
| `expect` statement | nothing | nothing |

Table overview new situation:
|    | Verification | CodeGen |
| ----- | ----- | ----- |
| body-less ghost function or method without `{:axiom}` or `{:extern}` | **warning** | **warning** |
| body-less compiled function or method without `{:axiom}` or `{:extern}` | **warning** | error |
| body-less ghost function or method with `{:axiom}` | nothing | nothing |
| body-less ghost function or method with `{:extern}` | nothing | nothing |
| body-less compiled function or method with `{:axiom}` | nothing | error |
| body-less compiled function or method with `{:extern}` | nothing | nothing |
| body-less loop | warning | **error** |
| `assume` statement without `{:axiom}` | **warning** | **warning** |
| `assume` statement with `{:axiom}` | nothing | nothing |
| `expect` statement | nothing | nothing |

We would like to report a warning for {:extern} on a ghost method or function, but we leave that for a future PR.

#### Refactoring changes:
- Move resolution code for Assert/Assume/Expect into the related type definition files

### Testing

- Add tests for the above behavior


Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -162,6 +162,15 @@ public interface IAttributeBearingDeclaration {
     Attributes Attributes { get; }
   }
 
+  static class AttributeBearingDeclaration {
+
+    public static bool IsExtern(this IAttributeBearingDeclaration me) =>
+      DafnyOptions.O.AllowExterns && Attributes.Contains(me.Attributes, "extern");  // TODO do earlier check to complain about any usages of :extern if allow externs is false.
+
+    public static bool IsExplicitAxiom(this IAttributeBearingDeclaration me) =>
+      Attributes.Contains(me.Attributes, "axiom");  // TODO do earlier check to complain about any usages of :extern if allow externs is false.
```
atomb, line 171: Is this comment relevant here?

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -162,6 +162,15 @@ public interface IAttributeBearingDeclaration {
     Attributes Attributes { get; }
   }
 
+  static class AttributeBearingDeclaration {
+
+    public static bool IsExtern(this IAttributeBearingDeclaration me) =>
+      DafnyOptions.O.AllowExterns && Attributes.Contains(me.Attributes, "extern");  // TODO do earlier check to complain about any usages of :extern if allow externs is false.
```
atomb, line 168: Would it be feasible to include this change in this PR, or is it a lot of extra work? If it's the latter, could we have an issue for it?

Source/DafnyCore/AST/Statements/AssignSuchThatStmt.cs
```diff
@@ -76,6 +76,10 @@ public AssignSuchThatStmt(RangeToken rangeToken, List<Expression> lhss, Expressi
     base.Resolve(resolver, resolutionContext);
 
     if (AssumeToken != null) {
+      if (!AssumeToken.IsExplicitAxiom()) {
+        resolver.Reporter.Warning(MessageSource.Resolver, ErrorDetail.ErrorID.None, AssumeToken.Token, "Assume statement has no {:axiom} annotation.");
```
atomb, line 80: It might be an improvement for this message to mention the context more explicitly, i.e. that it's in an assign-such-that statement.

atomb (commented):
I saw a few things that could potentially be improved. It looks like a bunch of tests also need to be updated, though.

Source/DafnyCore/AST/DafnyAst.cs
keyboardDrummer: I've removed the TODO since I'm sure whether this option is used. It's not part of the new CLI

docs/DafnyRef/Options.txt
```diff
@@ -192,6 +192,9 @@ Usage: dafny [ option ... ] [ filename ... ]
   
   ---- Verification options -------------------------------------------------
   
+  /allowAxioms:<value>
```
robin-aws, line 195: As per offline discussion, let's just make this default to true for now, and we'll switch it to false for Dafny 4.0

Test/allocated1/dafny0/CompilationErrors.dfy.expect
```diff
@@ -1,23 +1,25 @@
...
+CompilationErrors.dfy(27,2): Warning: assume statement has no {:axiom} annotation
+CompilationErrors.dfy(33,2): Warning: assume statement has no {:axiom} annotation
+CompilationErrors.dfy(43,6): Warning: assume statement has no {:axiom} annotation
+CompilationErrors.dfy(55,8): Warning: assume statement has no {:axiom} annotation
+CompilationErrors.dfy(60,16): Warning: assume keyword in assign-such-that statement has no {:axiom} annotation
 CompilationErrors.dfy(71,2): Warning: note, this loop has no body (loop frame: x)
 CompilationErrors.dfy(73,2): Warning: note, this loop has no body (loop frame: i, x)
 CompilationErrors.dfy(83,4): Warning: note, this loop has no body (loop frame: x)
 CompilationErrors.dfy(85,4): Warning: note, this loop has no body (loop frame: i, x)
+CompilationErrors.dfy(12,6): Warning: assume statement has no {:axiom} annotation
```
robin-aws, line 15: For the near future: it would be nice to have a warning code attached to make it easier to look up the explanation for why this is an issue.

For now: could we add an FAQ entry? https://github.com/dafny-lang/dafny/tree/master/docs/HowToFAQ. Even just one with all the new error messages with a link to the relevant reference manual section would be great.

Test/ast/method.dfy
```diff
@@ -0,0 +1,6 @@
+// RUN: %baredafny verify --use-basename-for-filename "%s" > "%t"
+// RUN: ! %baredafny run --use-basename-for-filename "%s" >> "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method Foo()
+method {:axiom} Bar()
```
robin-aws, line 6: {:axiom} is just about assuming post-conditions without proof, so couldn't we allow bodiless methods with no post-conditions?

For the record (for later) I think I'd like to support attaching `{:axiom}` to individual `ensures` clauses instead of the top-level declaration. There are two reasons:
1. The term "axiom" is supposed to mean "assume this is true without proof". So "this post-condition is an axiom" is much more natural than "this method is an axiom" (when in this case there's actually nothing being assumed at all).
2. It would allow you to make individual `ensures` clauses axioms while still proving others, which could be useful when methods DO have bodies.

robin-aws: I realize this is an orthogonal improvement to the feature you are just emitting warnings about, but if it's quick to implement it means fewer false-positive warnings so it's worth doing if possible.

Test/git-issues/git-issue-1637.dfy
```diff
@@ -7,10 +7,10 @@ datatype Thing = Thing
 type Foo =  s: seq<uint8> | |s| < 10 witness *
 type Fii =  real 
 
-function Bar(t: Thing): Foo
-function Bii(t: Thing): Fii
+function {:axiom} Bar(t: Thing): Foo
+function {:axiom} Bii(t: Thing): Fii
 
-predicate Baz(f: Thing -> seq<uint8>)
+predicate {:axiom} Baz(f: Thing -> seq<uint8>)
```
robin-aws, line 13: To my other point, this is especially confusing since `{:axiom}` DOESN'T mean this predicate is always true, it means all post-conditions always hold (which is vacuously true in this case). I think this emphasizes why not requiring it here as I suggest above is worth considering before we force all our users to add this.

robin-aws (changes requested):
(just reviewing the feature from a usability standpoint since @atomb reviewed the implementation)

This is a great feature, but it's hampered by the existing confusing design of `{:axiom}` through no fault of your own. I've made a suggestion for tweaking this I'd strongly consider if possible, as otherwise this may cause more confusion than it's worth.

Test/ast/method.dfy
keyboardDrummer: Nice suggestion! I will make this change

keyboardDrummer:
@robin-aws Could you check for potential soundness issue in the the changes in `traits/TraitCompileErrors.dfy.expect` specifically?

docs/DafnyRef/Options.txt
keyboardDrummer: I would have been OK with introducing this default even if we weren't doing a major version bump. I think it's OK to have minor breaking changes like this be part of minor version bumps, given the current level of maturity of Dafny.

However, since the tests are setting the value anyways, there's no effort in changing the default.

Test/allocated1/dafny0/CompilationErrors.dfy.expect
keyboardDrummer: > For the near future: it would be nice to have a warning code attached to make it easier to look up the explanation for why this is an issue.

IMO the mechanism needs some refactoring before we start using it more.

> For now: could we add an FAQ entry? https://github.com/dafny-lang/dafny/tree/master/docs/HowToFAQ. Even just one with all the new error messages with a link to the relevant reference manual section would be great.

Could you give me a sample of what you're looking for? I'd like to add a detailed description that can be shown on hover, but as a I mentioned I think that feature needs a little work.

robin-aws:
> @robin-aws Could you check for a potential soundness issue in the changes in traits/TraitCompileErrors.dfy.expect specifically?

Looking at that and other test behavior changes, I'm not convinced you actually want eveything in the last noted change in this PR:

> * Do not error on bodyless ghost functions, bodyless ghost methods and assume statements when generating code, even if they don't have an {:axiom} attribute

That goes against how things are currently described in the reference manual, at least for assumes. I can't seem to find a solid statement about bodiless ghost functions and methods that aren't lemmas - @RustanLeino do you think we should allow those when compiling?

You might have been led astray by the error you see in the old version of `git-issue-663.dfy.expect` for example: "Error: an assume statement cannot be compiled (use the {:axiom} attribute to let the compiler ignore the statement)". I'm pretty sure the error message was incorrect, as the compiler has been rejecting all assumes whether they have `{:axiom}` or not. `assert {:axiom} ...` CAN be compiled though, so that error message should be fixed.

keyboardDrummer:
> You might have been led astray by the error you see in the old version of `git-issue-663.dfy.expect` for example: "Error: an assume statement cannot be compiled (use the {:axiom} attribute to let the compiler ignore the statement)". I'm pretty sure the error message was incorrect, as the compiler has been rejecting all assumes whether they have `{:axiom}` or not. `assert {:axiom} ...` CAN be compiled though, so that error message should be fixed.

I think that message matches the current behavior. Look at [this test](https://github.com/dafny-lang/dafny/blob/master/Test/dafny0/CompilationErrors.dfy#L13) for example, `assume {:axiom} true` is compiled without error.

robin-aws:
> > You might have been led astray by the error you see in the old version of `git-issue-663.dfy.expect` for example: "Error: an assume statement cannot be compiled (use the {:axiom} attribute to let the compiler ignore the statement)". I'm pretty sure the error message was incorrect, as the compiler has been rejecting all assumes whether they have `{:axiom}` or not. `assert {:axiom} ...` CAN be compiled though, so that error message should be fixed.
> 
> I think that message matches the current behavior. Look at [this test](https://github.com/dafny-lang/dafny/blob/master/Test/dafny0/CompilationErrors.dfy#L13) for example, `assume {:axiom} true` is compiled without error, and [here](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Compilers/SinglePassCompiler.cs#L2914) is the code that implements that behavior.

Hmm, looks like you're right. But now it feels weird to me that we more recently added support for `assert {:axiom} P`, which seems to behave identically to `assume {:axiom} P`. Knowing the latter is an option I think it's a better way to express it: "assume this is true, and not just temporarily because I believe it is a true axiom" vs. "assert this is true so the verifier has to prove it...only no don't actually do that because it's an axiom". I'm tempted to immediately deprecate `assert {:axiom} P` support.

@RustanLeino what's your option here, on this and the question of "compiling" body-less ghost functions/methods? We've had different instincts here for a while, ever since I just wanted to compile `assume` instead of adding `expect`. :)

RustanLeino:
## Language-design philosophy

As a verifier, Dafny allows you to declare things and reason about them. For example, you can define a function and prove a lemma about it. To earn the status of being a "verifier", Dafny has to be careful in what kinds of declarations it admits. For example, if you declare a lemma with `ensures false`, then using this lemma in proofs would let you prove things that don't actually hold.

In typical cases, there's a straightforward way to make sure the declarations in a program are mathematically consistent: for each declaration, insist that the program include a "witness" that justifies admitting the declaration. For example, if the program declares a lemma, then the program also has to supply a proof of that lemma. (This proof is, in some sense, a "witness" that shows that such a lemma exists and can be admitted by the verifier.) As another example, if you declare a method with a specification, then the program also has to supply an implementation of that method. (Here, the method implementation is a mathematical "witness" that shows that there exists a method with the specification you gave.)

As a programming language, Dafny allows you to write programs that can be compiled and executed. A prerequisite of compilation is that the given declarations exist as mathematical things, which is ensured by what I said above. In fact, the straightforward way of ensuring mathematical consistency also helps compilation--the compiler might not know how to turn a specification into executable code, but it does know how to turn compositions of statements into executable code and these statements are, essentially, supplied as the same "witnesses" that guarantee mathematical consistency.

Dafny splits declarations into non-ghost and ghost declarations and has a system of rules that allows the compiler to erase the ghost declaration during compilation. The soundness of this erasure relies on knowing the existence--in principle, in mathematics--of values for the ghost parts, even if those parts aren't present at run time.

So, as long as every declaration has two sides--a signature/specification and a defining body/RHS--then you get something that's both mathematically consistent and can be compiled.

For various reasons, we want to write programs that don't strictly follow this two-sides discipline. Here are some reasons why we may want to stray from the strict two-sided discipline:
* During program development, we'd like to be able to verify a caller before we have implemented the callee. In such a case, the callee will be body-less, that is, it will have a declaration but not an implementation. Without having "both sides", there's no guarantee that there exists something that will do what the callee is specified to do.
* In what may be considered a special case of the previous bullet, a loop also has a specification and a body. One might want to author the context around the loop before implementation the loop. In that state of development, the loop does not yet have "two sides".
* Some declared things will be provided outside the Dafny language. We will never have both sides of such declarations.
* A program may want to introduce assumptions that are not checked by the verifier. Such assumptions may be used temporarily to aid in debugging a proof. Or they may state properties that appear "obvious" and yet hard to prove, so an engineer may want to trade the risk of being wrong for the time it would take to do the proof. Or the assumption may be about the environment that Dafny cannot control, perhaps some extern code or the behavior of an end user. Note that an assumption in ghost contexts is as much of a risk to mathematical consistency as an assumption in compiled code--that is, an assumption in ghost code is in no way more acceptable than those in compiled contexts.
* Another use case of Dafny is to do some kind of modeling. This may involve declaring some uninterpreted types, uninterpreted functions, and giving some axioms--in other words, types, functions, and lemmas without bodies.

To deal with these exceptions of the strict two-sides discipline, Dafny allows
0. Body-less declarations (methods, functions, lemmas, consts, opaque types).
1. Body-less statements (loops, forall statements).
2. `assume` statements.

As a friendly tool, Dafny wants to provide useful feedback to users, pointing out possible accidental omissions or errors. This seems best to do for features that are subtle. For example, consider the following program snippet:

``` dafny
while i < 10
  i := i + 1;
```

Did the programmer really intend to write this body-less loop that's followed by an increment of `i`, or did the programmer intend for `i := i + 1;` to be in the body of the loop but forgot to write `{` and `}`? Here, Dafny currently always generates a warning (or, I think's it's an "information message") whenever a program contains a body-less loop.

For body-less declarations, it's less clear. Today, Dafny offers no feedback about these until the compiler is invoked. Sometimes, this can be confusing to a user. For example, some beginners don't realize there's a difference between a lemma with an empty body and a lemma with no body. As another example, our IDE constantly runs the verifier, but not the compiler, so a programmer may celebrate a verified development too soon, having forgotten to fill in the proofs of some important lemmas. I could imagine doing more here, and the present PR is going in that direction.

The presence of an `assume` in a program does not happen accidentally, like leaving off curly braces might. But a programmer might accidentally forget to remove `assume` statements. We guard against this by highlighting `assume` statements in a conspicuous color in the IDE, and the compiler gives an error if it finds an `assume` in the program (including an `assume` in ghost declarations, since such an `assume` undermines the compiler's ability to safely erase the code).

I don't think compiler errors about body-less loops is a problem. But always getting compiler errors about body-less declarations and `assume` statements is a problem. Therefore, we allow compiler directives that tell the compiler, "please ignore, because the programmer is taking willing to take responsibility for the assumption or missing body". One such directive is `:extern`. If a body-less declaration bears an `:extern` attribute, then I think we should not generate any warning or complaint.

Another such directive is `:axiom`. The initial motivation for this attribute for functions and lemmas. Note that body-less functions pose a risk, even if the function has no postcondition. It is because of the possibility that the function's result value is an empty type. Similarly, a method without a postcondition may break mathematical consistency if one of its out-parameters may have a type that is empty. Note also that this "empty output types" condition may arise from an output type being a type parameter that could be instantiated with an empty type. (So, Robin's suggestion in the comments above to not warn about postcondition-less, body-less methods needs to also look at the types of the method's out-parameters.)

Having an `:axiom` attribute on an `assume` statement could also be useful. It would say that the programmer takes responsibility that the assumed condition will always hold, which suddenly makes the `assume` statement compilable (by a no-op!). Thus, an `assume {:axiom}` would just be ignored by the compiler.

The discussion above also mentioned an `assert {:axiom}`. What is it good for? (Is this what the `assert`+`expect` idea has turned into?) Such a statement would seem really strange to me.

## Changes in warnings

Given the push by the present PR do change where we generate warnings or errors about things that may risk mathematical consistency, my opinion is the following.

* I think it would be a good idea (as suggested by this PR) to give some indication to the user about body-less declarations, _before_ the compiler. Maybe just a gentle warning or information message (as is done for body-less loops), rather than an error. The message could say "... add a body or mark as `:extern` or `:axiom`".
* I see no reason to support `:axiom` for body-less loops.
* It should not be possible for a body-less declaration to make it all the way through the resolver-verifier-compiler pipeline without an error, unless the declaration has an `:extern` or `:axiom` attribute.
* An `abstract` module is like a template that is to be filled in later. We should not generate any warnings or errors about missing bodies or `assume` statements there. Instead, such warnings or errors are generated for the modules that refine the abstract module.
* Even a postcondition-less function or method without a body is a risk for mathematical consistency. It only poses no risk if the result types are guaranteed to be nonempty.
* I see no reason that any part of the pipeline should raise an eyebrow about `assume` statements, except the compiler. The `assume` keyword is enough of an indication that the programmer intended to include the statement.
* I think it would be nice to support an `assume {:axiom}`. No warning would be generated, and the compiler would turn it into a no-op.
* I question the `assert {:axiom}` combination.

RustanLeino:
Should this PR be linked with Issue https://github.com/dafny-lang/dafny/issues/2961?

robin-aws:
@RustanLeino Thank you for the fantastic exposition. This needs to be captured somewhere other than a comment on a random PR. :)

I see your point about possibly empty return types, and agree we would need this PR to cover that case as well. But I believe we have a real usability/readability issue with the {:axiom} attribute. It naturally reads as "assume this thing is true without proof", which is why I preferred to attach it to the boolean expressions that are assumed true (e.g. `ensures` clauses in this case). But as you point out the real meaning of {:axiom} is "assume that a definition of this specified thing exists without proof", which is a broader definition but also changes the type of thing the term is applied to. It's especially dangerous when applied to a thing where the two meanings don't align, since `predicate {:axiom} Foo` seems to imply Foo is always true by the first meaning (and I've seen multiple users confused by this).

Is there a different term we could use for this concept? We could keep "axiom" for places where it is unambiguous such as lemmas (and only literally `lemma`s, for `ghost method`s I think it already gets confusing), but perhaps something closer to the spirit of "extern" otherwise? It might also help to put the annotation in the place where the body would normally go instead, so your options are logically to either provide a body or state "assume a body exists without proof". What about something like:

```dafny
predicate Foo(x: int, y: int)
  ensures Foo(x, y) ==> P(x, y)
  witness *
```

Using the "witness" keyword seems to align with the actual semantics better, and I think it's more likely to make users say "what?" and go look up the meaning, rather than incorrectly assume they know what {:axiom} means. I don't think this exact idea is perfect, mind you, since `witness *` on a subset type means "don't try to prove this type is non-empty", whereas here we want to assume a witness does exist without proof. The idea we should be trying to communicate is "instead of a body here, I want to intentionally just assume one exists".

@keyboardDrummer: Given this issue, I'm not inclined to dig ourselves deeper in the wrong direction by forcing users to attach {:axiom} to things like methods where it will cause more confusion, so I'd rather hold off a bit rather than make the breaking change for Dafny 4.0. I wonder if we could focus on providing "warnings" through the auditor instead, if we can consider that not a breaking change @atomb?

> I question the assert {:axiom} combination.

I double checked and this was just my own confusion - `assume {:axiom}` was the recent addition and I just misremembered it as `assert {:axiom}`. Please ignore entirely! :)

> I see no reason to support :axiom for body-less loops.

My understanding that if a body-less while loop is allowed by the verifier, for example, the loop condition must be false, and this means you don't need a witness that a compatible loop body exists. So I agree that there's no reason to support :axiom. But independently, I feel we should emit a warning for what is almost certainly a mistake as you point out - the program may be correct but readers will surely be confused!

RustanLeino:
I'm still chewing on whether a body-less predicate is better confirmed with `{:axiom}` or `witness *` or something else. Other words that come to mind are "postulate" or "admit!".

Body-less loops always generate an information message today (regardless of what follows them). Semantically, a body-less loop is similar (but not identical) to a body-less method. Whereas a body-less method has the same meaning as adding a method body `{ assume false; }`, a body-less loop is like adding a loop body `{ assume false; }` _and_ looking at the loop guard and loop specification to guess the intended loop assignment targets rather than inferring the loop assignment targets from the loop body.

I'm fine with delaying decisions about these axiom warnings until after the Dafny 4.0 release.

keyboardDrummer:
> I think it would be a good idea (as suggested by this PR) to give some indication to the user about body-less declarations, before the compiler. 

Indeed! Users have requested this indication to occur as part of `dafny verify`, which is the main driver behind this PR.

> It should not be possible for a body-less declaration to make it all the way through the resolver-verifier-compiler pipeline without an error, unless the declaration has an :extern or :axiom attribute.

I find it inconsistent that something related to verification does not produce an error when verifying, but does produce an error on execution.

A piece of code that for development purposes is meant to temporarily allow verification to pass, should also allow temporarily running the program. Running is also useful during development.

This PR suggests using regular warnings for axiomatic constructs such as bodyless declarations and assume statements. In CI, `--warnings-as-error` should always be used, so programs containing such constructs, if they are not annotated to mute the warning, will fail will CI. I think this is fine from a soundness perspective.

(Side-note, I'm thinking of adding a `dafny continuous-integration` command that'll run `dafny verify`, `dafny test` and `dafny measure-complexity`, and has `--warnings-as-error` on by default.) 

#### Alternative
Although I'm not in favour of this, we could let axiomatic constructs cause execution commands such as `dafny run` to fail by default. In that case, we should introduce the concept of a "verification warning", which produces a warning during verification, an error on execution, and that does not occur when using `--no-verify`.

This way, you can have bodyless ghost declarations and assume statements, and still do `dafny run --no-verify` during development to get your program to run without errors, and do `dafny verify` to get your program to verify with warnings. If you add annotations to mute the verification warning, then you can do `dafny verify` and get no warnings, and do `dafny run` without warnings or errors.

> I'm still chewing on whether a body-less predicate is better confirmed with {:axiom} or witness * or something else. Other words that come to mind are "postulate" or "admit!".

Although I'm generally not a fan of attributes, since here we have a concept that we want to apply to very different language constructs, both declarations and statements, and which feels like it only slightly changes the meaning of the constructs, I'm in favor of using an attribute here, and `{:axiom}` seems to indicate precisely what is meant.

> `predicate {:axiom} Foo` seems to imply Foo is always true by the first meaning (and I've seen multiple users confused by this).

Does this confusion only occur for predicates? It seems like a corner-case to me. We can add a hover hint for `{:axiom}` to explain what it does, so there is less chance of confusion.

RustanLeino:
@keyboardDrummer I think the verifier should not give an _error_ for things like assume statements or body-less declarations. (That would seem misleading to me. For example, the reason for having assume statements in the first place is to give information to the verifier, so that should not be a verification error. Also, if the program has not supplied a body for a loop, function, or method, then the verifier has not been asked to verify anything, so that shouldn't be an error, either.) However, this discussion has convinced me it's a good idea for the verifier to give _warnings_ about these.

Unless marked with `{:axiom}` or `{:extern}`, the compiler has to give errors for the same constructs. The reason for that is that the compiler is supposed to generate code, and it can't generate code for a body-less function/method/loop or an assume statement.

So, what about the following messages from the verifier and compiler?

|    | Verification | Compilation |
| ----- | ----- | ----- |
| body-less function or method without `{:axiom}` or `{:extern}` | warning | error |
| body-less function or method with `{:axiom}` or `{:extern}` | | |
| body-less loop | warning | error |
| `assume` statement without `{:axiom}` | warning | error |
| `assume` statement with `{:axiom}` | | compiled as no-op |
| `expect` statement | | |
| RHS-less const of non-auto-initializing type without `{:axiom}` or `{:extern}` | warning | error |
| RHS-less const of non-auto-initializing type with `{:axiom}` or `{:extern}` |  |   |
| opaque type without `(0)` or `(00)` | | compiled as some arbitrary type, e.g., `bool` |
| opaque type with `(0)` | | compiled as some arbitrary auto-init type, e.g., `bool` |
| opaque type with `(00)` | | |
| any opaque type with {:extern} | | |

Remarks about the table:
* "function or method" refers to any variation of these (predicate, twostate predicate, extreme predicate, lemma, extreme lemma, function by method)
* "non-auto-initializing type" refers to "a non-ghost const whose type is not an auto-init type, or a ghost const whose type is a possibly empty type"
* Except for the meaning of auto-initializing, whether or not something is declared as ghost is irrelevant for the table
* Loops do not support `{:axiom}` or `{:extern}`
* Opaque types do not support `{:axiom}`
* An abstract module is ignored by the compiler
* In an abstract module, the verifier still follows the table above, except that it does not give any warnings for body-less function or method and no warnings for RHS-less consts (since functions, methods, and consts can be filled in by a refining module). Note, the verifier _does_ warn for body-less loops and assume statements, since we have deprecated statement-level refinements.
* Except for opaque types, the auditor may choose to give warnings about any of the above, including those decorated with `{:axiom}` or `{:extern}`.

keyboardDrummer:
> Unless marked with {:axiom} or {:extern}, the compiler has to give errors for the same constructs. The reason for that is that the compiler is supposed to generate code, and it can't generate code for a body-less function/method/loop or an assume statement.

This argument doesn't hold for ghost code since the compiler doesn't need to generate code for those. Previously, the compiler would error on ghost declarations without a body. With this PR, you just get a warning from the resolver.

Could you comment on what I said previously:

> I find it inconsistent that something related to verification does not produce an error when verifying, but does produce an error on execution. A piece of code that for development purposes is meant to temporarily allow verification to pass, should also allow temporarily running the program. Running is also useful during development.

My main question is, can you accept allowing `dafny run ProgramUnderDevelopmentWithEmptyGhostMethodAndNoAxiomAttribute.dfy` to run (with a warning), or do you want the user to add another option to allow this, like `dafny run --no-verify ProgramUnderDevelopmentWithEmptyGhostMethodAndNoAxiomAttribute.dfy`

RustanLeino:
> > Unless marked with {:axiom} or {:extern}, the compiler has to give errors for the same constructs. The reason for that is that the compiler is supposed to generate code, and it can't generate code for a body-less function/method/loop or an assume statement.

> This argument doesn't hold for ghost code since the compiler doesn't need to generate code for those, right? Previously, the compiler would error on ghost declarations without a body. With this PR, you just get a warning from the resolver.

Specifically regarding the role of the compiler, how about if I changed the two occurrences of "generate code" to "generate code or justify erasure"? Would you then agree the description accurately reflects the responsibility of the compiler?

RustanLeino:
> ...
> I think the trade-off is between making it easier to run during development, and preventing running code in production when there are still assume-like constructs left over from development that weren't intended for production. I think the warning provide enough protection against the latter so we should allow `dafny run ProgramUnderDevelopmentWithEmptyGhostMethodAndNoAxiomAttribute.dfy`

Okay, I buy this. But only for body-less ghost functions and body-less ghost methods (that is, the first row in the table above). For the other 3 rows with "error" in the "compiler" column, I think a compiler error still seems appropriate. Here's my thinking, for the 4 compiler-error rows, in order:

* While a body-less ghost declaration carries the risk of having used `false` in the verification, the non-ghost code can still be compiled. If a user is invoking Dafny for such a program-under-development, they are probably expecting that the program will nevertheless be compiled.
* We don't have any machinery to synthesize the body for a body-less loop. And we couldn't very well let the compiler just omit the loop. (Exception: We could still compile the program if body-less loops occur only in ghost code.)
* I think `assume` statements are a special case, and I would not like a user to get the mistaken impression that `assume` ordinarily compiles to a no-op or to an `assert`. Therefore, giving a compiler error here makes sense to me, with an additional hint like "if you really want to compile this, then add the `{:axiom}` attribute to compile it as a no-op or change the `assume` into an `expect` to compile it into a run-time check". (Exception: Again, we could allow compilation if such an `assume` occurs in ghost code.)
* If a user omits the initialization of a `const`, I'm guessing the omission was not intentional. That is, I'm thinking the programmer would appreciate if we then gave an error, just like we'd give an error if the compiler is asked to compile a body-less non-ghost method. Also, if we did compile such an uninitialized constant, my guess is that the resulting program behavior may end up being far more mysterious than the behavior you'd get in the first bullet above. (Exception: If the `const` is ghost, we could still proceed with compilation.)

In the "Exception" cases I've noted above, the resolver would still give a warning, if I understand your suggested plan correctly.

keyboardDrummer:
> I think assume statements are a special case

>  I would not like a user to get the mistaken impression that assume ordinarily compiles to a no-op

Why not? It's a ghost statement right? It can only compile to a no-op.

> with an additional hint like "if you really want to compile this, then add the {:axiom} attribute to compile it as a no-op

That seems dangerous, because the `{:axiom}` annotation will remove even the warning and allow this `assume` statement to pass CI checks, even though the user might want to run their program with this assume statement only to help develop it.

I'm suggesting the following table (compared to what you were already convinced by, one more error has been turned into a warning):

|    | Verification | Compilation |
| ----- | ----- | ----- |
| body-less ghost function or method without `{:axiom}` or `{:extern}` | warning | *warning* |
| body-less compiled function or method without `{:axiom}` or `{:extern}` | warning | error |
| body-less ghost function or method without `{:axiom}` or `{:extern}` | | |
| body-less compiled function or method without `{:axiom}` or `{:extern}` | | error |
| body-less loop | warning | error |
| `assume` statement without `{:axiom}` | warning | *warning* |
| `assume` statement with `{:axiom}` | | |
| `expect` statement | | |

MikaelMayer:
> I'm suggesting the following table (compared to what you were already convinced by, one more error has been turned into a warning):
> 
> Verification	Compilation
> body-less ghost function or method without `{:axiom}` or `{:extern}`	warning	_warning_
> body-less compiled function or method without `{:axiom}` or `{:extern}`	warning	error
> body-less ghost function or method without `{:axiom}` or `{:extern}`		
> body-less compiled function or method without `{:axiom}` or `{:extern}`		error
> body-less loop	warning	error
> `assume` statement without `{:axiom}`	warning	_warning_
> `assume` statement with `{:axiom}`		
> `expect` statement

The third and fourth row you meant `with` instead of `without`, right?

I'm suggesting a few changes for your table. Notably, I still want the compiler to emit an error for assume statement without axiom, because they are a smell that I did not give more thought to an assumption. So at least I would like the compiler to force me to think of an assumption as something I really need to take ownership of, e.g. through `{:axiom}`

Second, we can count on the auditor for anything annotated with `{:axiom}`

Third, a ghost declaration labelled `{:extern}` should trigger a warning, because extern is only for compiled code.



|    | Verification | Compilation | Auditor |
| ----- | ----- | ----- |------|
| body-less ghost function or method without `{:axiom}` or `{:extern}` | warning | warning | *warning* |
| body-less compiled function or method without `{:axiom}` or `{:extern}` | warning | error |  *error* |
| body-less ghost function or method with `{:axiom}` | | | *report* |
| body-less compiled function or method with `{:axiom}`  | | *error* |  *error* |
| body-less ghost function or method with `{:extern}` | *warning* | *warning* | *warning* |
| body-less compiled function or method with `{:extern}`  | | |  |
| body-less loop | warning | error |  *error* |
| `assume` statement without `{:axiom}` | warning | *error* |  *error* |
| `assume` statement with `{:axiom}` | | |  *report* |
| `expect` statement | | | |

keyboardDrummer:
> The third and fourth row you meant with instead of without, right?

Yes, thanks

> Notably, I still want the compiler to emit an error for assume statement without axiom, because they are a smell that I did not give more thought to an assumption.

Indeed it's an indication that you didn't think about the assumption, that's why I think it's safest to let this emit a warning and not an error. 

When it emits a warning, it won't get committed to the repository. An error also prevents you from running the program. Forcing the user to add `{:axiom}` to allow them to run motivates them to add it, just so they can run, without giving enough thought to the axiom, which creates the risk of this bad axiom being committed to the repo.

MikaelMayer:
> When it emits a warning, it won't get committed to the repository. An error also prevents you from running the program. Forcing the user to add {:axiom} to allow them to run motivates them to add it, just so they can run, without giving enough thought to the axiom, which creates the risk of this bad axiom being committed to the repo.

That's a fair point of view. With your remark, I see how originally the behavior of a compiler error was implemented to prevent assumptions to go in production, but since we can always check the error codes now, emitting a warning would be sufficient.

Let me use your remark to reuse an existing alternative, I think we already have a way to run a program without taking into consideration verification errors: `--no-verify`. What if we made `assume X;` as a warning in the compiler `--no-verify` is present, and as an error if that option is not present? The error message could even be clear for that case. Here is a suggestion of the interaction on the CLI:

```
> dafny run --show-snippets HelloWorld.dfy

Dafny finished with 5 verified, 0 errors
5:    assume X;
      ^^^^^^^^
Error: Unbacked assumption. If you want to run the program anyway, use the `--no-verify` option.

> dafny run --no-verify --show-snippets HelloWorld.dfy

Dafny did not attempt to prove anything.
5:    assume X;
      ^^^^^^^^
Warning: Unbacked assumption. If you want to take full responsibility of this assumption and remove this warning, annotate it with  `{:axiom}`
Hello World!
> sed -i 's/assume/assume {:axiom}/g' HelloWorld.dfy
> dafny run --show-snippets HelloWorld.dfy
Dafny finished with 5 verified, 0 errors
Hello World!
```

MikaelMayer:
As a datapoint anecdote, this morning, I wanted to execute my model to ensure it makes sense but it has a few "assume false" on things I still wanted to prove. Even with --no-verify, the compiler complained and prevented me from running my program, so I had to manually add "{axiom}" to each assumption. That's painful, knowing some other assertions were axioms and it's going to mix everything up.

For sure, with --no-verify, we should accept assumptions, even if it means a warning.

Source/DafnyCore/AST/MethodOrFunction.cs
```diff
@@ -0,0 +1,41 @@
...
+  protected abstract bool Bodyless { get; }
+  protected abstract string TypeName { get; }
+
+  public bool IsVirtual => EnclosingClass is TraitDecl && !IsStatic;
+
+  public virtual void Resolve(Resolver resolver) {
+    if (Bodyless && !IsVirtual && !this.IsExtern() && !this.IsExplicitAxiom()) {
+      foreach (var ensures in Ens) {
+        if (!ensures.IsExplicitAxiom() && !resolver.Options.Get(CommonOptionBag.AllowAxioms)) {
+          resolver.Reporter.Warning(MessageSource.Resolver, ErrorDetail.ErrorID.None, ensures.Tok,
```
MikaelMayer, line 32:
```suggestion
          resolver.Reporter.Warning(MessageSource.Verifier, ErrorDetail.ErrorID.None, ensures.Tok,
```
Let's switch to verification warnings rather than resolution warnings. That way,
1) we can still use warnings as errors while letting verification run.
2) I can use this information for changing the verification context of gutter icons

Source/DafnyCore/AST/Statements/AssignSuchThatStmt.cs
```diff
@@ -76,6 +76,10 @@ public AssignSuchThatStmt(RangeToken rangeToken, List<Expression> lhss, Expressi
     base.Resolve(resolver, resolutionContext);
 
     if (AssumeToken != null) {
+      if (!resolver.Options.Get(CommonOptionBag.AllowAxioms) && !AssumeToken.IsExplicitAxiom()) {
+        resolver.Reporter.Warning(MessageSource.Resolver, ErrorDetail.ErrorID.None, AssumeToken.Token, "assume keyword in assign-such-that statement has no {:axiom} annotation");
```
MikaelMayer, line 80:
```suggestion
        resolver.Reporter.Warning(MessageSource.Verifier, ErrorDetail.ErrorID.None, AssumeToken.Token, "assume keyword in assign-such-that statement has no {:axiom} annotation");
```

Source/DafnyCore/AST/Statements/AssumeStmt.cs
```diff
@@ -31,4 +31,12 @@ public AssumeStmt(RangeToken rangeToken, Expression expr, Attributes attrs)
   public bool SetIndent(int indentBefore, TokenNewIndentCollector formatter) {
     return formatter.SetIndentAssertLikeStatement(this, indentBefore);
   }
+
+  public override void Resolve(Resolver resolver, ResolutionContext context) {
+    base.Resolve(resolver, context);
+
+    if (!resolver.Options.Get(CommonOptionBag.AllowAxioms) && !this.IsExplicitAxiom()) {
+      resolver.Reporter.Warning(MessageSource.Resolver, ErrorDetail.ErrorID.None, Tok, "assume statement has no {:axiom} annotation");
```
MikaelMayer, line 39:
```suggestion
      resolver.Reporter.Warning(MessageSource.Verifier, ErrorDetail.ErrorID.None, Tok, "assume statement has no {:axiom} annotation");
```

MikaelMayer (changes requested):
I did not go through all the files as others have, but I have only one request. These new warnings have nothing to do with resolution but only with verification, so I would like (even if it's in the resolver phase), that they are labelled as such. It will also make it easier down the road to solve #3089 by implementing #3116 in a more generic way than I did.

keyboardDrummer:
> That's a fair point of view. With your remark, I see how originally the behavior of a compiler error was implemented to prevent assumptions to go in production, but since we can always check the error codes now, emitting a warning would be sufficient.

What do you mean by "since we can always check the error codes now" ? My reasoning was since in CI users should always use `warn-as-errors`, warnings won't be merged.

> Let me use your remark to reuse an existing alternative, I think we already have a way to run a program without taking into consideration verification errors: `--no-verify`. What if we made `assume X;` as a warning in the compiler `--no-verify` is present, and as an error if that option is not present?

That alternative means that if you have an `assume X`, verification can pass without errors, but code generation may still produce verification related errors. I think it's simpler to keep those two completely separated. Technically, I'd prefer to remove all verification related properties from the AST before moving to code generation.

What's the downside of what I proposed?

MikaelMayer:
> What do you mean by "since we can always check the error codes now" ? My reasoning was since in CI users should always use `warn-as-errors`, warnings won't be merged.

Yes I wasn't very precise in my choice of words but that's what I had in mind.

> That alternative means that if you have an `assume X`, verification can pass without errors, but code generation may still produce verification related errors. I think it's simpler to keep those two completely separated. Technically, I'd prefer to remove all verification related properties from the AST before moving to code generation.
> 
> What's the downside of what I proposed?

In your alternative , bodiless loops have this exact same property, (compiler to emit errors) why would you have an inconsistency?
Also, my workflow is typically work in the IDE, and when I want to run my code while in dev, I automatically use "git run --no-verify", because for me Dafny is sound, so unbacked assumptions should not let Dafny execute the program by default To me I would consider it to be a soundness issue if a code containing "assume" without the {:axiom:} attribute could execute, even if you considered "dafny run" as dev mode, unless the user assume their intentions via the attribute or via --no-verify.

Feel free to ask more opinions on that.

keyboardDrummer:
> Also, my workflow is typically work in the IDE, and when I want to run my code while in dev, I automatically use "git run --no-verify"

That works. I think it has the slight downside that you will have to run both `dafny verify` and `dafny run --no-verify` to get all the information you want during development, while with this PR's proposal you would only do `dafny run`.

> I would consider it to be a soundness issue if a code containing "assume" without the {:axiom:} attribute could execute, even if you considered "dafny run" as dev mode, unless the user assume their intentions via the attribute or via --no-verify.

What if we make `--warn-as-errors=true` the default, which I think is a good idea anyways (@robin-aws thoughts?), so only `dafny run --warn-as-errors=false` will allow you to run with `assume {no-axiom}` ? There could be an alias `--dev` as a shorthand for `--warn-as-errors=false`

keyboardDrummer:
Maybe we can have a (possibly hidden) option that determines whether non-axiom assumptions are errors or warnings, and let them be warnings when using `dafny verify` and errors when doing `dafny run`, then have them not turn up when doing `--no-verify`, so `dafny run --no-verify` works.

MikaelMayer:
> Maybe we can have a (possibly hidden) option that determines whether non-axiom assumptions are errors or warnings, and let them be warnings when using dafny verify and errors when doing dafny run, then have them not turn up when doing --no-verify, so dafny run --no-verify works.

You mean, "run" or "verify" and "--no-verify" would actually modify this (possibly hidden) option's default? That sound great. I think you have a good implementation there.

keyboardDrummer:
> > Maybe we can have a (possibly hidden) option that determines whether non-axiom assumptions are errors or warnings, and let them be warnings when using dafny verify and errors when doing dafny run, then have them not turn up when doing --no-verify, so dafny run --no-verify works.
> 
> You mean, "run" or "verify" and "--no-verify" would actually modify this (possibly hidden) option's default? That sound great. I think you have a good implementation there.

Yes. Cool ! : )

RustanLeino (changes requested):
Looks pretty good to me.

In addition to the `:| assume` statement (which is handled), there's also `:- assume` (which I didn't see code for).

Regarding the refactorings to move resolution code into `Resolve` methods of the particular AST nodes:  This is just for the legacy resolver, right? Were changes also needed for the new resolver?

Source/IntegrationTests/TestFiles/LitTests/LitTest/ast/function.dfy.expect
```diff
@@ -0,0 +1,15 @@
...
+function.dfy(8,50): Warning: This ensures clause is part of a bodyless function. Add the {:axiom} attribute to it or the enclosing function to suppress this warning
+
+Dafny program verifier finished with 9 verified, 0 errors
+function.dfy(7,33): Warning: This ensures clause is part of a bodyless function. Add the {:axiom} attribute to it or the enclosing function to suppress this warning
+function.dfy(7,47): Warning: This ensures clause is part of a bodyless function. Add the {:axiom} attribute to it or the enclosing function to suppress this warning
+function.dfy(8,36): Warning: This ensures clause is part of a bodyless function. Add the {:axiom} attribute to it or the enclosing function to suppress this warning
+function.dfy(8,50): Warning: This ensures clause is part of a bodyless function. Add the {:axiom} attribute to it or the enclosing function to suppress this warning
+
+Dafny program verifier finished with 9 verified, 0 errors
+function.dfy(7,9): Error: Function _module._default.Compiled has no body so it cannot be executed
```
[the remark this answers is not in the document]
keyboardDrummer: I think at some point we should adopt something other than 'compiled' here, because I think compiled is not correct. To me compilation means translation to another language, and the situation here is not exactly that. Even if Dafny had an interpreter, you still would not be able to run/execute this function. Also, the function has a contract, so it can be translated to Boogie. In other terms, it can be 'compiled' to Boogie.

keyboardDrummer:
> Regarding the refactorings to move resolution code into Resolve methods of the particular AST nodes: This is just for the legacy resolver, right? Were changes also needed for the new resolver?

Oof, I missed that, and so did the tests. I think we should drive towards removing the old resolve code ASAP because this duplication will keep biting us.

Source/IntegrationTests/TestFiles/LitTests/LitTest/ast/function.dfy.expect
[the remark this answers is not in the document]
keyboardDrummer: Anyways, I've changed it all to "compiled"

RustanLeino:
Speaking of the two resolvers, I should have thought of this before: Can you move this warning code to a separate pass that will run after "pass 0" (probably very late in the `ResolveTopLevelDecls_Core` method)?

keyboardDrummer:
> Speaking of the two resolvers, I should have thought of this before: Can you move this warning code to a separate pass that will run after "pass 0" (probably very late in the `ResolveTopLevelDecls_Core` method)?

I don't think this has enough weight to deserve a separate pass at this point, although indeed it would avoid the duplicate resolver issue.

RustanLeino:
> > Speaking of the two resolvers, I should have thought of this before: Can you move this warning code to a separate pass that will run after "pass 0" (probably very late in the `ResolveTopLevelDecls_Core` method)?
> 
> I don't think this has enough weight to deserve a separate pass at this point, and technically it does not require a separate pass, it does not depend on anything being done first, nor does anything else depend on this being done first, although indeed a separate pass would avoid the duplicate resolver issue.

In such a case, there may be some other pass that the new checks could piggyback on.

keyboardDrummer:
> > > Speaking of the two resolvers, I should have thought of this before: Can you move this warning code to a separate pass that will run after "pass 0" (probably very late in the `ResolveTopLevelDecls_Core` method)?
> > 
> > 
> > I don't think this has enough weight to deserve a separate pass at this point, and technically it does not require a separate pass, it does not depend on anything being done first, nor does anything else depend on this being done first, although indeed a separate pass would avoid the duplicate resolver issue.
> 
> In such a case, there may be some other pass that the new checks could piggyback on.

Would that improve anything?

I think unless we add an AxiomWarningsPass, these warnings are fine where they are.

RustanLeino (dismissed):

Source/DafnyCore/AST/Statements/Verification/AssumeStmt.cs
```diff
@@ -32,4 +32,12 @@ public AssumeStmt(RangeToken rangeToken, Expression expr, Attributes attrs)
   public bool SetIndent(int indentBefore, TokenNewIndentCollector formatter) {
     return formatter.SetIndentAssertLikeStatement(this, indentBefore);
   }
+
+  public override void GenResolve(INewOrOldResolver resolver, ResolutionContext context) {
+    base.GenResolve(resolver, context);
+
+    if (!resolver.Options.Get(CommonOptionBag.AllowAxioms) && !this.IsExplicitAxiom()) {
+      resolver.Reporter.Warning(MessageSource.Resolver, ResolutionErrors.ErrorId.r_assume_statement_without_axiom, Tok, "assume statement has no {:axiom} annotation");
```
MikaelMayer, line 40:
```suggestion
      resolver.Reporter.Warning(MessageSource.Verifier, ResolutionErrors.ErrorId.r_assume_statement_without_axiom, Tok, "assume statement has no {:axiom} annotation");
```

Source/DafnyCore/AST/Members/MethodOrFunction.cs
```diff
@@ -0,0 +1,49 @@
...
+  protected abstract string TypeName { get; }
+
+  public bool IsVirtual => EnclosingClass is TraitDecl && !IsStatic;
+  public bool IsAbstract => EnclosingClass.EnclosingModuleDefinition.ModuleKind != ModuleKindEnum.Concrete;
+
+  public virtual void Resolve(ModuleResolver resolver) {
+    if (Bodyless && !IsVirtual && !IsAbstract && !this.IsExtern(resolver.Options) && !this.IsExplicitAxiom()) {
+      foreach (var ensures in Ens) {
+        if (!ensures.IsExplicitAxiom() && !resolver.Options.Get(CommonOptionBag.AllowAxioms)) {
+          resolver.Reporter.Warning(MessageSource.Resolver, ResolutionErrors.ErrorId.none, ensures.Tok,
```
MikaelMayer, line 40:
```suggestion
          resolver.Reporter.Warning(MessageSource.Verifier, ResolutionErrors.ErrorId.none, ensures.Tok,
```

Source/DafnyCore/AST/Statements/Assignment/AssignOrReturnStmt.cs
```diff
@@ -237,6 +238,18 @@ public AssignOrReturnStmt(RangeToken rangeToken, List<Expression> lhss, ExprRhs
     resolver.EnsureSupportsErrorHandling(Tok, firstType, expectExtract, KeywordToken != null);
   }
 
+  public void ResolveKeywordToken(INewOrOldResolver resolver, ResolutionContext resolutionContext) {
+    if (KeywordToken == null) {
+      return;
+    }
+
+    if (!resolver.Options.Get(CommonOptionBag.AllowAxioms) && KeywordToken.Token.val == "assume" && !KeywordToken.IsExplicitAxiom()) {
+      resolver.Reporter.Warning(MessageSource.Resolver, ResolutionErrors.ErrorId.none, KeywordToken.Token, "assume keyword in update-with-failure statement has no {:axiom} annotation");
```
MikaelMayer, line 247:
```suggestion
      resolver.Reporter.Warning(MessageSource.Verifier, ResolutionErrors.ErrorId.none, KeywordToken.Token, "assume keyword in update-with-failure statement has no {:axiom} annotation");
```

robin-aws:
For the record (and thanks to @keyboardDrummer for the ping that this was about to go in), although I still don't love the particular way Dafny uses the word "axiom" in attributes for readability, the changes in this PR are orthogonally good to make.

Source/DafnyCore/AST/Statements/Verification/AssumeStmt.cs
keyboardDrummer: > ResolutionErrors to VerificationErrrors for clarity

I've moved it to `AssumeStmt`, where the warning is produced.

MikaelMayer (approved):
Thanks for accepting my request to make the messages part of the verifier!

MikaelMayer (approved):

Issue #3468, opened 2023-02-06
Bring new CLI options up to par with legacy options


atomb:
### Summary

Currently, the new top-level commands for Dafny provide most but not all of the frequently-used features of the tool. This issue is to track remaining options to fill in before 4.0.

### Background and Motivation

Particularly when working on proof debugging and solver performance investigation, several of the most useful options are missing.

### Proposed Feature

- [x] Add `--isolate-assertions` to `verify` (not just `check-complexity`)
- [x] Equivalent of `/verificationLogger`
  - Suggestion: `--log-format` and `--log-file` options to `verify`
- [x] Equivalent of `/vcsLoad`
  - Suggestion: `--solver-load` option to `verify`
- [x] Equivalent of `/errorLimit`
  - Suggestion: `--error-limit` option to `resolve` [DRC: to all commands] -- this one needed to support testing with new CLI
- [x] Equivalent of `/rLimit`
  - Suggestion: `--resource-limit` option to `verify`
- [ ] Equivalent of `/trace`
  - Suggestion `--verbosity` with `silent`, `quiet`, and `trace` options
- [x] Equivalent of `/proverDll`
  - Suggestion `--solver-plugin` option to `verify`
- [x] Equivalent of `/proverLog`
  - Suggestion: `--solver-log` option to `verify`
- [x] Equivalent of `/proverOpt`
  - Suggestion `--solver-option` option to `verify`
- [x] Add `--stdin` to `resolve` (and everything else, transitively)

DRC: Anything for verify needs to be available on translate/build/run in case those verify
DRC: Others to consider:
--random-seed
--help-attr
--deprecation
something like /noCheating:0 to allow mix of assert and expect and assume
something to easily /extractCounterexample

I fear if we don't supply what users need to tinker with proof performance, they may just not use the new CLI.
So maybe also
--legacy=" ... string of old options in old format "
just in case

davidcok:
My add: --use-basename-for-filename should not be jsut a developer option -- I would find it essential for testing as a user.

Also --error-limit, --random-seed

keyboardDrummer:
`/vcsLoad` is already covered by `--cores`

I would suggest `--solver-log` to be a hidden option

`/trace` is poorly defined. The documentation reads "blurt out various debug trace information". We should think about what it's used for and how to expose that with a good UI.

What does `/proverDll` add to the already existing `--solver-path` ?

davidcok:
[...]

PR #3568, opened 2023-02-17
Add several new-style command-line options

atomb:
* `--error-limit` is available everywhere
* `--solver-log` is available for `verify` and `measure-complexity`
* `--solver-plugin` is available for `verify` and `measure-complexity`
* `--resource-limit` is available for `verify` and `measure-complexity`
* `--isolate-assertions` and `--format` are also available for `verify` with `--format` renamed `--log-format`

Addresses #3468


davidcok (approved):

davidcok (dismissed):

davidcok (approved):

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -2,11 +2,15 @@
...
 using System.IO;
 using System.Linq;
+using Microsoft.Boogie;
 
 namespace Microsoft.Dafny;
 
 public class CommonOptionBag {
 
+  public static readonly Option<int> ErrorLimit =
+    new("--error-limit", () => 5, "Set the maximum number of errors to report (0 for unlimited).");
```
keyboardDrummer, line 12: Shouldn't this be called `verification-error-limit`, since it limits only verification errors?

Also, the use-case of the option is unclear to me. If it's only used for this then it would be better if it's a hidden option. What's the reason that the default is not unlimited? Is that to improve performance? If yes, then it might be good to mention performance in the description of this option.

Right now, if I read the description of this option then I'm confused as to why the reported errors are limited to begin with.

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -45,6 +49,21 @@ public class CommonOptionBag {
...
+
+The `trx` and `csv` formats automatically choose an output file name by default, and print the name of this file to the console. The `text` format prints its output to the console by default, but can send output to a file given the `LogFileName` option.
+
+The `text` format also includes a more detailed breakdown of what assertions appear in each assertion batch. When combined with the {CommonOptionBag.IsolateAssertions.Name} option, it will provide approximate time and resource use costs for each assertion, allowing identification of especially expensive assertions.".TrimStart()) {
+    ArgumentHelpName = "configuration"
+  };
+
+  public static readonly Option<uint> SolverResourceLimit = new("--resource-limit", @"Specify the maximum resource limit (rlimit) value to pass to Z3. Multiplied by 1000 before sending to Z3.");
+  public static readonly Option<string> SolverPlugin = new("--solver-plugin", @"Specify a plugin to use to solve verification conditions (instead of an external Z3 process).");
```
keyboardDrummer, line 64: I would use the name `--boogie-plugin` instead of `--solver-plugin` because as far as I understand, the reason we have this option is that is supports the exact same plugin format as Boogie does, so users can re-use their Boogie plugins with Dafny. If you don't care about re-using a Boogie plugin and you want to write a Dafny specific solver plugin, I would recommend using `--plugin` instead since it provides the full Dafny plugin interface.

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -45,6 +49,21 @@ public class CommonOptionBag {
...
+  public static readonly Option<string> SolverLog = new("--solver-log", @"Specify a file to use to log the SMT-Lib text sent to the solver.");
```
keyboardDrummer, line 65: Based on this comment: https://github.com/dafny-lang/dafny/pull/3537#issuecomment-1433501171

"The option that actually came up in our use case was /normalizeOptions:0, which makes the generated SMT files much easier to debug. Hopefully in the new CLI there remains a way to do this."

It seems like just exposing the raw SMT doesn't help users much. I suggest adding `normalizeOptions` as well and making both that and this hidden options, since they seem geared towards people who are developing Dafny or extensions to Dafny, not people who are only consuming Dafny.

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -45,6 +49,21 @@ public class CommonOptionBag {
...
+
+The `trx` and `csv` formats automatically choose an output file name by default, and print the name of this file to the console. The `text` format prints its output to the console by default, but can send output to a file given the `LogFileName` option.
+
+The `text` format also includes a more detailed breakdown of what assertions appear in each assertion batch. When combined with the {CommonOptionBag.IsolateAssertions.Name} option, it will provide approximate time and resource use costs for each assertion, allowing identification of especially expensive assertions.".TrimStart()) {
+    ArgumentHelpName = "configuration"
+  };
+
+  public static readonly Option<uint> SolverResourceLimit = new("--resource-limit", @"Specify the maximum resource limit (rlimit) value to pass to Z3. Multiplied by 1000 before sending to Z3.");
```
keyboardDrummer, line 63: How is a user supposed to use this option? There's no guidance for how to pick a value.

Source/DafnyCore/Options/CommonOptionBag.cs
atomb: I also think the default should be unlimited, and that perhaps the option isn't needed at all. My goal for now, though, was to replicate existing functionality in the new CLI, rather than change it. Since the default is to limit the number of errors presented, having an option to raise or eliminate that limit seems important.

Calling it `--verification-error-limit` makes sense. I think I'm going to have another PR to add some more options and could change it then.

Source/DafnyCore/Options/ICommandSpec.cs
```diff
@@ -59,13 +62,15 @@ public interface ICommandSpec {
...
   public static IReadOnlyList<Option> ParserOptions = new List<Option>(new Option[] {
+    CommonOptionBag.StdIn,
     BoogieOptionBag.Cores,
     CommonOptionBag.Libraries,
     CommonOptionBag.Plugin,
     CommonOptionBag.Prelude,
     Function.FunctionSyntaxOption,
     CommonOptionBag.QuantifierSyntax,
     CommonOptionBag.UnicodeCharacters,
+    CommonOptionBag.ErrorLimit,
```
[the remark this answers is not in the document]
atomb: Yes, you're right. I was mistakenly thinking it affected other sorts of errors, too.

Source/DafnyCore/Options/CommonOptionBag.cs
atomb: The only kind of plugin Boogie supports is a solver plugin, and I'd prefer that Dafny users not need to know about the existence of Boogie. So I'd prefer to leave it with the current naming.

Source/DafnyCore/Options/CommonOptionBag.cs
atomb: Many Dafny users I've talked to want to be able to look at the generated SMT (or store it to send to a different solver). I wouldn't entirely be opposed to making it hidden, but I'd want to make sure that the standard help has some sort of message in the spirit of "there are other, more advanced options, too. Used --help-hidden to see them". It's used often enough that I'd be reluctant to make it hard to find.

Source/DafnyCore/Options/CommonOptionBag.cs
atomb: I'll add more thorough help text in a later PR.

Source/DafnyCore/Options/CommonOptionBag.cs
davidcok: It can be quite important. If you encode a method as one big VC, then it reports one assertion failure. To get more assertion failures you adjust the VC to prevent the failure you have already seen and rerun. You can continue this until no more failures are found. That iteration takes time and a user might well want to stop after some small number. The problem is that the order in which results are found is non-deterministic. There are arguments for a default either way -- when I was working with production code, I would set it to a small number for a faster debug cycle. No limit only works for users if the results are presented as they are found. I vote to keep it. Like Aaron, I did not know it was only for verification errors.

Source/DafnyCore/Options/CommonOptionBag.cs
davidcok: I understand that some options are used only for development debugging of solver performance. But in general having hidden options smacks of being patriarchal "we know best what you need" to the users. Many tools have lots of options -- for myself (as a user) it is helpful to browse the list to see what is possible, but not dig into them unless I need them. I would bias towards not hiding options unless they real are deeply technical.

Source/DafnyCore/Options/CommonOptionBag.cs
keyboardDrummer: > I'd prefer that Dafny users not need to know about the existence of Boogie. 

If a user doesn't know about Boogie then they should use `--plugin` to add C# solver code. Inside their plugin they can set `DafnyOptions.O.TheProverFactory = ` and then there's no need for a `--solver-plugin` option. What am I missing?

Source/DafnyCore/Options/CommonOptionBag.cs
keyboardDrummer: > If you encode a method as one big VC, then it reports one assertion failure. To get more assertion failures you adjust the VC to prevent the failure you have already seen and rerun.

To be clear, I think this is what Boogie does automatically.

> That iteration takes time and a user might well want to stop after some small number.

So it's indeed a performance related option.

> The problem is that the order in which results are found is non-deterministic. 

Why is that a problem?

> I vote to keep it. Like Aaron, I did not know it was only for verification errors.

If we keep it then we should explain why we have it.

davidcok: The non-determinism causes this problem. Suppose you run with an error-limit of 1, find a problem and "fix" it. Then run again. You get a different problem. Did you fix the first one or did you randomly get a different problem?

Non-determinism is also a problem for tests.(And is why I write tests with just one problem, though that does not capture the full range of uses)

And yes, I would count it as a (user) performance optimization.

The user manual says this: limits the number of verification errors reported per procedure. Default is 5; 0 means as many as possible; a small positive number runs faster but a large positive number reports more errors per run

To me that concisely captures the value and tradeoff of the option

Source/DafnyCore/Options/CommonOptionBag.cs
davidcok: Has anyone ever used a solver plugin? And are there any instructions on how to do it? Although there is a SMT-lib standard, solvers differ quite a bit in ways that end up mattering that this would be a very non-trivial exercise.

davidcok: I'm inclined on this one to not expose it in the new CLI and see if anyone complains.

Source/DafnyCore/Options/CommonOptionBag.cs
keyboardDrummer: > But in general having hidden options smacks of being patriarchal "we know best what you need" to the users.

I don't feel that way. There are many options that are only relevant to Dafny language developers, take printing out the resolved Dafny program for example. As a user, for each option in the help that I read that isn't useful to me, I will be less excited to keep reading the help, and there's a higher chance I will miss options that are relevant.

> I'd want to make sure that the standard help has some sort of message in the spirit of "there are other, more advanced options, too. Used --help-hidden to see them". It's used often enough that I'd be reluctant to make it hard to find.

That sounds good, although I wouldn't say 'more advanced options' but 'options geared towards helping develop the Dafny language toolchain'

> (or store it to send to a different solver)

Could this be part of a Dafny toolchain developer use-case?

PR #3573, opened 2023-02-19
Marking options as allowing override behavior

### LINKED ISSUE #3571: Erroneous command-line error for repeated option
### Dafny version

3.11.0+dev

_No response_

```code
> dafny resolve --function-syntax:4 --function-syntax:4 x.dfy
Option '--function-syntax' expects a single argument but 2 were provided.
```

### What happened?

The second instance of the option should replace the first. This option should not accumulate a list of values.

This is a change in behavior from legacy options.

In fact it is common, even in testing, to have some built-in defaults, that then can be overridden by additional arguments.
Common also in other tools.
Looking at the System.CommandLine docs, it seems one has to specially configure options for overriding, which makes me suspect that the problem applies to many options. Indeed --unicode-char has the same behavior.

Mac

davidcok:
Fixes #3571


keyboardDrummer (commented):
A downside of this is that users might accidentally specify one option multiple times, and now won't be alerted to that.

It seems the only use-case of this change is to accommodate our way of testing. What's the testing issue we're trying to solve? Can we solve it in a different way?

Source/DafnyCore/Options/BoogieOptionBag.cs
```diff
@@ -11,21 +11,19 @@ public static class BoogieOptionBag {
...
   };
 
   public static readonly Option<uint> Cores = new("--cores", result => {
-    if (result.Tokens.Count != 1) {
-      result.ErrorMessage = $"Expected only one value for option {Cores.Name}";
-      return 1;
-    }
 
-    var value = result.Tokens[0].Value;
+    var value = result.Tokens[result.Tokens.Count - 1].Value;
```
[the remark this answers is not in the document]
davidcok: Made the change.

davidcok:
I'd say the reason for this change is so that dafny behaves like most other tools and users are not surprised by its behavior.
It is not just our way of testing. In my experience a common way of using tools is to create an alias that builds in desired defaults, which you can then override by adding overriding options when you use the alias. I agree it is not going to affect IDE users. Robin wanted it fixed for 3.12, so I researched and put in a fix promptly.

Source/DafnyCore/Options/BoogieOptionBag.cs
```diff
@@ -11,21 +11,19 @@ public static class BoogieOptionBag {
   public static readonly Option<IEnumerable<string>> BoogieFilter = new("--boogie-filter", @"
 (experimental) Only check proofs whose Boogie name is matched by pattern <p>. This option may be specified multiple times to match multiple patterns. The pattern <p> may contain * wildcards which match any character zero or more times. If you are unsure of how Boogie names are generated, please pre- and postfix your pattern with a wildcard to enable matching on Dafny proof names."
     .TrimStart()) {
-    ArgumentHelpName = "pattern"
+    ArgumentHelpName = "pattern",
+    AllowMultipleArgumentsPerToken = true,
```
keyboardDrummer, line 15: Could you set this for all options? 

You can write something like 
```c#
    foreach (var option in Commands.SelectMany(c => c.Options)) {
        option. AllowMultipleArgumentsPerToken = true;
    }
```

In `CommandRegistry.Create`

davidcok: Made this change

Source/DafnyDriver/Commands/CommandRegistry.cs
```diff
@@ -67,7 +67,12 @@ static class CommandRegistry {
       if (!allowHidden) {
         option.IsHidden = false;
       }
+      option.AllowMultipleArgumentsPerToken = true;
     }
+    RunCommand.Inputs.AllowMultipleArgumentsPerToken = false;
+    CommonOptionBag.Libraries.AllowMultipleArgumentsPerToken = false;
```
keyboardDrummer, line 73: Why are some of these false? Is that because they allow multiple values as inputs? Can we add that check on line 70 instead?

davidcok: No - because all of this setting is false by default; besides why check them on each iteration of the loop when they can just be set afterwards. 

These three are false because that is the current behavior.
We could say that they take a comma-separated list instead, as I've proposed to Aaron for --log-format. We have already implemented comma-separated-lists for --libraries, but not restricted its use. I can't see restricting --input as it is used in dafny run

keyboardDrummer: Maybe you could check whether the .Arity is `ZeroOrMore` or `OneOrMore`, and if so on that keep `AllowMultipleArgumentsPerToken` at false?

davidcok: Made this change.

keyboardDrummer (approved):

PR #3577, opened 2023-02-20
Fix formatting regression

### LINKED ISSUE #3563: Between 3.11.0 and now indentation regressed
### Dafny version

3.11.0+dev

```dafny
const c :=
           1111111111111111111111111111111111111111
```

```code
dafny format --print
```

### What happened?

This used to format as
```dafny
const c :=
  1111111111111111111111111111111111111111
```
but now is
```dafny
const c :=
1111111111111111111111111111111111111111
```
which seems like a regression to me.

Mac

MikaelMayer:
This PR fixes a formatting regression by actually implementing a missing formatting support for DatatypeValue.

Fixes #3563

### DIFF
--- a/Source/DafnyCore/AST/Expressions/Expressions.cs
+++ b/Source/DafnyCore/AST/Expressions/Expressions.cs
@@ -922,7 +922,7 @@ public class StringLiteralExpr : LiteralExpr {
   }
 }
 
-public class DatatypeValue : Expression, IHasUsages, ICloneable<DatatypeValue> {
+public class DatatypeValue : Expression, IHasUsages, ICloneable<DatatypeValue>, ICanFormat {
   public readonly string DatatypeName;
   public readonly string MemberName;
   public readonly ActualBindings Bindings;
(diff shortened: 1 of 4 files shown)


Source/DafnyCore/AST/MemberDecls.cs
```diff
@@ -128,7 +128,7 @@ public Field(RangeToken rangeToken, Name name, bool hasStaticKeyword, bool isGho
 
   public bool SetIndent(int indentBefore, TokenNewIndentCollector formatter) {
     formatter.SetOpeningIndentedRegion(StartToken, indentBefore);
-    formatter.SetClosingIndentedRegion(EndToken, indentBefore);
+    formatter.SetIndentations(EndToken, below: indentBefore);
```
MikaelMayer, line 131: Before, it would set the indent of the last token to (above:Indent + 2, inline:Indent, below:Indent), which works for sets and sequences, but not for constants.
Now I only set the indentation below that last token, which is the only thing really needed.

Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -986,6 +986,10 @@ public DatatypeValue(IToken tok, string datatypeName, string memberName, List<Ex
   }
 
   public IToken NameToken => tok;
+  public bool SetIndent(int indentBefore, TokenNewIndentCollector formatter) {
+    formatter.SetMethodLikeIndent(StartToken, OwnedTokens, indentBefore);
```
MikaelMayer, line 990: This method is the "default" formatting heuristic, which can take into account type parameters, named arguments, etc.

keyboardDrummer (approved):

PR #3581, opened 2023-02-21
Dafny 4: edits to Refinement and Attributes sections

davidcok:

### DIFF
--- a/docs/DafnyRef/Attributes.md
+++ b/docs/DafnyRef/Attributes.md
@@ -106,7 +106,7 @@ be added.
 ```
 
 ### 11.1.2. `{:nativeType}` {#sec-nativetype}
-The `{:nativeType}` attribute may only be used on a ``NewtypeDecl``
+The `{:nativeType}` attribute is only recognized by a `newtype` declaration
 where the base type is an integral type or a real type. For example:
 
 <!-- %check-resolve Attributes.1.expect -->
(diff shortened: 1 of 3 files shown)


docs/DafnyRef/Attributes.md
```diff
@@ -117,10 +117,10 @@ newtype {:nativeType "byte"} bad_ubyte = x : int | 0 <= x < 257 // Fails
 
 It can take one of the following forms:
 
-* `{:nativeType}` - With no parameters it has no effect and the ``NewtypeDecl``
-will have its default behavior which is to choose a native type that can hold any
+* `{:nativeType}` - With no parameters it has no effect and the declaration
+will have its default behavior, which is to choose a native type that can hold any
 value satisfying the constraints, if possible, otherwise BigInteger is used.
```
stefan-aws, line 122:
```suggestion
value satisfying the constraints, if possible, and otherwise to use BigInteger.
```

docs/DafnyRef/Topics.md
```diff
@@ -21,7 +21,7 @@ type List<T>
 function Elements(list: List): set
 ```
 In the latter case, Dafny knows that the already defined types `set` and `List` each take one type parameter
-so it fills in `<T>` (using some unique type parameter name) and then determines the the function itself needs
+so it fills in `<T>` (using some unique type parameter name) and then determines that the function itself needs
 a type parameter `<T>` also.
```
stefan-aws, line 25:
```suggestion
a type parameter `<T>` as well.
```

stefan-aws (approved):

PR #3586, opened 2023-02-21
Update to Boogie 2.16.3, and remove workaround code to deal with concurrency issues in Boogie.

keyboardDrummer:
Remove workaround code to deal with concurrency issues in Boogie.


keyboardDrummer:
@MikaelMayer note that without this block https://github.com/dafny-lang/dafny/pull/3586/files#diff-ba233aa922c57c1e25acf0a501ee2b494373271d715691f2c012c714f795f0f2R261 the tests still pass, but assertion batches will only be reported once the whole implementation has finished verifying. The tests don't seem to catch this issue.

Issue #3605, opened 2023-02-22
Range of bounds variables in set comprehensions not limited to allocated objects


RustanLeino:
### Dafny version

3.12.0.50221

### Code to produce this issue

```dafny
method M()
  ensures false
{
  ghost var s, r;
  s := set c: C | True(c);
  label L:
  r := Same(s);

  var c := new C;
  assert c in s; // because s includes all C's, both allocated and unallocated
  assert c !in r; // because t is known to include only allocated objects
  // yet, s and t are equal
}

class C { }

predicate True(s: C) {
  true
}

ghost method Same(s: set<C>) returns (r: set<C>)
  ensures r == s
{
  r := s;
}
```

### Command to run and resulting output

```code
% dafny test.dfy

Dafny program verifier finished with 2 verified, 0 errors
Compiled assembly into test.dll
```

### What happened?

The problem is that the set comprehension `set c: C | True(c)` is modeled as containing _all_ `C` objects, not just those `C` objects that have been allocated in the program. From that encoding, one can prove `false`, as is done in the postcondition of method `M`.

The problem comes about from the too-frugal encoding in the default `/allocated:3`. While the description of `/allocated:4` seems like it would cure the problem, `allocated:4` crashes all over the place.

### What type of operating system are you experiencing the problem on?

Mac

jtristan:
Fxied by #3609

Issue #3600, opened 2023-02-22
Improve /allocate for Dafny 4.0


robin-aws:
Placeholder based on a conversation with @RustanLeino about improvements he has already implemented and may be deliverable for Dafny 4.0.

robin-aws:
Main PR merged and task to switch the default tracked under #2548

PR #3609, opened 2023-02-22
Fix allocated 4

### LINKED ISSUE #19: quantifiers over types that might be heap types
If C is a heap-related type (such as a class), then there are restrictions on expressions "forall x:C :: ..." to prohibit quantifying over objects x that aren't yet allocated.  However, these restrictions aren't always enforced for type variables T that might later be instantiated with C.  In the example below, the function AllP quantifies over all x of type T, even though T is later instantiated with C.  Then AllP's frame axiom is used (in M3) to prove something that is not provable directly (as seen in M2).  Rustan and I discussed this and tentatively proposed that the rules for quantifying over type variables T should be tightened, unless the programmer annotates T to indicate that T can never be instantiated with a heap-related type.

```
predicate P<T>(x:T)

predicate AllP<T>() { forall x :: P<T>(x) }

class C {}

method M1()

method M2()
{
  assume (forall x :: P<C>(x));
  M1();
  assert (forall x :: P<C>(x)); // FAILS (perhaps correctly)
}

method M3()
{
  assume (forall x :: P<C>(x));
  M1();
  assert AllP<C>();             // SUCCEEDS (perhaps incorrectly)
  assert (forall x :: P<C>(x)); // SUCCEEDS (perhaps incorrectly)
}
```

### LINKED ISSUE #3606: Function in reads clause can undetectedly depend on allocation state
### Dafny version

3.12.0.50221

```dafny
class C { }

function G(): int
  // The following line is equivalent to
  //     reads set c: C, obj: object | obj in M(c) :: obj
  // and should therefore be disallowed.
  reads M
{
  5
}

function M(x: C): set<object> {
  {x}
}

ghost method Test()
  ensures false
{
  var gr := G.reads();
  var c := new C;
  assert c in M(c);
  assert c in gr;
}
```

```code
% dafny test.dfy

Dafny program verifier finished with 3 verified, 0 errors
Compiled assembly into test.dll
```

### What happened?

Dafny allows a `reads` clause to list not only objects and sets of objects (like a `modifies` clause does), but also _functions_ to objects or sets of objects. The primary motivation for this is to allow

``` dafny
reads f.reads
```

where `f` is a parameter to the enclosing function. (Note that `f.reads` is a function from the argument types of `f` to `set<object>`.)

So, a reads clause like the `reads M` in the repro above is desugared into a "big union", which is notated as a generalized set comprehension in Dafny:

``` dafny
reads set c: C, obj: object | obj in M(c) :: obj
```

If this `reads` clause was given this way by the programmer, Dafny would depend that the set comprehension depends on the allocation state--in particular, the set comprehension ranges over all allocated `c: C` values. (Note that the range of `obj: object` is not a problem, since it is constrained by the expression `obj in ...`.)

Dafny should disallow a function `M` in a `reads` clause if that function `M` has in-parameters that may contain references.

Because of this missing check, it is possible to prove `false`, as exhibited by the postcondition of method `Test()` above.

Mac

### LINKED ISSUE #3607: Specifications of lambda expressions not checked for dependency on allocation state
### Dafny version

3.12.0.50221

```dafny
class Cell {
  ghost var data: int
}

method M(cc: Cell)
  ensures false
{
  var f: () ~> set<Cell> :=
    ()
      reads set c: Cell | c in {c}
    => set c: Cell | c.data == 17;

  var s := f();
  var d := P(f);
}

ghost method P(f: () ~> set<Cell>) returns (d: Cell)
  requires f.requires()
  ensures f() == old(f())
  ensures d !in f() && d.data == 17
{
  var s := f();

  // The following creates and returns a Cell
  //  - that's previously not in f(), and
  //  - that now has .data equalling 17
  d := new Cell;
  assert d !in s;
  d.data := 17;

  // Since f is presumed not to depend on the allocation state, we can
  // prove the following:
  var t := f();
  assert s == t;
}
```

```code
% dafny test.dfy

Dafny program verifier finished with 3 verified, 0 errors
Compiled assembly into test.dll
```

### What happened?

All functions, include the function values defined by a lambda expression, are assumed to depend only on allocated objects. Indeed, most functions depend on far less than all allocated objects. In order for a function to depend on (the mutable state of) some object, the function must declare that object in the function's `reads` clause. Therefore, the `reads` clause itself must also (just like the body of the function) not depend on the allocation state. This check is missing from the specification of lambda expressions.

The missing check allows one to prove `false`, as the repro above shows. In it, method `P` gets to assume that its parameter, `f`, does not depend on the allocation state. Thus, the fact that `P` allocates a new object (`d`) should have no bearing on the value returned by `f()` (as is confirmed by `assert s == t;`). The postcondition of `P` lets caller `M` know that there's an object that ought to have been in `f()` but wasn't, and thus `M` can prove `false`.

Mac

RustanLeino:
Fixes #3606 
Fixes #3607 
Under the now-fixed `/allocated:4`, also fixes #19 and #3605.

This PR discovered a necessary change in the standard library: https://github.com/dafny-lang/libraries/pull/99 (which has been merged)

Also, this PR is what's referred to in #3600.

This PR fixes several previous problems with quantifications and comprehensions over reference types. For example, previously, the verifier encoded `set o: object | true` as the set of _all_ objects, both allocated and unallocated, which is wrong in spirit and also allowed for unsound verification.

The PR makes several general fixes and some fixes under just `/allocated:4` (which is not yet the default). We should make `/allocated:4` the default and remove the possibility for users to choose how allocation state is encoded (that is, we should remove the `allocated` flag altogether). However, it seems best to get this PR into Dafny version 3.Last, then make `/allocated:4` the default in Dafny 4.0 (and probably deprecate the `/allocated` flag at that time, too), and later remove the `/allocated` switch.

The general changes in this PR include three breaking changes and one change that may affect verification.
* One breaking change is that some functions used in `reads` clauses are no longer allowed. It was a bug that these were allowed before. This is checked by the resolver.
* Another breaking change is that lambda expressions, including the `requires` and `reads` clauses of lambda expressions, are no longer allowed to depend on the allocation state. The previous omission of this check was a bug. This is checked by the resolver.
* The third breaking change is that the verifier now include allocation antecedents for bound variables whose allocation isn't already implied by the body of the quantifier/comprehension. This is a virtue of (the buggy design of) `/allocated:3` and is fixed in `/allocated:4`. (Historically, this problem was not known at the time `/allocated:3` was designed. The intended design of `/allocated:4` does not have the problem, but the implementation of `/allocated:4` was previously incomplete and buggy.) So, switching from the default `/allocated:3` to `/allocated:4` (which will soon be the new default) changes (fixes!) the encoding of these allocation antecedents, which may be a breaking change for programs that (intentionally or inadvertently) made use of the previous, buggy encoding.
* The possible change to verification is that this PR generalizes one axiom. Logically, adding an axiom does not make any verifying program not verify. However, as with any change in the verifier, in cases where verification was previously unstable, the presence of this new axiom may change verification behavior (e.g., change in verification performance or verification outcome). This axiom is added for both `/allocated:3` and `/allocated:4`.

Unfortunately, there were no tests for `/allocated:4`, so no wonder it didn't work. Fortunately, this PR fixes `/allocated:4`. Unfortunately, the PR adds only a scant amount of specific `/allocated:4` tests. Fortunately, I did run the entire test suite locally with `/allocated:4` and it passes. So, the lack of many `/allocated:4` tests will be remedied as soon as we switch to `/allocated:4` as the default (in a separate PR).

The long-standing issue #19 is fixed with this PR. Actually, the original program reported in #19 has been fixed for a while, but the long discussion of #19 showed many more issues. With `/allocated:4`, all of those have been fixed (and all of them are recorded in tests in `git-issue-19[ab].dfy`).

## More details

The rewrite of functions in `reads` clauses was previously done in more than place. Now, this happens as a desugaring that's part of the resolver's bounds-discovery phase.

Here are descriptions of the changes in the PR, listed more or less in the order that github shows the changes:

* (code improvement) Add `ASTVisitor.VisitTopLevelFrameExpression`.
* (code improvement) Remove `requiredVirtues` parameter to `ComprehensionExpr.GetBest`, since it was always passed in as `0`.
* In `FrameExpression`, capture user-supplied frame expressions into `.OriginalExpression`, add `.DesugaredExpression` to hold any possible frame-expression rewrite (in particular, functions in `reads` clauses get desugared during resolution, and this desugaring is stored in `.DesugaredExpression`), and let `.E` return the "better" of these two fields.
* (code improvement) Move `FrameArrowToObjectSet` method from `ArrowType` to `BoundsDiscoveryVisitor`.
* (code improvement) Specialized the `AstVisitor` `Context` type parameter in the implementation of `BoundsDiscoveryVisitor`.
* (code improvement) Specialized the `AstVisitor` `Context` type parameter in the implementation of `TypeInferenceChecker`. This makes it clearer what parts of the context are important to type-inference checking.
* (fix) Fill in `.Bounds` for various quantifiers/comprehensions/forall-statements that are generated automatically. (These are needed, because `/allocated:4` expects `.Bounds` to be set.)
* Previously, each datatype constructor gave rise to an allocation axiom, roughly `forall x, xs :: IsAlloc(x) && IsAlloc(xs) ==> IsAlloc(Cons(x, xs))`. Now, if no constructor depends on the allocation state, the entire type gets one allocation axiom, rather than having a separate one for each constructor (roughly `forall c: Color :: IsAlloc(c)`). This makes the axiom more useful where there isn't already a case split for each constructor.
* (fix) The previous translation of `allocated(_)` made it sometimes not provable. (This has to do with boxing and has now been fixed.)
* (fix) There was a code path for generating an allocation-consequence axiom for functions. However, it was broken in many places. Apparently, this was not detected, because the axiom is generated only for `/allocated:4` and there were no tests for `/allocated:4`. This PR fixes the code.
* Make some error messages more specific.


jtristan:
Is it now the case that a function (can be used in a reads clause) if and only if ((it is of type ->) or (it is of type -->)) ?

jtristan:
Is it now the case that a reads function is (of type ->) or (or type -->)?

jtristan (commented):
Summary. A long-standing issue with Dafny's logic is that universal quantification over objects was defined so that it bounded both allocated and unallocated objects, allowing to state properties of objects before they even exist. This problem had been addressed with a more careful translation and the fix was available through the non-default flag /allocated:4. We plan on making /allocated:4 the default (and only) interpretation of universal quantification over objects in Dafny 4 and this PR is preparing for that change. To that end, this PR addresses a few issues with the current implementation of /allocated:4. In particular, it restricts which functions can be used or defined as reads clauses. This fixes must also be applied to implicit forms of quantification, such as set comprehension. The bulk of the modifications consist in tracking the set of bound objects explicitly. The PR also include a change to the translation of datatypes so that the generated allocation axioms are not defined recursively for pure datatypes.

Source/DafnyCore/Verifier/Translator.DataTypes.cs
```diff
@@ -16,7 +16,11 @@ partial class Translator {
       Contract.Requires(dt != null);
       Contract.Requires(sink != null && predef != null);
 
-      var constructorFunctions = dt.Ctors.ToDictionary(ctor => ctor, ctor => AddDataTypeConstructor(dt, ctor));
+      var mayInvolveReferences = UserDefinedType.FromTopLevelDecl(dt.tok, dt).MayInvolveReferences;
+      var constructorFunctions = dt.Ctors.ToDictionary(ctor => ctor, ctor => AddDataTypeConstructor(dt, ctor, mayInvolveReferences));
+      if (!mayInvolveReferences) {
+        AddCommonIsAllocConstructorAxiom(dt);
```
jtristan, line 22: Why does this work? Consider the value Cons(3,Cons(4,Nil)). Does it work because if Cons(3,Cons(4,Nil)) is allocated, then we can easily deduct that Cons(4,Nil) is as well? Or is it because it is always good enough to know that Cons(3,Cons(4,Nil)) is allocated?

jtristan (approved):

PR #3613, opened 2023-02-23
Make tests easier in preparation for Z3 4.12.1

atomb:
Make several of the test easier to prove by adding intermediate assertions, making things opaque, or adding `{:vcs_split_on_every_assert}`. This will be necessary for the upgrade to Z3 4.12.1, and is a portion of #3400, but is something that we should be able to merge independently and include both before and after the Z3 upgrade.


Test/dafny2/SnapshotableTrees.dfy.expect
```diff
@@ -1,25 +1,7 @@
 SnapshotableTrees.dfy(71,25): Error: A precondition for this call might not hold.
```
davidcok, line 1: Is this really the same test?

atomb: My understanding is that, previously, Z3 wasn't able to prove certain things, but that it was still a useful test. With the changes to the structure, Z3 can prove more. I could be wrong about that, but it seems plausible.

davidcok (approved):

alex-chew (approved):

PR #3617, opened 2023-02-23
Dispose Z3 processes after starting a language server in an XUnit test

keyboardDrummer:
This prevents dangling Z3 processes from accumulating when running the XUnit tests


Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
```diff
@@ -152,6 +152,9 @@ ensures true
 #pragma warning disable VSTHRD110 // Observe result of async calls
       Server.Initialize(CancellationToken);
 #pragma warning restore VSTHRD110 // Observe result of async calls
+
+      Disposable.Add(Server);
```
atomb, line 156: Nice! I started down this path, but it looks like I didn't go far enough. I didn't know about `Server.Services`.

atomb (approved):

atomb:
Oh, it looks like there's a formatting change needed for CI to pass.

atomb (approved):

Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
keyboardDrummer: I did some debugging on the library source to find this solution, starting from the knowledge that Boogie's ExecutionEngine has a CheckerPool that needs to be disposed, and then walking up the ownership chain to figure out which objects need to be disposed as well.

PR #3619, opened 2023-02-23
Let Java properly include runtime sources

### LINKED ISSUE #3611: `dafny build` with `extern` declarations
### Dafny version

3.10.0.41215

```dafny
module Test {
  function {:extern} inc(x:int) : (r:int)

  method aTest() {
    expect inc(1) == 1;
  }
}
```

```code
dafny build --function-syntax 4 test.dfy
```

This resulted in the following:

```
Errors compiling program into test
(5003,37): error CS0117: '__default' does not contain a definition for 'inc'
```
```


### What happened?

I don't necessarily believe the above was unexpected --- but I am stuck.  The question is: _how do I compile Dafny containing `{:extern}` functions to Java?_  

Instead of using `dafny build` I logically tried `dafny translate` next.  Unfortunately, this does not generate the `DafnyRuntime.jar` even with `--include-runtime` provided (perhaps that is the real bug here).  This means I cannot build just from the Dafny source, but would have to check the runtime jar into my repository.

My requirements for building the project are:

1. Dafny source files in one directory (e.g. `src/dafny`)
2. My Java source files in another directory (e.g. `src/java`) which provides implementation for `{:extern}` functions.
3. Generated Java files from Dafny placed into a build directory (e.g. `bin/java`).
4. At which point, I should be able to compile everything in `src/java` and `bin/java` together.

I should also note that `dafny test` has similar problems to `dafny build`.  However, I cannot control how `dafny test` is trying to build my project.  It does not even support the `--output` option from `dafny build`.

Anyway, any thoughts on this would be appreciated.

Linux

### PR DESCRIPTION
**Pull request by keyboardDrummer**
### Changes

1. Let `dafny translate java --include-runtime` produce emit the Dafny runtime as .java files, similar to the other back-ends. Previously ` --include-runtime` has no effect when using `dafny translate java`.
1. Let `dafny /compileTarget:java /compile:2 /spillTargetCode:3 /useRuntimeLib` emit sources translated from your Dafny files, a `UserCode.jar` that doesn't include the Dafny runtime, and a `DafnyRuntime.jar`. Previously it would fail on invoking `javac`.
1. Rename RuntimeLib to IncludeRuntimeSources, inverting its meaning

### Testing

- Update the test ManualCompile.dfy so that it no longer uses `DafnyRuntime.jar` from the binary directory, but instead only relies on source files generated by the Dafny binary.
- Add `JavaUseRuntimeLib.dfy` to verify change number 2 works.

Fixes #3611

### CODE REVIEW

**Comment by davidcok** [comment]
Are you now including the Java _source_? Or properly including the Java Runtime Jar? I hope the latter and would not be in favor of the former -- in java world I've never wanted to work with or have to compile the source for a library.

**Comment by keyboardDrummer** [comment]
> Are you now including the Java _source_? Or properly including the Java Runtime Jar? I hope the latter and would not be in favor of the former -- in java world I've never wanted to work with or have to compile the source for a library.

As with the other runtimes, if `IncludeRuntimeSource` is set to true, the runtime is emitted _as source_. If you leave that false, no runtime sources are emitted and instead you have to reference a DafnyRuntime file, which you can get from the various package manager, when building your sources using an external build system.

**Comment by davidcok** [comment]
I just tested this myself for java. build and run work fine. For translate
- to match our documentation, --include-runtime should cause a copy of DafnyRuntime.jar to be put with the translated sources (in A-java or whatever)
- to improve on current state, always include the DafnyRuntime.jar -- but I don't know if that matches expectations for non-compiled languages.

If we get rid of --include-runtime in favor of --include-source (or some such), I'd still say always include compiled libraries but for non-compiled languages include runtime source just when --include-source is specified.

But I don't think any change from intended past behavior should be made this close to 4.0 release.

**Comment by keyboardDrummer** [comment]
This PR is just maintaining the status quo, which is that the Dafny runtime is included as _source_ when compiling with `dafny /spillTargetCode:3` or `dafny translate --include-runtime`.

Previously for Java no runtime was included when using `dafny /spillTargetCode:3` or `dafny translate --include-runtime`, and now the runtime source is included, as it already is for the other back-ends.

I agree with you that it _might_ be better to include a more compiled runtime instead of the runtime's source, but then we should do that consistently across back-ends, not just for Java. I don't intend for this PR to have such a large scope. Its intention is purely to fix a bug, not to change the behavior of `--include-runtime`.

> to match our documentation

Could you point me to documentation that says a DafnyRuntime.jar is included when using `--include-runtime` ? Then I'll update that.

**Comment by davidcok** [comment]
If all you are doing is including source for Java in the translate, and not by default, I agree the change is small -- except that we are also making the breaking change of changing the name of the option.
However, as a longtime Java user, it is not a bit of functionality I would ever want. What we describe for Java users (against which we should measure bugs) is that the .jar is copied in with the generated source, to make it easy to compile the generated sources.

So let's think through the use cases and what the functionality of --include-runtime should be.

For dafny run/build -- to make a runnable executable, the library has to be available -- either built-in to the executable or linked in as an additional artifact. And if it is a library that is being created, then there is a real question as to whether the runtime is included in the constructed library or not (one might be making multiple libraries and not want the runtime in all or any of them).
So for run/build I think --include-runtime is a legitimate choice for which there are use cases either way, but the default should be to include it in the built executable. And that is the current behavior of --include-runtime.

Note that in the new CLI one can keep the build artifacts using --spill-translation, though it appears that we have not exposed that option -- so we should expose it, not make a new option.

So what about translate. The use case here is either (1) to be able to inspect the translation of Dafny to Target source and (2) to be able to build the end product oneself using ones own build system. In either case I would expect that the preferred artifact for the runtime is the binary rather than the source. But granting that someone wants to build the runtime from source, I would advocate an --include-runtime option with three values: none, binary, source (with the default being none, as now).

In summary:
- I'm still opposed to changing the name of the option, even just for translate. Make if --include-runtime:source if need be, but still keep/implement the behavior of --include-runtime:binary and --include-runtime:none
- Be sure to keep the current behavior for build/run -- include-runtime (binary) by default with the option of none.
- And of course, for non-compiled languages, the binary option in all cases is moot.

**Comment by keyboardDrummer** [comment]
> except that we are also making the breaking change of changing the name of the option.

What do you mean? No CLI options are renamed in this PR. Currently there is only `--include-runtime` for the new CLI and `/useRuntimeLib` for the old one, and after this PR that's still the case.

> So what about translate. The use case here is either (1) to be able to inspect the translation of Dafny to Target source and (2) to be able to build the end product oneself using ones own build system. In either case I would expect that the preferred artifact for the runtime is the binary rather than the source. But granting that someone wants to build the runtime from source, I would advocate an --include-runtime option with three values: none, binary, source (with the default being none, as now).

That seems good, but I won't add that to the scope of this PR. The existing meaning of `--include-runtime`, which is only available for `dafny translate`, is to have none or source.

File: Source/XUnitExtensions/Lit/ILitCommand.cs [hunk 1]
```diff
@@ -2,12 +2,31 @@
 using System.Collections.Generic;
```
davidcok, line 2: What does this change to Lit have to do with the PR?

Source/DafnyCore/Compilers/CSharp/CsharpBackend.cs
```diff
@@ -48,11 +48,6 @@ public class CsharpBackend : ExecutableBackend {
     compilation = compilation.WithOptions(compilation.Options.WithOutputKind(callToMain != null ? OutputKind.ConsoleApplication : OutputKind.DynamicallyLinkedLibrary));
 
     var tempCompilationResult = new CSharpCompilationResult();
-    var libPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
```
davidcok, line 51: Why are we no longer emitting the runtime dll?

davidcok (changes requested):

Source/DafnyCore/Compilers/CSharp/CsharpBackend.cs
keyboardDrummer: It's not being done on master either. Having these lines caused a crash.

Source/XUnitExtensions/Lit/ILitCommand.cs
keyboardDrummer: The PR uses path expansion "%S/**/*.java" in the Lit commands that wasn't well supported by our Lit emulation.

davidcok:
I still think that
- the intent of the --include-runtime option was to include the appropriate runtime for the backend (binary or source) along with the translated sources, and not just source necessarily; so the internal name change is inappropriate
- at some point in the past, this worked for Java - the DafnyRuntime.jar was copied in along with the generated sources, so it was easy to construct a command-line that built an executable
- as  Java user I don't want to be bothered with sources for libraries in Jar form, except when needed for debugging

But I don't have time or energy to debate this anymore.

davidcok (approved):

robin-aws (changes requested):
Want to have a look myself before this goes in

robin-aws:
I mostly agree with the changes here. The important meaning of `--include-runtime` is to include the Dafny runtime directly in the component being compiled from Dafny, so it has no extra dependencies. What we do to make that happen is implementation detail, so I don't think we need to be more specific with the public option.

For C# that has always meant literally prepending the source code to the single `.cs` file it outputs, or internally compiles and runs. For Java the equivalent behavior is outputting all the individual `.java` runtime files in the same location/scope as the `.java` files generated from Dafny source. For `dafny build` or `dafny build`, all of it should be compiled together into a single jar.

`ManualCompile.dfy` shows how Java not doing this was inconsistent with the rest of the backends, because the Java case was the only place where you had to include a reference to the runtime library.

@davidcok you're correct that this is a bug introduced in the new CLI (by having `--include-runtime` do nothing for Java). It also looks like the old CLI is confused because it does both: `dafny /compileTarget:java Hello.dfy` splits out the class files for the runtime right beside `Hello.class`, but also leaves a copy of `DafnyRuntime.jar` there too. So I can understand  why you think we should be outputting `DafnyRuntime.jar`. But with my interpretation of the option above, that is a bug too (just a mostly harmless one).

If you specify `--include-runtime:false`, that should mean you are expected to provide `DafnyRuntime.jar` from somewhere else. Given we distribute it in Dafny releases, and now have it available from Maven too, I think spitting out a copy of it in the build output is just weird and non-standard and won't help any realistic scenarios.

I don't fully agree with the internal option rename as referring to "source" is too specific about implementation strategy, so I'll just make a suggestion there.

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -313,7 +313,7 @@ public enum IncludesModes {
 
     public IncludesModes PrintIncludesMode = IncludesModes.None;
     public int OptimizeResolution = 2;
-    public bool UseRuntimeLib = false;
+    public bool IncludeRuntimeSource = true;
```
robin-aws, line 316: I do think "include runtime" is better and less ambiguous than "use runtime lib" (since the latter is talking about what the user has to do and not what our code is doing, but I don't think adding source helps. The important meaning is to embed the runtime somehow, and It's conceivable some future backend might want to achieve this by statically linking a pre-built binary or something like that.

I would make this `IncludeRuntime` with a comment about what that actually means.

robin-aws:
Oh I would add a release note too. It's fair to call this a bug fix, but it will definitely break anyone who switched to the new CLI and assumed the current behavior was intentional, so let's at least call it out loudly.

davidcok (changes requested):
I'll approve this once Robin's comments are addressed - which I think are just the name of the option field and the release note.

keyboardDrummer:
> it will definitely break anyone who switched to the new CLI

How would it break them? I'm not sure what to put in a release note except `--include-runtime` now works for Java.

Source/DafnyCore/Compilers/Java/JavaBackend.cs
```diff
@@ -81,19 +78,9 @@ public class JavaBackend : ExecutableBackend {
       return false;
     }
 
-    if (!DafnyOptions.O.UseRuntimeLib) {
```
davidcok, line 84: It looks like you are changing the Java build behavior to no longer include the runtime files in the built executable jar, but rather, requiring DafnyRuntime.jar on the CLASSPATH. If so (1) that ought to be a user choice (based on --include-runtime, and the default should be to include the runtime in the executable jar) and (2) the trend in Java is to package up everything in an executable Jar, or even in a natively executable exe by bundling in the JVM, so I think unbundling is the wrong direction. And this is definitely a change in behavior, as we document that one can just run the executable jar (without having to adjust the CLASSPATH).

If instead, you have included the sources and the user is recompiling the sources to include in the executable jar. then we have just added to the user's compilation burden (a backward step).

[ I though t we were just fixing the omission of not emitting runtime sources in 'dafny translate']

Test/dafny0/JavaUseRuntimeLib.dfy
```diff
@@ -0,0 +1,9 @@
+// RUN: %dafny "%s" /useRuntimeLib /out:%S/Output/DafnyConsole.jar /compileTarget:java > "%t"
```
davidcok, line 1: What does this look like in the new CLI?

davidcok (changes requested):

Test/comp/compile1verbose/CompileAndThenRun.dfy.expect
```diff
@@ -16,6 +16,39 @@ hello, Dafny
 
 Dafny program verifier finished with 0 verified, 0 errors
 Wrote textual form of target program to CompileAndThenRun-java/CompileAndThenRun.java
+Additional target code written to CompileAndThenRun-java/dafny/Array.java
```
davidcok, line 19: I know this output only happens in --verbose, but it still seems more noise than needed, unfortunately. As well as being fragile when the runtime sources change. I'd prefer the previous solution of just using the runtime .class files and not recompiling them.

Source/DafnyCore/Compilers/Java/JavaBackend.cs
keyboardDrummer: > It looks like you are changing the Java build behavior to no longer include the runtime files in the built executable jar

They are included through the changes in `Source/DafnyCore/Compilers/Java/Compiler-java.cs`. This change makes sure they're not included twice.

Test/dafny0/JavaUseRuntimeLib.dfy
keyboardDrummer: You can't do this with the new CLI, since there's no `--exclude-runtime` option in `dafny build`, and no `--compile` option in `dafny translate`

davidcok: One of the use cases for `dafny build` is to compile a library (rather than an executable). In that case, the user needs the choice, in java at least, whether to bundle the runtime into the new library or not. I'm not sure how each of these cases is accomplished with the new CLI and the changes you are implementing here.

davidcok: The intuitive meaning, to me, of --include-runtime for `dafny build` is whether to build the resulting library or executable with or without the runtime bundled in. Is that what your implementation accomplishes?

keyboardDrummer: For `dafny build` there shouldn't be an `--include-runtime`, because it includes the runtime by default There could be an `--exclude-runtime`, but I'm more inclined towards adding a `--compile` to `dafny translate`.

davidcok: We don't need --exclude-runtime, just --include-runtime:false for `dafny build`. That would solve the use case issue I have in mind. That build, internally, would have to be done with DafnyRuntime.jar on the classpath (and without the sources).

davidcok: Remy, if you concur with this use of --include-runtime for `dafny build`, I'm happy to put off implementing that behavior to a future PR. For Java at least, If the runtime library is included in multiple jars, Java just uses the class files from the first one -- so the only bad effects are library bloat and possible conflicts between versions.

keyboardDrummer: How do you feel about `--compile` for `dafny translate`, that way all your library use-cases are handled by `dafny translate`, instead of by both `dafny build` and `dafny translate` ?

Test/comp/compile1verbose/CompileAndThenRun.dfy.expect
keyboardDrummer: > I know this output only happens in --verbose, but it still seems more noise than needed, unfortunately.

I think `/compileVerbose:1` does what it's intended to do here. We can remove that option in the test to make the test more robust, regardless of this PR, but we can also do that when it breaks.

> I'd prefer the previous solution of just using the runtime .class files and not recompiling them.

Why? Do you think there is a noticeable difference?

Test/dafny0/JavaUseRuntimeLib.dfy
davidcok: I feel that adding --compile to translate is just turning it into build; just like build with not deleting sources is close to translate. And if the two are to be merged, I'd prefer the name build. I understood the distinction to be:
translate -- produce source code without building it, for inspection or because one needs to incorporate the source into a build system in a way that is more complex than `dafny build`. If we want we could also write out a csproj/gradle/Makefile/antfile or whatever to help the user build it.
build -- produce a library or executable, incorporating external code if desired, but ready to be used; that is, the user does not have to figure out how to build the Dafny translation artifacts into the library/executable
run -- a way to quickly translate-build-run programs with simple build systems.

And I still like that distinction.
- I like the idea of auto-detecting the platform
- Also like producing platform specific binaries as an option (i.e.,. as an alternate but not in place of an executable jar for java)
- On your third bullet: good with that as well, given the user has an option to pick the underlying mechanism
- And I'm all for a good library as well.

As for your comments on library use case. There are simple use cases as well. Just compile and package up some Dafny functionality into a library one uses locally (not distributed). And not nearly all Java uses require Gradle, .Net or maven. I would not want simple uses of Dafny to require users to be setup with build environments.

Source/DafnyCore/Compilers/Java/JavaBackend.cs
keyboardDrummer: I see, thanks.

davidcok (dismissed):
It is sort of messy that for compiled languages we still have to package all the runtime source into the release package. But we're open source anyway.

I'm going to need to experiment with this PR after merge and build to really understand what is changed and how to explain it in the docs, but I'll go with this.

robin-aws:
> > it will definitely break anyone who switched to the new CLI
> 
> How would it break them? I'm not sure what to put in a release note except `--include-runtime` now works for Java.

I was thinking that anyone using `--include-runtime` previously would find they had to include `DafnyRuntime.jar` on their CLASSPATH to build successfully, and then when we fixed the option to actually do something it would cause problems, but I now I recall that Java generally doesn't have a problem with multiple copies of the same code on the CLASSPATH. :P So it won't actually break anyone immediately, until many months later when the copies get out of sync and there's a subtle bug related to which copy gets loaded!

It will still be worth a release note to help users spot that they can avoid introducing a second copy by accident.

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
```diff
@@ -79,7 +79,11 @@ public class CsharpCompiler : SinglePassCompiler {
         CsharpSynthesizer.EmitImports(wr);
       }
       EmitDafnySourceAttribute(program, wr);
-      ReadRuntimeSystem(program, "DafnyRuntime.cs", wr);
+
+      if (DafnyOptions.O.IncludeRuntime) {
```
robin-aws, line 83: FYI this doesn't change anything, because `ReadRuntimeSystem` checks for the option and returns early itself.

keyboardDrummer: Not anymore :D

keyboardDrummer:
> I was thinking that anyone using --include-runtime previously would find they had to include DafnyRuntime.jar on their CLASSPATH to build successfully, and then when we fixed the option to actually do something it would cause problems

This seems quite far fetched, why would anyone be using `--include-runtime` if it didn't do anything?

robin-aws (approved):

PR #3625, opened 2023-02-23
Add `--verbosity` and `--solver-opt` flags

atomb:
Addresses most of #3468. Additional help flags (for solver options, attributes), control of deprecation, extraction of counterexamples, and something like `/noCheating:0` could still potentially be valuable.

Bumps the Boogie dependency along the way.

### DIFF
--- a/Source/DafnyCore/Options/CommonOptionBag.cs
+++ b/Source/DafnyCore/Options/CommonOptionBag.cs
@@ -11,6 +11,9 @@ public class CommonOptionBag {
   public static readonly Option<int> VerificationErrorLimit =
     new("--error-limit", () => 5, "Set the maximum number of verification errors to report (0 for unlimited).");
 
+  public static readonly Option<Boogie.CoreOptions.VerbosityLevel> Verbosity =
+    new("--verbosity", () => CoreOptions.VerbosityLevel.Normal, "Set the level of output produced. Ranges from nothing (Silent) to detailed (Trace). With 'Normal', produce warnings, errors, and some extra information. With 'Quiet', produce only errors and warnings.");
+
   public static readonly Option<bool> ManualLemmaInduction =
     new("--manual-lemma-induction", "Turn off automatic induction for lemmas.");
 
@@ -28,10 +31,6 @@ true - In the compiled target code, transform any non-extern
     non-ghost parameter into just that parameter. For example, the type
         datatype Record = Record(x: int)
     is transformed into just 'int' in the target code.".TrimStart());
-
-  public static readonly Option<bool> Verbose = new("--verbose",
-    "Print additional information such as which files are emitted where.") {
-  };
 
   public static readonly Option<bool> DisableNonLinearArithmetic = new("--disable-nonlinear-arithmetic",
     @"
@@ -186,6 +185,10 @@ Functionality is still being expanded. Currently only checks contracts on every 
       (options, value) => { options.DisallowConstructorCaseWithoutParentheses = value; });
     DafnyOptions.RegisterLegacyBinding(WarningAsErrors, (options, value) => { options.WarningsAsErrors = value; });
     DafnyOptions.RegisterLegacyBinding(VerificationErrorLimit, (options, value) => { options.ErrorLimit = value; });
+    DafnyOptions.RegisterLegacyBinding(Verbosity, (options, value) => {
+      options.Verbosity = value;
+      options.CompileVerbose = value >= CoreOptions.VerbosityLevel.Trace;
+    });
     DafnyOptions.RegisterLegacyBinding(VerifyIncludedFiles,
       (options, value) => { options.VerifyAllModules = value; });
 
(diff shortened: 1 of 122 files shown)


Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -60,9 +59,10 @@ public class CommonOptionBag {
     ArgumentHelpName = "configuration"
   };
 
-  public static readonly Option<uint> SolverResourceLimit = new("--resource-limit", @"Specify the maximum resource limit (rlimit) value to pass to Z3. Multiplied by 1000 before sending to Z3.");
+  public static readonly Option<uint> SolverResourceLimit = new("--resource-limit", @"Specify the maximum resource limit (rlimit) value to pass to Z3. A resource limit is a deterministic alternative to a time limit. The output produced by `--log-format csv` includes the resource use of each proof effort, which you can use to determine an appropriate limit for your program. Multiplied by 1000 before sending to Z3.");
   public static readonly Option<string> SolverPlugin = new("--solver-plugin", @"Specify a plugin to use to solve verification conditions (instead of an external Z3 process).");
   public static readonly Option<string> SolverLog = new("--solver-log", @"Specify a file to use to log the SMT-Lib text sent to the solver.");
+  public static readonly Option<IList<string>> SolverOpt = new("--solver-opt", @"Specify an option to control how the solver works.");
```
davidcok, line 65: I don't think there is a need to be cryptic: call it --solver-options

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -8,8 +8,11 @@ namespace Microsoft.Dafny;
 
 public class CommonOptionBag {
 
-  public static readonly Option<int> ErrorLimit =
-    new("--error-limit", () => 5, "Set the maximum number of errors to report (0 for unlimited).");
+  public static readonly Option<int> VerificationErrorLimit =
+    new("--error-limit", () => 5, "Set the maximum number of verification errors to report (0 for unlimited).");
+
+  public static readonly Option<Boogie.CoreOptions.VerbosityLevel> Verbosity =
+    new("--verbosity", () => CoreOptions.VerbosityLevel.Normal, "Set the level of output produced. Ranges from nothing (Silent) to detailed (Trace). With 'Normal', produce warnings, errors, and some extra information. With 'Quiet', produce only errors and warnings.");
```
davidcok, line 15: The definition of VerbosityLevel is somehow not in these diffs.
Between Normal and Trace there should be a Verbose level which matches whatever the previous --verbose option did. To me Trace connotes even more output - detailed events along the way.

davidcok (changes requested):

Source/DafnyCore/Options/CommonOptionBag.cs
atomb: The previous `--verbose` option affected only compilation, and was more like verification tracing, so it seemed like it made sense to put it in the same bucket.

davidcok: I still would like a level in between. The step from Normal to Trace seems huge. I don't expect anyone to use TRACE as a matter of course -- just when needing to understand what is going wrong  with something. Whereas verbose is just a bit more information than normal, but not an overwhelming amount.

For example, I need a way to turn on explanations of error messages on the command-line (like we are putting into hover text in the IDE). Not on by default. But then if accompanied by lots of other 'tracing' information would be too much.

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -8,8 +8,11 @@ namespace Microsoft.Dafny;
 
 public class CommonOptionBag {
 
-  public static readonly Option<int> ErrorLimit =
-    new("--error-limit", () => 5, "Set the maximum number of errors to report (0 for unlimited).");
+  public static readonly Option<int> VerificationErrorLimit =
+    new("--error-limit", () => 5, "Set the maximum number of verification errors to report (0 for unlimited).");
+
+  public static readonly Option<Boogie.CoreOptions.VerbosityLevel> Verbosity =
```
keyboardDrummer, line 14: Let's define our own VerbosityLevel type. I don't think this is the type of thing we want to import from Boogie.

atomb:
I'm now thinking that `--verbosity` should be global, not limited to verification, and therefore shouldn't re-use Boogie's verbosity levels. That leads me to lean in the direction of not including this in 4.0, since doing a good job of broader support for adjustable verbosity is a somewhat involved project. For the moment, `dafny verify --boogie -trace` provides the immediate functionality I wanted out of verbosity, and `--boogie -proverOpt:...` provides the `--prover-option` functionality. I'm still leaning toward adding Dafny-level options for those, but it's less urgent now, in my mind.

I'm going to create a new PR that _just_ updates Boogie.

davidcok:
verbosity (or a better name) should definitely be an option for every dafny command -- even if it doe snot control anything now, it may well in the future. I have some ideas in mind already.

atomb:
> verbosity (or a better name) should definitely be an option for every dafny command -- even if it doe snot control anything now, it may well in the future. I have some ideas in mind already.

Sounds good. Let's chat about it and do this properly after 4.0.

atomb:
Closing in favor of #3943.

PR #3626, opened 2023-02-24
chore: Enable all CI for main-* as well, disable push triggers

robin-aws:
Our model is to have branch protection for all important branches, and therefore there isn't any need to run CI on pushes (especially since we have a nightly build to test these branches directly anyway). This should save a fair bit of CI capacity.

Also adds concurrency control for unit tests too (since they are also getting expensive).

This does NOT yet enable the nightly build for `main-*` unfortunately, as scheduled builds only run on the default branch, so that will take some more thinking (plus ensuring that different branches produce distinct prerelease version identifiers).

### DIFF
--- a/.github/workflows/doc-tests.yml
+++ b/.github/workflows/doc-tests.yml
@@ -9,8 +9,7 @@ name: Test documentation
 on:
   workflow_dispatch:
   pull_request:
-    branches: [ master ]
-  push:
+    branches: [ master, main-* ]
 
 jobs:
   check-deep-tests:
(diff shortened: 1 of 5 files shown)


atomb (approved):
Looks good!

fabiomadge:
The push trigger actually had a purpose. We have an if clause that should prevent it from actually running something unless you are on a fork and have some global variable set, i.e. you know what you're doing. I agree that the pushes to master need no extra tests, but they were there before, so we preserved them.

PR #3628, opened 2023-02-24
Test improvements discovered as part of Z3 upgrade

atomb:
Merging these for both 3.x and 4.x will simplify #3400.

### DIFF
--- a/Test/cli/errorLimit.dfy
+++ b/Test/cli/errorLimit.dfy
@@ -1,4 +1,4 @@
-// RUN: %exits-with 4 %baredafny verify %args --error-limit=2 "%s" > "%t"
+// RUN: %exits-with 4 %baredafny verify %args --error-limit=0 "%s" > "%t"
 // RUN: %diff "%s.expect" "%t"
 method m(x: int) {
   if x == 1 {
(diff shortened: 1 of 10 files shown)


Test/cli/errorLimit.dfy
```diff
@@ -1,4 +1,4 @@
-// RUN: %exits-with 4 %baredafny verify %args --error-limit=2 "%s" > "%t"
+// RUN: %exits-with 4 %baredafny verify %args --error-limit=0 "%s" > "%t"
```
davidcok, line 1: It  looks like this one was intended to test the error limit. Is it too non-deterministic ?

davidcok (changes requested):

Test/cli/errorLimit.dfy
atomb: Yeah, I originally created that one to test the `--error-limit` flag, and then found that limiting it to a subset of the errors was non-deterministic, so I'd use it in the mode where it raised the limit, instead. At first, I only had that fix on the Z3 4.12.1 test update branch, but it really belongs everywhere.

alex-chew (approved):

PR #3629, opened 2023-02-24
Clean up output of auditor

atomb:
* Skip auto-generated members
* Properly indent members in IETF-style output
* Use more human-friendly names in IETF-style output


Source/DafnyCore/Auditor/AuditReport.cs
```diff
@@ -115,13 +115,21 @@ public class AuditReport {
     StringBuilder text = new StringBuilder();
 
     foreach (var module in modulesWithEntries) {
-      text.AppendLine($"# Module `{module.Name}`");
+      if (module.Name.Equals("_module")) {
```
keyboardDrummer, line 118: A Name comparison seems like a frail method to determine whether it's the default module.

Source/DafnyCore/AST/TopLevelDeclarations.cs
```diff
@@ -33,13 +33,15 @@ public abstract class Declaration : RangeNode, IAttributeBearingDeclaration, IDe
   private bool scopeIsInherited = false;
 
   public override IEnumerable<AssumptionDescription> Assumptions() {
-    if (Attributes.Contains(Attributes, "axiom")) {
+    if (Attributes.Contains(Attributes, "axiom") &&
```
keyboardDrummer, line 36: Magic strings like `axiom` and `auto_generated` should be avoided in favor of referring to a single constant where the string is defined.

Source/DafnyCore/Auditor/AuditReport.cs
davidcok: So at least use a const value somewhere, or an IsRootModule predicate or something.

atomb: Yeah, I strongly agree. I had initially planned to make this PR as minimal as possible, and this sort of string comparison is what the rest of the code does, but it wouldn't be hard to introduce a reusable predicate as part of this PR.

Source/DafnyCore/AST/TopLevelDeclarations.cs
atomb: Ditto my previous comment.

Source/DafnyCore/Auditor/AuditReport.cs
atomb: Oh, and it turns out that this and the related class property already existed, and I hadn't found them before because they do better than using string comparison.

alex-chew (approved):

PR #3630, opened 2023-02-24
fix: Java --unicode-char mode coercion improvements 

### LINKED ISSUE #3413: Missing coercions in edge cases for characters in Java with `--unicode-char`
Follow-up from https://github.com/dafny-lang/dafny/pull/2818. `Test/unicodechars/comp/Arrays.dfy` doesn't work for Java because of several cases of missing coercions between `int` and `CodePoint`. I have the fixes but they need clean up.

robin-aws:
Fixes #3413.

Addresses the issues uncovered during #2818 by adding a `--unicode-char` mode version of `Test/comp/Arrays.dfy`, which all stem from incomplete handling of manual boxing/unboxing of the `CodePoint` type at various Java code generation points. Note this is still incomplete as there must be other spots that do not handle coercion correctly in general, but these changes at least cover `Arrays.dfy`.

It turned out that handling arrow coercion as in the Go compiler was not necessary for Java, because the initial casting of a function reference as a lambda is where the necessary boxing/unboxing needs to happen instead anyway. That is, a reference to a Dafny `function f(x: char): char` has to end up as a Java `Function<CodePoint, CodePoint>` and therefore be eta expanded from the start.

Also removed the fail-fast behavior from `%testDafnyForEachCompiler` as I found it more useful for debugging that way.

Also implemented `ConcreteSyntaxTree.Comma` as part of an initial implementation of Java arrow conversion, which ended up not directly used but seems useful enough to keep (and I refactored one spot to use it as an example).

Edit: I've also added a few more similar cases exposed as part of enabling `/unicodeChar:1` by default for Dafny 4.0: https://github.com/dafny-lang/dafny/pull/3635


Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -3646,22 +3665,31 @@ protected class ClassWriter : IClassWriter {
...
+        if (IsObjectType(from) && IsCharBasedType(to)) {
+          wr.Write("((dafny.CodePoint)(");
           var w = wr.Fork();
           wr.Write(")).value()");
           return w;
         }
 
-        if (from is { IsCharType: true } && IsObjectType(to)) {
-          wr.Write($"dafny.CodePoint.valueOf(");
+        if (IsCharBasedType(from) && IsObjectType(to)) {
```
RustanLeino, line 3691: `IsObjectType` is not part of this PR, but I find its name misleading. It seems to return whether or not the given type is _compiled into_ Java's type `java.lang.Object`. Could the method be renamed to something like `IsObjectTargetType` or something like that?

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -2899,6 +2899,15 @@ void TrCasePatternOpt<VT>(CasePattern<VT> pat, Expression rhs, string rhs_string
       return TypeInitializationValue(simplifiedType, wr, tok, usePlaceboValue, constructTypeParameterDefaultsFromTypeDescriptors);
     }
 
+    protected string DefaultValueCoercedIfNecessary(Type type, ConcreteSyntaxTree wr, IToken tok,
+      bool constructTypeParameterDefaultsFromTypeDescriptors = false) {
+
+      var tempWr = new ConcreteSyntaxTree();
```
RustanLeino, line 2905: `tempWr` would seem better named `resultWr`.

Source/DafnyCore/ConcreteSyntax/ConcreteSyntaxTree.cs
```diff
@@ -73,6 +73,31 @@ public T Append<T>(T node)
...
+      var sep = "";
+      var index = 0;
+      foreach (var element in elements) {
+        Write(sep);
+        a(element, index);
+        sep = comma;
+        index++;
+      }
+
+      return this;
```
RustanLeino, line 98: These look useful. In fact, so useful that they might be used outside the compilers. I suggest moving the `Action`-based `Comma` methods to class `DafnyCore/Generic/Util.cs`, where there are already several other `Comma` methods. It would then make sense to make these new routines return a `string`, which the `ConcreteSyntaxTree` clients can then `Write(...)` themselves.

RustanLeino (changes requested):

RustanLeino:
This PR fixes https://github.com/dafny-lang/dafny/issues/3413, right?

Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -3646,22 +3665,31 @@ protected class ClassWriter : IClassWriter {
       return type == NativeObjectType || type.IsTypeParameter;
     }
 
+    private bool IsCharBasedType(Type type) {
+      return type switch {
+        { IsCharType: true } => true,
+        UserDefinedType { ResolvedClass: SubsetTypeDecl std } =>
+          IsCharBasedType(std.RhsWithArgument(type.TypeArgs)),
+        _ => false
+      };
```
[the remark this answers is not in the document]
robin-aws: I found I needed this for `newtype` declarations based on `char` as well, but now that I look at it I don't understand why this implementation, which looks for `SubsetTypeDecl`, works. :) I'll investigate.

Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -3646,22 +3665,31 @@ protected class ClassWriter : IClassWriter {
...
     protected override ConcreteSyntaxTree EmitCoercionIfNecessary(Type/*?*/ from, Type/*?*/ to, IToken tok, ConcreteSyntaxTree wr) {
       from = from == null ? null : DatatypeWrapperEraser.SimplifyType(from);
       to = to == null ? null : DatatypeWrapperEraser.SimplifyType(to);
 
       if (UnicodeCharEnabled) {
         // Need to box from int to CodePoint, or unbox from CodePoint to int
 
-        if (IsObjectType(from) && to is { IsCharType: true }) {
-          wr.Write($"((dafny.CodePoint)(");
+        if (IsObjectType(from) && IsCharBasedType(to)) {
```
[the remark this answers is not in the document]
robin-aws: The difference is when `to` is `null`: `to is { IsCharType: true }` returns in `false, but `to.IsCharType` throws a `NullReferenceException`.

Source/DafnyCore/Compilers/Java/Compiler-java.cs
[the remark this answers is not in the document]
robin-aws: I reverted and everything passes just fine, I must have misinterpreted a failure earlier, and I didn't realize `IsCharType` was calling `NormalizeExpand()`. Thanks for the catch!

Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -1339,7 +1347,8 @@ protected class ClassWriter : IClassWriter {
       } else if (member is Function fn) {
         var wr = new ConcreteSyntaxTree();
         EmitNameAndActualTypeArgs(IdName(member), TypeArgumentInstantiation.ToActuals(ForTypeParameters(typeArgs, member, false)), member.tok, wr);
-        if (typeArgs.Count == 0 && additionalCustomParameter == null) {
+        // TODO: needs to check for argument conversions as well
```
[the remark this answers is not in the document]
robin-aws: I intended to leave it for now, but you're right a negative test would be a good idea. But I think it's just easier to implement this too - it's a fluke that `Arrays.dfy` exposed the return value problem, I should really add dedicated tests for this in something like `Strings.dfy` instead.

Source/DafnyCore/ConcreteSyntax/ConcreteSyntaxTree.cs
robin-aws: No these versions are useful precisely it makes multiple writes to a `ConcreteSyntaxTree` instead of building up a string. It's useful for cases where the body of the "loop" is much more efficiently or naturally expressed by writing to a   `ConcreteSyntaxTree` (often by calling a fairly involved method like `TrExpr` which will write multiple times) instead of producing a string value. In cases where the body can return a string instead, then as you say you can just do something like `wr.Write(Comma(things, thing => MakeAString(thing)))` instead with the existing `Util.Comma` methods.

In a more recent commit I refactored some existing code to use it, which I think is a good illustrative example: https://github.com/dafny-lang/dafny/pull/3630/files#diff-703a5c7aa014afff3f25a9b9ebaee0546db822636c524b905406db519b2c1663L4709-R4729

Source/DafnyCore/Compilers/Java/Compiler-java.cs
robin-aws: I made it `IsNativeObjectType` which is an improvement, but I still don't love it. It's checking whether the type is `java.lang.Object` or a type parameter, and is meant to test whether boxing/unboxing is needed.

RustanLeino (approved):

PR #3631, opened 2023-02-24
Allow the use of zips created by old package.py

atomb:
We can undo this once updates to `package.py` are done


atomb:
This is made obsolete by #3633.

PR #3635, opened 2023-02-24
fix: Default to /unicodeChar:0 in legacy CLI as well

robin-aws:
This was accidentally missed from #3623.

Mostly I just updated expect files or added `/unicodeChar:0` where the testing coverage wasn't really relevant, but in a few cases I added `/unicodeChar:1` variants under `Test/unicodechars` where testing both modes was better.


Test/comp/Datatype.dfy.expect
```diff
@@ -18,7 +18,7 @@ Stream.Next
 {Berry.Smultron, Berry.Jordgubb, Berry.Hallon}
 CoBerry.Hjortron true true false
 false
-42 q hello
+42 'q' hello
```
atomb, line 21: It's strange to me for this to be an artifact to switching to proper Unicode characters, but I'm not opposed to it.

Test/dafny2/Z-BirthdayBook.dfy.expect
```diff
@@ -1,3 +1,3 @@
 
 Dafny program verifier finished with 21 verified, 0 errors
-Send cards to: {Mike, Susan}
+Send cards to: {['M', 'i', 'k', 'e'], ['S', 'u', 's', 'a', 'n']}
```
atomb, line 3: This is a little weirder than the character instance. Is this the best way to print strings? Does it only come about due to refinement?

Test/metatests/InconsistentCompilerBehavior.dfy.testdafny.expect
```diff
@@ -39,11 +39,4 @@ Diff (changing expected into actual):
  
 
 Executing on C++...
-Execution failed, for reasons other than known unsupported features. Output:
-
-Dafny program verifier did not attempt verification
-(0,-1): Error: UserDefinedTypeName _#Func1
-
-Error:
-
```
atomb, line 48: Oh, so does this work now? Nice!

atomb (approved):
I had a couple of questions, but I _think_ the current status is good to go, assuming the answers to those questions are what I expect.

Test/comp/Datatype.dfy.expect
robin-aws: It's an intentional side-effect, although only documented internally: https://github.com/dafny-lang/dafny/pull/3629

As you're not the first to ask it might be worth following up on with documentation, but since the behavior of `print` is all but unspecified in the reference manual or anywhere else I'm not sure it's worth it. :)

Test/dafny2/Z-BirthdayBook.dfy.expect
robin-aws: See other answer, same root cause. Nothing to do with refinement at least.

Test/metatests/InconsistentCompilerBehavior.dfy.testdafny.expect
robin-aws: No it's actually that the C++ compiler gives up earlier because it doesn't support `/unicodeChar:1` :)

Test/dafny2/Z-BirthdayBook.dfy.expect
atomb: It seems to me that other tests were printing strings like "Mike" (adding double quotes), which seems like a nice behavior to me. Was I misinterpreting that, or is something different going on here?

robin-aws: It's only because this is the result of (effectively) `print {"Mike", "Susan"};`. So it's printing a `set<string>` rather that `string` values. The runtimes COULD handle this better, but they'd need to accept a `areElementsStrings` parameter which the compiler would fill in when it knows the target `set<T>` type is in fact a `set<string>`.

PR #3637, opened 2023-02-25
Allocated 4 default

RustanLeino:
This PR makes `/allocated:4` the default, which addresses soundness issues with `/allocated:3`.

The PR also deprecates the `/allocated` option. Going forward, Dafny will support only one encoding of the allocation state (namely, the one that's `/allocated:4`). As far as I know, the previous users of `/allocated:1` or `/allocated:2` are no longer using an `/allocated` setting.


keyboardDrummer (dismissed):

robin-aws (approved):

robin-aws:
Note that two jobs were hung for over an hour - I cancelled them in order to attempt a re-run, but then the PR was auto-merged anyway for some reason. I suspect an internal GitHub error of some kind. We can rely on the deep tests on this branch to catch things if there was actually a failure lurking there.

PR #3638, opened 2023-02-25
Another attempt at #3622

atomb:
Attempt to make the changes from #3622 without failing the nightly CI. This changes both the GitHub configuration _and_ `package.py` to include both Z3 4.8.5 and 4.12.1, from our [`solver-builds`](https://github.com/dafny-lang/solver-builds) repository. This leaves the default at 4.8.5, but makes it easy to switch without disturbing the build system further.

Several caveats with these solvers:
* They don't include arm64 binaries. Just like with 3.x, we'll be distributing x64 Z3 binaries to run on arm64 macOS. That works just fine, even if arm64 binaries would likely perform better. It won't be hard to provide arm64 binaries for at least Z3 4.12.1 in the future, but it'd take a bit more time to do it now.
* The Linux version doesn't work on some older Linux distributions. The backward compatibility is better than available from the official Z3 4.12.1 binaries, but less good than available from the official Z3 4.8.5 binaries. Doing better would be hard, and would probably require building a Docker container with a different Linux distribution to run inside an action to perform the Z3 build. I think the versions here should be reasonably good as they'll run on any non-deprecated GitHub runner as well as other Linux distributions less than about 3 years old.
* The names of the Dafny packages now use different OS/architecture naming conventions, using names derived from GitHub runner names.

I think these are all reasonable caveats for now, and can be incrementally improved upon in the future.


Scripts/package.py
```diff
@@ -76,50 +68,47 @@ def flush(*args, **kwargs):
...
+        return path.join(CACHE_DIRECTORY, z3_zip)
 
     @property
-    def MB(self):
-        return self.size / 1e6
+    def cached(self):
+        for z3_zip in self.z3_zips:
+          if not path.exists(self.local_zip(z3_zip)):
+              return False
+        return True
```
alex-chew, line 94: nit: can we rename these to clarify that they're referring to Z3 and not Dafny or something else?

Scripts/package.py
```diff
@@ -173,17 +162,18 @@ def pack(self):
...
+                with zipfile.ZipFile(self.local_zip(z3_zip)) as Z3_archive:
+                    for fileinfo in Z3_archive.infolist():
+                        fname = fileinfo.filename
                         z3_files_count += 1
                         contents = Z3_archive.read(fileinfo)
                         fileinfo.filename = Release.zipify_path(path.join(DAFNY_PACKAGE_PREFIX, Z3_PACKAGE_PREFIX, fname))
+                        if self.os_name != 'windows':
+                            # http://stackoverflow.com/questions/434641/
+                            fileinfo.external_attr = 0o100755 << 16
+                            fileinfo.create_system = 3  # lie about this zip file's source OS to preserve permissions
```
alex-chew, line 176: nit: It'd be good to dedupe this.

Scripts/package.py
atomb: Good call. I'll do that.

Scripts/package.py
atomb: Agreed.

alex-chew (approved):

PR #3652, opened 2023-02-28
chore: Don't release from non-master branches

fabiomadge:
Don't release from non-master branches.


robin-aws (approved):
Good safety mechanism for now. We'll need more mechanisms for fetching the latest nightly before 4.0 besides this though, since once we merge `main-4.0` back to master the same problem will happen.

robin-aws (approved):

PR #3654, opened 2023-02-28
Cancel obsolete CI test runs

davidcok:

### DIFF
--- a/.github/workflows/doc-tests.yml
+++ b/.github/workflows/doc-tests.yml
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ master, main-* ]
 
+concurrency:
+  group: unit-tests-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check-deep-tests:
      uses: ./.github/workflows/check-deep-tests-reusable.yml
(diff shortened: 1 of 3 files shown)


.github/workflows/doc-tests.yml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ master, main-* ]
 
+concurrency:
+  group: unit-tests-${{ github.event.pull_request.number || github.ref }}
```
keyboardDrummer, line 15: The group name prefix `unit-tests` confuses me. It looks like you're defining a single group for anything that refers to the same commit, so `commit-` instead of `unit-tests-` might be more straightforward (Do we even need `github.event.pull_request.number || github.ref` instead of `github.ref` ?)

keyboardDrummer (approved):

davidcok:
The code was a recommendation from github action resources. But I agree that it seems github.ref should be sufficient. As for the prefix -- it could be anything and it is not something anyone sees.

.github/workflows/doc-tests.yml
davidcok: Fixed up these matters in a followup PR

PR #3656, opened 2023-02-28
Update release process to work on non-master branches

robin-aws:
Also split up a compound release note file that the script wouldn’t understand.

Tested with `--dry-run`.

### DIFF
--- a/Scripts/prepare_release.py
+++ b/Scripts/prepare_release.py
@@ -213,11 +213,12 @@ class Version(NamedTuple):
 
 class Release:
     REMOTE = "origin"
-    MASTER_BRANCH = "refs/heads/master"
     NEWSFRAGMENTS_PATH = "docs/dev/news"
 
-    def __init__(self, version: str) -> None:
+    def __init__(self, version: str, source_branch: str) -> None:
         self.version = version
+        self.source_branch_name = source_branch
+        self.source_branch_path = f"refs/heads/{self.source_branch_name}"
         self.branch_name = f"release-{version}"
         self.branch_path = f"refs/heads/{self.branch_name}"
         self.tag = f"v{version}"
@@ -275,7 +276,7 @@ class Release:
         return Version.from_string(self.version)
 
     def _is_release_branch(self) -> bool:
-        return self._get_branch(check=False) in (self.MASTER_BRANCH, self.branch_path)
+        return self._get_branch(check=False) in (self.source_branch_path, self.branch_path)
 
     @staticmethod
     def _is_repo_clean() -> bool:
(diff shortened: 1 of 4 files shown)


Scripts/prepare_release.py
```diff
@@ -213,11 +213,12 @@ def comment(self):
...
-    MASTER_BRANCH = "refs/heads/master"
     NEWSFRAGMENTS_PATH = "docs/dev/news"
 
-    def __init__(self, version: str) -> None:
+    def __init__(self, version: str, source_branch: str) -> None:
         self.version = version
+        self.source_branch_name = source_branch
+        self.source_branch_path = f"refs/heads/{self.source_branch_name}"
         self.branch_name = f"release-{version}"
         self.branch_path = f"refs/heads/{self.branch_name}"
```
alex-chew, line 223: Since we're introducing a new set of branch name/paths, we should rename the existing ones to `release_branch_{name,path}` or something along those lines in order to avoid confusion.

Scripts/prepare_release.py
```diff
@@ -450,6 +451,7 @@ def parse_arguments() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description="Dafny release helper")
     parser.add_argument("--dry-run", help="Do not make any modifications (no file updates, no git commands).",
                         action="store_true")
+    parser.add_argument("--source-branch", help="Which branch to release", default="master")
```
alex-chew, line 454:
```suggestion
    parser.add_argument("--source-branch", help="Which branch to release from", default="master")
```

docs/dev/RELEASE.md
```diff
@@ -5,14 +5,15 @@
...
+   * has the intended branch checked out (usually `master`,
+     but may be another mainline branch such as `main-3.x`).
 
 1. Select a version number `$VER` (e.g., "3.0.0" or "3.0.0-alpha"). 
    The `major`.`minor`.`patch` numbers may already have been
    incremented since the last release, so they do not necessarily need
    to be updated. However, you may want to increment them further
    depending the types of changes that are in the release.
-1. Run `Scripts/prepare_release.py $VER prepare` from the root of the
+1. Run `Scripts/prepare_release.py $VER prepare --source-branch <this branch>` from the root of the
```
alex-chew, line 16: Should we mention that the argument is optional, and is `master` by default?

docs/dev/news/3641.feat
```diff
@@ -0,0 +1 @@
+Expose non-relaxed definite assignment (`/definiteAssignment:4`) in legacy CLI
```
alex-chew, line 1: Does the script handle duplicate issue/PR numbers correctly (even if the change type is different)?

alex-chew (changes requested):

docs/dev/RELEASE.md
robin-aws: Dammit `argparse` why didn't you do this for me automatically... :)

docs/dev/news/3641.feat
robin-aws: Looks like it, it only groups and sorts by kind, it never uses the PR as a key. You can't have multiple news fragments for the same PR and kind together, but that's only because you can't have two `1234.feat` files at once. :)

docs/dev/RELEASE.md
alex-chew: (bump for this suggestion)

robin-aws: Ah I fixed it for the `--help` of prepare_release.py itself but not here, fixed now.

alex-chew (approved):
LGTM

PR #3659, opened 2023-03-01
Remove included lemma body

keyboardDrummer:
Refactoring to simplify `Method.Body`, in preparation for removing references to `DafnyOptions.O`

This PR has the side-effect of slightly increasing the amount of resolved code, namely lemma's in included files have their body resolved even when `/verifyIncludedModules=false`. This is slightly less performant, but also more consistent and what you would expect from the option `verifyIncludedModules`

Also includes a fix for `lit.site.cfg` so it can find Z3 again


Test/lit.site.cfg
```diff
@@ -207,16 +207,17 @@ elif "18.04" in ver and "Ubuntu" in ver:
...
+        for file in files:
+            if file.endswith(name):
+                return name
+    return ""
 
 solverPath = \
-    lit.util.which("z3-4.8.5", solverRoots) or \
-    lit.util.which("cvc4", solverRoots)
+    find("z3-4.8.5", binaryDir) or \
+    find("z3-4.12.1", binaryDir) or \
```
atomb, line 219: It seems reasonable to include both 4.8.5 and 4.12.1 in the list of alternatives, but were you having trouble where not having 4.12.1 was causing failures? On `master`, 4.8.5 should still be the default, and both versions should be available on all current branches at the moment. I'm wondering whether there's something deeper going wrong that we should fix.

atomb (approved):

atomb (dismissed):

atomb (approved):

PR #3660, opened 2023-03-01
Add auditor support for the `:concurrent` attribute

### LINKED ISSUE #3463: {:concurrent} attribute support in auditor
Dafny verification is currently based on sequential execution of code. It CAN be safe to concurrently execute the compiled code but only under certain assumptions. For example, the .NET AWS Encryption SDK exposes entry points that are threadsafe because they do not read or write any shared state. [This RFC](https://github.com/dafny-lang/rfcs/pull/6) is intended to improve the robustness of this mechanism.

This feature request is the very first baby step towards making Dafny aware of concurrency. We would provide a warning in `dafny audit` for any Dafny program element with a body (constructors, methods, and functions?) annotated with `{:concurrent}`. The attribute would be a way to document the fact that the code will be used in a concurrent setting, so that the auditor can point out this assumption. The suggested mitigation would be something like "target language concurrent testing and manual review", since Dafny can't yet offer any better.

Later on we could provide more options to make Dafny happy without any warnings. The RFC above would make this warning go away if the attributed element had both `reads {}` and (often implicitly) `modifies {}`. Even later on when Dafny has language features to support concurrent execution, there would be ways to synchronize so that arbitrary logic could be proven correct even under concurrent execution.

Calling this a language definition change even though the initial change would only be in the auditor, since it would still be establishing semantics that interact with future language features.

atomb:
Adds support to `dafny audit` for reporting instances of the `{:concurrent}` attribute, intended to mark code that should be, but can't be proven to be, safe for use in concurrent settings.

Fixes #3463


Test/auditor/TestAuditor.dfy-ietf.md.expect
```diff
@@ -53,3 +53,7 @@
 ### Member `LoopWithoutBody`
 
 * Definition contains loop with no body. MUST provide a body.
+
+### Member `ConcurrentMethod`
+
+* Declaration has `{:concurrent}` attribute. MUST manually review and test in a concurrent setting.
```
robin-aws, line 59: Does the auditor have a way of guiding the user to further documentation about a particular class of assumption and mitigation? I feel like this one deserves a bit of text about how Dafny can't (yet) analyze this for you etc.

docs/dev/news/3660.feat
```diff
@@ -0,0 +1 @@
+The `dafny audit` command now reports instances of the `{:concurrent}` attribute, intended to flag code that is intended, but can't be proven, to be safe for use in a concurrent setting.
```
robin-aws, line 1:
```suggestion
The `dafny audit` command now reports instances of the `{:concurrent}` attribute, intended to flag code that is intended to be safe for use in a concurrent setting.
```

"but can't be proven" implies there is could that could be proven safe, but Dafny can't do that yet.

robin-aws (changes requested):
We should add this to the reference manual too at a minimum (and hopefully guide users to that entry for details). Perhaps we could even point to https://github.com/dafny-lang/rfcs/pull/6 as the necessary change to support proving this in many cases?

Test/auditor/TestAuditor.dfy-ietf.md.expect
atomb: It might make sense to add some text including a link to documentation, for all of the assumption types. I'll look into that.

atomb: I updated the reference manual to describe all of the forms of assumption that the auditor reports, including a link to rfcs#6 in the bullet for `{:concurrent}`. The generated report currently doesn't include a link to that section of the manual, however.

docs/DafnyRef/UserGuide.md
```diff
@@ -370,6 +370,49 @@ The command emits exit codes of
...
+
+* Any declaration marked with the `{:concurrent}` attribute.
+This is intended to indicate that the code is safe for use in a concurrent setting, but Dafny currently cannot prove that form of safety.
+The [addition of `reads` clauses to methods](https://github.com/dafny-lang/rfcs/pull/6) will be a step toward being able to prove safety in this case,
+but until it exists, careful inspection and testing are appropriate.
+
+* Any definition with an `assume` statement in its body.
+To improve assurance, attempt to convert it to an `assert` statement and prove that it holds.
+Alternatively, converting it to an `expect` statement will cause it to be checked at runtime.
+Such a definition will not be compilable unless the statement is also marked with `{:axiom}`.
```
robin-aws, line 397:
```suggestion
Such a definition will not be compilable unless the statement is also marked with `{:axiom}`.
Alternatively, converting it to an `expect` statement will cause it to be checked at runtime.
```

As is it sounds like you're implying an `expect` statement is not compilable without `{:axiom}`.

Test/auditor/TestAuditor.dfy-ietf.md.expect
robin-aws: I think that's good enough for now at least, so a user can search the manual for that attribute and get a hit.

robin-aws (changes requested):
Thanks for the addition to the reference manual - just a quick edit to that content and this LGTM!

robin-aws (approved):

PR #3663, opened 2023-03-02
No static options

keyboardDrummer:
Remove `DafnyOptions.O` and all references to it.

Builds on https://github.com/dafny-lang/dafny/pull/3659


Source/DafnyTestGeneration/ProgramModifier.cs
```diff
@@ -119,7 +120,7 @@ public abstract class ProgramModifier {
         }
       }
       toRemove.ForEach(x => program.RemoveTopLevelDeclaration(x));
-      return Utils.DeepCloneProgramAndReresolve(program, DafnyOptions.O);
+      return Utils.DeepCloneResolvedProgram(program, options);
```
atomb, line 123: This seems like a potentially substantive change. Are you making this change because `DeepCloseResolvedProgram` didn't exist before but is more efficient now that it does exist? Or is there another reason?

Source/DafnyTestGeneration/Utils.cs
```diff
@@ -55,57 +55,55 @@ public static class Utils {
...
       return copy;
     }
 
     /// <summary>
     /// Deep clone and re-resolve a Boogie program.
     /// </summary>
-    public static Microsoft.Boogie.Program
-      DeepCloneProgramAndReresolve(Microsoft.Boogie.Program program, DafnyOptions options) {
-      program = DeepCloneProgram(program);
+    public static Microsoft.Boogie.Program DeepCloneResolvedProgram(Microsoft.Boogie.Program program, DafnyOptions options) {
```
atomb, line 91: Now I see the answer to my previous question! I'm not sure that this name change is an improvement in clarity.

Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -90,7 +90,7 @@ public abstract class Expression : TokenNode {
 
   public override string ToString() {
     try {
-      return Printer.ExprToString(this);
+      return Printer.ExprToString(DafnyOptions.DefaultImmutableOptions, this);
```
atomb, line 93: Is there a reason to be sure that this doesn't require any non-default options? The same goes for other places where `DefaultImmutableOptions` is used.

atomb (changes requested):
I think at least one of my comments likely might lead to changes, but I could be convinced otherwise. :)

Source/DafnyCore/AST/Expressions/Expressions.cs
keyboardDrummer: The only reason to be sure is that the build passes. I think `ToString()` should only be used for debugging purposes.

Source/DafnyCore/AST/ExtremeLemmaBodyCloner.cs
```diff
@@ -21,7 +21,7 @@ public ExtremeLemmaBodyCloner(ExtremeLemma context, Expression k, ISet<ExtremePr
   }
 
   public override Expression CloneExpr(Expression expr) {
-    if (DafnyOptions.O.RewriteFocalPredicates) {
+    if (reporter.Options.RewriteFocalPredicates) {
```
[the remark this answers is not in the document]
keyboardDrummer: I agree on both accounts :-D

Source/DafnyTestGeneration/Utils.cs
atomb: Fair enough!

atomb (approved):

stefan-aws (approved):
Previously approved

PR #3668, opened 2023-03-02
chore: Restore publishing language server to nuget, fix allow_on_mac.sh for z3 paths

robin-aws:
Other nuget packages depend on the language server, so we have to break those dependencies before we can stop publishing it.

Also patched allow_on_mac.sh to account for the change in Z3 packaging (and be more robust/idempotent in general)


atomb (approved):
Looks good!

keyboardDrummer:
> Other nuget packages depend on the language server, so we have to break those dependencies before we can stop publishing it.

Which packages are those?

PR #3669, opened 2023-03-02
Restoring the ability to test on local forks

davidcok:

### DIFF
--- a/.github/workflows/doc-tests.yml
+++ b/.github/workflows/doc-tests.yml
@@ -8,6 +8,7 @@ name: Test documentation
 
 on:
   workflow_dispatch:
+  push:
   pull_request:
     branches: [ master, main-* ]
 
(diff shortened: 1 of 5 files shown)


robin-aws (changes requested):
First off let me apologize for removing the `push` trigger without spotting the additional logic designed to restrict that only to forks. I agree we should support running CI on forks easily somehow.

My typical workflow has been to create a dummy draft PR on my fork, based on my fork's copy of master (which I occasionally update when necessary). Once I have the PR cut every subsequent push triggers the CI, and I find it handy to browse the current set of changes in the PR anyway.

Would this workflow be acceptable instead of this solution? Part of the reason I prefer it is that it's much simpler to implement - we don't need all these extra `if` conditions everywhere.

davidcok:
So with a draft PR:
- we can run the CI locally?
- and it does not run on master/main-* ?

davidcok:
Draft PRs do have the disadvantage of spamming other people's email

robin-aws:
Draft PR *on my fork* though, so it won't spam anybody but me. :)

davidcok:
Closing as abandoned

PR #3678, opened 2023-03-03
chore: Fix release downloads verification workflow

### LINKED ISSUE #3712: Test Release Downloads workflow has been failing for months
### Dafny version

4.0.0

_No response_

_No response_

### What happened?

See the history of the workflow: https://github.com/dafny-lang/dafny/actions/workflows/release-downloads.yml

It currently runs only when something is published. But either then or by manual triggering the workflow fails to download artifacts to test that the published artifacts actually work.

Furthermore, there seems to be no mechanism to alert anyone that they are failing.

Mac

robin-aws:
Fixes #3712 

These have been broken for several releases. Fixing two issues:

1. The OS identifiers were out of date.
2. Both the workflow steps and quicktest.sh created files like `a.dfy`, and on some OS' they interfered with each other (probably depending on how they decided on the current directory). The quick fix was to reorder them, but do we really need both `quicktest.sh` and what seems like equivalent workflow steps?

Also fixed the NuGet version, mostly related to Z3 path issues, but also lots of small bits of rot because of these workflows not being run.

Tested with runs from my fork:
* https://github.com/robin-aws/dafny/actions/runs/4368772189
* https://github.com/robin-aws/dafny/actions/runs/4368722815

Note that one set fails on non-windows machines because `quicktest.sh` actually changed behavior with the change in string and character printing in Dafny 4.0. I considered changing the workflow to use the checked-in quicktest.sh to fix this, but since this is a legitimate (although small) breakage in what we shipped, it's better to test that part of the release. Since this PR also fixes `quicktest.sh`, it should pass for the next release (and hopefully we can verify this with a nightly release after this is merged).


davidcok:
I agree that there is duplication in the smoke tests. Plus they should be upgraded to the new CLI. I created issue #3679 for this task.

robin-aws:
I think I can fully fix the nuget workflow too, don't merge this just yet...

jtristan (approved):

robin-aws:
Temporarily closing to save CI while I test on my fork

.github/workflows/release-downloads-nuget.yml
```diff
@@ -33,15 +33,6 @@ jobs:
...
         os: [ ubuntu-22.04, ubuntu-20.04, macos-11, windows-2019 ]
-        include:
-        - os:  'ubuntu-22.04'
-          osn: 'ubuntu\-22.04'
-        - os:  'ubuntu-20.04'
-          osn: 'ubuntu\-20.04'
-        - os:  'macos-latest'
-          osn: 'osx-11'
-        - os:  'windows-2019'
-          osn: 'win'
```
atomb, line 44: Nice simplification. :)

atomb (approved):
Thanks for fixing this!

PR #3706, opened 2023-03-07
Generate Rust-assert warning

RustanLeino:
In Rust, one writes an assertion using a function named `assert!`. I believe the `!` at the end of the name is a convention for telling the caller that this function may panic. Thus, to assert that the condition `x == 0` holds in Rust, one writes

``` rust
assert!(x == 0);
```

To a Dafny programmer, this looks totally wrong, because it seems the _negation_ of `x == 0` is being asserted. That is, to the eyes of a Dafny programmer, the Rust function call looks like

``` dafny
assert !(x == 0);
```

If someone used to Rust writes Dafny, there's a chance of accidentally using the Rust `assert!` and thus ending up with the negation of what's intended. This PR detects this situation and outputs a warning about it. More precisely, if a legal program contains `assert` followed by `!` followed `(`, with no space between the `assert` and the `!`, then a warning is output.

The warning is not likely to be spurious or be a nuisance for Dafny programmers, because it is unlikely that a Dafny program would omit this space on purpose, and even less likely that a Dafny programmer would have any objections to adding a space there to prevent the warning. 


Source/DafnyCore/Resolver/Resolver.cs
```diff
@@ -464,7 +464,7 @@ enum ValuetypeVariety {
       if (Options.DisallowConstructorCaseWithoutParentheses) {
         rewriters.Add(new ConstructorWarning(reporter));
       }
-      rewriters.Add(new UselessOldLinter(reporter));
+      rewriters.Add(new UselessOldAndRustAssertLinter(reporter));
```
MikaelMayer, line 467: Because you are combining seemingly two very different linters, and there might be case for other things like that in the future, could you rename this one to `LocalInter` ? `PrecedenceLinter` is everything but local so it can stay in its own class.

Source/DafnyCore/Rewriters/UselessOldAndRustAssertLinter.cs
```diff
@@ -46,5 +46,18 @@ class UselessOldLinterVisitor : TopDownVisitor<Unit> {
       }
       return base.VisitOneExpr(expr, ref st);
     }
+
+    protected override bool VisitOneStmt(Statement stmt, ref Unit st) {
+      // Check if the statement has the three tokens "assert", "!", and "(" next to each other. If so, that may
+      // be a mistake, especially if the programmer is coming from Rust.
+      if (stmt is AssertStmt { Expr: UnaryOpExpr { Op: UnaryOpExpr.Opcode.Not, E: ParensExpression } negationExpression }) {
+        if (stmt.tok.pos + 6 == negationExpression.tok.pos) {
```
MikaelMayer, line 54:
```suggestion
        if (stmt.tok.pos + 6/*="assert".length*/ == negationExpression.tok.pos) {
```

Source/DafnyCore/Rewriters/UselessOldAndRustAssertLinter.cs
```diff
@@ -46,5 +46,18 @@ class UselessOldLinterVisitor : TopDownVisitor<Unit> {
...
+          reporter.Warning(MessageSource.Rewriter, ErrorDetail.ErrorID.None, stmt.tok,
+            "You have written an assert statement with a negated condition, but the lack of whitespace between 'assert' and '!' " +
+            "suggests you may be used to Rust and have accidentally negated the asserted condition. Consider adding a space between these two.");
```
MikaelMayer, line 57:
```suggestion
            "suggests you may be used to Rust and have accidentally negated the asserted condition. If that is the case, please remove the "!" and the parentheses or, if it was intended, please add a space between these two.");
```
Otherwise, the "consider" is not in line with the previous statement.

MikaelMayer (changes requested):
What a nice effort to ensure Rust developers have a good experience in Dafny !
I just have a remark about consistency on the error message and another one, but otherwise, let's ship it !

Source/DafnyCore/Resolver/Resolver.cs
RustanLeino: "Local" -- nice way to make the distinction. I'll do that.

Source/DafnyCore/Rewriters/LocalLinter.cs
```diff
@@ -46,5 +46,19 @@ class UselessOldLinterVisitor : TopDownVisitor<Unit> {
...
+
+    protected override bool VisitOneStmt(Statement stmt, ref Unit st) {
+      // Check if the statement has the three tokens "assert", "!", and "(" next to each other. If so, that may
+      // be a mistake, especially if the programmer is coming from Rust.
+      if (stmt is AssertStmt { Expr: UnaryOpExpr { Op: UnaryOpExpr.Opcode.Not, E: ParensExpression } negationExpression }) {
+        if (stmt.tok.pos + 6/*="assert".Length*/ == negationExpression.tok.pos) {
+          reporter.Warning(MessageSource.Rewriter, ErrorDetail.ErrorID.None, stmt.tok,
+            "You have written an assert statement with a negated condition, but the lack of whitespace between 'assert' and '!' " +
+            "suggests you may be used to Rust and have accidentally negated the asserted condition. If you did not intend the negation, " +
+            "remove the '!' and the parentheses; if you do want the the negation, please add a space between the 'assert' and '!'.");
```
MikaelMayer, line 58:
```suggestion
            "remove the '!' and the parentheses; if you do want the negation, please add a space between the 'assert' and '!'.");
```

MikaelMayer (approved):

PR #3710, opened 2023-03-08
Ghost stutter is compilable

RustanLeino:
This PR expands the code that is considered legal. More specifically, based on an observation about ghost parameters, this PR treats some `if`-`then`-`else` expressions as _compiled_, whereas they previously were considered to be _ghost_.

## Illustrative example

For illustration, consider the following (contrived way to write an) identity function:

``` dafny
function F(x: int, ghost n: nat): int {
  if n < 7 then
    x
  else
    F(x, n - 1)
}
```

Function `F` is compiled (i.e., non-ghost), since its body is expected to be compilable. Ordinarily, an `if`-`then`-`else` (ITE) whose test (here, `n < 7`) is ghost is itself ghost. Under that rule, the body of `F` would be rejected (and that's what Dafny has done previously).

However, if the `else` branch is taken*, then the function makes a recursive call in which only the value of ghost parameters are different. Eventually, such calls must end up taking some other branch. In fact, the branch that eventually will be taken must be the other branch (here, the `then` branch), since, until this ITE, all branch tests (here, none) were based on compiled variables. Hence, at run time, the code might as well ignore the test and pick the `then` branch right away.**

*) To understand "taken" in the previous paragraph, you'll need to, in your head, augment the non-ghost computation with the ghosts that the compiler erases.
**) If you erase the ghost parameter, like the compiler does, then "taking" the `else` branch is like a _stuttering step_ in the compiled code.

From this observation, the entire ITE above turns into just the `then` branch, namely the expression `x`. Since `x` is not ghost, it's a fine body for `F`. This PR identifies the situation described and will admit the ITE above as a compiled expression.

## Motivating example

The identity function above does not motivate the special case. But here is a function that does. It comes from a tutorial @jtristan and I wrote recently. I won't give all the details here; all you need to know is that `Stream` is a possibly infinite list (defined by a `codatatype`) and `IsFinite` is a `least predicate` (so, the prefix function `IsFinite#` is a ghost function that takes a parameter `k` of type `ORDINAL`, syntactically given in square brackets). We'd like to express a function that computes the length of a finite `Stream`. Here's a way to do that:

``` dafny
function Length(ghost k: ORDINAL, s: Stream): nat
  requires IsFinite#[k](s)
{
  if s == Nil then
    0
  else if k.Offset != 0 then
    1 + Length(k - 1, s.tail)
  else
    var k' :| k' < k && IsFinite#[k'](s);
    Length(k', s)
}
```

Note that the ITE test `k.Offset != 0` is ghost, and note that the `else` branch evaluates to a recursive call where all the non-ghost parameters are unchanged. Hence, from a compiled perspective, the inner ITE can be thought of as just the expression `1 + Length(/*k - 1,*/ s.tail)`. Thus, `Length` can be compiled!

It possible to manually rewrite `Length` to make it compiled without this PR. However, the rewrite is cumbersome. The manual rewrite might be called the "reduce `k` trick". The idea is to replace the inner ITE with the expression

``` dafny
1 + Length(ReduceK(k, s), s.tail)
```

where `ReduceK(k, s)` returns the value of `k` after the stuttering is done. This PR renders the "reduce `k` trick" unnecessary, and the resulting Dafny code (namely, the original `Length` function above) is easier to understand.

## Details

Ghost stuttering occurs in the following situation:  The body of a function contains a number of choices that ultimate lead to an ITE where the ITE test is ghost and one of the ITE branches boils down to a recursive call where all non-ghost parameters are the same as those given.

* "are the same" is checked syntactically.
* The "choices" can be ITE or `match` expressions.
* In comparing the situation with what I just described,
    - any `StmtExpr` (like `assert P; E`) is treated just like the expression part (`E` in `assert P; E`), and
    - any let-expression is, in effect, treated like a substitution into the body of the let-expression.


jtristan:
@RustanLeino I'm trying to think about what the general principal is. Intuitively, as a first approximation, it seems like it could be something like this:
For any function F with concrete parameters x_1, ..., x_n and ghost parameters g_1, ..., g_m, if there exists a computable function C such that forall x_1, ..., x_n, g_1, ..., g_m F(x_1, ..., x_n,g_1, ..., g_m) == C(x_1, ..., x_n), then we can consider F to be compilable and replace F with C in the compiler. 

If so, then the rules you suggest above might indeed be correct because the syntactic check should imply that the ghost control flow does not affect the value returned by the function.

RustanLeino:
@jtristan Yes, that general principle seems correct.

Btw, for a particular instance of such an `F`, it may be possible to manually write the computable function `C` and prove the correspondence in Dafny. But it's not certain that you can do that in Dafny, because it could be that function `C` would not be well-formed in Dafny. For example, maybe proving termination of `C` requires the additional ghost parameters and additional logic (or perhaps requires the "`ReduceK` trick" that you and I used in the tutorial note we wrote). So, the reasoning that `F(x, g) == C(x)` probably needs to be done at a meta-level.

PR #3715, opened 2023-03-08
Run integration tests in single process

keyboardDrummer:
### Changes

- `dafny run --target=cs` now uses a separate process to run, so the statics of the compiled source code, such as `Console`, are not shared with the Dafny Compiler
- Uses of `Console.` are replaced with non-static references.
- Move `DafnyFile` class into its own file
- Remove `ThreadTaskScheduler` which creates one-shot threads, and instead use a thread pool. Also consistently use large threads for resolution, translation to Boogie, printing and compilation.
- Made counter example IDE support thread-safe
- The TestDafny project now uses the new CLI UI


robin-aws:
Awesome progress on a difficult and overdue problem.

Haven't looked too closely yet, but since `InvokeMainMethodsDirectly` is still stuck on `false`, I'd suggest we find a way to run SOME integration tests in this mode to avoid regressing (and I might push a commit for this myself).

keyboardDrummer:
> Awesome progress on a difficult and overdue problem.
> 
> Haven't looked too closely yet, but since `InvokeMainMethodsDirectly` is still stuck on `false`, I'd suggest we find a way to run SOME integration tests in this mode to avoid regressing (and I might push a commit for this myself).

The goal of this PR is to merge the made changes to avoid merge conflicts in the future. I think it's OK if there is some regressions after that. When we manage to complete this feature by changing the `InvokeMainMethodsDirectly` flag, we can resolve those regressions.

robin-aws (commented):
Just reviewed the test file and workflow changes in this batch

Source/DafnyCore/Compilers/Dafny/DafnyBackend.cs
```diff
@@ -41,13 +41,68 @@ public class DafnyBackend : ExecutableBackend {
...
    */
-  public override bool RunTargetProgram(string dafnyProgramName, string targetProgramText, string /*?*/ callToMain,
-    string targetFilename, ReadOnlyCollection<string> otherFileNames, object compilationResult, TextWriter outputWriter) {
+  public override bool RunTargetProgram(string dafnyProgramName, string targetProgramText, string callToMain, /*?*/
+    string targetFilename, ReadOnlyCollection<string> otherFileNames, object compilationResult, TextWriter outputWriter,
+    TextWriter errorWriter) {
     Contract.Requires(targetFilename != null || otherFileNames.Count == 0);
 
-    return RunTargetDafnyProgram(targetFilename, outputWriter);
+    /*
```
robin-aws, line 49: Bad merge, the code below is in `RunTargetDafnyProgram` now, which is also used by the `LibraryBackend`. Doesn't look like you had to change the code though? It's just fine if it's still using a subprocess.

Source/DafnyCore/Compilers/JavaScript/JavaScriptBackend.cs
```diff
@@ -73,9 +74,10 @@ public class JavaScriptBackend : ExecutableBackend {
       nodeProcess.StandardInput.Close();
       // Fixes a problem of Node on Windows, where Node does not prints to the parent console its standard outputs.
       var errorProcessing = Task.Run(() => {
-        PassthroughBuffer(nodeProcess.StandardError, Console.Error);
+        // TODO fix.
```
robin-aws, line 77: Fix what exactly?

Source/DafnyCore/Compilers/Library/LibraryBackend.cs
```diff
@@ -80,7 +80,7 @@ public class LibraryBackend : ExecutableBackend {
   }
 
   public override bool RunTargetProgram(string dafnyProgramName, string targetProgramText, string callToMain, string targetFilename,
-    ReadOnlyCollection<string> otherFileNames, object compilationResult, TextWriter outputWriter) {
+    ReadOnlyCollection<string> otherFileNames, object compilationResult, TextWriter outputWriter, TextWriter errorWriter) {
 
     var dooPath = DooFilePath(dafnyProgramName);
     return RunTargetDafnyProgram(dooPath, outputWriter);
```
robin-aws, line 86: This probably needs an `errorWriter` parameter too?

Source/DafnyCore/Compilers/Python/PythonBackend.cs
```diff
@@ -61,6 +61,14 @@ public class PythonBackend : ExecutableBackend {
     return true;
   }
 
+  public override void CleanSourceDirectory(string sourceDirectory) {
```
robin-aws, line 64: Shouldn't this just delete the whole `sourceDirectory`, just like the `JavaBackend`?

Source/DafnyCore/Compilers/CSharp/CsharpBackend.cs
```diff
@@ -93,54 +93,40 @@ public class CsharpBackend : ExecutableBackend {
     var outputDir = targetFilename == null ? Directory.GetCurrentDirectory() : Path.GetDirectoryName(Path.GetFullPath(targetFilename));
     var outputPath = Path.Join(outputDir, Path.GetFileNameWithoutExtension(Path.GetFileName(dafnyProgramName)) + ".dll");
     var outputJson = Path.Join(outputDir, Path.GetFileNameWithoutExtension(Path.GetFileName(dafnyProgramName)) + ".runtimeconfig.json");
-    if (inMemory) {
```
robin-aws, line 96: Might be worth mentioning this change of behavior in a release note (that C# always emits source code now). I can't find any note of it in documentation, and I do like making this more consistent with the other backends, but it could be surprising to some customers that have been using Dafny for a long time.

Source/DafnyCore/Rewriters/IRewriter.cs
```diff
@@ -13,6 +13,22 @@ public abstract class IRewriter {
...
+    /// * Have no PluginConfiguration  class, and an ErrorReporter will be provided to your class's constructor.
+    /// 
+    /// Then you can use the protected field "reporter" like the following:
+    /// 
+    ///     reporter.Error(MessageSource.Compiler, token, "[Your plugin] Your error message here");
+    ///
+    /// The token is usually obtained on expressions and statements in the field `tok`
+    /// If you do not have access to them, use moduleDefinition.GetFirstTopLevelToken()
+    /// </summary>
+    /// <param name="reporter">The error reporter. Usually outputs automatically to IDE or command-line</param>
```
robin-aws, line 31: Nice!

Source/DafnyCore/UndisposableTextWriter.cs
```diff
@@ -0,0 +1,247 @@
+using System;
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Dafny;
+
+public class UndisposableTextWriter : TextWriter {
```
robin-aws, line 9: What's the story behind this? I could search through the changes but I think it deserves a comment. :)

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -97,19 +100,29 @@ public enum ExitValue { SUCCESS = 0, PREPROCESSING_ERROR, DAFNY_ERROR, COMPILE_E
...
-        new System.Threading.ThreadStart(() => { ret = ThreadMain(args); }),
-          0x10000000); // 256MB stack size to prevent stack overflow
-      thread.Start();
-      thread.Join();
-      return ret;
+      return MainWithWriter(Console.Out, Console.Error, Console.In, args);
     }
 
-    public static int ThreadMain(string[] args) {
+    public static int MainWithWriter(TextWriter outputWriter, TextWriter errorWriter, TextReader inputReader,
```
robin-aws, line 106: nit:

```suggestion
    public static int MainWithWriters(TextWriter outputWriter, TextWriter errorWriter, TextReader inputReader,
```

Source/IntegrationTests/LitTests.cs
```diff
@@ -161,6 +157,12 @@ public class LitTests {
       Config = new LitTestConfiguration(substitutions, commands, features, DafnyDriver.ReferencedEnvironmentVariables);
     }
 
+    public static ILitCommand DafnyCommand(IEnumerable<string> arguments, LitTestConfiguration config, bool invokeDirectly) {
```
robin-aws, line 160: I've been trying to make "xlit" generic enough that we could one day apply it to Boogie as well. Would you mind adding a TODO to make this slightly more generic, by defining an interface with the `MainWithWriters` method that other test suites could define as well? That way we could have a single `MainMethodWithWritersCommand` that would point to such a method instead of implicitly "the main method of this assembly".

robin-aws (changes requested):
Reviewed the rest. Again excellent work slogging through a hard problem.

Test/git-issues/git-issue-684.dfy
```diff
@@ -1,7 +1,7 @@
 // RUN: %dafny /compile:0 "%s" > "%t"
 // RUN: %dafny /noVerify /compile:4 /compileTarget:cs "%s" >> "%t"
 // RUN: %dafny /noVerify /compile:4 /compileTarget:js "%s" >> "%t"
-// RUN: %exits-with 3 %dafny /noVerify /compile:4 /compileTarget:go "%s" >> "%t"
+// RUN: %dafny /noVerify /compile:4 /compileTarget:go "%s" >> "%t"
```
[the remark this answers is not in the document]
keyboardDrummer: I created an issue: https://github.com/dafny-lang/dafny/issues/3953 the bug is subtle and hard to reproduce.

Source/DafnyCore/Compilers/Dafny/DafnyBackend.cs
keyboardDrummer: Already on master, this code doesn't work for library back-end btw. It calls Dafny using the old CLI syntax and passing a .doo file, which causes it to crash on DooFile.Validate.

Source/DafnyCore/Compilers/JavaScript/JavaScriptBackend.cs
keyboardDrummer: No idea. Removed it.

Source/DafnyCore/UndisposableTextWriter.cs
keyboardDrummer: It's currently unused, only referenced in commented code:
```
      // Code that shouldn't be needed, but prevents some exceptions when running the integration tests in parallel
      // outputWriter = new UndisposableTextWriter(outputWriter);
      // errorWriter = new UndisposableTextWriter(errorWriter);
      // outputWriter = TextWriter.Synchronized(outputWriter);
      // errorWriter = TextWriter.Synchronized(errorWriter);
```

Basically, one exception that when running tests in parallel is usage of already disposed writers, so somehow the tests were sharing writers, and this wrapper prevents them from being disposed which removed those crashes, but it's a patch for problem that needs a better solution.

keyboardDrummer: I could remove this since it's currently not used, but it can be useful in debugging when we try get this to pass on CI. I'd rather keep it in for now.

Source/IntegrationTests/LitTests.cs
keyboardDrummer: I'm not a fan of leaving TODOs in code. I think a GH issue would be better. Also I'm sure what you're describing

Source/DafnyCore/UndisposableTextWriter.cs
robin-aws: Yeah once I spotted the commented out code I understood, we can leave it as is for now.

Source/TestDafny/MultiBackendTest.cs
```diff
@@ -89,6 +94,11 @@ public class TestDafny {
     var success = true;
     foreach (var plugin in dafnyOptions.Plugins) {
       foreach (var compiler in plugin.GetCompilers(dafnyOptions)) {
+        if (compiler.TargetId == "lib") {
+          // Some tests still fail when using the lib back-end, for example due to disallowed assumptions being present in the test,
+          // Such as empty constructors with ensures clauses, generated from iterators
+          continue;
+        }
```
robin-aws, line 101: Ah this is how you kept it all green :) This is reasonable for now, but it'll be good to dig into those tests to see if the other backends are too permissive about this.

Source/DafnyCore/Compilers/Dafny/DafnyBackend.cs
robin-aws: I got that fixed FYI: https://github.com/dafny-lang/dafny/pull/3967

robin-aws (approved):
LGTM, but FYI I turned off the auto-merge just to make @fabiomadge's life easier as he's trying to release 4.1. If he's fine releasing from a branch go ahead and merge!

PR #3718, opened 2023-03-08
Add associativity axioms for seq concatentation

### LINKED ISSUE #3394: Add axioms about associativity of concatenation
### Summary

Make verification more performant and less variable by telling Z3 that Dafny's sequence concatenation is associative.

### Background and Motivation

Code that heavily makes use of sequence concatenation often results in slow verification with unpredictable success. In many cases, re-grouping repeated concatenations to match the associativity used by another part of the goal can dramatically improve performance. This is tedious and error-prone, however.

### Proposed Feature

When the Dafny verifier translates a Dafny program to Boogie, it can look for any instance of nested concatenation in the program, something of the form `a + b + c`, and emit a specific instance of the associativity axiom equating `a + (b + c)` with `(a + b) + c`. Automatically instantiating the more general axiom for each instance will avoid a potential blowup in the proof process due to the search for useful quantifier instantiations.

### Alternatives

One alternative would be to require Dafny programmers to add such axioms (or proven lemmas) themselves. Many Dafny programmers do this already. However, this makes Dafny programs messier, and requires deep analysis by the programmer that's fundamentally unnecessary.

It would be possible to increase assurance by adding goals to prove each instance of the associativity theorem, so that these would no longer be axioms. That would result in re-proving the same facts many times, however.

atomb:
Every instance of (a + b) + c or a + (b + c) now gets an additional axiom about its equality with the other form.

Fixes #3394

Still a draft until I do more performance evaluation.

### DIFF
--- a/Source/DafnyCore/AST/Expressions/Expressions.cs
+++ b/Source/DafnyCore/AST/Expressions/Expressions.cs
@@ -196,6 +196,16 @@ public abstract class Expression : TokenNode {
     return s;
   }
 
+  public static Expression CreateConcat(Expression e0, Expression e1) {
+    Contract.Requires(e0 != null);
+    Contract.Requires(e1 != null);
+    Contract.Ensures(Contract.Result<Expression>() != null);
+    var s = new BinaryExpr(e0.tok, BinaryExpr.Opcode.Add, e0, e1);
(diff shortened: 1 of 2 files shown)


MikaelMayer:
Any update on this one?

atomb:
It generally didn't provide substantial benefit, last time I checked. I've been leaning toward closing this PR for a while. Any objections?

PR #3721, opened 2023-03-08
Small miscellaneous edits + edits for 3614, 3481, 3298

### LINKED ISSUE #3298: Replace `dafny --dev <verb> --help` by `dafny <verb> --help-dev`
### Summary

Improving the CL for hidden help

### Background and Motivation

The current `dafny --dev <verb> --help` is not intuitive and does not fit the rest of the new CLI design

### Proposed Feature

Implement `dafny <verb> --help-dev` instead

### Alternatives

_No response_

### LINKED ISSUE #3481: Erroneous example in Reference

The reference says

```
newtype int32 = x | -0x8000_0000 <= x < 0x8000_0000

then the code fragment is erroneous, since the result of the addition may fail to satisfy the predicate in the definition of int32. The code fragment can be rewritten as

var mid := lo + (hi - lo) / 2;

in which case it is legal for both int and int32.
```

but the following code, correctly, does not verify

```
newtype int32 = x | -0x8000_0000 <= x < 0x8000_0000

function method foo(x : int32, y : int32) : int32
{
  if x < y then
    x + (y - x) / 2
  else
    y + (x - y) / 2
}
```

Although it would verify with an unsigned type.

### LINKED ISSUE #3614: Declaring a synonym masks datatype
### Dafny version

3.12

```dafny
datatype D<T> = D(t: T)

type Dint = D<int>

function f(i: int): Dint { Dint(i) }
```

```code
dafny resolve test.dfy
```

### What happened?

temp.dfy(5,31): Error: name of type synonym (Dint) is used as a function

But we claim that synonyms act just like their expanded version, but here it does not

Mac

davidcok:
Fixes #3298. Fixes #3481. Fixes #3614.


davidcok:
Replaced by https://github.com/dafny-lang/dafny/pull/3761 with the same content

PR #3722, opened 2023-03-09
Use Xunit for all automated tests

keyboardDrummer:


Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -99,66 +98,66 @@ decreases y
...
-      Assert.AreEqual(0, resolutionDiagnostics.Length);
-
-      var verificationDiagnostics = await diagnosticsReceiver.AwaitNextDiagnosticsAsync(CancellationToken, documentItem);
-      Assert.AreEqual(0, verificationDiagnostics.Length);
-
-      await AssertNoDiagnosticsAreComing(CancellationToken);
-    }
-
-    [TestMethod, Timeout(MaxTestExecutionTimeMs)]
+    // [Fact(Timeout = MaxTestExecutionTimeMs)]
```
robin-aws, line 136: Why didn't this one work?

Source/XUnitExtensions/FileTestCase.cs
```diff
@@ -1,11 +1,11 @@
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
 using Xunit.Sdk;
+using SkipException = Xunit.SkipException;
```
robin-aws, line 8: Does this mean we can drop `Xunit.SkippableFact` from the csproj now (because it's part of xunit core now)?

robin-aws (changes requested):
🚀 (aside from the commented out test)

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
keyboardDrummer: It's turned off on master, and when I turned it on it failed, so keeping that like it is in this PR

Source/XUnitExtensions/FileTestCase.cs
keyboardDrummer: If I remove that package reference the. I have to replace `SkippableTestMessageBus` and `SkippableFactTestCase` with something and I'm not entirely sure what. If you're already aware of the new API, could you make that change in a follow-up PR?

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
robin-aws: Oh I missed that just the attribute was already commented out.

You can do `[Fact(Timeout = MaxTestExecutionTimeMs, Skip = "Not working")]` instead. That way it still shows up in the test run and we don't completely forget about it.

robin-aws (approved):

PR #3728, opened 2023-03-10
(Test Generation) Address #3716 by adding an option that controls inlining

### LINKED ISSUE #3716: Allow annotating methods/functions to control inlining for test generation
### Summary

Have an annotation or option to mark a given method as "to be covered" if it is reachable in the call graph of the entry point for test generation.

### Background and Motivation

Currently one way to control the coverage requirements of the test generation is to use the `--inline-depth n` argument, which asks for coverage of all the blocks of code within methods at most `n` steps away from the entry point in the call-graph. However there is a strong potential here for path explosion, the usual problem encountered with this kind of symbolic test generation.

Ideally things would be a bit more fine-grained, by marking exactly which methods should be covered, if reachable from the entry point.

### Proposed Feature

One could imagine either a command line feature like `--inline-for-tests SomeModule.SomeMethod`, or an attribute `{:test-inline}` attached to the method/function method declaration itself.

I can imagine 2 semantics for this annotation:

- The method, *and all methods above it in the call graph* are inlined, and the test generation tool attempts to cover each block of the annotated method using the usual approach.
- The method is inlined iff all the methods above it are also marked as inlined (or if it is directly reachable in 1 step from the entry point).

The second semantics seems more "honest" in terms of the semantics, but may lead to confusing behavior if the method is "hidden" by a non-inlined method. Maybe a warning is appropriate in this case.

### Alternatives

I don't currently have a case where this feature allows test generation to succeed (i.e. both terminate and cover the desired blocks) when it otherwise would fail, though I suspect I will soon run into this. I'm not sure what an alternative would be.

Dargones:
Fixes #3716

This PR adds a new command line option to test generation that allows the user to enter a comma-separated list of all methods that should be inlined for a given test generation run. @codyroux, would something like this work in your use case? I decided to prototype this as an argument since this allows for more flexibility but I can rewrite this to work as an attribute like you describe in the issue.

Consider the following Dafny program (added as a test as part of this PR):
```Dafny
module M {
  class Inlining {
    static function max(a:int, b:int):(c:int) {
      if (a > b) then a else b
    }
    static function min(a:int, b:int):(c:int) {
      if (a < b) then a else b
    }
    static method test(a:int, b:int, c:int, d:int) returns (r:int) {
      r := max(a, b);
      r := min(c, d);
    }
  }
}
```
Running `dafny generate-tests inline.dfy --inline:M.Inlining.min --target-method:M.Inlining.test` will produce the following two tests, since `min` is inlined but `max` is not:

```Dafny
method {:test} test0() {
  var r0 := M.Inlining.test(8856, 8855, 1236, 1237);
}
method {:test} test1() {
  var r0 := M.Inlining.test(2437, 2436, 7719, 6483);
}
```


keyboardDrummer:
Commented on the design in the related issue

Dargones:
Closing this PR for now in lieu of #3865

PR #3729, opened 2023-03-10
Add CLI option and code placeholder

RustanLeino:
This PR sets up the CLI option and initial call into the to-be-added new type inference. The new type inference will allow Dafny's type system to grow. For example, it will allow us to add traits as supertypes of datatypes.

For now, the new CLI does not do anything (except cause a crash).


Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -142,6 +142,12 @@ public class CommonOptionBag {
 true - The char type represents any Unicode scalar value.".TrimStart()) {
   };
 
+  public static readonly Option<bool> TypeSystemRefresh = new("--type-system-refresh", () => false,
+    @"
+false - The type-inference engine and supported types are those of Dafny 4.0.
+true - Use an updated type-inference engine. Warning: This mode is under construction and probably won't work at this time.".TrimStart()) {
+  };
```
atomb, line 149:
```suggestion
    IsHidden = true
  };
```

Given that this option is very much for our own use (and even the name will become dated at some point), making it hidden seems appropriate.

atomb: I don't know if there's a way to do a similar thing for the legacy version of the option.

atomb (changes requested):
I like the idea of adding a shell and incrementally filling it in. It seems to me like the option would be better as a hidden option, though, as it's mostly for the use of us developers.

jtristan (approved):

atomb (approved):

PR #3738, opened 2023-03-14
Fix: Right shift on native byte has the same consistent semantics even in Java

### LINKED ISSUE #3724: Java code generation incorrect
### Dafny version

3.10 and 4.0

```dafny
newtype uint8 = x : int | 0 <= x < 256

predicate method IsHex(ch : char)
{
  || '0' <= ch <= '9'
  || 'a' <= ch <= 'f'
}

// return the hex character for this hex value
function method HexVal(x : uint8) : (res : char)
  requires x < 16
  ensures IsHex(res)
{
  if x < 10 then
    '0' + x as char
  else
    'a' + (x - 10) as char
}

// return the hex string for this byte
function method HexStr(x : uint8) : (ret : string)
  ensures |ret| == 2
  ensures IsHex(ret[0]) && IsHex(ret[1])
{
  if x < 16 then
    var res := ['0', HexVal(x)];
    res
  else
    var y := x as bv8;
    var res := [HexVal((y >> 4) as uint8), HexVal((y & 0xf) as uint8)];
    res
}

// return the hex string for these bytes, keeping any leading zero
function method {:tailrecursion} HexFmt(val : seq<uint8>) : (ret : string)
  ensures |ret| == 2 * |val|
  ensures forall ch <- ret :: IsHex(ch)
{
  if |val| == 0 then
    []
  else
    HexStr(val[0]) + HexFmt(val[1..])
}

  method {:test} TestBeacon() {
    var binary := [0x00,0x11,0x22,0x33,0x44,0x55,0x66,0x77,0x88,0x99,0xaa,0xbb,0xcc,0xdd,0xee,0xff];
    var str := "00112233445566778899aabbccddeeff";
    var conv := HexFmt(binary);
    print "\n", str, "\n", conv, "\n";
    expect str == conv;
  }
```

```code
This passes the test
dafny /functionSyntax:3 /compileTarget:cs /runAllTests:1 /compile:3 HexFail.dfy

This fails the test
dafny /functionSyntax:3 /compileTarget:java /runAllTests:1 /compile:3 HexFail.dfy
```

### What happened?

I expected both languages to pass.

Given the 
     ensures forall ch <- ret :: IsHex(ch)
 I expected java to only produce hex characters, but instead of
00112233445566778899aabbccddeeff
java produces
0011223344556677ŏ8Ő9őaŒbœcŔdŕeŖf

In HexStr, where I have
    var y := x as bv8;
    var res := [HexVal((y >> 4) as uint8), HexVal((y & 0xf) as uint8)];
If I replace it with
    var res := [HexVal((x / 16) as uint8), HexVal((x % 16) as uint8)];
then both languages are happy, so I'm guessing (y >> 4), with the leftmost bit set, does the "signed int" thing and extends the sign bit.

Mac

### LINKED ISSUE #3734: Incorrect right shift operation for byte in Java
### Dafny version

4.0.0

```dafny
newtype uint8 = x : int | 0 <= x < 256

method Main() {
  var y := 0x80 as bv8;
  var x := (y >> 4) as uint8;
  assert x != 248; // What Java obtains
  assert x == 0x8;
  if x == 248 {
    var x := 1/0;
  }
  print "All right";
}
```

```code
dafny run --target java file.dfy
```

### What happened?

This is a minimum reproductible example for #3724

```
Exception in thread "main" java.lang.ArithmeticException: BigInteger divide by z
ero
        at java.base/java.math.MutableBigInteger.divideKnuth(MutableBigInteger.j
ava:1184)
        at java.base/java.math.BigInteger.divideKnuth(BigInteger.java:2318)
        at java.base/java.math.BigInteger.divide(BigInteger.java:2299)
        at dafny.DafnyEuclidean.EuclideanDivision(DafnyEuclidean.java:90)
        at _System.__default.Main(__default.java:22)
        at _System.__default.__Main(__default.java:27)
        at git_issu_3724bis.lambda$main$0(git_issu_3724bis.java:10)
        at dafny.Helpers.withHaltHandling(Helpers.java:279)
        at git_issu_3724bis.main(git_issu_3724bis.java:10
```

Looks like the generated Java code is the following:
```java
    byte _0_y;
    _0_y = (byte) 128;
    byte _1_x;
    _1_x = ((byte) (byte)  ((byte)((_0_y) >>> ((byte) 4))));
    if ((_1_x) == ((byte) 248)) {
      java.math.BigInteger _2_x = java.math.BigInteger.ZERO;
      _2_x = dafny.DafnyEuclidean.EuclideanDivision(java.math.BigInteger.ONE, java.math.BigInteger.ZERO);
    }
    System.out.print((dafny.DafnySequence.asUnicodeString("All right")).verbatimString());
```

Transforming "4" in "3" and "0x80" in "0x40" does not trigger the stack overflow, so it's really the last byte.

Windows, Mac

MikaelMayer:
This PR fixes #3734  and fixes #3724.

Here is what happened.
In Java, since bytes are signed, we have that, even for the unsigned right shift ">>>",
```
(0b1000,1000 >>> 1) == 0b1100,0100
(0b1000,1000 >>> 2) == 0b1110,0010
(0b1000,1000 >>> 3) == 0b1111,0001
(0b1000,1000 >>> 4) == 0b1111,1000
```
Since negative bytes are implemented with the complement of two, and 0b1111,1000 + 0b1000 == 0, it means that 0b1111,1000 + 8 == 0, so 0b1111,1000 == -8, which is equal to 248 modulo 2.
That explains that, the result of (0x88 >> 4) in java was 248, which gives the strange character encoding observed in #3724.

We actually want that 
```
(0b1000,1000 >> 1) == 0b0100,0100
(0b1000,1000 >> 2) == 0b0010,0010
(0b1000,1000 >> 3) == 0b0001,0001
(0b1000,1000 >> 4) == 0b0000,1000
```
So I added a helper in the runtime to compute exactly this.
The trick is that : -1 === 0b1111,1111, and by left-shifting it to the left by a correct amount, we can use it to xor it to remove the extra 1 introduced by ">>>". For that, we need to shift it left by 8 minus original amount. Indeed:
```
(0b1111,1111 << (8-1)) == 0b1000,0000
(0b1111,1111 << (8-2)) == 0b1100,0000
(0b1111,1111 << (8-3)) == 0b1110,0000
(0b1111,1111 << (8-4)) == 0b1111,0000
```
Hence, only in Java,
```
((0b1000,1000 >>> a) ^ (0b1111,1111 << (8-a))) ==((0b1000,1000 >>(dafny) a)
```


RustanLeino:
Thanks, great bugs to be fixing! I have some remarks:

## LHS of right-shift in Java

* The `>>>` operator in Java fills the new high-order bits with `0`'s. However, Java's shift operators want an `int` or `long` as the left-hand argument. So, a `byte` or `short` as the LHS will first be converted--and sign-extended!--to be an `int`.
* So, the reported problem occurs only when the target type is `byte` or `short` _and_ the high-order bit is set. This can only happen for Dafny's types `bv8` and `bv16` (not `bv7`, for example, since the high-order bit of the target `byte` is never set for a Dafny `bv7`).
* I suggest implementing `unsignedShiftRightByte(byte a, byte amount)` as
    ```
    (((int)a) & 0xFF) >>> amount
    ```
* Note, `unsignedShiftRightByte` is better called `unsignedShiftRightBv8`, since it's only necessary for `bv8`, not for smaller bitvectors. Actually, if its name contains the nod `bv8` to Dafny, then the "`unsigned`" is not relevant, so it could be called just `bv8ShiftRight`.
* Note, the analogous thing also needs to be done for `short` (using `(((int)a) & 0xFFFF) >>> amount`). More precisely, this needs to be done for `bv16` (and is not necessary for `bv15` or any other bitvector type).
* To look for `bv8` and `bv16`, the method `doPossiblyNativeBinOp` should inspect `resultType`, not the string `nativeName`.

## RHS of left- and right- shift in Java and C#.

I noticed that there's another problem, too, with both Java and C#:

* These languages say that only the least significant 5 (for `int` and `uint`) or 6 (for `long` and `ulong`) bits of the RHS of the (left and right) shift operators matter.
* Dafny places a restriction on the RHS of the shift operators; namely, for `bv<n>`, the value `amount` of the RHS must be in the range `0 <= range <= n`. Note that `bv32` is represented as `int` (in Java) or `uint` (in C#), but these languages won't let you shift an `int`/`uint` by 32 bits. And similar for `bv64`.
* So, specifically for `bv32`, the Java code should be
    ```
    amount == 32 ? 0 : (a >>> amount)
    ```
    and the same for C#. (But for C#, it's better to use `>>` here. The `>>>` operator requires C# version 11, and since we're using a `uint` anyhow, `>>` and `>>>` would have the same effect, so we might as well choose the operator that's not dependent on a particular version of C#.)
* Similarly, for `bv64`, we need `amount == 64 ? 0 : (a >>> amount)`.

## Summary

In summary, to fix these issues with bit-shifting, we need to

* special-case `bv8` and `bv16` for right-shift in Java
* special-case `bv32` and `bv64` for left- and right-shift in Java and C#.

## References:

* Java: https://docs.oracle.com/javase/specs/jls/se8/jls8.pdf, section 15.19, page 568.
* C#: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators

## Possible test file

```

method Main() {
  Right();
  Left();
}

method Right() {
  var m: bv0 := 0;
  print m >> 0, "\n"; // 0
  var n: bv1 := 1;
  print n >> 0, " ", n >> 1, "\n"; // 1 0

  var a: bv7 := 12;
  print a >> 0, " ", a >> 1, " ", a >> 6, " ", a >> 7, " "; // 12 6 0 0
  a := -1;
  print a >> 0, " ", a >> 6, "\n"; // 127 1

  var b: bv8 := 12;
  print b >> 0, " ", b >> 1, " ", b >> 7, " ", b >> 8, " "; // 12 6 0 0
  b := -1;
  print b >> 0, " ", b >> 7, "\n"; // 255 1

  var r: bv15 := 12;
  print r >> 0, " ", r >> 1, " ", r >> 14, " ", r >> 15, " "; // 12 6 0 0
  r := -1;
  print r >> 0, " ", r >> 14, "\n"; // 32767 1

  var s: bv16 := 12;
  print s >> 0, " ", s >> 1, " ", s >> 15, " ", s >> 16, " "; // 12 6 0 0
  s := -1;
  print s >> 0, " ", s >> 15, "\n"; // 65535 1

  var w: bv31 := 12;
  print w >> 0, " ", w >> 1, " ", w >> 30, " ", w >> 31, " "; // 12 6 0 0
  w := -1;
  print w >> 0, " ", w >> 30, "\n"; // 0x7fff_ffff 1

  var x: bv32 := 12;
  print x >> 0, " ", x >> 1, " ", x >> 31, " ", x >> 32, " "; // 12 6 0 0
  x := -1;
  print x >> 0, " ", x >> 31, "\n"; // 0xffff_ffff 1

  var y: bv63 := 12;
  print y >> 0, " ", y >> 1, " ", y >> 62, " ", y >> 62, " "; // 12 6 0 0
  y := -1;
  print y >> 0, " ", y >> 62, "\n"; // 0x7fff_ffff_ffff_ffff 1

  var z: bv64 := 12;
  print z >> 0, " ", z >> 1, " ", z >> 63, " ", z >> 64, " "; // 12 6 0 0
  z := -1;
  print z >> 0, " ", z >> 63, "\n"; // 0xffff_ffff_ffff_ffff 1

  var u: bv100 := 12;
  print u >> 0, " ", u >> 1, " ", u >> 99, " ", u >> 100, " "; // 12 6 0 0
  u := -1;
  print u >> 0, " ", u >> 99, "\n"; // 0xf_ffff_ffff_ffff_ffff_ffff_ffff 1
}

method Left() {
  var m: bv0 := 0;
  print m << 0, "\n"; // 0
  var n: bv1 := 1;
  print n << 0, " ", n << 1, "\n"; // 1 0

  var a: bv7 := 12;
  print a << 0, " ", a << 1, " ", a << 6, " ", a << 7, " "; // 12 24 0 0
  a := -1;
  print a << 0, " ", a << 6, "\n"; // 127 64

  var b: bv8 := 12;
  print b << 0, " ", b << 1, " ", b << 7, " ", b << 8, " "; // 12 24 0 0
  b := -1;
  print b << 0, " ", b << 7, "\n"; // 255 128

  var r: bv15 := 12;
  print r << 0, " ", r << 1, " ", r << 14, " ", r << 15, " "; // 12 24 0 0
  r := -1;
  print r << 0, " ", r << 14, "\n"; // 32767 16384

  var s: bv16 := 12;
  print s << 0, " ", s << 1, " ", s << 15, " ", s << 16, " "; // 12 24 0 0
  s := -1;
  print s << 0, " ", s << 15, "\n"; // 65535 32768

  var w: bv31 := 12;
  print w << 0, " ", w << 1, " ", w << 30, " ", w << 31, " "; // 12 24 0 0
  w := -1;
  print w << 0, " ", w << 30, "\n"; // 0x7fff_ffff 0x4000_0000

  var x: bv32 := 12;
  print x << 0, " ", x << 1, " ", x << 31, " ", x << 32, " "; // 12 24 0 0
  x := -1;
  print x << 0, " ", x << 31, "\n"; // 0xffff_ffff 0x8000_0000

  var y: bv63 := 12;
  print y << 0, " ", y << 1, " ", y << 62, " ", y << 62, " "; // 12 24 0 0
  y := -1;
  print y << 0, " ", y << 62, "\n"; // 0x7fff_ffff_ffff_ffff 0x4000_0000_0000_0000

  var z: bv64 := 12;
  print z << 0, " ", z << 1, " ", z << 63, " ", z << 64, " "; // 12 24 0 0
  z := -1;
  print z << 0, " ", z << 63, "\n"; // 0xffff_ffff_ffff_ffff 0x8000_0000_0000_0000

  var u: bv100 := 12;
  print u << 0, " ", u << 1, " ", u << 99, " ", u << 100, " "; // 12 24 0 0
  u := -1;
  print u << 0, " ", u << 99, "\n"; // 0xf_ffff_ffff_ffff_ffff_ffff_ffff 0x8_0000_0000_0000_0000_0000_0000 
}
```

## Expect file for the test file

```
0
1 0
12 6 0 0 127 1
12 6 0 0 255 1
12 6 0 0 32767 1
12 6 0 0 65535 1
12 6 0 0 2147483647 1
12 6 0 0 4294967295 1
12 6 0 0 9223372036854775807 1
12 6 0 0 18446744073709551615 1
12 6 0 0 1267650600228229401496703205375 1
0
1 0
12 24 0 0 127 64
12 24 0 0 255 128
12 24 0 0 32767 16384
12 24 0 0 65535 32768
12 24 0 0 2147483647 1073741824
12 24 0 0 4294967295 2147483648
12 24 0 0 9223372036854775807 4611686018427387904
12 24 0 0 18446744073709551615 9223372036854775808
12 24 0 0 1267650600228229401496703205375 633825300114114700748351602688
```

MikaelMayer:
I implemented everything, thanks, what an impressive insight !

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
```diff
@@ -2881,9 +2881,27 @@ protected class ClassWriter : IClassWriter {
           }
 
         case BinaryExpr.ResolvedOpcode.LeftShift:
-          opString = "<<"; truncateResult = true; convertE1_to_int = true; break;
+          if (resultType.NormalizeExpand() is
+              BitvectorType { Width: var width and (32 or 64) }) {
+            staticCallString = $"{DafnyHelpersClass}.Bv{width}ShiftLeft";
+            convertE1_to_int = true;
+            truncateResult = true;
```
RustanLeino, line 2888: For 32- and 64-bit values, the result should not need to be truncated. However, the optimization to omit truncation for these (and also for 16- and 8-bit values) better belongs in `EmitBitvectorTruncation` (which need not be part of this PR).

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
```diff
@@ -2881,9 +2881,27 @@ protected class ClassWriter : IClassWriter {
           }
 
         case BinaryExpr.ResolvedOpcode.LeftShift:
-          opString = "<<"; truncateResult = true; convertE1_to_int = true; break;
+          if (resultType.NormalizeExpand() is
+              BitvectorType { Width: var width and (32 or 64) }) {
```
RustanLeino, line 2885: Is there a way to use `resultType.AsBitvectorType` and still use the `{ Width ... }` form? For example, is it legal to say

``` c#
if (resultType.AsBitvectorType is { Width: var width and (32 or 64) }) {
```

Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -2622,30 +2622,58 @@ protected class ClassWriter : IClassWriter {
...
-          callS = null;
+          if (o == ">>>" && resultType.NormalizeExpand() is
+                BitvectorType { Width: var width and (8 or 16 or 32 or 64) }) {
+            /// Solves https://github.com/dafny-lang/dafny/issues/3734
+            preOpS = width switch {
+              8 => "(byte)",
+              16 => "(short)",
+              32 => "(int)",
+              _ /*64*/ => "(long)"
+            };
```
RustanLeino, line 2636: Can you use `CastIfSmallNativeType` for this? (I expect the `(int)` and `(long)` casts are identity casts anyway.)

RustanLeino (dismissed):
I have some comments, but they are optional.

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
MikaelMayer: Yes, let me do that.

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
MikaelMayer: I'm not sure how to optimize away the truncation. Could you perhaps file an issue with what you have in mind?

Source/DafnyCore/Compilers/Java/Compiler-java.cs
MikaelMayer: Looks like it works. Implemented!

RustanLeino (approved):
Nice!

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
RustanLeino: I took a look at our generated code and apparently we already do the optimization that I suggested here. In particular, the Dafny code

``` dafny
  var x: bv7 := 63;  //    011 1111
  var y: bv8 := 63;  //   0011 1111
  var z: bv9 := 63;  // 0 0011 1111
  // a 5 left-shift of    011 1111 in a  uint8 gives           1110 0000, which after truncation gives            110 0000
  // a 5 left-shift of   0011 1111 in a  uint8 gives           1110 0000 -- no truncation needed
  // a 5 left-shift of 0 0011 1111 in a uint16 gives 0000 0111 1110 0000, which after truncation gives 0000 0001 1110 0000
  print x << 5, " ", y << 5, " ", z << 5, "\n"; // 96 224 480
```

I noticed that the three shift operations are already emitted by the C# and Java compilers as:

``` c#
  (byte)(    ((byte)(x << 5)) & (byte)0x7F     )
  (byte)(    ((byte)(y << 5))                  )
(ushort)(  ((ushort)(z << 5)) & (ushort)0x1FF  )
```

So, we're good.

PR #3739, opened 2023-03-14
Fix deep tests (Windows)

MikaelMayer:
The auditor was using code that replaced Unix newlines to create tests, so it did not work on Windows.
This PR fixes that.

Moreover, the release brew seemed to be broken for quite some time (tested 10 commits behind master, still the same issue) The smoke script, in go, prints '\n' literally instead of a newline. I wrapped it to make it a string so that it's printed normally.
I'll file this issue later.

### DIFF
--- a/.github/workflows/release-brew.yml
+++ b/.github/workflows/release-brew.yml
@@ -21,7 +21,7 @@ jobs:
         brew install dafny
     - name: Make test program
       run: |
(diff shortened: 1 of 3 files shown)


keyboardDrummer (dismissed):

davidcok (approved):
SHould be Aaron's to approve, but I'll trust that tests will fail if this is wrong.

atomb (approved):
Good catch!

PR #3743, opened 2023-03-14
Disable gutter icons when verification turned off

jtristan:
Disable gutter icons in the IDE when the verification is turned off


jtristan:
Not done right

MikaelMayer (approved):
Yes indeed, that's better so that we don't get gutter icon issues

jtristan:
Not useful anymore

PR #3754, opened 2023-03-17
Adding tests for previously uncovered parser code

davidcok:
And fixing some error positions.


Source/DafnyCore/Dafny.atg
```diff
@@ -188,6 +188,7 @@ public static Parser SetupParser(string/*!*/ s, Uri/*!*/ uri, ModuleDecl module,
   return new Parser(scanner, errors, module, builtIns);
 }
 
+// Unused code
```
alex-chew, line 191: Any reason to keep this unused code at all?

davidcok: Well, It could well be handy, for example if an IDE allowed user's to write debugging expressions. So I did not want to just delete it because it is currently not called anywhere.

PR #3769, opened 2023-03-21
Exposing --output and --spill-translation in dafny test, per Issue 3612

### LINKED ISSUE #3612: `dafny test` and `--output` option
### Summary

_(This is related to #3611)_

Unlike `dafny build`, the `dafny test` command does not support the `--output` option.  When run with the `--target java` option it just dumps the generated `java` files into the directory where the file being tested is contained (in my case `src/dafny` directory).  This is not an ideal place to store generate files, and it would be nice to have some control here.

### Background and Motivation

I'm trying to update by project build file to use the new CLI syntax (e.g. `dafny build`, etc).

### Proposed Feature

Adding `--output` as an option for `dafny test`.

### Alternatives

_No response_

davidcok:
Fixes #3612

### DIFF
--- a/Source/DafnyCore/Options/CommonOptionBag.cs
+++ b/Source/DafnyCore/Options/CommonOptionBag.cs
@@ -154,6 +154,13 @@ true - Use an updated type-inference engine. Warning: This mode is under constru
   };
   public static readonly Option<bool> VerifyIncludedFiles = new("--verify-included-files",
     "Verify code in included files.");
+  public static readonly Option<bool> UseBaseFileName = new("--use-basename-for-filename",
+    "When parsing use basename of file for tokens instead of the path supplied on the command line") {
+  };
+  public static readonly Option<bool> SpillTranslation = new("--spill-translation",
+    @"In case the Dafny source code is translated to another language, emit that translation.") {
+    IsHidden = true
+  };
   public static readonly Option<bool> WarningAsErrors = new("--warn-as-errors",
     "Treat warnings as errors.");
   public static readonly Option<bool> WarnMissingConstructorParenthesis = new("--warn-missing-constructor-parentheses",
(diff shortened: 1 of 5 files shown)


Source/DafnyCore/Resolver/RunAllTestsMainMethod.cs
```diff
@@ -98,6 +99,12 @@ public class RunAllTestsMainMethod : IRewriter {
     foreach (var moduleDefinition in program.CompileModules) {
       foreach (var callable in ModuleDefinition.AllCallables(moduleDefinition.TopLevelDecls)) {
         if ((callable is Method method) && Attributes.Contains(method.Attributes, "test")) {
+          var regex = program.Options.MethodsToTest;
+          if (regex != null) {
+            string name = method.FullDafnyName;
+            if (!Regex.Match(name, regex).Success) continue;
```
atomb, line 105: This should have braces, as the CI complains about.

Source/DafnyCore/Resolver/RunAllTestsMainMethod.cs
```diff
@@ -19,17 +25,12 @@ public class RunAllTestsMainMethod : IRewriter {
   /// errors much earlier in the pipeline.
   /// </summary>
   internal override void PreResolve(Program program) {
-    var hasMain = Compilers.SinglePassCompiler.HasMain(program, out var mainMethod);
-    if (hasMain) {
-      Reporter.Error(MessageSource.Rewriter, mainMethod.tok, "Cannot use /runAllTests on a program with a main method");
-      return;
-    }
```
atomb, line 26: The code at the end of this file still calls `hasMain` and an assertion fails when running with `/testContracts`. I wonder whether the implementation of `HasMain` need to be updated to account for the possibility of the new synthetic name you're using?

atomb (changes requested):
Overall I think this looks good, but there are two CI failures, one trivial and one I'm not entirely sure the source of.

Source/DafnyCore/Resolver/RunAllTestsMainMethod.cs
atomb: That was a satisfyingly simple fix. :)

atomb (approved):

PR #3779, opened 2023-03-21
Fix: Opaque functions guaranteed to be opaque until revealed

### LINKED ISSUE #3719: `opaque` function called within lambda is not opaque!
### Dafny version

4.0.0

```dafny
// Function want to verify something about
opaque function inc(x:nat) : nat { x + 1 }

// Intermediatary to force lambda
predicate Check(p: ()->bool) { p() }

method test() {
    assert Check(() => inc(1) == 2);
}
```

_No response_

### What happened?

The assertion verifies when it should not because there is no specification of `inc()` and, furthermore, `inc()` is marked `opaque`.  The problem persists when using the attribute syntax `{:opaque}` as well.  Finally, observe that this assertion does not verify as expected:

```dafny
assert inc(1) == 2;
```

Linux

### LINKED ISSUE #3778: Change the implementation of reveal/opaque to binary trigger, not fuel
## How opaque is currently implemented

To ensure recursive functions do not unroll indefinitely, Dafny uses a "fuel" attribute, which is set to 2 in the context of an assert, and 1 in the context of an assume. It ensures the axioms of the function do not unroll the function more than required.
For example:
```dafny
function inc(i: nat): nat {
  if i == 0 then 0 else inc(i-1) + 1
}

method Test(i: nat)
  requires i >= 3
{
  assert inc(i) == inc(i-3) + 3;
}
```
Dafny is unable to prove the assertion. However, if we replace the 3s by 2s, Dafny would succeed.

The amount of fuel that can be used is currently an annotation on the function `{:fuel <lowfuel>,<highfuel>}`.
Interestingly, if we set up this fuel to 0,0 or _unknown constants_, the function body is never available. This insight made it possible to figure out an opaque mechanism: Pretend the fuel is an unknown constant. To reveal it, assume this unknown constant is `1+another constant` for assumptions or `1+(1+another constant)` for assertions, which let the axioms unfold the function as usual.

For opaque functions, the attribute `{:fuel }` actually sets up the increase from `another constant`. It's just that these values are revealed only when passed to reveal().

## Problems with opaque = assumption of non-zero fuel

* In #3719, the problem is that opaque functions are provided with constant non-zero fuel that comes from a lambda expression. Looking at the code, it looks like it's by design, as a lambda expression starts a built-in function `AtLayer` which automatically provides a global fuel for the expressions below
* The notion of opaqueness is binary, either we reveal something or we don't, whereas here we need to maintain the value of constants
* It's not possible to set up the fuel for solving an assert at the time of the assertion.

## Proposed solution

Currently, fuel is encoded as unary numbers: `0 = $LZ` and `1 + fuel = $LS(<fuel>)` and uninterpreted variables for opaque functions.
I suggest that the opaqueness of functions be not related to the fuel, so that there is no interference. Whenever we have an opaque function call `P(<fuel>, x)`, we would thus have as first fuel argument the following expression:

```
P(CanRevealFuel(<contextFuel>, is_revealed_p), x)
```

`is_revealed_p` is a `bool` constant

The axiom is that
```
CanRevealFuel($LS(ly), true) == $LS(CanRevealFuel(ly, true))
```
meaning that without knowing is_revealed_p, it's never possible to extract fuel from a `CanRevealFuel()` expression.
And the `reveal()` call would do nothing else than assuming `is_revealed_p`

MikaelMayer:
This PR fixes #3719 and (implements) fixes #3778

In an upcoming PR, we might be able to remove the old mechanism "StartFuelAssert"/"StartFuel" of picking the fuel from the defniition site and letting user define their fuel at the call site, but this is independent.

This is the first PR of a series that will ensure opaqueness is consistent.


Test/vstte2012/BreadthFirstSearch.dfy
```diff
@@ -86,9 +86,10 @@ class BreadthFirstSearch<Vertex(==)>
     {
       // remove a vertex "v" from "C"
       var v :| v in C;
+      assert v != dest && C == {v} ==>
+        (Processed + {v}) + (N + (set w | w in Succ(v) && w !in V)) == R(source, d+1, AllVertices) by { reveal R(); }
       C, Processed := C - {v}, Processed + {v};
       ghost var pathToV := Find(source, v, paths);
-
```
atomb, line 90: Is there a reason to remove these blank lines?

atomb (approved):
Nice improvement! I suspect it could be valuable to recommend Dafny developers put `assert false` on any path they expect to be unreachable, as a general practice. It makes things clearer, and sometimes makes proofs faster. And, in the case where we warn about unreachable code, it would be easy to filter out those that explicitly included `assert false`.

Test/vstte2012/BreadthFirstSearch.dfy
MikaelMayer: No indeed that does not improve verifiability :-)

Test/vstte2012/BreadthFirstSearch.dfy
```diff
@@ -98,13 +99,16 @@ class BreadthFirstSearch<Vertex(==)>
           if length(p) < length(pathToV) {
             // show that this branch is impossible
             RMonotonicity(source, length(p), d-1, AllVertices);
+            assert false;
           }
         }
         return d, pathToV;
       }
-
       // process newly encountered successors
```
MikaelMayer, line 107:
```suggestion

      // process newly encountered successors
```

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -1063,13 +1064,28 @@ internal class PredefinedDecls {
...
+                  MaybeCreateRevealableConstant(f);
+                }
               }
             }
           }
         }
       }
     }
 
+    private void MaybeCreateRevealableConstant(Function f) {
```
RustanLeino, line 1078: The "`Maybe`" part of the name suggests to me that it may or may not be appropriate to have a revealable constant at all. But here it just indicates that the operation is idempotent--it's performed only once. For that purpose, I suggest still just naming the method `CreateRevealableConstant`.

Test/git-issues/git-issue-3719.dfy
```diff
@@ -0,0 +1,19 @@
+// RUN: %exits-with 4 %baredafny verify %args "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+// Function want to verify something about
+opaque function inc(x:nat) : nat { x + 1 }
+
+// Intermediatary to force lambda
+predicate CheckThis(p: ()->bool) { p() }
```
RustanLeino, line 8: Formatting:
* Use one space after `:`
* Use no space before `:`
* Use a space before and after `->`

Test/git-issues/git-issue-3719a.dfy
```diff
@@ -0,0 +1,11 @@
+// RUN: %baredafny verify %args "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+ghost opaque function R(i: int): int {
+  i
+}
+
+lemma Test() {
+  assert 1 == R(1) by {reveal R(); }
```
RustanLeino, line 9: Formatting: Add one space between `{` and `reveal`.

Source/DafnyCore/DafnyPrelude.bpl
```diff
@@ -505,6 +505,21 @@ function AtLayer<A>([LayerType]A, LayerType): A;
...
+function CanRevealFuel(LayerType, bool) : LayerType;
+
+axiom (forall $ly: LayerType :: 
+  { CanRevealFuel($LS($ly), true) }
+  CanRevealFuel($LS($ly), true) == $LS(CanRevealFuel($ly, true)));
+
+axiom (forall $ly: LayerType, b: bool ::
+  { CanRevealFuel(AsFuelBottom($ly), b) }
+  CanRevealFuel(AsFuelBottom($ly), b)
+  == $LZ);
```
RustanLeino, line 521: Well done in figuring out the problem and suggesting a solution. I don't think we ever equate anything with `$LZ` elsewhere. Because of that, my hunch is that it's better to write the axiom as

``` boogie
axiom (forall $ly: LayerType, b: bool :: 
  { CanRevealFuel(AsFuelBottom($ly), b) } 
  CanRevealFuel(AsFuelBottom($ly), b) == AsFuelBottom($ly));
```

In fact, even more, since the first axiom (LL 514-516) says that `CanRevealFuel` commutes with `$LS`, provided the second parameter is known to equal `true`, I think it seems appropriate to do the analogous thing for the second axiom. That is, I suggest writing the second axiom as

``` boogie
axiom (forall $ly: LayerType, b: bool :: 
  { CanRevealFuel(AsFuelBottom($ly), b) } 
  CanRevealFuel(AsFuelBottom($ly), b) == AsFuelBottom(CanRevealFuel($ly, b)));
```

RustanLeino (changes requested):

Test/git-issues/git-issue-3719.dfy
MikaelMayer: Right, I had just taken the provided test without reformatting.

RustanLeino (dismissed):

atomb (dismissed):

atomb (approved):

atomb (approved):
Nice work!

PR #3784, opened 2023-03-22
Cok suspend pdf build2

davidcok:
Fixes # nightly failure


fabiomadge:
Why do we bother running `publish-release`, if there is nothing to publish?

fabiomadge:
I suppose the nuget releases make it worthwhile, but still, we should run `package.py` with `--skip_manual`.

davidcok:
I think the publish release check is so that the process is tested. As for skip-manual I was not aware of that, and would have used it. Where is that option put?

fabiomadge:
The upload step updates the [nightly release](https://github.com/dafny-lang/dafny/releases/tag/nightly) with fresh binaries, which we don't build in the current revision. I'd prefer to avoid having a green nightly build without accompanying binaries. Maybe we can merge #3785 instead.

davidcok:
Closed in favor of #3785

PR #3787, opened 2023-03-22
Using quicktest in release-brew

davidcok:
Replaces inline tests on each platform with an execution of quicktests.sh, as is done in release-downloads and release-downloads-nuget


.github/workflows/release-brew.yml
```diff
@@ -19,43 +19,35 @@ jobs:
...
-    - name: Check C# compile
-      run: |
-        dafny /compileVerbose:0 /compile:3 /compileTarget:cs /spillTargetCode:3 a.dfy
-    - name: Check Go compile
-      run: |
-        dafny /compile:3 /spillTargetCode:3 /compileTarget:go a.dfy
-        env GO111MODULE=auto GOPATH=$PWD/a-go go run a-go/src/a.go > actual.txt
-        diff expect.txt actual.txt
-    - name: Check Javascript compile
+        python --version
```
atomb, line 30: Seems odd to check the Python version right before installing it (presumably installing a newer version).

atomb (approved):
This is much nicer! My one comment isn't blocking. Feel free to change or not as you like.

PR #3795, opened 2023-03-24
Introduce pre-types

RustanLeino:
Depends on https://github.com/dafny-lang/dafny/pull/3793

This PR introduces _pre-types_, which are used in the new type inference. This affects AST nodes, which now keep track of both their pre-type (used during type inference) and their program-provided or final type.

As a reminder, the motivation for changing the type inference at all is that we have several planned enrichments of the Dafny type system (e.g., allowing a `datatype` to implement a `trait`). The current type inference is not only complicated (see below), but it also hardwires various assumptions (for example, a `trait` is always a reference type and a `newtype` is always a numeric type). To evolve the type system, it seemed timely to reimplement type inference, even though this is a dauntingly large task.

## Pre-types

Two major complications in the early stages of the resolver are:
* Name resolution and type inference depend on each other. For example, to determine the type of an identifier expression `x`, one needs to know what the name `x` refers to, and to figure out what the name `y` in the expression `e.y` refers to, one needs to know the type of expression `e`.
* For the type system, sometimes subset types play or a role and sometimes they do not. For example, if a class `C` takes one type parameter, then `C<int>` is a different type than `C<nat>`. On the hand, given variables `i: int` and `n: nat`, the type system allows both the assignment `i := n;` and the assignment `n := i;` (for the latter, the verifier will check that the RHS value is indeed a `nat`).

To get through name resolution as quickly as possible, this PR introduces _pre-types_. A pre-type is a type, but ignoring any subset types. Pre-types are sufficient for name resolution, and thus the mutual dependency between name resolution and types is reduced to a mutual dependency between name resolution and pre-types.

## AST

To keep track of pre-types during name resolution, this PR adds pre-type information to AST nodes. For example, each `Expression` has not just a `.Type` field (which will eventually hold the full type inferred for the expression) but also a `.PreType` field (which is used during type inference). Inferred pre-types/types are stored not in those `.PreType`/`.Type` fields, but also in fields that keep track of type-parameter instantiations. For example, in addition to the existing `.TypeApplication_AtEnclosingClass` field, the PR adds a neighboring `.PreTypeApplication_AtEnclosingClass` field.

It is unfortunate and somewhat ugly that the AST now contains both of these fields. Perhaps this can be improved in the future. For now, it's a price to pay for the simpler type inference and lets us move forward into an expanded type system.

## Programmer-supplied types

While some types are inferred, others are provided directly in the program. If a type is provided, it gets incorporated into the type-inference process as given. To date, the Dafny implementation has used just one `.Type` field to hold both program-supplied types and inferred types. These two cases can be distinguished, but in two different ways.

When the Dafny syntax has an optional type (for example, for a local variable), the parser fills the `.Type` field with the given type, if any, or otherwise fills the `.Type` field with a _type proxy_ (typically `InferredTypeProxy`). When type inference starts to figure out what the type should be, it fills in the `.T` field of the proxy. By gradually filling in proxies, rather than replacing the proxy with the actual type, it is possible in the implementation to build up other types that use the proxy. For example, if a local variable `x` is initially given a proxy type `α`, then an expression `[x]` will have type `seq<α>`. Later, when the type of `x` is figured out, say to be `int`, then `α.T` is set to `int`, which auto-magically has the effect of also updating the type of `[x]` to be `seq<int>`.

When a type is always inferred (for example, the type of an expression `x + y`), then the `.Type` field starts off as `null`. During type inference, it is fill in with a non-`null` value (which often starts off as a proxy and is gradually filled in).

Because the `.Type` field can contain a program-supplied type, this PR inspects that field and creates a corresponding pre-type. To do this, it uses the method `Type2PreType`. When name resolution and pre-type inference are done, the resulting pre-type is converted to a type. To do this, it uses the method `PreType2Type`.

## Type synonyms

A type synonym in Dafny is just that: a synonym. Type inference is thus easier to do if type synonyms are expanded as soon as they are encountered. However, type-synonym names are meaningful to a programmer and are good to include in error messages. For example, if a programmer has declared a field to be of type `string`, then it's nicer if the error message for an incorrect assignment to the field mentions `string` rather than `seq<char>`.

For this reason, the `DPreType` class includes a `.PrintablePreType` field. It is used to store the original type synonym, if any.

## More uniform treatment of types

The `PreType` class hierarchy is simpler than the corresponding `Type` hierarchy. In particular, the `PreType` hierarchy stores types more uniformly. For example, whereas the `Type` hierarchy has special nodes for built-in types (e.g., `SeqType`), the `PreType` hierarchy stores those types as general type constructors (e.g., a `DPreType` whose `.Decl` field is special `TopLevelDecl` with the name `"seq"`). See method `BuiltInTypeDecl` in `PreTypeResolve.cs`.

In the future, it would be nice to simplify the `Type` hierarchy in the same way. But that's not part of this PR (nor is it part of my planned series of PRs for the new type inference).

## Special types

By and large, a `PreType` is either a `PreTypeProxy` or a `DPreType` (the "`D`" is for "a pre-type with a `.Decl`"). But there are a few others as well.

Name resolution tries to resolve each prefix of an input like `A.B.C.D`. But not all of these prefixes might be an expression. For example, perhaps `A` names an import, `B` names a nested module inside `A`, `C` is a type inside `B`, and `D` is a `static const` in `C`. To handle these prefix "expressions" in a somewhat uniform way, pre-type inference still fills in their `.PreType` field. The value of such a "pre-type" is a `PreTypePlaceholder`, either a `PreTypePlaceholderModule` or a `PreTypePlaceholderType`.

In another special case, an input like `x := M(y);` might be an ordinary assignment with an expression RHS (here, `M` would then be referring to a function) or it might be a method call (if `M` denotes a method). In the process of figuring this out, pre-type inference will assign the `.PreType` field of the RHS even if `M` is a method. In that case, the "pre-type" assigned will be `UnusedPreType`.

The legacy type inference in Dafny also had special cases for what the previous two paragraphs described. In addition, the legacy type inference used a `SelfType` in the signature of the built-in bitvector methods `RotateLeft` and `RotateRight`. This allowed the legacy type inference to resolve these names even before the specific bitvector width had been inferred. That complication did not seem worth keeping for pre-types, so this PR instead provides some methods for creating the rotation methods for specific bitvector types. (These will be used in future PRs.) So, there is no equivalent of the `SelfType` in the new type inference.

## Reviewing this PR

This PR builds on https://github.com/dafny-lang/dafny/pull/3793. It makes sense to review and merge #3793 before reviewing this one.


Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -700,6 +703,47 @@ public abstract class Expression : TokenNode {
...
+    Contract.Requires(function.TypeArgs.Count == 0);
+
+    var call = new FunctionCallExpr(tok, function.Name, receiver, tok, tok, new List<Expression>()) {
+      Function = function,
+      Type = Type.Bool,
+      TypeApplication_AtEnclosingClass = receiver.Type.TypeArgs,
+      TypeApplication_JustFunction = new List<Type>()
+    };
+
+    // Wrap the resolved call in the usual unresolved structure, in case the expression in cloned and re-resolved.
```
jtristan, line 722:
```suggestion
    // Wrap the resolved call in the usual unresolved structure, in case the expression is cloned and re-resolved.
```

Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -700,6 +703,47 @@ public abstract class Expression : TokenNode {
     return ite;
   }
 
+  /// <summary>
+  /// Create a resolved function-call expression. The returned expression will have syntactic scaffolding, which
+  /// enables resolving a syntactic clone of this resolved expression.
+  /// Expects "receiver" to be a resolved expression.
+  /// Expects that "function" has no type parameters, but it's okay for the enclosing type to have type parameters.
```
jtristan, line 710: What if it does have type parameters?

Source/DafnyCore/AST/Expressions/Expressions.cs
jtristan, line 722:
```suggestion
    // Wrap the resolved call in the usual unresolved structure, in case the expression is cloned and re-resolved.
```

Source/DafnyCore/AST/Expressions/Expressions.cs
jtristan, line 710: What if it does have a type parameter?

Source/DafnyCore/Resolver/PreType/PreType.cs
```diff
@@ -0,0 +1,326 @@
...
+//-----------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using Bpl = Microsoft.Boogie;
+
+namespace Microsoft.Dafny {
+  public abstract class PreType {
```
jtristan, line 15:
```suggestion
  
  public abstract class PreType {
```

Source/DafnyCore/Resolver/PreType/PreType.cs
```diff
@@ -0,0 +1,326 @@
...
+    public PreType Normalize() {
```
jtristan, line 16:
```suggestion

    public PreType Normalize() {
```

Source/DafnyCore/Resolver/PreType/PreType.cs
```diff
@@ -0,0 +1,326 @@
...
+// SPDX-License-Identifier: MIT
+//
+//-----------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using Bpl = Microsoft.Boogie;
+
```
jtristan, line 13: Some big picture comments explaining what pre-types are about and what the classes and methods do would be welcome

Source/DafnyCore/Resolver/PreType/PreType.cs
```diff
@@ -0,0 +1,326 @@
...
+          if (!Same(ap.Arguments[i], bp.Arguments[i])) {
+            return false;
+          }
+        }
+        return true;
+      }
+      return a == b;
+    }
+
+    public PreType UrAncestor(PreTypeResolver ptResolver) {
```
jtristan, line 39: Example of where comments would be helpful. I assume we following definitions to find typing constraints, and if so it would be good to say so and explain what assumptions we are making to avoid going in cycles

Source/DafnyCore/Resolver/PreType/PreTypeToType.cs
```diff
@@ -0,0 +1,174 @@
+//-----------------------------------------------------------------------------
```
jtristan, line 1: As for the rest of the code in PreType, some high-level comments would be welcome

Source/DafnyCore/Resolver/PreType/PreTypeToType.cs
```diff
@@ -0,0 +1,174 @@
...
+        return new SetType(false, PreType2Type(pt.Arguments[0]));
+      case "multiset":
+        return new MultiSetType(PreType2Type(pt.Arguments[0]));
+      case "seq":
+        return new SeqType(PreType2Type(pt.Arguments[0]));
+      case "map":
+        return new MapType(true, PreType2Type(pt.Arguments[0]), PreType2Type(pt.Arguments[1]));
+      case "imap":
+        return new MapType(false, PreType2Type(pt.Arguments[0]), PreType2Type(pt.Arguments[1]));
+      default:
```
jtristan, line 51: no bv? string has already been desugared into seq<char>? nat treated as int?

Source/DafnyCore/Resolver/PreType/PreTypeToType.cs
```diff
@@ -0,0 +1,174 @@
...
+      Contract.Assert(datatypeValue.InferredTypeArgs.Count == 0 || datatypeValue.InferredTypeArgs.Count == datatypeValue.InferredPreTypeArgs.Count);
+      if (datatypeValue.InferredTypeArgs.Count == 0) {
+        foreach (var preTypeArgument in datatypeValue.InferredPreTypeArgs) {
+          datatypeValue.InferredTypeArgs.Add(PreType2Type(preTypeArgument));
+        }
+      }
+    }
+
+    if (expr.PreType is UnusedPreType) {
+      expr.Type = new InferredTypeProxy();
```
jtristan, line 115: Perhaps my biggest concern is related to this: is the benefit of not having to annotate a constant with its type worth the cost? Is it even beneficial? We require the users to annotate their functions with types in part because it makes the code more readable, why not constants as well?

jtristan (approved):
Is allowing the programmer to not provide a type annotation for a constant declaration (or subset type or newtype) actually a feature we should support? It makes the inference more complicated, which reflects that the code is more complicated to read and understand.

Source/DafnyCore/AST/Expressions/MemberSelectExpr.cs
```diff
@@ -11,6 +11,18 @@ public class MemberSelectExpr : Expression, IHasUsages, ICloneable<MemberSelectE
   [FilledInDuringResolution] public Label /*?*/ AtLabel;  // non-null for a two-state selection
   [FilledInDuringResolution] public bool InCompiledContext;
 
+  /// <summary>
+  /// TypeApplication_AtEnclosingClass is the list of type arguments used to instantiate the type that
```
stefan-aws, line 15:
```suggestion
  /// PreTypeApplication_AtEnclosingClass is the list of type arguments used to instantiate the type that
```

Source/DafnyCore/AST/Expressions/MemberSelectExpr.cs
```diff
@@ -11,6 +11,18 @@ public class MemberSelectExpr : Expression, IHasUsages, ICloneable<MemberSelectE
...
+  /// declares Member (which is some supertype of the receiver type).
+  /// </summary>
+  [FilledInDuringResolution] public List<PreType> PreTypeApplication_AtEnclosingClass;
+
+  /// <summary>
+  ///  TypeApplication_JustMember is the list of type arguments used to instantiate the type parameters
```
stefan-aws, line 21:
```suggestion
  ///  PreTypeApplication_JustMember is the list of type arguments used to instantiate the type parameters
```

Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -1212,6 +1257,7 @@ public Resolver_IdentifierExpr(IToken tok, TopLevelDecl decl, List<Type> typeArg
     Decl = decl;
     TypeArgs = typeArgs;
     Type = decl is ModuleDecl ? (Type)new ResolverType_Module() : new ResolverType_Type();
+    PreType = decl is ModuleDecl ? new PreTypePlaceholderModule() : new PreTypePlaceholderType();
```
stefan-aws, line 1260: Why are we using type-casting `(Type)` in the definition of `Type`, but no equivalent in the definition of `PreType`?

Source/DafnyCore/AST/Function.cs
```diff
@@ -84,6 +84,7 @@ public TailStatus
   public readonly List<TypeParameter> TypeArgs;
   public readonly List<Formal> Formals;
   public readonly Formal Result;
+  public PreType ResultPreType;
```
stefan-aws, line 87: Why is `ResultType` readonly, but `ResultPreType` isn't?

Source/DafnyCore/AST/MemberDecls.cs
```diff
@@ -97,6 +97,7 @@ public class Field : MemberDecl, ICanFormat, IHasDocstring {
   public override string WhatKind => "field";
   public readonly bool IsMutable;  // says whether or not the field can ever change values
   public readonly bool IsUserMutable;  // says whether or not code is allowed to assign to the field (IsUserMutable implies IsMutable)
+  public PreType PreType;
```
stefan-aws, line 100: Why is `Type` readonly, but `PreType` isn't?

Source/DafnyCore/Resolver/PreType/PreType.cs
stefan-aws: It seems to me that for the existing `Type`, the function `Normalize` is defined within a class `Type` in `Types.cs`. Why are taking a different approach for `PreType`?

Source/DafnyCore/Resolver/PreType/PreType.cs
stefan-aws, line 15: Add a `ToString` function?

Source/DafnyCore/Resolver/PreType/PreType.cs
stefan-aws: A comment with a description of `Normalize` would be good to have. For `PreType`, why is there no need for e.g. `NormalizeExpand`?

Source/DafnyCore/Resolver/PreType/PreType.cs
```diff
@@ -0,0 +1,326 @@
...
+          var basePreType = ptResolver.Type2PreType(newtypeDecl.BaseType);
+          pt = basePreType.Substitute(subst);
+        } else {
+          // nothing more we can do
+          return pt;
+        }
+      }
+    }
+
+    public DPreType AsCollectionType() {
```
stefan-aws, line 56: Given the naming, I expected the return type to be `CollectionType`, not `DPreType`.

Source/DafnyCore/Resolver/PreType/PreType.cs
```diff
@@ -0,0 +1,326 @@
...
+        t = proxy.PT;
+      }
+      return t;
+    }
+
+    public static bool Same(PreType a, PreType b) {
+      a = a.Normalize();
+      b = b.Normalize();
+      if (a is DPreType ap && b is DPreType bp && ap.Decl == bp.Decl) {
+        Contract.Assert(ap.Arguments.Count == bp.Arguments.Count);
```
stefan-aws, line 28: This inner part could be moved into a separate `Same` function in `DPreType`.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
```diff
@@ -68,8 +65,210 @@ public abstract class ResolverPass {
...
+        }
+      }
+
+      type = expandedType;
+      var decl = Type2PreTypeDecl(type);
+      var arguments = type.TypeArgs.ConvertAll(ty => Type2PreType(ty, null, Type2PreTypeOption.GoodForInference));
+      return new DPreType(decl, arguments, printablePreType);
+    }
+
+    TopLevelDecl Type2PreTypeDecl(Type type) {
```
stefan-aws, line 188: Given the return type, could perhaps be renamed more generally to  `Type2Decl`. Will there be a `Type2DPreType` function, too?

Source/DafnyCore/Resolver/PreType/PreTypeToType.cs
stefan-aws: +1

stefan-aws (approved):

keyboardDrummer:
Could we come up with a more descriptive name than pre-types? complete-types, fullset-types, unconstrainted-types ? I think the essence of these types is that they don't rely in any way on the verifier: for any expression we can always say whether it's of one of these types or not. That's where the name "complete-types" comes from.

Source/DafnyCore/Resolver/PreType/PreType.cs
keyboardDrummer, line 39: What does `Ur` mean?

Source/DafnyCore/Resolver/PreType/PreType.cs
```diff
@@ -0,0 +1,326 @@
...
+        } else {
+          // nothing more we can do
+          return pt;
+        }
+      }
+    }
+
+    public DPreType AsCollectionType() {
+      if (Normalize() is DPreType dp) {
+        switch (dp.Decl.Name) {
```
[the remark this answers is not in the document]
RustanLeino: No, `string` is a type synonym for `seq<char>`. But `seq` is included in this list, so that means strings will be handled by that case.

Source/DafnyCore/AST/Expressions/Expressions.cs
RustanLeino: Then, either this method will need to be expanded to handle that case, too, or we'd need a different method for it. For now, the only use of this `CreateResolvedCall` method is by `CreateResolvedValidCall` (which is part of the `:autocontract` machinery), and `Valid()` does not have any type parameters.

Source/DafnyCore/Resolver/PreType/PreType.cs
RustanLeino: I added (the suggested blank line and) a description of the method.

For `Type`, there are three operations:
* `Normalize()` follows type proxies to get to the type for which it is a proxy.
* `NormalizeExpandKeepConstraints()` additionally follows any type synonym to the type it stands for.
* `NormalizeExpand()` additionally follows a subset type to its base type.

For `PreType`, the `Normalize()` method is analogous. Pre-types don't keep track of type synonyms or subset types; any such types given in the input program are `Type.NormalizeExpand()`ed before as part of turning it into a pre-type. Therefore, there is no analogous `PreType` form of `NormalizeExpandKeepConstraints()` and `NormalizeExpand()`.

I added this to the comment above `PreType.Normalize()`.

Source/DafnyCore/Resolver/PreType/PreType.cs
RustanLeino: I added a comment in the code. In essence, pre-types are the part of types that is mutually dependent on name resolution. This means a pre-type expands type synonyms and subset types. (See also the comments at the top of this PR.)

Source/DafnyCore/Resolver/PreType/PreType.cs
RustanLeino: I added a comment.

Regarding cycles, this is a good observation. The method as it stands assumes there are no cycles. I added a comment about this. Some of the cycle detection did not make it into this PR, but are part of new upcoming PRs.

Source/DafnyCore/Resolver/PreType/PreTypeToType.cs
RustanLeino: Since pre-types don't include type synonyms (like `string`) or subset types (like `nat`), pre-type inference only comes up with types like `seq<char>` and `int`. Those are handled in this method.

Awkwardly, the `Type` hierarchy has special nodes for several different types. For example, `seq` is represented by a `SeqType` node. (This would be good to change in the future.) The pre-type hierarchy is simpler, treating all non-proxy pre-types as (what the literature would call) type constructors.

Bitvectors are represented as `ValuetypeDecl`s. To create such a type from a pre-type, the `ValuetypeDecl.CreateType` method is called.

Source/DafnyCore/Resolver/PreType/PreTypeToType.cs
RustanLeino: I hear your concern. Indeed, maybe the language should be changed to insist that every `const` (with or without a RHS) is given an explicit type, and that the base type of `newtype` and subset types are given explicitly. In my next PR, I handle the inference of these things, so I suggest we delay the decision of whether or not to change the language.

Source/DafnyCore/AST/Expressions/Expressions.cs
RustanLeino: Astute observation. It is because neither `ResolverType_Module` nor `ResolverType_Type` is a subclass of the other, whereas `PreTypePlaceholderModule` is a subclass of `PreTypePlaceholderModule`. The rules in C# for inferring the type of a `_ ? _ : _` expression is unable to figure out that both alternatives are a subtype of `Type` on L259.

(Dafny's rules for this are different. In a future PR, you will see that Dafny is willing to infer a pre-type that is common super-type of the two `if-then-else` alternatives. Dafny keeps going up the supertype chain until such a choice is unique. We'll discuss this choice later.)

Source/DafnyCore/AST/Function.cs
RustanLeino: `Result` is determined by the parser, whereas `ResultPreType` happens at the beginning of pre-type resolution.

Source/DafnyCore/AST/MemberDecls.cs
RustanLeino: Same as for `ResultPreType`, see above.

Source/DafnyCore/Resolver/PreType/PreType.cs
RustanLeino: Each `PreType` subclass already adds a `ToString` method.

Source/DafnyCore/Resolver/PreType/PreType.cs
RustanLeino: Good point. I'll at least rename it to `AsCollectionPreType`.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
RustanLeino: I like it. I'll rename it to `Type2Decl`.

Source/DafnyCore/Resolver/PreType/PreType.cs
RustanLeino: It's such a small piece of code that I'd prefer to keep it here. But I did replace the loop with one `... Zip ... All ...` construction.

RustanLeino:
@keyboardDrummer I think of pre-types as being the part of types that are mutually dependent on name resolution. But I like your "don't rely in any way on the verifier" description (which is also accurate). Let's consider a renaming, but I suggest we do so outside of this PR (and probably waiting until after the next few PRs in this same series come through, too). That way, we'll have the whole picture in front of us and there may be several things we'd like to rename at that time.

Source/DafnyCore/Resolver/PreType/PreType.cs
RustanLeino: From Oxford Dictionary:

prefix: ur-
primitive, original, or earliest.
Example: "urtext"

The dictionary says the prefix has a German origin. I think the Dutch prefix may be "ur-" or "oer-".

In short, I was trying to say "earliest ancestor" or "most primitive base type".

stefan-aws: As a native German speaker, I can confirm.

stefan-aws (approved):

PR #3799, opened 2023-03-27
Update description of --solver-plugin

keyboardDrummer:
Updated description reads:

```
  --solver-plugin <solver-plugin>              Dafny uses Boogie as part of its verification process. This 
                                               option allows customising that part using a Boogie plugin. 
                                               More information about Boogie can be found at 
                                               https://github.com/boogie-org/boogie. Information on how to 
                                               construct Boogie plugins can be found by looking at the code 
                                               in 
                                               https://github.com/boogie-org/boogie/blob/58ab09236f41894ee9a7
                                               93d30eb17617b0fce797/Source/Provers/SMTLib/ProverUtil.cs
```

### DIFF
--- a/Source/DafnyCore/Options/CommonOptionBag.cs
+++ b/Source/DafnyCore/Options/CommonOptionBag.cs
@@ -61,7 +61,8 @@ The `text` format also includes a more detailed breakdown of what assertions app
   };
 
   public static readonly Option<uint> SolverResourceLimit = new("--resource-limit", @"Specify the maximum resource limit (rlimit) value to pass to Z3. Multiplied by 1000 before sending to Z3.");
-  public static readonly Option<string> SolverPlugin = new("--solver-plugin", @"Specify a plugin to use to solve verification conditions (instead of an external Z3 process).");
+  public static readonly Option<string> SolverPlugin = new("--solver-plugin",
+    @"Dafny uses Boogie as part of its verification process. This option allows customising that part using a Boogie plugin. More information about Boogie can be found at https://github.com/boogie-org/boogie. Information on how to construct Boogie plugins can be found by looking at the code in https://github.com/boogie-org/boogie/blob/58ab09236f41894ee9a793d30eb17617b0fce797/Source/Provers/SMTLib/ProverUtil.cs");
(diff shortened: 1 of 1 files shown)


Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -61,7 +61,8 @@ public class CommonOptionBag {
   };
 
   public static readonly Option<uint> SolverResourceLimit = new("--resource-limit", @"Specify the maximum resource limit (rlimit) value to pass to Z3. Multiplied by 1000 before sending to Z3.");
-  public static readonly Option<string> SolverPlugin = new("--solver-plugin", @"Specify a plugin to use to solve verification conditions (instead of an external Z3 process).");
+  public static readonly Option<string> SolverPlugin = new("--solver-plugin",
+    @"Dafny uses Boogie as part of its verification process. This option allows customising that part using a Boogie plugin. More information about Boogie can be found at https://github.com/boogie-org/boogie. Information on how to construct Boogie plugins can be found by looking at the code in https://github.com/boogie-org/boogie/blob/58ab09236f41894ee9a793d30eb17617b0fce797/Source/Provers/SMTLib/ProverUtil.cs");
```
atomb, line 65: It seems to me that that link could get out of date quickly. Could you change it to a link that'd resolve to the latest version, instead?

atomb (changes requested):

Source/DafnyCore/Options/CommonOptionBag.cs
keyboardDrummer: An 'up-to-date' link will break when the file gets moved. I think a potentially out-of-date but unbreakable link is better. Neither of these two options is perfect. Perfect would be a stable link to documentation but Boogie doesn't have documentation on this topic.

atomb: Would it be possible to link to a file as it exists for a certain tag? Then we could keep things pointing at the tag of the most recent release.

keyboardDrummer: I'm not sure what you mean. Either you point to a fixed point in time, using a commit SGA or tag, and then you're sure the path doesn't break, if you point to a dynamic point in time, and then the file path may break. Right?

atomb (approved):
Love it!

PR #3802, opened 2023-03-27
Fix: Formatter consistent between command-line and VSCode

### LINKED ISSUE #3790: Dafny formatter produces incorrect indentation with match statement and :|
### Dafny version

4.0.0

```dafny
lemma Try(i: int)
{
  match i {
    case i' =>
      var x :| x == i';
      assert x == i';
  }
}
```

```code
dafny format file.dfy
```

### What happened?

When running dafny format, one gets the following result:

```dafny
lemma Try(i: int)
{
  match i {
    case i' =>
      var x :| x == i';
  assert x == i';
  }
}
```

Instead it shouldn't change the indentation.

When adding curly braces around the case statement body, then the indentation is correct, and correctly produces the following:

```dafny
lemma Try(i: int)
{
  match i {
    case i' => {
      var x :| x == i';
      assert x == i';
    }
  }
}
```

Mac

MikaelMayer:
This PR fixes #3790

Besides that bug, it looks like there was discrepancies between the formatter on the command-line and the formatter in VSCode. So I took the chance to fix it all at once because that bug was concerned as well.
I also modified the tests so that they also test the formatting post-resolution. It enabled me to find 13 failing tests that I all fixed as well.

### DIFF
--- a/Source/DafnyCore/AST/Expressions/Expressions.cs
+++ b/Source/DafnyCore/AST/Expressions/Expressions.cs
@@ -783,7 +783,7 @@ public abstract class Expression : TokenNode {
   /// </summary>
   public static Expression CreateIdentExpr(IVariable v) {
     Contract.Requires(v != null);
-    var e = new IdentifierExpr(v.RangeToken.StartToken, v.Name);
+    var e = new IdentifierExpr(v.Tok, v.Name);
     e.Var = v;  // resolve here
     e.type = v.Type;  // resolve here
     return e;
--- a/Source/DafnyCore/AST/Expressions/Expressions.cs
+++ b/Source/DafnyCore/AST/Expressions/Expressions.cs
@@ -3508,7 +3508,7 @@ public abstract class SuffixExpr : ConcreteSyntaxExpression {
   }
 }
 
-public class NameSegment : ConcreteSyntaxExpression, ICloneable<NameSegment> {
+public class NameSegment : ConcreteSyntaxExpression, ICloneable<NameSegment>, ICanFormat {
   public readonly string Name;
   public readonly List<Type> OptTypeArguments;
   public NameSegment(IToken tok, string name, List<Type> optTypeArguments)
@@ -3530,6 +3530,13 @@ public class NameSegment : ConcreteSyntaxExpression, ICloneable<NameSegment> {
   }
 
   public override IEnumerable<Node> PreResolveChildren => OptTypeArguments ?? new List<Type>();
+  public bool SetIndent(int indentBefore, TokenNewIndentCollector formatter) {
+    formatter.SetTypeLikeIndentation(indentBefore, OwnedTokens);
+    foreach (var subType in PreResolveChildren.OfType<Type>()) {
+      formatter.SetTypeIndentation(subType);
+    }
+    return false;
+  }
 }
 
 /// <summary>
--- a/Source/DafnyCore/AST/Grammar/TokenNewIndentCollector.cs
+++ b/Source/DafnyCore/AST/Grammar/TokenNewIndentCollector.cs
@@ -412,12 +412,14 @@ public class TokenNewIndentCollector : TopDownVisitor<int> {
       }
 
       var initialMemberIndent = declWithMembers.tok.line == 0 ? indent : indent2;
-      foreach (var member in declWithMembers.Members) {
-        if (member.tok is IncludeToken) {
+      foreach (var member in declWithMembers.PreResolveChildren) {
+        if (member.Tok is IncludeToken) {
           continue;
         }
 
-        SetMemberIndentation(member, initialMemberIndent);
+        if (member is MemberDecl memberDecl) {
+          SetMemberIndentation(memberDecl, initialMemberIndent);
+        }
       }
     } else if (topLevelDecl is SubsetTypeDecl subsetTypeDecl) {
       SetRedirectingTypeDeclDeclIndentation(indent, subsetTypeDecl);
@@ -660,7 +662,7 @@ public class TokenNewIndentCollector : TopDownVisitor<int> {
                   commaIndent = indent;
                 } else {
                   rightIndent = indent + SpaceTab;
-                  commaIndent = indent = SpaceTab;
+                  commaIndent = indent + SpaceTab;
                 }
 
                 SetIndentations(assignmentOperator, afterStartIndent, opIndentDefault, rightIndent);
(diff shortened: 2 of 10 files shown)


Source/DafnyCore/AST/Expressions/Expressions.cs
```diff
@@ -3530,6 +3530,13 @@ public NameSegment(IToken tok, string name, List<Type> optTypeArguments)
   }
 
   public override IEnumerable<Node> PreResolveChildren => OptTypeArguments ?? new List<Type>();
+  public bool SetIndent(int indentBefore, TokenNewIndentCollector formatter) {
+    formatter.SetTypeLikeIndentation(indentBefore, OwnedTokens);
+    foreach (var subType in PreResolveChildren.OfType<Type>()) {
+      formatter.SetTypeIndentation(subType);
+    }
+    return false;
+  }
```
MikaelMayer, line 3539: Without this line, the tests containing
```
  ConstructorName(
    a: LongName < map <
           b,
           c>>
```
Were incorrectly formatted post-resolution, because the user defined type becomes a NameSegment expression after resolution

Source/DafnyCore/AST/Grammar/TokenNewIndentCollector.cs
```diff
@@ -352,10 +342,27 @@ private class Indentations {
           }
       }
     }
+  }
+
+  public void SetTypeIndentation(Type type) {
+    var tokens = type.OwnedTokens.ToList();
```
MikaelMayer, line 348: Setting up type indentation was also not recursive post-resolution, so I had to change that.

Source/DafnyCore/AST/Grammar/TokenNewIndentCollector.cs
```diff
@@ -412,12 +419,14 @@ private class Indentations {
       }
 
       var initialMemberIndent = declWithMembers.tok.line == 0 ? indent : indent2;
-      foreach (var member in declWithMembers.Members) {
-        if (member.tok is IncludeToken) {
+      foreach (var member in declWithMembers.PreResolveChildren) {
```
MikaelMayer, line 422: Without this change, special fields such as "_reads" for iterators, which are being assigned the same token as the iterator itself, would cause the keyword "iterator" to be indented twice.

Source/DafnyCore/AST/Grammar/TokenNewIndentCollector.cs
```diff
@@ -660,7 +669,7 @@ private class Indentations {
                   commaIndent = indent;
                 } else {
                   rightIndent = indent + SpaceTab;
-                  commaIndent = indent = SpaceTab;
+                  commaIndent = indent + SpaceTab;
```
MikaelMayer, line 672: That's the bug I fixed ^^ but it was not sufficient.

Source/DafnyCore/AST/Grammar/TokenNewIndentCollector.cs
```diff
@@ -692,6 +701,10 @@ private class Indentations {
       }
     }
 
+    if (stmt is AssignSuchThatStmt assignSuchThatStmt) {
+      Visit(assignSuchThatStmt.Expr, rightIndent);
```
MikaelMayer, line 705: We were missing the visit of the right-hand-side of an assign such that statement. Now it's being tested

Source/DafnyCore/AST/TopLevelDeclarations.cs
```diff
@@ -1104,11 +1105,23 @@ public class ModuleDefinition : RangeNode, IDeclarationOrUsage, IAttributeBearin
...
   public IToken NameToken => tok;
-  public override IEnumerable<Node> Children => (Attributes != null ? new List<Node> { Attributes } : Enumerable.Empty<Node>()).Concat<Node>(TopLevelDecls).
+  public override IEnumerable<Node> Children => (Attributes != null ?
+      new List<Node> { Attributes } :
+      Enumerable.Empty<Node>()).Concat<Node>(TopLevelDecls).
     Concat(RefinementQId == null ? Enumerable.Empty<Node>() : new Node[] { RefinementQId });
 
-  public override IEnumerable<Node> PreResolveChildren => Includes.Concat<Node>(TopLevelDecls).Concat(
-    PrefixNamedModules.Select(tuple => tuple.Item2));
+  private IEnumerable<Node> preResolveTopLevelDecls;
```
MikaelMayer, line 1113: Resolving a modules implies to transfer all the PrefixNamedModules and putting them in nested modules under TopLevelDecls. However, the formatter deals with structural declarations only, not semantics, so it was messing up with the indentation of modules like `module Outer.A {}` because it would then believe that "module" belongs to "module A {}" which is _inside_ `module Outer` so would be wrongly indented by two spaces.

Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
```diff
@@ -336,7 +336,9 @@ public partial class Resolver {
       if (expr is ParensExpression) {
         var e = (ParensExpression)expr;
         ResolveExpression(e.E, resolutionContext);
-        e.ResolvedExpression = e.E;
+        var innerRange = e.E.RangeToken;
+        e.ResolvedExpression = e.E; // Overwrites the range, which is not suitable for ParensExpressions
```
MikaelMayer, line 340: I also found a post-resolution bug. Normally, when ConcreteExpressions are resolved, their replacement takes the range token of the original expression.
However, when writing `(E)`, during resolution, it would transform the range of `E` to actually contain the parentheses, which would mess up token ownership and prevent formatting from happening correctly post-resolution.

Source/DafnyCore/AST/Grammar/TokenNewIndentCollector.cs
atomb: Ooh, that's a tricky one.

Source/DafnyCore/AST/Statements/AssignSuchThatStmt.cs
```diff
@@ -8,6 +9,10 @@ public class AssignSuchThatStmt : ConcreteUpdateStatement, ICloneable<AssignSuch
   public readonly Expression Expr;
   public readonly AttributedToken AssumeToken;
 
+  public override IEnumerable<Node> PreResolveChildren =>
+  Lhss.Concat<Node>(new List<Node>() {
+  Expr });
```
atomb, line 14: The formatting of this declaration is a bit weird. Maybe indent line 13 and join line 14 with it?

Source/DafnyCore/AST/TopLevelDeclarations.cs
atomb: Another tricky one!

docs/dev/news/3790.fix
```diff
@@ -0,0 +1 @@
+Formatter consistent between command-line and VSCode
```
atomb, line 1: Maybe make this a little more detailed? How about:

```suggestion
The formatter now produces the same output whether invoked on the command-line or from VSCode
```

atomb (approved):
Nice debugging! I had a couple of minor formatting and wording suggestions, but I think it otherwise looks good.

PR #3808, opened 2023-03-28
Experiment with resolving matches without introducing let bindings

keyboardDrummer:


RustanLeino:
Just as a +1 for this draft PR. I too would be interested in not introducing any let bindings until after name resolution is done, because that will simplify the new type inference I'm working on.

Issue #3806, opened 2023-03-27
Assertions/reveal localized in expressions


MikaelMayer:
### Summary

This feature requests consists in changing the semantics of `StmtExpr` so that revealed assumptions are always local to the expression next to it and is not revealed in unexpected places.

### Background and Motivation

Proof durability is one of the major concerns of some of our high-level Dafny customers. Basically, it all boils down to ensure that the verifier is not provided with too many facts that could blow up verification time.
One very practical way to reduce the number of facts known to the verifier is to use opaque functions labelled assertions associated with reveal statements. Unfortunately, this is system is not very robust. Some efforts has been made but problems still remain (#2260, #3719, #3506, #2941, #3805 ...). One of the most emblematic ones is #1382, that indicates that revealing functions inside functions extends past the function definition itself.

We need better mechanisms to make things opaque and not blow up

### Proposed Feature

A `reveal` StmtExpr should only reveal its argument to its own Expr, and to verify a caller's assumption if at the root of an argument., but should not extend past this.
So for example, 

```dafny
opaque function F(x: int): nat {
  1
}

function AllArgumentsNotZero(x: int, y: int): int
  requires x != 0
  requires y != 0

function testCallF(): int {
    var r := AllArgumentsNotZero(reveal F(); F(0), F(1)); // Used to verify, should not verify anymore
    assert F(0) != 0; // Always verified
    assert F(1) != 0; // Always verified
    assert F(0) == 1; // Used to verify, should not verify anymore
    assert F(1) == 1; // Used to verify, should not verify anymore
    r
}
```

This should be the same for `assume` and `assert` StmtExpr.

### Alternatives

Not having this feature prevents some Dafny customers to scale up their designs and their proofs.
Workaround include more refactoring, which is not always practical or desirable, or increasing the number of cores.

Other solutions we considered included:
* A new construct (e.g. `by { proof }`) to prove things locally. However, compared to reusing an existing syntax and giving its a definite meaning, this is no longer the panacea.
* Locally opaque expressions. Still in the works, can be implemented another time

MikaelMayer:
@atomb  @alex-chew @fabiomadge  and myself discussed this today. Here is the fruit of our discussion.

# Decision

In a first phase, we will solve this issue by implementing the following decisions:
* We will use an attribute (alternative C) for the flexibility it provides in this experimentation phase.
* We will use `{:scoped}` so that it can have suitable antonym and is memorable.

### Notes about decision

[...]

PR #3838, opened 2023-04-04
Feat: Labelled assertions and requires in functions

### LINKED ISSUE #3804: Support for labelled assertions and requires in functions
### Summary

There are labelled assertions in statements, but not in expressions. Similarly, it's not possible to reveal "labelled requires" in functions.

### Background and Motivation

We often see a conversion from method to functions to ease reasoning.
However, labelling assertion is not possible in expressions. In the following example:

```dafny
predicate P(x: int)

function test(): (i: int)
  ensures P(i)
  requires rule: forall i | i % 2 == 0 :: P(i)
{
  assert rule2: forall i | i % 4 == 0 :: P(i) by {
    reveal rule;
  }
  assert P(12) by {
    reveal rule;
  }
  12
}
```

`assert rule2: ` is not permitted, and if removed, `reveal rule;` is not permitted either.
It ought to be available.

### Proposed Feature

* Labelled assertions in expressions
* Support for revealing labelled requires in functions

### Alternatives

Not having labelled assertions is not a problem for small codebases, but bigger codebases start to have this issue.
The workaround for not having labelled assertions is to have non-labelled assertions, but as functions grow, verification can become not durable. Not only that, but to figure out what assertions actually speed up proof, one absolutely needs to isolate assertions so that they don't interfere with other mechanisms.
Another workaround is to convert a function to a method, but it's not always possible, especially if the function was called by other functions.

MikaelMayer:
This PR fixes #3804 by implementing labelled assertions and labelled requires in functions.
Previously, labelled requires in functions were admissible by the parser but would always be available and could not be revealed. Also, assertions in expressions could not have labels.

Precisely, this PR adds and tests the following:
* Parsing support for _labelled assertions_
* Dominating labels are now pushed and popped when resolving the body of a function.
* Resolution support for _labelled requires_ (nothing needed to be done for assertions)
* Removed assumption _labelled requires_ in the well-formedness check of a function (this is the only relevant check, since functions don't have a separate implementation)
* Assumptions inside expressions cannot be revealed outside of these expressions. Doing so would require to store the _condition path_ and is strictly not necessary, one could simply lift those assertions for now (see change in documentation).
* I changed the error message for revealed constants, assertions and requires to be more specific than `"undefined identifier: reveal_p"`. The error message is now `Cannot reveal 'p' because no constant, assert label or requires label in the current scope is named 'p'`
* I added five integration tests

## Out of scope

* I considered enabling reveal of labelled assertions in inner expressions for outer expressions, but that will require to store the path condition if possible, so I'll keep this feature for later. It does not exist for statements either.
* Making a labelled assertion uniquely visible to preconditions of a function, e.g.
```
function call(arg0: int)
  requires LargePrecond(arg0)

assert precond: LargePrecond(arg0) by {...}; 
var x := (reveal precond; call(arg0));
// precond is still also revealed here, sorry !
...
```
This PR does not provide a way to reveal the precondition only for the function call. We'll follow-up on that later, e.g. by fixing #3806

## Impact on verification
The impact on verification is almost inexistent since it's a new feature. The generate Boogie code is identical for every program, except one special scenario that is extremely unlikely to happen, which yet previously verified and no longer verifies.
The example is the following:
```dafny
function DidVerify(i: int): nat
  requires precondition: i > 0
{
  i
}
```
We recorded the label in the requires of the function, but it was never a label that could be revealed (writing `reveal precondition;` would not even have passed resolution, so the requires was never opaque and the function was verified.
Now verification fails because we correctly not assume the precondition until it is revealed, so the fix is to reveal it:
```dafny
function DidVerify(i: int): nat
  requires precondition: i > 0
{
  reveal precondition;
  i
}
```
The reasons why this scenario is extremely unlikely to happen is the following:
* Labelled requires were until now the same as requires in functions, so there was no point using labelled requires for functions.
* It was not possible to reveal labelled requires anyway (which is always the effort that someone would do if they label a requires)
* Labelled requires that are not revealed, if they were working, would have very likely been removed as they don't participate in the verification condition, since they are opaque. The only exception to that is well-formedness check, but for functions, these are not separate verification phases but only a single one.

Still, it's worth noting, in case someone accidentally converted a method to a function and this method had a labelled requires that was never revealed, and that after doing the conversion the labelled requires started to be useful.
The fix is trivial though.


Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
```diff
@@ -5992,7 +5997,11 @@ public class ResolveTypeReturn {
       } else {
         // ----- None of the above
         if (complain) {
-          reporter.Error(MessageSource.Resolver, expr.tok, "unresolved identifier: {0}", name);
+          if (resolutionContext.InReveal) {
+            reporter.Error(MessageSource.Resolver, expr.tok, "cannot reveal '{0}' because no constant, assert label or requires label in the current scope is named '{0}'", expr.Name);
```
keyboardDrummer, line 6001: Is a label reference not an identifier? Previous message seemed alright to me.

docs/DafnyRef/UserGuide.md
```diff
@@ -932,6 +932,43 @@ method Slow(i: int, j: int)
 }
 ```
 
+Labelled assert statements are also available both in expressions and statements.
```
keyboardDrummer, line 935: Can you leave `also` out?

docs/DafnyRef/UserGuide.md
```diff
@@ -932,6 +932,43 @@ method Slow(i: int, j: int)
...
+Assertion labels are not accessible outside of scope of the assert statement.
```
keyboardDrummer, line 936: "the scope"

Might be better to say "the scope which the assert statement is in"

keyboardDrummer (approved):

Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
```diff
@@ -3268,8 +3271,10 @@ public XConstraint_EquatableArg(IToken tok, Type a, Type b, bool allowSuperSub,
           var ec = reporter.Count(ErrorLevel.Error);
           allTypeParameters.PushMarker();
           ResolveTypeParameters(function.TypeArgs, false, function);
+          DominatingStatementLabels.PushMarker();
```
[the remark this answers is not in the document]
MikaelMayer: I moved these push and pop inside function.Resolve.
Yes both are needed because "requires labels" need to be accessible within the by method body if there is one.

keyboardDrummer (dismissed):

Source/DafnyCore/Dafny.atg
```diff
@@ -3650,7 +3650,7 @@ AssertStmt<out Statement/*!*/ s, bool inExprContext>
   .)
   "assert"                                     (. x = t; startToken = t; .)
   { Attribute<ref attrs> }
-  ( [ IF(IsLabel(!inExprContext))
+  ( [ IF(IsLabel(true))
```
RustanLeino, line 3653: The `inExprContext` parameter of production `AssertStmt` is no longer needed and can be removed.

Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
```diff
@@ -5989,7 +5993,11 @@ public class ResolveTypeReturn {
       } else {
         // ----- None of the above
         if (complain) {
-          reporter.Error(MessageSource.Resolver, expr.tok, "unresolved identifier: {0}", name);
+          if (resolutionContext.InReveal) {
+            reporter.Error(MessageSource.Resolver, expr.tok, "cannot reveal '{0}' because no constant, assert label or requires label in the current scope is named '{0}'", expr.Name);
```
RustanLeino, line 5997:
```suggestion
            reporter.Error(MessageSource.Resolver, expr.tok, "cannot reveal '{0}' because no constant, assert label, or requires label in the current scope is named '{0}'", expr.Name);
```

Test/dafny0/LabeledAssertsResolution.dfy.expect
```diff
@@ -7,9 +7,9 @@ LabeledAssertsResolution.dfy(26,26): Error: assert label shadows a dominating la
 LabeledAssertsResolution.dfy(28,13): Error: assert label shadows a dominating label
 LabeledAssertsResolution.dfy(34,24): Error: assert label shadows a dominating label
 LabeledAssertsResolution.dfy(73,12): Error: break label is undefined or not in scope: ABC
-LabeledAssertsResolution.dfy(81,17): Error: unresolved identifier: reveal_C
+LabeledAssertsResolution.dfy(81,17): Error: cannot reveal 'C' because no constant, assert label or requires label in the current scope is named 'C'
```
RustanLeino, line 10: Much better! Thank you.

Test/git-issues/git-issue-3804.dfy
```diff
@@ -0,0 +1,64 @@
...
+  assert P(12) by {
+    reveal rule2;
+  }
+  12
+}
+
+function BothAssertNotProved(): (i: int)
+  ensures P(i)
+{
+  assert rule2: P(12) by {
```
RustanLeino, line 20: I expect this to fail as well. That is, the verifier should try to prove the asserted condition, whether or not there is a label. But if there is a label, then the asserted condition should be available downstream only if a `reveal` is used.

Test/git-issues/git-issue-3804.dfy
```diff
@@ -0,0 +1,64 @@
+// RUN: %exits-with 4 %baredafny verify %args "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+predicate P(x: int)
+
+function FirstAssertNotProved(): (i: int)
+  ensures P(i)
+{
+  assert rule2: P(12) by {
```
RustanLeino, line 9: This assertion should generate a verification error (see my comment on L20 below).

Test/git-issues/git-issue-3804.dfy
```diff
@@ -0,0 +1,64 @@
...
+  assert P(12) by { // Fails
+  }
+  12
+}
+
+function FirstAssertNotProvedBecauseRequiresNotRevealed(): (i: int)
+  requires rule: P(12)
+  ensures P(i)
+{
+  assert rule2: P(12) by {
```
RustanLeino, line 31: Again, the verifier should complain about this line.

Test/git-issues/git-issue-3804.dfy
```diff
@@ -0,0 +1,64 @@
...
+
+
+function SecondAssertNotProvedBecauseAssertNotRevealed(): (i: int)
+  requires rule: P(12)
+  ensures P(i)
+{
+  assert rule2: P(12) by {
+    reveal rule;
+  }
+  assert P(12) by {
```
RustanLeino, line 48: The verifier should complain about this line.

Test/git-issues/git-issue-3804.dfy.expect
```diff
@@ -0,0 +1,7 @@
+git-issue-3804.dfy(9,17): Error: assertion might not hold
+git-issue-3804.dfy(20,17): Error: assertion might not hold
+git-issue-3804.dfy(22,10): Error: assertion might not hold
+git-issue-3804.dfy(31,17): Error: assertion might not hold
+git-issue-3804.dfy(48,10): Error: assertion might not hold
```
RustanLeino, line 5: Oh, I see. The verifier _does_ complain about those lines. Good. Please add comments (like `// error: cannot be verified here`) for those lines, so that these intended failures are evident from just reading the test file.

RustanLeino, line 5: Oh, I see. The verifier _does_ complain about those lines. Good. Please add comments (like `// error: cannot be verified here`) for those lines, so that these intended failures are evident from just reading the test file.

Test/git-issues/git-issue-3804b.dfy
```diff
@@ -0,0 +1,42 @@
...
+
+function FunctionAndMethodVerify(i: int): (r: int)
+  requires p: i >= 0
+  ensures r >= 0
+{
+  reveal p;
+  i
+} by method {
+  reveal p;
+  r := i;
```
RustanLeino, line 12: Note, this `by method` clause should also verify in the absence of L11. The reason is that the `by method` is verified to return the exact same value as the function body, and the function's `ensures` clauses are checked only of the function body.

Test/git-issues/git-issue-3804b.dfy.expect
```diff
@@ -0,0 +1,10 @@
+git-issue-3804b.dfy(15,9): Error: a postcondition could not be proved on this return path
```
RustanLeino, line 1: Add comments to the test file to indicate that these errors are expected.

Test/git-issues/git-issue-3804b.dfy.expect
```diff
@@ -0,0 +1,10 @@
...
+git-issue-3804b.dfy(17,12): Related location: this is the postcondition that could not be proved
+git-issue-3804b.dfy(31,12): Error: a postcondition could not be proved on this return path
+git-issue-3804b.dfy(31,5): Related location: this is the postcondition that could not be proved
```
RustanLeino, line 4: I expect that this is _not_ an error. The reason is that the `by method` part should only need to establish that it returns the same value as the function body. It is the function body's responsibility to ensure that the function's `ensures` clauses follow from the function body.

Test/git-issues/git-issue-3804b.dfy.expect
```diff
@@ -0,0 +1,10 @@
...
+git-issue-3804b.dfy(35,9): Error: a postcondition could not be proved on this return path
+git-issue-3804b.dfy(37,12): Related location: this is the postcondition that could not be proved
+git-issue-3804b.dfy(40,12): Error: a postcondition could not be proved on this return path
+git-issue-3804b.dfy(40,5): Related location: this is the postcondition that could not be proved
```
RustanLeino, line 8: Same here. I expect this to verify.

Test/git-issues/git-issue-3804c.dfy
```diff
@@ -0,0 +1,26 @@
...
+{
+  assert p1: x ==> P(1) by { reveal r; }
+  assert p2: !x ==> P(2) by { reveal r; }
+  var x :=
+    match x {
+      case true =>
+        assert p11: P(1) by { reveal p1; }
+        assert p12: P(1) by { reveal p2; } // Previous assertion is not visible here
+        assert P(1) by { reveal p11; } // p11 can be revealed and this assertion is visible outside.
+        1
```
RustanLeino, line 17: Nice tests!

Test/git-issues/git-issue-3804c.dfy
```diff
@@ -0,0 +1,26 @@
...
+      case false => 
+        assert p22: P(2) by { reveal p2; }
+        assert p21: P(2) by { reveal p1; } // Previous assertion is not visible here
+        assert P(2) by { reveal p21; } // p21 can be revealed and this assertion is visible outside.
```
RustanLeino, line 21: What if L21 were changed to just `reveal p21;`? Would that still allow L24 to prove?

docs/DafnyRef/UserGuide.md
```diff
@@ -932,6 +932,43 @@ method Slow(i: int, j: int)
...
+```dafny
+ghost predicate P(i: int)
+
+method TestMethod(x: bool)
+  requires r: x <==> P(1)
+{
+  assert a: x ==> P(1) by { reveal r; } // Lifted assertion
+  if x {
+    assert a': P(1) by { reveal a; } // Proved without revealing the precondition
+  }
```
RustanLeino, line 965: It may be more natural to write these lines as

``` dafny
assert a: x ==> P(1) by {
  if x {
    assert P(1) by { reveal r; }
  }
}
```

RustanLeino (changes requested):
In addition to my comments below, please add some similar tests for an extreme predicate (e.g., `least predicate`) to make sure that the cloning of the extreme predicate into a prefix predicate works as expected (both for the specification and the body). (Btw, I wonder if this is tested for extreme lemmas today.)

Test/git-issues/git-issue-3804.dfy
MikaelMayer: It does, so I just added a comment on the line of the test to be consistent.

Test/git-issues/git-issue-3804b.dfy.expect
MikaelMayer: > I expect that this is not an error. The reason is that the by method part should only need to establish that it returns the same value as the function body. It is the function body's responsibility to ensure that the function's ensures clauses follow from the function body.

As explained above, it's an error because we cannot prove here the precondition of calling the function since the precondition is opaque.

Test/git-issues/git-issue-3804c.dfy
MikaelMayer: >What if L21 were changed to just reveal p21;? Would that still allow L24 to prove?

Yes it does as expected.

Test/git-issues/git-issue-3804b.dfy
RustanLeino: Hmm. Evidently, precondition labels are ignored during the well-formedness checking of specifications. Therefore, the use of `F` in the postcondition of method `M` above passes the well-formedness check, and so do the well-formedness checks of the specification in

``` dafny
method M'(i: int) returns (r: int)
  requires p: i >= 0
  requires 2 < 10 / if i >= 0 then 1 else 0
  ensures r == F(i)
```

It would seem consistent to say that the use of the function body in a `by method` still needs to prove the precondition of that function. But since that's the very same precondition that we're copying into the method that stands for the `by method` part, it seems that any reveals of the function's precondition ought to come for free. 

Would it be possible in our implementation (of the well-formedness check of `by method`) to `assume CanCall$F(i)` before checking the `r == F(i)` postcondition?

MikaelMayer: Looks like it should work. I'll implement it.

MikaelMayer: There is something I don't understand.

I changed the code a bit so that the by method declaration contains a conditional return
```dafny
function FunctionVerifiesButNotMethod(i: int): (r: int)
  requires p: i >= 0
  ensures r >= 0
{
  reveal p;
  i
} by method { // Error: Could not verify + related
  if(i == 5) {
    return  ; // Ensure we check every branch.
  }
  r := i;
}
```

Now, things start to get weird. In the generated boogie code, we have (I removed irrelevant pieces of code), I get an error on the last return path, but not the inner one.
```
procedure {:verboseName "FunctionVerifiesButNotMethod (correctness)"} Impl$$_module.__default.FunctionVerifiesButNotMethod(i#0: int) returns (defass#r#0: bool, r#0: int, $_reverifyPost: bool);
  // user-defined postconditions
  free ensures _module.__default.FunctionVerifiesButNotMethod#canCall(i#0);
  ensures r#0 == _module.__default.FunctionVerifiesButNotMethod(i#0);

implementation {:verboseName "FunctionVerifiesButNotMethod (correctness)"} Impl$$_module.__default.FunctionVerifiesButNotMethod(i#0: int) returns (defass#r#0: bool, r#0: int, $_reverifyPost: bool)
{
    $_reverifyPost := false;
    if (i#0 == LitInt(5))
    {
        r#0 := i#0;
        defass#r#0 := true;
        assert defass#r#0;
        return;
    }
    r#0 := i#0;
    defass#r#0 := true;
    assert defass#r#0;
} // Error here, cannot prove the precondition r#0 == _module.__default.FunctionVerifiesButNotMethod(i#0);
```
However, if I inline the free ensures as an assume and the ensures as an assert, Boogie is happy.
So it looks like there is actually a problem on the Boogie side.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -3417,6 +3417,35 @@ public enum StmtType { NONE, ASSERT, ASSUME };
...
+        List<Bpl.Expr> args = arguments.Select(arg => etran.TrExpr(arg)).ToList();
+        var formals = MkTyParamBinders(GetTypeParams(method), out var tyargs);
+        if (options.AlwaysUseHeap || method.FunctionFromWhichThisIsByMethodDecl.ReadsHeap) {
+          Contract.Assert(etran.HeapExpr != null);
+          tyargs.Add(etran.HeapExpr);
+        }
+
+        if (!method.IsStatic) {
+          Bpl.Expr th;
+          var thVar = BplBoundVar("this", TrReceiverType(method.FunctionFromWhichThisIsByMethodDecl), out th);
```
RustanLeino, line 3438:
```suggestion
          var thVar = BplBoundVar("this", TrReceiverType(method.FunctionFromWhichThisIsByMethodDecl), out var th);
```

Test/git-issues/git-issue-3804b.dfy
```diff
@@ -1,42 +1,61 @@
...
+  reveal p;
   i
-} by method { // Error: Could not verify + related
-  r := i;
+} by method {
+  assert problem: i >= 0; // Error: Cannot prove
+  assert cando: i >= 0 by {
+    reveal p;
+  }
+  if(i < -2) {
```
RustanLeino, line 44: No need for parentheses. Same on L52.

RustanLeino (approved):
Looks good. Thanks!

PR #3841, opened 2023-04-05
[Counterexamples] Support constant fields

Dargones:
Counterexamples should now correctly report values assigned to constant fields (where previously these values would not be filled in). As a bonus, this PR removes the need for special processing of arrays because array lengths are constant fields under the hood.

This PR adds the following test case for counterexample generation:

```Dafny
class Value {
  const v:int;
}
method a(v:Value) {
  assert v.v != 42;
}
```

With the expected output being: `v:_module.Value? = (v := 42)`

### DIFF
--- a/Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
+++ b/Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
@@ -34,7 +34,6 @@ namespace DafnyServer.CounterexampleGeneration {
       fSetEmpty, fSeqEmpty, fSeqBuild, fSeqAppend, fSeqDrop, fSeqTake,
       fSeqUpdate, fSeqCreate, fReal, fU2Real, fBool, fU2Bool, fInt, fU2Int,
       fMapDomain, fMapElements, fMapBuild, fIs, fIsBox;
-    private readonly Dictionary<Model.Element, Model.Element[]> arrayLengths = new();
     private readonly Dictionary<Model.Element, Model.FuncTuple> datatypeValues = new();
 
     // maps a numeric type (int, real, bv4, etc.) to the set of integer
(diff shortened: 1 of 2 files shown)


Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -124,23 +123,8 @@ public class DafnyModel {
...
-            var len = tpl.Result;
-            if (!arrayLengths.TryGetValue(elt, out var ar)) {
-              ar = new Model.Element[dims];
-              arrayLengths.Add(elt, ar);
-            }
-            Contract.Assert(ar[idx] == null);
-            ar[idx] = len;
-          }
-        } else if (fn.Name.StartsWith("#") && fn.Name.IndexOf('.') != -1 && fn.Name[1] != '#') {
+        if (fn.Name.StartsWith("#") && fn.Name.IndexOf('.') != -1 && fn.Name[1] != '#') {
```
keyboardDrummer, line 127: Could you put the magic string into named constants, so it's more clear where they come from?

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -879,12 +859,12 @@ public class DafnyModel {
...
 
     /// <summary>
-    /// Return all functions that map the datatype object to a particular
-    /// destructor value.
+    /// Return all functions that map an object to a destructor value.
     /// </summary>
     private List<Model.Func> GetDestructorFunctions(Model.Element datatypeElement) {
       var types = GetIsResults(datatypeElement).Select(isResult =>
         new DafnyModelTypeUtils.DatatypeType(ReconstructType(isResult) as UserDefinedType ?? UnknownType).Name);
+      types = types.Select(type => type.Length > 0 && type.Last() == '?' ? type[..^1] : type);
```
keyboardDrummer, line 867: Is it possible to chop off the `?` at the type level instead of at the string level? Something like `Type.NonNullable`

Could you put the "?" into a const variable?

Could you replace the reassignment with a fresh assignment?

For example using,

```
var possiblyNullableTypes = ..; 
var types = possiblyNullableTypes.Select(type => type.Length > 0 && type.Last() == '?' ? type[..^1] : type)
```

keyboardDrummer (dismissed):

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Done!

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
Dargones: Done - changed the code to use `DafnyModelTypeUtils.GetNonNullable` (I don't think there is API to perform this operation in the Type class itself, at least not on the unresolved level)

atomb (approved):

keyboardDrummer (approved):

Dargones:
Thank you, @atomb and @keyboardDrummer! I have fixed a small issue in my code that caused the tests to fail, so everything should be good to go.

Dargones:
@atomb , @keyboardDrummer  - could you take a brief look a this one? I made a [small change](https://github.com/dafny-lang/dafny/pull/3841/commits/380286e8db3039693755cc04be96e6687936c5a7) since you last approved the PR, which fixed a bug causing a test to fail but everything else should be the same.

keyboardDrummer (approved):

PR #3842, opened 2023-04-05
Give the --boogie option an arity of * instead of 0-1

### LINKED ISSUE #3836: Multiple Occurences of `--boogie`
### Dafny version

4.0

```dafny
function id(x:int) : (r:int) { x }
```

```code
dafny verify --boogie "/randomSeedIterations:0" --boogie "/vcsLoad:2" tmp.dfy
```

### What happened?
When multiple occurrences of the cli switch `--boogie` are given on the command-line, all but the last are ignored.  We can see this because the above `dafny` command returns successfully.  However, it should report an error on the malformed input given in `--boogie "/randomSeedIterations:0"`.  Specifically, this command line does fail as expected:

```
dafny verify --boogie "/randomSeedIterations:0" tmp.dfy

dafny: Error: Invalid argument "0" to option /randomSeedIterations
Use /help for available options
```

This behaviour is pretty confusing, and had me stumped for quite a while. 

Linux

keyboardDrummer:
Fixes #3836


atomb (approved):
This seems like the least surprising behavior to me. Thanks!

PR #3843, opened 2023-04-05
Translate else blocks with only havoc assignments to valid Python

keyboardDrummer:
Translate else blocks with only havoc assignments to valid Python


Test/ghost/elseWithHavocStatement.dfy
```diff
@@ -0,0 +1,13 @@
+// RUN: %baredafny run %args --target=py %s > %t
```
fabiomadge, line 1: Isn't that subsumed by the following line?

Source/DafnyCore/Compilers/Python/Compiler-python.cs
```diff
@@ -1574,10 +1574,13 @@ private class ClassWriter : IClassWriter {
     protected override void TrStmtList(List<Statement> stmts, ConcreteSyntaxTree writer) {
       Contract.Requires(cce.NonNullElements(stmts));
       Contract.Requires(writer != null);
-      if (stmts.All(s => s.IsGhost)) {
+      var listWriter = new ConcreteSyntaxTree();
+      base.TrStmtList(stmts, listWriter);
+      if (listWriter.Descendants.OfType<LineSegment>().Any()) {
```
fabiomadge, line 1579: Maybe we could add something like `IsEmpty`.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -2961,7 +2961,7 @@ public class CheckHasNoAssumes_Visitor : BottomUpVisitor {
       }
     }
 
-    protected internal void TrStmt(Statement stmt, ConcreteSyntaxTree wr, ConcreteSyntaxTree wStmts = null) {
+    protected void TrStmt(Statement stmt, ConcreteSyntaxTree wr, ConcreteSyntaxTree wStmts = null) {
```
fabiomadge, line 2964: Do we need that?

Test/ghost/elseWithHavocStatement.dfy
```diff
@@ -0,0 +1,13 @@
...
+method Main() {
+    var b: bool :| true; 
+    var r: int;
+    if b {
+      r := 13;
+    } else {
+      r := *;
+    }
+    print r;
+}
```
fabiomadge, line 13: `Test/comp/Various.dfy` might be a good fit. The whole problem is that this isn't ghost code, right?

fabiomadge (commented):
Thanks for the fix! The previous test was janky.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
keyboardDrummer: `protected internal` exposes more than `protected`, and `internal` wasn't needed here. It's suggested by Rider.

Source/DafnyCore/Compilers/Python/Compiler-python.cs
keyboardDrummer: This implementation isn't perfect so I'd rather not hide it behind something that looks better.

Test/ghost/elseWithHavocStatement.dfy
keyboardDrummer: Why would we put it in another file? `Various` doesn't tell me much. I would rather put it in a file related to `If` statements.

Source/DafnyCore/Compilers/Python/Compiler-python.cs
fabiomadge: If it's not, we could just amend it existing check.

Test/ghost/elseWithHavocStatement.dfy
fabiomadge: I prefer fewer files with multiple tests, because they make browsing the directory easier. I think you should at least put the file somewhere else. The `ghost` directory seems wrong.

Source/DafnyCore/Compilers/Python/Compiler-python.cs
fabiomadge: Quite frankly I'm tempted to just always print the `pass`.

keyboardDrummer: Since this is slightly better and we already wrote the code, shall we just leave it?

Test/ghost/elseWithHavocStatement.dfy
keyboardDrummer: Moved it to a sensible directory

fabiomadge: I don't think the tests are closely related. The others test the verifier, or the resolver, but not the compilers, for which we have a dedicated folder.

keyboardDrummer: What are you suggesting we do then?

I have placed the new test in `comp/` and left the existing tests unchanged.

fabiomadge (approved):

PR #3851, opened 2023-04-06
Add project files with includes/excludes and options support

### LINKED ISSUE #2907: Enable configuring Dafny options through a project file
### Problems
- It's not possible to open a Dafny project in the IDE and automatically let the IDE use the correct Dafny options for this project.
- When calling the Dafny CLI on a Dafny project, the correct options for this project always have to be specified, which means most Dafny project needs to use a wrapping build tool like `Make` or a Bash script to enable easily invoking the Dafny CLI.

### Prerequisites
- https://github.com/dafny-lang/dafny/issues/2906

### Out of scope
- Enable specifying the Dafny version to use in the project file

### Solution
Enable specifying options to the Dafny CLI in a Dafny project file, which can be passed as an argument when calling the Dafny CLI, also when using the `dafny server` command.

keyboardDrummer:
Fixes #2907

#### From the updated docs:

Commands on the Dafny CLI that can be passed a Dafny file, can also be passed a Dafny project file. Such a project file may define which Dafny files the project contains, and which Dafny options it uses. The project file must be a [TOML](https://toml.io/en/) file named `dafny.toml`. Here's an example of a Dafny project file:

```toml
includes = ["src/**/*.dfy"]
excludes = ["**/ignore.dfy"]
[options]
  enforce-determinism = true
  warn-shadowing = true
```
Options are applied to a command if they can be. Invalid options are ignored.

When using a Dafny IDE based on the `dafny server` command, the IDE will search for project files by traversing up the file tree, and options from the first found project file will override options passed to `dafny server`.

### Caveats

- The IDE will only use the saved-on-disk versions of project files, so unsaved changes are ignored.

### Testing

- Add CLI (through Lit) and `dafny server` tests (through XUnit)


docs/DafnyRef/UserGuide.md
```diff
@@ -571,6 +571,23 @@ Most output from `dafny` is directed to the standard output of the shell invokin
 - Dafny `expect` statements (when they fail) send a message to **standard-out**.
 - Dafny I/O libraries send output explicitly to either **standard-out or standard-error**
 
+### 13.5.5. Project files
+
+Commands on the Dafny CLI that can be passed a Dafny file, can also be passed a Dafny project file. Such a project file may define which Dafny files the project contains, and which Dafny options it uses. The project file must be a [TOML](https://toml.io/en/) file named `dafny.toml`. Here's an example of a Dafny project file:
```
robin-aws, line 576: It feels a bit weird to be passing the project file directly as just another file, rather than with a flag like `--project`. It also means you can't have multiple project files for a codebase in the same directory, which is a less common scenario but one I have seen in other systems frequently enough.

It also seems valuable to search for a project file automatically just as `dafny server` will. Future improvement or something you're intentionally omitting?

docs/dev/news/2907.feat
```diff
@@ -0,0 +1,3 @@
+Added support for Dafny project files. For now the project file only allows specifying which Dafny files to include and exclude, and what options to use.
+The CLI commands that take Dafny files as input, such as build, run, translate, will also accept Dafny project files.
+When using an IDE based on `dafny server`, such as the Dafny VSCode extension, the IDE will look for Dafny project files by traversing up the file tree from the currently opened file. The first found project file will override options specified in the IDE.
```
robin-aws, line 3: Will this be immediately true or will the extension need an update first?

Test/cli/projectFile/projectFile.dfy
robin-aws, line 1: Make sure you have a test case or two where an option is specified on the command line and in a project file. The former should win, and I suspect your implementation favours the latter instead.

Also make sure you have at least one test with Dafny files nested more than one level deep. Very common bug to assume your glob handler treats `**` as recursive descent when it doesn't (a trap I've fallen into on the command line myself several times :)

Source/DafnyLanguageServer.Test/DafnyLanguageServer.Test.csproj
```diff
@@ -68,6 +68,21 @@
     <None Update="Various\TestFiles\includesAincludesB.dfy">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="ProjectFiles\TestFiles\src\foo.dfy">
```
robin-aws, line 71: Could we replace all these with a a single pattern somehow? I'm betting newcomers to the code base will forget to do this and be confused why their test doesn't pass.

FYI IntegrationTests does this to pick up all lit test files: https://github.com/dafny-lang/dafny/blob/master/Source/IntegrationTests/IntegrationTests.csproj#L41-L43

Understandable if this is scope creep for this PR though.

Source/DafnyLanguageServer/Workspace/ITextDocumentLoader.cs
```diff
@@ -26,11 +26,13 @@ public interface ITextDocumentLoader {
     /// Loads the specified document item of the language server and applies the necessary steps
     /// to generate a dafny document instance.
     /// </summary>
+    /// <param name="options"></param>
```
robin-aws, line 29: I'm generally against these rote documentation elements, even when they do have text if the text just describes the name of the element. Is our build process forcing them?

docs/DafnyRef/UserGuide.md
```diff
@@ -571,6 +571,23 @@ Most output from `dafny` is directed to the standard output of the shell invokin
...
+
+```toml
+includes = ["src/**/*.dfy"]
+excludes = ["**/ignore.dfy"]
+
+[options]
+  enforce-determinism = true
```
robin-aws, line 583: Do these have to be indented?

docs/DafnyRef/UserGuide.md
```diff
@@ -571,6 +571,23 @@ Most output from `dafny` is directed to the standard output of the shell invokin
 - Dafny `expect` statements (when they fail) send a message to **standard-out**.
 - Dafny I/O libraries send output explicitly to either **standard-out or standard-error**
 
+### 13.5.5. Project files
+
+Commands on the Dafny CLI that can be passed a Dafny file, can also be passed a Dafny project file. Such a project file may define which Dafny files the project contains, and which Dafny options it uses. The project file must be a [TOML](https://toml.io/en/) file named `dafny.toml`. Here's an example of a Dafny project file:
+
+```toml
+includes = ["src/**/*.dfy"]
```
robin-aws, line 579: Is there a default pattern, or is the default no files?

robin-aws (changes requested):
Very important and long-awaited feature, and I'm happy we're starting off nice and small. Reviewed the feature design, docs and tests at a high level, but not the actual implementation just yet.

docs/dev/news/2907.feat
keyboardDrummer: Immediately true

docs/DafnyRef/UserGuide.md
keyboardDrummer: > It feels a bit weird to be passing the project file directly as just another file, rather than with a flag like --project. It also means you can't have multiple project files for a codebase in the same directory

Why would it mean that? Passing project files as just other files means that the extension needs to be used for Dafny to determine how to interpret the file. Any `.toml` file is considered a project file. That's fine right?

> It also means you can't have multiple project files for a codebase in the same directory, which is a less common scenario but one I have seen in other systems frequently enough.

The IDE needs to be able to quickly find a unique project file based on the path of a `.dfy` file. You could have all project files end in the same suffix, like `*.dafny.toml` or `*.dfyp`, but what do you when the IDE finds two project files at the same distance of a `.dfy` file? You can prompt the user to ask which one should be used, but that seems needlessly complicated to me.

Source/DafnyLanguageServer.Test/DafnyLanguageServer.Test.csproj
keyboardDrummer: What would be the pattern? Everything in `ProjectFiles/TestFiles` ? Then you might still forget to add another pattern for `NewFolder/TestFiles`

docs/DafnyRef/UserGuide.md
keyboardDrummer: Added a default

docs/DafnyRef/UserGuide.md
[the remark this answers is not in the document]
keyboardDrummer: You can use non-Dafny files as source too, using the same `includes` property.

Test/cli/projectFile/projectFile.dfy
keyboardDrummer: Check

docs/DafnyRef/UserGuide.md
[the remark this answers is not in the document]
robin-aws: Fair enough, but that won't work very well for projects with target language files for multiple target languages. That's also part of why I asked about having multiple project files, but as you say that complicates things for the IDE.

I think the right answer is that we'll likely want per-target language configuration of some kind, something like:

```toml
includes = ["src/**/*.dfy"]

[java]
includes = ["src/**/*.jar"]

[csharp]
includes = ["src/**/*.dll"]
```

I just want to make there's a clear path to supporting that somehow.

docs/DafnyRef/UserGuide.md
```diff
@@ -571,6 +571,23 @@ Most output from `dafny` is directed to the standard output of the shell invokin
...
+excludes = ["**/ignore.dfy"]
+
+[options]
+enforce-determinism = true
+warn-shadowing = true
```
robin-aws, line 584: I'd add just a bit more documentation on the exact syntax here and what's accepted, especially for users not familiar with TOML.

Does the legacy CLI support project files as well? I'm assuming not since the options don't align 1:1. We should document that too.

Source/DafnyLanguageServer.Test/DafnyLanguageServer.Test.csproj
robin-aws: I would reverse the ordering so you have `TestFile/ProjectFiles` and `TestFiles/NewFolder` instead. But again probably better not to add to this PR since it would then be a lot of edits.

robin-aws (commented):
Replied to comments, will review the implementation shortly

docs/DafnyRef/UserGuide.md
keyboardDrummer: Yes it has to be named `dafny.toml`, but that's so the IDE can easily find it, not because the project files are in the same CLI argument location as source files.

Source/DafnyLanguageServer.Test/DafnyLanguageServer.Test.csproj
keyboardDrummer: Indeed nicer to leave out of scope

Test/cli/projectFile/projectFile.dfy
robin-aws, line 1: (not for this PR, just an idea) I suspect it will be very useful to use project files as lit tests directly, and I THINK this syntax will work with `lit` and the xUnit test runner as is:

```toml
# RUN: %baredafny resolve --use-basename-for-filename "%s" > "%t"
# RUN: %diff "%s.expect" "%t"

includes = ["**/*"]

[options]
warn-shadowing = 3
```

We'd just have to tweak `lit.site.cfg` and `IntegrationTests` to pick up `*.toml` as well as `*.dfy`. We might have to separate project-based tests so we don't pick up source files twice mind you.

Test/cli/projectFile/projectFile.dfy
```diff
@@ -0,0 +1,22 @@
+// A project file can specify input files and configure options
+// RUN: %baredafny resolve --use-basename-for-filename "%S/dafny.toml" > "%t"
+
+// Test option override behavior
+// RUN: %baredafny resolve --use-basename-for-filename "%S/dafny.toml" --warn-shadowing=false >> "%t"
+
+// Multiple project files are not allowed
+// RUN: ! %baredafny resolve --use-basename-for-filename "%S/dafny.toml" "%S/broken/dafny.toml"
```
robin-aws, line 8: Don't you still want to assert the error output in the expect file for these errors too?

docs/DafnyRef/UserGuide.md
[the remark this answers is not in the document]
robin-aws: Okay cool, I think the nested `[a.b.c]` section style gives us a lot of flexibility in the future.

FYI I'm also tempted to find a way to include specific files when building a codebase with specific options such as `--unicode-char`, as I'm using that approach to address https://github.com/dafny-lang/libraries/issues/111

docs/DafnyRef/UserGuide.md
```diff
@@ -571,6 +571,25 @@ Most output from `dafny` is directed to the standard output of the shell invokin
...
+```
+
+Under the section `[options]`, any options from the Dafny CLI can be specified using the option's name without the `--` prefix. When executing a `dafny` command using a project file, any options specified in the file that can be applied to the command, will be. Options that can't be applied or are misspelled, are ignored.
```
robin-aws, line 587: Hmm, not catching misspellings is unfortunate. Couldn't we check that all options resolve after parsing, even if they aren't used for the current command?

Source/DafnyLanguageServer.Test/ProjectFiles/TestFiles/Nesting/dafny.toml
robin-aws, line 1: I'd rename at least one project file to have a base name other than `dafny`, just to ensure we're not assuming that anywhere

robin-aws (changes requested):
Finished reviewing the rest

Source/DafnyCore/Options/ProjectFile.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.Extensions.FileSystemGlobbing;
+using Microsoft.Extensions.FileSystemGlobbing.Abstractions;
+using Tomlyn;
+
+namespace Microsoft.Dafny; 
+
+public class ProjectFile {
+  [IgnoreDataMember]
+  public Uri Uri { get; set; }
```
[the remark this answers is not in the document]
keyboardDrummer: I'm not sure what change your comment is asking for. If it's a local path or a URL, isn't a `Uri` the right type? My impression is that `Uri` is great for paths since it's more platform independent than `string`.

Source/DafnyCore/Options/ProjectFile.cs
```diff
@@ -0,0 +1,77 @@
...
+    } catch (IOException e) {
+      errorWriter.WriteLine(e.Message);
+      return null;
+    } catch (TomlException tomlException) {
+      errorWriter.WriteLine($"The Dafny project file {uri.LocalPath} contains the following errors:");
+      var regex = new Regex(
+        @$"\((\d+),(\d+)\) : error : The property `(\w+)` was not found on object type {typeof(ProjectFile).FullName}");
+      var newMessage = regex.Replace(tomlException.Message,
+        match =>
+          $"({match.Groups[1].Value},{match.Groups[2].Value}): the property {match.Groups[3].Value} does not exist.");
```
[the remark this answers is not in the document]
keyboardDrummer: Until we update the TOML library to use customized error types, yes.

Source/DafnyDriver/Commands/CommandRegistry.cs
```diff
@@ -105,24 +105,36 @@ static class CommandRegistry {
...
+          if (!ProcessFile(dafnyOptions, file)) {
+            failedToProcessFile = true;
+            return;
+          }
         }
       }
 
       foreach (var option in command.Options) {
-        var value = context.ParseResult.GetValueForOption(option);
+        var result = context.ParseResult.FindResultFor(option);
```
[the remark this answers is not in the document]
keyboardDrummer: `GetValueForOption` always returns a value, the default value in case none was specified.

Source/DafnyLanguageServer.Test/Lookup/HoverTest.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Lookup {
   public class HoverTest : ClientBasedLanguageServerTest {
     protected override async Task SetUp(Action<DafnyOptions> modifyOptions = null) {
       void ModifyOptions(DafnyOptions options) {
-        options.ProverOptions.Add("-proverOpt:SOLVER=noop");
+        options.ProverOptions.Add("SOLVER=noop");
```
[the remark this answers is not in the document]
keyboardDrummer: I guess so. It produced an exception in the CI log that's how I found it. Guess it doesn't break the tests.

Test/cli/projectFile/projectFile.dfy
keyboardDrummer: Yes but the output contains, and should contain, absolute paths that differ between machines so I can't test against them. I'd have to filter them out before doing the diff.

Source/DafnyLanguageServer.Test/ProjectFiles/ProjectFilesTest.cs
```diff
@@ -0,0 +1,33 @@
...
+    var documentItem = CreateTestDocument(source, filePath);
+    await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
+    var diagnostics = await GetLastDiagnostics(documentItem, CancellationToken);
+
+    Assert.Single(diagnostics);
+    Assert.Equal("Shadowed local-variable name: x", diagnostics[0].Message);
+  }
+
+  [Fact]
+  public async Task ProjectFileOverridesOptions() {
```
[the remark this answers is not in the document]
keyboardDrummer: Added those two tests

Test/cli/projectFile/badTypes/dafny.toml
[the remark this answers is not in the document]
keyboardDrummer: There doesn't have to be, but this project file is valid.

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -127,6 +133,40 @@ public class DocumentManager {
     Compilation.Start();
   }
 
+  private static DafnyOptions DetermineDocumentOptions(DafnyOptions serverOptions, DocumentUri uri) {
+    ProjectFile? projectFile = null;
+
+    var folder = Path.GetDirectoryName(uri.GetFileSystemPath());
+    while (!string.IsNullOrEmpty(folder)) {
+      var children = Directory.GetFiles(folder, "dafny.toml");
```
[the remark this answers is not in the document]
keyboardDrummer: I don't think that's a good idea, then you can't have non-Dafny `*.toml` files in your filetree above a `.dfy` file, because the IDE would recognize them as Dafny project files.

We could allow any `*dafny.toml`, but that seems needlessly complicated. Is there a problem with only accepting `dafny.toml` ? If you want to have different project files that use different source files, there shouldn't be an issue in putting those in a directory as well. It's similar to TypeScript: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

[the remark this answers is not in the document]
keyboardDrummer: Anyways, if we leave it like it is now, we can always allow more later ;-)

Test/cli/projectFile/badTypes/dafny.toml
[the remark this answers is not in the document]
keyboardDrummer: > Regardless, we should also have test cases that explicit state what happens when your file isn't called dafny.toml.

Added the test, and a warning when a non `dafny.toml` project file is used.

Source/DafnyCore/Options/ProjectFile.cs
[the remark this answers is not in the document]
robin-aws: Ah if this is just a clean implementation for a local path that's fine too. But we should probably double check the protocol is always `file:///`.

robin-aws:
I wasn't sure from previous versions whether we were going to require project files to be named `dafny.toml`, or if we'd support any `*.toml` file. It sounds like you want the former, which I'm fine with and less of a one-way door, so let's go with that, but just make sure the documentation and tests are all consistent. :)

keyboardDrummer:
> I wasn't sure from previous versions whether we were going to require project files to be named `dafny.toml`, or if we'd support any `*.toml` file. It sounds like you want the former, which I'm fine with and less of a one-way door, so let's go with that, but just make sure the documentation and tests are all consistent. :)

I think it's consistent. A `*.toml` file is allowed on the CLI, but gives a warning that the IDE will only use `dafny.toml` files. The documentation says to use `dafny.toml` file.

Source/DafnyCore/Options/ProjectFile.cs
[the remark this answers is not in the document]
keyboardDrummer: Added a check to make sure passed project files exist on disk.

docs/dev/news/2907.feat
robin-aws, line 3: Should mention the file has to be called `dafny.toml` here.

docs/dev/news/2907.feat
```diff
@@ -0,0 +1,3 @@
+Added support for Dafny project files. For now the project file only allows specifying which Dafny files to include and exclude, and what options to use.
```
robin-aws, line 1: Should probably mention the supported format is `*.toml` here, or perhaps link to the refman section.

docs/DafnyRef/UserGuide.md
```diff
@@ -571,6 +571,25 @@ Most output from `dafny` is directed to the standard output of the shell invokin
 - Dafny `expect` statements (when they fail) send a message to **standard-out**.
 - Dafny I/O libraries send output explicitly to either **standard-out or standard-error**
 
+### 13.5.5. Project files
+
+Commands on the Dafny CLI that can be passed a Dafny file, can also be passed a Dafny project file. Such a project file may define which Dafny files the project contains, and which Dafny options it uses. The project file must be a [TOML](https://toml.io/en/) file named `dafny.toml`. Here's an example of a Dafny project file:
```
robin-aws, line 576: I think we're clear on the behaviour now, which means this should say that any `*.toml` file is accepted...

docs/DafnyRef/UserGuide.md
```diff
@@ -571,6 +571,25 @@ Most output from `dafny` is directed to the standard output of the shell invokin
...
+excludes = ["**/ignore.dfy"]
+
+[options]
+enforce-determinism = true
+warn-shadowing = true
+```
+
+Under the section `[options]`, any options from the Dafny CLI can be specified using the option's name without the `--` prefix. When executing a `dafny` command using a project file, any options specified in the file that can be applied to the command, will be. Options that can't be applied or are misspelled, are ignored.
+
+When using a Dafny IDE based on the `dafny server` command, the IDE will search for project files by traversing up the file tree, and options from the first found project file will override options passed to `dafny server`.
```
robin-aws, line 589: ...but this should say only `dafny.toml` files will be found.

robin-aws (commented):
Looks great aside from a couple last documentation corrections!

And one more quick thing before walking through the one-way door, hopefully just a quick find-and-replace: in the style of `tsconfig.json`, I feel like `dafnyconfig.toml` would be slightly more helpful. WDYT?

docs/DafnyRef/UserGuide.md
keyboardDrummer: Updated

docs/DafnyRef/UserGuide.md
keyboardDrummer: Updated

robin-aws (approved):
🎉

PR #3858, opened 2023-04-10
Remove IncludeToken as prep for caching parsing

keyboardDrummer:
Currently a file that's included is parsed differently. Its tokens are wrapped in an `IncludeToken`, that includes a reference to the including file.

To allow reusing parsed files in a cache, how the file is parsed should be independent of what triggered it to be parsed. To achieve that, this PR removes `IncludeToken`. To enable still determining whether a particular token was part of an include or not, tokens now store a `Uri` that uniquely identifies the document that token came from. The Uris of the root documents are stored in `DefaultModuleDefinition`, which is just below `Program`. Any token with a Uri that is not part of the root set is considered to be included.


Source/DafnyLanguageServer/Util/PathExtensions.cs
```diff
@@ -62,24 +52,24 @@ public static class PathExtensions {
...
 
     /// <summary>
     /// Gets the filename of the document where the given boogie token is located in.
     /// </summary>
     /// <param name="token">The token to get the document filename from.</param>
     /// <returns>The filename (without path) of the document containing the given token.</returns>
-    public static string GetDocumentFileName(this Boogie.IToken token) {
-      return Path.GetFileName(token.filename);
+    public static string? GetDocumentFileName(this Boogie.IToken token) {
+      if (token.filename == null) {
```
MikaelMayer, line 68: Same here, I would like the token's type to be filtered to ensure the cast does not throw a cast exception

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -55,8 +55,7 @@ LineVerificationStatus[] PerLineStatus
 
       // Render verification tree content into lines.
       foreach (var verificationTree in verificationTrees) {
-        if (verificationTree.Filename == uri.ToString() ||
-            "untitled:" + verificationTree.Filename == uri) {
+        if (verificationTree.Uri == uri) {
```
MikaelMayer, line 58: That sounds like a more robust comparison !

Source/DafnyPipeline.Test/Issue1355.cs
```diff
@@ -16,16 +18,11 @@ public class Issue1355 {
...
       options.DafnyPrelude = "../../../../../Binaries/DafnyPrelude.bpl";
 
       var programString = @"trait Trait<A, B> { }";
-      ModuleDecl module = new LiteralModuleDecl(new DefaultModuleDefinition(), null);
-      Microsoft.Dafny.Type.ResetScopes();
-      BuiltIns builtIns = new BuiltIns(options);
-      Parser.Parse(programString, "virtual", "virtual", module, builtIns, reporter);
-      var dafnyProgram = new Program("programName", module, builtIns, reporter);
-      Main.Resolve(dafnyProgram, reporter);
+      var dafnyProgram = Utils.Parse(options, programString, false);
```
MikaelMayer, line 24: Why is the parsing in Utils, but the Resolve is in Main? Could you move Parsing also in Main, or is there something I need to know?

Source/DafnyPipeline.Test/PluginsTest.cs
```diff
@@ -45,71 +47,49 @@ public class PluginsTest {
     return assemblyPath;
   }
 
-  class CollectionErrorReporter : BatchErrorReporter {
```
MikaelMayer, line 48: Makes sense to delete it. I wrote It originally because the BatchErrorReporter's error messages were private. Now that they are accessible, this class is no longer needed.

Source/DafnyPipeline.Test/Trivia.cs
```diff
@@ -71,11 +71,8 @@ ensures true
 ";
         programString = AdjustNewlines(programString);
 
-        ModuleDecl module = new LiteralModuleDecl(new DefaultModuleDefinition(), null);
-        Microsoft.Dafny.Type.ResetScopes();
-        BuiltIns builtIns = new BuiltIns(options);
-        Parser.Parse(programString, "virtual", "virtual", module, builtIns, reporter);
-        var dafnyProgram = new Program("programName", module, builtIns, reporter);
+        var dafnyProgram = Utils.Parse(options, programString, false);
+        var reporter = dafnyProgram.Reporter;
```
MikaelMayer, line 75: Good job doing this refactoring over and over, it's more readable.

Test/allocated1/dafny0/CompilationErrors.dfy.expect
```diff
@@ -5,6 +5,7 @@ CompilationErrors.dfy(83,4): Warning: note, this loop has no body (loop frame: x
 CompilationErrors.dfy(85,4): Warning: note, this loop has no body (loop frame: i, x)
 
 Dafny program verifier finished with 5 verified, 0 errors
+CompilationErrors.dfy(3,8): Error: the included file CompilationErrors.dfy contains error(s)
```
MikaelMayer, line 8: That message is weird. How come it references itself?

Source/DafnyCore/AST/IncludeHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+    if (token is RefinementToken) {
+      return false;
+    }
+
+    if (token == Token.NoToken) {
+      return false;
+    }
+
+    var files = outerModule.RootUris;
+    if (files.Contains(token.Uri)) {
```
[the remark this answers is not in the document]
keyboardDrummer: `RootUris` contains all the documents that are parsed for reasons other than being included, such as being passed in the command line, being specified in the project file, or being the open file in the IDE.

Source/DafnyCore/DafnyMain.cs
```diff
@@ -135,6 +138,7 @@ public class DafnyFile {
       foreach (var attrHandle in dllMetadataReader.CustomAttributes) {
         var attr = dllMetadataReader.GetCustomAttribute(attrHandle);
         try {
+          // The cast to MemberReferenceHandle is an explicit operator that uses internal properties and throws a cast exception in some cases.
```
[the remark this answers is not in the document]
keyboardDrummer: Properties that we can't access, so we cannot make our own version that doesn't throw an InvalidCastException.

Source/DafnyCore/Generic/Reporting.cs
```diff
@@ -38,16 +43,13 @@ public abstract class ErrorReporter {
...
-    public void Error(MessageSource source, string errorId, IToken tok, string msg) {
+    public virtual void Error(MessageSource source, string errorId, IToken tok, string msg) {
       Contract.Requires(tok != null);
       Contract.Requires(msg != null);
-      // if the tok is IncludeToken, we need to indicate to the including file
-      // that there are errors in the included file.
-      if (tok is IncludeToken) {
-        IncludeToken includeToken = (IncludeToken)tok;
-        Include include = includeToken.Include;
+      if (tok.IsIncludeToken(OuterModule) && OuterModule != null) {
```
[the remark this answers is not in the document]
keyboardDrummer: Sounds good, done.

Source/DafnyCore/JsonDiagnostics.cs
```diff
@@ -33,7 +33,7 @@ record DiagnosticMessageData(MessageSource source, ErrorLevel level, Boogie.ITok
 
   private static JsonObject SerializeToken(Boogie.IToken tok) {
     return new JsonObject {
-      ["filename"] = tok.filename,
+      ["uri"] = ((IToken)tok).Uri.AbsoluteUri,
```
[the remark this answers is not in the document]
keyboardDrummer: I'll keep it

Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
```diff
@@ -594,19 +594,6 @@ decreases y
       Assert.Contains("a postcondition", diagnostics[0].Message);
     }
 
-    [Fact]
-    public async Task OpeningDocumentThatIncludesDocumentWithSemanticErrorsReportsResolverErrorAtInclude() {
-      var source = "include \"syntaxError.dfy\"";
```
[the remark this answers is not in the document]
keyboardDrummer: It was a duplicate of `OpeningDocumentThatIncludesDocumentWithSyntaxErrorsReportsParserErrorAtInclude`

Also the title was wrong, since it uses the file `syntaxError` and the title contains `SemanticErrors`.

Test/allocated1/dafny0/CompilationErrors.dfy.expect
keyboardDrummer: There are two files with the same name, one includes the other.

Source/DafnyPipeline.Test/Issue1355.cs
keyboardDrummer: The Parse function is only used in tests, so it's in a testing `Utils` that's not in DafnyCore.

I'm hoping in the future there's be more consolidation, and everything can use `Compilation` to do their thing.

MikaelMayer (changes requested):
Getting closer! A few more items

Source/DafnyCore/Coco/Scanner.frame
```diff
@@ -415,48 +411,6 @@ public class Scanner {
         if (ch == '\r' && buffer.Peek() != '\n') ch = EOL;
         if (ch == EOL) {
           line++; col = 0;
-        } else if (false && ch == '#' && col == 1) {
```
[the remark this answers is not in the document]
RustanLeino: This code for recognizing `#line` pragmas was inherited from Boogie. We're currently not using it in Dafny, as evidenced by the `false &&` on L418. If the code only looked for `#line` (and not `#` followed by other things), then it does not interfere with otherwise legal Dafny programs. Still, since we're not using `#line`, it's fine with me if we remove the code. If we ever wanted this `#line` functionality in the future, it would be easy enough to copy the code from Boogie again or to write the code anew.

MikaelMayer (approved):
Great job and perseverance going through this refactoring!

PR #3864, opened 2023-04-12
Improve rules for nameonly/positional/nameless and required/optional …

### LINKED ISSUE #3859: `nameonly` should not constrain default values
### Dafny version

3.13

```dafny
datatype Foo = Foo(
  nameonly bar: Option<string> := None,
  nameonly baz: Option<string>
)


This is probably because `Foo("positional parameters", "have order")`.
However `nameonly` means that I MUST

```dafny
Foo(
  baz := "Look ma, no order!"
)
```
```


_No response_

### What happened?

In the above example `baz` gets:
`a required parameter must precede all optional parameters`

Mac

### PR DESCRIPTION
**Pull request by RustanLeino**
Fixes #3859

This PR allows more parameter declarations (for example, as suggested in #3859, a `nameonly` parameter with a default value can now be follow by other `nameonly` parameters), clarifies some error messages, and gives new errors for useless situations (for example, a parameter's default value can not be used if it is followed by a positional-only parameter).

## The new rules

Formal parameters have three ways to indicate how they are to be passed in:

* `nameonly`: the only way to give a specific argument value is to name the parameter
* positional only: these are nameless parameters (which are allowed only for datatype constructor parameters)
* either positional or by name: this is the most common parameter

A parameter is either required or optional:
* required: a caller has to supply an argument
* optional: the parameter has a default value that is used if a caller omits passing a specific argument

The syntax for giving a positional-only (i.e., nameless) parameter does not a default-value expression, so a positional-only parameter is always required.

At a call site, positional arguments are not allowed to follow named arguments. Therefore, if `x` is a nameonly parameter, then there is no way to supply the parameters after `x` by position. Thus, any parameter that follows `x` must either by passed by name or have a default value. That is, if a later parameter does not have a default value, it is _effectively_ nameonly. We impose the rule that

* an effectively nameonly parameter must be declared as nameonly

For a positional-only parameter `x`, every parameter preceding `x` is _effectively_ required. We impose the rule that

* an effectively required parameter must not have a default-value expression

### DIFF
--- a/Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
+++ b/Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
@@ -4681,12 +4681,12 @@ namespace Microsoft.Dafny {
       //   * required: a caller has to supply an argument
       //   * optional: the parameter has a default value that is used if a caller omits passing a specific argument
       //
-      // The syntax for giving a positional-only (i.e., nameless) parameter does not a default-value expression, so
+      // The syntax for giving a positional-only (i.e., nameless) parameter does not allow a default-value expression, so
       // a positional-only parameter is always required.
       //
       // At a call site, positional arguments are not allowed to follow named arguments. Therefore, if "x" is
       // a nameonly parameter, then there is no way to supply the parameters after "x" by position. Thus, any
-      // parameter that follows "x" must either by passed by name or have a default value. That is, if a later
+      // parameter that follows "x" must either be passed by name or have a default value. That is, if a later
       // parameter does not have a default value, it is _effectively_ nameonly. We impose the rule that
       //   * an effectively nameonly parameter must be declared as nameonly
       //
@@ -4716,7 +4716,7 @@ namespace Microsoft.Dafny {
           // "formal" is preceded by a nameonly parameter, but itself is neither nameonly nor has a default value
           reporter.Error(MessageSource.Resolver, formal.tok,
             $"this parameter is effectively nameonly (because of the earlier nameonly parameter '{nameOfMostRecentNameonlyParameter}'); " +
-            "declare it as nameonly or give it an default-value expression");
+            "declare it as nameonly or give it a default-value expression");
         }
         if (formal.IsNameOnly) {
           nameOfMostRecentNameonlyParameter = formal.Name;
--- a/Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
+++ b/Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
@@ -4673,26 +4673,53 @@ namespace Microsoft.Dafny {
 
       Contract.Assume(AllTypeConstraints.Count == 0);
 
+      // Formal parameters have three ways to indicate how they are to be passed in:
+      //   * nameonly: the only way to give a specific value is to name the parameter
+      //   * positional only: these are nameless arguments (which are allowed only for datatype constructor parameters)
+      //   * either positional or by name: this is the most common parameter
+      // A parameter is either required or optional:
+      //   * required: a caller has to supply an argument
+      //   * optional: the parameter has a default value that is used if a caller omits passing a specific argument
+      //
+      // The syntax for giving a positional-only (i.e., nameless) parameter does not a default-value expression, so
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by seebees — DISMISSED** [review]
LGTM

File: Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs [hunk 1]
```diff
@@ -4673,26 +4673,53 @@ public XConstraint_EquatableArg(IToken tok, Type a, Type b, bool allowSuperSub,
...
 
+      // Formal parameters have three ways to indicate how they are to be passed in:
+      //   * nameonly: the only way to give a specific argument value is to name the parameter
+      //   * positional only: these are nameless parameters (which are allowed only for datatype constructor parameters)
+      //   * either positional or by name: this is the most common parameter
+      // A parameter is either required or optional:
+      //   * required: a caller has to supply an argument
+      //   * optional: the parameter has a default value that is used if a caller omits passing a specific argument
+      //
+      // The syntax for giving a positional-only (i.e., nameless) parameter does not a default-value expression, so
```
keyboardDrummer, line 4684: `does not a` => `does not allow a`

Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
```diff
@@ -4673,26 +4673,53 @@ public XConstraint_EquatableArg(IToken tok, Type a, Type b, bool allowSuperSub,
...
+      // a positional-only parameter is always required.
+      //
+      // At a call site, positional arguments are not allowed to follow named arguments. Therefore, if "x" is
+      // a nameonly parameter, then there is no way to supply the parameters after "x" by position. Thus, any
+      // parameter that follows "x" must either by passed by name or have a default value. That is, if a later
```
keyboardDrummer, line 4689: `be passed`

Source/DafnyCore/Resolver/NameResolutionAndTypeInference.cs
```diff
@@ -4673,26 +4673,53 @@ public XConstraint_EquatableArg(IToken tok, Type a, Type b, bool allowSuperSub,
...
-          reporter.Error(MessageSource.Resolver, formal.tok, "a required parameter must precede all optional parameters");
-        }
-        if (!allowNamelessParameters && !formal.HasName) {
-          reporter.Error(MessageSource.Resolver, formal.tok, "a nameless parameter must precede all nameonly parameters");
+          previousParametersWithDefaultValue.Add(formal);
+        } else if (nameOfMostRecentNameonlyParameter != null && !formal.IsNameOnly) {
+          // "formal" is preceded by a nameonly parameter, but itself is neither nameonly nor has a default value
+          reporter.Error(MessageSource.Resolver, formal.tok,
+            $"this parameter is effectively nameonly (because of the earlier nameonly parameter '{nameOfMostRecentNameonlyParameter}'); " +
+            "declare it as nameonly or give it an default-value expression");
```
keyboardDrummer, line 4719: `a default-value expression`

Test/dafny0/ParameterResolution.dfy
```diff
@@ -275,15 +275,15 @@ module NameOnlyParameters {
     least predicate LP(a: int, nameonly b: int, c: int := 100)
     static greatest predicate GP(a: int, nameonly b: int := 75, c: int := 100)
   }
-  iterator Iter(nameonly u: int, x: int)
+  iterator Iter(nameonly u: int, x: int) // error (x2, because of the desugaring of iterators): x is effectively nameonly, but not declared as such
```
keyboardDrummer, line 278: Maybe we should come up with a testing format where you can specify the expected error messages inline : )

keyboardDrummer (dismissed):

keyboardDrummer:
Such a nice quality of life improvement, thanks !

Test/dafny0/ParameterResolution.dfy
RustanLeino: I do like being able to read the expected outcome in the test file itself. Therefore, I always put such "error" comments in the tests. But these comments aren't the exact error messages, don't capture column information, and in some cases are even placed only on a nearby line. It's plausible that we could turn these into a formally recognized testing file format.

PR #3870, opened 2023-04-14
Some auditor adjustments

atomb:
Makes several changes to the output of `dafny audit` that I'm pretty sure are the right call:
* Audit only explicitly-specified files, not included files, like the default of `dafny verify`
* Report the number of findings (especially useful when there are none)
* Report on any `function {:extern}` declaration (suggesting `method` with `modifies *` instead)
* Don't report separately on bodyless externs (since they'll be reported, anyway, for other reasons), and even an extern with a body is making an assumption
* Treat `assume {:axiom}` differently from a normal `assume`
* Have separate messages for IETF format (which are similar to, but not identical to the normal messages, particularly in the use of `SHOULD` vs `MUST`)

And one I'm still not sure about:
* If an `extern` declaration or definition is marked with `{:axiom}`, don't complain about contracts


Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -4,56 +4,90 @@
...
+) {
   public static AssumptionDescription HasAxiomAttribute = new(
     issue: "Declaration has explicit [{:axiom}] attribute.",
     mitigation: "Provide a proof or test.",
+    mitigationIETF: "SHOULD provide a proof or test.",
     isExplicit: true);
   public static AssumptionDescription HasVerifyFalseAttribute = new(
     issue: "Declaration has [{:verify false}] attribute.",
     mitigation: "Remove and prove if possible.",
+    mitigationIETF: "MUST remove and prove.",
```
MikaelMayer, line 21: replace by assertion and prove?

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -4,56 +4,90 @@
...
+    isExplicit: false);
+  public static AssumptionDescription ExternFunction = new(
+    issue: "Function with [{:extern}] attribute.",
+    mitigation: "Turn into a [method] or test thoroughly for lack of side effects.",
```
MikaelMayer, line 25: a method with `modifies *` ?
Personally, using a "function" instead of a "method" for externs is just making two more assumptions: determinism w.r.t. inputs and no side-effects.
Would that be better if we observed one more attribute, e.g. {:axiom} ?

MikaelMayer (commented):
Just two questions for now, otherwise looks good to me!

Source/DafnyCore/Auditor/Assumption.cs
atomb: I meant remove the _attribute_. I should update the wording to clarify that!

Source/DafnyCore/Auditor/Assumption.cs
atomb: Yeah, I think an `{:axiom}` annotation would make it clear that it's an intentional assumption.

So I'll update that message to suggest adding `modifies *` and testing for determinism, as well as potentially adding `{:axiom}`.

Test/auditor/TestAuditor.dfy.html.expect
```diff
@@ -41,15 +41,16 @@
...
-<tr><td>GenerateBytesWrapper</td><td>True</td><td>False</td><td>False</td><td>Definition has <code>assume</code> statement in body.</td><td>Replace with <code>assert</code> and prove.</td></tr>
+<tr><td>GenerateBytesWrapper</td><td>True</td><td>False</td><td>False</td><td>Definition has <code>assume {:axiom}</code> statement in body.</td><td>Replace with <code>assert</code> and prove.</td></tr>
 <tr><td>GenerateBytesNoGuarantee</td><td>True</td><td>False</td><td>True</td><td>Declaration with <code>{:extern}</code> has a requires clause.</td><td>Test any external callers (maybe with <code>/testContracts</code>).</td></tr>
-<tr><td>ExternFunction</td><td>True</td><td>False</td><td>True</td><td>Function with <code>{:extern}</code> attribute.</td><td>Turn into a <code>method</code> or test thoroughly for lack of side effects.</td></tr>
+<tr><td>ExternFunction</td><td>True</td><td>False</td><td>True</td><td>Function with <code>{:extern}</code> attribute.</td><td>Turn into a <code>method</code> with <code>modifies *</code> or test thoroughly for lack of side effects.</td></tr>
```
MikaelMayer, line 46: How can someone test thoroughly for the lack of side effects so that it can be acknowledged by the auditor? Is that even possible?

MikaelMayer (approved):
Looks good to me !

Test/auditor/TestAuditor.dfy.html.expect
atomb: The only current way to thoroughly make it acknowledged by the auditor is to use [Duvet](https://github.com/awslabs/duvet) and the `md-ietf` format. It allows you to explain (in text or code) how you satisfy all of a set of requirements.

PR #3886, opened 2023-04-19
fix: Better cloning of variable declarations in match statements

### LINKED ISSUE #3658: Compiling a disjunctive case gives a Boogie error that refers to a generated variable name
### Dafny version

3.12

```dafny
method mmm() returns (i: int) {
  match "abc"
    case "def"|"ghi" => { var st := 0; return st; }
    case _ => { return 1; }
}
```

```code
dafny build
```

### What happened?

An error message that refers to a generated variable name:

/Users/davidcok/test-dafny/z.dfy(18,30): Error: more than one declaration of variable name: st#0_0 

Mac

robin-aws:
Fixes #3658 

I'm not that familiar with the idioms around cloning so we might be able to improve this slightly. In general, cloning feels pretty tricky to get right in the current design, and I'd prefer in the longer term to be more disciplined about how we use the caching in a `Cloner` to avoid these kinds of bugs.

### DIFF
--- a/Source/DafnyCore/AST/Expressions/NestedMatchCaseStmt.cs
+++ b/Source/DafnyCore/AST/Expressions/NestedMatchCaseStmt.cs
@@ -5,7 +5,7 @@ using System.Linq;
 
 namespace Microsoft.Dafny;
 
-public class NestedMatchCaseStmt : NestedMatchCase, IAttributeBearingDeclaration {
+public class NestedMatchCaseStmt : NestedMatchCase, IAttributeBearingDeclaration, ICloneable<NestedMatchCaseStmt> {
   public readonly List<Statement> Body;
   public Attributes Attributes;
   Attributes IAttributeBearingDeclaration.Attributes => Attributes;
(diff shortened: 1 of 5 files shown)


atomb:
This looks pretty good to me, though I'm a little uncertain about the failing integration test. I'm inclined to wait for a look from @keyboardDrummer before finalizing this, since he understands the recent changes to pattern matching and the cloner more thoroughly than I do.

keyboardDrummer (dismissed):

Source/DafnyCore/AST/Expressions/NestedMatchCaseStmt.cs
```diff
@@ -21,6 +21,14 @@ public class NestedMatchCaseStmt : NestedMatchCase, IAttributeBearingDeclaration
     this.Attributes = attrs;
   }
 
+  private NestedMatchCaseStmt(Cloner cloner, NestedMatchCaseStmt original) : base(original.tok, original.Pat) {
```
keyboardDrummer, line 24: Would be better if this calls a cloning constructor in the base class.

keyboardDrummer:
Looking at the failing test:

I see

Print before PR: nested match, with a non-null flattened attribute, containing another nested match with a non-null flattened attribute

Print after PR: nested match, with a non-null flattened attribute, containing another nested match with a null flattened attribute

What's important is that the AST paths that always go through flattened, and never go through flattened, stay the same, and it seems like that's the case. The new printing behavior seems better, so let's update the print expectation.

I'm not sure where the change comes from. I think because of the additional cloning of statements, the traversal of the tree visitor at the top of `MatchFlattener` may have changed, since it keeps track of a set of visited nodes.

keyboardDrummer (approved):
Looks great, thanks

Source/DafnyCore/AST/Expressions/NestedMatchCaseStmt.cs
robin-aws: I had that thought too, but can be improved later at least.

robin-aws:
Thanks @keyboardDrummer! I suspected the same thing about the failing test but appreciate the second pair of eyes to confirm.

Dilan-s:
Hi @robin-aws, @keyboardDrummer 

I've attempted to update my clone of Dafny to the latest version and have noticed that this commit has lead to the assertion errors mentioned in #4004 (the error does not occur before this commit)

Especially this seems to have caused an issue with parallel assignments within match statements.

Dilan

PR #3888, opened 2023-04-20
Feat: Support for expressions in proof obligation descriptions

MikaelMayer:
This PR is a precursor of https://github.com/dafny-lang/dafny/pull/2153

- Support for proof obligation description to return a Dafny expression that would fail if asserted
- Tests in the case of a division by zero.

### DIFF
--- a/Source/DafnyCore/Verifier/ProofObligationDescription.cs
+++ b/Source/DafnyCore/Verifier/ProofObligationDescription.cs
@@ -5,6 +5,14 @@ using JetBrains.Annotations;
 namespace Microsoft.Dafny.ProofObligationDescription;
 
 public abstract class ProofObligationDescription : Boogie.ProofObligationDescription {
+  public abstract Expression GetAssertedExpr(DafnyOptions options);
+}
+
+// When there is no way to translate the asserted constraint in Dafny
+public abstract class ProofObligationDescriptionNotInDafny : ProofObligationDescription {
+  public override Expression GetAssertedExpr(DafnyOptions options) {
+    return null;
+  }
 }
 
 //// Arithmetic and logical operators, conversions
@@ -17,9 +25,19 @@ public class DivisorNonZero : ProofObligationDescription {
     "possible division by zero";
 
   public override string ShortDescription => "non-zero divisor";
+
+  private readonly Expression divisor;
+
+  public DivisorNonZero(Expression divisor) {
+    this.divisor = divisor;
+  }
+
+  public override Expression GetAssertedExpr(DafnyOptions options) {
+    return new BinaryExpr(divisor.tok, BinaryExpr.Opcode.Neq, divisor, new LiteralExpr(divisor.tok, 0));
+  }
 }
 
-public class ShiftLowerBound : ProofObligationDescription {
+public class ShiftLowerBound : ProofObligationDescriptionNotInDafny {
   public override string SuccessDescription =>
     "shift amount is always non-negative";
 
@@ -29,7 +47,7 @@ public class ShiftLowerBound : ProofObligationDescription {
   public override string ShortDescription => "shift lower bound";
 }
 
-public class ShiftUpperBound : ProofObligationDescription {
+public class ShiftUpperBound : ProofObligationDescriptionNotInDafny {
   public override string SuccessDescription =>
     $"shift amount is always within the width of the result ({width})";
 
@@ -45,7 +63,7 @@ public class ShiftUpperBound : ProofObligationDescription {
   }
 }
 
-public class ConversionIsNatural : ProofObligationDescription {
+public class ConversionIsNatural : ProofObligationDescriptionNotInDafny {
   public override string SuccessDescription =>
     $"{prefix}value to be converted is always a natural number";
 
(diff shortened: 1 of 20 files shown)


Source/DafnyCore/Verifier/Translator.ExpressionWellformed.cs
```diff
@@ -916,7 +916,7 @@ private class WFOptions {
                     zero = Bpl.Expr.Literal(0);
                   }
                   CheckWellformed(e.E1, wfOptions, locals, builder, etran);
-                  builder.Add(Assert(GetToken(expr), Bpl.Expr.Neq(etran.TrExpr(e.E1), zero), new PODesc.DivisorNonZero(), wfOptions.AssertKv));
+                  builder.Add(Assert(GetToken(expr), Bpl.Expr.Neq(etran.TrExpr(e.E1), zero), new PODesc.DivisorNonZero(e.E1), wfOptions.AssertKv));
```
atomb, line 919: My long-term wish would be that we could construct the expression once and then put that expression both in an assertion and in the proof obligation description. That way, we'd know that the two expressions always match. It's probably not worth holding this PR up over that, if only because the current side conditions are expressed as Boogie expressions and we really want Dafny expressions to show to the user. But it would be lovely to eventually be in a state where that was possible.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -7363,7 +7363,7 @@ public ForceCheckToken(IToken tok)
         }
         w = BplOr(body, w);
       }
-      builder.Add(Assert(tok, w, new PODesc.LetSuchThanExists()));
+      builder.Add(Assert(tok, w, new PODesc.LetSuchThatExists()));
```
atomb, line 7366: Good catch.

Source/DafnyCore/Verifier/ProofObligationDescription.cs
```diff
@@ -5,6 +5,14 @@
 namespace Microsoft.Dafny.ProofObligationDescription;
 
 public abstract class ProofObligationDescription : Boogie.ProofObligationDescription {
+  public abstract Expression GetAssertedExpr(DafnyOptions options);
+}
+
+// When there is no way to translate the asserted constraint in Dafny
+public abstract class ProofObligationDescriptionNotInDafny : ProofObligationDescription {
```
atomb, line 12: What do you think of this name, instead?
```suggestion
public abstract class ProofObligationDescriptionWithNoExpr : ProofObligationDescription {
```

docs/dev/news/expressions-on-assertions.feat
```diff
@@ -0,0 +1 @@
+API support for obtaining failed dafny expressions on assertions
```
atomb, line 1:
```suggestion
API support for obtaining the Dafny expression that is being checked by each assertion
```

Source/DafnyCore/Verifier/ProofObligationDescription.cs
```diff
@@ -290,7 +308,7 @@ public class PreconditionSatisfied : ProofObligationDescription {
   }
 }
 
-public class AssertStatement : ProofObligationDescription {
+public class AssertStatement : ProofObligationDescriptionNotInDafny {
```
atomb, line 311: Why don't we implement `GetAssertedExpr` for this one right away, since it should be readily available?

atomb (changes requested):
This is very nice. I just had a couple of small wording/naming suggestions, and a request that we do _two_ assertion types in this PR instead of just one. :)

Source/DafnyCore/Verifier/ProofObligationDescription.cs
MikaelMayer: The assertion is made of Boogie expressions... It is used in ensures and requires as well. This will require more work that I already started in a separate branch.

atomb: Ah, okay, let's leave it out for now, then. :)

atomb (approved):
Looks good!

Source/DafnyCore/Verifier/Translator.ExpressionWellformed.cs
robin-aws: I was troubled by this as well. Is there any way we could test (or even prove) that the Dafny-level expression is encoded as a Boogie expression that is equivalent to the actual Boogie expression?

MikaelMayer: For now, whenever it was built already, I will ensure I'm reusing the same expression that is first translated to Boogie. That said, proving is out of control is our code is not written in Dafny :-) Regarding testing, we cannot necessarily execute the code.
We could however emit an internal assertion in the code that the Dafny expression is equivalent to the Boogie expression.

However, this is not practical/desirable for many reasons
- **Not practical to generate the expression at translation level** Code actions will generate the assertion expression on demand on failed assertions, so that we avoid generating 2x more expressions. At this point, the verifier has already finished to run.
- **Not practical to always create a Dafny expression first** What we send to Boogie is often might not be not practical for the user. For example, for termination checks, we send a O(n^2) expression which is practical for Boogie to prove; but for the user, we want a linear nested expression
- **Not necessary because safeguards:** This expression will only be served as an hint. If it's too weak, anyway the user will see that it won't help proving the assertion. If it's too strong because we can't expose some Boogie concepts in Dafny, then not only we can't prove equivalence, but the user won't be able to prove it either, so the verifier acts as a safeguard
 in Boogie.
- **Not meaningful for many assertions** For `assert A == B`, we will want to generate `calc {A; B;}.` It does not make sense to generate the calc statement and convert it to Boogie Down the road, we will generate more than simple expressions, we will generate whole code actions, e.g. to prove decreases clauses in while loop or prove frame conditions.

PR #3890, opened 2023-04-21
Feat: Make assertions explicit in VSCode

MikaelMayer:
This PR is a precursor of https://github.com/dafny-lang/dafny/pull/2153 and builds on top of the recently merged #3888 to let users explicit failing assertions in the IDE.

For statements, it will add the assertion just above the first statement in which the assertion fails.
It works either by hovering and clicking
ImplicitASsertion1

Or navigating and using CTRL+; and ENTER
ImplicitASsertion2

 If not in a statement, it will add the assertion just above the enclosing expression before a branch.
Hence it won't create
`F(assert i + 1 != 0; 2 / (i + 1))` but instead
`assert i + 1 != 0; F(2/ (i + 1))`:
ImplicitASsertion3

Note that the assertion is added in a context that is maximal, that is, it should not escape any path condition, for example, in this case parentheses have to be added:
ImplicitASsertion4

I added tests on the language server side to capture this behavior.
For now, I am not using the formatter, only the pretty-printer to display the expression. However, I ensure the indentation is correct by mimicking the indentation of the expression being prefixed by the assertion.

I added the ability for code action tests to support multiple edits at once by testing only the final result.

There will be more code actions to come, but this is a good start.


Source/DafnyLanguageServer/Language/ImplicitFailingAssertionCodeActionProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+    private readonly Expression failingImplicitAssertion;
+    private readonly Node program;
+    private readonly Range selection;
+
+    public ExplicitAssertionDafnyCodeAction(
+      DafnyOptions options,
+      Dafny.Program program,
+      Expression failingImplicitAssertion,
+      Range selection
+      ) : base("Explicit failing assertion") {
```
robin-aws, line 52:
```suggestion
      ) : base("Insert explicit failing assertion") {
```

?

Source/DafnyLanguageServer/Language/ImplicitFailingAssertionCodeActionProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+    DafnyDiagnostic diagnostic, Range selection) {
+    var uri = input.Uri;
+    if (input.Program == null || diagnostic.Source != MessageSource.Verifier) {
+      return null;
+    }
+    var failingExpressions = new List<Expression>() { };
+    input.VerificationTree.ForEach(tree => {
+      if (tree is AssertionVerificationTree assertTree &&
+          assertTree.Finished &&
+            assertTree.Range.Intersects(selection) &&
```
robin-aws, line 102: This might match failed implicit assertions located in the specification text, right? Will it insert invalid syntax in that case?

This might work fine if you are able to insert `assert ...; <expr>` there, but we should at least have some test cases for that.

MikaelMayer: It will work fine in that case because it will do exactly what you suggest, that is, insert `assert ...;` before <expr>, even in specification expressions

Source/DafnyLanguageServer.Test/CodeActions/CodeActionsTest.cs
```diff
@@ -178,5 +295,32 @@ method f() returns (i: int)
           $"Did not find the code action '{expectedTitle}'. Available were:{string.Join(",", otherTitles)}");
       }
     }
+
+    private static List<string> ApplyEdits(TextDocumentEdit? textDocumentEdit, string output) {
+      var inversedEdits = textDocumentEdit.Edits.ToList()
+        .OrderBy(x => -1000 * x.Range.Start.Line - x.Range.Start.Character);
```
atomb, line 301: Would it be possible to do this ordering by lexicographic comparison on tuples, instead? If the code happened to have a very long long, this would miscalculate.

atomb (changes requested):
This is a nice feature to have! Thanks for implementing it. I think it mostly looks good, but there's one call to `OrderBy` that I think could be improved (in my comment below) and there are a few linter failures that should be easy to fix.

Source/DafnyLanguageServer.Test/CodeActions/CodeActionsTest.cs
MikaelMayer: I just found how, it's pretty elegant:
```
         .OrderByDescending(x => x.Range.Start.Line)
        .ThenByDescending(x => x.Range.Start.Character);
```

atomb: Nice!

atomb (approved):
Looks good!

PR #3895, opened 2023-04-21
Revising text about witness clauses

### LINKED ISSUE #3834: Misleading documentation about witness clause

Section 5.6.3. "Subset types" says

> The declaration of a subset type permits an optional [witness clause](https://dafny.org/latest/DafnyRef/DafnyRef#sec-witness), to declare default values that the compiler can use to initialize variables of the subset type, or to assert the non-emptiness of the subset type.

Unlike some other languages (like Java, C#, or Go), Dafny does not have the concept of a default value for a type. Therefore, the use of the phrase "to declare default values" above is misleading. The purpose of a `witness` clause is prove the existence of a value that the compiler can use to initialize a variable. The compiler may or may not choose this witness value when emitting code to auto-initialize a variable of the type.

I suggest changing the sentence above to

> The declaration of a subset type permits an optional [witness clause](https://dafny.org/latest/DafnyRef/DafnyRef#sec-witness), to declare an example value that the compiler may use to initialize variables of the subset type, or to assert the non-emptiness of the subset type.

or possibly something like

> The declaration of a subset type permits an optional [witness clause](https://dafny.org/latest/DafnyRef/DafnyRef#sec-witness), to declare an example value that the compiler may (but is not obligated to) use to initialize variables of the subset type, or to assert the non-emptiness of the subset type.

Another nit, while I'm at it. Section "5.6.3.4. Witness clauses", which is referred to in the above sentence, says

> For example, a numeric type that includes 0 is known by Dafny to be non-empty. 

It would be more precise to say

> For example, in the absence of a `witness` clause, a numeric type that includes 0 is known by Dafny to be non-empty. 

because `type MyUnknownIntegerType = x: int | true witness *` is not known by Dafny to be nonempty.

davidcok:
Fixes #3834

### DIFF
--- a/docs/DafnyRef/Types.md
+++ b/docs/DafnyRef/Types.md
@@ -1690,7 +1690,7 @@ In many cases, Dafny can determine that a newly declared type has
 some value. 
 For example, in the absence of a witness clause,
 a numeric type that includes 0 is known by Dafny
-to be non-empty,  
+to be non-empty.
 However, Dafny cannot always make this determination.
 If it cannot, a `witness` clause is required. The value given in
 the `witness` clause must be a valid value for the type and assures Dafny
--- a/docs/DafnyRef/Types.md
+++ b/docs/DafnyRef/Types.md
@@ -1518,7 +1518,10 @@ are never allowed, even if the value assigned is a value of the target
 type.  For such assignments, an explicit conversion must be used, see
 [Section 9.10](#sec-as-is-expression).)
 
-The declaration of a subset type permits an optional [`witness` clause](#sec-witness), to declare default values that the compiler can use to initialize variables of the subset type, or to assert the non-emptiness of the subset type.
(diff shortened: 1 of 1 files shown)


docs/DafnyRef/Types.md
```diff
@@ -1684,11 +1687,15 @@ The declaration of a subset type permits an optional `witness` clause.
 Types in Dafny are generally expected to be non-empty, in part because
 variables of any type are expected to have some value when they are used.
 In many cases, Dafny can determine that a newly declared type has 
-some value. For example, a numeric type that includes 0 is known by Dafny
-to be non-empty. However, Dafny cannot always make this determination.
+some value. 
+For example, in the absence of a witness clause,
+a numeric type that includes 0 is known by Dafny
+to be non-empty,
```
RustanLeino, line 1693:
```suggestion
to be non-empty.
```

RustanLeino (approved):

PR #3897, opened 2023-04-22
[Test Generation] Increase stack size during translation

Dargones:
Since test generation can now be done asynchronously, stack overflow is possible on large files due to the way the translator works. This PR makes sure the translator is called from a special thread that has enough stack space allocated for it.


Source/DafnyTestGeneration/Utils.cs
```diff
@@ -16,6 +16,23 @@ namespace DafnyTestGeneration {
...
 
+    /// <summary>
+    /// Call Translator with larger stack to prevent stack overflow
+    /// </summary>
+    public static List<Microsoft.Boogie.Program> Translate(Program program) {
+      var ret = new List<Microsoft.Boogie.Program> { };
+      var thread = new System.Threading.Thread(
+        () => {
+          ret = Translator
+    .Translate(program, program.Reporter)
```
atomb, line 27: This is a tiny nit, but could indent this line and the following beyond the indentation level of `ret`?

atomb (changes requested):

Source/DafnyTestGeneration/Utils.cs
Dargones: Done!

atomb (approved):
Looks good!

keyboardDrummer (approved):

PR #3902, opened 2023-04-24
Project aware ide

keyboardDrummer:


Source/DafnyLanguageServer/ServerCommand.cs
```diff
@@ -75,7 +75,13 @@ related locations
   public static void ConfigureDafnyOptionsForServer(DafnyOptions dafnyOptions) {
     dafnyOptions.RunLanguageServer = true;
     dafnyOptions.Set(DafnyConsolePrinter.ShowSnippets, true);
+
     dafnyOptions.PrintIncludesMode = DafnyOptions.IncludesModes.None;
+    // TODO This may be subject to change. See Microsoft.Boogie.Counterexample
+    //      A dash means write to the textwriter instead of a file.
+    // https://github.com/boogie-org/boogie/blob/b03dd2e4d5170757006eef94cbb07739ba50dddb/Source/VCGeneration/Couterexample.cs#L217
+    dafnyOptions.ModelViewFile = "-";
```
keyboardDrummer, line 83: Moved from elsewhere

Source/DafnyLanguageServer/Workspace/Documents/CompilationAfterParsing.cs
```diff
@@ -10,64 +10,34 @@
...
-  public virtual VerificationTree GetVerificationTree() {
-    return new DocumentVerificationTree(Program, DocumentIdentifier);
-  }
-
-  protected IEnumerable<Diagnostic> ComputeFileAndIncludesResolutionDiagnostics() {
-    var includeErrorDiagnostics = GetIncludeErrorDiagnostics();
-    return FileResolutionDiagnostics.Concat(includeErrorDiagnostics).Select(d => d.ToLspDiagnostic());
-  }
-
-  private IEnumerable<DafnyDiagnostic> GetIncludeErrorDiagnostics() {
```
keyboardDrummer, line 46: Moved to `NotificationPublisher.PublishImplicitProjectDiagnostics`

Source/DafnyLanguageServer.Test/Synchronization/CloseDocumentTest.cs
```diff
@@ -34,7 +35,14 @@ public class CloseDocumentTest : DafnyLanguageServerTestBase, IAsyncLifetime {
       var documentItem = CreateTestDocument(source);
       await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
       await CloseDocumentAndWaitAsync(documentItem);
-      Assert.Null(await Projects.GetResolvedDocumentAsync(documentItem.Uri));
+      for (int attempt = 0; attempt < 5; attempt++) {
```
alex-chew, line 38: nit: 500ms seems like a long time to allow

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
```diff
@@ -55,19 +128,28 @@ public class ClientBasedLanguageServerTest : DafnyLanguageServerTestBase, IAsync
...
-    var document = (await Projects.GetLastDocumentAsync(documentItem))!;
-    Assert.NotNull(document);
-    Assert.Equal(documentItem.Version, document.Version);
-    Diagnostic[] result;
-    do {
-      result = await diagnosticsReceiver.AwaitNextDiagnosticsAsync(cancellationToken);
-    } while (!document!.AllFileDiagnostics.Select(d => d.ToLspDiagnostic()).SequenceEqual(result));
+    var compilation = (await Projects.GetLastDocumentAsync(documentItem))!;
+    Assert.NotNull(compilation);
+    Assert.True(documentItem.Version <= compilation.Version);
```
alex-chew, line 135: Why is okay to allow `documentItem.Version < compilation.Version`?

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -131,7 +129,7 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
           return information;
         }
         // Ok no assertion here. Maybe a method?
-        if (node.Position.Line == position.Line && node.Uri == state.Uri) {
+        if (node.Position.Line == position.Line && state.Compilation.Project.IsImplicitProject && node.Uri == state.Compilation.Project.Uri) {
```
alex-chew, line 132: Why does/should this depend on whether the compilation is for an implicit project?

Source/DafnyLanguageServer/Language/Symbols/ISymbolResolver.cs
```diff
@@ -3,15 +3,6 @@
 
 namespace Microsoft.Dafny.LanguageServer.Language.Symbols {
   public interface ISymbolResolver {
-    /// <summary>
```
alex-chew, line 6: Why remove this documentation?

Source/DafnyCore/Options/InMemoryDirectoryInfoFromDotNet8.cs
```diff
@@ -0,0 +1,139 @@
+#nullable enable
```
alex-chew, line 1: Why the duplicate file (the other copy being `Source/DafnyLanguageServer/Workspace/InMemoryDirectoryInfoFromDotNet8.cs`)? Does it have to do with this `#nullable enable` pragma?

alex-chew (commented):
Partial review, more to come later.

Source/DafnyLanguageServer.Test/Synchronization/CloseDocumentTest.cs
keyboardDrummer: Does it matter? Some test timeouts are >60s

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
keyboardDrummer: Each time _any_ document in a project changes, the version field of the related ProjectManager is incremented, so that grows strictly faster than the version of any single document part of that project.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
keyboardDrummer: `state.Compilation.Project.Uri` maps to `.dfy` file only for implicit projects. Basically, in this PR, when project mode is on and you're using a project file, all features that depend on the VerificationTree, which is gutter icons and part of hovering, are turned off. This line will change in the future. I don't think we need to leave a todo. I will add tests in a future PR for the behavior that's still missing, and then remove the property `IsImplicitProject` and fix related code.

Source/DafnyLanguageServer/Language/Symbols/ISymbolResolver.cs
keyboardDrummer: I don't see any documentation that adds information beyond the signature, so to me this is just noise. Is there any piece of the documentation you want to keep?

Source/DafnyCore/Options/InMemoryDirectoryInfoFromDotNet8.cs
keyboardDrummer: Oops !

Source/DafnyLanguageServer/Language/Symbols/ISymbolResolver.cs
alex-chew: I think it's helpful to keep at least the summary section, even if the method's purpose seems obvious right now.

alex-chew (commented):
I've reviewed all the files now, though given the size of this PR it would be good for a second reviewer to take a look as well.

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
```diff
@@ -74,37 +79,80 @@ public class NotificationPublisher : INotificationPublisher {
...
+        foreach (var uri in uris) {
+          IEnumerable<Diagnostic> current = currentDiagnostics.GetOrDefault(uri, Enumerable.Empty<Diagnostic>);
+          IEnumerable<Diagnostic> previous = previousStateDiagnostics.GetOrDefault(uri, Enumerable.Empty<Diagnostic>);
+          if (previous.SequenceEqual(current)) {
+            continue;
+          }
+
+          languageServer.TextDocument.PublishDiagnostics(new PublishDiagnosticsParams {
+            Uri = uri,
+            Version = filesystem.GetVersion(uri),
```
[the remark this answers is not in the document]
keyboardDrummer: There's two different versions. There's the file version, one for each file, what is what the client cares about because it will use that to do client side migration of positions. The client decides when to update the file versions. 

There's also the compilation version, which is irrelevant to the client, but something the server can use to disregard it's own outdated notifications. This used to be the same version as the file version since there was only a single file triggering compilations, but now it's a separate integer. Maaaybe we could get rid of it but this way the logic has changed the least.

Source/DafnyLanguageServer/Workspace/ProjectManager.cs
```diff
@@ -2,158 +2,168 @@
...
     }
 
-    var dafnyOptions = DetermineDocumentOptions(options, documentChange.TextDocument.Uri);
+    StartNewCompilation(migratedVerificationTree, lastPublishedState);
+    TriggerVerificationForFile(documentChange.TextDocument.Uri.ToUri());
+  }
+
+  private void StartNewCompilation(VerificationTree? migratedVerificationTree,
+    IdeState lastPublishedState) {
+    version++;
```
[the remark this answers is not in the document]
keyboardDrummer: There are implicit locks around the CRUD methods in LSP, which trigger this StartNewCompilation

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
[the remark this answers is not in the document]
alex-chew: Thanks for the explanation, that makes sense.

Source/DafnyLanguageServer.Test/Synchronization/CloseDocumentTest.cs
alex-chew: I don't think it matters for the sake of CI, but I think setting too high an allowance could matter for catching performance regressions. Not a blocker though.

alex-chew (dismissed):

alex-chew (dismissed):

alex-chew (dismissed):

alex-chew (approved):

alex-chew (approved):

PR #3903, opened 2023-04-24
Documentation of nameonly and default values for formals

davidcok:
Fixes #


docs/DafnyRef/Types.md
```diff
@@ -4495,4 +4495,65 @@ imposed on the body of `ReachableVia` makes sure that, if the
 predicate returns `true`, then every object reference in `p` is as old
 as some object reference in another parameter to the predicate.
 
+## 6.5. Nameonly Formal Parameters and Default-Value Expressions
+
+A formal parametes of a method, constructor in a class, iterator,
+function, or datatype constructor can be declared with an expression
+denoting a _default value_. This makes the parameter _optional_,
+as opposed to _required_. All required parameters must be declared
+before any optional parameters.
```
RustanLeino, line 4504: No, this is no longer true (once https://github.com/dafny-lang/dafny/pull/3864 is merged). A required parameter is allowed to follow an optional parameter, provided the required parameter is `nameonly`. (It may be best to omit the sentence here and explain this below, after `nameonly` has been introduced.)

Actually, L4538 says what needs to be said.

docs/DafnyRef/Types.md
```diff
@@ -4495,4 +4495,65 @@ imposed on the body of `ReachableVia` makes sure that, if the
...
+using names; here, though there are no `nameonly` parameters so the names may be used 
+or the actual arguments may be given in order.
+
+If a function is declared with a `nameonly` formal, then that formal's value must be given with a named assignment.
+For example, a function `ff` declared as
+<!-- %check-resolve -->
+```dafny
+function ff(x: int, nameonly y: int): int
+```
+may be called only using `ff(0, y := 4)`. A `nameonly` formal may also have a default value and thus be optional.
```
RustanLeino, line 4536: "only" here feels a bit confusing. For one, the parameters don't have to be passed in as `0` and `4`. But also, one can call `ff` with `ff(y := 4, x := 0)`.

docs/DafnyRef/Types.md
```diff
@@ -4495,4 +4495,65 @@ imposed on the body of `ReachableVia` makes sure that, if the
...
+
+Any formals after a `nameonly` formal must either be `nameonly` themselves or have default values.
+Otherwise an error will result when an attempt is made to call the function or method with such a formal list.
```
RustanLeino, line 4539: L4539 describes the motivation for L4538, from the language-design perspective. If L4538 is violated, Dafny will give an error for the declaration of the formals, so what happens at a call site is a moot point. I suggest keeping just L4538 and deleting L4539.

docs/DafnyRef/Types.md
```diff
@@ -4495,4 +4495,65 @@ imposed on the body of `ReachableVia` makes sure that, if the
...
+
+The formals of datatype constructors are not required to have names. Such formals may have default values but
+may not be declared `nameonly`. Such nameless formals must be declared before `nameonly` formals.
```
RustanLeino, line 4542: No, a nameless formal is neither allowed to have a default value nor allowed to be declared `nameonly`. Also, missing here is the rule that a nameless formal parameter is not allowed to follow an optional parameter. I suggest replacing the second and third sentences with "A nameless parameter is not allowed to follow a parameter with a default value.".

docs/DafnyRef/Expressions.md
```diff
@@ -1825,6 +1825,21 @@ Method calls, object-allocation calls (`new`), function calls, and
...
+
+A parameter is either required or optional:
+- required: a caller has to supply an argument
+- optional: the parameter has a default value that is used if a caller omits passing a specific argument
+
+The syntax for giving a positional-only (i.e., nameless) parameter does not allow a default-value expression, so a positional-only parameter is always required.
+
+At a call site, positional arguments are not allowed to follow named arguments. Therefore, if `x` is a nameonly parameter, then there is no way to supply the parameters after `x` by position. 
+Thus, any parameter that follows `x` must either by passed by name or have a default value. 
+That is, if a later (in the formal parameter declaration) parameter does not have a default value, it is effectively nameonly.
```
RustanLeino, line 1841: Since this paragraph explains "effectively nameonly", you may want to add a similar paragraph that explains "effectively required" (see https://github.com/dafny-lang/dafny/pull/3864).

RustanLeino (changes requested):

docs/DafnyRef/Types.md
davidcok: Removed the sentence.

docs/DafnyRef/Types.md
```diff
@@ -4495,4 +4495,65 @@ imposed on the body of `ReachableVia` makes sure that, if the
...
+
+For example,
+<!-- %check-resolve %save f.tmp -->
+```dafny
+function f(x: int, y: int := 10): int
+```
+may be called as either
+<!-- %check-resolve %use f.tmp -->
+```dafny
+const i := f(1,2);
```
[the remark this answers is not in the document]
davidcok: OK

docs/DafnyRef/Types.md
```diff
@@ -4495,4 +4495,65 @@ imposed on the body of `ReachableVia` makes sure that, if the
...
+const j := f(1);
+```
+where `f(1)` is equivalent to `f(1,10)` in this case.
```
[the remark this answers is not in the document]
davidcok: OK

docs/DafnyRef/Types.md
```diff
@@ -4495,4 +4495,65 @@ imposed on the body of `ReachableVia` makes sure that, if the
...
+
+Formal parameters may also be declared `nameonly`, in which case a call site
+must explicitly name the formal when providing its actual argument.
```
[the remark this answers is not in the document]
davidcok: Done

docs/DafnyRef/Types.md
davidcok: Right. Fixed.

docs/DafnyRef/Types.md
davidcok: In this code
```
function f(x: int, nameonly y: int, z: int): int

const c := f(0, y:=0)
```
Dafny gives an error on the const declaration, but not on the function.

docs/DafnyRef/Types.md
davidcok: OK, but I would think a default value for a nameless formal to be a perfectly reasonable thing.

docs/DafnyRef/Expressions.md
davidcok: Actually, I found the 'effectively' terminology unhelpful -- better is simply stating and enforcing the rules about which formals can be optional and which can be nameonly.

docs/DafnyRef/Types.md
RustanLeino: Apparently, this has been fixed since Dafny 4.0. We now get an error on the function declaration as well.

docs/DafnyRef/Types.md
RustanLeino: I agree. It would be a perfectly reasonable thing. We could add it in the future.

RustanLeino (approved):

Issue #1983, opened 2022-04-07
Document minimum requirements for compiled code in each language


robin-aws:
Spin-off from #1951. For each target language, we should be very explicit about what minimum version of the language/runtime/etc. the compiled code needs, what third-party libraries are needed (e.g. the `bignumber.js` library), etc. We also need to back this up with CI to ensure we test on this minimal environment.

davidcok:
Note that the CI process includes constructing a new working environment and running compilations in it for each language. This clearly needs maintenance as new languages are added. But it is a check on the dependencies.

robin-aws:
Just to update, #2572 implements one approach that involves running the whole suite twice, once for the latest supported version of each language and once for the oldest supported version.

There are two issues with this:
1. Our CI resource constraints are already an issue, and even just doubling the nightly build cost could actually start to impact development further. Not to mention it is extremely wasteful to run tests that aren't even compiled twice for different compilation target runtimes (credit to @fabiomadge for pointing this out to me).
2. This approach only supports picking exactly two configurations, but .NET specifically needs at least two "oldest" configurations: `netstandard2.0` and `net472` IIRC.

I don't yet see a way to avoid significantly refactoring our test suite to address this.

robin-aws:
With most compiled tests converted to use `%testDafnyForEachCompiler` now (#4115), I can see a clearer path forwards on this:

* We implement a way to only run the subset of integration tests that use that command, and restrict it to only run for a given target language.
* We split the integration tests into shards differently:
  * Define a "target-environment" GitHub actions matrix dimension 
  * For each target language, add multiple `target-environment` values, trying to hit at least the oldest and newest supported versions (e.g. [`java-8`, `java-17`], [`dotnet-netstandard2.0`, `dotnet-net452`, `dotnet-7.0`], ...)
  * Execute the remaining tests (verification only, or NONUNIFORM compilation tests) on their own shard
* Individual shards could still be subdivided using `XUNIT_SHARD` and `XUNIT_SHARD_COUNT` as we do now, if they are disproportionately large

texastony:
Is there any chance this can be prioritized?
For years, we have been asking for Dafny to advertise the details of the supported compiled languages:
i.e.: Dafny's compiled output uses features from JavaScript X, Java Y, Go Z, etc.

atomb:
[...]

PR #3904, opened 2023-04-24
fix: Explicitly define inequality of `multiset`s in Python for better backwards compatibility

### LINKED ISSUE #3874: Python Runtime Error during Multiset Iteration
### Dafny version

4.0.0

```dafny
method Main() {
    var s: multiset<bool> := multiset{true};

    while (s != multiset{}) {
        var x :| x in s;
	s := s[x := 0];
    }
}
```

```code
dafny /compileTarget:py main.dfy && python main-py/main.py
```

### What happened?

Running the output Python produces the following error:
```
Exception: assign-such-that search produced no value (line 5)
```
Running on other backends terminates successfully. 
From running this with prints, it appears that the loop is entered twice which is why we get the assign-such-that exception.
I can reproduce this for the following Python versions: 3.8, 3.9. I can't reproduce it with 3.11.

Linux, Mac

fabiomadge:
Fixes #3874.


robin-aws (approved):
LGTM, thanks!

It would be even better if we could explicitly test against an older Python, but as per https://github.com/dafny-lang/dafny/issues/1983 I don't yet see an easy way to do that.

Issue #3556, opened 2023-02-16
Detect and report mixed use of unicode-char true and false


davidcok:
### Summary

Mixing code that has differing values of --unicode-char will liely not work and should be errored about.

Without detection this probably permits a soundness problem of some kind.

### Background and Motivation

Dafny 4 has a default of unicode-char:true, previous versions had unicode-char:false

### Proposed Feature

Detect whether modules are setting different values -- ideally only where it actually matters.

atomb:
I updated the text to generalize it a bit to any option that changes language semantics.

robin-aws:
Note that although a URL pointing to a `*.doo` file is technically enough for a remote dependency, I still wouldn't recommend that as a production-grade solution, since GitHub releases don't have any standards for including checksums or signatures. This makes building on top of existing systems like Maven/Gradle via plugins like #3169 appealing: https://docs.gradle.org/current/userguide/dependency_verification.html

robin-aws:
Added `breaking-change` since actually blocking the unsafe library usage by default, and hence actually closing the soundness issue, is breaking.

robin-aws:
We actually already made the breaking change to warn against non-doo files in `--library` (since warnings now stop the pipeline by default).

The remaining unsafe feature is including source files without verifying them (i.e. without `--verify-included-files`), which we still have to at least warn against to close this soundness issue.

PR #3913, opened 2023-04-27
feat: Build artifacts ("library" backend and .doo files)

robin-aws:
Provides a build artifact for outputting after successful verification, with enough metadata attached to support safe consumption as a library, as loosely described in this comment: https://github.com/dafny-lang/dafny/issues/3556#issuecomment-1513871567

Also refactors the functionality used by `dafny audit` to ensure the library backend rejects `assume` statements as the other backends do, without having to implement a proper SinglePassCompiler subclass.

Also hides the internal "Simple Dafny" backend, which isn't intended to be publicly documented, while the new library backend is.

Provides a solution for #3556, but I'm not calling that resolved until we at least deprecate separate verification features that don't use .doo files (which isn't in scope for this PR).

TODO:

- [x] Clean up doo file generation (generic Backend/Compiler interface is creating an extra unused file entry)
- [x] Eliminate useless `LibraryCompiler` if possible without rathole-ing on the general interface design
- [ ] ~Attempt asserting that `DooFile.ProgramSerializationOptions` doesn't change over time.~ (too tricky given options are only initialized for the current command)
- [x] Bit more testing
- [x] Address change in behavior in auditor
- [x] Reference manual updates
- [x] Release note


robin-aws:
Note the failing `TestAuditor.dfy` test: it's due to my filling in more cases of `Node.Assumptions()`. I think my changes improve the accuracy of that method so I'd prefer to keep them, but the `AuditReport` logic has to be adjusted a little to account for that to maintain current behavior. It would be simpler if we ditched the behavior of only reporting on explicit assumptions if there are any in a given "scope", since getting the scope right is now a bit trickier.

Source/DafnyCore/DooFile.cs
```diff
@@ -0,0 +1,238 @@
...
+  
+  public ManifestData Manifest { get; set; }
+  
+  public string ProgramText { get; set; }
+  
+  // This must be independent from any user-provided options,
+  // and remain fixed over the lifetime of a single .doo file format version.
+  // We don't want to attempt to read the program text using --function-syntax:3 for example.
+  // If we change default option values in future Dafny major version bumps,
+  // this must be configured to stay the same.
```
atomb, line 71: I wonder whether we can do something to enforce that.

Source/DafnyCore/DooFile.cs
```diff
@@ -0,0 +1,238 @@
...
+      
+      Options = new Dictionary<string, object>();
+      foreach (var (option, check) in OptionChecks) {
+        options.Options.OptionArguments.TryGetValue(option, out var optionValue);
+        Options.Add(option.Name, optionValue);
+      }
+    }
+
+    public static ManifestData Read(TextReader reader) {
+      return Toml.ToModel<ManifestData>(reader.ReadToEnd(), null, new TomlModelOptions());
```
atomb, line 55: Does this check that the text being read only contains entries that fit in `ManifestData`?

Test/separate-verification/assumptions.dfy.expect
```diff
@@ -0,0 +1,21 @@
...
+TestAuditor.dfy(94,4): Warning: note, this loop has no body (loop frame: i)
+TestAuditor.dfy(85,10): Warning: /!\ No terms found to trigger on.
+TestAuditor.dfy(87,4): Warning: /!\ No terms found to trigger on.
+
+Dafny program verifier did not attempt verification
+assumptions.dfy(6,8): Error: the included file TestAuditor.dfy contains error(s)
+TestAuditor.dfy(18,6): Error: Ghost declaration has no body and has an ensures clause.
+TestAuditor.dfy(23,15): Error: Ghost declaration has no body and has an ensures clause.
+TestAuditor.dfy(41,17): Error: Compiled declaration has no body and has an ensures clause.
+TestAuditor.dfy(41,17): Error: Declaration with [{:extern}] has a ensures clause.
```
atomb, line 11: I'm on the fence about this one. Do you think the standard for published libraries should be that they include only `bare` externs, and use something like `Need` to concretely check? Or should we instead allow contracts on externs but require that they're (automatically) tested, with `--test-assumptions Externs`?

atomb (changes requested):
I like this! I made a few suggestions, and had a few questions, but I'm keen to get this merged soon.

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -297,8 +298,55 @@ public enum TestAssumptionsMode {
...
+      QuantifierSyntax,
+      SolverLog,
+      SolverPath,
+      SolverPlugin,
+      SpillTranslation,
+      StdIn,
+      TestAssumptions,
+      WarnShadowing,
+      ManualLemmaInduction,
+      RelaxDefiniteAssignment,
```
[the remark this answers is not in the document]
robin-aws: Yeah and I had the same debate about `EnforceDeterminism`: it's not actually unsound but does feel like the kind of things users will want to apply transitively so I included it. It would be more consistent to include this too, and including an option is a two-way door whereas leaving one out is a one-way door (at least fixing it would likely require rebuilding artifacts, if not an artifact version bump).

Source/DafnyCore/Compilers/Library/LibraryBackend.cs
```diff
@@ -0,0 +1,89 @@
...
+
+  public override IReadOnlySet<string> SupportedExtensions => new HashSet<string> {};
+
+  public override string TargetName => "Dafny Library (.doo)";
+
+  public override string TargetExtension => "doo";
+  public override string TargetId => "lib";
+
+  public override string TargetBaseDir(string dafnyProgramName) =>
+    $"{Path.GetFileNameWithoutExtension(dafnyProgramName)}-lib";
```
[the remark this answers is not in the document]
robin-aws: Yeah, the cleanest way to model this was to imitate how Java works, for example: blurt the entry files to a directory, and then rely on `CompileTargetProgram` to create the jar. It makes debugging a little easier too.

I played around with loosening the `IExecutableBackend` assumption that a backend produces textual output of some kind, but it was a real rathole. If and when we want to serialize a program to a non-textual format in the future, we should probably do that first.

Source/DafnyCore/Compilers/Library/LibraryBackend.cs
```diff
@@ -0,0 +1,89 @@
...
+  }
+  
+  public override bool CompileTargetProgram(string dafnyProgramName, string targetProgramText, string callToMain, string targetFilename,
+    ReadOnlyCollection<string> otherFileNames, bool runAfterCompile, TextWriter outputWriter, out object compilationResult) {
+    
+    var targetDirectory = Path.GetFullPath(Path.GetDirectoryName(targetFilename));
+    var dooPath = DooFilePath(dafnyProgramName);
+    
+    File.Delete(dooPath);
+    ZipFile.CreateFromDirectory(targetDirectory, dooPath);
```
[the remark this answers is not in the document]
robin-aws: Leaves it around, which is again consistent with the other backends even though I'd prefer it didn't as per my other comment.

Test/separate-verification/assumptions.dfy.expect
robin-aws: I was aiming to default to the safer option for now. It'll let us publish `FileIO` from `dafny-lang/libraries` for example, but not `MutableMap`. I'd like to allow them too, but I'm not yet sure how to enforce at the time of building with `-t:lib` that you ran with `--test-assumptions Externs` "at some point".

Plus it's probably helpful to distinguish in some way as a consumer if you're consuming a "fully verified" library or not. Perhaps you can run the auditor in such a way that libraries are in scope too?

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -277,7 +282,8 @@ public enum CommandLineArgumentsResult {
         var nameToShow = useRelative ? relative : file;
         var supportedExtensions = options.Backend.SupportedExtensions;
         if (supportedExtensions.Contains(extension)) {
-          // .h files are not part of the build, they are just emitted as includes
+          // .h files are not This should be delegated to the backend insteadpart of the build, they are just emitted as includes
```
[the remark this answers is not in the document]
robin-aws: A wild copy and paste appeared!

Source/DafnyCore/DafnyMain.cs
```diff
@@ -78,9 +86,32 @@ public class DafnyFile {
...
+        if (!dooFile.Validate(filePathForErrors, options)) {
+          throw new IllegalDafnyFile(true);
+        }
+        
+        // For now it's simpler to let the rest of the pipeline parse the
+        // program text back into the AST representation.
+        // At some point we'll likely want to serialize a program
+        // more efficiently inside a .doo file, at which point
+        // the DooFile class should encapsulate the serialization logic better
+        // and expose a Program instead of the program text.
```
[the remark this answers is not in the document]
robin-aws: I also attempted to make a `DooFile` extend `DafnyFile` and add a `Parse` function, and that was kind of working nicely but also a big rathole, so punting on that for now too.

Source/DafnyCore.Test/DooFileTest.cs
```diff
@@ -0,0 +1,33 @@
...
+    dooFile.Write(path);
+    var loadedDooFile = DooFile.Read(path);
+    
+    Assert.Equal(loadedDooFile.Manifest.DooFileVersion, DooFile.ManifestData.CurrentDooFileVersion);
+    Assert.Equal(loadedDooFile.Manifest.DafnyVersion, options.VersionNumber);
+    Assert.Equal(loadedDooFile.Manifest.SolverIdentifier, options.SolverIdentifier);
+    Assert.Equal(loadedDooFile.Manifest.SolverVersion, options.SolverVersion.ToString());
+  }
+
+  private static Program ParseProgram(string dafnyProgramText, DafnyOptions options) {
```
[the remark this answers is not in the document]
robin-aws: Agreed, I mostly copied it from another test package. Perhaps we could have other test packages depend on this one in the future.

Source/DafnyCore/AST/Grammar/Printer.cs
```diff
@@ -399,10 +404,14 @@ public class Printer {
             wr.WriteLine("}");
           }
 
-        } else if (d is ModuleDecl) {
+        } else if (d is ModuleDecl md) {
           wr.WriteLine();
           Indent(indent);
           if (d is LiteralModuleDecl modDecl) {
+            if (printMode == PrintModes.DllEmbed && !modDecl.ModuleDef.IsToBeCompiled) {
```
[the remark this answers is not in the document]
robin-aws: I renamed the `DllEmbed` mode and added a comment.

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -756,7 +763,9 @@ private record TargetPaths(string Directory, string Filename) {
 
     static void WriteDafnyProgramToFiles(DafnyOptions options, TargetPaths paths, bool targetProgramHasErrors, string targetProgramText,
       string/*?*/ callToMain, Dictionary<string, string> otherFiles, TextWriter outputWriter) {
-      WriteFile(paths.Filename, targetProgramText, callToMain);
+      if (targetProgramText.Length != 0) {
```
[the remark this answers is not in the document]
robin-aws: The backend has no control over that unfortunately, the best it can do is not write anything to the `ConcreteSyntaxTree` its handed by the `Compile` method.

Source/DafnyCore/DooFile.cs
robin-aws: Yup! Adding a quick unit test for this though. :)

Source/DafnyCore/DooFile.cs
robin-aws: We could conceivably have a serialized form of DafnyOptions to check against in a test, but it might be another rathole to have the Toml encoding work for all 67 options :) I'll timebox giving it a shot.

atomb: Yeah, probably not urgent. :)

Test/separate-verification/assumptions.dfy.expect
atomb: Yeah, I think being stricter at first and then possibly making thing more lenient in the future is a good strategy.

atomb: Yeah, Haskell has a notion of "safe" modules, which have more restricted functionality (IIRC, this include no FFI, no unsafe things, etc.) and can only import other safe modules. We could imagine having different categories of libraries, and the first one we implement could be the strictest.

Source/DafnyCore/AST/Grammar/Printer.cs
```diff
@@ -408,7 +408,10 @@ public class Printer {
           wr.WriteLine();
           Indent(indent);
           if (d is LiteralModuleDecl modDecl) {
-            if (printMode == PrintModes.DllEmbed && !modDecl.ModuleDef.IsToBeCompiled) {
+            if (printMode == PrintModes.Serialization && !modDecl.ModuleDef.IsToBeCompiled) {
```
atomb, line 411: Does this mean that we wouldn't support proof-only libraries?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -333,6 +333,9 @@ public enum IncludesModes {
     public bool AuditProgram = false;
 
     public static string DefaultZ3Version = "4.12.1";
+    // Not directly user-configurable, only recorded once we discover it
+    public string SolverIdentifier { get; private set; }
```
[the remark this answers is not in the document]
robin-aws: It was added in https://github.com/dafny-lang/dafny/pull/3663/files#diff-ba72c3c1efed223d4216f91a154cd44c8e31134d76911321a965d27db8eeb8acR1119, and it looks like it was just a case of "hmm getting the version is relatively expensive, we should probably cache this safely in case it's called a lot". So I'm not too worried about a multi-solver workflow showing up, but I'll make `SetZ3Options` throw if invoked multiple times just in case.

Test/separate-verification/assumptions.dfy.expect
robin-aws: I actually anticipate needing an exact match on the Dafny version is going to be the biggest initial pain.

Source/DafnyCore/AST/Grammar/Printer.cs
robin-aws: Oh sure we will - the `IsToBeCompiled` field really just means "was not provided by `--library`". So you can work just with `.doo` files and never compile any of them with the other backends. Perhaps the name could be "IsToBeBuilt" instead, but I don't think that's enough of an improvement to bother with personally. :)

atomb: Ah, right. Sounds good!

atomb (approved):
I think this looks good. CI will ensure that we can't merge without fixing that last test. :)

fabiomadge (approved):

Source/DafnyCore/Compilers/Library/LibraryBackend.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using DafnyCore;
+
+namespace Microsoft.Dafny.Compilers; 
+
+public class LibraryBackend : ExecutableBackend {
```
keyboardDrummer, line 11: Hey, you don't have to inherit from SinglePassCompiler, that's convenient 😄

docs/DafnyRef/UserGuide.md
```diff
@@ -93,10 +93,43 @@ However, which files are verified, built, run, or processed by other
...
+a fully-verified program along with metadata about how it was verified.
+Such files use the extension `.doo`, for Dafny Output Object,
+and can be used as input anywhere a `.dfy` file can be.
+
+`.doo` files are produced by an additional backend called the "Dafny Library" backend,
+identified with the name `lib` on the command line. For example, to build multiple
+Dafny files into a single build artifact for shared reuse, the command would look something like:
+
+```bash
+dafny build -t:lib A.dfy B.dfy C.dfy --out MyLib.doo
```
keyboardDrummer, line 111: I think Dafny projects should specify whether they're an executable or a library, and depending on this, the _default_ for `dafny build` should produce an executable or a `*.doo` file.

robin-aws: > I think Dafny projects should specify whether they're an executable or a library, and depending on this, the _default_ for `dafny build` should produce an executable or a `*.doo` file.
> 
> Also, I think the "lib" target should not exist for `dafny run` and `dafny translate`. Currently, it's not mentioned in the `--help` documentation for those (and for build).
> 
> I think the ideal end-game is that Dafny's standard library is large enough that `dafny run` and `dafny build` can generally be used without supplying additional non-Dafny files that implement externs, and that `dafny translate` is used for integration scenarios. If that's the case, the `--target` option for `dafny run` and `build` would generally not be used, and `*.doo` files would be generated when doing `dafny build` on a Dafny project that's marked as a library.
> 
> For now, could you show an error when using the `lib` target with `run`, and have `lib` be the default for projects without a main method?

You make some good points, but I think the real issue is us using "lib" as an identifier for the ".doo target", because it implies the result is a library and not an executable, when that aspect should be orthogonal to which backend you're targeting. We could just call it "the doo backend": I didn't love that, but also couldn't think of a better label.

I'd much rather have the .doo backend support all operations, so that it's easier to get better coverage through generic utilities like `%testDafnyForEachCompiler`. Hopefully changing the label would make that seem more reasonable.

> Currently I get:
> 
> ```
> rwillems@bcd0745419f2 dafny2 % ./Scripts/dafny run -t:lib /Users/rwillems/SourceCode/dafny2/Test/c++/hello.dfy                
> 
> Dafny program verifier finished with 0 verified, 0 errors
> Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
>    at DafnyCore.DooFile.Validate(String filePath, DafnyOptions options, Command currentCommand) in /Users/rwillems/SourceCode/dafny2/Source/DafnyCore/DooFile.cs:line 122
>    at Microsoft.Dafny.DafnyFile..ctor(DafnyOptions options, String filePath, Boolean useStdin) in /Users/rwillems/SourceCode/dafny2/Source/DafnyCore/DafnyMain.cs:line 99
>    at Microsoft.Dafny.DafnyDriver.ProcessCommandLineArguments(String[] args, DafnyOptions& options, List`1& dafnyFiles, List`1& otherFiles) in /Users/rwillems/SourceCode/dafny2/Source/DafnyDriver/DafnyDriver.cs:line 263
>    at Microsoft.Dafny.DafnyDriver.ThreadMain(String[] args) in /Users/rwillems/SourceCode/dafny2/Source/DafnyDriver/DafnyDriver.cs:line 112
>    at Microsoft.Dafny.DafnyDriver.<>c__DisplayClass10_0.<Main>b__0() in /Users/rwillems/SourceCode/dafny2/Source/DafnyDriver/DafnyDriver.cs:line 102
>    at System.Threading.Thread.StartCallback()
>  Process exited with exit code 134
> ```
> 
> and
> 
> ```
> rwillems@bcd0745419f2 dafny2 % ./Scripts/dafny translate lib /Users/rwillems/SourceCode/dafny2/Test/c++/hello.dfy 
> 
> Dafny program verifier finished with 0 verified, 0 errors
> ```

The first case is a legit bug which I've got a fix for, thanks for finding that: https://github.com/dafny-lang/dafny/pull/3967

The second case is working as advertised AFAICT: it should have created a `hello-lib` directory with the entries for a doo file (which I don't love as discussed above, but I didn't think it was actively harmful, and I found it handy for debugging).

keyboardDrummer: OK, I like your view of things. We should change the `TargetId` `lib` into `doo` or `dfy` though.

keyboardDrummer: Based on offline discussions with @robin-aws , created this issue: https://github.com/dafny-lang/dafny/issues/3976

PR #3919, opened 2023-04-28
feat: Semantics for dots in `{:extern}` module names in Python

### LINKED ISSUE #2963: Nested extern attribute creates stub with incorrect file name in Python
Follow-up from #2952.

After #2952, ` module {:extern "A.B.C"} A.B.C` works in Python, but creates an unnecessary and confusing stub named `A.B.C.py` (as opposed to where you actually implement the extern, `A/B/C.py`). As clean up we should have the Python code generation not create this stub.

fabiomadge:
Fixes #2963.


Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -866,7 +866,7 @@ private record TargetPaths(string Directory, string Filename) {
         var callToMainTree = new ConcreteSyntaxTree();
         string baseName = Path.GetFileNameWithoutExtension(dafnyProgramName);
         compiler.EmitCallToMain(mainMethod, baseName, callToMainTree);
-        callToMain = callToMainTree.ToString(); // assume there aren't multiple files just to call main
+        callToMain = callToMainTree.MakeString(compiler.TargetIndentSize); // assume there aren't multiple files just to call main
```
robin-aws, line 869: 👍

robin-aws (commented):
Definitely an improvement and the implementation LGTM. Does this fix #2963?

How about an update to `docs/DafnyRef/integration-py/IntegrationPython.md`?

What happens if you also include an `Nested.Library.Child` module? Is it impossible since then we'd want `Nested/Library/__init__.py` instead?

fabiomadge:
Yes, this only allows for definitions at the leafs, which seems to be the way to structure modules in Python anyway.

robin-aws (approved):

PR #3925, opened 2023-04-29
Changing proof stability language

davidcok:


docs/DafnyRef/UserGuide.md
```diff
@@ -1225,15 +1225,16 @@ proof attempt multiple times after automatically making minor changes to
...
+#### 13.6.5.1. Measuring proof variability
 
-To measure the stability of your proofs, start by using the
+To measure the variability of your proofs, start by using the
 `-randomSeedIterations:N` flag to instruct Dafny to attempt each proof
 goal `N` times, using a different random seed each time. The random seed
 used for each attempt is derived from the global random seed `S`
-specified with `-randomSeed:S`, which defaults to `0`.
+specified with `-randomSeed:S`, which defaults to `0` (which means use
+an arbitrary -- e.g. clock-based -- seed).
```
atomb, line 1237: I don't think this is true. At least on Dafny's side of the randomization process, `0` is the actual seed used, which helps ensure deterministic results. Dafny then sends this seed to Z3, as well, but I don't think it interprets `0` specially (though I could be wrong about that).

atomb (changes requested):
The changes look just right, but I don't think the additional comment about a clock-based seed is actually true.

davidcok:
The SMT-LIB standard (http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf <http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf>) states that 0 means to choose an arbitrary starting seed, whereas non-zero is used to tell the solver to be deterministic (or at least to reduce non-determinism). My understanding and experience is that Z3 is compliant to SMT in this regard.

atomb:
> The SMT-LIB standard (http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf <http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf>) states that 0 means to choose an arbitrary starting seed, whereas non-zero is used to tell the solver to be deterministic (or at least to reduce non-determinism). My understanding and experience is that Z3 is compliant to SMT in this regard.

Good to know! I wonder whether we should modify some of the code so that both Dafny and Z3 are either deterministic (or at at least more deterministic) or non-deterministic under the same circumstances as each other.

atomb (approved):

PR #3927, opened 2023-04-29
Trial fix for issue 1352

### LINKED ISSUE #1352: `Expect` not allowed in function methods?
Dafny 3.2.0.30713 (b96750a1cbf4a084ed61b345ff68b42dca13759e)

I'm following the [Expect example in the Dafny Documentation](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-expect-statement), but changing the methods for function methods. Expect then causes an error.
The error message doesn't seem to make sense in the context.

example.dfy:
```dafny
function method {:external} fm(b: int): int

predicate method pm(a: int)
{
    true
}

function method fm'(n: int): int
ensures pm(n)
{
    var v := fm(n);
    //assert pm(v);   // OK
    //assume pm(v);   // OK
    expect pm(v);   // error
    v
}
```

```bash
dafny example.dfy 
XXX/example.dfy(14,4): Error: expect statement is not allowed in this context (because this is a ghost method or because the statement is guarded by a specification-only expression)
1 resolution/type errors detected in example.dfy
```

davidcok:
Fixes #1352


davidcok:
Abandoning this PR, per the thoughts that Leino recorded in Issue #1352.

Note that the work so far stopped the error message for expect in functions, but was stalled at the major work needed to actually compile expect in functions.

PR #3933, opened 2023-05-01
Feat: `assert {:only}` to temporarily work on a single assertion

### LINKED ISSUE #3095: [Feature request]: `{:only}` on assertions

MikaelMayer:
This PR fixes #3095 by introducing `assert {:only}`. As stated in the documentation:

`assert {:only} X;` temporarily transforms all other non-`{:only}` assertions in the surrounding declaration into assumptions.

```dafny
method Test() {
  assert true;                  // Unchecked
  assert {:only} true by {      // Checked
    assert true;                // Checked
  }
  assert true;                  // Unchecked
  assert {:only "after"} true;  // Checked
  assert true;                  // Checked
  assert {:only "before"} true; // Checked
  assert true;                  // Unchecked
}
```

`{:only}` can help focusing on a particular proof or a particular branch, as it transforms not only other explicit assertions, but also other implicit assertions, and call requirements, into assumptions.
Since it's meant to be a temporary construct, it always emits a warning.
It also has two variants `assert {:only "before"}` and `assert {:only "after"}`.
Here is precisely how Dafny determines what to verify or not.
`{:only}` annotations define "verification interval" which are visual:

* `assert {:only} X [by {...} | ;]` sets a verification interval that starts at the keyword `assert` and ends either at the end of the proof `}` or the semicolon `;`, depending on which variant of `assert` is being used.
* `assert {:only} ...` inside an other verification interval removes that verification interval and sets a new one.
* `assert {:only "before"} ...` inside another verification interval finishes that verification interval earlier at the end of this assertion. Outside verification intervals, it sets a verification interval from the beginning of the declaration to the end of this assertion, but only if there were no other verification intervals before.
* `assert {:only "after"} ...` inside another verification interval moves the start of that verification interval to the start of this new assert. Outside verification interval, it sets a verification interval from the beginning of this `assert` to the end of the declaration.

The start of an asserted expression is used to determines if it's inside a verification interval or not.
For example, in `assert B ==> (assert {:only "after"} true; C)`, `C` is actually the start of the asserted expression, so it is verified because it's after `assert {:only "after"} true`.

As soon as a declaration contains one `assert {:only}`, none of the postconditions are verified; you'd need to explicit them with assertions if you wanted to verify them at the same time.

## Integration

* I updated the auditor to report usages of `assert {:only}`
* I updated the doc
* I added a code action to suggest to remove the `{:only ...}` in VSCode.

## Alternatives considered

It's currently possible to use ` {:selective_checking true}` on a declaration and and `{:start_checking_there}` on an assertion to "comment out" all previous assertions.
However, this mechanism is not suitable for us, even transpiling our attributes to that, for the following reasons:. It works at the Boogie level, but at the Dafny level it's often misleading. For example, `{:start_checking_there}` on an `assert {:start_checking_there} X by { Y }` will actually start checking X but not the proof Y because in Boogie the proof is written before the assertion.


Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -1353,6 +1355,49 @@ public partial class Translator {
       builder.Add(Assert(tok, q, new PODesc.TraitFrame(m.WhatKind, true), kv));
     }
 
+    // Return a way to know if an assertion should be converted to an assumptions
+    private void SetAssertionOnlyFilter(Node m) {
+      List<RangeToken> rangesOnly = new List<RangeToken>();
+      m.Visit(node => {
+        if (node is AssertStmt assertStmt && Attributes.Find(assertStmt.Attributes, "only") is UserSuppliedAttributes attribute) {
+          var (isBefore, isAfter) = attribute.Args.Count == 1 && attribute.Args[0] is LiteralExpr { Value: var value } &&
+                                  value is "after" or "before" ? (value is "before", value is "after") : (false, false);
```
atomb, line 1364: If it were me, I'd move the logic for extracting the `{:only}` attribute to a separate definition. Then maybe you'd have a datatype `AssertOnlyKind = After | Before | Only` and a `bool HasAssertOnlyAttribute(out var assertOnlyKind)`, or something like that.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -1353,6 +1355,49 @@ public partial class Translator {
       builder.Add(Assert(tok, q, new PODesc.TraitFrame(m.WhatKind, true), kv));
     }
 
+    // Return a way to know if an assertion should be converted to an assumptions
```
atomb, line 1358: The notion of "before" and "after" here seems potentially confusing. For instance, if you were to say `assert {:only "before"} P;` in the `else` branch of an `if` statement, it would enable assertions in the other branch of the `if` statement. It's a simple notion, though, so perhaps the strangeness of it isn't fatal.

atomb (changes requested):
This is a feature I expect I'll use frequently. :) I had a few nits about wording, and a broader comment about whether the notion of "range" used here is the right one. It might be, but I'm not yet 100% convinced.

Source/DafnyCore/Generic/ErrorRegistry.cs
```diff
@@ -26,6 +26,7 @@ public record DafnyAction(string Title, IReadOnlyList<DafnyCodeActionEdit> Edits
 
 public static class ErrorRegistry {
 
+  // Replace any NoneId by ParseErrors.p_... or ResolutionErrors.r_
```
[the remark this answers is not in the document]
MikaelMayer: I came across warnings and errors which were using ErrorRegistry.NoneId as identifier. However, at first, I did not know how to put an identifier that has some meaning, so I had to do research and found out that meaningful identifiers are actually defined in ParseErrors.p_.... so I created the same for ResolutionErrors.r_.... and put a comment so that if a resolution error is using ErrorRegistry.NoneId and someone wants to create a rich error (e.g. that contains extended error text or a code action), they know where to go and what to do.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
MikaelMayer: Yeah I thought about that too. The advantage is that the notions are straightforward indeed for now. I primarily envision users to use `{:only}` by itself for a single proof, and if they really want, they can delimit without surprise which assertions should be verified using `{:only "before"}` and `{:only "after"}`. It's just that I had to give meaning to single instances of  `{:only "before"}` and  `{:only "after"}`.

I also thought of a more coarse-grained numerical notion of "level", which is one that @jtristan suggested one day. For example, `assert {:only 0}` will be the same as `assert {:only}`, whereas `assert {:only 1}` will define a verification interval on the surrounding block, `assert {:only 2}` on the block containing that block, etc. Because we traverse the program using a visitor pattern and nodes, this would be quite easy to implement.
If you like this idea, that would make three of us to like it so I would happily implement it as part of this PR.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -67,8 +67,14 @@ public record AssumptionDescription(string issue, string mitigation, bool isExpl
       isExplicit: false,
       allowedInLibraries: false);
   }
-}
 
+  public static AssumptionDescription AssertOnly = new(
```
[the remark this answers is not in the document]
atomb: Oh, right! Sometimes helpful things aren't as helpful. :)

Source/DafnyCore/Generic/ErrorRegistry.cs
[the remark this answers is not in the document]
atomb: Got it.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
atomb: Yeah, that makes sense. And adding those would be backward-compatible, so no need to hold up this PR for it.

atomb (approved):
Thanks for implementing this!

PR #3934, opened 2023-05-01
First draft of proof optimization guide

atomb:
Adds a new document describing techniques for optimizing proofs to reduce resource use and variability.

This in some ways overlaps with the sections in the manual about fixing failing proofs, optimizing slow proofs, and reducing variability. I'd ultimately like it to supersede those, once we ensure that we're not losing material that way. Those DafnyRef sections can be replaced by a link to this new document.


docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,597 @@
...
+
+Dafny provides several attributes that tell it to verify certain assertions separately, rather than verifying everything about the correctness of a particular definition in one query to the solver.
+
+* The `{:split_here}` attribute on an `assert` statement tells Dafny to verify all assertions leading to this point in one batch and all assertions after this point (or up to the next `{:split_here}`, if there is one) in one batch.
+
+* The `{:focus}` attribute on an `assert` or `assume` statement tells Dafny to verify all assertions in the containing block, and everything that _always_ follows it, in one batch, and the rest of the definition in another batch (potentially subject to further splitting due to other occurrences of `{:focus}`).
+
+* The `{:vcs_split_on_every_assert}` attribute on a definition tells Dafny to verify each assertion in that definition in its own batch. You can think of this as being similar to having many `{:focus}` or `{:split_here}` occurrences, including on assertions that arise implicitly, such as preconditions of calls or side conditions on partial operations.
```
MikaelMayer, line 77:
```suggestion
* The [`{:vcs_split_on_every_assert}`](../DafnyRef/DafnyRef#sec-vcs_split_on_every_assert) attribute on a definition tells Dafny to verify each assertion in that definition in its own batch. You can think of this as being similar to having many [`{:focus}`](../DafnyRef/DafnyRef#sec-focus) or [`{:split_here}`](../DafnyRef/DafnyRef#sec-split_here) occurrences, including on assertions that arise implicitly, such as preconditions of calls or side conditions on partial operations.
```

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,597 @@
...
+    ensures x * z == y * z
+  {
+    LemmaMulEquality(x, y, z);
+  }
+}
+```
+
+There was another instance of code in the repository where a proof began to fail after upgrading the version of Z3 included with Dafny. That code originally used `LemmaMulEqualityAuto`. We [updated it](https://github.com/dafny-lang/libraries/blob/7c386fa0b4a267715f9bd49948d1af68e1631e6b/src/dafny/Collections/LittleEndianNat.dfy#L532) to call `LemmaMulEquality` with the appropriate parameters and the proof reliably succeeded.
+
+In general, creating and using explicit lemmas like `LemmaMulEquality`, and skipping lemmas like `LemmaMulEqualityAuto` altogether, tends to lead to proofs with less variability.
```
MikaelMayer, line 563: Indeed I've seen that too!

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,597 @@
...
+
+As we'll describe in more detail later, however, it can sometimes be better to avoid the quantifiers in the first place than to help Dafny instantiate them.
+
+# Hiding information
+
+There are several ways to structure a Dafny program to limit the number of facts that are available in scope at any particular point. First we'll go over a few that can be applied as small changes to existing programs, and then we'll cover some that are better applied early in the development process (though they can be applied to existing programs with more significant refactoring).
+
+## Subsumption
+
+The default behavior of an `assert` statement is to instruct the verifier to prove that a predicate is always true and then to assume that same predicate is true in later portions of the code, making it available to help prove other facts. This is called _subsumption_. On occasion, however, you may have an `assert` statement that exists to show a desired condition for its own sake, and that does not help prove later facts. In this case, you can disable subsumption, using `assert {:subsumption 0}`, to reduce the number of facts in scope in subsequent code.
```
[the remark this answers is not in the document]
atomb: They are mentioned below (in the section "Opaque specification clauses", but perhaps it would make sense to move the "Subsumption" section after that one?

MikaelMayer (dismissed):
This first version of proof optimization is good enough!
Would you please add some links from the documentation to it?

atomb:
> This first version of proof optimization is good enough! Would you please add some links from the documentation to it?

I added some links. At some point we should make a more complete editing pass to reduce redundancy, but I think that can wait until a future PR.

MikaelMayer (dismissed):
Perfect! That's a well-integrated topic now.

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,599 @@
...
+ProveSomeArithmetic (correctness) (assertion batch 9),Passed,00:00:00.0708400,67341
+ProveSomeArithmetic (correctness) (assertion batch 8),Passed,00:00:00.0644100,64792
+ProveSomeArithmetic (correctness) (assertion batch 7),Passed,00:00:00.0622940,65952
+```
+
+In all of these output formats, you can see that batch 5, corresponding to the assertion about conversion to and from bit vectors, is the most expensive to prove. Later sections will go into why this is the case, and how to make proofs involving bit vectors more straightforward.
+
+# Identifying highly variable assertions
+
+Sometimes, verifying a particular assertion is fast, initially, but becomes very slow (or fails) after a small change. We refer to this phenomenon as _verification variability_. The techniques in this document can help to fix proofs when it occurs. However, sometimes it would be more helpful to predict when verification of your program is highly variable, and adapt it proactively, when you can intentionally set aside time, rather than needing to fix it retroactively.
```
robin-aws, line 161:
```suggestion
Sometimes, verifying a particular assertion succeeds, initially, but fails after a small change. We refer to this phenomenon as _verification variability_. The techniques in this document can help to fix proofs when it occurs. However, sometimes it would be more helpful to predict when verification of your program is likely to be highly variable, and adapt it proactively, when you can intentionally set aside time, rather than needing to fix it retroactively.
```

Often variable "proofs" are already slow to verify.

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,599 @@
...
+
+If you have a difficult non-linear fact to prove, and proving it from the ground up is infeasible, it is possible, as a last resort, to adjust how Z3 attempts to prove arithmetic formulas. It contains a parameter, `smt.arith.solver`, that allows you to choose between one of six different arithmetic decision procedures. Dafny has historically used solver "2", and continues to do so by default. This peforms most effectively on the corpus of examples that we have experimented with. However, solver "6", which is the default for Z3 when not being invoked from Dafny, performs better at non-linear arithmetic. Currently, it's possible to choose this solver using the Dafny flag `--boogie /proverOpt:O:smt.arith.solver=6` (or just `/proverOpt:O:smt.arith.solver=6`, when not using top-level commands).
+
+## Relationships between integers and bit vectors
+
+Dafny contains two distinct ways to describe what might be stored in a machine word in other programming languages.
+
+* The `int` type denotes the unbounded mathematical integers.
```
robin-aws, line 255: Perhaps mention that in practice you'd use a bounded `newtype` and not `int` directly, especially since then you really are directly equivalent with the bit vector option.

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,599 @@
...
+    LemmaMulEquality(x, y, z);
+  }
+}
+```
+
+There was another instance of code in the repository where a proof began to fail after upgrading the version of Z3 included with Dafny. That code originally used `LemmaMulEqualityAuto`. We [updated it](https://github.com/dafny-lang/libraries/blob/7c386fa0b4a267715f9bd49948d1af68e1631e6b/src/dafny/Collections/LittleEndianNat.dfy#L532) to call `LemmaMulEquality` with the appropriate parameters and the proof reliably succeeded.
+
+In general, creating and using explicit lemmas like `LemmaMulEquality`, and skipping lemmas like `LemmaMulEqualityAuto` altogether, tends to lead to proofs with less variability. 
+
+## Adding triggers
```
robin-aws, line 567: I would explicitly flag this as a last resort, and perhaps even say if you find you HAVE to use triggers you should reach out to us to provide a better solution.

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,599 @@
...
+
+* Break large functions into smaller functions, especially when they're ghost functions, and make at least some of them opaque. I often consider a ghost function of more than a handful of lines to be a code smell. There are exceptions where larger functions are necessary, or the most straightforward alternative, though, such when handling all of the many cases of a complex algebraic data type.
+
+* Prove lemmas about functions rather than having numerous or complex postconditions. The individual lemmas can then be invoked only when needed, and other facts that are true but irrelevant will not bog down the solver.
+
+* Break methods up into smaller pieces, as well. Because methods are always opaque, this will, in many cases, require adding a contract to each broken-out method. To make this more tractable, consider abstracting the concepts used in those contracts into separate functions, some of which may be opaque.
```
robin-aws, line 599: I think such a detailed and multi-faceted document deserves a concluding section, especially just to list all the individual techniques for quick reference.

robin-aws (changes requested):
Love it.

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,599 @@
...
+
+When Dafny fails to prove an assertion, it will point out the location of that assertion in the program and describe what it was trying to prove at that point.
+
+However, when verification times out, Dafny will (by default) tell you only which definition timed out. This is similarly true in the case where a proof takes longer to complete than is practical for your development cycle. To get more fine-grained information about exactly which assertion was difficult to prove, it's possible to tell Dafny to _split_ the verification process into several batches, each of which can succeed, fail, or time out independently. This also makes independent statistics about resource use available for each batch.
+
+Dafny provides several attributes that tell it to verify certain assertions separately, rather than verifying everything about the correctness of a particular definition in one query to the solver.
+
+* The [`{:split_here}`](../DafnyRef/DafnyRef#sec-split_here) attribute on an `assert` statement tells Dafny to verify all assertions leading to this point in one batch and all assertions after this point (or up to the next `{:split_here}`, if there is one) in one batch.
```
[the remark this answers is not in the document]
atomb: I think these links will work when everything's rendered to HTML. I'm not entirely sure how to make them work both locally and in HTML.

I took that text from `Attributes.md`, and I'm honestly not entirely sure what the precise behavior is! Maybe @RustanLeino knows?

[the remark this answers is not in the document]
robin-aws: Happy to improve this later, it's technically correct if a bit vague, and more important to improve the description in the reference manual anyway.

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,599 @@
...
+
+If the time or resource count taken to verify a given assertion batch varies widely depending on the random seed, this is frequently predictive of later failure after small modifications. To measure the variation in resource use for the method from the previous section, on 10 different random seeds, we could use the following command.
+
+```sh
+$ dafny measure-complexity ProveSomeArithmetic.dfy --log-format csv --iterations 10
+```
+
+This will produce a CSV file and print out its name, as in the previous `dafny verify` case. With the `measure-complexity` command, however, each assertion batch will appear multiple times in the file. Given a CSV report containing the results of multiple iterations of verification, the [`dafny-reportgenerator`](https://github.com/dafny-lang/dafny-reportgenerator) tool can perform statistical analysis on this data, and can even be used to fail a build if the statistics exceed specified bounds.
+
+We generally recommend keeping the coefficient of variation (a.k.a. normalized standard deviation) of the resource count under 20%. This can be enforced with the following command.
```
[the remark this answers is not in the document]
robin-aws: Thanks, I like that you suggest `200000` as a default resource count even if that's likely too low for serious codebases - I'd rather suggest too low than too high. :)

docs/VerificationOptimization/VerificationOptimization.md
```diff
@@ -0,0 +1,625 @@
...
+
+If you have a difficult non-linear fact to prove, and proving it from the ground up is infeasible, it is possible, as a last resort, to adjust how Z3 attempts to prove arithmetic formulas. It contains a parameter, `smt.arith.solver`, that allows you to choose between one of six different arithmetic decision procedures. Dafny has historically used solver "2", and continues to do so by default. This peforms most effectively on the corpus of examples that we have experimented with. However, solver "6", which is the default for Z3 when not being invoked from Dafny, performs better at non-linear arithmetic. Currently, it's possible to choose this solver using the Dafny flag `--boogie /proverOpt:O:smt.arith.solver=6` (or just `/proverOpt:O:smt.arith.solver=6`, when not using top-level commands).
+
+## Relationships between integers and bit vectors
+
+Dafny contains two distinct ways to describe what might be stored in a machine word in other programming languages.
+
+* The `int` type denotes the unbounded mathematical integers. A subtype of `int` can be used to represent the subset of integers that can be represented in a machine word of width _n_.
```
robin-aws, line 257: Thanks, but this is *slightly* off: a subset type based on `int` would indeed be a subtype of `int`, but can't be compiled to native types safely, so it's not used in practice. And a `newtype` based on `int`, which is what folks do use, is NOT a subtype of `int` but a distinct type you can cast to and from `int`.

docs/ProofOptimization/ProofOptimization.md
```diff
@@ -0,0 +1,599 @@
...
+  ensures r > 0
+{
+  r := x + 1;
+  assert r > 0 by { reveal P; }
+}
+```
+
+## Subsumption
+
+The default behavior of an `assert` statement is to instruct the verifier to prove that a predicate is always true and then to assume that same predicate is true in later portions of the code, making it available to help prove other facts. This is called _subsumption_. On occasion, however, you may have an `assert` statement that exists to show a desired condition for its own sake, and that does not help prove later facts. In this case, you can [disable subsumption](../DafnyRef/DafnyRef#1133-subsumption-n), using `assert {:subsumption 0} P`, to reduce the number of facts in scope in subsequent code. This is similar to `assert L: P`, except that in the case of a labeled assertion the predicate can be selectively revealed later. If you know you'll _never_ need `P`, rather than _usually_ not needing it, subsumption could be a better option.
```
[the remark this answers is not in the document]
robin-aws: Thanks, the expanded description is just as good as an example.

docs/VerificationOptimization/VerificationOptimization.md
```diff
@@ -0,0 +1,625 @@
...
+
+* Prove lemmas about functions rather than having numerous or complex postconditions. The individual lemmas can then be invoked only when needed, and other facts that are true but irrelevant will not bog down the solver.
+
+* Break methods up into smaller pieces, as well. Because methods are always opaque, this will, in many cases, require adding a contract to each broken-out method. To make this more tractable, consider abstracting the concepts used in those contracts into separate functions, some of which may be opaque.
+
+# Summary
+
+When you're dealing with a program that exhibits high verification variability, or simply slow or unsuccessful verification, careful encapsulation and information hiding combined with hints to add information in the context of difficult constructs is frequently the best solution. Breaking one large definition with extensive contracts down into several, smaller definitions, each with simpler contracts, can be very valuable. Within each smaller definition, a few internal, locally-encapsulated hints to help the prover make progress in a more limited context can be very valuable.
```
robin-aws, line 611: Slighty awk: ending two consecutive sentences with "can be very valuable" :)

docs/VerificationOptimization/VerificationOptimization.md
atomb: Ah, right! Thanks for catching that.

docs/VerificationOptimization/VerificationOptimization.md
atomb: Another good catch.

robin-aws (approved):
🎉

docs/ProofOptimization/ProofOptimization.md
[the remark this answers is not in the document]
atomb: It's just a little above the maximum value in the CSV file shown. :) That's generally how I do things: set the maximum a little above where things are now, to make sure they don't get worse, and try to ratchet it down if possible.

Issue #3938, opened 2023-05-02
Expert mode with additional controls over how verification is performed


atomb:
### Summary

Provide some additional flags to control how verification is performed, perhaps on a project, file, module, or definition level (appropriate scope TBD), to improve performance and reduce variability when needed.

### Background and Motivation

Dafny currently takes a "best default for full automation" approach to encoding problems in Boogie and SMT, and providing parameters to the underlying solver. This approach works remarkably well for many programs, and it's important that we continue to make it work as well as possible, and leave it as the default.

However, as the programs created by Dafny users get larger and more sophisticated, highly variable proofs, as well as slow or unsuccessful proofs, are increasingly blocking Dafny users from making progress.

Allowing more control over encodings and solver parameters could allow tricky programs to be verified at all, to allow other programs to be verified more quickly, and to reduce the variability in verification performance. However, this control would require deeper understanding, so we'd want to be careful to avoid overwhelming users who don't need it.

### Proposed Feature

We could provide a number of additional flags to Dafny that are shown only when the expert mode is enabled. These could include (based on existing examples where different defaults can help, or educated guesses about where they might):

* Changing the arithmetic solver used by Z3. (Dafny currently chooses solver "2", but the Z3 default is "6", and we've seen examples where each performs better.)
* Changing the encoding of sequences between axiomatization and direct uses of solver theories.
* Changing between WP and, say, symbolic execution for VC generation.

### Alternatives

1. We could continue to search for default settings that work better than the current ones.
    * For example, we could change the encoding of sequences, once and for all, if a different encoding performs better.
1. We could attempt to infer different values for various settings based on the structure of the programs.
    * For example, straight-line, non-iterative code that works only on types directly supported by the solver could be handled differently than more complex code.
1. We could add the flags described above but make them visible by default.

None of these are mutually exclusive with each other or the proposed feature (except that the last would change the UI slightly), but they could be higher or lower priority.

robin-aws:
I'm less convinced that hiding all of these features by default is the right call, especially as I think there's no catch-all "expert" vs. "non-expert" classification. The first two flags you suggest, for example, are related to arithmetic and sequences respectively, which are very distinct fields where lots of users will be "experts" in one but not the other, or more importantly be writing code that exercises one or the other very deeply. If we're worried about the proliferation of options or them being distracting, perhaps we can just do things like group them differently in the output of `dafny verify --help`.

[...]

PR #3943, opened 2023-05-02
Add --solver-option flag and move others around

atomb:
Adds a `--solver-option` flag to correspond to the old (Boogie) `/proverOpt` flag. In the process, this also refactors a few things:

* Rearranges some of the option code to move some Boogie-related options from `CommonOptionBag` to `BoogieOptionBag`. 
* Makes `--solver-option`, `--solver-path`, `--solver-plugin`, `--solver-log`, and `--resource-limit` available to the  `generate-tests` and `find-dead-code` commands.

Includes some refactoring to DafnyTestGenerator, thanks to @Dargones, to make it easier to support arbitrary options there. 

Addresses #3468 but doesn't entirely resolve it.


Dargones:
@atomb, I have made a PR to this PR's branch that should make `--solver-option`, `--solver-path`, `--solver-log`, and `--resource-limit` work with test generation (I also did some refactoring that should hopefully make addition of new options easier in the future). 

Re: `--boogie` and `--solver-plugin` options - would you have any pointer on how to copy these from one DafnyOptions instance to another?

atomb:
> Thank you, @atomb! I have made a PR to this PR's branch that should make `--solver-option`, `--solver-path`, `--solver-log`, and `--resource-limit` work with test generation (I also did some refactoring that should hopefully make addition of new options easier in the future).

Thanks!
 
> Re: `--boogie` and `--solver-plugin` options - would you have any pointer on how to copy these from one DafnyOptions instance to another?

I think, for `--solver-plugin`, it should be sufficient to copy the `ProverDllName` and `TheProverFactory` fields. I think it should be safe for multiple threads to share a reference to the same factory, since we're already doing that when solving multiple goals in parallel during normal Dafny verification.

As for `--boogie`, it may be more trouble than it's worth. I considered adding that as a catch-all for any Boogie option that I didn't think to add, but I don't specifically know that it's needed.

Source/DafnyTestGeneration/GenerateTestsCommand.cs
```diff
@@ -10,9 +10,16 @@ namespace Microsoft.Dafny;
 public class GenerateTestsCommand : ICommandSpec {
   public IEnumerable<Option> Options =>
     new Option[] {
+      // IMPORTANT: Before adding new options, make sure they are
```
keyboardDrummer, line 13: How would a developer know that they should do that? Why doesn't `CopyForProcedure` use the DafnyOptions clone constructor instead

Source/DafnyTestGeneration/DeadCodeCommand.cs
```diff
@@ -8,9 +8,15 @@ namespace Microsoft.Dafny;
 public class DeadCodeCommand : ICommandSpec {
   public IEnumerable<Option> Options =>
     new Option[] {
+      // IMPORTANT: Before adding new options, make sure they are
```
keyboardDrummer, line 11: This seems frail. Why not make the code more robust instead of adding this comment? Why doesn't CopyForProcedure use the DafnyOptions clone constructor?

Source/DafnyTestGeneration/DeadCodeCommand.cs
```diff
@@ -21,6 +27,8 @@ public class DeadCodeCommand : ICommandSpec {
   }
 
   public void PostProcess(DafnyOptions dafnyOptions, Options options, InvocationContext context) {
+    // IMPORTANT: Before adding new default options, make sure they are
```
keyboardDrummer, line 30: This seems frail. Why not make the code more robust instead of adding this comment? Why doesn't CopyForProcedure use the DafnyOptions clone constructor?

Source/DafnyCore/Options/BoogieOptionBag.cs
```diff
@@ -58,6 +59,32 @@ public static class BoogieOptionBag {
...
+      IsHidden = true
+    };
+
+  public static readonly Option<bool> IsolateAssertions = new("--isolate-assertions", @"Verify each assertion in isolation.");
+
+  public static readonly Option<FileInfo> SolverPath = new("--solver-path",
+    "Can be used to specify a custom SMT solver to use for verifying Dafny proofs.") {
+  };
+  public static readonly Option<IList<string>> SolverOption = new("--solver-option",
+    "Specify an option to control how the solver works.") {
```
keyboardDrummer, line 84: This line doesn't tell me how to use this. Could you mention that these arguments are passed directly to the solver CLI; that the default solver is Z3 with a link to its CLI documentation; and that Dafny is already configuring some options, so not all of options might be usable.

Source/DafnyCore/Options/BoogieOptionBag.cs
```diff
@@ -58,6 +59,32 @@ public static class BoogieOptionBag {
...
+    @"Dafny uses Boogie as part of its verification process. This option allows customising that part using a Boogie plugin. More information about Boogie can be found at https://github.com/boogie-org/boogie. Information on how to construct Boogie plugins can be found by looking at the code in https://github.com/boogie-org/boogie/blob/v2.16.3/Source/Provers/SMTLib/ProverUtil.cs#L316");
+
+  public static readonly Option<string> SolverLog =
+    new("--solver-log", @"Specify a file to use to log the SMT-Lib text sent to the solver.") {
+      IsHidden = true
+    };
+
+  public static readonly Option<bool> IsolateAssertions = new("--isolate-assertions", @"Verify each assertion in isolation.");
+
+  public static readonly Option<FileInfo> SolverPath = new("--solver-path",
```
keyboardDrummer, line 80: I would set `SolverPath`, `SolverOption` and `SolverPlugin` (which I still think should be named BoogiePlugin) to hidden, since using them requires looking up additional documentation outside of Dafny. If there are particular Z3 options that we think users are likely to get a benefit front, I suggest we expose them as first class Dafny options.

Source/DafnyTestGeneration/GenerateTestsCommand.cs
Dargones: In short, merging [this one line PR to Boogie](https://github.com/boogie-org/boogie/pull/725) would allow using the DafnyOption's clone constructor and indeed make the code more robust.

The problem is that the `ProcsToCheck` field of `DafnyOptions` (inherited from Boogie's `CommadLineOptions`) does not have a setter. The clone constructor copies the field anyway using reflection but `ProcsToCheck` is a list, so the constructor copies the reference and one cannot modify the 'copied' list without modifying the original. The best solution would be to add a setter, like I do in the PR to Boogie above.

Source/DafnyCore/Options/BoogieOptionBag.cs
atomb: Yeah, my goal with #3938 is to expose more control over the verification process in a friendly way. But this allows expert users to get that control immediately, if they know they could benefit from it. And it allows us to conveniently experiment with various options to help decide which might be worth exposing in a nicer way.

I agree that `SolverOption` should be hidden, and I've made it that way. I'm less convinced about the two `Path` options, which I think will be used more frequently.

As far as the name of `SolverPlugin`, Boogie plugins can only control how it interfaces with the solver. It doesn't provide any mechanism for them to do anything else. So I still prefer `--solver-plugin`.

Source/DafnyCore/Options/BoogieOptionBag.cs
atomb: I added a `--solver-option-help` flag that describes what all the options are, and expanded the help text for `--solver-option`.

keyboardDrummer: Thanks

Source/DafnyCore/Options/BoogieOptionBag.cs
keyboardDrummer: Thanks

PR #3945, opened 2023-05-03
[Test Generation] Allow users to specify methods used to construct objects

Dargones:
This PR allows the user to specify methods that test generation can use to construct objects. My understanding is that a feature like this would improve the test generation experience (@codyroux, @hmijail, @aleks-AWS) but I mark the PR as a draft as I would want to make sure the proposed approach is helpful before asking to review the code. 

The proposed approach involves the user annotating such methods (which must be static, have one return parameter and no preconditions) with `:testConstructor` attribute. The test generation will use the provided method to construct objects of the corresponding type whose value is not constrained by the counterexample.

For instance, given the following Dafny program:

```Dafny
module M {
  class String { 
    var s:string;
    var len:int;
    // this constructor has a precondition,
    // so test generation will not use it
    constructor(s:string, len:int)  
      requires |s| == len
    {
      this.s := s;
      this.len := len;
    }
    // test generation will use this method to construct a string
    // whenever this string is unconstrained
    static method {:testConstructor} getString() returns (s:String) {
      s := new String("example", 7);
    }
  }
  function IsNull(s:String?):bool { 
    if s == null then false else true
  }
}
```
The tests generated for the `isNull` function will look like this (this test is added as part of the PR):

```Dafny
method {:test} test0() {
  var r0 := M.IsNull(null);
}
method {:test} test1() {
  var string0 : M.String := M.String.getString();
  var r0 := M.IsNull(string0);
}
```

### DIFF
--- a/Source/DafnyCore/TestGenerationOptions.cs
+++ b/Source/DafnyCore/TestGenerationOptions.cs
@@ -16,7 +16,8 @@ namespace Microsoft.Dafny {
     [CanBeNull] public string PrintBpl = null;
     [CanBeNull] public string PrintStats = null;
     public bool DisablePrune = false;
-    public static readonly uint DefaultTimeLimit = 10;
+    public const uint DefaultTimeLimit = 10;
+    public const string TestConstructorAttribute = "testConstructor";
 
     public bool ParseOption(string name, Bpl.CommandLineParseState ps) {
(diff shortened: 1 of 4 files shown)


Dargones:
Closing in favor of #5226

PR #3946, opened 2023-05-03
Fix: Formatting for parameter default values

### LINKED ISSUE #3944: Formatter bug
### Dafny version

4.0.0

```dafny
function Example(
  newNames : map<string, string> := map[],
  a: int
) : bool
{
  true
}
```

```code
dafny format

function Example(
  newNames : map<string, string> := map[],
                                        a: int
) : bool
{
  true
}
```

### What happened?

The second argument shouldn't be badly formatted

Windows

MikaelMayer:
This PR fixes #3944. I first recreated the issue with a test, and then fixed it.
It turned out, formals had a field "DefaultValue" which was never considered as a child, so its tokens were considered to be part of the declaration, which messed up indentation.
The fix was to ensure Formals now take the default value also in their range, and also in their children.


robin-aws (approved):
🚀

PR #3948, opened 2023-05-03
Check for underspecified pre-types

RustanLeino:
The new type inference works as follows:

A0. Resolve names and infer pre-types (that is, the parts of types that don't depend on theorem proving; in other words, ignoring subset types). These two things are mutually dependent. The inference of pre-types is done by laying down type constraints that are solved. A0 detects any over-constrained pre-types (for example, having to be both `int` and `bool`).
A1. Check for any under-constrained pre-types.
A2. Convert the inferred pre-types into the `Type`s that the rest of the Dafny pipeline expects.
B. "Improve" the inferred types by also considering subset types.

This PR contains the code for A1. There are no changes to code paths that don't use the new type inference. Note that A0 is not yet merged into master, so it would be pointless to run `/typeSystemRefresh:1` at this time (A1 would at best report underspecified pre-types everywhere, but may also crash in some places, because it's expecting A0 to fill in certain information). Nevertheless, it seemed to make sense to carve out A1 at this time into its own PR to be reviewed and merged separately.


Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
```diff
@@ -218,6 +218,63 @@ public enum Type2PreTypeOption { GoodForInference, GoodForPrinting, GoodForBoth
...
+          break;
+        }
+        var subst = PreType.PreTypeSubstMap(newtypeDecl.TypeArgs, preType.Arguments);
+        preType = (DPreType)parent.Substitute(subst);
+      }
+      return preType;
+    }
+
+    /// <summary>
+    /// AllParentTraits(decl) is like decl.ParentTraits, but also returns "object" is "decl" is a reference type.
```
stefan-aws, line 263:
```suggestion
    /// AllParentTraits(decl) is like decl.ParentTraits, but also returns "object" if "decl" is a reference type.
```

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
```diff
@@ -218,6 +218,63 @@ public enum Type2PreTypeOption { GoodForInference, GoodForPrinting, GoodForBoth
       return decl;
     }
 
+    /// <summary>
+    /// Returns the non-newtype ancestor of "cecl".
+    /// </summary>
+    public static TopLevelDecl AncestorDecl(TopLevelDecl decl) {
+      while (decl is NewtypeDecl newtypeDecl) {
+        var parent = newtypeDecl.BasePreType.Normalize();
```
stefan-aws, line 226: This part looks very similar to what is happening around line 251 in `NewTypeAncestor` below. Why do we have to account for cycles in the latter, but not here?

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
```diff
@@ -237,6 +294,25 @@ public enum Type2PreTypeOption { GoodForInference, GoodForPrinting, GoodForBoth
...
     }
 
+    /// <summary>
+    /// Add to "ancestors" every TopLevelDecl that is a reflexive, transitive parent of "d",
+    /// but not exploring past any TopLevelDecl that is already in "ancestors".
+    /// An ancestor
+    void ComputeAncestors(TopLevelDecl d, ISet<TopLevelDecl> ancestors) {
+      if (!ancestors.Contains(d)) {
+        ancestors.Add(d);
+        if (d is TopLevelDeclWithMembers dm) {
```
stefan-aws, line 304: nit: variable names `d`, `decl`, `dm` are used interchangeably for `TopLevelDecl` and `TopLevelDeclWithMembers`.

Source/DafnyCore/Resolver/PreType/UnderspecificationDetector.cs
```diff
@@ -0,0 +1,584 @@
...
+          if (f is ExtremePredicate cop) {
+            CheckMember(cop.PrefixPredicate);
+          } else if (f.ByMethodDecl != null) {
+            CheckMember(f.ByMethodDecl);
+          }
+        }
+      }
+    }
+
+    private void CheckPreType(PreType preType, ICodeContext codeContext, IToken tok, string what) {
```
stefan-aws, line 163: `what` could perhaps be renamed to e.g. `stringType`

Source/DafnyCore/Resolver/PreType/UnderspecificationDetector.cs
```diff
@@ -0,0 +1,584 @@
...
+      var visitor = new DetectUnderspecificationVisitor(this, codeContext);
+      visitor.CheckPreTypeIsDetermined(tok, preType, what);
+    }
+
+    private void CheckExpressions(Statement stmt, ICodeContext codeContext) {
```
stefan-aws, line 168:
```suggestion
    private void CheckStatement(Statement stmt, ICodeContext codeContext) {
```

Source/DafnyCore/Resolver/PreType/UnderspecificationDetector.cs
```diff
@@ -0,0 +1,584 @@
...
+  public class UnderspecificationDetector : ResolverPass {
+    public UnderspecificationDetector(Resolver resolver)
+      : base(resolver) {
+    }
+
+    /// <summary>
+    /// This method
+    ///     * checks that type inference was able to determine all types
+    ///     * checks that shared destructors in datatypes are in agreement
+    ///     * checks that ORDINAL is used only where allowed
```
stefan-aws, line 25: Most of these checks seems to be implemented as methods in the class `DetectUnderspecificationVisitor`. Might be good to move / add comments closer to the definitions.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
RustanLeino: `NewTypeAncestor` below detects cycles as it goes along. This is needed somewhere in the resolver. In contrast, `AncestorDecl` assumes that there are no cycles, so it should only be called after cycle detection. I will add comments to this effect.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
RustanLeino: Thanks. I changed these to use (our now more typical) longer names.

Source/DafnyCore/Resolver/PreType/UnderspecificationDetector.cs
RustanLeino: :O Thanks!

Source/DafnyCore/Resolver/PreType/UnderspecificationDetector.cs
```diff
@@ -0,0 +1,584 @@
...
+    protected override void VisitOneStmt(Statement stmt) {
+      if (stmt is VarDeclStmt) {
+        var s = (VarDeclStmt)stmt;
+        foreach (var local in s.Locals) {
+          CheckPreTypeIsDetermined(local.Tok, local.PreType, "local variable");
+          CheckTypeArgsContainNoOrdinal(local.Tok, local.PreType);
+        }
+      } else if (stmt is VarDeclPattern) {
+        var s = (VarDeclPattern)stmt;
+        s.LocalVars.Iter(local => CheckPreTypeIsDetermined(local.Tok, local.PreType, "local variable"));
```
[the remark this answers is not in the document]
RustanLeino: Hmm, I actually like `local` better.

Source/DafnyCore/Resolver/PreType/UnderspecificationDetector.cs
RustanLeino: I added further comments (here) that explain, more specifically, where those checks are performed. In doing so, I noticed that the last thing on the list isn't performed here at all (it's done in Pass 0 of resolution, but well after types have been inferred), so I removed that comment.

Source/DafnyCore/Resolver/PreType/UnderspecificationDetector.cs
RustanLeino: I renamed it to `whatIsBeingChecked`

Source/DafnyCore/Resolver/PreType/UnderspecificationDetector.cs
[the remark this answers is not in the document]
stefan-aws: Yeah, that's fine. I suggested `lv` (as in `LocalVars`) because you used `bv` (as in `BoundVars`) elsewhere.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
```diff
@@ -237,6 +239,8 @@ public enum Type2PreTypeOption { GoodForInference, GoodForPrinting, GoodForBoth
 
     /// <summary>
     /// Returns the non-newtype ancestor of "preType".
+    /// If the ancestor chain has a cycle or if some part of the chain hasn't yet been resolved, this method stops ends the traversal
```
stefan-aws, line 242:
```suggestion
    /// If the ancestor chain has a cycle or if some part of the chain hasn't yet been resolved, this method ends the traversal
```

stefan-aws (dismissed):

stefan-aws (approved):

PR #3967, opened 2023-05-06
fix: Block .doo files as input in legacy CLI

robin-aws:
Also use the new CLI when executing Dafny files for the library and Simple Dafny backends.

### DIFF
--- a/Source/DafnyCore/Compilers/ExecutableBackend.cs
+++ b/Source/DafnyCore/Compilers/ExecutableBackend.cs
@@ -116,7 +116,7 @@ public abstract class ExecutableBackend : Plugins.IExecutableBackend {
     SinglePassCompiler.WriteFromStream(rd, outputWriter);
   }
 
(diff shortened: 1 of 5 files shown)


Source/DafnyCore/Compilers/ExecutableBackend.cs
```diff
@@ -128,15 +128,17 @@ public abstract class ExecutableBackend : Plugins.IExecutableBackend {
...
-
+    var args = opt.MainArgs
+      .Prepend(targetFilename);
+    if (!verify) {
+      args = args.Prepend("--no-verify");
+    }
+    args = args
+      .Prepend("--target:cs")
+      .Prepend("run")
+      .Prepend(dafny);
```
atomb, line 139: Nice for us to migrate internally to more use of the new CLI! Apart from fixing a bug, this seems worthwhile for its own sake.

atomb (approved):

PR #3983, opened 2023-05-10
fix: Test and CI

fabiomadge:


robin-aws (approved):
Cut https://github.com/dafny-lang/dafny/issues/3984 for putting the test back (which we're removing because it will break every time the release script changes the version number)

PR #3993, opened 2023-05-10
Try using Boogie's "argument" type encoding scheme

atomb:
This seems to provide much better performance on the libraries code base, with no problems so far.

**Draft PR for testing, for the moment.**


atomb:
This will be easier to re-do at this point.

PR #3996, opened 2023-05-11
chore: Remove libraries submodule

robin-aws:
#3715 broken the nightly build because various tests in Test/libraries failed, but we already decided it made more sense to remove the submodule and rely on the libraries repo's own nightly build to catch issues instead.


robin-aws:
Doesn't fix the nightly build after all, but still good to do anyway.

PR #4014, opened 2023-05-15
fix: Remove multiset cardinality cap in Python

fabiomadge:
Fixes 4012.


stefan-aws (approved):
Can confirm it fixes the issue.

PR #4020, opened 2023-05-16
Make the Boogie representation of the Dafny heap use only Box

atomb:
This PR makes a simple but significant change to the Boogie representation of the Dafny heap, ensuring that all values stored in  it are of type `Box`. This makes the polymorphism in the Boogie `$Heap` variable irrelevant, but does not remove it. Removing it will, unfortunately, require slightly more significant changes to Dafny itself, more than the prelude. I plan that in a separate PR, but this one already seems to reduce resource use and resource use variation a little bit, so I thought it'd be good to merge it now. These changes seem to work best if done in very small steps.

The `ReadHeap` method now takes an optional type, which will ultimately be provided at all call sites. Once it is, the `Field` type in the prelude will no longer need a parameter and the heap can become truly monomorphic. The correct value to pass in isn't obvious at all call sites, however, so I'm filling each one in incrementally as I conclude what the right parameter should be.

(Note: there are a few tests that currently fail, but I'm working on fixing them. They're independent of the main content of this PR, just perturbed by essentially any verification change.)


Test/concurrency/11-MutexGuard2.dfy
```diff
@@ -1,3 +1,4 @@
+// UNSUPPORTED: windows,posix
```
atomb, line 1: This test repeatedly fails essentially whenever we make any change to the Boogie encoding. So I'm disabling it for now. I think it could be refactored to be less difficult, but that would be a somewhat involved job.

Source/DafnyCore/DafnyPrelude.bpl
```diff
@@ -649,7 +649,7 @@ procedure $IterHavoc1(this: ref, modi: Set Box, nw: Set Box);
 procedure $IterCollectNewObjects(prevHeap: Heap, newHeap: Heap, this: ref, NW: Field (Set Box))
                         returns (s: Set Box);
   ensures (forall bx: Box :: { s[bx] } s[bx] <==>
-              read(newHeap, this, NW)[bx] ||
+              (read(newHeap, this, NW) : Set Box)[bx] ||
```
MikaelMayer, line 652: This is not needed, but because you will eventually remove the alpha from read, you add it preventively, right?

Test/concurrency/11-MutexGuard2.dfy
MikaelMayer: At least it's still tested on other OS, right?

MikaelMayer (approved):
Looks good! If it helps performance, I think monomorphization is a good process, especially as it will make the code more maintainable (I have some other ideas)

Source/DafnyCore/DafnyPrelude.bpl
atomb: Exactly. And the code in Dafny that generates Boogie code does this for some but not all instances of `read`, as well. Once it does it everywhere, we can remove the type parameter to `Field`.

Test/concurrency/11-MutexGuard2.dfy
atomb: No, it's not tested anywhere. One of the other tests in that directory is disabled, too, because it doesn't reliably verify. Ultimately, it's interesting code but I'm not sure it's testing something particularly valuable. And it's so brittle that changing essentially anything in the encoding causes it to fail, even if it makes other things better.

RustanLeino (approved):
Thanks for doing this work. I look forward to the polymorphic map types being removed, since they add a Boogie-level of boxing on top of the boxing that Dafny already does.

RustanLeino (approved):

PR #4021, opened 2023-05-16
Adding error explanations for resolution errors

davidcok:


docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+## **Error: array selection requires an array_n_ (got _type_)**
+
+<!-- TODO - fix - different error message for this example -->
+
+```dafny
+const a: int
+const x: ORDINAL
+method m() 
+{  
+var c := a[0,0];
```
atomb, line 206: Indentation would make this slightly nicer.

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+## **Error: type conversion to an int-based type is allowed only from numeric and bitvector types, char, and ORDINAL (got _type_)**
+
+```dafny
+const x: int := true as int
+```
+
+Not all pairs of types have implicit or even explicit conversions. But there are conversions
+to int types from numeric types, including the ORDINAL type; for any source type, the value of 
+the numeric expression must be in the range for the int type (if it is a subset type or a newtype).
+Even `char` values have an integer representation (and thus a represetnation as a `real`)
```
atomb, line 369:
```suggestion
Even `char` values have an integer representation (and thus a representation as an `int`) 
```

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+
+## **Error: type conversion to a real-based type is allowed only from numeric and bitvector types, char, and ORDINAL (got _type_)**
+
+```dafny
+const x: real := true as real
+```
+
+Not all pairs of types have implicit or even explicit conversions. But there are conversions
+to real types from numeric types, including the ORDINAL type; for any source type, the value of 
+the numeric expression must be in the range for the int type (if it is a subset type or a newtype).
```
atomb, line 380:
```suggestion
the numeric expression must be in the range for the real type (if it is a subset type or a newtype).
```

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+Even `char` values have an integer representation corresponding to their unicode value.
```
atomb, line 381:
```suggestion
Even `char` values have a real representation corresponding to their unicode value.
```

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+## **Error: type conversion to a bitvector-based type is allowed only from numeric and bitvector types, char, and ORDINAL (got _type_)**
+
+```dafny
+const x: bv1 := true as bv1
+```
+
+Not all pairs of types have implicit or even explicit conversions. But there are explicit conversions
+to bitvector types from numeric types, including the ORDINAL type; for any source type, the value of 
+the numeric expression must be in the range for the bitvector type. Even `char` values have an
+integer representation corresponding to their unicode value.
```
atomb, line 393:
```suggestion
bitvector representation corresponding to their unicode value.
```
With "a" instead of "an" on the previous line.

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+
+```dafny
+const x: char := true as char
+```
+
+Not all pairs of types have implicit or even explicit conversions. But there are explicit conversions
+to the char type from numeric types, including the ORDINAL type; for any source type, the value of 
+the numeric expression must be in the range for the char type. The numeric values for a given
+`char` is its numeric unicode representation, which (for `--unicode-char=true`) is the range 
+0 to 0x10FFFF, inclusive.
```
atomb, line 405: There's a gap in the middle, to Z3's great sadness.

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+```
+
+The logical operators `||`, `&&`, `==>`, `<==`, `<==>` take only `bool` arguments.
+Dafny does not have any implicit conversion to or from `bool` values.
+
+## **Error: range of quantified variable must be of type bool (instead got _type_)**
+
+<!-- %no-check TODO - fix - different error message for this example -->
+```dafny
+const c := forall i: int | true && (i==0) :: true
```
atomb, line 482: In this example, the range _does_ have a `bool` type.

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+<!--two instances, for < <= and for > >= -->
+
+## **Error: type of _op_ must be a bitvector type (instead got _type_)**
+
+```dafny
+const z :=  0 << 1
+```
+
+The `<<` and `>>` operators are left- and right-shift operations. 
+They shift a bit-vector value by a given integer number of bits.
```
atomb, line 571: The number of bits can also have bit-vector type, as you mention below.

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -0,0 +1,678 @@
...
+The left-hand operand must be a value of a bit-vector type.
+Even int literals are not implicitly converted to bitvectors 
+(because Dafny wuld not know which bit-vector type to use).
```
atomb, line 574:
```suggestion
(because Dafny would not know which bit-vector type to use).
```

atomb (approved):
Great change (and a huge amount of work)! I saw a few tiny things to improve, but it otherwise looks good.

docs/check-examples
```diff
@@ -119,7 +119,7 @@ do
   let n++
 
   ## For header lines, extract the error message from the header text
-  ## everything but the initial ##[ ]*, the enclosing ** and the appended {#_errorid_}
+  ## (everything but the initial ##[ ]*, the enclosing ** and the appended {#_errorid_}
```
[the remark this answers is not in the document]
davidcok: Added

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -0,0 +1,763 @@
...
+## **Error: fresh expressions are allowed only in specification and ghost contexts**
+
+```dafny
+class A {}
+method m(a: A) returns (b: bool) {
+  b := fresh(a);
+}
+```
+
+The `old` construct is only used in ghost contexts. Typically using `old`
```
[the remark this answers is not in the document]
davidcok: Yes - fixed the C&P error

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -0,0 +1,763 @@
...
+
+```dafny
+const b: bv4 := 30
+```
+
+An integer literal can be converted implicitly to a value of a bitvector type,
+but only if the integer literal is in the range for the target type.
+For example, the type `bv4` has 4 bits and holds the values 0 through 15 inclusive.
+So a `bv4` can be initialized with a value in that range.
+Negative values are allowed: a value of -n corresponds to the bit vector
```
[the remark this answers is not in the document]
davidcok: Actually they are. Take up your argument with Rustan I guess, though the horse is a bit out of the barn.

docs/HowToFAQ/Errors-Resolver3.md
davidcok: Indented

docs/HowToFAQ/Errors-Resolver3.md
davidcok: Fixed up the text for int and real

docs/HowToFAQ/Errors-Resolver3.md
davidcok: Fixed

docs/HowToFAQ/Errors-Resolver3.md
davidcok: I hope the distinction between an integer value and an int type and an int-base type are clear (and consistently used)

davidcok:
Thanks for the reading and comments. Please double-check that I addressed them to your satisfaction.

docs/HowToFAQ/Errors-Resolution.md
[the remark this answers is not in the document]
atomb: I wrote up an example to try out how that works. It's a little weird, but I don't think you can prove false without intervening verification errors, so I'm a little less worried now. :)

PR #4025, opened 2023-05-17
refactor: Improve ClassDecl hierarchy

RustanLeino:
This PR changes the class hierarchy of the classes that previously rolled up to `ClassDecl`. This both improves the organization of the code and paves the way for traits to move away from always being reference types.

The old class hierarchy looked like:

```
   - TopLevelDecl
     - TopLevelDeclWithMembers
       - ValuetypeDecl
       - ClassDecl
         - ArrayClassDecl
         - IteratorDecl
         - DefaultClassDecl
         - TraitDecl
         - ArrowTypeDecl
```

Of these classes, `ArrowTypeDecl` and `DefaultClassDecl` are not really classes, but were historically placed under `ClassDecl` when `ClassDecl` was the only thing that contained members. (A further improvement, not in this PR, would be to rename `DefaultClassDecl` to something like `ModuleMembers`.) Also, while it previously was okay to treat `TraitDecl` as a kind of `ClassDecl`, this is about to change in Dafny, since traits will soon be available for non-reference types, too.

The new class hierarchy looks like:

```
   - TopLevelDecl
     - TopLevelDeclWithMembers
       - ValuetypeDecl
         - ArrowTypeDecl
       - DefaultClassDecl
       - ClassLikeDecl
         - ClassDecl
           - ArrayClassDecl
           - IteratorDecl
         - TraitDecl
```


Source/DafnyCore/AST/Cloner.cs
```diff
@@ -112,18 +112,19 @@ public class Cloner {
       } else if (d is TraitDecl) {
         var dd = (TraitDecl)d;
         var tps = dd.TypeArgs.ConvertAll(CloneTypeParam);
-        var mm = dd.Members.ConvertAll(d => CloneMember(d, false));
+        var mm = dd.Members.ConvertAll(member => CloneMember(member, false));
         var cl = new TraitDecl(Range(dd.RangeToken), dd.NameNode.Clone(this), m, tps, mm, CloneAttributes(dd.Attributes), dd.IsRefining, dd.ParentTraits.ConvertAll(CloneType));
         return cl;
+      } else if (d is DefaultClassDecl) {
```
keyboardDrummer, line 118: I think it's better to have the clone methods in the classes themselves, by implementing the `ICloneable` interface and letting `CloneDeclaration` check for the presence of that, but this is fine too.

By moving the cloning into the classes, you can also use a cloning constructor that calls into the cloning constructor of the base class. For `DefaultClassDecl`, that is only copying members inherited from base classes, that would mean the clone constructor can be empty and only call the clone constructor from its base class.

Source/DafnyCore/AST/Grammar/Printer.cs
```diff
@@ -382,18 +382,19 @@ public class Printer {
...
+        } else if (d is DefaultClassDecl defaultClassDecl) {
+          if (defaultClassDecl.Members.Count == 0) {
             // print nothing
           } else {
             if (i++ != 0) { wr.WriteLine(); }
-            PrintMembers(cl.Members, indent, fileBeingPrinted);
+            PrintMembers(defaultClassDecl.Members, indent, fileBeingPrinted);
           }
 
+        } else if (d is ClassDecl or TraitDecl) {
```
keyboardDrummer, line 393: Why not `d is ClassLikeDecl` ?

Source/DafnyCore/AST/Types/Types.cs
```diff
@@ -1101,8 +1101,8 @@ public enum AutoInitInfo { MaybeEmpty, Nonempty, CompilableValue }
                 return true;
               }
             }
-          } else if (udtSub.ResolvedClass is ClassDecl) {
-            var cl = (ClassDecl)udtSub.ResolvedClass;
+          } else if (udtSub.ResolvedClass is ClassDecl or TraitDecl) {
```
keyboardDrummer, line 1104: why not `is ClassLikeDecl` ? I get a bunch more hits on `ClassDecl or TraitDecl`, so if you change one consider changing all of them.

Test/dafny0/AutoContracts.dfy.expect
```diff
@@ -31,7 +31,7 @@ module _System {
   type {:axiom} array(==)<arg> = c: array?<arg> | c != null /*special witness*/
   */
 
-  class {:compile false} /*_#Func1*/ -T0 ~> +R {
+  type {:compile false} /*_#Func1*/ -T0 ~> +R {
```
keyboardDrummer, line 34: Seems like this PR is more than just a refactoring

Source/DafnyCore/Resolver/Resolver.cs
```diff
@@ -481,12 +481,19 @@ enum ValuetypeVariety {
       prog.CompileModules.Add(prog.BuiltIns.SystemModule);
       RevealAllInScope(prog.BuiltIns.SystemModule.TopLevelDecls, systemNameInfo.VisibilityScope);
       ResolveValuetypeDecls();
+
+      if (Options.Get(CommonOptionBag.TypeSystemRefresh)) {
```
keyboardDrummer, line 485: I didn't expect this change based on the PR details

keyboardDrummer (approved):

Source/DafnyCore/AST/Cloner.cs
RustanLeino: I've seen that you've done that with `Expression` and that looks nice. If a `Cloner` subclass needs to do something special with the cloning of a declaration (or expression, for that matter), how does it do that? Would it then have the `if (d is ...)` structure that `CloneDeclaration` currently has? In other words, is it just the standard cloning that would be done via the `IClonable` interface?

Source/DafnyCore/AST/Grammar/Printer.cs
RustanLeino: Good point. I changed it globally.

Source/DafnyCore/AST/Types/Types.cs
RustanLeino: Done.

Test/dafny0/AutoContracts.dfy.expect
RustanLeino: This change is because `ArrowTypeDecl` used to inherit from `ClassDecl` and now inherits from `ValuetypeDecl`. Really, arrow types were never classes in the first place.

Source/DafnyCore/Resolver/Resolver.cs
RustanLeino: My bad. That wasn't intended until a future PR.

Source/DafnyCore/AST/Cloner.cs
keyboardDrummer: Yes and yes. Usually such subclasses of `Cloner` only override the cloning behavior of particular AST nodes, so this doesn't bring us back to a giant `if ... else if.. etc`

keyboardDrummer (approved):

keyboardDrummer (dismissed):

Source/DafnyCore/AST/Cloner.cs
RustanLeino: Okay, then I agree that would be a good change. But since you say you're fine with the current version, too, I would like to request that we do it in a future PR instead.

keyboardDrummer: OK !

keyboardDrummer:
This PR still fails some of the XUnit tests.

keyboardDrummer (approved):

PR #4028, opened 2023-05-17
Stabilizing a flaky test

davidcok:


fabiomadge (approved):

davidcok:
This is a metatest so I thought that the point was to check that the %testForAllCompilers macro worked,
which this change does, though it is true that an ideal fix would use an program for which each compiler gave a different result. 
If I'm wrong, someone is welcome to provide an alternate fix.

PR #4031, opened 2023-05-17
Add additional IDE status messages, so users aren't wondering what ha…

keyboardDrummer:
Add additional IDE status messages, so users aren't wondering what happens between resolution and verification

I've tested the behavior with the current IDE and it still behaves the same, and with an updated IDE it shows the new states.

### DIFF
--- a/Source/DafnyLanguageServer/Workspace/Compilation.cs
+++ b/Source/DafnyLanguageServer/Workspace/Compilation.cs
@@ -110,6 +110,7 @@ public class Compilation {
     }
 
     try {
+      statusPublisher.SendStatusNotification(resolvedCompilation.TextDocumentItem, CompilationStatus.PreparingVerification);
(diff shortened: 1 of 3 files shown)


MikaelMayer (dismissed):
This will be a great addition for users to understand that it takes time to prepare verification. Good idea!

Also, but unrelated, I found that, when verification is onchange, there can be several seconds until resolution (parsing) starts again. I'm wondering what the IDE is doing during all this time.

keyboardDrummer:
> This will be a great addition for users to understand that it takes time to prepare verification. Good idea!
> 
> Also, but unrelated, I found that, when verification is onchange, there can be several seconds until resolution (parsing) starts again. I'm wondering what the IDE is doing during all this time.

Some explanation might be found in the description of this PR: https://github.com/dafny-lang/dafny/pull/4002

I've noticed that `UpdateDocument` is not called as quickly as I expected, and the cancellation mechanic on master is not reliable.

MikaelMayer (approved):

PR #4033, opened 2023-05-17
Fix: Ensures override checks have access to fuel constant equivalences

### LINKED ISSUE #2941: "the method must provide an equal or more permissive precondition than in its parent trait" when marking a predicate opaque
The code:

```dafny
trait T {
  predicate Valid(x: int)
  method MyMethod(x: int) requires Valid(x)
}

class C extends T {
  predicate {:opaque} Valid(x: int) { true }
  method MyMethod(x: int) requires Valid(x) {
  }
}
```

Incorrectly shows the error:
```
the method must provide an equal or more permissive precondition than in its parent trait
```

Boogie code:

```
    assume _module.T.Valid(this, x#0);
    assert _module.C.Valid(StartFuelAssert__module.C.Valid, this, x#0);
```

With `{:opaque} the following axioms are missing:

```
// layer synonym axiom
axiom (forall $ly: LayerType, this: ref, x#0: int ::
  { _module.C.Valid($LS($ly), this, x#0) }
  _module.C.Valid($LS($ly), this, x#0) == _module.C.Valid($ly, this, x#0));

// fuel synonym axiom
axiom (forall $ly: LayerType, this: ref, x#0: int ::
  { _module.C.Valid(AsFuelBottom($ly), this, x#0) }
  _module.C.Valid($ly, this, x#0) == _module.C.Valid($LZ, this, x#0));
```

… but they are not enough (since we don't know the value of `StartFuelAssert__module.C.Valid`

Reported by @MikaelMayer .

### LINKED ISSUE #3995: Opaque insufficiently revealing - Verification issue
Failing code
------------

```dafny
trait Animal {
    function           not_opaque() : ( rv : int)
    function {:opaque} iam_opaque() : ( rv : int) 

    predicate depends_on_not_opaque() 
       ensures  not_opaque() == 42 
    predicate depends_on_iam_opaque() 
       ensures  iam_opaque() == 42 
}

class Cat extends Animal {
    function           not_opaque() : ( rv : int) { 42 } 
    function {:opaque} iam_opaque() : ( rv : int) 
         //ensures rv == 42
         { reveal iam_opaque();  42 } 

    predicate depends_on_not_opaque() 
       ensures not_opaque() == 42 {true}
    predicate depends_on_iam_opaque() 
       ensures reveal iam_opaque(); iam_opaque() == 42 {reveal iam_opaque(); true}
}
```

Steps to reproduce the issue
----------------------------

* Dafny version: 4.0.0.50303
* Dafny VSCode extension version: 3.1.0
* 

Expected behavior
-----------------
I've made two sets of two parallel functions across a trait and a class
depends_on_not_opaque()'s postcondition depends on the return value of the not_opaque() function

depends_on_iam_opaque()'s postcondition is exactly the same,
except that it depends on the return value iam_opaque() function, which is opaque.
I put in "reveal iam_opaque()" everywhere I could think of at the time...

I expected both opaque and nonopaque versions would behave the same

Actual behavior
---------------

the not_opaque version works;  the iam_opaque version doesn't, complaining that

> the function must provide an equal or more detailed postcondition than in its parent trait

even though the postconditions is textually identical..

adding in a postcondition on the opaque version (uncommentiung the commented line  `//ensures rv == 42`)
means everything works. I sort-of understand why, but I think I should be able to do this with reveals... 

Here's what may be a related issue:

```dafny
class Rat {
    function           not_opaque() : ( rv : int) { 42 } 
    function {:opaque} iam_opaque() : ( rv : int) 
         { 42 } 

    predicate depends_on_not_opaque() 
       ensures  not_opaque() == 42 {true}
    predicate depends_on_iam_opaque() 
    //   requires reveal iam_opaque(); true
      ensures reveal iam_opaque(); iam_opaque() == 42 { true}   //reveal iam_opaque(); true
}
```

in an attempt to get to a single-class version I tried the above. 
This has a different problem: it does not verify if  `iam_opaque();` is revealed in the postcondition; (or in a separate ensures clause);
Error is 

> A postcondition might not hold on this return path.

It *does* verify if  `iam_opaque();` is revealed as a separate precondition `requires reveal iam_opaque(); true`
or in the body `{ reveal iam_opaque(); true }`.

MikaelMayer:
This PR fixes #3995 and fixes #2941

### DIFF
--- a/Source/DafnyCore/Dafny.atg
+++ b/Source/DafnyCore/Dafny.atg
@@ -430,19 +430,19 @@ bool IsExpliesOp() => IsExpliesOp(la);
 bool IsAndOp() => IsAndOp(la);
 bool IsOrOp() => IsOrOp(la);
 static bool IsEquivOp(IToken la) {
-  return la.val == "<==>" || la.val == "\u21d4";
+  return la.val == "<==>";
 }
 static bool IsImpliesOp(IToken la) {
-  return la.val == "==>" || la.val == "\u21d2";
+  return la.val == "==>";
 }
 static bool IsExpliesOp(IToken la) {
-  return la.val == "<==" || la.val == "\u21d0";
+  return la.val == "<==";
 }
 static bool IsAndOp(IToken la) {
-  return la.val == "&&" || la.val == "\u2227";
+  return la.val == "&&";
 }
 static bool IsOrOp(IToken la) {
-  return la.val == "||" || la.val == "\u2228";
+  return la.val == "||";
 }
 bool IsBitwiseAndOp() {
   return la.val == "&";
@@ -468,10 +468,7 @@ bool IsRelOp() {
       || la.val == "!="
       || la.val == "in"
       || la.kind == _notIn
-      || la.val =="!"
-      || la.val == "\u2260"
-      || la.val == "\u2264"
-      || la.val == "\u2265";
+      || la.val == "!";
 }
 bool IsShiftOp() {
   if (la.kind == _openAngleBracket) {
@@ -493,7 +490,7 @@ bool IsMulOp() {
   return la.kind == _star || la.val == "/" || la.val == "%";
 }
 bool IsQSep() {
-  return la.kind == _doublecolon || la.kind == _bullet;
+  return la.kind == _doublecolon;
 }
 
 bool IsNonFinalColon() {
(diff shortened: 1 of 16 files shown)


atomb (approved):
Wow. I'm happy to see it was such a simple fix, once you understood what was going on.

PR #4044, opened 2023-05-19
Bump Boogie to v2.16.8

atomb:
No major user-visible changes, but it should make the pool of solver threads behave better, and may bring us a step closer to being able to use Boogie's `/monomorphize` option with Dafny.


atomb:
Unfortunately, it seems like the latest Boogie is causing stack overflows in Dafny for some programs.

keyboardDrummer (approved):

PR #4050, opened 2023-05-21
Fixing a crash on using a non-existent doo file -- #4048

### LINKED ISSUE #4048: Crash on trying to produce library
### Dafny version

4.1.0

```dafny
dafny build -t:lib test2.dfy --out:M.doo
```

```code
Gives a crash: `--out:M.doo` is interpreted as a file to include as input, which does not exist.
```

### What happened?

Crashed

Mac

davidcok:
Fixes #4048


Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -283,6 +283,9 @@ public enum CommandLineArgumentsResult {
           }
           dafnyFiles.Add(df);
           isDafnyFile = true;
+        } catch (ArgumentException e) {
+          options.Printer.ErrorWriteLine(options.ErrorWriter, "*** Error: " + e.Message);
+          return CommandLineArgumentsResult.PREPROCESSING_ERROR;
```
robin-aws, line 288: Instead of using exceptions for control flow here, I'd actually prefer we do the same thing as in the preceding if branch: check `if (!File.Exists(filePath))` first and print out the same error message, before attempting the `DooFile.Read` (which then doesn't have to change).

Test/git-issues/git-issue-4048.dfy
```diff
@@ -0,0 +1,6 @@
+// RUN: %exits-with 1 %baredafny build -t:lib "%s" x.doo > "%t"
```
robin-aws, line 1: You should do `2>` instead so that the diff can actually check the stderr output.

robin-aws (changes requested):
Good catch (and glad some good came out of my example typo at least :) )

docs/DafnyRef/UserGuide.md
```diff
@@ -108,7 +108,7 @@ identified with the name `lib` on the command line. For example, to build multip
 Dafny files into a single build artifact for shared reuse, the command would look something like:
 
 ```bash
-dafny build -t:lib A.dfy B.dfy C.dfy --out MyLib.doo
+dafny build -t:lib A.dfy B.dfy C.dfy --output:MyLib.doo
```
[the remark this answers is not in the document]
davidcok: Yes it can and does. In this case it would need to be marked with  %check-cli and it would need to be a valid command - that is, A.dfy etc. would need to exist.

Test/git-issues/git-issue-4048.dfy
davidcok: Fixed

Source/DafnyDriver/DafnyDriver.cs
davidcok: I put in the check for File.Exists, but left the exception catching.
- things can happen in the file system between checking that a file exists and reading it
- in reading/unzipping the .doo file all sorts of bad things can happen, many of which are not worth trying to predict

davidcok:
I encountered another form of error, for which I'll create an issue. The use case has an error being produced that uses a token for some position in some file in the .doo file -- e.g. running dafny on A.dfy and a .doo file that contains A.dfy, so there are duplicate declarations. It appears that the token file names in the .doo file do not obey some invariants that have been otherwise expected. It seemed a rather large mission creep to fix it in this PR. (Issue #4070)

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -283,17 +287,19 @@ public enum CommandLineArgumentsResult {
...
+          options.Printer.ErrorWriteLine(options.ErrorWriter, "*** Error: {0}: ", nameToShow, e.Message);
+          return CommandLineArgumentsResult.PREPROCESSING_ERROR;
         } catch (IllegalDafnyFile e) {
           if (e.ProcessingError) {
             return CommandLineArgumentsResult.PREPROCESSING_ERROR;
           }
           // Fall through and try to handle the file as an "other file"
+        } catch (Exception e) {
+          options.Printer.ErrorWriteLine(options.ErrorWriter, "*** Error: {0}: {1}", nameToShow, e.Message);
+          return CommandLineArgumentsResult.PREPROCESSING_ERROR;
```
robin-aws, line 300: For the record I generally don't like losing the stack trace for unexpected exceptions, but I appreciate that this code is already a bit of a mess w.r.t. error messages so I'm not going to block on it. I've especially never been a fan of the only-used-for-control-flow exception `IllegalDafnyFile` :)

Test/doofiles/Test4.dfy
```diff
@@ -0,0 +1,8 @@
+// RUN: %baredafny build -t:lib --use-basename-for-filename "%s" --output "%S/test2" > "%t"
+// RUN: %baredafny resolve --use-basename-for-filename "%s"  "%S/test2.doo" >> "%t"
+// RUN: %diff "%s.expect" "%t"
+// XFAIL: *
```
robin-aws, line 4: Again I won't block on it, but couldn't we use `!` or `%exits-with` to be more precise about what fails? I'd generally prefer to avoid `XFAIL`.

robin-aws (approved):
A couple of questions/comments but nothing blocking. Appreciate the additional test coverage, and I think we can add more when we address #4070.

Test/doofiles/Test4.dfy
davidcok: This is not really an expected failure -- it is a bug waiting to be fixed, as described in the relevant issue. I agree that a real XFAIL is better written with %exists-with (which was the purpose of developing %exits-with in the first place).

PR #4058, opened 2023-05-22
Remove verify and compile fields from modules to enable caching parsing

keyboardDrummer:
### Changes

1) Modules no longer contain properties that indicate whether they should be verified and/or compiled, which is required for the upcoming change to cache parsing.

2) The API of Parser is simplified, so it always takes a TextReader as input. 

3) Change `SourceProcessor` so it doesn't need to open a file separately to determine what newlines it has, but can instead ride on the existing TextReader that is also used for parsing the file.

### Testing

The above is purely refactoring, so no test changes have been made.


Source/DafnyCore/DafnyFile.cs
```diff
@@ -14,63 +14,43 @@ public class DafnyFile {
...
-  }
-  public DafnyFile(DafnyOptions options, string filePath, bool useStdin = false) {
-    UseStdin = useStdin;
-    Uri = useStdin ? new Uri("stdin:///") : new Uri(filePath);
-    BaseName = useStdin ? "<stdin>" : Path.GetFileName(filePath);
-
-    var extension = useStdin ? ".dfy" : Path.GetExtension(filePath);
+  public TextReader Content { get; }
+  public Uri Uri { get; }
+  public DafnyFile(DafnyOptions options, string filePath, TextReader contentOverride = null) {
```
alex-chew, line 19: (copied from partial review of #4002)

nit: It'd be good to name contentOverride to suggest that it should only be used for stdin (unless that's not true, but it seems so based on the implementation)

Source/DafnyCore/DafnyMain.cs
```diff
@@ -79,27 +79,24 @@ public class DafnyMain {
...
-            col = 1,
-            line = 0
-          }, null, dafnyFile.SourceFilePath, false)
-          : null;
+        // We model a precompiled file, a library, as an include
+        var include = dafnyFile.IsPrecompiled ? new Include(new Token {
+          Uri = dafnyFile.Uri,
+          col = 1,
+          line = 0
+        }, new Uri("cli://"), dafnyFile.FilePath) : null;
```
alex-chew, line 87: (copied from partial review of #4002)

Instead of "cli://" can we make the URI suggest that it's from a precompiled file or library?

Source/DafnyCore/AST/ShouldCompileOrVerify.cs
```diff
@@ -0,0 +1,101 @@
...
+      program.UrisToCompile = ComputeUrisToCompile(program);
+    }
+
+    if (module.FullName == "_System") {
+      return true;
+    }
+
+    if (module is DefaultModuleDefinition) {
+      // If things from precompiled files live in the default module, that can cause downstream compilation issues:
+      // https://github.com/dafny-lang/dafny/issues/4009
```
alex-chew, line 21: Is this meant to be a TODO/FIXME?

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -391,13 +391,13 @@ public enum CommandLineArgumentsResult {
         return exitValue;
       }
 
-      if (0 <= options.VerifySnapshots && lookForSnapshots) {
+      if (0 < options.VerifySnapshots && lookForSnapshots) {
```
alex-chew, line 394: Is this a bug fix? And does it mean that we've accidentally been calling `ExecutionEngine.LookForSnapshots()` by default since the introduction of this code?

Source/DafnyCore/DafnyFile.cs
keyboardDrummer: In the current implementation that is indeed how it should be used. I'm working on a follow-up PR now where `DafnyFile` has all the hardcoded `stdin` code removed. Would you allow us to temporarily be in the awkward situation that you've identified, until we merge the follow-up PR ?

Source/DafnyCore/AST/ShouldCompileOrVerify.cs
keyboardDrummer: The GH issue itself is meant to be the 'TODO', without putting an actual TODO in the code. I prefer this way of planning future work.

Source/DafnyDriver/DafnyDriver.cs
keyboardDrummer: Yes it's a bug fix but possibly only for a latent bug. 

According to the documentation, `0` means the future is turned off. `options.VerifySnapshots` is `-1` by default, which makes the previous code work well unless someone explicitly assigns `0` to this option.

Source/DafnyCore/DafnyMain.cs
keyboardDrummer: This code is hacky because it's creating an `include` while there is no include, so there is no sensible value for the URI of the `includer`. I will remove this code in a follow-up.

Source/DafnyCore/DafnyFile.cs
alex-chew: Sounds good, thanks for the clarification.

PR #4059, opened 2023-05-22
Do not let `Parser.Parse` edit the default module

keyboardDrummer:
- `Parser.Parse` no longer takes the default module as an argument which it then appends to, but instead returns a new type `FileModuleDefinition`, which acts as a container for file contents.
- `SourceProcessor.ProcessDirectives` now makes sure not to remove or add newlines at the end of the file.
- Parsing code is shared between CLI and IDE
- Module cloning code has been moved out of `Cloner` and into the module classes.


Source/DafnyCore/DafnyMain.cs
```diff
@@ -140,83 +98,22 @@ public class DafnyMain {
...
+    //       return $"{errorCount} parse errors detected in {fn}";
+    //     }
+    //   } catch (IOException e) {
+    //     IToken tok = include == null ? Token.NoToken : include.tok;
+    //     errorReporter.Error(MessageSource.Parser, tok, "Unable to open included file");
+    //     return $"Error opening file \"{fn}\": {e.Message}";
+    //   }
+    //
+    //   return null; // Success
+    // }
```
alex-chew, line 116: Is this still needed?

Source/DafnyPipeline.Test/Trivia.cs
```diff
@@ -147,7 +147,6 @@ ensures true
       }
 
       AreAllTokensOwned(program);
-      Assert.Equal(9, count); // Sanity check
```
alex-chew, line 150: Why does this need to be removed?

Source/DafnyCore/Rewriters/RefinementTransformer.cs
```diff
@@ -401,31 +402,31 @@ public RefinementTransformer(Program p)
...
         if (d is ClassDecl) {
-          m.TopLevelDecls[index] = MergeClass((ClassDecl)nw, (ClassDecl)d);
+          nwPointer.Set(MergeClass((ClassDecl)nw, (ClassDecl)d));
         } else {
           Reporter.Error(MessageSource.RefinementTransformer, nw, commonMsg, nw.WhatKind, nw.Name);
         }
       } else if (nw is DefaultClassDecl) {
         if (d is DefaultClassDecl) {
-          m.TopLevelDecls[index] = MergeClass((DefaultClassDecl)nw, (DefaultClassDecl)d);
+          m.DefaultClass = (DefaultClassDecl)MergeClass((DefaultClassDecl)nw, (DefaultClassDecl)d);
```
alex-chew, line 429: I'm not very familiar with this part of the code, but is it not necessary to `nwPointer.Set(m.DefaultClass)` here?

alex-chew (approved):
Please split up big PRs like this, especially since much of the diff here is refactoring and moving code around, as opposed to implementing the changes described in the PR description.

Source/DafnyCore/Rewriters/RefinementTransformer.cs
keyboardDrummer: This will work as well, since there's only one `DefaultClass` field so assigning to it will do the same as using the `Set` of that field's pointer, but what you suggest looks better. I will change that in the follow-up PR.

Source/DafnyCore/DafnyMain.cs
keyboardDrummer: Ugh, woops. Will remove in the follow-up PR (https://github.com/dafny-lang/dafny/pull/4083)

Source/DafnyPipeline.Test/Trivia.cs
keyboardDrummer: Because the DefaultModule now has a filename in its StartToken, the traversal doesn't go deeper than that Node and count only gets up to 1. This is fine though. I'm not sure why `Assert.Equal(9, count)` was in there because it wasn't asserting any required behavior.

keyboardDrummer:
> Please split up big PRs like this, especially since much of the diff here is refactoring and moving code around, as opposed to implementing the changes described in the PR description.

The pure moves are just `ScopeCloner` and `AbstractSignatureCloner`, that were placed in their own file, and indeed that didn't need to be in this PR. The core of the PR change is the new parsing logic in `ParseUtils`, which is a combination of what was in `DafnyLangParser` and in `DafnyMain`, and the remaining changes were just to get the tests to pass again.

PR #4060, opened 2023-05-22
fix: Missing substitution in the subset check of the function override check

### LINKED ISSUE #4056: Framing problem with recursive reads-clause
### Dafny version

4.1.0

```dafny
trait ADT {
  ghost function ReprFamily(n: nat): set<object>
    decreases n
    ensures n > 0 ==> ReprFamily(n) >= ReprFamily(n-1)
    reads this, if n == 0 then {} else ReprFamily(n-1)
}

class P extends ADT {
  ghost function ReprFamily(n: nat): set<object>
    decreases n
    ensures n > 0 ==> ReprFamily(n) >= ReprFamily(n-1)
    reads this, if n == 0 then {} else ReprFamily(n-1)
}
```

```code
Error: function might read an object not in the parent trait context's reads clause
```

### What happened?

In versions prior to 3.13.0 this code was verified by Dafny, but since that version Dafny complains on the `P` class that implements the trait. It is not straightforward code but it should work. Each call to `ReprFamily(n)` reads `ReprFamily(n-1)`, until `ReprFamily(0)` where it reads `{}`, so it is well-founded recursion. In fact, Dafny accepts the code of the trait, but not the code of the class that implements that trait with exactly the same specification.

I've noticed that in version 3.13.0 a pull request was accepted that changes the checking of traits: #3479. Maybe it has something to do with this problem.

Linux

fabiomadge:
Fixes #4056.


MikaelMayer (approved):
Well done fixing this, I'm happy to rereview after merging with latest master

PR #4064, opened 2023-05-23
Enabling show-snippet option when using console (#3304)

### LINKED ISSUE #3304: --show-snippets does not work for parse errors
### Dafny version

3.20.0+dev

```dafny
const c = 5
```

```code
dafny verify --show-snippets test.dfy
```

### What happened?

Got:
/Users/davidcok/projects/dafny/dafny/Test/tmp.dfy(1,10): Error: a const field should be initialized using ':=', not '='
1 parse errors detected in /Users/davidcok/projects/dafny/dafny/Test/tmp.dfy

Expected to see a display of source code, like this for verification errors:
  |
3 |   assert false;
  |          ^^^^^

Mac

davidcok:
Fixes #3304


keyboardDrummer:
Can you comment on when this option should be used and when it shouldn't be? Should this be the default? Is there a reason not to use this option?

davidcok:
If you propose that it be the default, I suggest merging this PR and having the discussion with the group -- should we make this the default -- which will then require at least adjusting the tests and perhaps having a discussion with users. A later PR can turn the option on by default.

davidcok:
Fine with me actually. But please let's do this in two steps.

keyboardDrummer:
> If you propose that it be the default, I suggest merging this PR and having the discussion with the group -- should we make this the default -- which will then require at least adjusting the tests and perhaps having a discussion with users. A later PR can turn the option on by default.

I think we should only add options that we expect to be used. What's the use-case for having this option off ?

davidcok:
Besides in testing, the choice to limit verbosity. In other tools I use an equivalent feature both on and off, depending on what I'm doing.

keyboardDrummer:
> Besides in testing, the choice to limit verbosity. In other tools I use an equivalent feature both on and off, depending on what I'm doing.

I know it limits verbosity, but what is the situation in which you would use it?

davidcok:
For example, a lot of error messages about some syntax item -- say extraneous semicolons. The line number is just fine for context. I prefer a compact list of the error messages to work from. I work as much from command-line as I do within an IDE -- which leads me at least to like compact output and visual presentation.

davidcok:
Also scripts that process output from tools like dafny usually want a more predictable output.
Also, if we are going to spring a new output format on users, we should give them a way to turn it off.

robin-aws:
Since this PR is arguably fixing a bug in the existing `--show-snippets` option (to respect it in all forms of errors related to Dafny source code), we should merge as is, and deal with the independent question of whether the option should be on by default (which I'm also in favor of) in a separate change.

robin-aws:
To further clarify:

> I think we should only add options that we expect to be used. What's the use-case for having this option off ?

This PR doesn't add the option, just fixes its implementation.

davidcok:
Well on that point -- we had the option in old CLI, but not in new CLI. And Remy is right to raise the question whether old CLI options should be kept around. Though in this case I think it should be. In addition to comments above, I'm biased in favor of giving users options.

robin-aws:
Right, although again (unless I'm missing something) this PR itself doesn't add the option to the new CLI? I'm still in favor of merging the bug fix and then having the conversation about whether it's early enough to change the default without anyone being affected in practice.

davidcok:
No, this PR does add a new-CLI option that wraps the pre-existing old CLI option.

keyboardDrummer:
> I work as much from command-line as I do within an IDE 

I think that's a use-case we should support a 100%, so no concern there. I'm as much committed to development without as with the IDE.

> I prefer a compact list of the error messages to work from.

Do you like that because it allows you to look at multiple error messages at the same time? What's the use of that? I would think you always solve errors one by one, and after solving each error it is OK to scroll to the next one, even if the text is verbose.

> For example, a lot of error messages about some syntax item -- say extraneous semicolons.

Those would be warnings right? They should show up below the errors, so you don't have to scroll over them. Also, if you want to ignore that warning and look at other warnings, we should enable you to disable seeing that warning in the first place.

keyboardDrummer (approved):

PR #4073, opened 2023-05-24
Fix: Assert only applies locally only.

### LINKED ISSUE #4055: Assert {:only}'s effect are past the enclosing method
### Dafny version

latest nightly

```dafny
lemma LemmaOutsideofScaffolding2(a: int, b: int)
  ensures a == b
{
  assert false;
}

lemma LemmaWithScaffolding()
{
  assert {:only "after"} true;
}

lemma LemmaOutsideofScaffolding(a: int, b: int)
  ensures a == b
{
  assert false;
}
```

```code
Paste the code in VSCode.
```

### What happened?

image

- LemmaOutsideofScaffolding is not offered options to verify manually.
- It is displayed as verified when it is not
- Removing assert {:only} removes this issue.

Normally, {:only} should be local to a method, and moreover, I'm surprised that since it's {:after}, I would have imagined that, in the presence of a bug, only methods before would not verify.
Also, I wish to have {:only} also on top-level declarations.

Windows

MikaelMayer:
This PR quickly fixes #4055


atomb (approved):
Yeah, this seems like a more convenient semantics now that `{:only}` can apply to members, too.

PR #4080, opened 2023-05-24
More documentation about using project files

davidcok:

### DIFF
--- a/docs/DafnyRef/UserGuide.md
+++ b/docs/DafnyRef/UserGuide.md
@@ -606,7 +606,8 @@ Most output from `dafny` is directed to the standard output of the shell invokin
 
 ### 13.5.5. Project files
 
-Commands on the Dafny CLI that can be passed a Dafny file, can also be passed a Dafny project file. Such a project file may define which Dafny files the project contains, and which Dafny options it uses. The project file must be a [TOML](https://toml.io/en/) file named `dfyconfig.toml` for it to work on both the CLI and in the Dafny IDE, although the CLI will accept any `.toml` file. Here's an example of a Dafny project file:
+Commands on the Dafny CLI that can be passed a Dafny file can also be passed a Dafny project file. Such a project file may define which Dafny files the project contains and which Dafny options it uses. The project file must be a [TOML](https://toml.io/en/) file named `dfyconfig.toml` for it to work on both the CLI and in the Dafny IDE, although the CLI will accept any `.toml` file. 
+Here's an example of a Dafny project file:
 
 ```toml
 includes = ["src/**/*.dfy"]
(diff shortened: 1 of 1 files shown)


docs/DafnyRef/UserGuide.md
```diff
@@ -617,9 +618,15 @@ enforce-determinism = true
 warn-shadowing = true
 ```
 
-Under the section `[options]`, any options from the Dafny CLI can be specified using the option's name without the `--` prefix. When executing a `dafny` command using a project file, any options specified in the file that can be applied to the command, will be. Options that can't be applied or are misspelled, are ignored.
-
-When using a Dafny IDE based on the `dafny server` command, the IDE will search for project files by traversing up the file tree looking for the closest `dfyconfig.toml` file it can find. Options from the project file will override options passed to `dafny server`.
+- At most one `.toml` file may be named on the command-line; when using the command-line no `.toml` file is used by default.
+- `.toml` files with syntax errors will cause the invocation of dafny to fail (with a preprocessing error).
```
keyboardDrummer, line 622: This is so much in line with my expectations that I don't think it's useful to document.

docs/DafnyRef/UserGuide.md
```diff
@@ -617,9 +618,15 @@ enforce-determinism = true
 warn-shadowing = true
 ```
 
-Under the section `[options]`, any options from the Dafny CLI can be specified using the option's name without the `--` prefix. When executing a `dafny` command using a project file, any options specified in the file that can be applied to the command, will be. Options that can't be applied or are misspelled, are ignored.
-
-When using a Dafny IDE based on the `dafny server` command, the IDE will search for project files by traversing up the file tree looking for the closest `dfyconfig.toml` file it can find. Options from the project file will override options passed to `dafny server`.
+- At most one `.toml` file may be named on the command-line; when using the command-line no `.toml` file is used by default.
```
keyboardDrummer, line 621: What's the reason to use a list (`-`) for these .toml file related attributes? Why not one or multiple paragraphs? The items on line 623-625 belong together, as do those on line 626-628.

keyboardDrummer, line 621: > At most one `.toml` file may be named on the command-line

Is that good to mention in the documentation? I will also find out if I try it on the command line. I feel 99% of users won't try this and are just slowed down by larger documentation, while the 1% that doesn't attempt to do this because of this documentation only has a minor benefit.

keyboardDrummer:
Thanks!

docs/DafnyRef/UserGuide.md
davidcok: Its a reference manual, not a tutorial -- so my expectations are that the document defines the behavior.
Reasonable alternate behavior is that later things on the command-line supersede earlier ones, as options do.

davidcok: I'll omit line 622 though. (I have encountered tools that silently omit badly formatted input files, though I don't remember where at the moment).

docs/DafnyRef/UserGuide.md
davidcok: Its a reference manual, not a tutorial -- so my expectations are that the document defines the behavior.
Reasonable alternate behavior is that later things on the command-line supersede earlier ones, as options do.

keyboardDrummer (approved):

PR #4085, opened 2023-05-26
Cache parsing

keyboardDrummer:
### Changes
- Let the Dafny language server use a cache for parsing. The cache stores items as long as they were used in the previous compilation. This means that commenting out an include can cause a large piece of cache to be pruned, leading to a possibly unexpected slowdown when it's uncommented again.
- Fix the cloner so it clones the RangeToken property of Expression objects. This is required for parse caching since we need to clone the AST before further processing since it is mutable.
- Many fixes to the cloner code
- Add telemetry that records how long parsing a particular compilation took
- Add test for included file has error message for indirect includes, something that was still missing from https://github.com/dafny-lang/dafny/pull/4083

### Caveats
I've decided not to add an option to turn off caching, since I don't see the use-case. Of course caching is a computation vs storage trade-off, and we could provide many options to configure the caching behavior, but I think the current caching is cheap enough (at most storing two sets of parse results at a time) that there is no reason to want to turn it off.

Please carefully review the code that determines the cache key, since a bug in that would lead to outdated compilation results.

### Testing
- Add a test that checks if parsing is cached, and that the cache prunes entries when they're not hit.
- Add a test that checks that for larger files, the start and end of the file is taken into account for computing the caching hash, so presumably the rest of the file as well. Also check that the Uri changes the hash.
- There are many existing tests that make changes to a file which will trigger the new hashing code.


Source/DafnyLanguageServer/Caching/ByteArrayEquality.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Microsoft.Dafny.LanguageServer.Language;
+
+class ByteArrayEquality : IEqualityComparer<byte[]> {
+  public int Compare(byte[]? x, byte[]? y) {
+    return (x, y) switch {
+      (null, null) => 0,
+      (null, _) => 1,
+      (_, null) => -1,
+      _ => x.SequenceEqual(y) ? 0 : 1
+    };
```
alex-chew, line 15: 1. It's a bit strange that if both `x` and `y` are both non-null and unequal, then both `Compare(x, y)` and `Compare(y, x)` return 1.
2. `Compare` is defined not in `IEqualityComparer<T>` but instead in `IComparer<T>`, did you mean to also add the latter to the list of derived interfaces?

Source/DafnyLanguageServer/Caching/TextReaderFromCharArrays.cs
```diff
@@ -0,0 +1,74 @@
...
+    var array = arrays[arrayIndex];
+    if (elementIndex == array.Length) {
+      return -1;
+    }
+    var result = array[elementIndex++];
+    if (array.Length == elementIndex) {
+      arrayIndex++;
+      elementIndex = 0;
+    }
+    return result;
```
alex-chew, line 53: I think this `Read` implementation won't work as expected if there's an empty `char[]` anywhere in `arrays` (except perhaps at the end). Is that intentional, and if so can we document that?

Source/DafnyLanguageServer/Caching/TextReaderFromCharArrays.cs
```diff
@@ -0,0 +1,74 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks.Dataflow;
+
+namespace Microsoft.Dafny.LanguageServer.Language;
+
+class TextReaderFromCharArrays : TextReader {
```
alex-chew, line 10: Please consider adding unit tests for this class.

Source/DafnyLanguageServer/Caching/ByteArrayEquality.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Policy;
+
+namespace Microsoft.Dafny.LanguageServer.Language;
+
+class ByteArrayEquality : IEqualityComparer<byte[]> {
```
alex-chew, line 8: Please consider adding unit tests for this class.

Source/DafnyLanguageServer/Caching/ByteArrayEquality.cs
```diff
@@ -0,0 +1,34 @@
...
+    };
+    ;
+  }
+
+  public int GetHashCode(byte[] obj) {
+    if (obj.Length == 0) {
+      return 0;
+    }
+    return BitConverter.ToInt32(obj, 0);
+  }
```
alex-chew, line 33: Can you add some documentation on how this hash code is being used? It seems unusual to use the first four bytes of a byte array as a "hash".

alex-chew (commented):
I've reviewed the first 30 files (as they appear in the diff view) - that's up to and including `Source/DafnyLanguageServer/Caching/TextReaderFromCharArrays.cs`. I'll continue reviewing the rest.

Source/DafnyCore/AST/Grammar/ProgramParser.cs
```diff
@@ -130,7 +143,7 @@ public class ParseUtils {
     defaultModule.DefaultClass.SetMembersBeforeResolution();
   }
 
-  public static IList<DfyParseResult> TryParseIncludes(
+  public IList<DfyParseResult> TryParseIncludes(
```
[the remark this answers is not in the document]
keyboardDrummer: `ParseFile` is virtual, and overwritten by `CachingParser`

Source/DafnyLanguageServer/Caching/ByteArrayEquality.cs
keyboardDrummer: Removed the `Compare` method, it wasn't used.

Source/DafnyLanguageServer/Language/CachingParser.cs
```diff
@@ -0,0 +1,69 @@
...
+
+  /// <summary>
+  /// We read the contents of the reader and store them in memory using chunks
+  /// to prevent allocating a large array of memory.
+  /// </summary>
+  private static (TextReader reader, byte[] hash) ComputeHashFromReader(Uri uri, TextReader reader, HashAlgorithm hashAlgorithm) {
+    var result = new List<char[]>();
+    const int chunkSize = 1024;
+    hashAlgorithm.Initialize();
+    var uriBytes = Encoding.UTF8.GetBytes(uri.ToString());
```
alex-chew, line 48: To be safe I'd suggest adding a delimiter between the URI and the text content, in order to prevent hash collisions where the URI and text content might be "confused":
```suggestion
    // Add NUL delimiter to avoid collisions (otherwise hash("A" + "BC") == hash("AB" + "C"))
    var uriBytes = Encoding.UTF8.GetBytes(uri.ToString() + "\0");
```
(This should be okay on filesystems that disallow NUL bytes in filenames, which is most filesystems AFAIK.)

Such collisions are unlikely to occur naturally, but it's easy to construct a malicious example. Without a delimiter, a file named `foo.dfy//.dfy` with contents `X`, would have the same hash as another file named `foo.dfy` with contents `//.dfyX`.

alex-chew:
Most of the changes LGTM. The remaining blocking issues are:

- adding a bounds check in `TextReaderFromCharArrays.Peek`
- checking that empty `char[]`s are handled correctly in `TextReaderFromCharArrays.Read` (it looks to me like they shouldn't, but the added unit test will tell)
- preventing trivial collisions in the cache key hashing function

alex-chew (approved):

PR #4087, opened 2023-05-26
--show-snippets with default true and off in tests

davidcok:
Adjusts the tests so that if --show-snippets is on by default, the tests do not fail.
Also sets --show-snippets on by default.


keyboardDrummer:
I'm OK with this change, but requested some others to review as well.

docs/check-examples
```diff
@@ -379,11 +380,11 @@ do
         else
           if [ "$command" == "%check-legacy" ]; then
             isverbose=-compileVerbose:$checkIds
-            (cd "$dir"; "$dafny" $verb $dOptions $isverbose ${options[@]} $text   ) > "$actualOut" 2> "$actualError"
+            (cd "$dir"; "$dafny" $verb /showSnippets:0 $dOptions $isverbose ${options[@]} $text   ) > "$actualOut" 2> "$actualError"
```
MikaelMayer, line 383: @keyboardDrummer  This looks suspicious, it should probably be --show-snippets:false

davidcok: $verb is always empty at this point -- this is the branch for legacy CLI tests. SO it is correct, but perhaps $verb could be removed.

davidcok: Removed $verb on this line to avoid confusion.

davidcok: And tests now pass after nightlies have been fixed.  Mikael - can you approve now?

robin-aws (commented):
I'm fine with changing the default here since most users aren't going to be parsing console output, and even if they are broken by this it's just a matter of adding `--show-snippets:false`.

But on that note, do add a release note to explicitly call this change out, so it's easier for users to spot it.

(not explicitly requesting changes so another reviewer is free to approve once this is done)

davidcok:
Mikael (or Robin) - I have added a release note per Robin’s request. Please approve the PR or let me know what else needs improvement.

Thanks - David

keyboardDrummer (approved):

PR #4089, opened 2023-05-27
Adding documentation for some resolution errors

davidcok:


Source/DafnyCore/AST/Grammar/ParseErrors.cs
```diff
@@ -10,6 +10,8 @@ public class ParseErrors {
 
   public enum ErrorId {
     g_include_has_errors, // In Reporting.cs
+    p_generic_syntax_error,
+    p_generic_semantic_error,
```
robin-aws, line 14: Will we want an exception for these ones to avoid having to document them? I get that we need an enum value for these locations in code but they don't seem helpful for users.

Source/DafnyCore/Resolver/ResolutionErrors.cs
```diff
@@ -8,9 +8,77 @@ namespace Microsoft.Dafny;
 public class ResolutionErrors {
 
   public enum ErrorId {
-    // ReSharper disable once InconsistentNaming
```
robin-aws, line 11: Don't we still need this?

docs/DafnyRef/Attributes.md
```diff
@@ -669,7 +669,7 @@ Overrides the `/subsumption` command-line setting for this assertion.
 `{:subsumption 0}` checks an assertion but does not assume it after proving it.
 You can achieve the same effect using [labelled assertions](#sec-labeling-revealing-assertions).
 
-## 11.4. Attributes on variable declarations
+## 11.4. Attributes on variable declarations {#sec-assumption}
```
robin-aws, line 672: This doesn't seem like the right label here - shouldn't it be on 11.4.1 below?

docs/HowToFAQ/check-errors
robin-aws, line 1: Doesn't have to be in this PR, but can you add a bit of text at the top of this file and `check-errors-a` to describe what it does/how to use it? Especially since it's not used in CI yet.

docs/HowToFAQ/check-errors-a
```diff
@@ -34,7 +34,7 @@ done
 ## ErrorIds from Add
 
 for f in $errfiles ; do
-  echo ENUM FILE $f
+  echo ENUM FILE ADD $f
```
robin-aws, line 37: I can't figure out what these symbols are for and hence what `ADD` is doing.

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -652,9 +760,9 @@ Hence, indications of non-terminating loops, that is, `decreases *`, are not per
 
 This does mean that the specifier has to do the work of designing a valid terminating condition and proving it.
 
-<!-- 2 instances -->
+<!-- 3 instances -->
```
robin-aws, line 763: What are these lines doing?

robin-aws (commented):
Still have to go through the main Errors-Resolution.md changes, but here are some comments/questions now that I understand the system here.

Source/DafnyCore/AST/Grammar/ParseErrors.cs
davidcok: These are intended to be replaced by more specific errorids. Where that is not possible, probably for the syntax errors automatically generated by CoCo, I agree that we should think about what to say -- there is something useful to say though: there is a syntax error but CoCo doesn't really know the real problem.

Source/DafnyCore/Resolver/ResolutionErrors.cs
davidcok: Perhaps - but I don't know what it means.

docs/DafnyRef/Attributes.md
davidcok: Correct. Fixed.

docs/HowToFAQ/check-errors
davidcok: Will do - I'm using them but they are not polished yet.

docs/HowToFAQ/check-errors-a
davidcok: Just debugging output. I'll add comments in a future PR.

docs/HowToFAQ/Errors-Resolution.md
davidcok: They remind me that the same error message is produced by multiple spots in the code -- so I should make sure that the description is appropriately general, and perhaps add multiple examples. Keep in mind that these examples all are converted into lit scenarios that feed into the coverage measurements. When browsing the coverage results, if I see a error message that is uncovered, I try to determine whether I'm just missing a test or whether it is defensive programming that is (currently) unreachable.

robin-aws: Cool! You've got some great automation on all of this. Another +1 for a bit more documentation on it. :)

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -29,43 +32,64 @@ method m(a: A) returns (r: int) {
...
+method m(a:A) 
+{
+  print a.f;
+}
+```
 
-## **Error: _what_ '_name_' can be used only in specification contexts**
+Functions may have some (or all) formal parameters be ghost. Such parameters can only be used in ghost expressions
+within the function body. There are limits though on where such a function may be used.
+For example, passing the value of the function itself (not all call of the function) is restricted to ghost contexts.
```
robin-aws, line 49:
```suggestion
For example, passing the value of the function itself (not a call of the function) is restricted to ghost contexts.
```

?

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -423,34 +431,77 @@ are ORDINALs that are larger than any `nat`. Logical reasoning with ORDINALs is
...
+## **Error: the value returned by an abstemious function must come from invoking a co-constructor** {#r_abstemious_needs_conconstructor}
 
-## **the value returned by an abstemious function must come from invoking a co-constructor**
+```dafny
+codatatype D = A | B
+function {:abstemious} f(): int {0}
+```
 
 <!-- TODO -->
+_Abstemious functions are not documented. Please report occurences of this error message._
```
robin-aws, line 444: That's a relief, first I've ever heard of them :)

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -483,9 +534,12 @@ But a ghost context can also be implicit, and not so obvious: if something ghost
 such as the condition of an `if` statement or loop or the expression being matched in a match 
 statement, then the rest of the statement may be required to be ghost.
 
-<!-- TODO: comment on tracking of printeffects -->
+In addition, methods must be marked with the `{:print}` attribute if 
+`--track-print-effects` is enabled and it has 
+`print` statements or calls methods marked with `{:print}`.
```
robin-aws, line 539:
```suggestion
In addition, methods must be marked with the `{:print}` attribute if 
it has `print` statements or calls methods marked with `{:print}`
and `--track-print-effects` is enabled.
```

Reads better IMO to have the "it" closer to the subject.

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -519,15 +573,67 @@ considered ghost. And then the statement can contain no substatements that are f
...
+}
+```
 
-## **_var_ cannot be assigned a value that depends on a ghost**
+No changes to non-ghost variables may occur in ghost contexts.
+Ghost context can be implicit, such as the branches of an if-statement
+whose condition is a ghost expression.
 
-<!-- TODO -->
+## **Error: non-ghost _var_ cannot be assigned a value that depends on a ghost** {#r_no_assign_ghost_to_var}
```
robin-aws, line 590: I find "a value that depends on a ghost" confusing, as it implies value have lingering dependencies on other values, which doesn't make sense to me in a language without lazy evaluation semantics. Would it be accurate to just say "a ghost value"?

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -519,15 +573,67 @@ considered ghost. And then the statement can contain no substatements that are f
...
+
+## **Error: assumption variable must be of type 'bool'** {#r_assumption_var_must_be_bool}
+
+```dafny
+method m() {
+  ghost var {:assumption} b: int;
+}
+```
+
+Variables marked with `{:assumption} must have bool type.
```
robin-aws, line 612:
```suggestion
Variables marked with `{:assumption}` must have bool type.
```

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -519,15 +573,67 @@ considered ghost. And then the statement can contain no substatements that are f
...
+See [the reference manual](#../DafnyRef/DafnyRef#sec-assumption) for more detail on the use of this attribute.
+
+
+## **Error: assumption variable must be ghost** {#r_assumption_var_must_be_ghost}
```
robin-aws, line 616:
```suggestion
## **Error: assumption variables must be ghost** {#r_assumption_var_must_be_ghost}
```

Or is the typo in the actual error message?

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -519,15 +573,67 @@ considered ghost. And then the statement can contain no substatements that are f
...
+
+```dafny
+method m() {
+  var {:assumption} b: bool;
+}
+```
+
+Variables marked with `{:assumption} must be ghost.
```
robin-aws, line 624:
```suggestion
Variables marked with `{:assumption}` must be ghost.
```

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -519,15 +573,67 @@ considered ghost. And then the statement can contain no substatements that are f
...
+See [the reference manual](#../DafnyRef/DafnyRef#sec-assumption) for more detail on the use of this attribute.
+
+
+## **Error: assumption variable can only be declared in a method** {#r_assumption_var_must_be_in_method}
```
robin-aws, line 628:
```suggestion
## **Error: assumption variables can only be declared in a method** {#r_assumption_var_must_be_in_method}
```

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -519,15 +573,67 @@ considered ghost. And then the statement can contain no substatements that are f
...
+
+<!-- TODO - not sure this is reachable -->
+
+Variables marked with `{:assumption} must be declared within methods.
```
robin-aws, line 632:
```suggestion
Variables marked with `{:assumption}` must be declared within methods.
```

robin-aws (changes requested):
Great improvements!

Source/DafnyCore/AST/Grammar/ParseErrors.cs
davidcok: OK

docs/HowToFAQ/Errors-Resolution.md
davidcok: The typo is in the error message, some of which I have been fixing (if they don't involve fixing too many tests).
Actually in this form it is just cryptic: could be 'this assumption variable must be ghost'

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -304,50 +325,54 @@ More often one means `exists i :: 0 <= i < 10 && a[i] == 0`, that is, is there a
...
 
 <!-- TODO -->
 
-## **Error: type parameter '_name_' (inferred to be '_type_') in the function call to '_name_' could not be determined. If you are making an opaque function, make sure that the function can be called.**
+## **Error: type parameter '_name_' (passed in as '_type_') to function call '_name_' is not allowed to use ORDINAL** {#r_function_type_parameter_may_not_be_ORDINAL}
 
 <!-- TODO -->
 
-## **Error: type parameter '_name_' (passed in as '_type_') to function call '_name_' is not allowed to use ORDINAL**
+## **Error: the type of the bound variable '_var_' could not be determined** {#r_bound_variable_type_undetermined}
```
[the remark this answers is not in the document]
davidcok: Absolutely. That is why the messages are in a stylized form. The heading in the text is converted to a regular expression that is then matched against the output of running dafny against the code example. CI will fail if they do not match. The checking also ensures that the errorid in the text, which is also the HTML anchor in the markdown document, matches the errorid in the C# code.

docs/HowToFAQ/Errors-Resolution.md
```diff
@@ -304,50 +325,54 @@ More often one means `exists i :: 0 <= i < 10 && a[i] == 0`, that is, is there a
...
+const i := 10
+const b := i is Small
+```
 
-<!-- TODO -->
+The `is` type test is currently somewhat limited in Dafny, and more limited than the companion `as` conversion.
+In particular, `is` is not allowed to test that a value is a member of a subset type.
 
-## **Warning: the type of the other operand is a non-null type, so this comparison with 'null' will always return '_bool_'**
+## **Warning: the type of the other operand is a non-null type, so this comparison with 'null' will always return '_bool_'_hint_** {#r_trivial_null_test}
```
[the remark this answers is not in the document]
davidcok: The _hint_ is there to match against variable dafny output that explains a bit more of the context.

docs/HowToFAQ/Errors-Resolution.md
davidcok: But they are a thing (related to codatatypes) -- Rustan has a task to add a bit of documentation.

docs/HowToFAQ/Errors-Resolution.md
davidcok: FIne - though 'they' is better

docs/HowToFAQ/Errors-Resolution.md
robin-aws: Agreed and I almost skipped this one since error messages are often this cryptic/abstract, but really that's no excuse and they SHOULD be more readable :)

robin-aws (approved):

robin-aws (approved):

PR #4092, opened 2023-05-28
Documenting rewriter errors

davidcok:


docs/HowToFAQ/check-errors
```diff
@@ -3,6 +3,13 @@
 d=$(dirname ${BASH_SOURCE[0]})
 cd $d
 
+echo ">>>> REWRITER"
+
+csfiles="../../Source/DafnyCore/Rewriters/*.cs ../../Source/DafnyCore/Rewriters/*/*.cs"
+./check-errors-a Errors-Rewriter.md "$csfiles" ../../Source/DafnyCore/Rewriters/RewriterErrors.cs
+
+exit 0
```
fabiomadge, line 11: When do we run this?

docs/HowToFAQ/Errors-Rewriter.md
```diff
@@ -37,10 +37,23 @@ has no effect and is likely to lead to misunderstanding. Consequently, Dafny war
...
+<!-- %check-resolve-warn -->
+```dafny
+method m() {
+  assert!(1 == 1);
+}
+```
 
-## **Warning: _item_s given as :induction arguments must be given in the same order as in the quantifier; ignoring attribute**
+This message warns about a common syntax error for people familiar with Rust. 
+In Rust `assert! e` indicate checking that expression `e` holds. But in Dafny the `!` negates the expression.
```
fabiomadge, line 50:
```suggestion
In Rust `assert! e` indicates checking that expression `e` holds. But in Dafny the `!` negates the expression.
```

fabiomadge (approved):

docs/HowToFAQ/check-errors
davidcok: Well, it will be -- I'm using it to check all the error Ids, but it needs a bit of cleanup (and unification across all the files that use errorids) becfore I put it in CI.

docs/HowToFAQ/Errors-Rewriter.md
davidcok: OK - I'll fix in another PR.

PR #4093, opened 2023-05-28
More detail for resolution errors

davidcok:


docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -355,6 +379,12 @@ size operator -- not arrays, `iset`, or `imap`.
 
 ## **Error: a _what_ definition is not allowed to depend on the set of allocated references**
 
+```dafny
+class B {}
+const bbb:B
```
fabiomadge, line 384:
```suggestion
const bbb: B
```

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -477,15 +507,14 @@ Dafny does not have any implicit conversion to or from `bool` values.
...
+function f(i: set<int>): set<int> { set k: int <- i |  true || k  }
 ```
 
-In a quantified expression, the expression between the `|` and the `::` is the 
-_range_ expression. It serves to limit the values of the quantified variable(s)
-that are to be considered in evaluating the expression after the `::`. 
-As such, this range expression must always have a `bool` type.
+In a quantification using the `<-` syntax, the type of the quantified variable is
+determined by its explicit declaration or by the type of the elements of the container
+(the right-hand operand). If then the quantified variable is used as a `bool` value
```
fabiomadge, line 516:
```suggestion
(the right-hand operand). If the quantified variable is used as a `bool` value
```

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -661,14 +695,48 @@ collection of that type.
...
 <!-- TODO -->
 
 ## **Error: type of right argument to * (_type_) must agree with the result type (_type_)**
 
+```dafny
+function ff(i: int, j: real): real { j * i }
+```
+
+The types of the two arguments of `*` must be the same (or implicitly convertible to be the same).
+Typically the result of the expression is detrermined by the left operand.
```
fabiomadge, line 715:
```suggestion
Typically the result of the expression is determined by the left operand.
```

docs/HowToFAQ/Errors-Resolver3.md
```diff
@@ -661,14 +695,48 @@ collection of that type.
...
+This message then is stating that the right operand has a different type.
+
+
+## **Error: second argument to _op_ must be a set, multiset, or sequence with elements of type _type_, or a map with domain _type_ (instead got _type_)**
+
+```dafny
+function ff(i: int, j: real): bool { i in j }
+```
+
+The operators `in` and `!in`test membership of a value in a container,
```
fabiomadge, line 725:
```suggestion
The operators `in` and `!in` test membership of a value in a container,
```

fabiomadge (approved):

PR #4097, opened 2023-05-29
Enables output of errorId

davidcok:
Adds --verbose to parsing and enabling it to emit the errorId and explanation


Source/DafnyCore/Generic/Reporting.cs
```diff
@@ -218,7 +218,8 @@ public class ConsoleErrorReporter : BatchErrorReporter {
           errorLine += $" {msg} {tok.TokenToString(Options)}";
         }
 
-        if (Options.Verbose && false) { // Need to control tests better before we enable this
+        if (Options.Verbose) { // Need to control tests better before we enable this
+          errorLine += "\nID: " + errorId;
```
robin-aws, line 222: I think it's a bit confusing to have just "ID:" on a separate line - less obvious that's it's an error id about the previous line.  I think this would be clearer:

```
formatting1-3.dfy(1,0): Warning: File contains no code (ID: ...)
```

This would probably be even better, to put the id next to "Warning" or "Error", but may require more plumbing:

```
formatting1-3.dfy(1,0): Warning (ID: ...): File contains no code 
```

Test/dafny0/formatting1.dfy.expect
```diff
@@ -23,6 +23,7 @@ module A {
   }
 }
 formatting1-3.dfy(1,0): Warning: File contains no code
+ID:
```
robin-aws, line 26: I'd say we shouldn't print this at all if there is no error code yet.

Test/server/symbols.transcript.expect
```diff
@@ -1,5 +1,6 @@
 # Reading from symbols.transcript
 bank.dfy(4,25): Warning: constructors no longer need 'this' to be listed in modifies clauses
+ID: none
```
robin-aws, line 3: Where is "none" coming from?

Source/DafnyCore/Generic/Reporting.cs
davidcok: I'll put it after the message, so the important part of the message is upfront.
It is still only included when --verbose is on, so it won't affect existsing tests, but will take a bit more sedding to tease it out of the message.

robin-aws (approved):
Approving knowing that other pending PRs will add tests shortly (since there are now no changes to existing tests, which is good!)

PR #4110, opened 2023-06-01
Move AST classes into smaller files and add some directory hierarchy

keyboardDrummer:
Move classes in the files `Expression.cs`, `Statements.cs`, `MemberDecls.cs` and `TopLevelDeclarations.cs`, into separate files. I have applied a preliminary grouping of classes into folders.

> Why? 

Smaller classes improve performance of IDEs and GitHub; make it easier to keep multiple cursors in place when switching between code contexts; and allow for more hierarchical structure in the code. Also, I plan to move more code into the AST classes. I think everything up to and including resolution, including parsing and printing, that's specific to a particular AST node, should live inside or next to the definition of that AST node type.


Source/DafnyCore/AST/Statements/Verification/AssertLabel.cs
robin-aws, line 1: Slight nit (commenting on folder): I might prefer "Specification" rather than "Verification", since the former is really what these language elements are about, and the latter is just one possible means of evaluating whether the specification is satisfied. But not enough to block on (can always rename later).

Source/DafnyCore/AST/Types/TypeParameter.cs
robin-aws, line 1: I would put this under TypeDeclarations too personally, rather than have a separate folder just for this. And perhaps TypeDeclarations could just be Types. But again not blocking.

robin-aws (approved):
Lovely! I had minor quibbles with some of the categorization but not enough to block on - I'm confident developers will be able to find nodes wherever they are, and it's just so beneficial to break down more monsters files into smaller ones.

Issue #4108, opened 2023-05-31
Inconsistent printing of unordered collections across backends


robin-aws:
This is one of those issues that most of us are fully aware of but have lived with for a long time without writing down (unless I just can't find the existing issue).

```dafny
method Main() {
   print {"abc", "def"}, "\n";
}
```

```
% dafny run -t:java src/Scratch.dfy

Dafny program verifier finished with 0 verified, 0 errors
{['a', 'b', 'c'], ['d', 'e', 'f']}
% dafny run -t:cs src/Scratch.dfy

Dafny program verifier finished with 0 verified, 0 errors
{['d', 'e', 'f'], ['a', 'b', 'c']}
```

Not really a bug per se since the behavior of `print` is underspecified, but this makes consistent differential testing really annoying, whether testing a user program or the backends themselves.

My proposal for the most beneficial solution: implement the collections such as `set`, `multiset` and `map` in Dafny in the same way we now have a Dafny implementation of `seq`: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntimeDafny/src/dafnyRuntime.dfy. This would depend on having a common implementation of hash code functions for all Dafny types, instead of relying on target language runtime hash codes. We'd then have unpredictable but deterministic ordering for these collections. This is also worth doing in order to provide more efficient implementations of these immutable data types, rather than wrapping mutable versions from the target runtimes and making expensive defensive copies everywhere. 

See also #2062.

robin-aws:
Even worse, Python salts its hashes so its orderings are not even deterministic from one run to the next. This makes it effectively impossible to use `%testDafnyForEachCompiler` for such tests.

MikaelMayer:
I understand it's an issue for printing, but as you present it, it looks like you are only concerned about tests for now.
For tests, I think it could be solved by doing the following. For every collection where we expect a result for print like this:

```dafny
...
  print myset;
```

```
// expect file
{"abc", "def"}
```
1) Define a sequence of the expected result
2) Verify that each element of the expected sequence is in the collection (with appropriate value if it's a map
3) Verify that all keys of the sequence are different
4) Verify that the length of the sequence is equal to the length of the map.

```dafny
...
  var expected := ["abc", "def"];
  for i <- 0 to |expected| {
    if expected[i] !in myset {
      print "bad";
    }
  }
  if |myset| != |expected| || |myset| != |set k <- expected| {
    print "bad"
  }
```

[...]

PR #4115, opened 2023-06-02
fix: Uniform backend testing

### LINKED ISSUE #4103: Crash when specifying `--split-translation` on Java
### Dafny version

4.1.0

```dafny
// No code necessary
```

```code
% dafny build -t:java src/Scratch.dfy --spill-translation

Unhandled exception. System.AggregateException: One or more errors occurred. (One or more errors occurred. (Could not find a part of the path '/Users/salkeldr/Documents/GitHub/libraries/_System/__default.class'.))
 ---> System.AggregateException: One or more errors occurred. (Could not find a part of the path '/Users/salkeldr/Documents/GitHub/libraries/_System/__default.class'.)
 ---> System.IO.DirectoryNotFoundException: Could not find a part of the path '/Users/salkeldr/Documents/GitHub/libraries/_System/__default.class'.
   at System.IO.FileSystem.DeleteFile(String fullPath)
   at System.IO.File.Delete(String path)
   at Microsoft.Dafny.Compilers.JavaBackend.<>c.<CompileTargetProgram>b__15_1(String f) in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyCore/Compilers/Java/JavaBackend.cs:line 100
   at System.Collections.Generic.List`1.ForEach(Action`1 action)
   at Microsoft.Dafny.Compilers.JavaBackend.CompileTargetProgram(String dafnyProgramName, String targetProgramText, String callToMain, String targetFilename, ReadOnlyCollection`1 otherFileNames, Boolean runAfterCompile, TextWriter outputWriter, Object& compilationResult) in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyCore/Compilers/Java/JavaBackend.cs:line 100
   at Microsoft.Dafny.DafnyDriver.CompileDafnyProgram(Program dafnyProgram, String dafnyProgramName, ReadOnlyCollection`1 otherFileNames, Boolean invokeCompiler) in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyDriver/DafnyDriver.cs:line 914
   at Microsoft.Dafny.DafnyDriver.Compile(String fileName, ReadOnlyCollection`1 otherFileNames, Program dafnyProgram, PipelineOutcome oc, IDictionary`2 moduleStats, Boolean verified) in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyDriver/DafnyDriver.cs:line 736
   at Microsoft.Dafny.DafnyDriver.ProcessFilesAsync(IReadOnlyList`1 dafnyFiles, ReadOnlyCollection`1 otherFileNames, DafnyOptions options, Boolean lookForSnapshots, String programId) in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyDriver/DafnyDriver.cs:line 435
   at Microsoft.Dafny.DafnyDriver.ThreadMain(TextWriter outputWriter, TextWriter errorWriter, TextReader inputReader, String[] args) in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyDriver/DafnyDriver.cs:line 135
   at Microsoft.Dafny.DafnyDriver.<>c__DisplayClass11_0.<MainWithWriters>b__0() in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyDriver/DafnyDriver.cs:line 112
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__272_0(Object obj)
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
   at Microsoft.Dafny.DafnyDriver.MainWithWriters(TextWriter outputWriter, TextWriter errorWriter, TextReader inputReader, String[] args) in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyDriver/DafnyDriver.cs:line 112
   at Microsoft.Dafny.DafnyDriver.Main(String[] args) in /Users/salkeldr/Documents/GitHub/dafny/Source/DafnyDriver/DafnyDriver.cs:line 99
   at Dafny.Dafny.Main(String[] args) in /Users/salkeldr/Documents/GitHub/dafny/Source/Dafny/Dafny.cs:line 7
```

### What happened?

Bug in the logic that's trying to delete the "temporary" class files.

Mac

robin-aws:
Converts most existing integration tests that touch backends to run for every supported backend, as a relatively easy way to discover backend-specific bugs and increase testing coverage. Implements a conversion process to rewrite existing testing scripts like:

```
// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:cs "%s" >> "%t"
// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:js "%s" >> "%t"
// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:go "%s" >> "%t"
// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:java "%s" >> "%t"
// RUN: %dafny /noVerify /compile:4 /spillTargetCode:2 /compileTarget:py "%s" >> "%t"
```

To the equivalent:

```
// RUN: %testDafnyForEachCompiler "%s" -- --relax-definite-assignment --spill-translation
```

Overview of changes:

* Added `DAFNY_INTEGRATION_TESTS_MODE` environmment variable to support these uniformity requirements:
  * `uniformity-check` - Verifies that every test that touched backends either uses `%testDafnyForEachCompiler` or includes a `// RUN: NONUNIFORM: <reason>` command to flag it as intentionally non-uniform. This check is run as a `singletons` test in CI.
  * `uniformity-convert` - Convert any existing non-uniform tests to use `%testDafnyForEachCompiler`, by splitting up the existing `<test file>.expect` file into the chunks output by each backend and extracting any common extra options from all test commands. See `LitTests.ConvertToMultiBackendTestIfNecessary`.
  * (no value or blank) - executes tests as usual
* Converted `LitTests.InvokeMainMethodsDirectly` to an environment variable as well (for consistency and to avoid having to edit source code to use this feature locally)
* Added functionality to `%testDafnyForEachCompiler` to support known inconsistencies/bugs in backends (see updates to `TestDafny/README.md`
* Converted lots of private state in the `ILitCommand` hierarchy to publicly readable attributes, to support the analysis/conversion logic
  * Refactored `OutputCheckCommand` implementation to be usable from `TestDafny`

To review, I recommend looking first at the README updates, then at the source changes, then spot-checking the test conversions. The last commit is the result of applying the conversion logic to the whole test suite (and I'll try to keep it that way even after feedback edits), so it probably best to look at the diff up until but not including that commit first.

FYI, see https://github.com/dafny-lang/dafny/labels/testing-method%3A%20uniform-backend-testing for the list of 22 bugs discovered by this change (some of which were generally known without explicit issues, and even if explicitly known I still added the label to record the fact that it was pointed out by this technique).

Resolves #4103 (since it was blocking so many tests against Java). I otherwise tried not to be distracted by fixing bugs even if they appeared easy to fix. :)


Test/VerifyThis2015/Problem3.dfy
```diff
@@ -1,5 +1,4 @@
-// RUN: %dafny /compile:3 /print:"%t.print" /dprint:"%t.dprint" "%s" > "%t"
-// RUN: %diff "%s.expect" "%t"
+// RUN: %testDafnyForEachCompiler "%s" -- --relax-definite-assignment
```
MikaelMayer, line 1: I wasn't expecting Problem3.dfy.expect to disappear. what happened?

Source/TestDafny/MultiBackendTest.cs
```diff
@@ -74,10 +76,20 @@ public class MultiBackendTest {
     // but this was never meaningful and only added maintenance burden.
     // Here we only ensure that the exit code is 0.
 
+    // We also use --(r|b)print to catch bugs with valid but unprintable programs.
```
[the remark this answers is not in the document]
robin-aws: That's exactly the intention, it's another feature that should work for every valid Dafny program but where historically there have been bugs. It's particularly valuable since the current `.doo` file format relies on printing programs as well now. Doing this for every test is the simplest way to make sure existing tests that use these options for the same testing motivation still provide at least as much testing coverage.

I don't think there's added complexity for test authors, but there is a runtime cost when running the tests. As per your other comment I'll evaluate that if it seems expensive.

Source/XUnitExtensions/Lit/OutputCheckCommand.cs
```diff
@@ -185,6 +185,19 @@ private record CheckNotLiteral(string File, int LineNumber, string Literal) : Ch
     return result!;
   }
 
+  public static IList<CheckDirective> ParseCheckFile(string fileName) {
+    var result = File.ReadAllLines(fileName)
+      .Select((line, index) => CheckDirective.Parse(fileName, index + 1, line))
+      .Where(e => e != null).Cast<CheckDirective>()
+      .Select(e => e!)
```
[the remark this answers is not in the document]
robin-aws: I think you're right - I added this because I swear the compiler forced me to at some point, but in the current build configuration there's no warning or error without this line. It's possible that's because of this annotation on the return type parameter of `Cast` that might be new:

```
#nullable disable // there's no way to annotate the connection of the nullability of TResult to that of the source
                TResult
#nullable restore
```

It should be impossible for `e!` to throw by construction so I'm comfortable removing it, especially in testing-only code.

Source/DafnyCore/Compilers/Cplusplus/Compiler-cpp.cs
```diff
@@ -228,7 +228,7 @@ class CppCompiler : SinglePassCompiler {
         throw new UnsupportedFeatureException(tok, Feature.ExternalClasses, String.Format("extern in class {0}", name));
       }
       if (superClasses != null && superClasses.Any(trait => !trait.IsObject)) {
-        throw new UnsupportedFeatureException(tok, Feature.Traits, String.Format("traits in class {0}", name));
+        throw new UnsupportedFeatureException(tok, Feature.Traits);
```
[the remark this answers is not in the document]
robin-aws: It's marginally beneficial, but as long as we have an accurate line number for these errors I didn't think the extra content added much that isn't obvious. It's a bit more work to support these messages, which is worthwhile by already in flight in this PR (which needs non-trivial test fixes): https://github.com/dafny-lang/dafny/pull/2739

Test/VerifyThis2015/Problem3.dfy
MikaelMayer: Ah ok good thanks for the explanations!

robin-aws:
> How much longer does it take now to run the CI?

Comparing to another recent CI run (https://github.com/dafny-lang/dafny/pull/4172), it looks like all I've done is shuffle how the tests are assigned to shards, and the runtime before and after is still 20-27 minutes.

As an aside, it is likely far more tractable to execute a `MultiBackendTest` fully in-process, running the frontend on a program once and then each backend on the resulting resolved and verified program, than it is to switch the more general `DAFNY_INTEGRATION_TESTS_IN_PROCESS` mode on everywhere. It may be worth doing that to save some CI time after this too!

Test/comp/Arrays.dfy.go.expect
```diff
@@ -0,0 +1,96 @@
+0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 
+17
+[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22]
+[2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
```
MikaelMayer, line 4: Why is Go so different starting from here?!

Test/comp/ByMethodCompilation.dfy.expect
```diff
@@ -1,35 +1,3 @@
...
-Recursive(7) = 14
-NonCompilableFunction: 4 7
-Sums: 14 3
-
-Dafny program verifier did not attempt verification
-hello
-four is 4
-Recursive(7) = 14
-NonCompilableFunction: 4 7
-Sums: 14 3
```
MikaelMayer, line 30: NonCompilableFunction and Sums is no longer there.

MikaelMayer: Sorry please ignore this I did not see they were hidden in GitHub

Test/comp/firstSteps/7_Dt_Algebraic.dfy.go.expect
```diff
@@ -0,0 +1,11 @@
+List.Nil
+List.Cons(5, List.Nil)
+(List.Nil, List.Cons(5, List.Nil))
+(List.Cons(5, List.Nil), List.Nil)
+()
+5 List.Nil
+List.Cons(0, List.Cons(1, List.Cons(2, List.Cons(3, List.Cons(4, List.Cons(5, List.Cons(6, List.Nil)))))))
+0 + 1 + 2 + 3 + 4 + 5 + 6 == 21 (once more, that's 21)
+{Berry.Smultron, Berry.Jordgubb, Berry.Hallon}
```
MikaelMayer, line 9: I see your point here :-)
I think we can also leave the printing of sets unspecified (as in theory, there is no order in set), but have a function setToSeq that takes an order on the elements and transforms it into a sequence. No needed now.

Test/exports/FIFO.dfy.py.check
```diff
@@ -0,0 +1,3 @@
+// https://github.com/dafny-lang/dafny/issues/4162
+// CHECK-L: assert "FIFO" == __name__
```
MikaelMayer, line 2: What is CHECK-L ?

Test/unicodechars/comp/NativeNumbers.dfy.js.check
```diff
@@ -0,0 +1 @@
+// CHECK: Error: None of the types given in :nativeType arguments is supported by the current compilation target. Try supplying others.
```
MikaelMayer, line 1:
```suggestion
// CHECK-L: Error: None of the types given in :nativeType arguments is supported by the current compilation target. Try supplying others.
```

MikaelMayer (commented):
Do you confirm that all the tests files were generated by your script except for the ".check" one that I checked manually?

Otherwise, everything looks good so far.

Test/comp/Arrays.dfy.go.expect
robin-aws: The only difference is due to how Go prints strings when `--unicode-char` is `false`:

```
% diff Test/comp/Arrays.dfy.expect Test/comp/Arrays.dfy.go.expect 
88c88
< DDDDDDDDDDagggggaggg
---
> [D, D, D, D, D, D, D, D, D, D, a, g, g, g, g, g, a, g, g, g]
```

That does mean the NONUNIFORM tag on `Arrays.dfy` itself is pointing to the wrong issue though. I am second-guessing the decision to have the conversion script point to https://github.com/dafny-lang/dafny/issues/4108 by default, I think it would be better to force a manual review of any inconsistent output.

Test/comp/firstSteps/7_Dt_Algebraic.dfy.go.expect
robin-aws: Yup, it's definitely a solvable problem in stdlib code (and we should provide good solutions). For testing we may want to transition to using `expect` more and diffing print output less instead.

Test/exports/FIFO.dfy.py.check
robin-aws: https://github.com/stp/OutputCheck#check-l-string - like CHECK but looks for a "literal", exact match (aside from leading or trailing whitespace) rather than a regular expression.

Test/unicodechars/comp/NativeNumbers.dfy.js.check
robin-aws: That actually wouldn't work since the actual output is:

```
Test/comp/NativeNumbers.dfy(14,30): Error: None of the types given in :nativeType arguments is supported by the current compilation target. Try supplying others.
Test/comp/NativeNumbers.dfy(15,30): Error: None of the types given in :nativeType arguments is supported by the current compilation target. Try supplying others.
...
```

CHECK-L requires the whole trimmed line matches, not just part. We could add the source location prefixes, but that would just make this test more fragile, so I opted not to.

MikaelMayer: Ah I did not know. In that, case you might just want to escape the dots since it's a regex :-)

MikaelMayer (approved):
Great way to make our compilers more robust! Thanks for doing that.

PR #4116, opened 2023-06-02
Combine module resolution loops

keyboardDrummer:
Combine the different loops over modules that occur in resolution. There is now a method `LiteralModuleDecl.Resolve` that can be used for basing caching module resolution on.

### DIFF
--- a/Source/DafnyCore/AST/Modules/LiteralModuleDecl.cs
+++ b/Source/DafnyCore/AST/Modules/LiteralModuleDecl.cs
@@ -179,7 +179,6 @@ public class LiteralModuleDecl : ModuleDecl, ICanFormat {
 
     Type.PushScope(tempVis);
     resolver.ComputeIsRecursiveBit(module);
-    resolver.SolveAllTypeConstraints();
     resolver.FillInDecreasesClauses(module);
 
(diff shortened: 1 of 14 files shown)


Source/DafnyCore/AST/Modules/LiteralModuleDecl.cs
```diff
@@ -78,4 +79,119 @@ public LiteralModuleDecl(ModuleDefinition module, ModuleDefinition parent)
 
     return true;
   }
+
+  public void Resolve(Resolver resolver, Program prog, int beforeModuleResolutionErrorCount) {
+    // The declaration is a literal module, so it has members and such that we need
+    // to resolve. First we do refinement transformation. Then we construct the signature
+    // of the module. This is the public, externally visible signature. Then we add in
+    // everything that the system defines, as well as any "import" (i.e. "opened" modules)
+    // directives (currently not supported, but this is where we would do it.) This signature,
```
atomb, line 88: If I'm parsing this correctly, does it mean to say that `import opened` isn't supported? I think that's obsolete, if so.

atomb (approved):
In addition to making caching easier, I think this makes the code easier understand.

davidcok:
This commit causes a regression: https://github.com/dafny-lang/dafny/issues/4144
can we revert it until the regression is fixed?

PR #4134, opened 2023-06-05
Add pre-type resolution machinery

RustanLeino:
This PR adds the API for adding and solving pre-type constraints. These are part of the new type system, but this PR adds only the mentioned API, not the addition of the constraints to be solved.

The main logic for solving the constraints is found in `PartiallySolveTypeConstraints`:

``` csharp
      bool anythingChanged;
      do {
        anythingChanged = false;
        anythingChanged |= ApplySubtypeConstraints();
        anythingChanged |= ApplyEqualityConstraints();
        anythingChanged |= ApplyGuardedConstraints();
        if (makeDecisions) {
          if (DecideHeadsFromBounds(true)) {
            anythingChanged = true;
          } else if (DecideHeadsFromBounds(false)) {
            anythingChanged = true;
          }
        }
      } while (anythingChanged);
```

This is called from `SolveAllTypeConstraints`, which additionally make use of _advice_ (which in the absence of other constraints, chooses `int` instead of `bv73` or `ORDINAL` or `int32` as the type of integer literals, for example):

``` csharp
      PartiallySolveTypeConstraints(null, makeDecisions: false);
      PartiallySolveTypeConstraints(null, makeDecisions: true);
      if (ApplyDefaultAdvice()) {
        PartiallySolveTypeConstraints(null, makeDecisions: true);
      }
      ConfirmTypeConstraints();
```

(This logic replaced the far more complicated logic of the legacy type system, found in the 350-lines long method `PartiallySolveTypeConstraints` in `NameResolutionAndTypeInference.cs`.)


keyboardDrummer:
This type of PR is curious. Without it hooking into existing code, I'm not sure how to review this.

RustanLeino:
@keyboardDrummer It's an attempt to cut up all my new code into smaller pieces. Indeed, this code is not (here) called by anything. Should I perhaps dump the rest of my code into a large PR to give this PR more context?

keyboardDrummer:
> @keyboardDrummer It's an attempt to cut up all my new code into smaller pieces. Indeed, this code is not (here) called by anything. Should I perhaps dump the rest of my code into a large PR to give this PR more context?

I think if the cut-off code doesn't make sense by itself then it's not a good cut-off. If you publish the large PR first, then the reviewer can also help think of a good way of splitting it up, if there is one.

RustanLeino:
I have now published the larger PR as https://github.com/dafny-lang/dafny/pull/4323. It includes the changes in this PR. If it's nice to review this PR separately, we can do that. Otherwise, we can close this PR in favor of https://github.com/dafny-lang/dafny/pull/4323.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+
+    public static string TokToShortLocation(IToken tok) {
+      return $"{System.IO.Path.GetFileName(tok.filename)}({tok.line},{tok.col - 1})";
+    }
+
+    void PrintList<T>(string rubric, List<T> list, Func<T, string> formatter) {
+      if (!resolver.Options.Get(CommonOptionBag.NewTypeInferenceDebug)) {
+        return;
+      }
+      Console.WriteLine($"    {rubric}:");
```
keyboardDrummer, line 109: Please don't use `Console.` anywhere. You can use `resolver.Options.OutputWriter` or `resolver.Options.ErrorWriter` instead.

I would enjoy adding a tiny Roslyn plugin to our codebase to disallow using `Console.` unless you add a pragma, but haven't gotten to that.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+
+    void AddDefaultAdvice(PreType preType, AdviceTarget advice) {
+      Contract.Requires(preType != null);
+      defaultAdvice.Add(new Advice(preType, advice));
+    }
+
+    bool ApplyDefaultAdvice() {
+      bool anythingChanged = false;
+      foreach (var advice in defaultAdvice) {
+        var preType = advice.PreType.Normalize();
```
keyboardDrummer, line 707: Can you move the body of the loop into a method that's inside the class `Advice`, and move that class into a separate file together with `AdviceTarget`?

Also, consider moving `AdviceTarget` inside `Advice` and renaming it to `Target`, so you reference it using `Advice.Target`

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+//-----------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Diagnostics.Contracts;
+using Microsoft.Boogie;
+
+namespace Microsoft.Dafny {
+  public partial class PreTypeResolver {
```
keyboardDrummer, line 15: Please remove the partial. We can add it back in a future PR is if turns out to be warranted there, but I expect it'll be better to pass `PreTypeResolver` around more.

The downside of having a lot of `partial class` declarations is that the fields of this class become spread of many files, and it becomes difficult to both see and limit the size of the class's state.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+        A = a;
+        B = b;
+      }
+
+      public override string ErrorMessage() {
+        return string.Format(ErrorFormatString, A, B);
+      }
+    }
+
+    private List<EqualityConstraint> equalityConstraints = new();
```
keyboardDrummer, line 147: Please move all fields to the top of the class.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+        }
+        if (direction <= 0 && constraint.Sub.Normalize() == tproxy) {
+          yield return constraint.Super;
+        }
+      }
+    }
+
+    // ---------------------------------------- Guarded constraints ----------------------------------------
+
+    private List<Func<bool>> guardedConstraints = new();
```
keyboardDrummer, line 654: The division between equality, subtype and guarded constraints, the last of which seems to contain many types of constraints, seems somewhat arbitrary to me.

The only one of the three that stands out to me are subtype constraints since they're also used as an information store.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+      if (printableContext != null) {
+        PrintTypeInferenceState("(partial) " + printableContext);
+      }
+      bool anythingChanged;
+      do {
+        anythingChanged = false;
+        anythingChanged |= ApplySubtypeConstraints();
+        anythingChanged |= ApplyEqualityConstraints();
+        anythingChanged |= ApplyGuardedConstraints();
+        if (makeDecisions) {
```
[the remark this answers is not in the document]
RustanLeino: Excellent point!

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+    /// Returns "true" if "proxy" is among the free variables of "t".
+    /// "proxy" is expected to be normalized.
+    /// </summary>
+    private bool Occurs(PreTypeProxy proxy, PreType t) {
+      Contract.Requires(t != null);
+      Contract.Requires(proxy != null);
+      return Reaches(t, proxy, 1, new HashSet<PreTypeProxy>());
+    }
+
+    int _reaches_recursion;
```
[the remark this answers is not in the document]
RustanLeino: I think I initially intended this to be a debug-only variable, so perhaps that's why I gave it such a strange name. I've now converted it to a method parameter, renamed it `recursionDepth`, and added a comment.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+      if (_reaches_recursion == 20) {
+        Contract.Assume(false);  // possible infinite recursion
+      }
+      _reaches_recursion++;
+      var b = Reaches_aux(t, proxy, direction, visited);
+      _reaches_recursion--;
+      return b;
+    }
+    private bool Reaches_aux(PreType t, PreTypeProxy proxy, int direction, HashSet<PreTypeProxy> visited) {
+      Contract.Requires(t != null);
```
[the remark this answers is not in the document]
RustanLeino: I removed it altogether. Such non-nullness preconditions are our normal coding style. If there's a possibility that a parameter may be null, I mark it as such.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+    /// caused a change some pre-type proxy).
+    /// </summary>
+    void PartiallySolveTypeConstraints(string printableContext = null, bool makeDecisions = false) {
+      if (printableContext != null) {
+        PrintTypeInferenceState("(partial) " + printableContext);
+      }
+      bool anythingChanged;
+      do {
+        anythingChanged = false;
+        anythingChanged |= ApplySubtypeConstraints();
```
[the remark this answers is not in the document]
RustanLeino: It's rather arbitrary. On the one hand, equality constraints give more information, so they may help reach the end result faster. On the other hand, when subtype constraints are applied, they often turn into equality constraints, so there will need to be another round of equality-constraint solving afterwards. For this reason, I chose to do the subtype constraints first.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+        } else if (tp.Variance == TypeParameter.TPVariance.Co) {
+          AddSubtypeConstraint(arg0, arg1, tok, "covariance would require {0} :> {1}");
+        } else {
+          AddSubtypeConstraint(arg1, arg0, tok, "contravariance would require {0} :> {1}");
+        }
+      }
+    }
+
+    bool DecideHeadsFromBounds(bool fromSubBounds) {
+      // For each proxy, compute the join/meet of its sub/super-bound heads
```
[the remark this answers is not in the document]
RustanLeino: The comment describes the first loop of the method body. I improved the comment and also added one for the second loop.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+      return b;
+    }
+    private bool Reaches_aux(PreType t, PreTypeProxy proxy, int direction, HashSet<PreTypeProxy> visited) {
+      Contract.Requires(t != null);
+      Contract.Requires(proxy != null);
+      Contract.Requires(visited != null);
+      t = t.Normalize();
+      var tproxy = t as PreTypeProxy;
+      if (tproxy == null) {
+        var dp = (DPreType)t;
```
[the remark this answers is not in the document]
RustanLeino: Throwing the exception doesn't protect against bad casts, either.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+        return false;
+      } else {
+        visited.Add(tproxy);
+        return DirectionalBounds(tproxy, direction).Any(su => Reaches(su, proxy, direction, visited));
+      }
+    }
+
+    // ---------------------------------------- Comparable constraints ----------------------------------------
+
+    void AddComparableConstraint(PreType a, PreType b, IToken tok, string errorFormatString) {
```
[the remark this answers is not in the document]
RustanLeino: It may be that such "comparable constraints" won't actually be used. So, I didn't make any change here at this time.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+        return true;
+      } else {
+        // not enough information to determine
+        return false;
+      }
+    }
+
+    // ---------------------------------------- State with error message ----------------------------------------
+
+    abstract class PreTypeStateWithErrorMessage {
```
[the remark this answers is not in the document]
RustanLeino: See response to one of your last questions below.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,886 @@
...
+          AddSubtypeConstraint(arg1, arg0, tok, "contravariance would require {0} :> {1}");
+        }
+      }
+    }
+
+    bool DecideHeadsFromBounds(bool fromSubBounds) {
+      // For each proxy, compute the join/meet of its sub/super-bound heads
+      Dictionary<PreTypeProxy, TopLevelDecl> candidateHeads = new();
+      Dictionary<PreTypeProxy, SubtypeConstraint> constraintOrigins = new();
+      foreach (var constraint in unnormalizedSubtypeConstraints) {
```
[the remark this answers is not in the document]
RustanLeino: Subtype and equality constraints are processed before "guarded constraints", so there can't be just a single `constraints` queue. It could be a priority queue.

There is one other consideration, which is that, sometimes, _all_ of the subset constraints are searched when trying to find a bound for a proxy pre-type. This makes subset constraints special.

It's entirely possible that there's a better data structure (like the separate variables for upper/lower bounds) that would be more efficient. For now, the simple lists are used.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
RustanLeino: I tried to reduce the number of such categories as much as possible, and the current division is what I came up with.

With a priority queue (along your suggestion in another comment above), it may be possible to combine these constraints further.

The way I've been thinking of a "guarded constraint" is that "if the guard evaluates the true, then the 'guarded constraint' turns into an actual constraint". Perhaps a name like "constraint that's may or may not materialize" would be more to the point.

The last category, "post-inference confirmations", is one of the conceptual clarifications from the previous type-inference system in Dafny. Previously, there was an attempt to try to make every piece of information help the inference along, even though some information never did. As a (possibly fictitious) illustration, there could be a constraint "bitvector or integer" and a constraint "bitvector or char", and then one could, in principle, take the intersection of those constraints and figure out that something is a bitvector. In the new type-inference system, those kinds of constraints are made into "post-inference confirmations", which means they don't aid the _inference_ at all, but are just there to be checked after inference has done its thing.

RustanLeino:
@keyboardDrummer Thanks for your many and helpful comments! I've acted or responded to all of them.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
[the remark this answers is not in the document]
keyboardDrummer: For the sake of time we can leave it as is, but my intuition is that it would both be easier to understand and more performant if all of the iterations over the subset constraints would be replaced with separate state.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+namespace Microsoft.Dafny {
+  public class PreTypeInferenceState {
+    public readonly PreTypeResolver PreTypeResolver;
+    private readonly DafnyOptions options;
+
+    private List<SubtypeConstraint> unnormalizedSubtypeConstraints = new();
+    private List<EqualityConstraint> equalityConstraints = new();
+    private List<Func<bool>> guardedConstraints = new();
+    private readonly List<Advice> defaultAdvice = new();
+    private List<System.Action> confirmations = new();
```
keyboardDrummer, line 22: The concept of a confirmation seems too generic, but I'm not sure how to improve it without seeing the consuming code.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+      if (preType is PreTypeProxy) {
+        return false;
+      }
+
+      var pt = (DPreType)preType;
+      var ancestorPt = PreTypeResolver.NewTypeAncestor(pt);
+      var ancestorDecl = ancestorPt.Decl;
+      var familyDeclName = ancestorDecl.Name;
+      switch (check) {
+        case "InIntFamily":
```
keyboardDrummer, line 436: The usage of string literals here seems like a code-smell to me, but without he consuming code I'm not sure what alternative to suggest.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+    void ClearState() {
+      unnormalizedSubtypeConstraints.Clear();
+      equalityConstraints.Clear();
+      guardedConstraints.Clear();
+      defaultAdvice.Clear();
+      confirmations.Clear();
+      PreTypeResolver.allPreTypeProxies.Clear();
+    }
+
+    public void PrintTypeInferenceState(string/*?*/ header = null) {
```
keyboardDrummer, line 84: I would move all the printing code to the bottom of the file, or otherwise sort the methods topologically.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+    // ---------------------------------------- Pre-type inference state ----------------------------------------
+
+    /// <summary>
+    /// Solves or simplifies as many type constraints as possible.
+    /// If "allowDecisions" is "false", then no decisions, only determined inferences, are made; this mode is
+    /// appropriate for the partial solving that's done before a member lookup.
+    /// Returns "true" if anything changed (that is, if any of the constraints in the type-inference state
+    /// caused a change some pre-type proxy).
+    /// </summary>
+    void PartiallySolveTypeConstraints(string printableContext = null, bool makeDecisions = false) {
```
keyboardDrummer, line 38: I would extract the logic related to `makeDecisions` to an outer method. Having boolean parameters that change behavior is a smell, because it ends up with calls like `CanDoAnything(true,true,true,false,false,true)`

```c#
PartiallySettleTypeConstraints(string printableContext = null) {
  bool anythingChanged = false;
  do {
    if (TryResolveTypeProxiesUsingKnownBounds(true)) {
      // something changed, so do another round of Apply... calls below
    } else if (TryResolveTypeProxiesUsingKnownBounds(false)) {
       // something changed, so do another round of Apply... calls below
    } else {
      return;
     }
    anythingChanged = PartiallySolveTypeConstraints(printableContext);
  } while(anythingChanged);
}

bool PartiallySolveTypeConstraints(string printableContext = null) {
  // same as before but without the `if(makeDecisions)` statement, and returning if anythingChanged in any of the loops
}

```

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+      equalityConstraints = new();
+      foreach (var constraint in constraints) {
+        equalityConstraints.AddRange(constraint.Apply(this));
+      }
+      return true;
+    }
+
+    // ---------------------------------------- Comparable constraints ----------------------------------------
+
+    public void AddComparableConstraint(PreType a, PreType b, IToken tok, string errorFormatString) {
```
keyboardDrummer, line 157: I didn't understand your reply to my previous comment, but I don't think this code belongs in this file.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+      }
+    }
+
+    public void DebugPrint(string format, params object[] args) {
+      if (options.Get(CommonOptionBag.NewTypeInferenceDebug)) {
+        Console.WriteLine(format, args);
+      }
+    }
+
+    // ---------------------------------------- Equality constraints ----------------------------------------
```
keyboardDrummer, line 137: I find these topic headers to be a code smell. If you really need them then the code should probably in a separate class. I think you should remove them in this file.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+        // any other reference type implicitly has "object" as a parent, so the height is 1
+        return 1;
+      } else {
+        return 0;
+      }
+    }
+
+    IEnumerable<DPreType> AllSubBounds(PreTypeProxy proxy, ISet<PreTypeProxy> visited) {
+      Contract.Requires(proxy.PT == null);
+      if (!visited.Contains(proxy)) {
```
keyboardDrummer, line 310: Please invert the condition and exit early to reduce indentation

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+              yield return (DPreType)sub;
+            }
+          }
+        }
+      }
+    }
+
+    IEnumerable<DPreType> AllSuperBounds(PreTypeProxy proxy, ISet<PreTypeProxy> visited) {
+      Contract.Requires(proxy.PT == null);
+      if (!visited.Contains(proxy)) {
```
keyboardDrummer, line 329: Please invert the condition and exit early to reduce indentation

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+
+    void AddConfirmation(string check, PreType preType, IToken tok, string errorFormatString) {
+      confirmations.Add(() => {
+        if (!ConfirmConstraint(check, preType, null)) {
+          PreTypeResolver.ReportError(tok, errorFormatString, preType);
+        }
+      });
+    }
+
+    void AddConfirmation2(string check, PreType preType, Type toType, IToken tok, string errorFormatString) {
```
keyboardDrummer, line 405: The suffix `2` doesn't seem to add anything, consider removing it.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
```diff
@@ -316,6 +302,37 @@ public enum Type2PreTypeOption { GoodForInference, GoodForPrinting, GoodForBoth
...
+    /// Otherwise, return "false".
+    /// Note, if either "super" or "sub" contains a type proxy, then "false" is returned.
+    /// </summary>
+    public bool IsSuperPreTypeOf(DPreType super, DPreType sub) {
+      var subAncestors = new HashSet<TopLevelDecl>();
+      ComputeAncestors(sub.Decl, subAncestors);
+      if (!subAncestors.Contains(super.Decl)) {
+        return false;
+      }
+      var s = sub.AsParentType(super.Decl, this);
```
keyboardDrummer, line 316: Please don't use single letter variable names

Source/DafnyCore/Resolver/PreType/PreTypeStateWithErrorMessage.cs
```diff
@@ -0,0 +1,37 @@
...
+// Copyright by the contributors to the Dafny Project
+// SPDX-License-Identifier: MIT
+//
+//-----------------------------------------------------------------------------
+
+using System;
+using System.Diagnostics.Contracts;
+
+namespace Microsoft.Dafny {
+  abstract class PreTypeStateWithErrorMessage {
```
keyboardDrummer, line 12: I didn't understand your reply to my previous comment, but the name of this class doesn't make sense to me, both on its own and in the way it is used. Constraints and state seem like two different things to me.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
```diff
@@ -0,0 +1,589 @@
...
+//
+//-----------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Diagnostics.Contracts;
+
+namespace Microsoft.Dafny {
+  public class PreTypeInferenceState {
```
keyboardDrummer, line 14: This class does more than hold state. Consider renaming to `PreTypeInferencer` ?

Could you add a piece of prose in a summary comment above it that explains some of its capabilities?

keyboardDrummer (commented):
Thanks for applying the feedback.

However, the PR is not yet in a state I would normally approve. The main problem is that it's not a self-contained whole: it adds dead code and doesn't have tests for the code it's adding. Reviewing code that's not called is difficult because I can't judge the signatures well.

Secondly, even though we don't have performance tests, I think the code should keep performance complexity within `O(NLogN)`, and I suspect this implementation doesn't do that. I see bad performance complexity as a bug that only occurs when you write large codebases, as opposed to other bugs that trigger when you combine specific language constructs.

I can approve it with the idea that it's faster to get to the ideal state by merging and then improving the code instead of reworking it first, but then I think it would be good to ask someone with a high bar for maintainability to approve as well.

Or, if you want to improve this PR, what you could do is pick a small file from our test suite that resolves with this new resolver in the bigger PR, and then add just enough constraints from the bigger PR to this one so the test file passes.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
RustanLeino: I moved method `DebugPrint` to the end of the file. The others are already sorted topologically, and I think it makes sense to have those methods near where other operations that concern the entire state is placed.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
RustanLeino: I can change these string literals to an `enum`, but I prefer to do that in the large PR that that also contains the call sites.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
RustanLeino: So, you're suggesting that the 3 applications that would go into `PartiallySolveTypeConstraints` would always run to a fixpoint before attempting another decision? In other words, there would still be a loop inside `PartiallySolveTypeConstraints`, right? I don't know if that behavior would compare with the current behavior, but I think the code is clearer the way it is.

I also don't agree that parameterized behavior of methods is in general a code smell.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
RustanLeino: I removed the code.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
RustanLeino: I renamed it (and the file) `PreTypeConstraints` and added a summary description.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.State.cs
keyboardDrummer: > So, you're suggesting that the 3 applications that would go into PartiallySolveTypeConstraints would always run to a fixpoint before attempting another decision?

Yes. I don't understand the process well enough to understand the implication. Could fixpointing before making additional decisions lead to better decisions?

keyboardDrummer (approved):

Source/DafnyCore/Resolver/PreType/PreTypeConstraints.cs
```diff
@@ -0,0 +1,546 @@
...
+//
+//-----------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Diagnostics.Contracts;
+
+namespace Microsoft.Dafny {
+  /// <summary>
```
keyboardDrummer, line 14: I think this comment should try to give me enough knowledge, possibly through links to other documents, to be able to easily read and understand this class, which it doesn't do for me yet.

Could you explain why we have a constraint solving and a decision phase?

RustanLeino: Excellent question. In this use of the words, "constraint solving" just sorts out things that follow from the given constraints, whereas "decisions" makes choices between two or more ways to proceed. For example, if `a` is known to be an integer, then the constraint imposed by a program expression `a == b` says that `b` has to be an integer as well. In contrast, an assignment like `c := 10;` has a choice in what type to give to the literal `10` and `c`, because the type could be `int`, `bv32`, `ORDINAL`, or many other things. Once a decision (that is, a choice) has been made, the rest of the type inference must abide by that decision, so it's good to precede decisions with constraint solving.

The reason there is sometimes constraint solving that is _not_ followed by making decisions (in order words, what the `PartiallySolve...` method does) is that there is a small number of Dafny expressions where the type inference wants to look at the consequences of the constraints laid down so far, but without making any decisions. The prime example of this is member lookup, like in the expression `o.M`. The old type inference in Dafny always insisted on having the ("head" part of the) type of `o` when encountering `o.M`; if it couldn't determine this type, then it would immediately give an error. To determine the type of `o` in this case, one typically needs constraint solving. If that's sufficient, then it's best not to make any hasty decisions at that time.

In the new type inference, there is a general facility of "guarded constraints", which are constraints waiting to happen. Using this facility, I have changed many of the old ways of urgently needing a type. Essentially, this means that the resolver can continue with other expressions and then come back to a guarded constraint, at a time when, hopefully, there is enough inferred information to evaluate the "guard". If we would change to using such guarded constraints everywhere in the type inference, then (to answer your question) we would not need a `PartiallySolve...` method. But I haven't changed the way that member expressions like `o.M` are handled (or the way that the source expression in a `match` is handled). It would be good to change type inference to use the guarded constraints for those, too, but I wanted to see that everything else worked before embarking on those large changes. Thus, for now, we still need `PartiallySolve...`. But in the future, I hope we can improve both the code and the inference itself by changing those places to use guarded constraints, too.

Here is an example program that we don't accept today (neither with the old type inference nor with `/typeSystemRefresh:1` in PR now), but that we could accept in the future if we changing member lookup to use a guarded constraint:

``` dafny
class Cell { var data: int }

method M() {
  var c := null;
  for i := 0 to 10 {
    if c != null {
      c.data := c.data + 1; // this currently gives an error, saying it doesn't know what .data is
    }
    if i == 7 {
      c := new Cell; // this line, which comes after c.data above, is what's needed to infer the type of c
    }
  }
}
```

PR #4136, opened 2023-06-05
Remove compile modules by letting expect contracts produce resolved code

keyboardDrummer:
Step towards caching resolution


Source/DafnyCore/Verifier/Translator.TrStatement.cs
```diff
@@ -505,7 +505,6 @@ public partial class Translator {
         // try/recover statements are currently internal-only AST nodes that cannot be
         // directly used in user Dafny code. They are only generated by rewriters, and verifying
         // their use is low value.
-        throw new NotSupportedException("Verification of try/recover statements is not supported");
```
keyboardDrummer, line 508: `RunAllTestsMainMethod` runs post resolution, before verification, and no longer changes just the CompileModules, so the try/recover statements end up in verified code.

Running that transformation after verification causes slight changes in warning/error behavior, since then diagnostics from `RunAllTestsMainMethod` are reported after resolution.

Source/DafnyCore/Rewriters/ExpectContracts.cs
```diff
@@ -232,62 +304,41 @@ private class CallRedirector : TopDownVisitor<MemberDecl> {
...
+          CalledWrappers.Add(newTarget);
         }
       }
 
       return true;
     }
   }
 
-  internal override void PostCompileCloneAndResolve(ModuleDefinition moduleDefinition) {
-    foreach (var topLevelDecl in moduleDefinition.TopLevelDecls.OfType<TopLevelDeclWithMembers>()) {
```
atomb, line 248: I'm so happy to see this code go away! It made me unhappy to write it.

Source/DafnyCore/Verifier/Translator.TrStatement.cs
atomb: It seems to me that it'd be better to run it after verification, even if it changes warning/error behavior, since here we're adding some code to what will be verified and then just skipping verification of it. That seems fragile to me, and like it could potentially introduce soundness issues. Am I missing something?

atomb (changes requested):
This is a wonderful simplification (and I especially like that it removed the messiest part of the runtime contract checking code). I have one thing about try/recover that I'd like to discuss before merging it, though.

Source/DafnyCore/Verifier/Translator.TrStatement.cs
keyboardDrummer: Previously the code added by `RunAllTestsMainMethod` was entirely unverified, and after this PR it is verified except for the try/recover statement since we didn't implement verification for it. Doesn't that make it more instead of less sound?

atomb (approved):
Looks great!

PR #4142, opened 2023-06-07
Extract classes from Resolver.cs and NameResolutionAndTypeInference.cs

keyboardDrummer:


alex-chew:
LGTM up to and including commit e5c3b80, and I will come back to review the remaining few commits.

Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
```diff
@@ -5852,6 +5852,8 @@ public record ResolveTypeReturn(Type ReplacementType, ExprDotName LastComponent)
       return rr;
     }
 
+    public record MethodCallInformation(IToken Tok, MemberSelectExpr Callee, List<ActualBinding> ActualParameters);
```
alex-chew, line 5855: Was the `ObjectInvariant` method not being used, or should we add it to this?

alex-chew (commented):
Mostly LGTM. The indentation changes in XConstraint seem to greatly confuse the diff algorithm, and by extension, me...

Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
keyboardDrummer: `[ContractInvariantMethod]` is a Noop, so indeed `ObjectInvariant()` was purely for documentation purposes. I don't think we should keep it though.

keyboardDrummer:
> The indentation changes in XConstraint seem to greatly confuse the diff algorithm, and by extension, me...

I don't recall making any changes in XConstraint, besides extracting it to a separate file which changed the indentation.

Source/DafnyCore/Resolver/Resolver.cs
```diff
@@ -5818,6 +4612,7 @@ public class ResolveTypeOption {
     /// context. 
     /// </summary>
     public void CheckLocalityUpdates(Statement stmt, ISet<LocalVariable> localsAllowedInUpdates, string where) {
+      // TODO it looks like this method has no side-effects and doesn't return anything.
```
atomb, line 4615: Mysterious. We should probably investigate that more deeply at some point, so I'm adding a comment to make it slightly more visible. Given that this is a purely refactoring PR, I don't think we need to include resolving it here, though.

Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
atomb: Yeah, it'd be nice to start taking more advantage of those features in this code base.

atomb (approved):
I haven't compared the source and destination text of the classes you extracted, but assuming those are all pure extractions I think this looks good.

PR #4145, opened 2023-06-07
Fix: Resource units never negative and 3 digit precision

### LINKED ISSUE #4143: Resource unit's precision is not constant
When I hover over a method using more than 1M resource unit, I don't get any digits after the 1, so the next precision is 2M RU, from simple to double !
However, when the resource unit count is less than that, I get for example 872K RU.
I would like to get at least 2 or 3 digit precision every time.

MikaelMayer:
This PR fixes #4143 and also another reported problem when resource units are greater than MAX_INT, they were displayed negatively


Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -313,20 +313,34 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
       return information;
     }
 
-    private string FormatResourceCount(int nodeResourceCount) {
+    public static string FormatResourceCount(long nodeResourceCount) {
       var suffix = 0;
+      var fractional = 0;
+      // First, round up nodeResourceCount to 3 significant digits.
```
keyboardDrummer, line 319: Could you introduce a `roundToDigits` function so we don't need the comment? Also, would another way of rounding be to divide by `nDigits - 3`, then to do `Math.floor(x + 0.5)`, and then multiply by `nDigits - 3` ?

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -313,20 +313,34 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
...
+      var nDigits = ("" + nodeResourceCount).Length;
+      if (nDigits > 3) {
+        var toRemove = (long)Math.Pow(10, nDigits - 3);
+        nodeResourceCount += toRemove / 2;
+        nodeResourceCount -= (nodeResourceCount % toRemove);
+      }
+
       while (nodeResourceCount / 1000 >= 1 && suffix < 4) {
+        fractional = (int)(nodeResourceCount % 1000);
```
keyboardDrummer, line 328: Why do you assign to fractional in the loop instead of after it?

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -313,20 +313,34 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
...
         nodeResourceCount /= 1000;
         suffix += 1;
       }
+      var nodeResourceCountStr = $"{nodeResourceCount:n0}";
+      var fractionalStr = nodeResourceCountStr.Length >= 3 || suffix == 0 ?
```
keyboardDrummer, line 333: Is this the same `3` as before? Can we put it in a variable?

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -313,20 +313,34 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
...
+        "" :
+        ($".{fractional:000}").Remove(4 - nodeResourceCountStr.Length);
```
keyboardDrummer, line 335: Where does the `4` come from?

keyboardDrummer (approved):

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: Because it depends on a number that is modified after this assignment.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: > Could you introduce a roundToDigits function so we don't need the comment? 

Done

> Also, would another way of rounding be to divide by nDigits - 3, then to do Math.floor(x + 0.5), and then multiply by nDigits - 3 ?

image

Indeed. I'll keep my version because it contains only one modulo compared to two division/multiplication in base 10 :-)

keyboardDrummer (approved):

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -313,20 +313,41 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
...
+      if (nDigits > SignificativeDigits) {
+        var toRemove = (long)Math.Pow(10, nDigits - SignificativeDigits);
+        nodeResourceCount += toRemove / 2;
+        nodeResourceCount -= (nodeResourceCount % toRemove);
+      }
+
+      return nodeResourceCount;
+    }
+
+    public static string FormatResourceCount(long nodeResourceCount) {
```
keyboardDrummer, line 329: This is so much complicated code, even though .NET has quite a lot of number formatting utilities. Wouldn't `nodeResourceCount.ToString("G4")` give you more or less the same result? https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings

keyboardDrummer (approved):

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: It would but without the prefixed K, M, G which are very handy.

keyboardDrummer: Is that easier to read than `e+XXX` notation? 

It might be easier to compare `8.72e+005` with `8.72e+006` than it is to compare `872K` with `8.72M`

MikaelMayer: That would be much simpler to encode, true. But is it more readable?

For me the scientific notation is like.... 5.25e005 RU? What is that? Oh that's just 525K RU? Should have told earlier :-) It's even worse when we switch between the scientific notation to the normal notation, it makes it hard to compare. For example 525*10 = 5.25e003. Hard to see where the 003 comes from or compare the two right away.
Or if we delayed the scientific notation, 525, 5250, 52500, 5.25e005.
On the other side, when the unit is the same, (which happens most of the times),  the number of digits before the comma immediately tells about the magnitude, it's a visual comparison, not something to be interpreter from numbers like 5, 6.

Here are some comparisons:
* On the Windows file system as of today, everything is expressed in KB, even if it's millions.
* Total valuations of businesses are always expressed in terms of K (kilos) M (millions) or B (billions)

Happy to discuss this further if you think the current unit system can be enhanced.

keyboardDrummer: > It's even worse when we switch between the scientific notation to the normal notation, it makes it hard to compare.

I agree, I would always do the scientific one.

> For me the scientific notation is like.... 5.25e005 RU? What is that? Oh that's just 525K RU? Should have told earlier :-)

I don't think anyone using resource counts in Dafny will have trouble reading scientific notation.

I you want to compare magnitude, comparing e00X with e00Y, versus comparing two instances of K/M/G/etc. + a number for everything between the 1000s, than the former seems simpler to me.

PR #4149, opened 2023-06-08
Add test and fix for 4144

### LINKED ISSUE #4144: Resolution regression
### Dafny version

28397e8

```dafny
module P {
  predicate m() reads {}
} 
module Q refines P {
  predicate m() { true }
}
```

```code
dafny resolve

The above example ran without error in past versions of dafny, 
but now fails with recent commit 28397e8 on 6/6.
```

### What happened?

Recent commit broke at least one valid example.

Mac

keyboardDrummer:
Fixes #4144


davidcok (approved):

RustanLeino:
This is not the right fix. For example, the problem still remains if you change the declaration of `P.m()` to:

``` dafny
  predicate m() reads []
```

or

``` dafny
  predicate m() reads multiset{}
```

The problem is that Dafny, in order:
* Resolves the `reads` clause of `P.m()` and infers the type of `{}` to be `set<object?>`.
* Checks to see that types of all expressions were inferred, which they were.
* Notices the absence of a `decreases` clause of `P.m()` and thus fills one in (based on the `reads` clause).
* Merges `P` into `Q`. In more detail, it calls `RefinementTransformer.MergeClass` for the `nwMember is Function` case, which on L853 makes a call to `CloneFunction`, which by

    ``` csharp
    refinementCloner.CloneSpecExpr(previousFunction.Decreases)
    ```

    copies the automatically generated `decreases` clause from `P.m()`. This is the part that's wrong. It should not clone the `decreases` clause if it consists of `AutoGeneratedExpression`s.
* By design, the previous step clones the expression, but not the type. Next, Dafny infers types for `Q`, including for the `decreases` clause that was cloned. This will not infer a type for the `{}` that's part of the `decreases` clause.
* Dafny checks to see that types of all expressions in `Q` were inferred, but they were not, and hence the error mentioned in the bug report is reported.

So, note that the `{}` in `reads {}` is expected to be a set of references, so Dafny happily infers its type to be `set<object?>`. However, an expression `{}` occurring in another place---here, in a `decreases` clause---is inferred to be a set of an unknown element type.

keyboardDrummer:
Are you sure letting the generatedness determine whether to copy something during refinement is the right way to go? The whole concept of autogenerated seems vague to me. You could add some new construct to the Dafny language that causes lots of code to be generated early on in resolution, which does need to be copied over for refinement.

Why not blame the decreases clause generator for generating Dafny code that by itself doesn't resolve? That seems like a more robust solution to me. We just have to extend the decreases clause generator to the other causes that it shouldn't generate, like the ones you provided.

RustanLeino:
The idea of starting over from the syntactic things that were given in the original module seems like a fine principle to me. To do that, one needs to keep track of what was auto-generated. (I think Dafny uses two ways of indicating auto-generation: `AutoGeneratedToken` and `AutoGeneratedExpression`. I don't know why, or if, we need both instead of just one.)

I think some auto-generated things happen every time, so if we don't omit them from the clone, then a second refinement (that is, a module `B` refines `A` and module `C` in turn refines `B`) would end up with two copies of them. (`:autoReq` is such an example, I believe, but we probably won't keep `:autoReq` for long.) There are also some auto-generated things that depend on the presence of absence of other things. An example is `:autocontracts`, which looks for `{:autocontracts false}` in some places. Those may or may not work for a second refinement, if the first refinement manually adds an `{:autocontracts false}`. But perhaps these sorts of things are better off following some stricter discipline than what we're doing today, which could then mean that the "don't clone auto-generated things" principle is not the best.

Your suggestion also has a good principle: Make sure any auto-generated code contains information to be resolved by itself. For the frames, this means either disallowing the empty collections, omitting empty collections, or adding the necessary type information. I wonder how many more cases there would be? For decreases generation, would it perhaps be possible to always turn every clause `reads S` into a `decreases` term `var s: set<object> := S; s`? (Here are two test cases to include, if we go that way: `reads {}` and `reads { var o :| true; o }`.)

keyboardDrummer:
> The idea of starting over from the syntactic things that were given in the original module seems like a fine principle to me. To do that, one needs to keep track of what was auto-generated.

I don't think you need to keep track of that. You can do refinement cloning before all module resolution, so refinement only clones the parsed modules. That way we don't have to think about how to track what was changed after parsing, which seems fragile to me.

What I don't like about that is that it breaks performance. If you then have a long chain of refinement, you will end up doing N^2 amount of resolution work instead of N. 

However, because of meta-programming constructs like `{:autocontracts}`, I'm not sure we can do better. You can't apply `{:autocontracts}` on a class and then use the result of that when making a clone for refinement, because the refining class might add extra fields. Do I understand that correctly?

I guess for now it might be better to let refinement cloning happen before module resolution, it seems like it was designed with that behavior in mind and that will avoid any bugs, even though it means module refinement is relatively slow.

keyboardDrummer:
Some testing shows that on 4.1.0, refinement and autoReq, and refinement and autocontracts, already don't play well together. Made two issues: https://github.com/dafny-lang/dafny/issues/4156 and https://github.com/dafny-lang/dafny/issues/4157.

RustanLeino:
> You can do refinement cloning before all module resolution, so refinement only clones the parsed modules. 

That would be great. (I had assumed refinement cloning wanted a resolved module to start with, but if it doesn't, then doing the cloning directly from parsed module is nicer.)

> ... Similarly for {:autoReq}, the refining function might add extra calls which changes the behavior of {:autoReq}. Do I understand that correctly?

Yes, you're understanding that correctly. I'm not so fond of `:autoReq`, so I would not be unhappy if we remove it altogether. As for `:autocontracts`, it can be convenient, but I'd say it's still in a prototype stage, because we've never been clear about what's supposed to happen in many cases (including refinements modules that use `:autocontract`).

> ... What are your thoughts?

This first-do-refinement-cloning approach seems like a good direction to me.

keyboardDrummer:
> That would be great. (I had assumed refinement cloning wanted a resolved module to start with, but if it doesn't, then doing the cloning directly from parsed module is nicer.)

What about the O(N^2)ness of that, does that not bother you?

I think we should identify which transformations conflicts when done inside the same module transformation loop, such as refinement cloning and `{:autocontract}`, and put them in separate module transformation loops. We can even have a mechanism that allows transformations to specify which others they conflict with and have that determine the amount of module transformation loops. We could do caching of module transformations in each of those loops to get the feeling of O(1) resolution time when working in the IDE.

PR #4151, opened 2023-06-08
Cache resolution

keyboardDrummer:
Fixes #


Source/DafnyLanguageServer/Language/Symbols/DafnyLangSymbolResolver.cs
```diff
@@ -16,53 +16,53 @@ namespace Microsoft.Dafny.LanguageServer.Language.Symbols {
...
 
-    public DafnyLangSymbolResolver(ILogger<DafnyLangSymbolResolver> logger, ITelemetryPublisher telemetryPublisher) {
-      this.logger = logger;
+    public DafnyLangSymbolResolver(ILoggerFactory loggerFactory, ITelemetryPublisher telemetryPublisher) {
+      this.loggerFactory = loggerFactory;
+      logger = loggerFactory.CreateLogger<DafnyLangSymbolResolver>();
       this.telemetryPublisher = telemetryPublisher;
     }
 
-    public CompilationUnit ResolveSymbols(TextDocumentItem textDocument, Dafny.Program program, out bool canDoVerification, CancellationToken cancellationToken) {
```
keyboardDrummer, line 28: `canDoVerification` was unused

Source/DafnyCore/AST/SystemModuleManager.cs
```diff
@@ -18,8 +22,42 @@ public class SystemModuleManager {
...
+    get {
+      if (hash == null) {
+
+        var tupleInts = tupleTypeDecls.Keys.Select(tuple => {
+          var vector32 = new BitVector32();
+          for (var index = 0; index < tuple.Count; index++) {
+            vector32[index] = tuple[index];
+          }
+          return vector32.Data;
+        });
```
alex-chew, line 42: I think the hash function makes sense to me (and appears safe), but can you clarify what the keys of `tupleTypeDecls` represent and/or how they're derived?

Source/DafnyLanguageServer.Test/Synchronization/IncludeFileTest.cs
```diff
@@ -69,7 +69,6 @@ public class IncludeFileTest : ClientBasedLanguageServerTest {
   [Fact]
   public async Task MethodWhosePostConditionFailsAndDependsOnIncludedFile() {
     var temp = (Path.GetTempFileName() + ".dfy").Replace("\\", "/");
-    Console.WriteLine("temp file is: " + temp);
```
alex-chew, line 72: Just curious - was this causing issues or was it just a leftover debug artifact?

alex-chew (commented):
I've reviewed the majority of the files but I'll need to come back to review the rest tomorrow.

Source/DafnyLanguageServer.Test/Synchronization/IncludeFileTest.cs
keyboardDrummer: The latter

Source/DafnyLanguageServer/Language/CachingResolver.cs
```diff
@@ -0,0 +1,73 @@
...
+      var dependencyHash = hashes[dependency];
+      hashAlgorithm.TransformBlock(dependencyHash, 0, dependencyHash.Length, null, 0);
+    }
+
+    hashAlgorithm.TransformBlock(cache.Builtins!.MyHash, 0, cache.Builtins!.MyHash.Length, null, 0);
+    var parseHash = moduleDeclaration.CloneId.ToByteArray();
+    hashAlgorithm.TransformFinalBlock(parseHash, 0, parseHash.Length);
+
+    return hashAlgorithm.Hash!;
+  }
```
alex-chew, line 72: Can you write some documentation about what this is doing? It's not quite clear to me (for example) why the order of dependencies' hashes doesn't need to be canonicalized (for example, by sorting `moduleVertex.Successors` or by sorting the resulting hashes), and why we can assume that `hashes[dependency]` exists.

Source/DafnyLanguageServer/Language/DafnyLangParser.cs
```diff
@@ -60,23 +60,23 @@ public sealed class DafnyLangParser : IDafnyParser, IDisposable {
             new(errorReporter.Options, document.Uri.ToUri(), document.Content)
           },
           errorReporter, cancellationToken);
+        cachingParser.Prune();
         return result;
       }
       finally {
         telemetryPublisher.PublishTime("Parse", document.Uri.ToString(), DateTime.Now - beforeParsing);
-        cachingParser.Prune();
```
alex-chew, line 67: Why not prune the cache even in case of parse errors?

alex-chew (commented):
Mostly LGTM, though I left some questions about a couple snippets related to module/ModuleDecl hashing.

Source/DafnyLanguageServer/Language/DafnyLangParser.cs
keyboardDrummer: `ParseFiles` doesn't throw an exception if there are parse errors, but it does do that if parsing was cancelled, which means some of the includes are likely not to be traversed which would change the effect of pruning, which will prunes any entry that wasn't hit since the last prune.

Source/DafnyLanguageServer/Language/CachingResolver.cs
keyboardDrummer: The order of `moduleVertex.Successors` currently depends on the order that symbols have in the source files, so it's fixed. However, you could imagine a future change to `moduleVertex.Successors` that turns it into an unordered datatype like a set, so it might be better to already order it here. Made that change.

> why we can assume that hashes[dependency] exists.

Depends on how ProgramResolver calls `ResolveModuleDeclaration`, but I have changed `CachingResolver` so it doesn't depend on that.

Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
```diff
@@ -5323,7 +5323,7 @@ public record ResolveTypeReturn(Type ReplacementType, ExprDotName LastComponent)
           }
         }
         r = new IdentifierExpr(expr.tok, v);
-      } else if (currentClass is TopLevelDeclWithMembers cl && classMembers.TryGetValue(cl, out var members) && members.TryGetValue(name, out member)) {
+      } else if (currentClass is TopLevelDeclWithMembers cl && GetClassMembers(cl)?.TryGetValue(name, out member) == true) {
```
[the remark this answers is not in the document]
keyboardDrummer: It is necessary to transform the type `bool?` to `bool`

alex-chew (approved):
LGTM

PR #4155, opened 2023-06-08
feat: Run Python byte-code compiler during `/compile:1`

fabiomadge:


robin-aws (changes requested):
Great! The only change I'm requesting is that this doesn't close those two issues, just lets us take the soundness labels off. I think it's better to leave them open since they are still cases where a valid Dafny program produces code that can't execute, even if they are easy to avoid in practice. Better to have the issues open so folks can find them and understand they are known issues IMO.

robin-aws (approved):

PR #4158, opened 2023-06-09
Expand test for decreases clause inference and module refinement

### LINKED ISSUE #4156: Refinement and autoReq don't work well together
### Dafny version

4.1.0

```dafny
// RUN: %baredafny verify %args --rprint:"%t.rprint" "%s" > "%t"
// RUN: %OutputCheck --file-to-check "%t.rprint" "%s"
// Two 'requires x > 3' occur in 'CalledFunction'
// And two should be generated in 'RefinedFunction'
// So there should be no fifth
// CHECK: requires x > 3
// CHECK: requires x > 3
// CHECK: requires x > 3
// CHECK: requires x > 3
// CHECK-NOT: requires x > 3

module Refined {
  function {:autoReq} RefinedFunction(x: int): int
}

module Refining refines Refined {
  function RefinedFunction(x: int): int {
    CalledFunction(x)
  }
  
  function CalledFunction(x: int): int
    requires x > 3 {
    2
  }
}

module RefiningSquare refines Refining {

}
```

```code
The above test fails because in RefiningSquare, RefinedFunction has 'requires x > 3' twice
```

### What happened?

_

Linux

keyboardDrummer:
Fixed #4156


RustanLeino (changes requested):
I'm actually a bit surprised that this change has any effect at all. But even so, three problems remain:

One problem is that the `decreases` clause generated from `multiset{}` does not have a trigger. The trigger should be the `_s2s_0 in multiset{}` expression that's part of the generated set comprehension.

The idea of adjusting the generated `decreases` clause (as you argued in PR #4149) is growing on me. That is, resolution of `reads` clauses is a little special because it's specifically looking for expressions that can be turned into a set of `object`. This should then be taken into account when generating a `decreases` clause from the `reads` clause. But--and here lies the second problem--there are more cases to consider than the three given in the new tests, because `reads` clauses also allow functions to be listed (in particular, each expression in a `reads` clause can be of type `object`, collection of `object`, or function to a collection of `object`). These need to be considered, too. Here are some tests for them:

``` dafny
abstract module P {
  predicate k(o: object) reads (() => {o})
  predicate l(o: object) reads ((x: int) => [o])
  predicate n(o: object) reads ((y: real, z: real) => multiset{o})

  predicate k'() reads (() => {})
  predicate l'() reads ((x: int) => [])
  predicate n'() reads ((y: real, z: real) => multiset{})
} 

module Q refines P {
  predicate k(o: object) { true }
  predicate l(o: object) { true }
  predicate n(o: object) { true }

  predicate k'() { true }
  predicate l'() { true }
  predicate n'() { true }
}
```

A possible way out here is to never support the function-to-object-collection expressions in the reads-to-decreases clause generation (which I think would be fine).

The third problem is that removing the empty-set literal, as the change in PR #4149 does, changes the behavior of Dafny. The following program illustrates:

``` dafny
predicate m()
  reads {}
{
  m'({}, 0)
}

predicate m'(s: set<object?>, j: nat)
  reads {}
  decreases s, j // note, explicit decreases clause
{
  if s == {} then true else m()
}
```

Previously, Dafny inferred `decreases {}` for `m()`, which makes this program verify. But not inferring any `decreases` clause for `m()` causes this program to not verify. Clearly, my program here is rather artificial, but I don't see any good reason to make a special case for the empty-set literal. (I'd prefer generating `var empty: set<object> := {}; empty`.)

keyboardDrummer:
```
predicate m()
  reads {}
{
  m'({}, 0)
}

predicate m'(s: set<object?>, j: nat)
  reads {}
  decreases s, j // note, explicit decreases clause
{
  if s == {} then true else m()
}
```

It's hard for me to decide how I would like the decreases inference for this program to work. Indeed the previous inference that inferred `decreases {}` for `m` works, but that's only because the `reads {}` clause is there which isn't needed in the first place, so it seems coincidental. In general, I expect generating decreases clauses on literals not to be an effective way of passing the termination check.

That said, I'll bring back the old behavior.

> I'd prefer generating `var empty: set<object> := {}; empty`

Out of scope: it would be nice if we could skip introducing the variable, to just write `{}: set<object>`. This could be a compiler only expression construct, or maybe even a user facing one, if the syntax still fits in our existing grammar.

> One problem is that the decreases clause generated from multiset{} does not have a trigger. The trigger should be the _s2s_0 in multiset{} expression that's part of the generated set comprehension.

I think I've done this for the set comprehension expressions generated by decreases clause generation, but not in `BoundsDiscovery.FrameArrowToObjectSet`, where it seems useful as well. I haven't done this because I don't understand why these trigger annotations have to be added in the first place.

RustanLeino (dismissed):
Good work! It's interesting that, in the end, the fix consisted of using `NormalizeExpand()` a couple of times, adding the `IsEmptySet` disjunct to one of the guards, and filling in the trigger explicitly.

In the Boogie code we now generate, I investigated three things:
* One could imagine generating `set x | x in S` instead of `set x | x in S :: s`. However, the current facility for doing so creates an unresolved set comprehension and we want a resolved one. It does not seem worth it to add another `SetComprehension` constructor, since the extra `:: s` makes no difference other than when pretty printing the Dafny code.
* One could imagine that Dafny's trigger-selection machinery would pick out the trigger in `set x | x in S :: s` automatically. If so, we wouldn't have to generate one explicitly here. Dafny does do this if `S` is a sequence, but it does not do it if `S` is a set display or multiset display. The reason for that is that the translator rewrites such expressions into a disjunction of equalities; for example, `x in {a, b, c}` is rewritten into `x == a || x == b || x == c`, and--relevant for our purposes here--`x in {}` is rewritten into `false`. Such expressions cannot be used as triggers, so Dafny instead generates a "no trigger" warning. To avoid such warnings, we need to generate the triggers explicitly, which this PR now does.
* The set comprehensions generated from the _functions_ in the reads clauses don't have good triggers. However, this was the case before this PR, too, so the only difference is that there's now one such warning for the reads clause and one such warning for the inferred decreases clause. That seems fine.

Regarding your "out of scope" comment, I agree. It would be nice if Dafny supported more `as` expressions, like `{} as set<object?>`.

robin-aws (approved):
Re-approving after test fix from master merge

PR #4167, opened 2023-06-12
Improve errors on includes

keyboardDrummer:
### Changes
- Don't show a chain of errors of the form 'the included file ... contains errors', when really there is one at the bottom that has errors and the other fail to resolve because of that, with errors like `not resolving module '{0}' because there were errors in resolving its import`.

### Caveat
- To make the implementation simpler, only one error of the form 'the included file ... contains errors' is shown. Could be improved when there are multiple leaf files with errors.

### Testing
- Added a unit test


Source/DafnyLanguageServer.Test/Synchronization/IncludeFileTest.cs
```diff
@@ -25,17 +25,17 @@ public class IncludeFileTest : ClientBasedLanguageServerTest {
...
+include ""./includesSemanticError.dfy""
+include ""./semanticError.dfy""
 ".TrimStart();
     var documentItem = CreateTestDocument(source, Path.Combine(Directory.GetCurrentDirectory(), "Synchronization/TestFiles/test.dfy"));
     await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
     var diagnostics = await diagnosticsReceiver.AwaitNextDiagnosticsAsync(CancellationToken);
     Assert.Single(diagnostics);
     Assert.Contains("the included file", diagnostics[0].Message);
-    Assert.Contains("syntaxError.dfy", diagnostics[0].Message);
+    Assert.Contains("semanticError.dfy", diagnostics[0].Message);
```
RustanLeino, line 38: I don't understand what this does. Does it check for the mention of `"semanticError.dfy"` in the error message? If that's all it does, then how does that test whether or not only one error message is being produced, as promised by the PR description?

stefan-aws (commented):
Can confirm this removes the "Error: the included file .../Source/DafnyLanguageServer.Test/Synchronization/TestFiles/semanticError.dfy contains error(s)" error message.

PR #4168, opened 2023-06-12
Fix: Better error message when incorrect number of out parameters

### LINKED ISSUE #3835: Confusing message for out parameters for new users
We got several confused new Dafny users that, in the following program:

```
method Callee() returns (i: int) {
  i := 0;
}
method Caller() {
  Callee();
}
```
would not understand the error message
```
wrong number of method result arguments (got 0, expected 1)
```
because they are passing the correct number of arguments to the method.

What is confusing is actually
* They skip the word "result"
* In no other language I know is there such a thing of returning multiple values (not out parameters), so this is very novel.
* We are calling these parameters which should be to the left of the method "result arguments", which is weird. Arguments are passed usually to the right of the method name inside parentheses. At best, result arguments should be prefixed with "out" like in C#. However, here they are not arguments.

Here are alternative messages that could help reduce the friction for new users

* 1️⃣  The method returns 1 value(s) but is assigned to 0 variable(s). All return values must be assigned.
* 2️⃣  Wrong number of assignment variables to store the result of the method (expected 1, got 0)
* 3️⃣  The number of assignment variables is not the same as the number of return values (expected 1, got 0)
* 4️⃣  Mismatch between returned values (1) and assignments (0)

Please add your alternatives and/or vote so that we can make a better error message

**Update**
Based on three votes, we choose 

**The method returns 1 value(s) but is assigned to 0 variable(s). All return values must be assigned.**

MikaelMayer:
This PR fixes #3835

### DIFF
--- a/Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
+++ b/Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
@@ -3096,7 +3096,9 @@ namespace Microsoft.Dafny {
         if (isInitCall) {
           reporter.Error(MessageSource.Resolver, s, "a method called as an initialization method must not have any result arguments");
         } else {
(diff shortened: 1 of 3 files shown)


keyboardDrummer:
Nice! ❤️

PR #4169, opened 2023-06-12
Fix: Making assertion explicit work for nested statements

### LINKED ISSUE #4016: Making assertions explicit fails for nested statements
### Dafny version

4.1.0

```dafny
datatype D = C(value: int) | N

function Test(e: D, inputs: map<int, int>): bool {
  match e
  case N => true
  case C(index) => inputs[index] == index
}
```

```code
Paste this code in VSCode, click on input[index], select "Insert explicit failing assertion", it will modify the code incorrectly.
```

### What happened?

Here is the wrong result. `index#Z` should be replaced by `index`
```
datatype D = C(value: int) | N

function Test(e: D, inputs: map<int, int>): bool {
  match e
  case N => true
  case C(index) => assert index#Z in inputs;
                   inputs[index] == index
}
```

Windows

MikaelMayer:
This PR fixes #4016


alex-chew (approved):
LGTM, just one nitpick

Source/DafnyPipeline.Test/ImplicitAssertionTest.cs
```diff
@@ -0,0 +1,196 @@
...
+
+  [Fact]
+  public void ElementNotInDomain() {
+    ShouldHaveImplicitCode(@"
+method Test(m: map<int, int>, x: int) {
+  var b := m[x + 2]; // Here
+}
+", "x + 2 in m");
+  }
+  private void ShouldHaveImplicitCode(string program, string expected, DafnyOptions options = null) {
```
MikaelMayer, line 141:
```suggestion
  /** Look at the line containing "// Here", and for every assertion there,
  look if they contain an implicit assertion that, if printed,
  would generate the string "expected" */
  private void ShouldHaveImplicitCode(string program, string expected, DafnyOptions options = null) {
```

MikaelMayer:
I had to fix other tests that were failing, because this PR was printing the original range so creating an identifier with an original range and a new name did not work anymore (for printing `b'` for example). So I had to perform a few workarounds.

Source/DafnyCore/Verifier/ProofObligationDescription.cs
```diff
@@ -10,6 +10,14 @@ public abstract class ProofObligationDescription : Boogie.ProofObligationDescrip
   // An expression that, if verified, would trigger a success for this ProofObligationDescription
   // It is only printed for the user, so it does not need to be resolved.
   public abstract Expression GetAssertedExpr(DafnyOptions options);
+  
+  // Substituting replaces the token of a substituting expression by the token of the identifierExpr being susbstituted,
+  // Since the printer requires the token of IdentifierExpr to be Token.NoToken to print the custom name in Dafny mode,
+  // we just wrap the identifierExpr into a ParensExpression, as it is the case for any other expression.
+  protected static Expression ToSubstitutableExpression(BoundVar bvar) {
+    var expression = new IdentifierExpr(bvar.tok, bvar);
+    return new ParensExpression(bvar.tok, expression) { Type = bvar.Type, ResolvedExpression = expression};
```
MikaelMayer, line 19:
```suggestion
    return new ParensExpression(bvar.tok, expression) { Type = bvar.Type, ResolvedExpression = expression };
```

Source/DafnyCore/Verifier/ProofObligationDescription.cs
```diff
@@ -10,6 +10,14 @@ public abstract class ProofObligationDescription : Boogie.ProofObligationDescrip
   // An expression that, if verified, would trigger a success for this ProofObligationDescription
   // It is only printed for the user, so it does not need to be resolved.
   public abstract Expression GetAssertedExpr(DafnyOptions options);
+  
+  // Substituting replaces the token of a substituting expression by the token of the identifierExpr being susbstituted,
+  // Since the printer requires the token of IdentifierExpr to be Token.NoToken to print the custom name in Dafny mode,
+  // we just wrap the identifierExpr into a ParensExpression, as it is the case for any other expression.
+  protected static Expression ToSubstitutableExpression(BoundVar bvar) {
```
MikaelMayer, line 17:
```suggestion

  // Substituting replaces the token of a substituting expression by the token of the identifierExpr being susbstituted,
  // Since the printer requires the token of IdentifierExpr to be Token.NoToken to print the custom name in Dafny mode,
  // we just wrap the identifierExpr into a ParensExpression, as it is the case for any other expression.
  protected static Expression ToSubstitutableExpression(BoundVar bvar) {
```

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -686,7 +698,7 @@ public ThisSurrogate(IToken tok, Type type)
   [DebuggerDisplay("Bound<{name}>")]
   public class BoundVar : NonglobalVariable {
     public override bool IsMutable => false;
-
+
```
MikaelMayer, line 701:
```suggestion

```

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -503,6 +512,9 @@ public abstract class NonglobalVariable : TokenNode, IVariable {
         return name;
       }
     }
+    
+    public string DafnyName => RangeToken == null || tok.line == 0 ? Name : RangeToken.PrintOriginal();
+
```
MikaelMayer, line 517:
```suggestion

```

Source/DafnyCore/AST/DafnyAst.cs
```diff
@@ -503,6 +512,9 @@ public abstract class NonglobalVariable : TokenNode, IVariable {
         return name;
       }
     }
+
```
MikaelMayer, line 515:
```suggestion

```

alex-chew (approved):

PR #4171, opened 2023-06-12
Fix: Compilation of continue labels no longer crashing in Go

### LINKED ISSUE #3978: Go "continue" miscompilation error
### Dafny version

4.1.0

```dafny
method Main() returns ()
{
 for v_int_7 := 3 to 18
 {
  if (false) {
    continue;
  }
  var v_int_93 := 1;
  print v_int_93, "\n";
 }
}
```

```code
$ dafny /compile:2 /compileTarget:go t.dfy

# command-line-arguments
t-go/src/t.go:49:12: goto C0 jumps over declaration of _1_v__int__93 at t-go/src/t.go:51:9
```

### What happened?

Compiling with a continue within a for loop in Go leads to the message shown above. This error does not occur in any other language. I would expect this program to compile correctly as the continue is wrapped in an if (false) statement and hence all code below the if should be able to be compiled as it would be run.

Linux

MikaelMayer:
This PR fixes #3978
Our compiler hit the problem stated there: https://github.com/golang/go/issues/27165
I haven't tested if this problem is fixed in a more recent version of the Go compiler, but a workaround I found is to wrap the labelled code with curly braces, so that the goto escape the scope when jumping to the label, hence Go does not complain anymore about jumping over variable declaration


Test/git-issues/git-issue-3978.dfy
```diff
@@ -0,0 +1,13 @@
+// RUN: %testDafnyForEachCompiler "%s"
```
MikaelMayer, line 1: @robin-aws  your script required me to change my test for the Go compiler to every compiler. Nice job!

Source/DafnyCore/Compilers/GoLang/Compiler-go.cs
```diff
@@ -1835,9 +1835,11 @@ protected class ClassWriter : IClassWriter {
     }
 
     protected override ConcreteSyntaxTree CreateLabeledCode(string label, bool createContinueLabel, ConcreteSyntaxTree wr) {
+      wr.WriteLine("{");
```
RustanLeino, line 1838: Instead of explicitly adding the open and close curly braces, change the `var w = wr.Fork();` line to

    var w = wr.NewBlock(open: BlockStyle.Brace);

Test/git-issues/git-issue-3978.dfy
```diff
@@ -0,0 +1,13 @@
...
+
+method Main() returns ()
+{
+ for v_int_7 := 3 to 18
+ {
+  if (false) {
```
RustanLeino, line 7: Remove parentheses

Test/git-issues/git-issue-3978.dfy
```diff
@@ -0,0 +1,13 @@
+// RUN: %testDafnyForEachCompiler "%s"
+
+method Main() returns ()
+{
+ for v_int_7 := 3 to 18
```
RustanLeino, line 5: I see this is how the bug report showed the code. Still, I would suggest renaming `v_int_7` and `v_int_93` to something like `x` and `y`.

Test/git-issues/git-issue-3978.dfy
```diff
@@ -0,0 +1,13 @@
+// RUN: %testDafnyForEachCompiler "%s"
+
+method Main() returns ()
```
RustanLeino, line 3: Remove `returns ()`

RustanLeino (changes requested):

Source/DafnyCore/Compilers/GoLang/Compiler-go.cs
MikaelMayer: I can't because we need to add a label right after the final closing brace (line 1843), unless you have another suggestion?

Test/git-issues/git-issue-3978.dfy
```diff
@@ -0,0 +1,13 @@
...
+{
+ for v_int_7 := 3 to 18
+ {
+  if (false) {
+    continue;
+  }
+  var v_int_93 := 1;
+  print v_int_93, "\n";
+ }
```
MikaelMayer, line 12:
```suggestion
method Main()
{
 for v := 3 to 18
 {
  if false {
    continue;
  }
  var x := 1;
  print x, "\n";
 }
```

Source/DafnyCore/Compilers/GoLang/Compiler-go.cs
RustanLeino: I don't see that the label _after_ the closing brace is a problem, but I now see that the goto between the fork point and the closing curly is. So, the `Fork` _is_ necessary, after all. My bad.

However, when I now look at that, it seems that both the forked `w` and the `goto ...` are not indented inside the curly braces. The best way to fix that is to use a `NewBlock`.

In summary, would the following body of `CreateLabeledCode` work for you?

``` c#
var wBody = wr.NewBlock(open: BlockStyle.Brace);
var w = wBody.Fork();
var prefix = createContinueLabel ? "C" : "L";
wBody.WriteLine($"goto {prefix}{label};");
wr.Fork(-1).WriteLine($"{prefix}{label}:");
return w;
```

RustanLeino (changes requested):

RustanLeino (approved):

PR #4172, opened 2023-06-12
Chore: Unused identifiers is an advice, not a formal definition

### LINKED ISSUE #3949: Unexpected handling of variables in match cases
### Dafny version

4.1.0

```dafny
method Main() returns () {
  var v: int := 1;

  print v, " \n";
  match 0 {
    case v => {
      print v, " UNEXPECTED\n";
    }
    case _ => {
      print v, " EXPECTED\n";

    }
  }
  print v, " \n";

}
```

```code
$ dafny /compile:4 test.dfy
test.dfy(9,4): Warning: this branch is redundant

Running...

1 
0 UNEXPECTED
1
```

### What happened?

When a case has a variable which has previously been assigned, dafny appears to reassign the variables value for the scope of the case instead of testing whether its value is equal to the one being matched.

According to the [documentation](https://dafny.org/latest/DafnyRef/DafnyRef#sec-match-statement), this behaviour should only occur on "unused" identifiers. 

In addition this leads to the warning being produced, where the intended behaviour is that the default case is required.

Linux

MikaelMayer:
Fixes #3949 


docs/DafnyRef/Statements.md
```diff
@@ -922,7 +922,7 @@ the same as for match expressions and is described in
 
 The selector need not be enclosed in parentheses; the sequence of cases may but need not be enclosed in braces.
 The cases need not be disjoint.
-The cases must be exhaustive, but you can use a wild variable (`_`) or an as yet unused simple identifier to indicate "match anything".
+The cases must be exhaustive, but you can use a wild variable (`_`) or a simple identifier to indicate "match anything". It is possible to shadow existing variable names.
```
alex-chew, line 925: While the proposed change is a little clearer than the original, I think it would be overall better to instead direct the reader to section 9.31.2 (as the previous paragraph does) for details about binding behavior, since the proposed change is still imprecise - for example, does a simple identifier shadow constants or matching parameterless constructors?

Even better, section 9.31.2 should include examples of each of the described patterns.

MikaelMayer: Good idea for the first one. I rephrased to add a link to that section. Adding examples will be a bit more complicated since it involves also describing all the cases of shadowing, etc, so I suggest we hold off for now.

alex-chew: That sounds good to me, thanks.

alex-chew (approved):

PR #4175, opened 2023-06-14
Add basic support for cancellation of resolution

keyboardDrummer:
### Changes
- Add support for cancelling resolution work at module granularity. Before this PR, the language server would always finish it's currently running resolution job before starting resolution on the most recent changes, skipping resolution work of obsolete changes that was not yet started. With this PR, the language server can also stop it's current resolution work at the module boundaries.
- Add a telemetry publication that says how much time was spent on resolution

### Testing

I see improved behavior on large Dafny codebases with this change. The IDE generally becomes more responsive, switching away from 'Resolving...' into other states faster.

Given that from a user perspective the resolution process is black box, there's no contract for when in that process cancellation should be supported. I think the right way to test this is to have a codebase that takes a long time to resolve, and to test that changes will always cancel the current resolution within a time smaller than X. However, making such a test is time consuming and I would like to deliver this fix to users already, so I'm not adding such a test yet.

I also haven't added a test for the telemetry change but that is not a user facing change.


MikaelMayer (approved):
Looks like a small but great improvement that should definitely enhance the responsiveness of the resolution.
I would advise as well to later put these throwIfCancellationRequested also when starting to resolve individual declarations, because there could be a lot in a module.

PR #4185, opened 2023-06-15
Chore: Fix Gradle build flagging Java 8 compatibility as "obsolete"

MikaelMayer:
Without this fix, every other PR cannot be merged.


robin-aws (changes requested):
We need to be careful that we're not breaking compatibility of the runtime with Java 8 with this change, or that it didn't happen already. I think sourceCompatibility is fine to change since it only affects maintainers, but does this mean targetCompatibility was already automatically upgraded?

PR #4187, opened 2023-06-15
Fix: Removed wrong "related position" precision when dealing with regrouped quantifiers

### LINKED ISSUE #2211: Wrong conjunct mentioned in "related location" under a quantifier
Consider this Dafny program.

```dafny
predicate P1(a: int)
predicate P2(a: int)

predicate Q(s: seq<int>) {
  forall i | 0 <= i < |s| ::
    (P1(s[i]) || false) && (P2(s[i]) || false)  // Line 6
}

method M(s: seq<int>)
  requires forall i | 0 <= i < |s| :: P1(s[i])
  // requires forall i | 0 <= i < |s| :: P2(s[i])
  ensures Q(s)
{}
```

The method `M` has an error because it is missing the commented-out precondition about `P2`. But Dafny reports the following error misleading messages:

```
issue2211.dfy(13,0): Error: A postcondition might not hold on this return path.
issue2211.dfy(12,10): Related location: This is the postcondition that might not hold.
issue2211.dfy(5,2): Related location
issue2211.dfy(6,14): Related location

Dafny program verifier finished with 2 verified, 1 error
```

The last "Related location" points to the *first* conjunct (the 14th character on the 6th line), but in fact the first conjunct is true and the problem is with the second conjunct.

On the other hand, Dafny behaves correctly when not under a quantifier, as in this example:

```dafny
predicate P1(a: int)
predicate P2(a: int)

predicate Q(s: seq<int>) {
  |s| > 0 && P1(s[0]) && P2(s[0])
}

method M(s: seq<int>)
  requires |s| > 0
  requires P1(s[0])
  ensures Q(s)
{}
```

which generates

```
issue2211.dfy(12,0): Error: A postcondition might not hold on this return path.
issue2211.dfy(11,10): Related location: This is the postcondition that might not hold.
issue2211.dfy(5,25): Related location

Dafny program verifier finished with 2 verified, 1 error
```

The 25th character of the 5th line correctly points to the third conjunct, because the first two are true.

This is a relatively serious usability issue, and I feel like it is a regression that used to work?

I reproduced this error on the main branch.

MikaelMayer:
This PR fixes #2211
After investigation it looks like the following expression
```dafny
forall i | 0 <= i < |s| :: // Normal token
    && P1(s[i])
    && P2(s[i])
```
was split in two while searching for triggers 
```dafny
forall i | 0 <= i < |s| :: // Nested token, outer pointing to "forall", inner pointing to "P1"
    P1(s[i])
forall i | 0 <= i < |s| :: // Nested token, outer pointing to "forall", inner pointing to "P2"
    P2(s[i])
```
But another pass for searching for triggers regroups quantifiers which have the same bound variables and same triggers. Since the two expressions have the same trigger s[i], they are regroupped with an AND back to:
```dafny
forall i | 0 <= i < |s| :: // Nested token, outer pointing to "forall", inner pointing to "P1"
    P1(s[i]) && P2(s[i])
```
As you can see, the problem of the regrouping is that it takes the token of the first element of the group, which had an indication that it pointed to P1.
The last related location of #2211 thus was an artefact of the token being copied without paying attention to that detail. Since we regroup the tokens, there is no nested token there to have so this PR removes nested tokens when regrouping comprehensions to avoid giving false information.


RustanLeino (approved):
Thanks for the good description. From that description, this fix looks good.

Can you add another test that will exercise the `is ExistsExpr` branch of the code as well?

MikaelMayer:
> Thanks for the good description. From that description, this fix looks good.
> 
> Can you add another test that will exercise the `is ExistsExpr` branch of the code as well?

I added a test and verified that the output was different before this PR, it included the unwanted error location.

RustanLeino (approved):

RustanLeino (approved):

PR #4190, opened 2023-06-16
fix: Generate dummy constraint for `newtype`s to enable axiom generation

### LINKED ISSUE #4188: Dafny unable to produce arbitrary witness element from set of nats in custom datatype
### Dafny version

4.1.0.0

```dafny
newtype Int = nat
datatype IntWrapper = Wrap(s: set<Int>)

method m(x: IntWrapper) {
  var iter := x.s;
  while iter != {}
    decreases |iter|
  {
    var i: Int :| i in iter;
    iter := iter - {i};
  }
}
```

```code
VSCode verifier shows the following error message: 
cannot establish the existence of LHS values that satisfy the such-that predicate
```

### What happened?

The current code fails at the line `var i: Int :| i in iter;` -- Dafny cannot find an arbitrary witness from the set `iter` of type `set<nat>`. However, if I change the newtype definition from `newtype Int = nat` to `newtype Int = int`, verification succeeds. 

Mac

fabiomadge:
Fixes #4188.


fabiomadge:
Before merging we need to figure out how to pass the wellformedness check, now that we don't conveniently skip it anymore.

Test/git-issues/git-issue-4188.dfy
```diff
@@ -0,0 +1,14 @@
+// RUN: %baredafny verify %args "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+newtype Int = nat
```
RustanLeino, line 4: I suggest making two more variations of this example, namely the ones mentioned in the Issue discussion (`newtype Int = int` and `newtype Int = n: nat | true`, perhaps also with `newtype Int = x: int | 0 <= x`).

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -1274,14 +1274,15 @@ internal class PredefinedDecls {
...
-      }
+      dd.Var ??= new BoundVar(dd.tok, "x", dd.BaseType) { IsTypeExplicit = true };
+      dd.Constraint ??= Expression.CreateBoolLiteral(dd.tok, true);
+
+      AddWellformednessCheck(dd);
+      currentModule = dd.EnclosingModuleDefinition;
+      // Add $Is and $IsAlloc axioms for the newtype
+      AddRedirectingTypeDeclAxioms(false, dd, dd.FullName);
+      AddRedirectingTypeDeclAxioms(true, dd, dd.FullName);
+      currentModule = null;
```
RustanLeino, line 1285: Well identified! These type axioms are what's missing. However, it's not a good idea to change the AST at this time. Also, I think it's fine to continue to omit a well-formedness check for these var-less newtypes (hence, addressing your question in the PR comment).

So, I suggest making the `if (dd.Var != null)` control only whether or not `AddWellformednessCheck` is called. Then, do the other lines in either case (as you have in your new L1281-L1285).

Then, adjust `AddRedirectingTypeDeclAxioms` so that it's able to handle the `dd.Var == null` case. Instead of `dd.Var.Type` (which occurs many times in that method), use `baseType` defined as

``` c#
var baseType = dd.Var != null ? dd.Var.Type : ((NewtypeDecl)(object)dd).BaseType;
```

One of the method's preconditions thus needs to change to

``` c#
Contract.Requires((dd.Var != null && dd.Constraint != null) || dd is NewtypeDecl);
```

There are two other occurrences of `dd.Var` that you'll need to handle in some way.

FInally, if `dd.Constraint` is `null`, use `Bpl.Expr.True` instead of `etran.TrExpr(dd.Constraint)`.

RustanLeino (approved):

PR #4195, opened 2023-06-19
Add the axiom stating that `Seq#Take(s, Seq#Length(s)) == s)` for all…

### LINKED ISSUE #4184: `assert(multiset(arr[..]) == multiset(arr[..arr.Length]))` fails!
### Dafny version

4.1.0

```dafny
method main() {
  var arr: array<int>;
  assert(multiset(arr[..]) == multiset(arr[..arr.Length]));
}
```

_No response_

### What happened?

The assertion cannot be proven correct which make it more difficult to specify and verify certain type of programs (e.g. array merge).

Mac

MikaelMayer:
… sequence `s` (#4194)

A trivial axiom stating `Seq#Take(s, Seq#Length(s)) == s)` for all sequence `s`. This will help deducing automatically and without using any workaround that "multiset(arr[..]) == multiset(arr[..arr.Length])" for any array `arr`.

Fixes #4184

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT
license](https://github.com/dafny-lang/dafny/blob/master/LICENSE.txt).</small>

Fixes #

### DIFF
--- a/Test/git-issues/git-issue-4184.dfy
+++ b/Test/git-issues/git-issue-4184.dfy
@@ -1,0 +1,7 @@
+// RUN: %baredafny verify %args "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method main() {
+  var arr: array<int>;
+  assert(multiset(arr[..]) == multiset(arr[..arr.Length]));
+}
--- a/Source/DafnyCore/DafnyPrelude.bpl
+++ b/Source/DafnyCore/DafnyPrelude.bpl
@@ -1106,6 +1106,10 @@ axiom (forall<T> s, t: Seq T, n: int ::
   Seq#Take(Seq#Append(s, t), n) == s &&
   Seq#Drop(Seq#Append(s, t), n) == t);
 
+axiom (forall<T> s : Seq T ::
+  { Seq#Take(s, Seq#Length(s)) }
+  Seq#Take(s, Seq#Length(s)) == s);
+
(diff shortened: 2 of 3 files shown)


Test/git-issues/git-issue-4184.dfy
```diff
@@ -0,0 +1,7 @@
+// RUN: %baredafny verify %args "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method main() {
+  var arr: array<int>;
```
MikaelMayer, line 5:
```suggestion
method test(arr: array<int>) {
```

btwael:
It looks that the [some-proofs-only-work-without-autoTriggers.dfy](https://github.com/dafny-lang/dafny/blob/5679b3dbbba3f9f4ed3acc139e659d26a9c19aff/Test/triggers/some-proofs-only-work-without-autoTriggers.dfy] is what cause the failure of test, adding the axiom in line 23 to the prelude make the whole test file verified (with the added axiom in this PR):
```
axiom (forall<T> s: Seq T, i: int ::
  { Seq#Index(s, i) }
  0 <= i && i < Seq#Length(s) ==>
  Seq#Contains(s, Seq#Index(s, i)));
```
If we add this axiom, maybe we should just modify "some-proofs-only-work-without-autoTriggers.dfy.expect"!?
I wonder why those axioms are not already part of the default prelude file?

PR #4199, opened 2023-06-20
Do not run translation if there are resolution errors

keyboardDrummer:
### Changes
In the language server, do not run translation if there are resolution errors

### Testing
Add XUnit test


MikaelMayer (approved):
I approve it because it seems obvious but... What was the previous behavior? Did your test identify a crash?

keyboardDrummer:
> I approve it because it seems obvious but... What was the previous behavior? Did your test identify a crash?

The test crashes without the change yes.

PR #4201, opened 2023-06-20
Virtual filesystem

### LINKED ISSUE #4111: Let unsaved changes from included files be taken into account by the language server
Currently all Dafny projects read all files except the root from disk, so unsaved changed from files are only visible to the project that has this file as its root.

Instead, Dafny should access files through a filesystem interface, for which the language server has an implementation that is shared between projects, so that unsaved file changes are visible to all the projects.

### Implements hints

- See where the constructor of `DafnyFile` is called, and update those locations so any calls to `new StreamReader(filePath)` use an interface method `IFileSystem.OpenFile` instead.
- Use an instance `RegularFileSystem : IFileSystem` in `DafnyDriver` that does nothing but access the existing filesystem.
- Update `DocumentManagerDatabase` so it holds an instance of `InMemoryFileSystemOverlay : IFileSystem`, which it updates when its `OpenDocument`, `UpdateDocument` and `CloseDocument` methods are called. `InMemoryFileSystemOverlay` can use the `TextBuffer` class to store text for individual documents.
- Use the language server's dependency injection to propagate an instance of `DocumentManagerDatabase` to where it's needed for the language server, which probably only in `DafnyLangParser`

keyboardDrummer:
Fixes #4111

### Changes
- Let all unsaved file changes be visible to all `DocumentManager`s. Previously only the unsaved changes of the document being viewed were visible.
- None of the language server code depends on accessing the contents of any files directly any more. All textual access is done through the tokens resulting from the parser.
- Simplify `SymbolGuesser` so it no longer does some form of partial parsing but instead uses the AST.
- Let `ExprDotName.Children` work well when it is not yet resolved.
- Let parser return a more complete AST when statements are missing a `;` at the end, which is useful for code completion.
- Code for related location messages is more robust since it no longer relies on IO
- Remove `TextChangeProcessor` since it did so little that it wasn't worth its weight.
- DafnyHoverHandler consistently uses no group separator every 3 digits
- No longer let tests access internal text buffers

### Testing
- Add XUnit test that opens, updates and closes an included file, and verifies that the including file sees the changes.


Source/DafnyLanguageServer.Test/Synchronization/TextInsertionTest.cs
```diff
@@ -35,11 +26,11 @@ public class TextInsertionTest : SynchronizationTestBase {
 function GetConstant(): int {
   1
 }".TrimStart();
-      Assert.Equal(expected, document.TextDocumentItem.Text);
+      AssertCorrectBufferUpdate(source, range, change, expected);
```
MikaelMayer, line 29: Nice refactoring!

Source/DafnyLanguageServer.Test/Synchronization/TextReplacementTest.cs
```diff
@@ -1,110 +1,90 @@
...
+using Microsoft.Dafny.LanguageServer.Workspace;
 using Xunit;
-using Xunit.Abstractions;
 
 namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Synchronization {
-  public class TextReplacementTest : SynchronizationTestBase {
+  public class TextReplacementTest {
     [Fact]
-    public async Task ReplaceSingleLineTextAtStart() {
+    public void ReplaceSingleLineTextAtStart() {
```
MikaelMayer, line 9: If I understand correctly ,you completely remove the dependency of this test to the language server. I feel very suspicious about that as these are the language server tests, the transmission protocol, not tests about Omnisharp. Please justify.

Source/DafnyLanguageServer/Language/DiagnosticErrorReporter.cs
```diff
@@ -89,29 +87,11 @@ public class DiagnosticErrorReporter : ErrorReporter {
...
-            } catch (IOException) {
-              message = message + "(could not open file " + fileName + ")";
-            } catch (ArgumentOutOfRangeException) {
-              message = "Related location (could not read position in file " + fileName + ")";
-            }
-            if (message == "") {
-              message = "Related location (could not read file " + fileName + ")";
-            }
-          }
+          message = FormatRelated(range.PrintOriginal());
```
MikaelMayer, line 94: Indeed, no need to read the file system anymore.

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
```diff
@@ -75,26 +73,15 @@ public class DocumentManager {
   private const int MaxRememberedChanges = 100;
   private const int MaxRememberedChangedVerifiables = 5;
   public void UpdateDocument(DidChangeTextDocumentParams documentChange) {
-    // According to the LSP specification, document versions should increase monotonically but may be non-consecutive.
-    // See: https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md?plain=1#L1195
-    var oldVer = Compilation.TextBuffer.Version;
-    var newVer = documentChange.TextDocument.Version;
-    var documentUri = documentChange.TextDocument.Uri;
-    if (oldVer >= newVer) {
```
MikaelMayer, line 83: Why removing this version check? Seemed useful.

Source/DafnyLanguageServer/Workspace/TextBuffer.cs
```diff
@@ -99,7 +99,7 @@ public class TextBuffer {
     if (end < start) {
       throw new ArgumentException();
     }
-    if (start < 0 || end >= Text.Length) {
+    if (start < 0 || end > Text.Length) {
```
MikaelMayer, line 102: Good catch !

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -41,13 +41,11 @@ ILogger<SignatureAndCompletionTable> loggerSymbolTable
     }
 
     public VerificationTree RelocateVerificationTree(VerificationTree originalVerificationTree,
-      int lines,
```
MikaelMayer, line 44: I'm surprised you got away by removing line numbers here because it was needed, but since the file system is virtual, I understand you can afford to recompute this number every time.

MikaelMayer (commented):
Looks like a great improvement.
 We might have complains of people saying "my code verifies on VSCode but not on the CLI" because they forgot to save a file though. Could we perhaps autosave the files alternatively?

Source/DafnyCore/Dafny.atg
```diff
@@ -3006,6 +3005,12 @@ UpdateStmt<out Statement/*!*/ s>
     )
     ";"                            (. endTok = t; .)
   | ":"                            (. SemErr(ErrorId.p_invalid_colon, new RangeToken(startToken, t), "invalid statement beginning here (is a 'label' keyword missing? or a 'const' or 'var' keyword?)"); .)
+  | { Attribute<ref attrs> }       (. endToken = t; .)
+    ";"                            (. endTok = t; rhss.Add(new ExprRhs(e, attrs) { RangeToken = new RangeToken(e.StartToken, endToken) }); .)   
+  | { Attribute<ref attrs> }       (. endToken = t; .)
+                                   (. endTok = t; rhss.Add(new ExprRhs(e, attrs) { RangeToken = new RangeToken(e.StartToken, endToken) });
+                                      SemErr("statementMissingSemicolon", new RangeToken(startToken, t), "missing semicolon at end of statement");
```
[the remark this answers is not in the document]
keyboardDrummer: What's the benefit of using an enum? Seems like it centralized unrelated code to me, without a benefit.

Source/DafnyLanguageServer/Workspace/DocumentManager.cs
keyboardDrummer: It's moved to LanguageServerFileSystem

Source/DafnyCore/Generic/NodeExtensions.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Linq;
+
+namespace Microsoft.Dafny;
+
+public static class NodeExtensions {
+
+  public static INode FindNode(this INode node, DafnyPosition position) {
```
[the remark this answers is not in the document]
keyboardDrummer: I feel the deduplication opportunity is already in the implementation of `Children`. Having a foreach loop over `Children` both here and in `Visit` seems like a negligible amount of duplication to me, and removing that doesn't seem worth the benefit that we get from having this method be more specialized than `Visit`.

> rather have one common method to call

I can make one of them private, might be less confusing.

Source/DafnyCore/Coco/Parser.frame
```diff
@@ -167,6 +173,13 @@ public class Errors {
     return s;
   }
 
+  public void SemErr(string errorId, IToken tok, string msg) {  // semantic errors
```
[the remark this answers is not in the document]
keyboardDrummer: Doesn't this also mean you always have errorIds ? The string parameter is not optional

keyboardDrummer:
> We might have complains of people saying "my code verifies on VSCode but not on the CLI" because they forgot to save a file though. Could we perhaps autosave the files alternatively?

Maybe let's wait and see if that's an issue?

Source/DafnyLanguageServer.Test/Synchronization/TextReplacementTest.cs
keyboardDrummer: The previous test was pretending to be a high level using LSP APIs, but was actually mixing them with C# APIs to access internal state. I don't think it made sense. Anyways, the internal APIs it was using don't exist anymore, and now it's using more stable TextBuffer C# APIs.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
keyboardDrummer: I'm surprised too, so much so that I took a look at the code, found issues, added a new test and spend several hours fixing the problem.

Source/DafnyLanguageServer.Test/Synchronization/TextReplacementTest.cs
MikaelMayer: Ok, so actually these tests are not testing anything of our own code we wrote, so if they are not useful anymore, what about we remove them? It looks like we are testing an existing library. Unless I miss something.

Source/DafnyCore/Dafny.atg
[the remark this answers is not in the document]
MikaelMayer: The benefit of using an enum is that you can then relate it to code actions which help as fixing the code semi-automatically.

Source/DafnyCore/Coco/Parser.frame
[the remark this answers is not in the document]
MikaelMayer: Plugins will use `Reporting.Error` directly, they won't use the internal method `SemErr`. For our own code, every method should get an error id so that it can be linked to documentation, and also possibly used by code actions.

Source/DafnyCore/Dafny.atg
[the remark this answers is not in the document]
keyboardDrummer: You can use a string constant for the same purpose

Source/DafnyCore/Coco/Parser.frame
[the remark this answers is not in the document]
keyboardDrummer: Even though they can't now, plugins should be able to add syntax as well. I don't think we should have two almost the same but slightly different variations of an API, one for plugins and one for code from this repo.

Source/DafnyLanguageServer.Test/Synchronization/TextReplacementTest.cs
keyboardDrummer: TextBuffer is our own code.

MikaelMayer (approved):
Perfect, looks great, thanks!

stefan-aws (approved):

PR #4212, opened 2023-06-22
Fix: Modules verified in the correct order to prevent Boogie Crash

### LINKED ISSUE #4139: Regression: dafny run not working with include
### Dafny version

4.1.0

```dafny
test.dfy:
include "main.dfy"

method Main() {
  var undertest := new TestLogic''TopLevel();
  print "COMPLETED OK\n";
}

main.dfy:
class TestLogic''TopLevel {
  constructor () 
  {
  }
}
```

```code
dafny run --no-verify test.dfy

test.dfy(4,19): Error: call to undeclared procedure: Call$$_module.TestLogic_k_kTopLevel.__ctor
1 name resolution errors detected in /var/folders/89/bc5ryf6x55l_pm4w3347kd200000gn/T/test__module.bpl

/var/folders/89/bc5ryf6x55l_pm4w3347kd200000gn/T/test__module.bpl(2882,4): Error: call to undeclared procedure: Call$$_module.TestLogic_k_kTopLevel.__ctor
1 name resolution errors detected in /var/folders/89/bc5ryf6x55l_pm4w3347kd200000gn/
```

### What happened?

1) It does not work as it should
2) Why is there any reference to bpl files at all? No verification is called for.
3) It works fine if the files are combined into one file, without using include.

4!!!) If one includes the option --verify-included-files, enven though one also has --no-verify, then it works.

Mac

MikaelMayer:
This PR fixes #4139
I'm not sure of the fix right now, need to see if it works on other tests. I suspect it will fail and we might need another way to separate the TopLevelDecls order for verification as well. But let's see.


Source/DafnyCore/AST/Modules/ModuleDefinition.cs
```diff
@@ -58,12 +57,8 @@ public class ModuleDefinition : RangeNode, IDeclarationOrUsage, IAttributeBearin
   public readonly List<TopLevelDecl> ResolvedPrefixNamedModules = new();
   [FilledInDuringResolution]
   public readonly List<PrefixNameModule> PrefixNamedModules = new();  // filled in by the parser; emptied by the resolver
-  public virtual IEnumerable<TopLevelDecl> TopLevelDecls =>
-    defaultClassFirst ? DefaultClasses.
+  public virtual IEnumerable<TopLevelDecl> TopLevelDecls => DefaultClasses.
```
keyboardDrummer, line 60: I like this this PR simplifies the code, but Translator shouldn't depend on the order of TopLevelDecls, should it?

MikaelMayer: The order does not matter, except for one thing: The default declaration should be first, according to the current optimization of Boogie pruning.
Before the pruning of unused declarations in boogie, the order surely did not matter.

keyboardDrummer: How is the pruning affected by order? I don't think it should. Is that a bug?

MikaelMayer: The pruning is affected by order because of this condition here:

https://github.com/dafny-lang/dafny/blob/2b26b9d1cbb228a5bc47df5b1e0acb64287ec8b3/Source/DafnyCore/Verifier/Translator.ClassMembers.cs#L165-L169

When the default class is first, it will populate the `referenceMembers` for every definition that is referenced, so even if the method is in another file (!InVerificationScope), that method will still be added.
When the default class is not first, the `referenceMembers` will be empty when dealing with the dependencies first, so the method is ignored.

I don't have any more bandwidth in the next weeks to take care of this issue. Would you know of a quicker fix?

keyboardDrummer (approved):

PR #4220, opened 2023-06-27
Issue 4035 revisited

### LINKED ISSUE #4035: Missing antecedent in type axioms
### Dafny version

4.0.0

```dafny
type ReallyEmpty = x: int | false witness *

trait UnimplementableTrait {
  const x: ReallyEmpty
}

method M(o: object)
  ensures false
{
  if o is UnimplementableTrait {
    var u := o as UnimplementableTrait;
    var r := u.x as ReallyEmpty;
  }
}

method Main() {
  var obj := new object;
  M(obj);
  print 2 / 0, "\n"; // ouch!
}
```

```code
% dafny run test.dfy

Error: Execution resulted in exception: Exception has been thrown by the target of an invocation.
System.DivideByZeroException: Attempted to divide by zero.
   at System.Numerics.BigInteger.op_Division(BigInteger dividend, BigInteger divisor)
   at System.Numerics.BigInteger.Divide(BigInteger dividend, BigInteger divisor)
   at Dafny.Helpers.EuclideanDivision(BigInteger a, BigInteger b)
   at _module.__default._Main(ISequence`1 __noArgsParameter)
   at __CallToMain.<>c__DisplayClass0_0.<Main>b__0()
   at Dafny.Helpers.WithHaltHandling(Action action)
   at __CallToMain.Main(String[] args)
```

### What happened?

It's okay to have an unimplementable trait. Here, the trait is unimplementable because therre's no way an implementing type can furnish a value for `x` (of type `ReallyEmpty`). However, the presence of an unimplementable trait should not make verification unsound.

Alas, the type/allocation axioms generated for types in fields are missing an antecedent:

``` boogie
// UnimplementableTrait.x: Type axiom
axiom 1 < $FunctionContextHeight
   ==> (forall $o: ref :: 
    { _module.UnimplementableTrait.x($o) } 
    $Is(_module.UnimplementableTrait.x($o), Tclass._module.ReallyEmpty()));

// UnimplementableTrait.x: Allocation axiom
axiom 1 < $FunctionContextHeight
   ==> (forall $h: Heap, $o: ref :: 
    { _module.UnimplementableTrait.x($o), read($h, $o, alloc) } 
    $IsGoodHeap($h) && read($h, $o, alloc)
       ==> $IsAlloc(_module.UnimplementableTrait.x($o), Tclass._module.ReallyEmpty(), $h));
```

The first of these axioms needs an antecedent

``` boogie
$Is($o, Tclass._module.UnimplementableTrait()) ==>
```

That omission is what makes the bug exploitable. However, the second of the axioms should probably change `read($h, $o, alloc)` to

``` boogie
$IsAlloc($o, Tclass._module.UnimplementableTrait(), $h)
```

Mac

### LINKED ISSUE #4217: Issue raised in
### Dafny version

b2bb23417e49552349170b4bea2e9114cc8276e6

```dafny
function SpecConverter(input: string): int {
  1
}

type StringWrapper {
  ghost const base: string
}

method Converter(value: StringWrapper) returns (res: int)
  ensures res == SpecConverter(value.base)
{
  res := 1;
  //assert res == SpecConverter(value.base); // Uncomment this to verify
}
```

```code
dafny verify or vscode
```

### What happened?

The soundness issue fix in #4036 makes the code above to not verify (but uncommenting the assertion makes it verify). For now, the mitigation was to revert it #4192, and this issue is ONLY about a problem #4036 introduced and should be verified if #4036 get reinserted again.

Based on our observations @RustanLeino  and I, there are two things that can fix it and both should be fixed
- `free ensures` are currently ignored, which prevents a #canCall to be visible and prove the postcondition, whereas the assumption after the assertion adds the #canCall
- There is a type mismatch when the new boogie axiom wants `Is($value, $StringWrapper)` when it's assumed that `IsBox($value, $StringWrapper)`, and both are not interchangeable.

Other

### LINKED ISSUE #4224: Provides-exported type is encoded as injective in its type arguments
### Dafny version

4.0.0

```dafny
module Library {
  export OnlySome
    provides NeedlesslyParameterizedType
  export Everything
    reveals NeedlesslyParameterizedType

  type NeedlesslyParameterizedType<X> = bool
}

module Client {
  import Library`OnlySome

  class Cell<G> {
    constructor () { }
  }

  lemma Different(xx: Cell<Library.NeedlesslyParameterizedType<int>>, yy: Cell<Library.NeedlesslyParameterizedType<real>>)
    ensures xx != yy as object
  {
    // If Library.NeedlesslyParameterizedType is an injective type constructor, then
    //     Library.NeedlesslyParameterizedType<int>
    // and
    //     Library.NeedlesslyParameterizedType<real>
    // are different types. From this, it follows that xx and yy are are also of different types, and thus
    // xx and yy must be different objects.
  }
}

module Contradiction {
  import Library`Everything
  import Client

  method Main() {
    var c := new Client.Cell<bool>();
    Client.Different(c, c);
    assert false;
    print 1/0, "\n"; // ouch!
  }
}
```

```code
% dafny run test.dfy

Error: Execution resulted in exception: Exception has been thrown by the target of an invocation.
System.DivideByZeroException: Attempted to divide by zero.
   at System.Numerics.BigInteger.op_Division(BigInteger dividend, BigInteger divisor)
   at System.Numerics.BigInteger.Divide(BigInteger dividend, BigInteger divisor)
   at Dafny.Helpers.EuclideanDivision(BigInteger a, BigInteger b)
   at Contradiction_Compile.__default._Main(ISequence`1 __noArgsParameter)
   at __CallToMain.<>c__DisplayClass0_0.<Main>b__0()
   at Dafny.Helpers.WithHaltHandling(Action action)
   at __CallToMain.Main(String[] args)
```

### What happened?

For importers of the ``Library`OnlySome`` export set, type constructor `NeedlesslyParameterizedType` is encoded as being injective in its type argument. But the type constructor is _not_ injective, since it does not use its type argument. Taking advantage of this encoding, one can prove `false`.

The type-constructor injectivity axioms should not be generated for abstract types or internal type synonyms. Note, the `AddTypeDecl_Aux` method in `Translator.cs` contains comments that talk about this issue (see "`not unique`" and "... `NOT necessarily injective`"); yet, the injectivity axioms are nonetheless emitted from somewhere.

Mac

RustanLeino:
Fixes #4035 
Fixes #4217
Fixes #4224

There were two problems with the previous fix of Issue #4035 (PR #4036, which was reverted in PR #4192), both related to abstract types. One problem was that the previous fix used `$Is`, not `$IsBox`, for abstract types. The other problem was that Dafny has (for a long time, it seems) generated two Boogie type terms for an abstract type, one of the form `#$MyAbstractType` and the other of the form `TMyAbstractType()`. This PR eliminates the former form. As part of this change, the `#$InternalTypeSynonym` type terms have now also changed to `TInternalTypeSynonym()`.

In addition, this PR fixes a soundness problem (#4224), where injectivity axioms had been generated for export-provided types (which shouldn't happen).


MikaelMayer (approved):
Thanks for fixing the soundness issue!

PR #4225, opened 2023-06-28
Add minimal Rust backend that supports basic Dafny programs

shadaj:
Extends #3907


shadaj:
Alright, CI is failing because of the bootstrapping needed to run Dafny now :(

I might attempt to hack something together to fix this, but it will be good to have a deeper discussion on how we want to structure the build.

MikaelMayer:
> Alright, CI is failing because of the bootstrapping needed to run Dafny now :(
> 
> I might attempt to hack something together to fix this, but it will be good to have a deeper discussion on how we want to structure the build.

Wait, something is not accurate. We did not "boostrap" Dafny in the usual sense e.g. depending on an earlier version of Dafny.
Currently, there are only 2 uses of Dafny inside Dafny itself, and they don't require dependencies
- DafnyRunTime is compiled using DafnyCore/DafnyDriver
- GeneratedFromDafny.cs is a file that is checked out with the source code, and just verified to be equal to all internal Dafny files compiled with the current version of Dafny.

shadaj:
> > Alright, CI is failing because of the bootstrapping needed to run Dafny now :(
> > I might attempt to hack something together to fix this, but it will be good to have a deeper discussion on how we want to structure the build.
> 
> Wait, something is not accurate. We did not "boostrap" Dafny in the usual sense e.g. depending on an earlier version of Dafny. Currently, there are only 2 uses of Dafny inside Dafny itself, and they don't require dependencies
> 
> * DafnyRunTime is compiled using DafnyCore/DafnyDriver
> * GeneratedFromDafny.cs is a file that is checked out with the source code, and just verified to be equal to all internal Dafny files compiled with the current version of Dafny.

Yeah, ended up using the same approach as `GeneratedFromDafny.cs` (originally, I was trying to integrate the compilation of the Dafny sources into the C# build itself). So CI should work now (minus flakiness due to gradle).

Source/DafnyCore/Compilers/InternalCompilersPluginConfiguration.cs
```diff
@@ -14,7 +14,7 @@ internal class InternalCompilersPluginConfiguration : Plugins.PluginConfiguratio
       new PythonBackend(options),
       new CppCompilerBackend(options),
       new LibraryBackend(options),
-      new DafnyBackend(options),
+      new RustBackend(options),
```
MikaelMayer, line 17: Why removing the Dafny Backend?

shadaj: The "Dafny backend" no longer spits out a text representation of the program, but rather a desugared AST that can be consumed by other backends implemented in Dafny (i.e. Rust). So it's only meant for use internally by `DafnyExecutableBackend`.

shadaj:
Alright, CI is green and I think this PR is in a decent state to be reviewed @MikaelMayer! What works so far: simple printing, variables, binary arithmetic, creation of very simple datatypes, and method calls without args or returns.

I'd like to get this merged in sooner than later, because minor refactorings of `SinglePassCompiler` have been needed to properly extract the desugared AST and I'd like to avoid messy merges later on.

jtristan (approved):

PR #4227, opened 2023-06-29
Turn off snippets for lit testing

RustanLeino:
A recent change (#4087) explicitly turned on snippets in `lit`-based testing. The change causes many tests to fail, when run using `lit`. This PR turns them back off.


robin-aws (approved):
Either this was a typo, or it was an attempt to ensure more coverage for `--show-snippets`, which had no affect on CI anyway. :)

PR #4229, opened 2023-06-30
Update Counterexample Parsing to Support Different Polymorphism Encodings in Boogie (anticipating #4210)

Dargones:
This PR updates counterexample parsing so that it works with different polymorphism encodings, anticipating #4210. 

Here is a list of changes that were necessary to support the new type encoding:
1) Create a wrapper around Boogie's `Model.Func` that ignores type arguments, making counterexample extraction independent from the type encoding.
2) For the same purpose, remove all references to `type` function, which Boogie only emits when using the `Predicates` type encoding.
3) Update all counterexample and test generation tests to run twice, once under each encoding.
4) Remove several bugs and sources of nondeterminism in tests that surface after these changes (including having to do with the choice between reporting a nullable or a non-nullable type, a base type vs subset type, sorting counterexamples by position in the program, etc.). Add corresponding tests and fix existing ones. 
5) Remove DatatypeType class that was used by counterexample code to distinguish between class types and datatype types but which is no longer necessary given how test generation works. Propagate this change to test generation.


Dargones:
The [xunit-tests / osx](https://github.com/dafny-lang/dafny/actions/runs/5424651294/jobs/9864318869?pr=4229#logs) took 44 minutes to complete, not sure if this is a fluke or a potential problem. I don't think running counterexample tests twice should have such an effect (and the same tests run much faster on Ubuntu and Windows). An alternative would be to only run them for whatever encoding is the default and switch the setting once https://github.com/dafny-lang/dafny/pull/4210 is merged.

Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
```diff
@@ -58,14 +78,17 @@ method Abs(x: int) returns (y: int)
       await SetUp(o => o.Set(CommonOptionBag.RelaxDefiniteAssignment, true));
       var documentItem = CreateTestDocument(source);
       await client.OpenDocumentAndWaitAsync(documentItem, CancellationToken);
-      var counterExamples = (await RequestCounterExamples(documentItem.Uri)).ToArray();
+      var counterExamples = (await RequestCounterExamples(documentItem.Uri)).
+        OrderBy(counterexample => counterexample.Position).ToArray();
```
atomb, line 82: Wonderful. This has been a consistent source of issues.

Source/DafnyTestGeneration.Test/Various.cs
```diff
@@ -277,65 +287,68 @@ method eightPaths (i:int)
...
-      Assert.True(methods.Count >= 2);
+      Assert.True(methods.Count >= 5);
       Assert.True(methods.All(m =>
-        m.MethodName == "Objects.List.IsACircleOfLessThanThree"));
+        m.MethodName == "Objects.List.IsACircleOfTwoOrLessNodes"));
       Assert.True(methods.All(m =>
-        m.DafnyInfo.IsStatic("Objects.List.IsACircleOfLessThanThree")));
+        m.DafnyInfo.IsStatic("Objects.List.IsACircleOfTwoOrLessNodes")));
       Assert.True(methods.All(m => m.ArgValues.Count == 1));
-      // This test is too specific. A test input may be valid and still not satisfy it.
```
atomb, line 316: Thanks for fixing this! I didn't like just commenting that part out...

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -52,47 +47,45 @@ public class DafnyModel {
...
-      fChar = model.MkFunc("charType", 0);
-      fReal = model.MkFunc("realType", 0);
-      fU2Real = model.MkFunc("U_2_real", 1);
-      fBool = model.MkFunc("boolType", 0);
-      fU2Bool = model.MkFunc("U_2_bool", 1);
-      fInt = model.MkFunc("intType", 0);
-      fU2Int = model.MkFunc("U_2_int", 1);
-      fTag = model.MkFunc("Tag", 1);
-      fBv = model.MkFunc("TBitvector", 1);
+      var tyArgs = options.TypeEncodingMethod switch {
```
atomb, line 53:
```suggestion
      var tyArgMultiplier = options.TypeEncodingMethod switch {
```
I came up with the original name for this, but I think I didn't choose a good one. Maybe this is better?

atomb (approved):
This is very nice! It looks like it was a lot of work. Thanks for spending the time on it!

I made one suggestion to rename a variable that I originally named badly, but feel free to apply or skip that, as you like.

Source/DafnyLanguageServer.Test/Various/CounterExampleTest.cs
Dargones: I didn't realize that was the issue until one of the tests was failing with the new encoding - it is a good variability stress test.

Dargones:
Thank you, @atomb, I have made the change. On a rerun, the duration of OSX tests got down to 30 minutes, so the 44 minutes before must have been a fluke.

atomb (approved):
Thanks for this!

PR #4230, opened 2023-06-30
Replace default class with implicit classes

### LINKED ISSUE #4139: Regression: dafny run not working with include

keyboardDrummer:
Fixes #4139 because it causes declarations in the default module from different included files, to be sorted in topological order of the include graph.

Alternative to: https://github.com/dafny-lang/dafny/issues/4139

### Changes

- Instead of each module having a single default class, have each class member in a module be wrapped by an implicit class. This improves the order in which the CLI presents errors by making it follow the order of the source code more closely. 

### Testing
- Adapted test expect files to match the improved error message order. 


davidcok (commented):
I suggest that you check that this change in internal structure is OK for dafny doc

keyboardDrummer:
Closing for now because for code generation, it would be nicest to combine the implicit classes after resolution, but the current compiler architecture makes that difficult to implement because it breaks the references to the implicit classes and they can not easily be updated.

PR #4231, opened 2023-06-30
Renames related to project aware ide

keyboardDrummer:


robin-aws:
I'm not sure I'm sold on the various `Document` -> `Compilation` renamings actually, just because to me "Dafny compilation" means cross-compiling to one of the various target languages instead. The name is probably justified here but I'd like to understand a bit better.

keyboardDrummer:
> I'm not sure I'm sold on the various `Document` -> `Compilation` renamings actually, just because to me "Dafny compilation" means cross-compiling to one of the various target languages instead. The name is probably justified here but I'd like to understand a bit better.

The `CompilationManager` organizes parsing resolution translation and verification, and `Compilation` and its various subclasses contain the results of those runs.

robin-aws (approved):

PR #4234, opened 2023-06-30
fix: division for newtypes in JS and PY

### LINKED ISSUE #4233: JS/Py select incorrect division for newtype real
### Dafny version

4.0.0

```dafny
newtype MyReal = real

method Main() {
  var x: MyReal := 12.0 / 10.0;
  print x, "\n"; // expect 1.2

  if x == 1.0 {
    assert false;
    print "unreachable\n";
  }
}
```

```code
dafny run test.dfy --target:js

OR

dafny run test.dfy --target:py
```

### What happened?

Both JavaScript and Python emit malformed code. This causes the executing JavaScript code to crash and the Python code to compute the wrong result.

Mac

RustanLeino:
Fixes #4233

This PR corrects the type tests that the JS and PY compilers had done on newtypes.


Test/comp/Numbers.dfy
```diff
@@ -562,3 +563,99 @@ method CastRegressions() {
...
+    print h0, " ", h1, " ", h2, " ", h3, "\n"; // 98382635059784275282 0 98382635059784275282 0
+  }
+
+  method TestWraparound() {
+    var s0: bv3 := 6 + 3; // 1
+    var s1: bv3 := -1; // 7
+    var s2: bv3 := 4 + 5 - 2; // 7
+    var s3: bv7 := 126 + 3; // 1
+    var s4: bv7 := -1; // 127
+    var s5: bv7 := 64 + 65 - 2; // 127
```
fabiomadge, line 638:
```suggestion
    var s0: bv3 := 6 + 3;
    var s1: bv3 := -1;
    var s2: bv3 := 4 + 5 - 2;
    var s3: bv7 := 126 + 3;
    var s4: bv7 := -1;
    var s5: bv7 := 64 + 65 - 2;
```
Skipping them would be consistent with the other methods.

docs/dev/news/4234.fix
```diff
@@ -0,0 +1 @@
+Select proper division for newtype reals
```
fabiomadge, line 1:
```suggestion
Select proper division for real-based newtypes
```

fabiomadge (approved):
Thanks for fixing this! I was under the impression that real-based newtypes are a recent addition, which is probably why I made the mistake in the first place.

RustanLeino:
Thanks for the quick review, and for the comments. I'm afraid the web page showed me your comments only after I merged. I'll include them in my next PR.

PR #4235, opened 2023-06-30
Improve printing of real numbers

RustanLeino:
This PR improves how the runtimes on the various target platforms prints real numbers. In particular, the PR prints more numbers in decimal notation, rather than as fractions, and it makes the algorithm used more uniform across the various targets. Previously, some runtimes used decimal notation if the denominator is a power of 10, whereas others used decimal notation if _some multiple of_ the denominator is a power of 10. Now, all of them do the latter.

Out of scope for this PR: Some runtimes simplify the rational numbers (dividing both numerator and denominator by their GCD) before printing, whereas others do not. It would be nice to make that consistent, too. But since this simplification comes from some native target-platform library, the change is out of scope for this PR.


fabiomadge (approved):
Thanks!

PR #4237, opened 2023-07-03
Do not publish an error saying module resolution is not done because …

keyboardDrummer:
Preparation for https://github.com/dafny-lang/dafny/pull/3902

### Changes

Error messages of the form `ErrorsInRelatedModules.dfy(135,9): Error: not resolving module 'ClientOfErroneousModule5' because there were errors in resolving its import 'ModuleWithErrors'`, are no longer emitted. The user can expect that some resolution errors will block the resolution of other parts of the program, so this error is not useful. Moreover, this error can produce too much noise because it can cascade. Suppose you have 1 module with an error and a large tree of modules that are importing this one module. Previously all of the modules would produce an error and only 1 of those would be useful, now only that useful one is shown.

### Testing
Updated existing tests


robin-aws (changes requested):
Heavily in favor of this change. Only thing I'm concerned about is that I expected more changes to the test files, and that this could use a bit more testing. One of the two failing tests is an accidental test from the reference manual, and the content of the reference manual states that `Client` shows errors that it won't because it is never resolved! 

I propose we add at least one explicit regression test for this, and think through whether removing this error could cause unsoundness: if we have `A.dfy` with no errors that imports a module with errors from `B.dfy` using `--library`, could this change result in incorrect successful verification?

We should also update the reference manual to note that modules that import modules with errors won't be resolved - the example that `Modules.5.expect` is from seems to be http://dafny.org/dafny/DafnyRef/DafnyRef#sec-provided-and-revealed-names

keyboardDrummer:
> I propose we add at least one explicit regression test for this, and think through whether removing this error could cause unsoundness: if we have `A.dfy` with no errors that imports a module with errors from `B.dfy` using `--library`, could this change result in incorrect successful verification?

For resolution the rule is that everything gets resolved and all errors get reported, regardless of whether you're using `--library`. As such, with this change, in your example errors from `B.dfy` would still get reported and verification wouldn't start.

I'm not sure what test you're thinking of. Are you asking for one that specifically uses `--library`? I'll add it.

> We should also update the reference manual to note that modules that import modules with errors won't be resolved

Is that useful? I'm guessing there's many places where we chose to do an early exit. For example, when there's a parse failure then we won't do any resolution, which I expect we'll change in the future. Do you want to find all early exits and list them in the manual? How about generically saying that some errors in a program may prevent Dafny from reporting others?

> the example that Modules.5.expect is from seems to be http://dafny.org/dafny/DafnyRef/DafnyRef#sec-provided-and-revealed-names

I don't think the example needs to be updated. It's fine to have `// error: ...` in the example even if that error is only shown after resolving other errors. Before this PR that example already didn't show the errors in Client even though they're mentioned in the comments.

davidcok:
I think that is already stated somewhere. And is certainly the case with many other tools.

robin-aws (approved):
Alright, I can live with the current state of documentation. :)

stefan-aws (approved):
Reapproved after a small change.

Issue #3055, opened 2022-11-14
[Bug]: Default values of nested arrays broken in Java


RustanLeino:
### Dafny version

3.9.1.41027

### Code to produce this issue

```dafny
method Main() {
  ArrayX<bool>();
  ArrayInt();
  ArrayBool();
  ArrayClass();
}

class Class { }

method ArrayX<X>() {
  var x1: array<X>; // works!
  var x2: array2<X>; // works!

  var x11: array<array<X>>; // works!
  var x12: array<array2<X>>;
  var x21: array2<array<X>>; // works!
  var x22: array2<array2<X>>;

  var x1111: array<array<array<array<X>>>>;
  var x1112: array<array<array<array2<X>>>>;
  var x1121: array<array<array2<array<X>>>>;
  var x1122: array<array<array2<array2<X>>>>;
  var x1211: array<array2<array<array<X>>>>;
  var x1212: array<array2<array<array2<X>>>>;
  var x1221: array<array2<array2<array<X>>>>;
  var x1222: array<array2<array2<array2<X>>>>;
  var x2111: array2<array<array<array<X>>>>;
  var x2112: array2<array<array<array2<X>>>>;
  var x2121: array2<array<array2<array<X>>>>;
  var x2122: array2<array<array2<array2<X>>>>;
  var x2211: array2<array2<array<array<X>>>>;
  var x2212: array2<array2<array<array2<X>>>>;
  var x2221: array2<array2<array2<array<X>>>>;
  var x2222: array2<array2<array2<array2<X>>>>;
}

method ArrayInt() {
  var i1: array<int>; // works!
  var i2: array2<int>; // works!

  var i11: array<array<int>>;
  var i12: array<array2<int>>;
  var i21: array2<array<int>>;
  var i22: array2<array2<int>>;

  var i1111: array<array<array<array<int>>>>;
  var i1112: array<array<array<array2<int>>>>;
  var i1121: array<array<array2<array<int>>>>;
  var i1122: array<array<array2<array2<int>>>>;
  var i1211: array<array2<array<array<int>>>>;
  var i1212: array<array2<array<array2<int>>>>;
  var i1221: array<array2<array2<array<int>>>>;
  var i1222: array<array2<array2<array2<int>>>>;
  var i2111: array2<array<array<array<int>>>>;
  var i2112: array2<array<array<array2<int>>>>;
  var i2121: array2<array<array2<array<int>>>>;
[...]

PR #4240, opened 2023-07-04
feat: Compile general traits with type parameters

RustanLeino:
This PR implements support in all 5 compilers for type parameters in types that extend traits. In particular, the compilation support covers datatypes and codatatypes that extend traits. This functionality is available as part of the `/generalTraits:1` beta.

(After this PR, there will be one more PR to complete compilation support for general traits. It has to do with `newtype`s that extend traits.)

## More details

Here follows a description of what this PR implements and why.

Consider the following types, which are already supported by Dafny:

``` dafny
class Class<T(0)> { // note the auto-init type parameter
  method M() {
    var t: T; // the compiler auto-initializes "t", since "T" is an auto-init type
    print t;
  }
}

datatype Datatype<T(0)> = Datatype { // note the auto-init type parameter
  method M() {
    var t: T; // the compiler auto-initializes "t", since "T" is an auto-init type
    print t;
  }
}
```

The implementations of methods `Class.M()` and `Datatype.M()` need a type descriptor for `T`. Before this PR, where this type descriptor comes from is different from a class and a datatype.

For the class, the type descriptor is stored in each object. Let's call it "Design A". This design makes the type descriptor easily accessible from the implementation of `Class.M()` and saves the caller from having to create such a type descriptor with each call to `M()`. On the downside, it means that the type descriptor occupies space in each object.

For the datatype, the design is the opposite. Let's call it "Design B". Under this design, the type parameter is passed into `Datatype.M()` as a parameter at run time, which has the pros/cons of Design A's cons/pros.

To support traits and dynamic dispatch, Design B is no longer an option. Consider the following variation of the types above:

``` dafny
trait Parent {
  method M()
}

class Class<T(0)> extends Parent {
  method M() {
    var t: T;
    print t;
  }
}

datatype Datatype<T(0)> extends Parent = Datatype {
  method M() {
    var t: T;
    print t;
  }
}

method Main() {
  var c: Class<int> := new Class<int>;
  var d: Datatype<bool> := Datatype;
  var p: Parent;
  p := c;
  p.M();
  p := d;
  p.M();
}
```

Here, the calls to `Parent.M()` (which are dynamically dispatched to `Caller.M()` and `Datatype.M()`, respectively) do not take any type-descriptor parameters, because there are no type parameters involved in `Parent.M()`. Hence, the only way for an implementation of `M()` in a child type that extends `Parent` to get hold of the necessary type parameters (which are specific to the child type) is for the child-type value to contain the type descriptors. So, we must choose Design A for any type (class, datatype, ...) that can extend a trait.

So, this PR changes the run-time representation of (co)datatypes to include, in the (co)datatype values themselves, the necessary type descriptors. While there is a cost to this at run time, it is not common for a (co)datatype to have an auto-init type parameter, so the target code in most cases will be unchanged. _Except_ for Java, which clumsily adds type descriptors for _all_ type parameters (not just auto-init type parameters). For several reasons (for example, see https://github.com/dafny-lang/dafny/issues/3055), this would be good to change in the (hopefully near) future.

This change also affects the "downward clone" mechanism used by the C# target.

## Alternative

Rather than adding type descriptors as fields in the run-time representation of classes and datatypes, one could include such type descriptors in the vtable that each such class/datatype already has a pointer to at run time. However, there are two issues that make such a change difficult. One issue is that such vtables are controlled by the target language. Thus, for each required instantiation of the type descriptors, one would need to declare a separate leaf type in the target language. The other issue is that the required instantiations may be difficult to compute statically.

So, while this alternative may be considered in the future (especially for a native compiler for Dafny), it is out of scope for this PR.

### DIFF
--- a/Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
+++ b/Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
@@ -265,27 +265,57 @@ namespace Microsoft.Dafny.Compilers {
 
           var wCtorParams = new ConcreteSyntaxTree();
           wCtorBody = wBody.Format($"public {name}({wCtorParams})").NewBlock();
-
-          if (typeParameters != null) {
-            var sep = "";
-            foreach (var ta in TypeArgumentInstantiation.ListFromFormals(typeParameters)) {
-              if (NeedsTypeDescriptor(ta.Formal)) {
-                var fieldName = FormatTypeDescriptorVariable(ta.Formal.GetCompileName(Options));
-                var decl = $"{DafnyTypeDescriptor}<{TypeName(ta.Actual, wTypeFields, ta.Formal.tok)}> {fieldName}";
-                wTypeFields.WriteLine($"private {decl};");
-                if (ta.Formal.Parent == cls) {
-                  wCtorParams.Write($"{sep}{decl}");
-                }
-
-                wCtorBody.WriteLine($"this.{fieldName} = {TypeDescriptor(ta.Actual, wCtorBody, ta.Formal.tok)};");
-                sep = ", ";
-              }
-            }
-          }
+          EmitTypeDescriptorsForClass(typeParameters, cls, wTypeFields, wCtorParams, null, wCtorBody);
         }
       }
 
       return new ClassWriter(this, wBody, wCtorBody);
+    }
+
+    /// <summary>
+    /// For each type parameter X in "typeParametersForClass" that needs a type descriptor,
+    ///   * Write "protected TypeDescriptor<X> _td_X;" to wTypeFields
+    ///     -- each entry is terminated by a newline
+    ///   * Write "TypeDescriptor<X> _td_X" to wCtorParams
+    ///     -- entries are separated by a comma
+    ///   * Write "_td_X" to wCallArguments
+    ///     -- entries are separated by a comma
+    ///   * Write "this._td_X := _td_X;" to wCtorBody
+    ///     -- each entry is terminated by a newline
+    /// Any of the writer parameters may be null, so long as at least one is non-null.
+    /// The method returns the number type descriptors written.
+    /// </summary>
+    int EmitTypeDescriptorsForClass(List<TypeParameter> typeParametersForClass, TopLevelDecl cls,
+      [CanBeNull] ConcreteSyntaxTree wTypeFields, [CanBeNull] ConcreteSyntaxTree wCtorParams,
+      [CanBeNull] ConcreteSyntaxTree wCallArguments, [CanBeNull] ConcreteSyntaxTree wCtorBody,
+      List<TypeParameter> alternateTypeParameters = null) {
+
+      var wError = wTypeFields ?? wCtorParams ?? wCallArguments ?? wCtorBody;
+      int numberOfEmittedTypeDescriptors = 0;
+      if (typeParametersForClass != null) {
+        var sep = "";
+        var j = 0;
+        foreach (var ta in TypeArgumentInstantiation.ListFromFormals(typeParametersForClass)) {
+          if (NeedsTypeDescriptor(ta.Formal)) {
+            var fieldName = FormatTypeDescriptorVariable((alternateTypeParameters == null ? ta.Formal : alternateTypeParameters[j]).GetCompileName(Options));
+            var actualType = alternateTypeParameters == null ? ta.Actual : new UserDefinedType(ta.Formal.tok, alternateTypeParameters[j]);
+            var paramName = TypeDescriptor(actualType, wError, ta.Formal.tok);
+            var decl = $"{DafnyTypeDescriptor}<{TypeName(actualType, wError, ta.Formal.tok)}> {fieldName}";
+
+            wTypeFields?.WriteLine($"protected {decl};");
+            if (ta.Formal.Parent == cls) {
+              wCtorParams?.Write($"{sep}{decl}");
+            }
+            wCtorBody?.WriteLine($"this.{fieldName} = {paramName};");
+            wCallArguments?.Write($"{sep}{paramName}");
+
+            sep = ", ";
+            numberOfEmittedTypeDescriptors++;
+          }
+          j++;
+        }
+      }
+      return numberOfEmittedTypeDescriptors;
     }
 
     /// <summary>
@@ -308,7 +338,7 @@ namespace Microsoft.Dafny.Compilers {
 
       List<TypeParameter> typeDescriptorParams;
       if (enclosingTypeDecl is DatatypeDecl dtDecl) {
-        typeDescriptorParams = UsedTypeParameters(dtDecl);
+        typeDescriptorParams = UsedTypeParameters(dtDecl, true);
       } else {
         typeDescriptorParams = enclosingTypeDecl.TypeArgs;
       }
@@ -483,7 +513,7 @@ namespace Microsoft.Dafny.Compilers {
       var simplifiedType = DatatypeWrapperEraser.SimplifyType(Options, UserDefinedType.FromTopLevelDecl(dt.tok, dt));
       var simplifiedTypeName = TypeName(simplifiedType, wr, dt.tok);
 
-      // ContreteSyntaxTree for the interface
+      // ConcreteSyntaxTree for the interface
       wr.Write($"public interface _I{dt.GetCompileName(Options)}{TypeParameters(nonGhostTypeArgs, true)}");
       var superTraits = dt.ParentTypeInformation.UniqueParentTraits();
       if (superTraits.Any()) {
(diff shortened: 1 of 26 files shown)


Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
```diff
@@ -265,27 +265,57 @@ public class CsharpCompiler : SinglePassCompiler {
...
+    int EmitTypeDescriptorsForClass(List<TypeParameter> typeParametersForClass, TopLevelDecl cls,
+      [CanBeNull] ConcreteSyntaxTree wTypeFields, [CanBeNull] ConcreteSyntaxTree wCtorParams,
+      [CanBeNull] ConcreteSyntaxTree wCallArguments, [CanBeNull] ConcreteSyntaxTree wCtorBody,
+      List<TypeParameter> alternateTypeParameters = null) {
+
+      var wError = wTypeFields ?? wCtorParams ?? wCallArguments ?? wCtorBody;
+      int numberOfEmittedTypeDescriptors = 0;
+      if (typeParametersForClass != null) {
+        var sep = "";
+        var j = 0;
```
keyboardDrummer, line 297: Consider defining a method `Indexed` in `Util.cs`:

```c#
    public static IEnumerable<(T, int)> Indexed<T>(this IEnumerable<T> enumerable) {
      return enumerable.Select((index, value) => (index, value));
    }
```

and then here writing

```c#
foreach (var (index, ta) in TypeArgumentInstantiation.ListFromFormals(typeParametersForClass).Indexed()) {
```

and removing the variable `j`.

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
```diff
@@ -265,27 +265,57 @@ public class CsharpCompiler : SinglePassCompiler {
...
+    ///   * Write "TypeDescriptor<X> _td_X" to wCtorParams
+    ///     -- entries are separated by a comma
+    ///   * Write "_td_X" to wCallArguments
+    ///     -- entries are separated by a comma
+    ///   * Write "this._td_X := _td_X;" to wCtorBody
+    ///     -- each entry is terminated by a newline
+    /// Any of the writer parameters may be null, so long as at least one is non-null.
+    /// The method returns the number type descriptors written.
+    /// </summary>
+    int EmitTypeDescriptorsForClass(List<TypeParameter> typeParametersForClass, TopLevelDecl cls,
```
keyboardDrummer, line 288: Consider returning values instead of mutating arguments. It moves the work of creating some of these SyntaxTrees from the caller to the callee.

```c#

(int writtenDescriptors, 
 ICanRender typeFields, 
 ICanRender constructorParams,
 ICanRender callArguments,
 ICanRender constructorBody) EmitTypeDescriptorsForClass(
  List<TypeParameter> typeParametersForClass, 
  TopLevelDecl cls,
  List<TypeParameter> alternateTypeParameters = null)
{
  var typeFields = new ConcreteSyntaxTree();
  ..
}

```

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
```diff
@@ -265,27 +265,57 @@ public class CsharpCompiler : SinglePassCompiler {
...
+        foreach (var ta in TypeArgumentInstantiation.ListFromFormals(typeParametersForClass)) {
+          if (NeedsTypeDescriptor(ta.Formal)) {
+            var fieldName = FormatTypeDescriptorVariable((alternateTypeParameters == null ? ta.Formal : alternateTypeParameters[j]).GetCompileName(Options));
+            var actualType = alternateTypeParameters == null ? ta.Actual : new UserDefinedType(ta.Formal.tok, alternateTypeParameters[j]);
+            var paramName = TypeDescriptor(actualType, wError, ta.Formal.tok);
+            var decl = $"{DafnyTypeDescriptor}<{TypeName(actualType, wError, ta.Formal.tok)}> {fieldName}";
+
+            wTypeFields?.WriteLine($"protected {decl};");
+            if (ta.Formal.Parent == cls) {
+              wCtorParams?.Write($"{sep}{decl}");
```
keyboardDrummer, line 307: You can imagine a class `CommaSeparated` with a method `void Append(ICanRender element)` that automatically manages the `,` separator, so you replace:

```c#
        var wTypeFields = wr.Fork();
        var wCtorParams = new ConcreteSyntaxTree();
        var wCtorBody = wr.Format($"public {IdName(dt)}({wCtorParams})").NewBlock();
        EmitTypeDescriptorsForClass(dt.TypeArgs, dt, wTypeFields, wCtorParams, null, wCtorBody);
```

with

```c#
        var wTypeFields = wr.Fork();
        var wCtorParams = new CommaSeparated();
        var wCtorBody = wr.Format($"public {IdName(dt)}({wCtorParams})").NewBlock();
        EmitTypeDescriptorsForClass(dt.TypeArgs, dt, wTypeFields, wCtorParams, null, wCtorBody);
```

and 

```c#
int EmitTypeDescriptorsForClass(List<TypeParameter> typeParametersForClass, TopLevelDecl cls,
      [CanBeNull] ConcreteSyntaxTree wTypeFields, [CanBeNull] ConcreteSyntaxTree wCtorParams,
      [CanBeNull] ConcreteSyntaxTree wCallArguments, [CanBeNull] ConcreteSyntaxTree wCtorBody,
      List<TypeParameter> alternateTypeParameters = null) {
```

with 

```c#
int EmitTypeDescriptorsForClass(List<TypeParameter> typeParametersForClass, TopLevelDecl cls,
      [CanBeNull] ConcreteSyntaxTree wTypeFields, [CanBeNull] CommaSeparated wCtorParams,
      [CanBeNull] CommaSeparated wCallArguments, [CanBeNull] ConcreteSyntaxTree wCtorBody,
      List<TypeParameter> alternateTypeParameters = null) {
```

and then remove the variable `sep` in `EmitTypeDescriptorsForClass`.

And if you do, you can also consider adding an argument to `CommaSeparated` to manage to enclosing delimiters, in this case parenthesis, so you get

```c#
        var wTypeFields = wr.Fork();
        var wCtorParams = new CommaSeparated('(');
        var wCtorBody = wr.Format($"public {IdName(dt)}{wCtorParams}").NewBlock();
        EmitTypeDescriptorsForClass(dt.TypeArgs, dt, wTypeFields, wCtorParams, null, wCtorBody);
````

File: Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs [hunk 4]
```diff
@@ -265,27 +265,57 @@ public class CsharpCompiler : SinglePassCompiler {
...
-                if (ta.Formal.Parent == cls) {
-                  wCtorParams.Write($"{sep}{decl}");
-                }
+      return new ClassWriter(this, wBody, wCtorBody);
+    }
 
-                wCtorBody.WriteLine($"this.{fieldName} = {TypeDescriptor(ta.Actual, wCtorBody, ta.Formal.tok)};");
-                sep = ", ";
-              }
+    /// <summary>
```
keyboardDrummer, line 275: Adding constructor initialized fields to a class doesn't seem specific to C#. Can't this code be made more language agnostic?

Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -945,6 +933,51 @@ protected class ClassWriter : IClassWriter {
...
+    ///   * Write "TypeDescriptor<X> _td_X" to wCtorParams
+    ///     -- entries are separated by a comma
+    ///   * Write "_td_X" to wCallArguments
+    ///     -- entries are separated by a comma
+    ///   * Write "this._td_X := _td_X;" to wCtorBody
+    ///     -- each entry is terminated by a newline
+    /// Any of the writer parameters may be null, so long as at least one is non-null.
+    /// The method returns the number type descriptors written.
+    /// </summary>
+    int EmitTypeDescriptorsForClass(List<TypeParameter> typeParametersForClass, TopLevelDecl cls,
```
keyboardDrummer, line 949: What's the difference between this code and the C# version?

Source/DafnyCore/Compilers/GoLang/Compiler-go.cs
```diff
@@ -1281,37 +1306,32 @@ protected class ClassWriter : IClassWriter {
...
+          wInit?.WriteLine("_this.{0} = {0}", name);
+          wParams?.Write($"{sep}{name} {DafnyTypeDescriptor}");
+          sep = ", ";
           count++;
         }
       }
       return count;
     }
 
-    void WriteRuntimeTypeDescriptorsFormals(List<TypeParameter> typeParams, bool useAllTypeArgs, ConcreteSyntaxTree wr) {
```
keyboardDrummer, line 1304: Consider returning `ICanRender` instead of taking and mutating `ConcreteSyntaxTree`, so more work is done by the callee.

Source/DafnyRuntime/DafnyRuntimeDafny/src/dafnyRuntime.dfy
```diff
@@ -376,7 +376,7 @@ abstract module {:options "/functionSyntax:4"} Dafny {
       requires inv(t)
   }
 
-  trait {:extern} Sequence<T> {
+  trait {:extern} Sequence<T(==)> {
```
robin-aws, line 379: Note from offline conversation: this isn't actually true in general, so the better fix is to remove the `UniqueElements()` function and workaround that instead.

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
RustanLeino: I changed the in-parameters to out-parameters and reworked the call sites.

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
RustanLeino: I like this suggestion and I had thought about it when writing the code (because SO much of this code is spent getting commas and other delimiters in place!). However, it wasn't clear to me when `CommaSeparated` would look to see when to insert commas. That still isn't clear to me. If we could work this out, we'd save a lot of code in the compilers.

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
RustanLeino: This would be great to rework for all the compilers. A much better architecture would be to rewrite the Dafny code into some Dafny intermediate form, and then each compiler would have a much simpler task. But I consider such a change a separate endeavor.

Source/DafnyCore/Compilers/Java/Compiler-java.cs
RustanLeino: There is indeed a LOT of near-duplication of code throughout the compilers. But I find it out of scope for this PR to fix that.

keyboardDrummer: Even when introducing more of that near duplication? Are you proposing some future larger change that would resolve multiple duplication problems at once?

keyboardDrummer (approved):

keyboardDrummer (approved):

PR #4244, opened 2023-07-04
fix: Handle depending on modules from --library files

### LINKED ISSUE #4183: `/library` lets various compilers forget to include imports
### Dafny version

4.1.0

```dafny
// Bug_lib.dfy

module M {
  const i := 1
}
```

```dafny
// Bug.dfy

include "Bug_lib.dfy"

module CC {
  import M
  const i := M.i
}

method Main(){
  print CC.i, "\n";  
}
```

```code
dafny /compile:4 /compileTarget:go /library:Bug_lib.dfy Bug.dfy
```

### What happened?

The flag hinders the module discovery during compilation.

Mac

robin-aws:
Fixes #4183

The Go and Python backends, and likely others in the future, have a simple strategy of tracking all modules they have been asked to compile, and inserting imports for all previous modules every time they create a new one. If a module is referenced but not to be compiled because of `--library`, though, there is no call to `CreateModule`, and future modules that may reference it fail to emit imports. This change adds a new `DependOnModule` hook in `SinglePassCompiler` so that such backends can just track the necessary imports without creating the module.

Also expanded the `comp/separate-compilation` test case to cover more languages, as it exposes this exact problem. It turns out because the Javascript compiler emits code that only works in a single file (i.e. does not export any symbols) it doesn't support separate compilation at all. Cut https://github.com/dafny-lang/dafny/issues/4245 to track that and added a Feature to document this explicitly.

Also addressed the fact that the uniform backend testing scan wasn't flagging tests that use `dafny translate`, and marked/converted a handful of additional tests as a result.


keyboardDrummer (approved):
Looks great as usual :-)

PR #4255, opened 2023-07-06
Inlining overhaul for test generation

Dargones:
This PR overhauls implementation of inlining, on which test generation relies to generate system-level tests, and which can in the future be used for other purposes, such as bounded model checking. Inlining, in this context, means adjusting the Dafny to Boogie translation in such a way that the resulting Boogie program, when you call Boogie's `Inline` method on it, merges all specified Dafny functions and methods, including all corresponding well-formedness checks, into a single procedure. The key AST transformation necessary to make this happen is removal of short circuiting expressions from the original program. Most of what this PR does has to do with lifting this short-circuit-removal pass from Boogie to Dafny. This achieves three things:

- The new implementation is much more efficient. The old implementation can generate unit tests and inline one or two methods but not entire programs. Once Boogie 3.0 is merged into Dafny, the speed up should increase further.
- The new implementation preserves Dafny line/column information to report coverage in terms of Dafny source rather than its Boogie representation. Making this change has been one of the main asks pertaining to test generation.
- The new code should be, I think, easier to maintain going forward. It is split into disjoined transformations over Dafny/Boogie ASTs (here is the [main coordinating method](https://github.com/Dargones/dafny/blob/a5004d6d62d324433b57937363605b8ba975dca6/Source/DafnyTestGeneration/Inlining/InliningTranslator.cs#L22-L47)) and makes as few assumptions about the translation process as possible (whereas the previous implementation would rely very heavily on position and existence of various constructs that Dafny translator creates)


Source/DafnyTestGeneration/ProgramModification.cs
```diff
@@ -131,25 +156,35 @@ private enum Status { Success, Failure, Untested }
         return counterexampleLog;
       }
       var log = writer.ToString();
+      // If Dafny finds several assertion violations (e.g. because of inlining one trap assertion multiple times),
+      // pick the first execution trace and extract the counterexample from it
```
atomb, line 160: Is this always the right trace?

Source/DafnyTestGeneration/ProgramModification.cs
```diff
@@ -173,40 +208,26 @@ private enum Status { Success, Failure, Untested }
...
       }
       if (Options.Verbose) {
         await dafnyInfo.Options.OutputWriter.WriteLineAsync(
           $"// Test for {uniqueId} matches a test previously generated " +
-          $"for {duplicate.uniqueId}. This happens when test generation tool " +
-          $"does not know how to differentiate between counterexamples, " +
-          $"e.g. if branching is conditional on the result of a trait instance " +
-          $"method call.");
+          $"for {duplicate.uniqueId} - this may occur if the code under test is non-deterministic, " +
+          $"if a method/function is not inlined, or if test generation cannot extract a value from a counterexample.");
```
atomb, line 225: Same comment about "counterexample" as before.

atomb (changes requested):
This looks like a great improvement. I have a couple of suggestions about error message wording, and a couple of questions, but nothing major.

Source/DafnyTestGeneration/ProgramModification.cs
Dargones: The choice of the trace shouldn't matter since I replace all native assertions with assumptions and so the only assertion that can fail is the one that is being targeted. One can imagine an optimization that picks a trace covering the most blocks but in terms correctness the choice should not affect the result.

Source/DafnyTestGeneration/Inlining/AddByMethodRewriter.cs
```diff
@@ -0,0 +1,78 @@
...
+  // determines whether byMethod body should be added to a function
+  private readonly Func<MemberDecl, bool> shouldProcessPredicate;
+
+  protected internal AddByMethodRewriter(ErrorReporter reporter, Func<MemberDecl, bool> shouldProcessPredicate)
+    : base(reporter) {
+    this.shouldProcessPredicate = shouldProcessPredicate;
+  }
+
+  internal void PreResolve(Program program) {
+    AddByMethod(program.DefaultModule);
```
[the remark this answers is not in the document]
Dargones: It shouldn't - all modules are submodules of the default module

Source/DafnyTestGeneration/Main.cs
```diff
@@ -169,20 +164,41 @@ public static class Main {
...
+          ProgramModification.Status.Success);
+        int blocks = implementation.Blocks.Where(block => Utils.GetBlockId(block) != block.Label).ToHashSet().Count;
+        int coveredByCounterexamples = cache.NumberOfBlocksCovered(implementation);
+        int coveredByTests = cache.NumberOfBlocksCovered(implementation, onlyIfTestsExists: true);
+        yield return $"// Out of {blocks} locations in the " +
+                     $"{implementation.VerboseName.Split(" ")[0]} method, " +
+                     $"{coveredByTests} should be covered by tests " +
+                     $"(assuming no tests were found to be duplicates of each other). " +
+                     $"Moreover, {coveredByCounterexamples} locations have been found to be reachable " +
+                     $"(i.e. the verifier returned a counterexample and did not timeout). " +
```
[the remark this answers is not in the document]
Dargones: Done - paraphrased to "...the verifier did not timeout and produced example inputs to reach these locations"

atomb:
Those two failures aren't your fault. One will be fixed by #4317 and the other is non-deterministic.

atomb (approved):

Dargones:
Got it, thank you @atomb!

atomb (approved):

PR #4265, opened 2023-07-07
Add the compile suffix option

### LINKED ISSUE #4252: No longer adding `_Compile` when compiling non-extern modules breaks native code interoperabiility
Changed in this PR: https://github.com/dafny-lang/dafny/pull/4136. Too late to revert realistically and a good change, but very breaking for any projects like https://github.com/aws/aws-encryption-sdk-dafny that have target language code that refers to names like `Wrappers_Compile`.

I propose adding a compiler plugin or CLI option to modify the compile name of all modules, so affected users can add `_Compile` back if desired. This will be a useful feature for other reasons anyway, such as compiling all Dafny code under a desired parent package/namespace/etc. on a per target language basis.

keyboardDrummer:
Fixes #4252 (at least offers an easy fix for affected users)

### Changes
- Add the option `--compile-suffix` that brings back the old behavior of appending `_Compile` to module names when doing code generation

### Testing
- Manually tested on large customer codebase


robin-aws (dismissed):
Really appreciate this!

robin-aws (dismissed):

robin-aws (approved):

PR #4271, opened 2023-07-11
feat: support Find References LSP request

### LINKED ISSUE #2320: Support the "Find references" LSP request
Support this LSP request type: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references

### Prerequisites
- https://github.com/dafny-lang/dafny/issues/2329

### Implementation hints
AST Nodes can implement `IHasUsages` which allow them to report which declarations their references resolved to. For example a `break;` statement reports a usage to the loop which it is breaking out of, and as such you can ctrl+click on a break to jump to the loop.

An `IdentifierExpr` reports a usage to the variable which it points to.

The IHasUsages ASTNodes are used to create a `SymbolTable` which already has methods `Location? GetDeclaration(Position position)` and `ISet<Location> GetUsages(Position position)`, that can be and is available from `IdeState`. GetUsages can be used to implement "Find references"

alex-chew:
Fixes #2320.

- [x] Implement basic functionality
- [x] Add tests (WIP)
- [x] Update release notes


Source/DafnyLanguageServer/Handlers/DafnyReferencesHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+    public override async Task<LocationContainer> Handle(ReferenceParams request, CancellationToken cancellationToken) {
+      var document = await projects.GetResolvedDocumentAsync(request.TextDocument);
+      if (document == null) {
+        logger.LogWarning("location requested for unloaded document {DocumentUri}", request.TextDocument.Uri);
+        return new LocationContainer();
+      }
+
+      // If the position is not on a definition, return references to its definition
+      var definition = document.SymbolTable.GetDeclaration(request.Position);
+      if (definition != null) {
```
keyboardDrummer, line 39: Consider having only a single call to `GetUsages`, with a `var definitionPosition = definition == null ? request.Position : definition.Range.Start`

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
```diff
@@ -0,0 +1,278 @@
...
+          Position = position
+        }, CancellationToken);
+    }
+
+    /// <summary>
+    /// Assert that when finding-references at each cursor position,
+    /// the client returns all ranges marked with regular spans.
+    /// </summary>
+    /// <param name="source"></param>
+    private async Task AssertReferences(string source) {
```
keyboardDrummer, line 29: Sweet :)

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
```diff
@@ -0,0 +1,278 @@
...
+      MarkupTestFile.GetPositionsAndRanges(
+        source, out var cleanSource, out var positions, out var expectedRangesArray);
+      var expectedRanges = new HashSet<Range>(expectedRangesArray);
```
keyboardDrummer, line 32: Seems like you could extract positions using `expectedRangesArray.Select(r => r.Start)[.Concat(expectedRangesArray.Select(r => r.End))]` to replace many of `><` tokens you have in the tests.

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
```diff
@@ -0,0 +1,278 @@
...
+    case None => 0
+    case Some(value) => value + o.[>><value<]
+  ;
+}
+".TrimStart();
+
+      await AssertReferences(source);
+    }
+
+    // It seems that type declaration/usage information is not tracked
```
keyboardDrummer, line 163: Probably resolved by adding implementing `IHasUsages` on more AST nodes, but we can fix that in another PR

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
```diff
@@ -0,0 +1,278 @@
...
+      Assert.Single(positionsA);
+      Assert.Single(positionsB);
+      Assert.Single(rangesB);
+
+      var cwd = Directory.GetCurrentDirectory();
+      var pathA = Path.Combine(cwd, "Lookup/TestFiles/A.dfy");
+      var pathB = Path.Combine(cwd, "Lookup/TestFiles/B.dfy");
+
+      var documentItemA = CreateTestDocument(cleanSourceA, pathA);
+      await client.OpenDocumentAndWaitAsync(documentItemA, CancellationToken);
```
keyboardDrummer, line 262: Only `pathA` is visible to the language server at this point right? `pathB` is neither in memory nor on disc

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
```diff
@@ -0,0 +1,278 @@
...
+method M(><x: int) {
+  print [>><x<];
+}
+".TrimStart();
+
+      await AssertReferences(source);
+    }
+
+    [Fact]
+    public async Task Module() {
```
keyboardDrummer, line 101: I wonder whether find references on definitions that may be visible outside of the file, should be enabled when not in project mode, since it might cause too few results to be returned leading to confusion. 

One option is to do the least amount of work, which is to have it be enabled as you have now, and then gather feedback from users.

The other option is to for each definition say whether it's local to the file or not, and only if it is to return references results outside of project mode.

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
alex-chew: Good idea, I've implemented something along those lines to simplify the tests' markup.

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
alex-chew: I think that's a good idea, and I agree that it would be better to address in another PR.

On a related note, it would be nice if the declaration graph were reflexive on declarations, i.e. `GetDeclarations(myDecl.Position)` would include `myDecl` itself. This seems moreintuitive to me, and it would also simplify the implementation of `DafnyDefinitionHandler.Handle`.

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
alex-chew: Thanks, I misunderstood that `CreateTestDocument` doesn't automatically make the contents available for the resolver/LS to read. I've created appropriate files in `TestFiles` to make the test work.

Source/DafnyLanguageServer/Handlers/DafnyReferencesHandler.cs
alex-chew: This method ended up becoming more complex in order to handle cross-file references, so I don't think this suggestion is a good idea anymore (it would make the new code harder to follow).

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
alex-chew: I think whether or not we return references of definitions outside of the current file, it will be true that (a) the list of references will be incomplete in the general case (and so users may be confused), and (b) we will want to somehow encourage users to adopt the project mode.

To address (b) I'll look into ways we can warn the users that results may be incomplete, and guide the users towards documentation on how to add an appropriate project file. But then under those circumstances, I'd rather give a "best effort" list of references, i.e. including some-but-maybe-not-all references/declarations outside the current file, rather than intentionally withholding them.

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
keyboardDrummer: I don't follow. Instead of `GetDeclarations ` and `DafnyDefinitionHandler.Handle`, do you mean `GetUsages` and `DafnyReferencesHandler.Handle` ?

Note that LSP allows the client to request usages of a declaration that doesn't return the declaration itself:

image

Source/DafnyLanguageServer/Handlers/DafnyReferencesHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    protected override ReferenceRegistrationOptions CreateRegistrationOptions(
+      ReferenceCapability capability, ClientCapabilities clientCapabilities) {
+      return new ReferenceRegistrationOptions {
+        DocumentSelector = DocumentSelector.ForLanguage("dafny")
+      };
+    }
+
+    public override async Task<LocationContainer> Handle(ReferenceParams request, CancellationToken cancellationToken) {
```
keyboardDrummer, line 30: This implementation doesn't respect `request.context.includeDeclaration`

Source/DafnyLanguageServer/Handlers/DafnyReferencesHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+      // The declaration graph is not reflexive, so the position might be on a declaration; return references to it
+      if (declaration == null) {
+        return document.SymbolTable.GetUsages(request.Position).ToArray();
+      }
+
+      // If the position is not on a declaration, return references to its declaration
+      var definingDocument = declaration.Uri == document.Uri
+        ? document
+        : await projects.GetResolvedDocumentAsync(declaration.Uri);
+      return definingDocument?.SymbolTable.GetUsages(declaration.Range.Start).ToArray() ?? new LocationContainer();
```
keyboardDrummer, line 48: After the project aware PR you won't need to do this document switching, because then `document.SymbolTable.GetUsages(declaration.Range.Start)` will work even if `declaration.Uri != document.Uri`

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
keyboardDrummer: As long as you keep the document open after using `client.OpenDocumentAndWaitAsync`, it should be visible to all projects, as if it is on disk, so that's an alternative.

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
keyboardDrummer: For (b), I'm thinking of a notification that pops up whenever a file is opened that's not using project mode, that does include other files, and then ideally you could say "OK" or "OK and don't warn me again"

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
alex-chew: Yes, I did mean `GetUsages` and `DafnyReferencesHandle.Handle`.

I'll implement support for `includeDeclaration` being false.

alex-chew:
(The force-push rebases this branch onto master.)

Source/DafnyLanguageServer.Test/DafnyLanguageServer.Test.csproj
```diff
@@ -39,6 +39,12 @@
     <None Update="Lookup\TestFiles\foreign-verify.dfy">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="Lookup\TestFiles\find-refs-a.dfy">
```
keyboardDrummer, line 42: Nit: I think it's easier to read tests if all the relevant code is in the test. Alternatives to this are either opening these files in memory, or writing them to disk in the test.

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
```diff
@@ -0,0 +1,267 @@
...
+using OmniSharp.Extensions.LanguageServer.Protocol.Document;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Lookup {
+  public class ReferencesTest : ClientBasedLanguageServerTest {
+    private async Task<LocationContainer> RequestReferences(
+      TextDocumentItem documentItem, Position position) {
+      // We don't want resolution errors, but other diagnostics (like a cyclic-include warning) are okay
```
keyboardDrummer, line 17: Nit: I feel the method call implies what the comment says.

Source/DafnyLanguageServer.Test/Lookup/ReferencesTest.cs
```diff
@@ -0,0 +1,267 @@
...
+using Microsoft.Dafny.LanguageServer.IntegrationTest.Util;
+using OmniSharp.Extensions.LanguageServer.Protocol;
+using OmniSharp.Extensions.LanguageServer.Protocol.Document;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Lookup {
+  public class ReferencesTest : ClientBasedLanguageServerTest {
+    private async Task<LocationContainer> RequestReferences(
```
keyboardDrummer, line 15: Nit: I would move this and the next method to the bottom of the file, so the file is more callgraph topologically ordered.

Source/DafnyLanguageServer/Handlers/DafnyReferencesHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+      if (state == null) {
+        logger.LogWarning("location requested for unloaded document {DocumentUri}", request.TextDocument.Uri);
+        return new LocationContainer();
+      }
+
+      var requestUri = request.TextDocument.Uri.ToUri();
+      var declaration = state.SymbolTable.GetDeclaration(requestUri, request.Position);
+
+      // The declaration graph is not reflexive, so the position might be on a declaration; return references to it
+      if (declaration == null) {
```
keyboardDrummer, line 41: Nit: I would replace this with:
```c#
// The declaration graph is not reflexive, so the position might be on a declaration
FilePosition declarationPosition = declaration == null 
  ? new FilePosition(requestUri, request.Position) 
  : new FilePosition(declaration.Uri.ToUri(), declaration.Range.Start)

return state.SymbolTable.GetUsages(declarationPosition.Uri, declarationPosition.Position).ToArray();
```
```

keyboardDrummer (dismissed):
Looks great! Amazing

keyboardDrummer (approved):

PR #4274, opened 2023-07-11
Fix: Formatting within the IDE wasn't working anymore

### LINKED ISSUE #4269: Format `lemma` in the IDE
The following is what the IDE Format Document does.

The CLI does what you would expect.
(tested on both latest and nightly)

Note the indentation of `{` on `lemma`

```dafny
module Foo {

  function Baz()
    :(a: string)
    ensures a == "asdf"
  {
    "asdf"
  }

  lemma Bar(t: string)
    requires t == Baz()
    ensures t == "asdf"
    {}
}

```

### LINKED ISSUE #4270: Format parameters  the IDE
The following is what the IDE `Format Document` does.

The CLI does what you would expect.
(tested on both `latest` and `nightly`)

```dafny
module Foo {
  method Bar(
    a: map<string, string>,
           b: map<string, string>,
                  c: map<string, string>
  )
}

```

MikaelMayer:
This PR fixes #4269 and fixes #4270
I first added unit tests for the cases that were failing. What really solved the issue was that I had to implement proper cloning for 
- methods, that were not keeping BodyStartTok
- formals, that were not keeping RangeToken

Then I also added for all formatting test a case when it tries to clone all members before formatting, like VSCode does, which unveiled 3 more formatting issues:
```
method Test() {
  g := new ClassName.ConstructorName(
      argument1b, // Two extra spaces
      argument2b,
      argument3b
    );
  var g := new ClassName.ConstructorName(
           argument1, // Missing two extra spaces (when block mode activated)
           argument2,
           argument3
         );
...
  && unchanged(
       this,
       c
     )
     && old( // Extra undesired space here
         c.c
       ) == c.c
  && fresh(
       c.c
     )
```
I also solved the above issues by:
* Ensuring the `RangeToken` of `TypeRHS` is preserved after cloning by modifying its topmost cloning parent `AssignmentRHS`
* Ensure we iterate on pre resolved children to get `OwnedToken` (otherwise some cloned expressions were crashing)
* Ensure `UnchangedExpr` is cloned with the same pattern as everyone else (otherwise the `RangeToken` wasn't kept)


Source/DafnyCore/AST/Expressions/Operators/BinaryExpr.cs
```diff
@@ -323,6 +323,8 @@ public enum AccumulationOperand { None, Left, Right }
     if (cloner.CloneResolvedFields) {
       ResolvedOp = original.ResolvedOp;
     }
+
+    RangeToken = original.rangeToken;
```
keyboardDrummer, line 327: The call to base already does `RangeToken = cloner.Range(original.RangeToken);`, so this line seems obsolete

Source/DafnyCore/AST/Members/Method.cs
```diff
@@ -101,6 +102,22 @@ public class Method : MemberDecl, TypeParameter.ParentType, IMethodCodeContext,
     Contract.Invariant(Decreases != null);
   }
 
+  public Method(Cloner cloner, Method m) : base(cloner, m) {
```
keyboardDrummer, line 105: Could you rename `m` to `original`?

keyboardDrummer (changes requested):
Looks great, just one request for change :)

keyboardDrummer (approved):

PR #4276, opened 2023-07-11
Ensure that all code coverage class filters are applied

shadaj:
Multiple class filters must be passed in as a [single argument](https://github.com/danielpalme/ReportGenerator#usage--command-line-parameters), but previously they were passed as multiple `-classfilters:` of which only the first was actually applied.


shadaj:
For example, see logs from a previous build:
Screenshot 2023-07-11 at 5 36 24 PM

atomb (approved):
Nice catch. :)

PR #4283, opened 2023-07-13
feat: Benchmarking plugin (take two)

robin-aws:
Redoes #4215, which had to be reverted. Weakens the test that caused trouble due to the non-determinism of benchmarking (see second commit).


atomb (approved):
This seems like a reasonable change, since we can't guarantee to get a failure from JMH.

PR #4288, opened 2023-07-14
Fix: Gutter icons more robust

### LINKED ISSUE #4287: Gutter highlighting broken
This bug wasn't there in 4.1.0 but is in the latest nightly.

Create a new Dafny file.
Write manually (don't copy paste)
```
method Test() {
  assert false;
}
```

You'll get this::

image

MikaelMayer:
This PR fixes #4287
The computation of the number of lines was done improperly for empty files and wasn't migrated properly for non-empty files, which resulted in gutter icons to never expand as before.
I modified the tests so that it's now possible to fully simulate a variable number of line and more general insertions, which helped me first reproduce the issue and then ensure that the fix was correct.

Also, I renamed "Next" to "Replace" in the tests so that it better reflects what the edit is supposed to do: replace the entire line. I added the variant "Insert" and described https://github.com/dafny-lang/dafny/pull/4288/files#diff-81426824e7dbbe14c1b83ee3acd4998ddd7b565b09c355e835c7e101b930f4c5R184 what the variants mean
I made it possible also for the tests to perform several changes at once, so that I can simulate not just removing a comment like "//Insert:", but also replacing \n by newlines and \\ by single backslashes.


Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -370,14 +370,23 @@ VersionedTextDocumentIdentifier DocumentIdentifier
...
+    internal static Range ComputeRange(INode program, Uri uri) {
+      IToken? end = null;
+      if (program is Program dafnyProgram) {
+        end = dafnyProgram.DefaultModuleDef.RangeToken.StartToken;
+        while (end != null && end.Uri != uri) {
+          end = end.Next;
+        }
+      }
+      if (end == null) {
+        var fileNode = FindFileNode(program, uri);
```
keyboardDrummer, line 382: What was unreliable about the previous code? Using `dafnyProgram.DefaultModuleDef.RangeToken` seems unreliable to me. With projects there's no single main file. I'm not sure what the value of `dafnyProgram.DefaultModuleDef.RangeToken` will be.

MikaelMayer: FindFileNode returns null after the first time. I couldn't understand what was wrong there. I found at least one problematic case, which is when the parser fails so there is strictly nothing in dafnyProgram, so looking at the first node with an adequate token is not robust.
So, I based my change on what the formatter is doing, which is to obtain the first token of a file based on DefaultModuleDecl.
If you have better ideas I'll happily consider them.

keyboardDrummer: I see.

I think the better solution would be for Program to contain a list of FileModuleDefinition, where each has a StartToken with file specific URI even if parsing fails. These could replace DefaultModuleDef at some point but initially could be just for the token purpose.

I don't think `dafnyProgram.DefaultModuleDef.RangeToken` will work when in project mode, since there's no root file.

MikaelMayer:
I'm closing this one because the logic has been merged into main-4.2
The following commits did not make it though as they were linked to merging after the projects PR (@keyboardDrummer) so you might want to include them if you merge the projects PR first on master before mergin main-4.2
- https://github.com/dafny-lang/dafny/pull/4288/commits/095e688cfd4fbb5acb04da1be9a9f09e25403eac
- https://github.com/dafny-lang/dafny/pull/4288/commits/e5d0621b3087a748c447482fcbefcfc09c08ebae
- https://github.com/dafny-lang/dafny/pull/4288/commits/35a9b1e58b5978beb465e83cf40a7045ae2ed635
- https://github.com/dafny-lang/dafny/pull/4288/commits/953d0f43a71b1e3925d03f8514a43e1bdfa62c83

keyboardDrummer (approved):

Issue #3954, opened 2023-05-05
Output of text logger doesn't include as much context as normal Dafny errors


atomb:
### Dafny version

4.0.0

### Code to produce this issue

```dafny
method BadAbs(x: int) returns (r: int)
  ensures r > 0
{
    if(x < 0) {
        return -x;
    } else {
        return x;
    }
}
```

### Command to run and resulting output

```code
$ dafny verify --use-basename-for-filename --log-format text --isolate-assertions badabs.dfy
badabs.dfy(3,0): Error: a postcondition could not be proved on this return path
badabs.dfy(2,12): Related location: this is the postcondition that could not be proved

Dafny program verifier finished with 0 verified, 1 error

...
 
  Assertion batch 3:
    Outcome: Invalid
    Duration: 00:00:00.0671044
    Resource count: 6416

    Assertions:
      badabs.dfy(2,13): error is impossible: this is the postcondition that could not be proved

...
```

### What happened?

Note that the normal error output shows both the postcondition that could not be proved and the path on which it could not be proved. The text logger output does not.

To fix this, the `ProofObligationDescription` data type probably needs to allow related locations, and the main error in this case should be the one prefixed by `Error: ` in the normal Dafny error output.

### What type of operating system are you experiencing the problem on?

Mac

PR #4292, opened 2023-07-17
Don't say "error is impossible" anymore

atomb:
This was a confusing message and it's possible to do better.

This addresses #3954, but I don't think it entirely resolves it. To do that, I think it'd be best to extend ProofObligationDescription to include related locations, which will require Boogie changes.

This change helps the situation in the meantime, however.


Source/DafnyCore/Verifier/ProofObligationDescription.cs
```diff
@@ -1109,15 +1109,20 @@ public class AssignmentShrinks : ProofObligationDescriptionWithNoExpr {
...
-  public override string FailureDescription => msg;
+  public override string FailureDescription =>
+    comment + (errorMessage is not null ? $" ({errorMessage})" : "");
 
   public override string ShortDescription => "boilerplate triple";
 
-  private readonly string msg;
+  private readonly string errorMessage;
+  private readonly string successMessage;
+  private readonly string comment;
```
MikaelMayer, line 1121: There is a class named "ProofObligationDescriptionCustomMessages" that you can use to implement custom success and failure messages, so that we have a consistent interface. Could you make use it? I see you have a comment as well there, feel free to override and use comment + base.SuccessMessage

MikaelMayer (changes requested):
I had missed this "error is impossible", glad you fixed it. Just a small refactoring improvement and I'm good to go.

Source/DafnyCore/Verifier/ProofObligationDescription.cs
atomb: Ah, thanks for pointing that out! I hadn't noticed that you'd added that. :)

keyboardDrummer (approved):

keyboardDrummer (approved):

MikaelMayer (approved):
Great I like that. thanks for doing it !

PR #4294, opened 2023-07-17
fix: Avoid race condition in Java runtime sequence concatenation optimization

### LINKED ISSUE #1454: Lazy sequence concatenation in C# and Java runtimes is not thread-safe

robin-aws:
Fixes #1454.

Uses the same technique as the C# runtime, the only wrinkle being finding the left-most node in order to call `newCopied(length)` on it. This allows us to actually assert success in the multithreaded benchmark for this case.

This required being able to install the Java runtime to Maven local, so I've also applied the same changes @alex-chew had queued up for publishing this package.

Also added support for a `{:benchmarkTearDown}` attribute in the plugin, for once-per-benchmark cleanup, statistics reporting, etc.

### DIFF
--- a/Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/DafnySequence.java
+++ b/Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/DafnySequence.java
@@ -882,10 +882,10 @@ final class ConcatDafnySequence<T> extends LazyDafnySequence<T> {
                 leftBuffer = cseq.left;
                 rightBuffer = cseq.right;
                 if (leftBuffer == null || rightBuffer == null) {
+                    copier.copyFrom(cseq.ans);
+                } else {
                     toVisit.push(rightBuffer);
                     toVisit.push(leftBuffer);
-                } else {
-                    copier.copyFrom(cseq.ans);
                 }
             } else {
                 copier.copyFrom(seq);
--- a/Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/DafnySequence.java
+++ b/Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/DafnySequence.java
@@ -801,8 +801,8 @@ abstract class LazyDafnySequence<T> extends DafnySequence<T> {
 final class ConcatDafnySequence<T> extends LazyDafnySequence<T> {
     // INVARIANT: Either these are both non-null and ans is null or both are
     // null and ans is non-null.
-    private DafnySequence<T> left, right;
-    private volatile NonLazyDafnySequence<T> ans = null;
+    private volatile DafnySequence<T> left, right;
+    private NonLazyDafnySequence<T> ans = null;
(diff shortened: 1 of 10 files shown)


Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/DafnySequence.java
```diff
@@ -859,11 +879,13 @@ private NonLazyDafnySequence<T> computeElements() {
             if (seq instanceof ConcatDafnySequence<?>) {
                 ConcatDafnySequence<T> cseq = (ConcatDafnySequence<T>) seq;
 
-                if (cseq.ans != null) {
-                    copier.copyFrom(cseq.ans);
+                leftBuffer = cseq.left;
+                rightBuffer = cseq.right;
+                if (leftBuffer == null || rightBuffer == null) {
```
alex-chew, line 884: Isn't this condition backwards? i.e.
```suggestion
                if (leftBuffer != null && rightBuffer != null) {
```

alex-chew (changes requested):

Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/DafnySequence.java
robin-aws: Good catch. I flipped the if and else bodies instead to stay consistent with the other cases, though.

alex-chew (approved):
It's unfortunate that we don't have better tooling to test concurrency-related changes like this, but otherwise this LGTM.

keyboardDrummer (approved):

keyboardDrummer:
I don't follow what the Maven publishing is for. Can you elaborate?

PR #4302, opened 2023-07-18
Revert "Project aware ide (#3902)" in 4.2 branch

keyboardDrummer:
I'd rather not have this in a release before we do a bit more IDE testing.


robin-aws (approved):
Good call, better safe than sorry

PR #4303, opened 2023-07-19
Prevent a race condition related to ProjectManagerDatabase.OpenDocument

keyboardDrummer:
### Changes
1. Update `DafnyTextDocumentHandler.cs` so it properly awaits `ProjectManagerDatabase.OpenDocument`, so no new requests can be handled until the ProjectManager's constructor and `OpenDocument` method have been called. Without this change, a call to `CreateAndOpenTestDocument` could deadlock, because the hover request would come in while the OpenDocument request would still create a new compilation, causing the hover to wait for a dud compilation that would never be started. It's not clear why the hover wouldn't return as cancelled.
1. Update `CompilationManager` so it calls its ResolvedDocument and TranslatedDocument tasks when the `CompilationManager` is cancelled before being started.
1. Use `LanguageProtocolTestBase` instead of `LanguageServerTestBase`, which means more code re-use from the library
1. Do not Dispose the entire `ServiceCollection` but only the `ProjectManagerDatabase`, which is safer because we're disposing less but still enough. Hopefully this leads to fewer exceptions in the test runs.

### Testing
1. It's difficult to consistently reproduce the problem, but this change stabilizes the test `OnDiskChangesToProjectFileAffectCodeNavigation`
1. Add a test `CancelUnstartedCompilationLeadsToCancelledTasks`, which tests the second change in the above list.
1. Refactoring, no test needed
1. Refactoring, no test needed

- Add a test `UpdateDuringARequestWillCancelTheRequest`, which tests behavior that's hardcoded into the LSP library we're using, but is good to be aware of nonetheless.

### DIFF
--- a/Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
+++ b/Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
@@ -14,6 +14,7 @@ using OmniSharp.Extensions.LanguageServer.Protocol.Server;
 using OmniSharp.Extensions.LanguageServer.Server;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Dafny.LanguageServer.IntegrationTest.Various;
 using Microsoft.Dafny.LanguageServer.Language;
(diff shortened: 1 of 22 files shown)


Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
```diff
@@ -60,30 +59,47 @@ ensures true
...
-        ServerCommand.ConfigureDafnyOptionsForServer(dafnyOptions);
-        options.Services.AddSingleton(dafnyOptions);
-        ServerOptionsAction(options);
+      public AnonymousDisposable(Action action) {
+        this.action = action;
       }
 
-      var client = CreateClient(clientOptionsAction, NewServerOptionsAction);
-      await client.Initialize(CancellationToken).ConfigureAwait(false);
+      public void Dispose() => Interlocked.Exchange(ref action, null)?.Invoke();
```
MikaelMayer, line 86: What is Interlocked.Exchange doing?

MikaelMayer (approved):

Source/DafnyLanguageServer.Test/DafnyLanguageServerTestBase.cs
keyboardDrummer: Making sure the dispose action is only run once. Don't think it's relevant here but good defensive programming.

PR #4311, opened 2023-07-20
Refactoring: simplify computing the priority of implementation tasks

keyboardDrummer:
Store a token for the name symbol when creating a Boogie Implementation, and use this to compute the priority of evaluating implementation tasks


MikaelMayer (approved):
Simple but nice. I think BoogieRangeToken is the appropriate place to store three tokens, start/end and name.

PR #4319, opened 2023-07-21
Fix: #4315, allowing static reveal lemmas to reference non-static original procedures

### LINKED ISSUE #4315: Refining a module with a class containing opaque functions causes resolution error
### Dafny version

4.2.0

```dafny
module M1 {
  class C {
    opaque function f() : bool { true }
  }
}

module M2 refines M1 {
}
```

```code
$ ./Binaries/Dafny verify ~/test/scratch.dfy
    |
173 | <nonexistent line>
    |                     ^

/Users/ekanshdg/test/scratch.dfy[M2](173,20): Error: 'this' is not allowed in a 'static' context
1 resolution/type errors detected in scratch.dfy
```

### What happened?

Due to the change making reveal lemmas static, there is a resolution error when the static reveal lemma tries to set the fuel for its non-static originating function. This only gets triggered when a module is resolved again, like when it is being refined, because the reveal lemma is added after the resolution is complete.

Mac

EkanshdeepGupta:
This PR fixes #4315


Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
```diff
@@ -316,7 +316,49 @@ public partial class ModuleResolver {
...
             Contract.Assert(arg != null);
-            ResolveExpression(arg, resolutionContext);
+            if (!(Attributes.Contains(attributeHost.Attributes, "opaque_reveal") && attr.Name == "fuel" && arg is NameSegment)) {
+              ResolveExpression(arg, resolutionContext);
+            } else {
+              // Manually resolving NameSegments that are present in fuel attributes of reveal lemmas.
+              // This is because reveal lemmas are static and we want to allow a refence to non-static original procedures
+              // in static context in this setting.
+              //
+              // Most of the following code is copied from AnnotateRevealFunction() in OpaqueMemberRewriter.cs.
```
MikaelMayer, line 326: Can you please put this branch in a separate function, so that you can put it as the function's comment?

Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
```diff
@@ -326,6 +330,49 @@ public partial class ModuleResolver {
       }
     }
 
+    // <summary>
+    // Manually resolving NameSegments that are present in fuel attributes of reveal lemmas.
+    // This is because reveal lemmas are static and we want to allow a refence to non-static original procedures
```
MikaelMayer, line 335:
```suggestion
    // This is because reveal lemmas are static and we want to allow a reference to non-static original procedures
```

Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
```diff
@@ -326,6 +330,49 @@ public partial class ModuleResolver {
...
+    // in static context in this setting.
+    //
+    // Most of the following code is copied from AnnotateRevealFunction() in OpaqueMemberRewriter.cs.
```
MikaelMayer, line 338: Would you be able to refactor the common part out if the code is similar, so that we avoid code duplication? I haven't looked at the other side of the code.

docs/dev/news/4315.fix
```diff
@@ -0,0 +1 @@
+Fixed #4315, allowing static reveal lemmas to reference non-static original procedures
```
MikaelMayer, line 1:
```suggestion
Refining an abstract module with a class with an opaque function no longer crashes
```
Here it should be user-oriented.

MikaelMayer (changes requested):
A few requests, otherwise the fix looks good!

Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs
EkanshdeepGupta: I looked but cannot think of a clean way to refactor the common code. Also it is not so big to warrant refactoring I think.

MikaelMayer (approved):

PR #4323, opened 2023-07-22
Add new type inference, under /typeSystemRefresh:1

RustanLeino:
This PR contains the bulk of the long-awaited changes of the Dafny's new type system. Most of the changes are to implement a new type inference. This was necessary, because the previous implementation was very complicated and difficult to understand and modify. The new type inference is more straightforward (see, e.g., the description in https://github.com/dafny-lang/dafny/pull/4134). Unlike the previous type system, the new one supports traits over datatypes and lays the groundwork for some other type-system changes we planning in the months ahead.

The changes in this PR are under the option `--type-system-refresh` (legacy `/typeSystemRefresh:1`), except for some minor improvements (e.g., a better error source location for modify statements). (Also, to use traits over datatypes, it is also necessary to use `--general-traits` / `/generalTraits:1`.)

While this PR contains the core parts of the new type system, these changes are not final. However, **it makes sense to get these changes into the master branch at this time**, to more easily keep other changes in synch. For this reason, I don't know at what level of detail this PR ought to be reviewed. Because the changes are isolated to the `--type-system-refresh` option and because we should do a comprehensive review of the the system before we switch to making it the default in Dafny, there will be another chance to review anything that is more shallowly or leniently reviewed at this time.

From a functionality standpoint, the biggest thing missing from this PR is inference of subset types. For example, for a class `C`, this PR may infer the type `C?` rather than the non-null type `C`. In some cases, this will just make the verifier work harder, but in other cases (e.g., when a local variable has such a type and the local variable is modified inside a loop), the new type inference in this PR may generate an error. The additional precision that comes from considering subset types will be added in a separate PR.

One actual _change_ in the type system is worth highlighting: Consider a `trait Tr` and a `class Cl extends Tr`. In the type system of Dafny today, `Tr` is assignable to `Cl` without an explicit cast. But in the new type system, this require an explicit cast. (Note, what I just said applies to traits, not subset types. So, for example, there is no change in how an `int` can be assigned to a `nat`.)

## Basic approach

The basic approach (see also the description of https://github.com/dafny-lang/dafny/pull/4134, whose changes are included in this PR) is to

* First, infer _pre-types_. These are the parts of types that don't depend on theorem proving. In other words, they are what you get if you replace subset types with their base types.
* Then, make the types more precise by considering subset types. (This step is not included in this PR.)

## Additional notes

In the implementation, some of the resolver's fields had to be made more accessible (e.g., changing method `FreeVariables` to be `public`).

Error messages generated by the new type inference include the string `PRE-TYPE:`. This continues to be useful in debugging, but will be removed by the time we start recommending the new type system.

The PR disables CI coverage tests for the new code, because the test suite has not yet been changed to use the new type system.

The code contains some TODO's that will be filled in by later PRs. This is not standard practice, but I don't want to lose track of these TODO's and the code surrounding them is need to make everything compile and run.


keyboardDrummer:
> While this PR contains the core parts of the new type system, these changes are not final. However, it makes sense to get these changes into the master branch at this time, to more easily keep other changes in synch. For this reason, I don't know at what level of detail this PR ought to be reviewed. Because the changes are isolated to the --type-system-refresh option and because we should do a comprehensive review of the the system before we switch to making it the default in Dafny, there will be another chance to review anything that is more shallowly or leniently reviewed at this time.

Without having looked at the PR, I think the code should be reviewed at the same bar as any code, regardless of whether it's currently turned on or not or whether more changes are expected.

I don't expect this to hold for this PR, but suppose the state of the PR would be "it compiles but doesn't make sense apart from that", then I'd recommend not merging the PR, and dealing with worries about merge conflicts by making a smaller PR that only applies code moves in the original to prevent conflicts caused by that.

keyboardDrummer:
Is a significant part of the new code in this PR a copy of the previous resolver code? I'm asking because of the size of the PR. If so, how do you plan to keep the two in sync if bugfixes are done?

Also, I think this PR is too large to review to a normal bar. I think a more practical approach to developing this would have been to pick a tiny subset of the language and only make the pre-types work for that. Is there a subset of the tests that would still pass for a subset of the language? How much effort would it be to split those off?

Also, the approach of using `partial` classes to create smaller files is something I've used previously as a stop-gap, to work towards a more idiomatic solution of having smaller classes. I'd prefer actually having smaller files, and generally one class per file. There's lots of classes in this PR that could be moved into their own file.

Also, I think the resolution code that's specific to particular language constructs should reside near the definition of those constructs, such as in the class definition of the AST type for that construct, as an implementation of an abstract method.

jtristan (commented):
I suggest adding a markdown file in the PreType directory (maybe README.md for better integration with GitHub) that will provide a high-level description of the pre-type inference. You have shared a lot of useful information in the prelude of the PR, in discussions as part of previous PRs, in comments in this PR, sometimes in the code. We should collect all of this in one place. Also, while most of the code is very readable and doesn't need a lot of comments, it would be nice to comment a little more the basic classes used throughout, such as DPreType for example.

jtristan (dismissed):

jtristan (approved):

PR #4325, opened 2023-07-24
Dafny Test Coverage Report

Dargones:
This PR adds functionality for emitting JaCoCo-style coverage reports for Dafny code. A coverage report is a set of RangeTokens/CoverageLabel pairs, which are rendered like in the image below:

Screenshot 2023-07-24 at 1 56 12 AM

Each report is a time-stamped directory with an index.html file that summarizes the coverage and contains links to all other files:

Screenshot 2023-07-24 at 1 56 00 AM

Because there might be several notions of coverage in Dafny (coverage that Dafny expects automatically generated tests to have, actual runtime coverage on C#, Java, Python, etc.), I add a new command, called `merge-coverage-reports` that takes in several existing coverage reports and links them together so that it is easy to cross-reference them (see images above).


Source/DafnyCore/CoverageReporter/CoverageReport.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.Dafny;
+
+namespace DafnyCore.CoverageReporter; 
+
+public class CoverageReport {
+
+  private static int nextUniqueId = 0;
+
+  // INVARIANT: CoverageSpans are sorted within each list by the position of the StartToken
+  private readonly Dictionary<string, List<CoverageSpan>> labeledFiles;
```
atomb, line 15: I found this name confusing. Maybe `labelsByFile` or something like that would be better?

atomb (approved):
I'm a little wary that merging HTML files rather than a simpler data format could be fragile, but there is something appealing about it from the UI standpoint: there's no difference between what the computer interprets and what the human reads.

Source/DafnyCore/CoverageReporter/CoverageReport.cs
Dargones: Done!

Source/DafnyTestGeneration/GenerateTestsCommand.cs
```diff
@@ -86,6 +87,10 @@ private enum Mode {
     "Print the Boogie code used during test generation.") {
     ArgumentHelpName = "filename"
   };
+  public static readonly Option<string> PrintCoverage = new("--print-coverage",
```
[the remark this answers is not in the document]
Dargones: Have changed the option to `--coverage-report`

Source/DafnyCore/CoverageReporter/CoverageLabel.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using System.Linq;
+
+namespace DafnyCore.CoverageReporter;
+
+public enum CoverageLabel {
+  FullyCovered,
+  NotCovered,
+  PartiallyCovered
```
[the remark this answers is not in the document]
Dargones: I think that would make most sense. I will experiment with merging different coverage metrics - probably makes sense to separate that into its own PR

Dargones:
Thank you, @atomb! I am somewhat wary of the HTML parsing code as well but I think it would have to be done either way, unless there is some standard coverage format that fits all the criteria we need and an external tool that can automatically produce an html report from that same format?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -354,6 +354,8 @@ public enum IncludesModes {
     public bool UseStdin = false;
     public bool WarningsAsErrors = false;
     [CanBeNull] private TestGenerationOptions testGenOptions = null;
+    public List<string> CoverageReportsToMerge = new();
```
keyboardDrummer, line 357: Why do you add these fields here instead of a location that's specific to coverage?

Source/DafnyCore/assets/.resources/coverage.css
```diff
@@ -0,0 +1,90 @@
+body, td {
```
keyboardDrummer, line 1: Shouldn't this CLI UI be in DafnyDriver instead of DafnyCore?

Source/DafnyCore/CoverageReporter/CoverageReport.cs
```diff
@@ -0,0 +1,119 @@
...
+
+public class CoverageReport {
+
+  private static int nextUniqueId = 0;
+
+  // INVARIANT: CoverageSpans are sorted within each list by the position of the StartToken
+  private readonly Dictionary<string, List<CoverageSpan>> labelsByFile;
+  public readonly string Name; // the name to assign to this coverage report
+  public readonly string Units; // the units of coverage (plural). This will be written in the coverage report table.
+  private readonly int uniqueId = nextUniqueId++;
```
keyboardDrummer, line 18: I think this isn't thread-safe, and I imagine we could at some point run multiple tests concurrency in a single process that would generate coverage reports.

Consider using `Interlocked.Increment`

Source/DafnyCore/DafnyOptions.cs
Dargones: Done! I have moved them to `CoverageReportCommand.cs`

Source/DafnyCore/assets/.resources/coverage.css
Dargones: I think if I were to move these files to `DafnyDriver`, I would have to move the coverage reporting code there as well, so that they can be accessed from within the same assembly? My reason to put the assets in `DafnyCore` was that this mimics the set up for the Auditor.

Source/DafnyCore/CoverageReporter/CoverageReport.cs
Dargones: Done!

Source/DafnyCore/CoverageReporter/CoverageReport.cs
```diff
@@ -0,0 +1,119 @@
...
+  public string UniqueId => "." + (uniqueId == 0 ? "" : (uniqueId + ".")); // to add as postfix to files
```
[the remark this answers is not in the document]
Dargones: When writing to disk, I always create a new directory, so overwriting shouldn't be a problem. I added the option to use a user-supplied suffix in my latest commit, so the set up should be better now. Even so, the numerical suffix is still necessary because when you merge arbitrary coverage reports from disk, there must be some way to differentiate between them, and they might have been created with the same user-provided suffix.

Source/DafnyCore/assets/.resources/coverage.css
keyboardDrummer: > I think if I were to move these files to DafnyDriver, I would have to move the coverage reporting code there as well, so that they can be accessed from within the same assembly?

Sounds good.

> My reason to put the assets in DafnyCore was that this mimics the set up for the Auditor.

Isn't the auditor code a little all over the place?

Source/DafnyCore/CoverageReporter/CoverageReport.cs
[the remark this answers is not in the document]
keyboardDrummer: > Even so, the numerical suffix is still necessary because when you merge arbitrary coverage reports from disk, there must be some way to differentiate between them, and they might have been created with the same user-provided suffix.

Sorry I don't understand this. Why can't you merge files with the same filename that are in different directories?

Source/DafnyCore/CoverageReporter/CoverageReportCommand.cs
```diff
@@ -13,6 +13,9 @@ namespace DafnyCore.CoverageReporter;
 
 public class CoverageReportCommand : ICommandSpec {
 
+  public static List<string> CoverageReportsToMerge = new();
```
keyboardDrummer, line 16: Ah, but now you need statics, which aren't great either.  

Could you allow generically storing arguments in DafnyOptions. I put a draft of this idea in this PR: https://github.com/dafny-lang/dafny/pull/4410/files ?

Source/DafnyCore/assets/.resources/coverage.css
Dargones: I see, sorry, I think I misunderstood your original comment. I have now moved the `CoverageReportCommand.cs` and all the HTML parsing to `DafnyDriver`, with the tools for creating an in-memory coverage report remaining in `Dafny Core`.

Source/DafnyCore/CoverageReporter/CoverageReport.cs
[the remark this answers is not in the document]
Dargones: When serializing a report, I create an `.html` file for each `.dfy` file in the sources. When serializing several reports, I therefore have several `.html` files for a single original `.dfy` file and so I have to differentiate between them in some way. By default, I will use the user-provided suffix indicating what kind of coverage is being reported by each `.html` file. But if these suffixes are equal as well, I propose to use a numerical index.

Source/DafnyCore/CoverageReporter/CoverageReportCommand.cs
Dargones: Thank you, this has been super helpful!

Dargones:
Thank you, @keyboardDrummer! I have now isolated all HTML parsing inside `DafnyDriver`, with all abstract functionality remaining in `DafnyCore`. I also made use of the code you linked above to store options generically - this has been super useful.

keyboardDrummer (approved):

keyboardDrummer:
Thanks for changes. I would have moved the coverage code that's currently in Core to DafnyTestGeneration, since AFAIK that's the only place where it's used. I'm guessing you put it in core because you expect it will be used there in the future. However, it's cheap to move code later and I always prefer designing for today and adapting things to future needs when they come into play.

Also, I want to apologize for not having a crisp explanation of what should be in which assembly. My feeling is that we currently don't have the assembly responsibilities defined well. However, I think avoiding putting things in Core will make it easier to reorganize things later, so I generally advocate for that.

Source/DafnyCore/CoverageReporter/CoverageReportCommand.cs
keyboardDrummer: Woop :) thanks. Closing the PR then.

Dargones:
Thank you, Remy! I have moved the files from DafnyCore to DafnyTestGeneration. Like you have said, it should be very easy to relocate them later if necessary. @keyboardDrummer, @atomb - could you reapprove/merge?

keyboardDrummer (approved):

PR #4326, opened 2023-07-24
[Test Generation] Change the set of coverage criteria that can be targeted

Dargones:
This PR changes what coverage criteria a user of test generation may target. 

- Path coverage: Dafny will now attempt complete path coverage of the (inlined part of the) program as oppose to covering paths through each method in isolation. As a means of dealing with path explosion, Dafny will now try generating tests for increasingly longer paths, skipping over any paths whose subpaths have been found to be infeasible. 

- Branch coverage: As an middle ground between path coverage and block coverage, Dafny can now attempt call-graph-sensitive version of block coverage (i.e. it tries to cover every block in the program for every path through the call graph that can lead to that block). I call this "branch" coverage but am not sure this is the best way to call it - would be grateful for any suggestions.

This PR relies on a [recent change](https://github.com/boogie-org/boogie/pull/753) in Boogie that fixes a bug in how inlined programs are printed to a file. 


Source/DafnyCore/TestGenerationOptions.cs
```diff
@@ -8,7 +8,7 @@ public class TestGenerationOptions {
     public const string TestInlineAttribute = "testInline";
     public const string TestEntryAttribute = "testEntry";
     public bool WarnDeadCode = false;
-    public enum Modes { None, Block, Path };
+    public enum Modes { None, Block, Branch, Path };
```
atomb, line 11: You could call it `CallGraph` coverage?

Source/DafnyTestGeneration/PathBasedModifier.cs
```diff
@@ -23,54 +24,61 @@ public class PathBasedModifier : ProgramModifier {
...
-    /// </summary>
-    private void VisitImplementation(Implementation node) {
-      if (!ImplementationIsToBeTested(node) ||
-          !DafnyInfo.IsAccessible(node.VerboseName.Split(" ")[0])) {
-        return;
+        int pathLength = PathLengthStep;
+        bool newPathsFound = true;
+        // Consider paths of increasing length, pruning out infeasible sub-paths in the process:
+        while (newPathsFound) {
+          List<Path> pathsToConsider = new(); // paths without nown unfeasible subpaths
```
atomb, line 46:
```suggestion
          List<Path> pathsToConsider = new(); // paths without known unfeasible subpaths
```

Source/DafnyTestGeneration/PathBasedModifier.cs
```diff
@@ -90,55 +98,92 @@ public class PathBasedModifier : ProgramModifier {
...
+
+      public string ToString(DafnyOptions options) {
+        return $"{Impl.VerboseName.Split(" ")[0]} path through " +
+               $"{string.Join(",", pathBlocks.ConvertAll(block => Utils.GetBlockId(block, options) ?? block.UniqueId.ToString()))}";
+      }
+
+      /// <summary>
+      /// Constructs a binary tree of disjunctions made up of <param name="clauses"></param>
+      /// This limits the depth of the resulting AST and prevents stack overflow during verification for large trees
+      /// </summary>
```
atomb, line 172: The fact that this is necessary suggests there's something in Boogie to fix.

atomb (dismissed):
I made one suggestion about the name for the new mode. I'm not sure if it's a good one. ;)

Source/DafnyCore/TestGenerationOptions.cs
Dargones: I think CallGraph works great, thank you! Have renamed it

Source/DafnyTestGeneration/PathBasedModifier.cs
Dargones: I only encountered a stack overflow once when I had to construct a conjunction of about 1000 variables, so I don't think this is a common scenario but yes, it should be possible to optimize Boogie to avoid this.

Source/DafnyTestGeneration/GenerateTestsCommand.cs
```diff
@@ -49,9 +49,9 @@ private enum Mode {
   }
 
   private readonly Argument<Mode> modeArgument = new("mode", @"
-block - Prints block-coverage tests for the given program.
-branch - Prints branch-coverage tests for the given program.
-path - Prints path-coverage tests for the given program.");
+Block - Prints block-coverage tests for the given program.
+Branch - Prints call-graph-coverage tests for the given program.
```
atomb, line 53: Is this text right, now that you've renamed the constructor?

atomb (changes requested):
One tiny change!

Source/DafnyTestGeneration/GenerateTestsCommand.cs
Dargones: Oh, right, thank you for catching this. Have fixed it in my latest commit.

atomb (approved):

PR #4327, opened 2023-07-24
Improve handling of string options in Project file

keyboardDrummer:
### Changes

Previously
```
[options]
function-syntax = 3
```

Didn't do anything, because `function-syntax` is a string option so it expected `"3"` instead of `3`

### Testing
Adapted the test `ProjectFileOverridesOptions` so it tests the new behavior


jtristan (dismissed):

stefan-aws (approved):
Approved after offline discussion.

PR #4332, opened 2023-07-25
Document symbol for projects

keyboardDrummer:
### Changes
- Stop using `LspSymbolGeneratingVisitor` and instead rely on the AST to generate document symbols.

### Tests
- Add test for document symbols in multifile project


Source/DafnyCore/AST/IHasUsages.cs
```diff
@@ -0,0 +1,44 @@
...
+  File = 1,
+  Module = 2,
+  Namespace = 3,
+  Package = 4,
+  Class = 5,
+  Method = 6,
+  Property = 7,
+  Field = 8,
+  Constructor = 9,
+  Enum = 10,
```
MikaelMayer, line 27: What's an enum? A datatype?

Source/DafnyCore/AST/IHasUsages.cs
```diff
@@ -0,0 +1,44 @@
...
+  Variable = 13,
+  Constant = 14,
+  String = 15,
+  Number = 16,
+  Boolean = 17,
+  Array = 18,
+  Object = 19,
+  Key = 20,
+  Null = 21,
+  EnumMember = 22,
```
MikaelMayer, line 39: Same here. Is an enumMember a datatype constructor?

Source/DafnyCore/AST/IHasUsages.cs
```diff
@@ -0,0 +1,44 @@
...
+
+public interface ISymbol : IDeclarationOrUsage {
+  DafnySymbolKind Kind { get; }
+}
+
+public interface IHasSymbolChildren : ISymbol {
+  IEnumerable<ISymbol> ChildSymbols { get; }
+}
+
+public enum DafnySymbolKind {
```
MikaelMayer, line 17: Is this enumeration based off some language server numbering? If please, please consider putting an URL to the specification so that we can keep it in sync and ignore my two questions below.

Source/DafnyLanguageServer.Test/Lookup/DocumentSymbolTest.cs
```diff
@@ -30,7 +44,7 @@ class Y {
       var classChildren = classSymbol.Children.ToArray();
       Assert.Equal("Y", classSymbol.Name);
       Assert.Equal(new Range((0, 6), (0, 7)), classSymbol.SelectionRange);
-      Assert.Equal(new Range((0, 0), (9, 0)), classSymbol.Range);
+      Assert.Equal(new Range((0, 0), (9, 1)), classSymbol.Range);
```
MikaelMayer, line 47: Nice inclusion of the last character!

MikaelMayer (commented):
Great use of the INode construct to get rid of duplicate symbol traversers! And we will get the attributes for free, which we did not have before.

Since you actually modify core to make symbols available from there, could you please write a few tests in DafnyCore.Test as well?

keyboardDrummer:
> Since you actually modify core to make symbols available from there, could you please write a few tests in DafnyCore.Test as well?

Any suggestions for the shape of those? I'm skeptical that it's useful to add unit tests that act directly on the types to confirm that `((ClassDecl)obj).Kind == DafnySymbolKind.Class`

MikaelMayer (approved):
Nice addition. It might also make the semantic tokens easier to reimplement when I'll give it a go next time.

PR #4337, opened 2023-07-26
Hover and gutter icon project support

keyboardDrummer:
### Changes
- Enable complex hover messages and gutter icons when in project mode
- Do not publish gutter icons if the exact same icons have been previously published.

### Testing
- Update hover and gutter icon tests so they have project mode turned on


Source/DafnyCore/AST/TypeDeclarations/DatatypeDecl.cs
```diff
@@ -80,9 +80,6 @@ public abstract class DatatypeDecl : TopLevelDeclWithMembers, RevealableTypeDecl
     return base.IsEssentiallyEmpty();
   }
 
-  public override IEnumerable<ISymbol> ChildSymbols => base.ChildSymbols.Concat(Ctors);
-  public override DafnySymbolKind Kind => DafnySymbolKind.Enum;
```
MikaelMayer, line 84: This is not tested, so why deleting it?

keyboardDrummer: Err, that's accidental.

MikaelMayer (approved):

MikaelMayer (approved):

Source/DafnyCore/Options/DafnyProject.cs
```diff
@@ -19,8 +19,6 @@ public class DafnyProject : IEquatable<DafnyProject> {
 
   public string ProjectName => Uri.ToString();
 
-  public bool IsImplicitProject { get; set; }
```
MikaelMayer, line 22: What was this deletion about, by the way?

MikaelMayer (approved):

Source/DafnyCore/Options/DafnyProject.cs
keyboardDrummer: It makes it so that the IDE doesn't differentiate between implicit and explicit projects. Previously only implicit projects had features depending on the VerificationTree turned on

stefan-aws (dismissed):
Has been approved prior to a small change.

MikaelMayer (approved):
There is still a few CI check warnings

MikaelMayer (approved):

PR #4342, opened 2023-07-26
Feat: Default function opacity setting + autoreveal option

EkanshdeepGupta:
This PR adds an additional cmdline flag `--default-function-opacity` which controls the default opacity of functions from `transparent`, `autoRevealDependencies`, or `opaque`.

`transparent` is synonymous with the present behavior. `autoRevealDependencies` is a mode which makes functions opaque by default and inserts reveal stmts in each callable for each function dependency. `opaque` is a mode which makes functions opaque by default, leaving the user to insert all required function dependencies.


Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+
+            foreach (var expression in subExpressions) {
+              if (expression is FunctionCallExpr funcExpr) {
+                var func = funcExpr.Function;
+
+                if (isRevealedAlongPath(func, moduleDefinition, new List<ModuleDefinition> { func.EnclosingClass.EnclosingModuleDefinition })) {
+                  if (!func.IsOpaque && func.DoesAllOpaqueMakeOpaque(Options)) {
+                    var expr = cf.Rhs;
+
+                    var revealStmt0 = BuildRevealStmt(new FunctionReveal(func, new List<NameSegment>()),
```
MikaelMayer, line 42: Can you please create an overload of BuildRevealStmt so that you don't need to create an object just to call it? The first argument could just be "func".

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+                      var newExpr = new StmtExpr(cf.Rhs.Tok, revealStmt, origExpr) {
+                        Type = origExpr.Type
+                      };
+                      cf.Rhs = newExpr;
+                    }
+                  }
+                }
+              }
+            }
+          }
```
MikaelMayer, line 70: There is a lot of nesting here (maximum in this scope is 24 spaces). Could you please find ways to decrease nesting level to at most 8 spaces typically, perhaps 12 maximum?

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+        }
+      }
+    }
+  }
+
+  private class GraphTraversalVertex {
```
MikaelMayer, line 111: Can you please add a comment to this class, why it's needed?

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+          if (m is Constructor c) {
+            c.BodyInit.Insert(0, revealStmt);
+          } else {
+            m.Body.Body.Insert(0, revealStmt);
+          }
+        }
+      }
+    }
+
+    if (autoRevealDeps) {
```
MikaelMayer, line 160: Can you please merge these two ifs to reduce indentation?

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+          reqExpr = new StmtExpr(reqExpr.tok, revealStmt0, reqExpr);
+          reqExpr.Type = Type.Bool;
+        }
+
+        m.Req.Insert(0, new AttributedExpression(reqExpr));
+      }
+    }
+
+    if (addedReveals.Any()) {
+      reporter.Message(MessageSource.Rewriter, ErrorLevel.Info, null, m.tok, addedReveals.Count + " reveal stmt(s) inserted: \n" + RenderRevealStmts(addedReveals));
```
MikaelMayer, line 174:
```suggestion
      reporter.Message(MessageSource.Rewriter, ErrorLevel.Info, null, m.tok, $"{addedReveals.Count} reveal statement{(addedReveals.Count > 1 ? "s" : "")} inserted: \n{RenderRevealStmts(addedReveals)}");
```

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+
+  private void AddFunctionReveals(Function f, ErrorReporter reporter) {
+    Contract.Requires(f != null);
+
+    bool autoRevealDeps = Attributes.ContainsBoolAtAnyLevel(f, "autoRevealDependencies", true);
+
+    var currentClass = f.EnclosingClass;
+    List<RevealStmt> addedReveals = new List<RevealStmt>();
+
+    using IEnumerator<FunctionReveal> enumerator = GetEnumerator(f, currentClass, f.SubExpressions);
```
MikaelMayer, line 186: I see you using this using/while loop idiom a lot. Could you perhaps move this iteration logic somewhere else and just call a higher-order function? That might make things more legible and maintainable.

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    if (addedReveals.Any()) {
+      reporter.Message(MessageSource.Rewriter, ErrorLevel.Info, null, f.tok, addedReveals.Count + " reveal stmt(s) inserted: \n" + RenderRevealStmts(addedReveals));
+    }
+  }
+
+  private static string RenderRevealStmts(List<RevealStmt> revealStmtList) {
+    Contract.Requires(revealStmtList.Any());
+    return revealStmtList.Aggregate("", (s, stmt) => s + " " + stmt)[1..];
```
MikaelMayer, line 229: Use string.join() instead.

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+    typeList.ForEach(newType => {
+      if (newType is Type { AsSubsetType: not null }) {
+        foreach (var subexpression in newType.AsSubsetType.Constraint.SubExpressions) {
+          if (subexpression is FunctionCallExpr funcExpr) {
+
+            var func = funcExpr.Function;
+            var newVertex = func.EnclosingClass.EnclosingModuleDefinition.CallGraph.FindVertex(func);
+            if (newVertex is not null) {
+              queue.Enqueue(new GraphTraversalVertex(newVertex, false, defaultNameSegmentPath.ToList(), defaultModulePath.ToList()));
+              //TODO implement proper name lookup above
```
MikaelMayer, line 274: TODO here?
Also, please consider inversing ifs to reduce nesting.

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+    }
+  }
+
+
+  private static IEnumerable<Type> ExprListToTypeList(IEnumerable<Expression> exprList) {
+    if (exprList is null || !exprList.Any()) {
+      return new List<Type>();
+    }
+
+    var subSubexpressionTypeList = exprList.Select(expr => ExprListToTypeList(expr.SubExpressions));
```
MikaelMayer, line 469: What about "typeList" for a variable name? Since it's what you return.

Source/DafnyCore/Rewriters/OpaqueMemberRewriter.cs
```diff
@@ -80,7 +80,8 @@ public OpaqueMemberRewriter(ErrorReporter reporter)
     Contract.Requires(c != null);
     var newDecls = new List<MemberDecl>();
     foreach (var member in c.Members.Where(member => member is Function or ConstantField)) {
-      if (!Attributes.Contains(member.Attributes, "opaque") && !member.IsOpaque) {
+      if (!Attributes.Contains(member.Attributes, "opaque") && !member.IsOpaque &&
```
MikaelMayer, line 83: Can you please put this new conjunct in a separate function since it starts to be more complicated?

MikaelMayer (changes requested):
Ok this is in great shape! Comments are mostly about making code looking better.
I assume you will still add tests to test the two new possible command line switches as well.
Will you do a separate PR for the {:noAutoReveal} attribute?

Source/DafnyCore/AST/CompilationData.cs
```diff
@@ -75,6 +75,10 @@ public class CompilationData {
     result.Add(new LocalLinter(reporter));
     result.Add(new PrecedenceLinter(reporter, this));
 
+    if (Options.Get(CommonOptionBag.AllOpaque) == CommonOptionBag.DefaultFunctionOpacity.AutoRevealDependencies) {
```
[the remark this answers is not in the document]
EkanshdeepGupta: Ah, must've missed this when I migrated the interface!

EkanshdeepGupta:
> Ok this is in great shape! Comments are mostly about making code looking better. I assume you will still add tests to test the two new possible command line switches as well. Will you do a separate PR for the {:noAutoReveal} attribute?

We decided to go with `{:autoRevealDependencies false}`, to prevent ambiguity about whether the attribute prevents this function from being revealed elsewhere or its dependencies being revealed here.

I did add support for the `{:autoRevealDependencies}` attribute in this PR.

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -127,6 +129,20 @@ public class DafnyOptions : Bpl.CommandLineOptions {
       RegisterLegacyUi(CommonOptionBag.AddCompileSuffix, ParseBoolean, "Compilation options", "compileSuffix");
     }
 
+    private static void ParseDefaultFunctionOpacity(Option<CommonOptionBag.DefaultFunctionOpacity> option, Bpl.CommandLineParseState ps, DafnyOptions options) {
+      if (ps.ConfirmArgumentCount(1)) {
+        if (ps.args[ps.i].Equals("Transparent")) {
```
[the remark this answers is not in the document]
EkanshdeepGupta: Ah actually I copied the structure of this code from the ParsePrintMode() method on line 111. If you want I can change that and other parsers as well?

Source/DafnyCore/Options/ICommandSpec.cs
```diff
@@ -42,6 +42,7 @@ public interface ICommandSpec {
     BoogieOptionBag.SolverLog,
     CommonOptionBag.JsonDiagnostics,
     BoogieOptionBag.VerificationErrorLimit,
+    CommonOptionBag.AllOpaque
```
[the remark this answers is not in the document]
EkanshdeepGupta: Oh where are the language server's options stored? FWIW, I have been testing it with the IDE and it works smoothly.

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
EkanshdeepGupta: Tysm! I love this attention to detail.

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
```diff
@@ -0,0 +1,531 @@
...
+          if (importName is not null) {
+            newNameSegmentPath.Add(importName);
+          }
+
+          queue.Enqueue(new GraphTraversalVertex(callable, false, newNameSegmentPath, newModulePath.ToList()));
+        }
+      }
+    }
+
+    while (queue.Any()) {
```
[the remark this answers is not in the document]
EkanshdeepGupta: I'm not entirely sure what you mean.

Source/DafnyCore/DafnyOptions.cs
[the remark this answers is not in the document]
MikaelMayer: Ok if you just did copy paste this code from another place, let's leave it as it then. We will do the entire refactoring later.

Source/DafnyCore/Options/ICommandSpec.cs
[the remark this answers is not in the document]
MikaelMayer: Source/DafnyLanguageServer/ServerCommand.cs

Source/DafnyCore/AST/Members/Function.cs
```diff
@@ -33,6 +33,17 @@ public class Function : MemberDecl, TypeParameter.ParentType, ICallable, ICanFor
 
   public override bool IsOpaque { get; }
 
+  public bool DoesAllOpaqueMakeOpaque(DafnyOptions options) {
```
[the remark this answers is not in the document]
EkanshdeepGupta: Done.

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -106,7 +106,7 @@ public class DafnyOptions : Bpl.CommandLineOptions {
         argumentName: "Everything|DllEmbed|NoIncludes|NoGhost",
         defaultValue: PrintModes.Everything);
 
-      RegisterLegacyUi(CommonOptionBag.DefaultFunctionOpacity, ParseDefaultFunctionOpacity, "Language feature selection", "defaultFunctionOpacity", null, CommonOptionBag.DefaultFunctionOpacityOptions.Transparent);
+      RegisterLegacyUi(CommonOptionBag.DefaultFunctionOpacity, ParseDefaultFunctionOpacity, "Language feature selection", "defaultFunctionOpacity", null, CommonOptionBag.DefaultFunctionOpacityOptions.AutoRevealDependencies);
```
MikaelMayer: I think it should still say Transparent here, right?

MikaelMayer (changes requested):
Great job putting the documentation together.
Could you also link to your attributes from the "reveal" section?

Source/DafnyCore/DafnyOptions.cs
EkanshdeepGupta: You're right, I changed it to test something, then forgot to change back before merging!

docs/DafnyRef/Statements.md
```diff
@@ -2104,8 +2104,8 @@ an extended illustration of this technique to make all the dependencies of an `a
 
 ### 8.20.3. Revealing function bodies
 
-Normally function bodies are transparent and available for constructing proofs of assertions that use those functions.
-However, sometimes it is helpful to mark a function [`{:opaque}`](#sec-opaque) and treat it as an uninterpreted function, whose properties are
+By default, function bodies are transparent and available for constructing proofs of assertions that use those functions. 
+This can be changed using the `--defaul-function-opacity` commandline flag, or by using the [`:{opaque}`](#sec-opaque) attribute and treat it as an uninterpreted function, whose properties are
```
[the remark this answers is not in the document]
EkanshdeepGupta: Done.

Source/DafnyCore/AST/Modules/ModuleDefinition.cs
```diff
@@ -465,27 +465,30 @@ public class AccessibleMember {
     foreach (var d in TopLevelDecls) {
       if (d is AliasModuleDecl || d is AbstractModuleDecl) {
         ModuleSignature importSig;
-        if (d is AliasModuleDecl) {
-          var alias = (AliasModuleDecl)d;
-          importSig = alias.TargetQId.Root != null ? alias.TargetQId.Root.Signature : alias.Signature;
+        if (d is AliasModuleDecl alias) {
+          if (alias.TargetQId.Decl is not null) {
+            importSig = alias.TargetQId.Decl.Signature;
+          } else if (alias.TargetQId.Root is not null) {
```
MikaelMayer, line 471: It looks like a bug fix, is that the case? If so, what is the test file that tests this?

EkanshdeepGupta: This is part of the code constructing the AccessibleMembers list. It is not a bugfix.

Test/autoRevealDependencies/power.dfy
```diff
@@ -0,0 +1,13 @@
...
+
+function power(b: int, n: nat): int
+{
+  if (n==0) then 1 else b*power(b, n-1)
+}
+
+method test_power()
+  ensures power(power(2, 2), 2)==16
+{
+  //assert power(8, 5) == 32768;
```
MikaelMayer, line 12: Add a comment why this commented line is there or remove it

docs/DafnyRef/Attributes.md
```diff
@@ -487,7 +479,11 @@ where `Y` is `X` times either the default verification time limit
 for a function or method, or times the value specified by the
 Boogie `-timeLimit` command-line option.
 
-### 11.2.19. `{:verify false}` {#sec-verify}
+### 11.2.19. `{:transparent}` {#sec-transparent}
+
+By default, the body of a function is transparent to its users. This can be overridden using the `--default-function-opacity` command line flag. If default function opacity is set to `opaque` or `autoRevealDependencies`, then this attribute can be used on functions to make certain functions transparent.
```
MikaelMayer, line 484:
```suggestion
By default, the body of a function is transparent to its users. This can be overridden using the `--default-function-opacity` command line flag. If default function opacity is set to `opaque` or `autoRevealDependencies`, then this attribute can be used on functions to make them always non-opaque.
```

docs/DafnyRef/Types.md
```diff
@@ -4262,23 +4267,35 @@ A function is usually transparent up to some unrolling level (up to
...
-visible only if it was listed in the export set by which the contents
-of its module was imported. In that case, if the function was exported
-with `reveals`, the rules are the same within the importing module as when the function is used inside
-its declaring module. If the function is exported only with `provides` it is
-always opaque and is not permitted to be used in a reveal statement.
+Inside the module where the function is declared:
+  - If `--default-function-opacity` is set to `transparent` (default), then:
+     - if there is no `opaque` modifier, the function is transparent.
+     - if there is an `opaque` modifier, then the function is opaque. If the function is mentioned in a `reveal` statement, then
+     it's body is available in the callable where the `reveal` statement occurs.
```
MikaelMayer, line 4280:
```suggestion
     its body is available starting at that `reveal` statement.
```

MikaelMayer (changes requested):
All right, there are all the changes I will request, well done for this excellent job!

Source/DafnyCore/Resolver/CallGraphBuilder.cs
```diff
@@ -266,7 +266,16 @@ private class CallGraphASTVisitor : ASTVisitor<CallGraphBuilderContext> {
         ModuleDefinition callerModule = context.CodeContext.EnclosingModule;
         ModuleDefinition calleeModule = ((IASTVisitorContext)callee).EnclosingModule;
         if (callerModule != calleeModule) {
-          // inter-module call; don't record in call graph
+          // inter-module call; add edge in module's inter-module call graph
+          if (context.CodeContext is ICallable caller0) {
```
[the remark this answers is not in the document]
EkanshdeepGupta: I tried adding this guard, only to discover that the CallGraphBuilder class does not have access to Dafny options. Moreover, functions of this class are always called statically from outside. 

To implement this extra guard, I will have to add an extra parameter to every callable in this class to store whether to build the interModuleCallGraph. Should I do that? In my opinion that is a bigger change which probably isn't justified by the relatively small savings from not building the intermodule call graph.

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
```diff
@@ -0,0 +1,515 @@
...
+                  foreach (var newFunc in GetEnumerator(func, func.EnclosingClass, new List<Expression> { cf.Rhs },
+                             moduleDefinition)) {
+                    var origExpr = cf.Rhs;
+                    var revealStmt = BuildRevealStmt(newFunc.Function, cf.Rhs.Tok, moduleDefinition);
+
+                    if (revealStmt is not null) {
+                      var newExpr = new StmtExpr(cf.Rhs.Tok, revealStmt, origExpr) {
+                        Type = origExpr.Type
+                      };
+                      cf.Rhs = newExpr;
```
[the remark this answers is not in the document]
EkanshdeepGupta: Why is that? If we have `const c = p() + q()`, then we would want to insert reveal stmts for both p() and q(), right?

docs/DafnyRef/Attributes.md
EkanshdeepGupta: Done

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
```diff
@@ -0,0 +1,515 @@
...
+/// </summary>
+public class AutoRevealFunctionDependencies : IRewriter {
+  public AutoRevealFunctionDependencies(ErrorReporter reporter) : base(reporter) {
+  }
+
+  internal override void PostResolveIntermediate(ModuleDefinition moduleDefinition) {
+    Contract.Requires(moduleDefinition != null);
+
+    foreach (var decl in moduleDefinition.TopLevelDecls) {
+      if (decl is TopLevelDeclWithMembers cl) {
```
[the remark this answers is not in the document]
EkanshdeepGupta: Done. Added new interface and moved functionality inside each class

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -204,6 +204,20 @@ public enum TestAssumptionsMode {
...
+    AutoRevealDependencies,
+    Opaque
+  }
+
+  public static readonly Option<DefaultFunctionOpacityOptions> DefaultFunctionOpacity = new("--default-function-opacity", () => DefaultFunctionOpacityOptions.Transparent,
+    @"
+Change the default opacity of functions. 
+`Transparent` (default) means functions are transparent, can be manually made opaque and then revealed. 
+`AutoRevealDependencies` makes all functions non explicitly labelled as opaque to be opaque but reveals them automatically in scopes which do not have `{:autoRevealDependencies false}`. 
+`Opaque` means functions are always opaque so the opaque keyword is not needed, function must be revealed everywhere needed for a proof.".TrimStart()) {
```
atomb, line 218:
```suggestion
`Opaque` means functions are always opaque so the opaque keyword is not needed, and functions must be revealed everywhere needed for a proof.".TrimStart()) {
```

docs/DafnyRef/Types.md
```diff
@@ -4262,23 +4267,35 @@ A function is usually transparent up to some unrolling level (up to
...
-visible only if it was listed in the export set by which the contents
-of its module was imported. In that case, if the function was exported
-with `reveals`, the rules are the same within the importing module as when the function is used inside
-its declaring module. If the function is exported only with `provides` it is
-always opaque and is not permitted to be used in a reveal statement.
+Inside the module where the function is declared:
+  - If `--default-function-opacity` is set to `transparent` (default), then:
+     - if there is no `opaque` modifier, the function is transparent.
+     - if there is an `opaque` modifier, then the function is opaque. If the function is mentioned in a `reveal` statement, then
+     it's body is available starting at that `reveal` statement.
```
atomb, line 4280:
```suggestion
     its body is available starting at that `reveal` statement.
```

docs/DafnyRef/Attributes.md
```diff
@@ -218,7 +217,26 @@ function g(y:int, b:bool) : bool
 }
 ```
 
-### 11.2.3. `{:axiom}` {#sec-axiom}
+### 11.2.3. `{:autoRevealDependencies k}` {#sec-autorevealdependencies}
+When setting `--default-function-opacity` to `autoRevealDependencies`, the `{:autoRevealDependencies k}` attribute can be set on methods and functions to make sure that only function dependencies of depth `k` in the call-graph or less are revealed automatically. As special cases, one can also use `{:autoRevealDependencies false}` (or `{:autoRevealDependencies 0}`) to make sure that no dependencies are revealed, and `{:autoRevealDependencies true}` to make sure that all dependencies are revealed automatically.
```
atomb, line 221: Throughout the document, is the argument to `--default-function-opacity` written `AutoRevealDependencies` or `autoRevealDependencies`?

docs/DafnyRef/Statements.md
atomb, line 2108:
```suggestion
This can be changed using the `--default-function-opacity` command line flag, or by using the [`:{opaque}`](#sec-opaque) attribute and treat it as an uninterpreted function, whose properties are
```

Also, should this reference the `opaque` keyword instead of the attribute now?

Test/autoRevealDependencies/power.dfy.expect
```diff
@@ -0,0 +1,4 @@
+power.dfy(11,0): Error: a postcondition could not be proved on this return path
+power.dfy(10,31): Related location: this is the postcondition that could not be proved
```
atomb, line 2: It's not clear to me what purpose there is to testing that verification fails in this case. Would it be more valuable to test that it succeeds when the assertion is uncommented, to show that `autoRevealDependencies` hasn't broken the handling of recursive functions?

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
```diff
@@ -0,0 +1,309 @@
...
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Numerics;
+
+namespace Microsoft.Dafny;
+
+/// <summary>
+/// When Dafny is called with `--default-function-opacity autoRevealDependencies`, this rewriter computes
+/// all transitive functional dependencies for each callable, and inserts (in-memory, on the AST) reveal stmts for each such dependency
+/// at the top of the callable body, and also in a precondition.
```
atomb, line 12:
```suggestion
/// at the top of the callable body, and also in any preconditions.
```

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
```diff
@@ -0,0 +1,309 @@
...
+    // Removing the initial `\n\n`
+    return result[2..];
+  }
+
+  public static string GenerateMessage(List<RevealStmtWithDepth> addedReveals, int autoRevealDepth) {
+    var numInsertedReveals = addedReveals.Count(stmt => stmt.Depth <= autoRevealDepth);
+    var message = "";
+
+    message +=
+      $"Total {addedReveals.Count} function {(addedReveals.Count == 1 ? "dependency" : "dependencies")} found.";
```
atomb, line 102:
```suggestion
      $"Found {addedReveals.Count} function {(addedReveals.Count == 1 ? "dependency" : "dependencies")}.";
```

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
```diff
@@ -0,0 +1,309 @@
...
+
+    message +=
+      $" {numInsertedReveals} reveal {(numInsertedReveals == 1 ? "statement" : "statements")} inserted implicitly.";
+
+    // if (numInsertedReveals < addedReveals.Count) {
+    //   message += $" Remaining:\n{RenderRevealStmts(addedReveals, 1 + autoRevealDepth)}";
+    // } else {
+    message += $" Total stmts:\n{RenderRevealStmts(addedReveals)}";
```
atomb, line 110:
```suggestion
    message += $" Total number of reveal statements:\n{RenderRevealStmts(addedReveals)}";
```

Source/DafnyCore/AST/TypeDeclarations/SubsetTypeDecl.cs
```diff
@@ -50,4 +50,48 @@ public enum WKind { CompiledZero, Compiled, Ghost, OptOut, Special }
   public virtual string GetDescription(DafnyOptions options) {
     return "subset type";
   }
+
+  public void AutoRevealDependencies(AutoRevealFunctionDependencies Rewriter, DafnyOptions Options, ErrorReporter Reporter) {
+    if (Witness is null) {
+      return;
+    }
+
+    var expressions = Constraint.SubExpressions.ToList().Concat(Witness.SubExpressions.ToList());
```
atomb, line 59: It seems to me that there might be some shareable code between the function, method, constant, and subset type implementations of this method. We can leave that to a future PR, but I just wanted to note it for the record.

atomb (changes requested):
I requested a few tiny changes, mostly in documentation or error messages. Otherwise this looks great, though!

Source/DafnyCore/Rewriters/AllOpaqueRevealStmtInserter.cs
EkanshdeepGupta: Done

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
```diff
@@ -0,0 +1,515 @@
...
+                      };
+                      cf.Rhs = newExpr;
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      } else if (decl is SubsetTypeDecl { Witness: not null } expr) {
```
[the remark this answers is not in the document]
EkanshdeepGupta: Yes, I'm on it.

docs/DafnyRef/Statements.md
EkanshdeepGupta: Yes. I have moved the opaque attribute to the section of deprecated attributes, and changed the link to `#sec-opaque` to point to the section which discusses function transparency.

Source/DafnyCore/AST/TypeDeclarations/SubsetTypeDecl.cs
EkanshdeepGupta: It is just a sanity check to ensure that the feature is actually enabled. The program would successfully verify if functions were not made opaque, and the test would fail.

Test/autoRevealDependencies/power.dfy.expect
EkanshdeepGupta: It is a sanity check that the feature is actually working. The program would successfully verify if functions are not made opaque, and so the test would fail.

Source/DafnyCore/AST/TypeDeclarations/SubsetTypeDecl.cs
EkanshdeepGupta: Agreed.

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
EkanshdeepGupta: So actually, all the reveal stmts are collected in a single requires stmt (like `requires reveal f1(); reveal f2(); reveal f3(); true`), and inserted at the beginning, so I believe the original is more accurate.

docs/DafnyRef/Attributes.md
EkanshdeepGupta: That is actually an important point. Technically both work, but I am not sure which the canonical choice should be. The command-line documentation which picks the argument names (I think) from the underlying enum datatype, has everything capitalized (because in the codebase that is the norm), whereas in the command-line I think the norm is to keep the first letter lowercase in camelCase. What do you think?

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
MikaelMayer: Maybe "at the top of all clauses"?

docs/DafnyRef/Attributes.md
MikaelMayer: For me, on the CLI, it should start with a lowercase if possible

Source/DafnyCore/Resolver/CallGraphBuilder.cs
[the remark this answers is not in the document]
MikaelMayer: Since this is an instance function, you should be able to store a Dafny options as a field, so you don't need to change all instance functions.
For static methods, it's ok to put a DafnyOption as an arguments.

I don't know how much time is spent building the inter module call graph but it might be pretty long for codebases with 6000+ methods. If you don't want to do this change, could you please confirm that the extra time of doing that computation for a big project does not exceed 100ms?

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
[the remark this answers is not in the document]
MikaelMayer: `p() + q()` is a single RHS expression.
I'm talking about this iteration:
https://github.com/dafny-lang/dafny/pull/4342/files/853e95bdb48d77af644c0d53abbd251f83443c6f#diff-e931599e06715129f69d04cfaeb028c1e7ccd8c992890f6fc20809434b1e59cfR54

Source/DafnyCore/AST/Modules/ModuleDefinition.cs
```diff
@@ -73,12 +73,18 @@ public class ModuleDefinition : RangeNode, IDeclarationOrUsage, IAttributeBearin
...
   [FilledInDuringResolution]
   public readonly Graph<ICallable> InterModuleCallGraph = new();
 
   [FilledInDuringResolution]
   public int Height;  // height in the topological sorting of modules;
 
+  /// <summary>
+  /// The following class stores the relative name of any Declaration that is reachable from this module
+  /// as list of NameSegments, along with a flag for if the Declaration is revealed or merely provided.
+  /// For example, if "A" is a module, a function "A.f()" will be stored as f ~> {"A", "_default"} in the AccessibleMembers field.
```
MikaelMayer, line 86: What does that syntax `f ~> {"A", "_default"}` mean? Please explain.

Source/DafnyCore/AST/TypeDeclarations/SubsetTypeDecl.cs
MikaelMayer, line 59: Careful: Is SubExpressions taking ALL subexpressions or just the immediate sub-expressions?

EkanshdeepGupta: Ah good point. At some point I had implemented a method which iteratively collects subexpressions of a given expression (and its subexpressions' subexpressions and so on) and flattens them to a list. I should use that here.

Source/DafnyCore/Resolver/CallGraphBuilder.cs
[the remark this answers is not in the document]
EkanshdeepGupta: I think the cost of this computation should be small, since only the functions directly called by a method are added to the intermodule call graph, and each individual method usually has only about 6 to 10 "depth 1" dependencies.

What would be the best way to check the computation overhead for the entire project? Anything more sophisticated than running `time dafny verify ...` on a few files multiple times to get an average, and comparing the time taken on master vs my branch?

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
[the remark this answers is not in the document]
EkanshdeepGupta: Done. Refactored bunch of code to make dealing with expressions easier. Added uniform code for constFields and datatype constructors. Also added test case for datatype constructors.

Source/DafnyCore/AST/TypeDeclarations/SubsetTypeDecl.cs
EkanshdeepGupta: Done. Now it first flattens expressions to list of all subexpressions, then does the rest of the computation.

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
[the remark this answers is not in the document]
EkanshdeepGupta: Refactored the code. Now there is a separate method that takes an expression and returns a list containing it and all its subexpressions, and then we iterate over this list now.

Source/DafnyCore/AST/Modules/ModuleDefinition.cs
EkanshdeepGupta: AccessibleMembers is a `Dictionary<Declarations, AccessibleMember>` By `f ~> {"A", "_default"}` I meant that in this dictionary, `f` will be mapped to the AccessibleMember whose AccessPath list contains (the NameSegment versions of) "A" and "_default". I guess I should write it this verbosely in the comment itself?

EkanshdeepGupta: Done.

Source/DafnyCore/Resolver/CallGraphBuilder.cs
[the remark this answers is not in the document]
EkanshdeepGupta: Okay, I ran it multiple times both with and without the intermodule graph stuff commented out. The runtime numbers were noisy so hard to estimate exactly how much it is costing, but nothing appreciable since about half the time the present code was faster and half the time, the code with it commented out was faster.

docs/DafnyRef/Attributes.md
EkanshdeepGupta: Done. Uniformly changed it to "autoRevealDependencies".

MikaelMayer (approved):
Approved, only one small refactoring asked! Good job.

Source/DafnyCore/Resolver/CallGraphBuilder.cs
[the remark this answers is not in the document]
MikaelMayer: Thanks! I feel better.

Source/DafnyCore/Rewriters/AutoRevealFunctionDependencies.cs
```diff
@@ -439,7 +252,65 @@ private record FunctionWithDepth(Function Function, int Depth);
     return typeList.Concat(subExprTypeListConcat);
   }
 
-  private static RevealStmt BuildRevealStmt(Function func, IToken tok, ModuleDefinition rootModule) {
+  private static IEnumerable<Expression> ExprToSubexpressions(Expression expr) {
```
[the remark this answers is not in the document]
EkanshdeepGupta: Actually the AllSubExpressions method there works perfectly for what I want to do. Let me just use that.

MikaelMayer (dismissed):
Well done

atomb (dismissed):
Very nice work!

MikaelMayer (approved):

PR #4345, opened 2023-07-27
fix: Use a unique name for the entry point of programs compiled to Python

### LINKED ISSUE #4162: Python runtime error possibly due to refinement and all-caps module names
### Dafny version

4.1.0

```dafny
// Queue.dfyi

abstract module Queue {
    type Item
    type Queue = seq<Item>

    method Init() returns (q: Queue)
        ensures |q| == 0

    method Push(item: Item, q: Queue) returns (q': Queue)
        ensures |q'| == |q| + 1

    method Pop(q: Queue) returns (item: Item, q': Queue)
        requires |q| > 0
        ensures item in q
        ensures |q'| == |q| - 1
}

abstract module MainSpec {
    import Q : Queue
}

// LIFO.dfy

include "Queue.dfyi"

module LIFO refines Queue {
    type Item = int

    method Init() returns (q: Queue) {
        q := [];
    }

    method Push(item: Item, q: Queue) returns (q': Queue) {
        return [item] + q;
    }

    method Pop(q: Queue) returns (item: Item, q': Queue)
        ensures item == q[0]
    {
        item := q[0];
        q' := q[1..];
    }
}

module MainImpl refines MainSpec {
    import Q = LIFO

    method Main()
    {
        var q := Q.Init();
        q := Q.Push(0, q);
        q := Q.Push(1, q);
        q := Q.Push(2, q);

        var n: int;
        n, q := Q.Pop(q);
        print n, "\n";
        n, q := Q.Pop(q);
        print n, "\n";
        n, q := Q.Pop(q);
        print n, "\n";
    }
}
```

```code
% dafny run -t:py Test/exports/LIFO.dfy

Traceback (most recent call last):
  File "/Users/salkeldr/Documents/GitHub/dafny/Test/exports/LIFO-py/LIFO.py", line 10, in <module>
    assert "LIFO" == __name__
           ^^^^^^^^^^^^^^^^^^
AssertionError
```

### What happened?

Not yet sure what's going on, possibly some name mangling in Python due to the all-caps module name. Changing it to `Lifo` in code (without changing the file name) leads to a different error: 

```
% dafny run -t:py Test/exports/LIFO.dfy

Traceback (most recent call last):
  File "/Users/salkeldr/Documents/GitHub/dafny/Test/exports/LIFO-py/LIFO.py", line 6, in <module>
    import module_
  File "/Users/salkeldr/Documents/GitHub/dafny/Test/exports/LIFO-py/module_.py", line 9, in <module>
    import Lifo
ModuleNotFoundError: No module named 'Lifo'
```

Mac

fabiomadge:
Fixes #4162.


Source/DafnyDriver/Commands/RunCommand.cs
```diff
@@ -30,6 +30,7 @@ class RunCommand : ICommandSpec {
   public IEnumerable<Option> Options =>
     new Option[] {
       Inputs,
+      CommonOptionBag.Output
```
keyboardDrummer, line 33: Are you using this option? I think the concept of `output` for `dafny run` is different than for existing commands where it occurs, since run doesn't have producing files as a purpose. It just has intermediate files it may produce to run the program.

Could the ideal be that we have a build folder for intermediate artifacts, and a separate output folder for commands that produce output files?

fabiomadge: As you can probably tell, I need it to hit the bug this PR fixes. We could rewrite the compiler tester to manually run the generated sources, but I don't think this duplication of functionality is worth keeping the options list smaller.

If you go down the route you've sketched, I suggest generating a unique folder name and deleting it after finishing running.

keyboardDrummer: > As you can probably tell, I need it to hit the bug this PR fixes. We could rewrite the compiler tester to manually run the generated sources, but I don't think this duplication of functionality is worth keeping the options list smaller.

Could you add this option using a new option that's hidden? That way, you can use it in the test but it's not exposed in the public API. Also, maybe you can already make it be named `--build` so it's a little like what I sketched.

> If you go down the route you've sketched, I suggest generating a unique folder name and deleting it after finishing running. 

If we go down that route? :D

I think a unique folder that's deleted after running would be good yes, as a behavior for when you don't specify an explicit build directory.

keyboardDrummer (approved):

PR #4350, opened 2023-07-28
Compute static hover using ast

keyboardDrummer:
### Changes
- Let the statically computed part of hover information also work with project mode
- Compute statically part of hover mode using the Dafny AST instead of the `LegacySymbolTable`

### Testing
- Turned on project mode in `HoverTest.cs`
- Let half of tests in `HoverTest.cs` use a project file


Source/DafnyCore/AST/Expressions/Comprehensions/ComprehensionExpr.cs
```diff
@@ -441,8 +441,11 @@ public ComprehensionExpr(IToken tok, RangeToken rangeToken, List<BoundVar> bvars
       Bounds = original.Bounds?.Select(b => b?.Clone(cloner)).ToList();
     }
   }
-  public override IEnumerable<Node> Children => (Attributes != null ? new List<Node> { Attributes } : Enumerable.Empty<Node>()).Concat(SubExpressions);
-  public override IEnumerable<Node> PreResolveChildren =>
+  public override IEnumerable<INode> Children =>
+    (Attributes != null ? new List<Node> { Attributes } : Enumerable.Empty<Node>()).
+    Concat(BoundVars).Concat(SubExpressions);
```
MikaelMayer, line 446: Good work of finding the missing Boundars!

Source/DafnyCore/AST/Expressions/Variables/CasePattern.cs
```diff
@@ -102,6 +102,6 @@ public class CasePattern<VT> : TokenNode
     }
   }
 
-  public override IEnumerable<Node> Children => Arguments ?? Enumerable.Empty<Node>();
-  public override IEnumerable<Node> PreResolveChildren => Children;
+  public override IEnumerable<INode> Children => Var == null ? (Arguments ?? Enumerable.Empty<Node>()) : new[] { Var };
```
MikaelMayer, line 105: nice

Source/DafnyCore/AST/IHasUsages.cs
```diff
@@ -10,8 +11,33 @@ public interface IHasUsages : IDeclarationOrUsage {
...
+      type = $"<Internal error: {e.Message}>";
+    }
+    return $"{ghost}{variable.Name}: {type}";
+  }
+}
+
 public interface ISymbol : IDeclarationOrUsage {
   DafnySymbolKind Kind { get; }
+
+  string GetHoverText(DafnyOptions options);
```
MikaelMayer, line 40: Since this is in the core file and not in the LSP, would it make sense to rename it GetSymbolInfo? There is no hover in the command line.

Source/DafnyCore/AST/Program.cs
```diff
@@ -71,7 +71,7 @@ public class Program : TokenNode {
   /// Get the first token that is in the same file as the DefaultModule.RootToken.FileName
   /// (skips included tokens)
   public IToken GetFirstTopLevelToken() {
-    var rootToken = DefaultModuleDef.RangeToken.StartToken;
+    var rootToken = DefaultModuleDef.StartToken;
```
MikaelMayer, line 74: Nice simplification.

Source/DafnyCore/AST/TypeDeclarations/DatatypeCtor.cs
```diff
@@ -50,4 +50,8 @@ public DatatypeCtor(RangeToken rangeToken, Name name, bool isGhost, [Captured] L
   }
 
   public DafnySymbolKind Kind => DafnySymbolKind.EnumMember;
+  public string GetHoverText(DafnyOptions options) {
+    var formals = string.Join(", ", Formals.Select(f => f.AsText()));
```
MikaelMayer, line 54: I saw this code snippet several times in this PR. Would it make sense to refactor it somewhere?

Source/DafnyCore/Generic/Node.cs
```diff
@@ -10,9 +10,15 @@
...
+  public IToken StartToken => RangeToken.StartToken;
+  public IToken EndToken => RangeToken.EndToken;
+  IEnumerable<IToken> OwnedTokens { get; }
   RangeToken RangeToken { get; }
   IToken Tok { get; }
-  IEnumerable<Node> Children { get; }
+  IEnumerable<INode> Children { get; }
+  IEnumerable<INode> PreResolveChildren { get; }
+
+  IEnumerable<INode> GetConcreteChildren();
```
MikaelMayer, line 21: What's the difference between this and Children and PreResolveChildren? A comment would be appreciated.

Source/DafnyLanguageServer.Test/Lookup/HoverTest.cs
```diff
@@ -597,7 +602,7 @@ class C {
   C.m(); // TODO
  //  ^[Unformatted comment] // Does not work yet.
   var c: C := new C();
-//                ^[The class C. Should be used like this:\n```dafny\nnew C();\n```]
+//                ^[This is the constructor\n```dafny\nconstructor C()\n```]
```
MikaelMayer, line 605: Nice. Clicking on the symbol now redirect to the constructor instead of the class, I presume?

MikaelMayer (changes requested):
This is great.

Source/DafnyLanguageServer.Test/Lookup/HoverTest.cs
keyboardDrummer: I think it already did that

Source/DafnyCore/AST/TypeDeclarations/DatatypeCtor.cs
keyboardDrummer: It occurs twice, in DatatypeCtor and Function, but the only part that's exactly the same is `string.Join(", ", Formals.Select(f => f.AsText()))`. If there was multiple inheritance I might introduce something like `HasParameters` and include this method and the `Formals` field there, but given that there's not I'm more inclined to leave this like it is. Note there's also `string.Join(", ", Ins.Select(i => i.AsText()))` in `Method`.

I could introduce `interface IHasInputs { List<Formal> Inputs {get;} }` and have Function Method and DatatypeCtor inherit from it, and try to find all the existing duplications between these three, but that'd be better to do in a follow-up.

keyboardDrummer:
> This is great.

Thanks, you're sweet :-)

MikaelMayer (dismissed):
It will make it much easier to traverse now that we merged the symbol tree with the Node tree. Thanks for this refactoring!

MikaelMayer (approved):

Issue #3460, opened 2023-02-04
Implement a --quiet option


davidcok:
### Summary

Permit less noisy output

### Background and Motivation

Less noisy output is sometimes useful -- more robust tests; easier to see errors, easier in scripts.

### Proposed Feature

--quiet:0 -- the default, current behavior
--quiet:1 -- only errors and warnings, no informational output. Implies --verbose:0
--quiet:2 -- absolutely silent (no error, warning messages). only an exit code to indicate the results

PR #4356, opened 2023-07-29
fix: Continue to compilation tests after comparing .verifier.expect

### LINKED ISSUE #4355: .verifier.expect file cause compilation tests to be ignored
### Dafny version

4.2.0

```dafny
===== TestIt.dfy

method Main() {
  ghost var n := 15;
  if j :| 0 <= j < n { // this gives a no-trigger warning
  }
  print "hello\n";
}
===== TestIt.dfy.verifier.expect
TestIt.dfy(5,5): Warning: /!\ No terms found to trigger on.
===== TestIt.dfy.expect
good bye
```

```code
DAFNY/Test/dafny0> lit TestIt.dfy
...
-- Testing: 1 tests, 1 workers --
PASS: Dafny :: dafny0/TestIt.dfy (1 of 1)

Testing Time: 2.07s
  Passed: 1
```

### What happened?

When a test in the Dafny test suite has a `.verifier.expect` file that consists of only warning messages, then `TestDafny for-each-compiler` never calls the compiler.

As demonstrated by the repro above, the compiled program prints `hello` and the `.expect` file expects `good bye`. Yet, the test passes.

Mac

RustanLeino:
This PR fixes and changes the behavior of `%testDafnyForEachCompiler` tests. The bug that is fixed had caused 16 compiler tests in the Dafny test suite to not run.

Here is a description of the new behavior, where _italicized_ parts indicate changes from before:

* Verification is expected to go through with a 0 exit code.
* If there is a `.verifier.expect` file, then the actual output is expected to contain a blank line followed by a line that starts with `Dafny program verifier`, and this portion of the output will be removed before comparing the output. The remaining output is compared with the contents of the `.verifier.expect` file and is expected to match.
* If there is no `.verifier.expect` file, the verifier output minus the `Dafny program verifier` line (see previous bullet) is expected to be empty.
* When running the test for each compiler, if the output contains a line starting with `Dafny program verifier`, then _everything up to and include that line is removed_ before comparing the output.

This PR adds add test in `Test/metatests`.

Fixes #4355


robin-aws (changes requested):
Oof, glad you caught this, thanks for fixing.

> we're not set up to test the behavior of TestDafny itself.

Au contraire! Check out https://github.com/dafny-lang/dafny/tree/master/Test/metatests (which is where I should have added more tests as I added features to this utility in the first place)

Source/TestDafny/MultiBackendTest.cs
```diff
@@ -185,6 +181,10 @@ public class MultiBackendTest {
     }.Concat(options.OtherArgs);
 
     var (exitCode, outputString, error) = RunDafny(options.DafnyCliPath, dafnyArgs);
+    var compilationOutputPrior = new Regex("\r?\nDafny program verifier[^\r\n]*\r?\n").Match(outputString);
+    if (compilationOutputPrior.Success) {
+      outputString = outputString.Remove(0, compilationOutputPrior.Index + compilationOutputPrior.Length);
+    }
```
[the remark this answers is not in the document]
RustanLeino: This removes the "Dafny program verifier..." line from anywhere in the actual output, rather that prepending this line to the expected output (which had been done elsewhere in the code).

Source/TestDafny/MultiBackendTest.cs
```diff
@@ -94,33 +94,32 @@ public class MultiBackendTest {
...
-        return 0;
+      if (diffMessage != null) {
+        output.WriteLine(diffMessage);
+        return 1;
       }
-
-      output.WriteLine(diffMessage);
-      return 1;
+    }
+    if (exitCode != 0) {
```
robin-aws, line 111: Just to be clear, this change only means that if the output and exit code don't match, we'll report the former instead of the latter, correct? That's just fine with me if it happens to be useful for you.

Source/TestDafny/MultiBackendTest.cs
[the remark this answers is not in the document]
robin-aws: Aha I see from the metatest change now how this is different, makes sense.

This does make me think we should really have an option to suppress that output rather than hacking around it here (i.e. https://github.com/dafny-lang/dafny/issues/3460), but that can be done later.

Test/metatests/TestBeyondVerifierExpect.dfy.testdafny.expect
robin-aws, line 1: ❤️

robin-aws (dismissed):

Source/TestDafny/MultiBackendTest.cs
RustanLeino: I forgot to implement and test that the output is empty if there is no `.verifier.expect` file. Now done.

robin-aws (approved):

PR #4359, opened 2023-07-31
Chore: Make translation diagnostics available in IDE no matter what

MikaelMayer:
There are currently only two types translation diagnostics,
- "Some instances of this call are not inlined.", which can happen only if there is a trigger in the file, which generates an info diagnostic during resolution.
- "Fuel can only increase within a given scope." which can be triggered only if there is already a function with fuel, which thus has a "decreases" info diagnostic during resolution.

The fact that they make it to the IDE is almost exceptional because of this remark.
Indeed, In the CompilationManager.cs, normally, after verification is prepared, only _previous_ resolution diagnostics are sent back to the IDE under  the type `IReadOnlyDictionary<Uri, List<Diagnostic>>`
It's just that resolution diagnostics are are using the same list of diagnostics as the resolution ones, so the translation will add information to them, and that list will be updated _if it existed before_. If not, it will get created, but the resolution diagnostics won't have it since it's a snapshot of a dictionary.

This PR fixes that behavior, by ensuring the diagnostics sent after translation are the most up to date ones.
I don't think it's possible to write a reasonable test of this feature at this stage, because there are no other translation diagnostics that could be triggered without a previous resolution information diagnostic. But that will change later. This PR only prepares the way with a simple change.


keyboardDrummer:
I'm currently removing translation to Boogie as a phase that's always done by the IDE. Could you move these translation diagnostics to a post-resolve step?

MikaelMayer:
> I'm currently removing translation to Boogie as a phase that's always done by the IDE. Instead it will be done on-demand depending on what you're trying to verify. Could you move the two translation diagnostics you mentioned to a post-resolve step?

It would not be very practical, as these errors impact verification but not resolution or compilation. For example, the first one is an info, not an error, and the second one is an error only because of the encoding in Boogie. It should not be part of the resolution pipeline, especially if we change the verification back-end as it is being suggested these days.

.What you describe is fine, we don't need to emit these errors if verification/translation is not triggered anyway. I like the idea of translating only the things you need to verify what you want.
What can become true in the future is that translation itself could have more checks, including quick verification if possible to avoid delegating verification to Z3 when unnecessary. And currently, these checks would be visible only if there is already something that was emitted during resolution about that file, which it should not depend on.

keyboardDrummer:
> It would not be very practical, as these errors impact verification but not resolution or compilation. For example, the first one is an info, not an error, and the second one is an error only because of the encoding in Boogie. It should not be part of the resolution pipeline, especially if we change the verification back-end as it is being suggested these days.

> What you describe is fine, we don't need to emit these errors if verification/translation is not triggered anyway. I like the idea of translating only the things you need/want to verify. What can become true in the future is that translation itself could have more checks, including quick verification if possible to avoid delegating verification to Z3 when unnecessary. And currently, these checks would be visible only if there is already something that was emitted during resolution about that file, which it should not depend on.

Okay, I can go with that. 

> I don't think it's possible to write a reasonable test of this feature at this stage, because there are no other translation diagnostics that could be triggered without a previous resolution information diagnostic.

I don't understand. Can't we add a test that checks for one of the two translation diagnostics you mentioned?

keyboardDrummer:
> > I don't understand. Can't we add a test that checks for one of the two translation diagnostics you mentioned?
> 
> We can, but that test would already pass as of today without this PR as I explained earlier, so that's why I say we can't test this PR in particular. I'm not particularly inclined to add a test that was not failing before to a PR pretending that the PR solves it when it does not. But if you insist, I'll add it

I think we need a test for the fix.

What triggered you to make this PR, given that the fix currently has no visible affect? 

Is there a plugin can make the bug visible? Maybe add a test that uses this plugin?

Source/DafnyCore/Plugins/Rewriter.cs
```diff
@@ -31,5 +31,15 @@ public abstract class Rewriter : ErrorReportingBase {
     public virtual void PostResolve(Program program) {
       Contract.Requires(program != null);
     }
+
+    /// <summary>
+    /// Override this method to obtain the final program before the translation pipeline occurs the entire resolution pipeline
```
keyboardDrummer, line 36: Can you fix the grammar in this comment?

Source/DafnyCore/Plugins/Rewriter.cs
```diff
@@ -31,5 +31,15 @@ public abstract class Rewriter : ErrorReportingBase {
...
+    /// after the individual PostResolve on every module
+    /// You can also report warnings and errors using reporter.Error
+    /// </summary>
+    /// <param name="program">The entire program before it is translated</param>
```
keyboardDrummer, line 40: "entire" can be left out

keyboardDrummer (approved):

Source/DafnyCore/Plugins/Rewriter.cs
MikaelMayer, line 36:
```suggestion
    /// Override this method to obtain the resolved program before the translation pipeline occurs
```

Source/DafnyCore/Plugins/Rewriter.cs
MikaelMayer:
```suggestion
    /// <param name="program">The program before it is translated</param>
```

keyboardDrummer (approved):

PR #4368, opened 2023-08-01
Support character literals and basic built-in collection types

shadaj:
Support character literals and basic built-in collection types


Test/c++/returns.dfy
```diff
@@ -11,5 +11,5 @@ method multi() returns (x:uint64, y:uint64)
 method Main()
 {
   var x, y := multi();
-  print x, y, '\n';
+  print x, y, "\n";
```
shadaj, line 14: Printing of characters is pretty inconsistent; this test expects to see an actual newline printed but in other situations characters are printed with their surrounding quotes. So this just changes the test to use a string which has more consistent semantics.

jtristan (approved):

PR #4370, opened 2023-08-02
fix: Return `BigInteger` when computing the `multiset` cardinality in C#

### LINKED ISSUE #4125: OverflowException in C# for large multiset value cardinality
### Dafny version

4.1.0

```dafny
method Main() {
    var x: multiset<int> := multiset{};
    x := x[1 := 18446744073709551616];
    print |x|, "\n";
}
```

```code
% dafny run -t:cs src/Scratch.dfy

Unhandled exception. System.OverflowException: Value was either too large or too small for an Int32.
   at System.Numerics.BigInteger.op_Explicit(BigInteger value)
   at Dafny.MultiSet`1.get_Count()
   at _module.__default._Main(ISequence`1 __noArgsParameter)
   at __CallToMain.<>c__DisplayClass0_0.<Main>b__0()
   at Dafny.Helpers.WithHaltHandling(Action action)
   at __CallToMain.Main(String[] args)
```

### What happened?

C# backend doesn't handle multiset value cardinality outside of the Int32 range.

Mac

fabiomadge:
Fixes #4125. We could also apply this to the other collection types, but the benefit would only lie in the simpler code generation.

### DIFF
--- a/Source/DafnyRuntime/DafnyRuntime.cs
+++ b/Source/DafnyRuntime/DafnyRuntime.cs
@@ -374,6 +374,7 @@ namespace Dafny {
     bool IsEmpty { get; }
     int Count { get; }
     long LongCount { get; }
+    BigInteger ElementCount { get; }
     IEnumerable<T> Elements { get; }
     IEnumerable<T> UniqueElements { get; }
     bool Contains<G>(G t);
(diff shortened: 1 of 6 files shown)


Test/git-issues/git-issue-4012.dfy
```diff
@@ -1,7 +1,7 @@
-// RUN: %testDafnyForEachCompiler "%s" -- --relax-definite-assignment
```
keyboardDrummer, line 1: I'm confused because this PR says it fixes issue #4125, and in that issue the reproduction example given is:

```
method Main() {
    var x: multiset<int> := multiset{};
    x := x[1 := 18446744073709551616];
    print |x|, "\n";
}
```

But apparently that example was already in our test-suite. How come it was not triggering the bug?

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -652,18 +653,21 @@ public class MultiSet<T> : IMultiSet<T> {
...
     }
-    private BigInteger ElementCount() {
-      // This is inefficient
-      var c = occurrencesOfNull;
-      foreach (var item in dict) {
-        c += item.Value;
+
+    public BigInteger ElementCount {
+      get {
+        // This is inefficient
```
keyboardDrummer, line 664: If computing `ElementCount` is `inefficient`, shouldn't it be a method instead of a property?

Test/git-issues/git-issue-4012.dfy
fabiomadge: Robin checked in `Test/git-issues/git-issue-4012.dfy.cs.check`, which grants an exemption to the tester. At the time I only fixed it for Python.

Source/DafnyRuntime/DafnyRuntime.cs
fabiomadge: Maybe, but the other two properties just cast down the exact solution computed here, so there's a precedent.

keyboardDrummer (approved):

PR #4378, opened 2023-08-03
Let the IDE translate to and resolve Boogie only when attempting to verify

keyboardDrummer:
### Changes

- Let the IDE translate to and resolve Boogie only when attempting to verify, instead of doing those steps for all code in the project right after resolution, and before allowing any verification to be done.

  - Using the experimental verification caching feature to detect which tasks were already verified before the user triggers any verification, no longer works. This is because that caching relies on pre-verification which we now only do when the user triggers verification. However, when trying to verify these cached tasks, they'll be completed immediately after pre-verification finishes. In the future, we will have to implement caching of verification based on the Dafny AST if we want it to show up before pre-verification.
  - The Dafny AST nodes must now specify whether they can be verified, even before being translated to Boogie. This creates a slight duplication of effort, since this information was implicitly also provided by the Boogie translation
- Properly close files after they are read, both `.dfy` and `.dfyconfig.toml` files
- XUnit tests now all run sequentially. This does not seem to have a significant affect on total test runtime and makes debugging test timeouts easier because it excludes lack of available threads as a cause.
- The class `ConstantField` is moved to a separate file
- Add AutoGeneratedToken where they were missing.
- Configure the `dotnet test` call on the CLI so individual server tests won't run for more than 6 minutes.
- Split the existing `OnChange` mode into `OnChangeFile` and `OnChangeProject`, which let verification automatically run if either their containing file or containing project has changed, respectively.
- `Stringify`, a method to help with debugging, has been improved
- Small cleanup of `CounterExampleTest`, changing a `List<Action>` into an `Action`
- Many improvements to logging

### Testing
- Existing tests have been adapted for the new behavior
- The performance part improvement of this PR is not tested, since we don't have a good framework for testing performance.
- Add tests for the behavior of `OnChangeFile` and `OnChangeProject` mode.
- Add tests for `Stringify`
- Add tests for `FindNode`


MikaelMayer (changes requested):
I'm adding myself as a reviewer just to ensure we port the fix made in
https://github.com/dafny-lang/dafny/pull/4359
because that fix is not tested.
Happy to insert it once all the tests pass.

Source/DafnyCore/AST/Members/ConstantField.cs
```diff
@@ -0,0 +1,54 @@
...
+    set { throw new cce.UnreachableException(); }
+  }
+  public bool AllowsAllocation => true;
+
+  public override IEnumerable<INode> Children => base.Children.Concat(new[] { Rhs }.Where(x => x != null));
+  public override DafnySymbolKind Kind => DafnySymbolKind.Constant;
+
+  public override IEnumerable<INode> PreResolveChildren => Children;
+  public ModuleDefinition ContainingModule => EnclosingModule;
+  public bool ShouldVerify => true; // TODO fix.
```
MikaelMayer, line 53: TODO here? A constant with a RHS should verify.

Source/DafnyCore/AST/TypeDeclarations/DatatypeCtor.cs
```diff
@@ -54,4 +54,7 @@ public DatatypeCtor(RangeToken rangeToken, Name name, bool isGhost, [Captured] L
     var formals = string.Join(", ", Formals.Select(f => f.AsText()));
     return $"{EnclosingDatatype.Name}.{Name}({formals})";
   }
+
+  public ModuleDefinition ContainingModule => EnclosingDatatype.EnclosingModuleDefinition;
+  public bool ShouldVerify => Formals.Any(f => f.DefaultValue != null); // TODO fix
```
MikaelMayer, line 59: TODO?

Source/DafnyCore/AST/TypeDeclarations/NewtypeDecl.cs
```diff
@@ -110,6 +110,9 @@ public class NewtypeDecl : TopLevelDeclWithMembers, RevealableTypeDecl, Redirect
 
     return GetTriviaContainingDocstringFromStartTokenOrNull();
   }
+
+  public ModuleDefinition ContainingModule => EnclosingModuleDefinition;
+  public bool ShouldVerify => true; // TODO fix
```
MikaelMayer, line 115: TODO?

Source/DafnyCore/AST/TypeDeclarations/SubsetTypeDecl.cs
```diff
@@ -33,12 +33,18 @@ public enum WKind { CompiledZero, Compiled, Ghost, OptOut, Special }
...
-  Expression RedirectingTypeDecl.Witness { get { return Witness; } }
+  BoundVar RedirectingTypeDecl.Var => Var;
+  Expression RedirectingTypeDecl.Constraint => Constraint;
+  WKind RedirectingTypeDecl.WitnessKind => WitnessKind;
+  Expression RedirectingTypeDecl.Witness => Witness;
 
   public override List<Type> ParentTypes(List<Type> typeArgs) {
     return new List<Type> { RhsWithArgument(typeArgs) };
   }
+  public bool ShouldVerify => true; // TODO fix.
```
MikaelMayer, line 44: TODO?

Source/DafnyCore/Rewriters/PluginRewriter.cs
```diff
@@ -15,7 +15,7 @@ class PluginRewriter : IRewriter {
     internalRewriter.PostResolve(program);
   }
 
-  internal override void PreTranslate(Program program) {
-    internalRewriter.PreTranslate(program);
+  internal override void PreTranslate(ModuleDefinition module) {
+    internalRewriter.PreTranslate(module);
```
MikaelMayer, line 19:
```suggestion
  internal override void PreVerify(ModuleDefinition module) {
    internalRewriter.PreVerify(module);
```

Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs
```diff
@@ -443,15 +455,18 @@ iterator ThatIterator(x: int) yields (y: int, z: int)
...
-    Assert.Equal(new Range(12, 8, 12, 17), status.NamedVerifiables[2].NameRange);
-    Assert.Equal(new Range(15, 9, 15, 30), status.NamedVerifiables[3].NameRange);
-    Assert.Equal(new Range(20, 11, 20, 34), status.NamedVerifiables[4].NameRange);
-    Assert.Equal(new Range(27, 5, 27, 10), status.NamedVerifiables[5].NameRange);
-    Assert.Equal(new Range(30, 8, 30, 16), status.NamedVerifiables[6].NameRange);
-    Assert.Equal(new Range(33, 9, 33, 21), status.NamedVerifiables[7].NameRange);
+    Assert.Equal(9, status.NamedVerifiables.Count);
+    var index = 0;
+    Assert.Equal(new Range(1, 17, 1, 23), status.NamedVerifiables[index++].NameRange);
+    // TODO this entry doesn't actually have anything to verify, but it's hard to determine
```
MikaelMayer, line 461: TODO here?

Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs
```diff
@@ -443,15 +455,18 @@ iterator ThatIterator(x: int) yields (y: int, z: int)
...
+    Assert.Equal(new Range(4, 9, 4, 30), status.NamedVerifiables[index++].NameRange);
+    Assert.Equal(new Range(7, 11, 7, 34), status.NamedVerifiables[index++].NameRange);
+    Assert.Equal(new Range(12, 8, 12, 17), status.NamedVerifiables[index++].NameRange);
+    Assert.Equal(new Range(15, 9, 15, 30), status.NamedVerifiables[index++].NameRange);
+    Assert.Equal(new Range(20, 11, 20, 34), status.NamedVerifiables[index++].NameRange);
+    Assert.Equal(new Range(27, 5, 27, 10), status.NamedVerifiables[index++].NameRange);
+    Assert.Equal(new Range(30, 8, 30, 16), status.NamedVerifiables[index++].NameRange);
+    Assert.Equal(new Range(33, 9, 33, 21), status.NamedVerifiables[index].NameRange);
```
MikaelMayer, line 469: Can be done in a follow-up PR, but I would love if you did not explicit the ranges there but obtain them from the test case instead. It would make the code easier to maintain.

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
```diff
@@ -155,11 +157,14 @@ protected ClientBasedLanguageServerTest(ITestOutputHelper output, LogLevel dafny
     await client.WaitForNotificationCompletionAsync(documentItem.Uri, cancellationToken);
     var compilation = (await Projects.GetLastDocumentAsync(documentItem))!;
     Assert.NotNull(compilation);
-    var expectedDiagnostics = compilation.GetDiagnostics(documentItem.Uri.ToUri()).Select(d => d.ToLspDiagnostic()).ToList();
+    var expectedDiagnostics = compilation.GetDiagnostics(documentItem.Uri.ToUri()).
+      Select(d => d.ToLspDiagnostic()).
+      OrderBy(d => d.Range.Start).ToList();
```
MikaelMayer, line 162: Minor style preference, you do what you want here, but I prefer the dot to be put before the method name rather than after the object, when inserting newlines. I know it makes it looks like sentences, but that would be true only if the last call finished with a dot.

Source/DafnyLanguageServer.Test/Util/StringifyTest.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.Dafny.LanguageServer.IntegrationTest.Util; 
+
+public class StringifyTest {
+
+  record Inner(int Age, string Name);
+
+  record Outer(Inner[] InnersArray, IDictionary<string, Inner> MoreInners);
+
+  [Fact]
+  public void StringifyThing() {
```
MikaelMayer, line 14: Interesting helper. You're using it for logging, I presume?

Source/DafnyLanguageServer.Test/GutterStatus/ConcurrentLinearVerificationGutterStatusTester.cs
```diff
@@ -44,14 +44,14 @@ public class ConcurrentLinearVerificationGutterStatusTester : LinearVerification
...
- .  S [=][=][-][~][=]:{//Replace: { assert false;
- .  S [S][ ][I][S][ ]:}", false, $"EnsuresManyDocumentsCanBeVerifiedAtOnce{i}.dfy", true, true, verificationStatusGutterReceivers[i]));
+ .  |  |  |  |  I  |  |  | :predicate F(i: int) {
+ .  |  |  |  |  I  |  |  | :  false // Should not be highlighted in gutter.
+ .  |  |  |  |  I  |  |  | :}
+    |  |  |  |  I  |  |  | :
+ .  .  S [S][ ][I][I][S][ ]:method H()
+ .  .  S [=][=][-][-][~][O]:  ensures F(1)
+ .  .  S [=][=][-][-][~][=]:{//Replace: { assert false;
+ .  .  S [S][ ][I][I][S][ ]:}", false, $"EnsuresManyDocumentsCanBeVerifiedAtOnce{i}.dfy", true, true, verificationStatusGutterReceivers[i]));
```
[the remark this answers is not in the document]
keyboardDrummer: When the server detects that it's testing gutter icons, it will translate each module one by one, and then schedule tasks to be verified one by one.

Outside of this testing mode, the server will translate and schedule tasks one by one, so a task can begin verification while another is still translating. This mode of operation, while correct, makes it impossible to use the current testing format since that format assumes too much determinism. I didn't want to change the tests too much so I introduced a special mode on the server that allows the current tests to pass.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -20,42 +20,51 @@ namespace Microsoft.Dafny.LanguageServer.Language {
...
+      }
+
       await mutex.WaitAsync(cancellationToken);
       try {
 
         var program = compilation.Program;
         var errorReporter = (DiagnosticErrorReporter)program.Reporter;
 
-        cancellationToken.ThrowIfCancellationRequested();
+        cancellationToken.ThrowIfCancellationRequested(); // TODO remove ?
```
MikaelMayer: TODO? If you want to leave it, please put more explanations on what to do there.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
```diff
@@ -139,8 +138,6 @@ CancellationToken cancellationToken
         var newRelatedInformation = diagnostic.RelatedInformation?.SelectMany(related =>
           MigrateRelatedInformation(change, related)).ToList();
         yield return diagnostic with {
-          Message = diagnostic.Message.StartsWith(OutdatedPrefix) ? diagnostic.Message : OutdatedPrefix + diagnostic.Message,
```
MikaelMayer, line 142: I haven't tested it, but is it actually removing the "outdated" messages of verification? I think i liked them a lot.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -60,73 +60,108 @@ ILogger<ITextDocumentLoader> logger
...
+        new()
+      );
     }
 
     private IdeState CreateDocumentWithEmptySymbolTable(Compilation compilation,
       IReadOnlyDictionary<Uri, IReadOnlyList<Diagnostic>> resolutionDiagnostics) {
       var dafnyOptions = DafnyOptions.Default;
       var program = new EmptyNode();
       return new IdeState(
+        compilation.Version,
```
MikaelMayer, line 153: Isn't that what was sufficient to ensure that gutter icons are handled in order?

Source/DafnyLanguageServer/Workspace/IdeStateObserver.cs
```diff
@@ -80,6 +85,50 @@ public class IdeStateObserver : IObserver<IdeState> {
       notificationPublisher.PublishNotifications(LastPublishedState, snapshot).Wait();
 #pragma warning restore VSTHRD002
       LastPublishedState = snapshot;
+      logger.LogDebug($"Updated LastPublishedState to version {snapshot.Version}, uri {compilation.Uri}");
     }
   }
+
+  public void Migrate(DidChangeTextDocumentParams documentChange, int version) {
```
MikaelMayer, line 92: It's interesting that you migrated the migration code in the IdeStateObserver. Could you compare the versions to ensure they are increasing? Is it necessary? I remember it was in other places like for gutter icons.

Source/DafnyLanguageServer/Workspace/CompilationManager.cs
```diff
@@ -86,208 +87,219 @@ public class CompilationManager {
...
+        tasksForModule.GetValueOrDefault(verifiable.NameToken.GetFilePosition()) ?? new List<IImplementationTask>(0);
 
-    verificationProgressReporter.RecomputeVerificationTrees(translated);
+      verificationProgressReporter.ReportImplementationsBeforeVerification(compilation,
+        verifiable, tasksForVerifiable.Select(t => t.Implementation).ToArray());
 
-    if (ReportGutterStatus) {
-      foreach (var uri in translated.RootUris) {
-        verificationProgressReporter.ReportRealtimeDiagnostics(translated, uri, true);
+      foreach (var uri in compilation.RootUris) { // TODO scope to one uri ?
```
MikaelMayer, line 203: TODO: Remove or detail your explanation, or create an issue.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
```diff
@@ -60,73 +59,108 @@ ILogger<ITextDocumentLoader> logger
       return CreateDocumentWithEmptySymbolTable(compilation, ImmutableDictionary<Uri, IReadOnlyList<Diagnostic>>.Empty);
     }
 
-    public async Task<CompilationAfterParsing> LoadAsync(DafnyOptions options, Compilation compilation,
-      CancellationToken cancellationToken) {
+    public async Task<CompilationAfterParsing> ParseAsync(DafnyOptions options, Compilation compilation,
```
MikaelMayer, line 62: Finally a method name that matches what it's actually doing :-)
Also, well done splitting parsing and resolving.

MikaelMayer (commented):
Just finished my review. I did not test this PR myself yet, so I'm relying on the CI tests, although I did not find any obvious or potential issues in the code itself.

Source/DafnyLanguageServer.Test/GutterStatus/ConcurrentLinearVerificationGutterStatusTester.cs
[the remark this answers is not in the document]
MikaelMayer: Ok thanks for the explanation, it makes sense. Glad you introduced a more deterministic testing mode.

Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs
keyboardDrummer: I would rather leave that for when this ranges are updated.

Source/DafnyLanguageServer/Workspace/ChangeProcessors/Relocator.cs
keyboardDrummer: The code is just moved.

Source/DafnyLanguageServer/Workspace/IdeStateObserver.cs
keyboardDrummer: The version number checks ensure that events from an older version don't publish information after events from a newer version.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
keyboardDrummer: Can you elaborate?

Source/DafnyLanguageServer.Test/Util/ClientBasedLanguageServerTest.cs
keyboardDrummer: I don't have a strong opinion on this, but I would rather discuss this outside of this PR and then try to have a formatting rule for.

Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs
MikaelMayer: It looks like before the Ide state did not have this field and I remember you mentioning to me that because versioning had been lost, gutter icons were sometimes published in the wrong order. I was wondering if this fixes that.

MikaelMayer (commented):
One minor typo and I think it's good to go!

keyboardDrummer:
> It looks like before the Ide state did not have this field and I remember you mentioning to me that because versioning had been lost, gutter icons were sometimes published in the wrong order. I was wondering if this fixes that.

It previously sort of had this field, through its other field `Compilation` which has a version. This field is only needed because `IdeState` can now be migrated while keeping the same compilation, which is used in `IdeObserverState.Migrate`

MikaelMayer (dismissed):
Thanks for this excellent work.

MikaelMayer (approved):

PR #4381, opened 2023-08-03
Fix z3 download

atomb:
Fixes the download URLs for Z3 in a couple of places.


fabiomadge (approved):
Do we still package the same z3 binaries we download with the Makefile?

atomb:
> Do we still package the same z3 binaries we download with the Makefile?

If I understand you correctly, yes. The binaries downloaded by the Makefile and the GitHub workflows are all binaries we build and package. And we still build and package 4.8.5, though we could probably stop doing that soon.

fabiomadge:
Yes, I just had a look at `package.py` myself. It contains code that implements the fallback to x86.

PR #4385, opened 2023-08-04
[Test Generation] Deprecate support for old command line interface

Dargones:
This PR deprecates support for old command line interface in test generation. Test generation hardcodes some non-standard command-line-options ("predicates" type encoding, enforced determinism, disabled axiom pruning), which is much more easier to enforce via the new command line interface. This also removes quite a bit of code and makes test generation easier to maintain.

### DIFF
--- a/Source/DafnyTestGeneration/GenerateTestsCommand.cs
+++ b/Source/DafnyTestGeneration/GenerateTestsCommand.cs
@@ -4,6 +4,8 @@ using System.CommandLine;
 using System.CommandLine.Invocation;
 using System.Linq;
 using DafnyCore;
+using Microsoft.Boogie;
+
 // Copyright by the contributors to the Dafny Project
 // SPDX-License-Identifier: MIT
 
@@ -23,7 +25,7 @@ public class GenerateTestsCommand : ICommandSpec {
       BoogieOptionBag.SolverResourceLimit,
       BoogieOptionBag.VerificationTimeLimit,
       PrintBpl,
-      DisablePrune
+      ForcePrune
     }.Concat(ICommandSpec.ConsoleOutputOptions).
       Concat(ICommandSpec.ResolverOptions);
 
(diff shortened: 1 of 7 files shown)


Source/DafnyTestGeneration/GenerateTestsCommand.cs
```diff
@@ -57,23 +59,25 @@ private enum Mode {
...
+
+  internal static void PostProcess(DafnyOptions dafnyOptions, Options options, InvocationContext context, TestGenerationOptions.Modes mode) {
     dafnyOptions.CompilerName = "cs";
     dafnyOptions.Compile = true;
     dafnyOptions.RunAfterCompile = false;
     dafnyOptions.ForceCompile = false;
     dafnyOptions.DeprecationNoise = 0;
     dafnyOptions.ForbidNondeterminism = true;
     dafnyOptions.DefiniteAssignmentLevel = 2;
+    dafnyOptions.TypeEncodingMethod = CoreOptions.TypeEncoding.Predicates;
```
atomb, line 78: I hope it's possible eventually to use the same encoding as verification does, but this seems fine for now.

Source/DafnyTestGeneration/GenerateTestsCommand.cs
```diff
@@ -86,8 +90,8 @@ private enum Mode {
     "Print the Boogie code used during test generation.") {
     ArgumentHelpName = "filename"
   };
-  public static readonly Option<bool> DisablePrune = new("--no-prune",
-    "Disable axiom pruning that Dafny uses to speed up verification.") {
+  public static readonly Option<bool> ForcePrune = new("--force-prune",
+    "Enable axiom pruning that Dafny uses to speed up verification. This may negatively affect the quality of tests.") {
```
atomb, line 94: Do you have a summary of why pruning negatively affects tests?

atomb (approved):
It's nice to see so much code removed!

Source/DafnyTestGeneration/GenerateTestsCommand.cs
Dargones: It used to be the case that Boogie would not print the `{:uses}` attributes when saving an AST to file causing pruning to be way to aggressive if you then run Boogie on the same file. But I just discovered a [recent PR to Boogie](https://github.com/boogie-org/boogie/pull/765) that might have fixed it. I will look into it and remove the ForcePrune flag if that is the case once the newer Boogie is merged.

Source/DafnyTestGeneration/GenerateTestsCommand.cs
Dargones: Yes, I will hope to revert this change. The one test failing on Windows with the arguments encoding is still a mystery to me as I wasn't able to reproduce it with Boogie alone... But I will try to dig deeper into that.

Source/DafnyTestGeneration/GenerateTestsCommand.cs
atomb: Ah, I'm glad that fix might help!

PR #4388, opened 2023-08-04
Stop sharing verification trees between compilations

keyboardDrummer:
Hopefully fixes non-deterministic failures in verification tree based tests.

What we're seeing is that events triggered by update X also end up affecting notifications belonging to update X+1, so this change is a good candidate for fixing those issues.


MikaelMayer (approved):
Thats does the same but it's more reliable, thanks

PR #4395, opened 2023-08-06
Improve test stability

keyboardDrummer:
### Changes
- Various changes to help with debugging test stability
- A fix to prevent outdated diagnostics from being sent, which could happen in a race condition. This makes the test `OnDiskProducerVerificationErrors` stable.
- There's a bug that allows gutter icon notifications for document version X to be sent after those for version X+1 have been sent. This PR contains a change to testing code so tests will wait for all X gutter icon notifications to have been sent before applying update X+1, which works around the bug, allowing gutter icon tests to run stably.
- An improvement to code related to gutter icons that makes bugs in it less likely

### Caveat
- `IncrementalVerificationDiagnosticsBetweenMethods` still fails randomly
- One problem that this PR doesn't fix is tests on CI run at a higher concurrency level than we want, but I haven't figured out how to change that. SO post about it: https://stackoverflow.com/questions/76853188/how-do-i-configure-the-concurrency-level-of-xunit-through-the-the-dotnet-test

### Testing
No additional testing


Source/DafnyLanguageServer.Test/GutterStatus/CachedLinearVerificationGutterStatusTester.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
```
MikaelMayer, line 2:
```suggestion
```

Source/DafnyLanguageServer.Test/GutterStatus/SimpleLinearVerificationGutterStatusTester.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System.Threading.Tasks;
 using Microsoft.Dafny.LanguageServer.IntegrationTest.Extensions;
+using Microsoft.Extensions.Logging;
```
MikaelMayer, line 3:
```suggestion
```

Source/DafnyLanguageServer/Workspace/IdeStateObserver.cs
```diff
@@ -72,7 +74,11 @@ public class IdeStateObserver : IObserver<IdeState> {
         return;
       }
 
-      notificationPublisher.PublishNotifications(LastPublishedState, snapshot);
+      // To prevent older updates from being sent after newer ones, we can only run one PublishNotifications at a time.
+      // So we wait for it here to finish, and the lock in this method prevents more than one from running at a time.
+#pragma warning disable VSTHRD002
+      notificationPublisher.PublishNotifications(LastPublishedState, snapshot).Wait();
```
MikaelMayer, line 80: Nice

MikaelMayer (approved):
Two non-blocking comments, otherwise yes I think waiting is a good idea. I hope it does not block the thread though? Is it still async wait?

keyboardDrummer:
> Two non-blocking comments, otherwise yes I think waiting is a good idea. I hope it does not block the thread though? Is it still async wait?

There's one thread that handles Boogie events and publishes notifications, and that one is now doing a synchronous wait for the diagnostic publication code to check whether a particular Uri is owned by the current project, which requires a tiny bit of IO and influences which diagnostics get sent. I think it's not the prettiest but otherwise fine.

PR #4397, opened 2023-08-07
fix: Use correct equality for sequences in Java

### LINKED ISSUE #4141: Java Inconsistent equality between return values of methods
### Dafny version

4.1.0

```dafny
method m_method_12() returns (ret_1: seq<array<int>>)
	ensures (|ret_1| == 2 && ret_1[0].Length == 5 && (ret_1[0][0] == 25) && (ret_1[0][1] == 2) && (ret_1[0][2] == 18) && (ret_1[0][3] == 21) && (ret_1[0][4] == 12) && ret_1[1].Length == 4 && (ret_1[1][0] == 16) && (ret_1[1][1] == 9) && (ret_1[1][2] == 17) && (ret_1[1][3] == 8))
{
	var v_array_20: array<int> := new int[5] [25, 2, 18, 21, 12];
	var v_array_21: array<int> := new int[4] [16, 9, 17, 8];
	var v_seq_32: seq<array<int>> := [v_array_20, v_array_21];
	return (v_seq_32);
}

method Main() returns ()
{
	var v_seq_34: seq<array<int>> := m_method_12();
	var v_seq_36: seq<array<int>> := m_method_12();
	print v_seq_34 == v_seq_36;
}
```

```code
$ dafny /compileVerbose:0 /compile:4 /compileTarget:cs test.dfy

Dafny program verifier finished with 3 verified, 0 errors
false
$ dafny /compileVerbose:0 /compile:4 /compileTarget:java test.dfy

Dafny program verifier finished with 3 verified, 0 errors
true
```

### What happened?

In backends py, go, cs and js the output is false (i.e. the returned values from calling the same function is false), however in java the result is true.

I would expect java to have false here as each call to m_method_12 is distinct.

Linux

fabiomadge:
Fixes #4141.


RustanLeino (approved):
Good fix.

PR #4398, opened 2023-08-08
Add test automation for type-system refresh

RustanLeino:
As the new type system (`/typeSystemRefresh:1`) is coming into place, we need to have tests for it. Until we permanently switch over to the new type system, we'd like to run tests for both the old type system ("legacy") and the new one ("refresh").

To make it simple for tests to be run twice, once for the legacy resolver and once for the refresh resolver, this PR contains two new features of `TestDafny`:

* The `for-each-compiler` verb now accepts a `--refresh-exit-code` flag. If this flag is present, the "verify" part of the test will be run twice, once for each resolver. Then, each of the compilers will be run, using the legacy resolver.

    The "verify" part of the test is expected to have exit code 0 and to generate no output. However, there was already a way to indicate an exception: by supplying a `.verifier.expect` file, the output is compared against it instead of being compared with the empty output. This PR extends this functionality for the "verify" pass that uses the refresh resolver. In particular, the output is compared to (a) the contents of a `.refresh.expect` file, if present, or else (b) the contents of the `.verifier.expect` file, if present, or else (c) the empty output. In addition, the value of the given `--refresh-exit-code` flag is the expected exit code when the refresh resolver is used.

* There's now a new verb, `for-each-resolver` (invoked from `lit` by `%testDafnyForEachResolver`). It runs `dafny verify` twice, once with the legacy resolver and once with the refresh resolver.

    The output of the legacy resolver is given in the `.expect` file, which must exist. The exit code of this run is expected to be 0, but can be overridden using the flag `--expect-exit-code`.

    The output of the refresh resolver is given in a `.refresh.expect` file, if present, or else in the `.expect` file. The exit code of this run is expected to be the same as for the legacy resolver, but can be overridden using the flag `--refresh-exit-code`.

The PR contains some new tests in the `Test/metatests` directory. In addition, as a proof of concept, this PR converts 3 of the previous test files to using the new functionality. As the new type system is coming into place, we anticipate changing the majority of the test files to using this new functionality.


Source/TestDafny/MultiBackendTest.cs
```diff
@@ -153,13 +200,80 @@ public class MultiBackendTest {
...
+      $"--bprint:{tmpPrint}"
+    }.Concat(options.OtherArgs.Where(OptionAppliesToVerifyCommand)).ToArray();
+
+    output.WriteLine("Verifying...");
+
+    var resolutionOptions = new List<ResolutionSetting>() {
+      new ResolutionSetting(new string[] { }, new string[] { ".expect" },
+        options.ExpectExitCode != null ? (int)options.ExpectExitCode : 0),
+      new ResolutionSetting(new string[] { "--type-system-refresh" }, new string[] { ".refresh.expect", ".expect" },
+        options.RefreshExitCode != null ? (int)options.RefreshExitCode : options.ExpectExitCode != null ? (int)options.ExpectExitCode : 0)
```
alex-chew, line 236: nit: can be simplified to
```suggestion
      new ResolutionSetting(new string[] { }, new string[] { ".expect" },
        (int)(options.ExpectExitCode ?? 0)),
      new ResolutionSetting(new string[] { "--type-system-refresh" }, new string[] { ".refresh.expect", ".expect" },
        (int)(options.RefreshExitCode ?? options.ExpectExitCode ?? 0))
```

Source/TestDafny/MultiBackendTest.cs
```diff
@@ -153,13 +200,80 @@ public class MultiBackendTest {
...
+    };
+
+    foreach (var resolutionOption in resolutionOptions) {
```
alex-chew, line 239: nit: Can we log which resolver is being used during each run in order to aid debugging?

alex-chew (approved):
LGTM. I added a couple of non-blocking comments.

Source/TestDafny/MultiBackendTest.cs
RustanLeino: Excellent suggestion!

alex-chew (approved):
LGTM

PR #4399, opened 2023-08-08
Logic fix

jtristan:
Fixes #


jtristan:
Here's some explanation that should make its way into comments and ref manual.

WLOG, consider the following abstract pre/post conditions and body.
heap() stands for *, or, reading the entire heap.

predicate P<T>(x: T)
function R<T>(x: T): set<object>

function body<T>(x: T): T 

function heap(): set<object> 

For a function such as:

function f<T>(x: T): T 
  reads R(x)
  requires P(x)
  {
    body(x)
  }

To derive the axioms, we consider how the f.reads and f.requires functions would be **defined**
and capture that into axioms. 

function f_reads<T>(x: T): set<object> 
  reads R(x)
  requires P(x)
  {
    R(x)
  }

// Axiom 1
// REQUIRES(f_reads) == P == REQUIRES(f)

// Axiom 2
// Since READS(f_reads) == R == READS(f) and that reads clauses are checked under assumption of precondition
// P ==> READS(f_reads) == READS(f)
// That is,
// REQUIRES(f) ==> READS(f_reads) == READS(f)


predicate f_requires<T>(x: T)
  requires true 
  reads if P(x) then R(x) else heap()
  {
    P(x)
  }

// Axiom 3
// REQUIRES(f_requires) == True

// Axiom 4
// READS(f_requires) == if P then R else * == if REQUIRES(f) then READS(f) else *
// Which we turn into
// REQUIRES(f) ==> READS(f_requires) == READS(f)

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -5873,27 +5873,43 @@ void CheckCasePatternShape<VT>(CasePattern<VT> pat, Bpl.Expr rhs, IToken rhsTok,
...
 
           var fhandle = FunctionCall(f.tok, name, predef.HandleType, SnocSelf(SnocPrevH(args)));
           var lhs = FunctionCall(f.tok, Requires(arity), Bpl.Type.Bool, Concat(tyargs, Cons(h, Cons(fhandle, lhs_args))));
           Bpl.Expr rhs;
-          if (f.EnclosingClass is ArrowTypeDecl) {
-            // In case this is the /requires/ or /reads/ function, then there is no precondition
-            rhs = Bpl.Expr.True;
+          if (f.EnclosingClass is ArrowTypeDecl && f.Name == "requires") {
+            AddOtherDefinition(GetOrCreateFunction(f), (new Axiom(f.tok,
+                BplForall(Concat(vars, bvars), BplTrigger(lhs), Bpl.Expr.Eq(lhs, Bpl.Expr.True)))));
```
RustanLeino, line 5889:
```suggestion
            AddOtherDefinition(GetOrCreateFunction(f), new Axiom(f.tok,
                BplForall(Concat(vars, bvars), BplTrigger(lhs), Bpl.Expr.Eq(lhs, Bpl.Expr.True))));
```

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -5873,27 +5873,43 @@ void CheckCasePatternShape<VT>(CasePattern<VT> pat, Bpl.Expr rhs, IToken rhsTok,
...
+          } else if (f.EnclosingClass is ArrowTypeDecl && f.Name == "reads") {
+            var args_h = f.ReadsHeap ? Snoc(SnocPrevH(argsRequires), h) : argsRequires;
+            var pre = FunctionCall(f.tok, Requires(arity), Bpl.Type.Bool, Concat(SnocSelf(args_h), lhs_args));
+            AddOtherDefinition(GetOrCreateFunction(f), (new Axiom(f.tok,
+              BplForall(Concat(vars, bvars), BplTrigger(lhs), Bpl.Expr.Imp(pre, Bpl.Expr.Eq(lhs, Bpl.Expr.True))))));
           } else {
             var args_h = f.ReadsHeap ? Snoc(SnocPrevH(argsRequires), h) : argsRequires;
             rhs = FunctionCall(f.tok, RequiresName(f), Bpl.Type.Bool, Concat(SnocSelf(args_h), rhs_args));
+            AddOtherDefinition(GetOrCreateFunction(f), (new Axiom(f.tok,
+              BplForall(Concat(vars, bvars), BplTrigger(lhs), Bpl.Expr.Eq(lhs, rhs)))));
```
RustanLeino, line 5899:
```suggestion
            AddOtherDefinition(GetOrCreateFunction(f), new Axiom(f.tok,
              BplForall(Concat(vars, bvars), BplTrigger(lhs), Bpl.Expr.Eq(lhs, rhs))));
```

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -5873,27 +5873,43 @@ void CheckCasePatternShape<VT>(CasePattern<VT> pat, Bpl.Expr rhs, IToken rhsTok,
...
-            // In case this is the /requires/ or /reads/ function, then there is no precondition
-            rhs = Bpl.Expr.True;
+          if (f.EnclosingClass is ArrowTypeDecl && f.Name == "requires") {
+            AddOtherDefinition(GetOrCreateFunction(f), (new Axiom(f.tok,
+                BplForall(Concat(vars, bvars), BplTrigger(lhs), Bpl.Expr.Eq(lhs, Bpl.Expr.True)))));
+          } else if (f.EnclosingClass is ArrowTypeDecl && f.Name == "reads") {
+            var args_h = f.ReadsHeap ? Snoc(SnocPrevH(argsRequires), h) : argsRequires;
+            var pre = FunctionCall(f.tok, Requires(arity), Bpl.Type.Bool, Concat(SnocSelf(args_h), lhs_args));
+            AddOtherDefinition(GetOrCreateFunction(f), (new Axiom(f.tok,
+              BplForall(Concat(vars, bvars), BplTrigger(lhs), Bpl.Expr.Imp(pre, Bpl.Expr.Eq(lhs, Bpl.Expr.True))))));
```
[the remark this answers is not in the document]
jtristan: I think it is REQUIRES(f) ==> REQUIRES(f_reads) == true, this case is in the block that generates the f.x.requires axioms.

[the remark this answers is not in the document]
RustanLeino: Thanks. I read it wrong. So, then, this is _half_ of what the PR description calls AXIOM 1.

RustanLeino (changes requested):
I think half of AXIOM 1 is not encoded (which would be simple to fix).

While reviewing the PR, I wrote some additional test cases (see below -- since I have written them, I suggest you incorporate them into the test suite). The tests give errors on exactly those lines with "error", as expected, with four exceptions. Each exception is marked with `TODO`.

Two of those stem from that only half of AXIOM 1 is encoded.

I don't know why the TODO in `CallIfPossible2` does verifies, but I think it should NOT.

The line marked TODO in `Usability5` works as expected. But I'm wondering if that's what we want. (It probably is, but it may be surprising to users.)

``` dafny
// Tests for well-formedness of FunctionCallExpr

function MyFunction(x: int, S: set<object>): real
  requires x < 100
  reads S
{
  assert true;
  (x + 12) as real
}

// In methods, there are no reads checks:

method MCaller0(x: int, S: set<object>) {
  var myFunctionRequires := MyFunction.requires(x, S);
  var myFunctionReads := MyFunction.reads(x, S); // error: precondition (x < 100) may not hold
}

method MCaller1(S: set<object>) {
  var myFunctionRequires := MyFunction.requires(5, S);
  var myFunctionReads := MyFunction.reads(5, S);
}

// In functions, reads checks also matter:

// .requires

function FCallerRequires0(x: int, S: set<object>): real
  reads S
{
  var myFunctionRequires := MyFunction.requires(x, S); // error: insufficient reads clause to call MyFunction.requires, since it may return false
  3.14
}

function FCallerRequires1(S: set<object>): real {
  var myFunctionRequires := MyFunction.requires(5, S); // error: insufficient reads clause to call MyFunction.requires
  3.14
}

function FCallerRequires2(S: set<object>): real
  reads S
{
  var myFunctionRequires := MyFunction.requires(5, S);
  assert myFunctionRequires;
  3.14
}

function FCallerRequires3(S: set<object>): real
  requires MyFunction.requires(5, S) // error: insufficient reads clause
{
  var myFunctionRequires := MyFunction.requires(5, S);
  assert myFunctionRequires;
  3.14
}

function FCallerRequires4(S: set<object>): real
  requires MyFunction.requires(5, S)
  reads S
{
  var myFunctionRequires := MyFunction.requires(5, S);
  assert myFunctionRequires;
  3.14
}

// .reads

function FCallerReads0(x: int, S: set<object>): real {
  var myFunctionReads := MyFunction.reads(x, S); // error (x2): precondition failure and insufficient reads clause
  3.14
}

function FCallerReads1(S: set<object>): real {
  var myFunctionReads := MyFunction.reads(5, S); // error: insufficient reads clause
  3.14
}

function FCallerReads2(x: int, S: set<object>): real
  reads S
{
  var myFunctionReads := MyFunction.reads(x, S); // error: precondition failure (but under the assumption that precondition holds, reads clause is good)
  3.14
}

function FCallerReads3(S: set<object>): real
  reads S
{
  var myFunctionReads := MyFunction.reads(5, S);
  3.14
}

function FCallerReads4(S: set<object>): real
  requires MyFunction.requires(5, S)
  reads MyFunction.reads(5, S)
{
  var myFunctionRequires := MyFunction.requires(5, S);
  var myFunctionReads := MyFunction.reads(5, S);
  assert myFunctionRequires;
  assert myFunctionReads == S;
  3.14
}

// Usability tests

ghost function CallIfPossible0(x: int, S: set<object>): real {
  if MyFunction.requires(x, S) then // error: insufficient reads clause to call MyFunction.requires
    MyFunction(x, S)
  else
    2.7
}

ghost function CallIfPossible1(x: int, S: set<object>): real
  reads MyFunction.reads(x, S) // error: precondition may not hold
{
  if MyFunction.requires(x, S) then // under the assumption that the specification of CallIfPossible1 is well-formed, this line is fine
    MyFunction(x, S)
  else
    2.7
}

ghost function CallIfPossible2(x: int, S: set<object>): real
  reads S
{
  if MyFunction.requires(x, S) then // error: if MyFunction.requires(x, S) returns false, then this could read anything (TODO)
    MyFunction(x, S)
  else
    assert false; // error: control can indeed reach this point
    2.7
}

ghost function CallIfPossible3(x: int, S: set<object>): real
  requires x < 100
{
  if MyFunction.requires(x, S) then // error: insufficient reads clause
    MyFunction(x, S)
  else
    2.7
}

ghost function CallIfPossible4(x: int, S: set<object>): real
  requires x < 100
  reads S
{
  if MyFunction.requires(x, S) then // fine
    MyFunction(x, S)
  else
    assert false; // fine, since control never reaches this point
    2.7
}

// Tests for axioms that can be used to reason about ApplyExpr

method MApply0(x: int, S: set<object>) {
  var req := MyFunction.requires;
  assert req(5, S);
  assert req(x, S); // error: this assertion may not hold
}

method MApply1(x: int, S: set<object>) {
  var rds := MyFunction.reads;
  assert rds(5, S) == S;
  var s := rds(x, S); // error: precondition may not hold
}

method MApply2(x: int, S: set<object>) {
  var req := MyFunction.requires;
  var rds := MyFunction.reads;
  assert req(x, S) ==> rds(x, S) == S;
}

method MApply3(x: int, S: set<object>) {
  var req := MyFunction.requires;
  assert req(x, S) <==> x < 100;
  var rds := MyFunction.reads;
  assert req(x, S) ==> rds(x, S) == S;
}

// Usability tests

method Usability0(x: int, f: int ~> real, g: int --> real, h: int -> real) returns (ghost r: real) {
  if f.requires(x) {
    r := f(x);
  }
  if g.requires(x) {
    r := g(x);
  }
  if
  case true =>
    r := f(x); // error: precondition may not hold
  case true =>
    r := g(x); // error: precondition may not hold
  case true =>
    r := h(x);
}

ghost function Usability1(x: int, f: int ~> real, g: int --> real, h: int -> real): real {
  var r0 := if f.requires(x) then f(x) else 30.0; // error: insufficient reads clause to call f.requires(x)
  var r1 := if g.requires(x) then g(x) else 30.0;
  var r2 := f(x); // error: precondition may not hold
  var r3 := g(x); // error: precondition may not hold
  var r4 := h(x);
  r0 + r1 + r2 + r3 + r4
}

ghost function Usability2(x: int, f: int ~> real, g: int --> real, h: int -> real): real
  reads f.requires.reads(x)
{
  var r0 := if f.requires(x) then f(x) else 30.0; // fine
  var r1 := if g.requires(x) then g(x) else 30.0;
  var r2 := f(x); // error: precondition may not hold
  var r3 := g(x); // error: precondition may not hold
  var r4 := h(x);
  r0 + r1 + r2 + r3 + r4
}

ghost function Usability3(x: int, f: int ~> real, g: int --> real, h: int -> real): real
  reads f.reads(x) // error: precondition of f.reads(x) may not hold
{
  var r0 := if f.requires(x) then f(x) else 30.0;
  var r1 := if g.requires(x) then g(x) else 30.0;
  var r2 := f(x); // under the assumption that the specification of Usability3 is well-formed, this line is fine
  var r3 := g(x); // error: precondition may not hold
  var r4 := h(x);
  r0 + r1 + r2 + r3 + r4
}

ghost function Usability4(x: int, f: int ~> real, g: int --> real, h: int -> real): real
  requires f.requires(x)
  reads f.reads(x)
{
  var r0 := if f.requires(x) then f(x) else 30.0;
  var r1 := if g.requires(x) then g(x) else 30.0;
  var r2 := f(x);
  var r3 := g(x); // error: precondition may not hold
  var r4 := h(x);
  r0 + r1 + r2 + r3 + r4
}

method Usability5(x: int, g: int --> real, h: int -> real) {
  assert g.requires(x) ==> g.reads(x) == {};
  assert g.reads(x) == {}; // error: this is known to hold only if g.requires(x) holds (TODO: is this a surprise or awkward?)

  assert h.requires(x);
  assert h.reads(x) == {};
}

method Usability6(x: int, f: int ~> real) {
  // The following is called Axiom 1 in the PR description:
  assert f.requires(x) ==> f.reads.requires(x);
  assert f.reads.requires(x) ==> f.requires(x); // TODO

  // The following is called Axiom 2 in the PR description:
  assert f.requires(x) ==> f.reads.reads(x) == f.reads(x);

  // The following is called Axiom 3 in the PR description:
  assert f.requires.requires(x);

  // The following is called Axiom 4 in the PR description:
  assert f.requires(x) ==> f.requires.reads(x) == f.reads(x);
}

method Usability7(x: int, g: int --> real) {
  assert g.requires(x) ==> g.reads(x) == {};

  assert g.requires(x) ==> g.reads.requires(x);
  assert g.reads.requires(x) ==> g.requires(x); // TODO

  assert g.requires(x) ==> g.reads.reads(x) == {};

  assert g.requires.requires(x);
  assert g.requires(x) ==> g.requires.reads(x) == {};
}

method Usability8(x: int, h: int -> real) {
  assert h.requires(x);
  assert h.reads(x) == {};

  assert h.reads.requires(x);
  assert h.reads.reads(x) == {};

  assert h.requires.requires(x);
  assert h.requires.reads(x) == {};
}

// Framing

class Cell {
  var data: int
}

method Framing0(x: int, f: int ~> real, cell: Cell)
  requires f.requires(x) && f(x) == 8.0
  requires cell !in f.reads(x)
  modifies cell
{
  cell.data := cell.data + 1;
  assert f.requires(x) && f(x) == 8.0;
}

method Framing1(x: int, f: int ~> real)
  requires f.requires(x) && f(x) == 8.0
{
  var cell := new Cell;
  cell.data := cell.data + 1;
  assert f.requires(x) && f(x) == 8.0;
}

method Framing2(x: int, f: int ~> real, cell: Cell)
  requires f.requires(x) && f(x) == 8.0
  modifies cell
{
  cell.data := cell.data + 1;
  assert f.requires(x); // error: assertion may not hold
  assert f(x) == 8.0; // error: assertion may not hold
}
```

RustanLeino:
The `assert false` I had included in test `CallIfPossible2` had confused me. In case it confuses anyone else, too, here are some replacements of that test, with comments:

``` dafny
function MyFunction(x: int, S: set<object>): real
  requires x < 100
  reads S
{
  assert true;
  (x + 12) as real
}

ghost function CallIfPossible2'0(x: int, S: set<object>): real
  reads S
{
  if MyFunction.requires(x, S) then // error: insufficient reads clause for CallIfPossible2'0 if this condition returns false
    MyFunction(x, S)
  else
    2.7
}

ghost function CallIfPossible2'1(x: int, S: set<object>): real
  reads *
{
  if MyFunction.requires(x, S) then // fine, since CallIfPossible2'1 can read *
    MyFunction(x, S)
  else
    2.7
}

ghost function CallIfPossible2'2(x: int, S: set<object>): real
  reads S
{
  if MyFunction.requires(x, S) then // the reads error here is performed at the end of the function body and is thus masked by the "assert false"
    MyFunction(x, S)
  else
    assert false; // error: control can indeed reach this point
    2.7
}

ghost function CallIfPossible2'3(x: int, S: set<object>, b: bool): real
  reads S
{
  if MyFunction.requires(x, S) then // error: insufficient reads (this error is not masked by "assert b")
    MyFunction(x, S)
  else
    assert b; // error: control can indeed reach this point
    2.7
}
```

Test/hofs/Examples.dfy
```diff
@@ -1,57 +0,0 @@
-// RUN: %testDafnyForEachCompiler "%s" --refresh-exit-code=0 -- --relax-definite-assignment
-
-function Apply<A,B>(f: A ~> B, x: A): B
-  reads f.reads(x)
```
MikaelMayer, line 4: Why is this example not supported? It's no longer sound?

Test/hofs/Twice.dfy
```diff
@@ -1,37 +0,0 @@
-// RUN: %exits-with 4 %dafny /print:"%t.print" "%s" > "%t"
-// RUN: %diff "%s.expect" "%t"
-
-function Twice<A>(f : A ~> A): A ~> A
-{
-  x requires f.requires(x) && f.requires(f(x))
-    reads f.reads(x), if f.requires(x) then f.reads(f(x)) else {}
```
MikaelMayer, line 7: Same here, I'm surprise this example is deleted.
Is it because we can't interleave requires and reads? It seems to me we should be able to write:

```dafny
x requires f.requires(x)
   reads f.reads(x)
   requires f.requires(f(x))
   reads f.reads(f(x)) => f(f(x))
```

MikaelMayer (commented):
I'm not sure why some files were removed. Can you please give more context?

Test/hofs/Examples.dfy
jtristan: Correct. If you want to call method f.reads, you need to know that its precondition holds. It used to be that the axioms were incorrect and that you could call that function without its precondition being met, but not anymore.

stefan-aws: It seems as if 
```
function Apply'<A,B>(f: A ~> B) : A ~> B
{
  x requires f.requires(x)
    reads assume f.requires(x); f.reads(x)
    => f(x)
}
```
verifies, but 
```
function Apply'<A,B>(f: A ~> B) : A ~> B
{
  x requires f.requires(x)
    reads assert f.requires(x); f.reads(x)
    => f(x)
}
```
fails. Shouldn't I be able to deduce the assertions from the precondition?

Test/hofs/Twice.dfy
stefan-aws: Similar to `Examples.dfy`, I find it odd that `TwicePointwise` verifies, but `Twice` doesn't:

```
function Twice<A>(f : A ~> A): A ~> A {
  x requires f.requires(x) && f.requires(f(x))
    reads f.reads(x) + f.reads(f(x)) 
    => TwicePointwise(f, x)
}

function TwicePointwise<A>(f : A ~> A, x: A): A 
  requires f.requires(x) && f.requires(f(x))
  reads f.reads(x) + f.reads(f(x))
{
  f(f(x))
}
```

Why is `f.requires(x)` shared between the `requires` and `reads`-clauses of `Twicepointwise`, but not between the clauses of the lambda inside of `Twice`?

Test/hofs/Examples.dfy
jtristan: Perhaps, but let's not confuse distinct problems. The purpose of this PR is to ensure that when a .reads function is called, its precondition is checked accordingly. This is a matter of soundness. It is not a goal of this PR to solve incompleteness problems with higher-order functions, including ones that result from now using sound axioms. We should instead file an issue when we believe that something ought to be verified and doesn't.

Test/hofs/Twice.dfy
jtristan: Twice is deleted because now that the axioms for .reads and .requires have been corrected to be sound, it has revealed that there are incompleteness issues. It is not a goal of this PR to fix such issues. Instead, if we believe that it ought to verify, then we should file a distinct issue to solve this incompleteness problem.

stefan-aws: Fair point -- I'll approve this PR and file a separate issue.

stefan-aws (approved):

PR #4403, opened 2023-08-09
Revert #4359

keyboardDrummer:
I approved that PR but on second thought I don't think it should be possible to generate diagnostics during Boogie translation.

You mentioned these two relevant diagnostics:
- "Some instances of this call are not inlined.", which can happen only if there is a trigger in the file, which generates an info diagnostic during resolution.
- "Fuel can only increase within a given scope." which can be triggered only if there is already a function with fuel, which thus has a "decreases" info diagnostic during resolution. This is an error only because of the encoding in Boogie.

If it is performance wise practical to already generate these errors during resolution, then I think we should. Even if the second one is only a result of the encoding in Boogie, right now it's still a limitation of Dafny in general. What's the benefit of delaying the publication of these errors until the user decides to trigger verification of a method?


MikaelMayer:
Thank you for raising that point and your legitimate concern. Let me give some lights on why this is necessary.

First There is a benefit of leaving errors for verification only. Resolution errors are there both for the verifier and the compiler. We normally don't output verification errors as part of resolution, because what the compiler needs is a resolved program, not a verified one. So there is room for there to be verification errors that don't impact resolver/compilation pipeline.

For the first point, we inline triggers because of the way Boogie works, but if we had an interactive prover, this error would not be needed. So it should not make it to resolution, it's really a limitation of the verifier.
Similarly for the fuel, our current Boogie encoding does not enable to decrease the fuel because we currently encode the fuel as Peano numbers and assume things about that, and it's not possible to revert assumptions. The resolution should not be aware of that, But if the encoding was different, e.g. we started to define fuel as a variable instead and modify it, then this error would not be needed. Again, it should not make it to the resolution, it's a limitation of the verifier.

Second, I would like soon to emit verification warnings about trivially true or trivially false statements. This is not possible if, during the translation phase, we could not emit warnings and errors, and it does not belong to the resolution phase.

The fact that we are using the Boogie verifier is an implementation detail, but for now it's our only source of truth. My goal is that it will change.

To mitigate your concern relative to plugins, I suggest we rename `PreTranslate` to `PreVerify` so that, from the user's perspective, they are not dealing with the translation phase but the verification phase.

MikaelMayer:
If you are satisfied with my answer, then we can close this PR.

keyboardDrummer:
> Second, I would like soon to emit verification warnings about trivially true or trivially false statements.

If these warnings are so trivial to emit, why would you not emit them immediately when opening a file in the IDE, instead of waiting for the user to trigger verification for the symbols that contain the warnings?

> For the first point, we inline triggers because of the way Boogie works, but if we had an interactive prover, this error would not be needed. So it should not make it to resolution, it's really a limitation of the verifier.

How would using different verification back-ends work? Suppose I have a piece of code that's verified with one back-end, can I re-use that in a piece of code verified with another back-end? I think the answer has to be yes to prevent a language split.

I think I'm on board with what you're suggesting but the concept of a program having errors with back-end#1 but not with with back-end#2 made me wonder whether there was a language split. I think the answer is that in both cases the program is correct, but back-end#1 was not able to prove while back-end#2 was.

MikaelMayer:
> > Second, I would like soon to emit verification warnings about trivially true or trivially false statements.
> 
> If these warnings are so trivial to emit, why would you not emit them immediately when opening a file in the IDE, instead of waiting for the user to trigger verification for the symbols that contain the warnings?

These warnings don't require verification, they require only translation, Since translation is a post-resolution phase, we could as well try to migrate them there indeed, but that would require to migrate a lot of machinery.

For example: the following assertion is trivially false:

```dafny
method Test(a: A) 
  requires a == this
  requires a.x != this.x // Trivially false
{
}
```
However, to detect it's trivially false, we still need to reason about conjuncts, and this is better done on my taste on a post-resolution phase.

> > For the first point, we inline triggers because of the way Boogie works, but if we had an interactive prover, this error would not be needed. So it should not make it to resolution, it's really a limitation of the verifier.
> 
> How would using different verification back-ends work? Suppose I have a piece of code that's verified with one back-end, can I re-use that in a piece of code verified with another back-end? I think the answer has to be yes to prevent a language split.

Yes, indeed, for the same reason that in the *.doo files we store the version of Z3, of Boogie, the seed used to verify the artifacts, the version of Dafny. The goal is that if one dafny file was verified once, then it's no longer necessary to reverify it even if Z3 changes, Boogie changes, or Dafny's internal translation changes.

> I think I'm on board with what you're suggesting but the concept of a program having errors with back-end#1 but not with with back-end#2 made me wonder whether there was a language split. I think the answer is that in both cases the program is correct, but back-end#1 was not able to prove it while back-end#2 was.

If two verification back-end disagree, it has to be indeed on one being able to verify and the other saying "unknown", meaning it could not find a proof.
That's also why I plan to emit warning in the case above and not an error because it's challenging the hypothesis; not the result.

PR #4411, opened 2023-08-11
Improve error correction in the parser

### LINKED ISSUE #4408: Parser error on IDE without any indication where it occurred
### Dafny version

4.2.0

```dafny
module Test {
  method {:test} () {
    
  }
}
```

```code
Paste in VSCode. you'll get a "Parse error" but no diagnostics and no visible crash.
```

### What happened?

I would have expected diagnostics telling me where the parse error is.

Windows

keyboardDrummer:
Fixes #4408

### Testing
Did not add a test since we need a more general mechanism to test the cloner code that's currently only run by the language server, instead of single tests that only cover a single situation.


MikaelMayer (approved):
Oh yeah thank you very much, I hit this one literally every day.

PR #4412, opened 2023-08-11
Fix backwards compatibility with regards to the VerifyOnMode option

keyboardDrummer:
Rename CLI option back to its previous name, for backwards compatibility reasons.


MikaelMayer (approved):
yeah I could have caught that on the review of the last PR. Good quick fix.

PR #4413, opened 2023-08-11
Fix flickering and incorrect results in the verification status bar, and improve responsiveness of diagnostics

keyboardDrummer:
Fixes https://github.com/dafny-lang/ide-vscode/issues/409

### Changes
- Previously, due to a bug in `StatusFromBoogieStatus`, a BatchCompleted notification would cause the status bar to consider this verifiable to have finished verification, even if other batches were still running. This is the cause for the behavior "status bar flickers between Verification Succeeded" and "Could not prove the declaration"", because it will show the status of the last completed batch.
- Update diagnostics after each completed batch. This improves the behavior "whole lot of errors finally appear in the "problems" window"

### Testing
- Add test `NoFlickeringWhenMixingCorrectAndErrorBatches`
- Add test `IncrementalBatchDiagnostics`

### DIFF
--- a/Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
+++ b/Source/DafnyLanguageServer.Test/Synchronization/DiagnosticsTest.cs
@@ -958,24 +958,6 @@ method test()
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


MikaelMayer (approved):
To be merged ASAP! This bug has been annoying me for a while as well.

PR #4414, opened 2023-08-14
Use more caching when migrating information between document updates

keyboardDrummer:
### Changes

Small improvement to performance by doing more caching when migrating information between document updates.

Specifically, the cache `Dictionary<TextDocumentContentChangeEvent, Position> getPositionAtEndOfAppliedChangeCache;` is re-used when migrating different ranges/positions/verification trees/symbol tables

### Testing

No performance tests have been added/changes. We currently don't have any.


MikaelMayer (approved):
I understand you transformed the class not to be a service, but what are the rules for a class to be a service? Is it that it would be shared and might need expensive resources to set up?

PR #4418, opened 2023-08-14
Chore: Adding well formedness in the documentation of Dafny

### LINKED ISSUE #4333: Document what well-formedness is about

I suggest we add to the document what well-formedness is about so that people can reason about it.
We mention well-formedness 4 times but never explain it.

A simple example could help as well
```dafny
class A {
  var y: int
}

method Test(a: A, x: int)
  requires x > 0 && a.y / x > 0 // Must be verified independently
{
  assert x * a.y > 0;
}

lemma TestWellFormedness(a: A, x: int)
{
  assume x > 0;
  assert x != 0;
}
```

MikaelMayer:
Fixes #4333

### DIFF
--- a/docs/DafnyRef/Specifications.md
+++ b/docs/DafnyRef/Specifications.md
@@ -844,3 +844,58 @@ add:
    requires Valid()
    reads Repr
 ```
+
+## 7.8. Well-formedness of specifications {#sec-well-formedness}
+
+Dafny verifies that the requires and ensures clauses make sense by themselves, independently of the body.
+Hence, by declaring the following method:
+
+```dafny
+method Test(a: array<int>) returns (j: int)
+  requires a.Length >= 1
+  ensures a.Length % 2 == 0 ==> j >= 10 / a.Length
+{
+  j := 20;
+  var divisor := a.Length;
+  if divisor % 2 == 0 {
+    j := j / divisor;
+  }
+}
+```
+
+Dafny will split the verification in two [assertion batches](#sec-assertion-batches)
+that will roughly look like the following lemmas:
+
+```dafny
+lemma Test_WellFormed(a: array<int>)
+{
+  assert a != null;       // for the `requires a.Length >= 1``
+  assert a != null;       // Again for the `a.Length % 2``
+  if a.Length % 2 == 0 {
+    assert a != null;     // Again for the final `a.Length``
+    assert a.Length != 0; // Because of the 10 / a.Length
+  }
+}
+
+lemma Test_Correctness(a: array<int>)
+{ // Here we assume the well-formedness of the condition
+  assume a != null;       // for the `requires a.Length >= 1``
+  assume a != null;       // Again for the `a.Length % 2``
+  if a.Length % 2 == 0 {
+    assume a != null;     // Again for the final `a.Length``
+    assume a.Length != 0; // Because of the 10 / a.Length
+  }
+
+  // Now the body is translated
+
+  assert a != null; // For `var divisor := a.Length;`
+  var divisor := a.Length;
+  if * {
+    assume divisor % 2 == 0;
+    assert divisor != 0;
+  }
+  assume divisor % 2 == 0 ==> divisor != 0;
+}
+```
+
+If, when hovering over a method name, you see two assertion batches, it's usually because of that decomposition.
--- a/docs/DafnyRef/Attributes.md
+++ b/docs/DafnyRef/Attributes.md
@@ -225,7 +225,7 @@ without proof. In that case also the body of the function or
 method may be omitted.
 
 The `{:axiom}` attribute only prevents Dafny from verifying that the body matches the post-condition.
-Dafny still verifies the well-formedness of pre-conditions, of post-conditions, and of the body if provided.
+Dafny still verifies the [well-formedness](#sec-well-formedness) of pre-conditions, of post-conditions, and of the body if provided.
 To prevent Dafny from running all these checks, one would use [`{:verify false}`](#sec-verify), which is not recommended.
 
 The compiler will still emit code for an [`{:axiom}`](#sec-axiom), if it is a [`function`, a `method` or a `function by method`](#sec-function-declarations) with a body.
@@ -490,7 +490,7 @@ Boogie `-timeLimit` command-line option.
 ### 11.2.19. `{:verify false}` {#sec-verify}
      
 Skip verification of a function or a method altogether,
-not even trying to verify the well-formedness of postconditions and preconditions.
+not even trying to verify the [well-formedness](#sec-well-formedness) of postconditions and preconditions.
 We discourage using this attribute and prefer [`{:axiom}`](#sec-axiom),
 which performs these minimal checks while not checking that the body satisfies the postconditions.
 
--- a/docs/DafnyRef/Expressions.md
+++ b/docs/DafnyRef/Expressions.md
@@ -459,7 +459,7 @@ The _specification_ is a list of clauses `requires E` or
 expression.
 
 _body_ is an expression that defines the function's return
-value.  The body must be well-formed for all possible values of the
+value.  The body must be [well-formed](#sec-well-formedness) for all possible values of the
 parameters that satisfy the precondition (just like the bodies of
 named functions and methods).  In some cases, this means it is
 necessary to write explicit `requires` and `reads` clauses.  For
@@ -468,7 +468,7 @@ example, the lambda expression
 ```dafny
 x requires x != 0 => 100 / x
 ```
-would not be well-formed if the `requires` clause were omitted,
+would not be [well-formed](#sec-well-formedness) if the `requires` clause were omitted,
 because of the possibility of division-by-zero.
 
 In settings where functions cannot be partial and there are no
--- a/docs/DafnyRef/Grammar.md
+++ b/docs/DafnyRef/Grammar.md
@@ -511,7 +511,7 @@ may also include any of three optional elements:
 
 Note that a variable's domain expression may reference any variable declared before it,
 and a variable's range expression may reference the attached variable (and usually does) and any variable declared before it.
-For example, in the quantifier domain `i | 0 <= i < |s|, y <- s[i] | i < y`, the expression `s[i]` is well-formed
+For example, in the quantifier domain `i | 0 <= i < |s|, y <- s[i] | i < y`, the expression `s[i]` is [well-formed](#sec-well-formedness)
 because the range attached to `i` ensures `i` is a valid index in the sequence `s`.
 
 Allowing per-variable ranges is not fully backwards compatible, and so it is not yet allowed by default;
--- a/docs/DafnyRef/Statements.md
+++ b/docs/DafnyRef/Statements.md
@@ -1643,7 +1643,7 @@ the loop needs to be declared either with `invariant c.data == 100` or with
 When a loop has an explicit `modifies` clause, there is, at the top of
 every iteration, a proof obligation that
 
-* the expressions given in the `modifies` clause are well-formed, and
+* the expressions given in the `modifies` clause are [well-formed](#sec-well-formedness), and
 * everything indicated in the loop `modifies` clause is allowed to be modified by the
   (effective `modifies` clause of the) enclosing loop or method.
 
--- a/docs/DafnyRef/Types.md
+++ b/docs/DafnyRef/Types.md
@@ -181,7 +181,7 @@ C <== B <== A
 To illustrate the short-circuiting rules, note that the expression
 `a.Length` is defined for an array `a` only if `a` is not `null` (see
 [Section 5.1.2](#sec-reference-types)), which means the following two
-expressions are well-formed:
+expressions are [well-formed](#sec-well-formedness):
 <!-- %no-check -->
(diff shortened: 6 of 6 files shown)


docs/DafnyRef/Specifications.md
```diff
@@ -844,3 +844,58 @@ add:
...
+  var divisor := a.Length;
+  if * {
+    assume divisor % 2 == 0;
+    assert divisor != 0;
+  }
+  assume divisor % 2 == 0 ==> divisor != 0;
+}
+```
+
+If, when hovering over a method name, you see two assertion batches, it's usually because of that decomposition.
```
fabiomadge, line 901:
```suggestion
For this reason the IDE typically reports at least two assertion batches when hovering a method.
```

docs/DafnyRef/Specifications.md
```diff
@@ -844,3 +844,58 @@ add:
    requires Valid()
    reads Repr
 ```
+
+## 7.8. Well-formedness of specifications {#sec-well-formedness}
```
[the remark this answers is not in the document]
MikaelMayer: Ok, in the next version, I reworded the section about assertion batches so that I clarify that implicit assertions are actually the well-formedness of expressions, and changed the links to them. Indeed, it's better. That way, I can keep this title intact. Thanks for this useful feedback.

docs/DafnyRef/Specifications.md
```diff
@@ -844,3 +844,58 @@ add:
...
+```
+
+Dafny will split the verification in two [assertion batches](#sec-assertion-batches)
+that will roughly look like the following lemmas:
+
+```dafny
+lemma Test_WellFormed(a: array<int>)
+{
+  assert a != null;       // for the `requires a.Length >= 1``
+  assert a != null;       // Again for the `a.Length % 2``
```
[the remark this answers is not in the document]
MikaelMayer: Excellent remark.

docs/DafnyRef/Specifications.md
```diff
@@ -844,3 +844,58 @@ add:
...
+}
+```
+
+Dafny will split the verification in two [assertion batches](#sec-assertion-batches)
+that will roughly look like the following lemmas:
+
+```dafny
+lemma Test_WellFormed(a: array<int>)
+{
+  assert a != null;       // for the `requires a.Length >= 1``
```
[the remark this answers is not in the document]
MikaelMayer: No it does not because it's in triple backticks anyway. But I see that in many places where I intended only one backtick, it writes two, let me fix that.

docs/DafnyRef/UserGuide.md
```diff
@@ -1147,25 +1147,30 @@ method innerLoop()
...
+### 13.6.3. Assertion batches, well-formedness, correctness {#sec-assertion-batches}
 
 To understand how to control verification,
 it is first useful to understand how `dafny` verifies functions and methods.
 
-For every method (or function, constructor, etc.), `dafny` extracts _assertions_. Here is a non-exhaustive list of such extracted assertions:
+For every method (or function, constructor, etc.), `dafny` extracts _assertions_.
+Assertions can either be *implicit* or *explicit*.
+Implicit assertions, also called *well-formedness assertions*, ensure expressions make sense.
+Explicit assertions, also called *correctness assertions*, are provided by the user, and can either help prove implicit assertions, or prove other desired higher-level properties.
```
fabiomadge, line 1158:
```suggestion
Explicit assertions, also called *correctness assertions*, are provided by the user, and can either help prove other assertions, or the specification.
```
Do we use the term _correctness assertion_ elsewhere?

docs/DafnyRef/Specifications.md
```diff
@@ -844,3 +844,61 @@ add:
    requires Valid()
    reads Repr
 ```
+
+## 7.8. Well-formedness of specifications {#sec-well-formedness}
+
+Dafny ensures that the [`requires` clauses](#sec-requires-clause) and [`ensures` clauses](#sec-ensures-clauses) are [well-formed](#sec-assertion-batches) independent of the body they belong to. Examples of conditions this rules out are null pointers, out-of-bounds array access, and division by zero.
```
fabiomadge, line 850:
```suggestion
Dafny ensures that the [`requires` clauses](#sec-requires-clause) and [`ensures` clauses](#sec-ensures-clauses) are [well-formed](#sec-assertion-batches) independent of the body they belong to. Examples of conditions this rules out are null pointer dereferencing, out-of-bounds array access, and division by zero.
```

docs/DafnyRef/Specifications.md
```diff
@@ -844,3 +844,61 @@ add:
...
+  assert a != null;       // for the `requires a.Length >= 1`
+  assume a.Length >= 1;   // After well-formedness, we assume the requires
+  assert a != null;       // Again for the `a.Length % 2`
+  if a.Length % 2 == 0 {
+    assert a != null;     // Again for the final `a.Length`
+    assert a.Length != 0; // Because of the 10 / a.Length
+  }
+}
+
+lemma Test_Correctness(a: array<int>)
```
fabiomadge, line 881: Is this actually legal within a lemma? Maybe this needs to be a (ghost) method.

docs/DafnyRef/UserGuide.md
MikaelMayer: my new version:
Explicit assertions, also called *correctness assertions*, are provided by the user, and can help prove other implicit and explicit assertions.

> Do we use the term correctness assertion elsewhere?

yes in boogie 😸

docs/DafnyRef/Specifications.md
MikaelMayer: With a var in front of the j it is legal. but I did convert it to a method.
I also added check-verify.

docs/DafnyRef/UserGuide.md
MikaelMayer: We have the duality implicit/explicit assertion.
We have a term to qualify checking implicit assertions, which is "well-formedness check". But often that confuses users because they don't make the distinction, for them Dafny is just verifying everything.
So we need another term to say that an assertion is not part of the well-formedness check, and it's the correctness check, which is the official term we use when turning on logging in the CLI.

MikaelMayer: As well-formedness does not convey much meaning to you, I recommend you read the following section of the documentation which use the term "well-formed" in context with examples and might give you some intuition about why this term exist.
https://dafny.org/dafny/DafnyRef/DafnyRef#sec-implication-and-reverse-implication
https://dafny.org/dafny/DafnyRef/DafnyRef#sec-multi-dimensional-arrays
https://dafny.org/dafny/DafnyRef/DafnyRef#sec-inductive-datatypes

Under the hood, this term is different from regular correctness because expressions in Dafny are translated almost raw into mathematical formulas, and formulas can be constructed only if we can independently prove they make sense in the first place, and this step is called well-formedness check.

> Also, what's the reason we need a separate correctness and well-formedness check again? 

It should be totally doable and equivalent to do the verification of both things in a single implementation, and we probably should down the road as there are ways to do that in Boogie. But merging the two checks in a single Boogie implementation does not intrinsically make them the same thing.

Moreover, the distinction between well-formedness and correctness is a bit blurred. For example, if there is an explicit assert StmtExpr as part of the requires or ensures clause, it is still checked during the well-formedness of the pre/post condition even though it's explicit. For expressions, this distinction exists as well, although well-formedness is checked in the same scope as correctness.

keyboardDrummer: > I recommend you read the following section of the documentation which use the term "well-formed" in context with examples and might give you some intuition about why this term exist.

Thanks

Do I understand correctly that code is well-formed if it satisfies all requires clauses of the calls it makes, be it calls to native code such as indexing and division, or calls to user-code? If the answer is yes, then we should add that definition to the documentation.

> For example, if there is an explicit assert StmtExpr as part of the requires or ensures clause, it is still checked during the well-formedness of the pre/post condition even though it's explicit. For expressions, this distinction exists as well, although well-formedness is checked in the same scope as correctness.

I think that's to be expected, because a requires clause translates to an assume statement in the callee, and all calls in the body of the callee are part of its WF check, including that assume statement.

MikaelMayer: > Do I understand correctly that code is well-formed if it satisfies all requires clauses of the calls it makes, be it calls to native code such as indexing and division, or calls to user-code? 

That made me realize that well-formedness makes a distinction between expressions and statements, which are both code. There is no well-formedness of statements (this is correctness) and there is no correctness of expressions (this is well-formedness).
I'll reword it

keyboardDrummer: What do you mean with there is no well-formedness of statements? If I make a call to a method, which is a statement, then that introduces a check in the well-formedness Boogie `implementation`, does it not?

MikaelMayer: What check does it introduce in the case you mention?
Note that I'm taking expressions in a large sense, for example ExprRHS is an expression, so for a method call to be validated, one has to prove its well-formedness, i.e. the requires clause holds. But that's not the well-formedness of the statement, only expressions.
Similarly, when having an assignment like A.x := 3;, A is an expression that computes an object, and a part of the well-formedness is that A is not null.

keyboardDrummer: Okay.. well, regardless of what we call expressions or statements, how do feel about what I said before?

> Do I understand correctly that code is well-formed if it satisfies all requires clauses of the calls it makes, be it calls to native code such as indexing and division, or calls to user-code? 

> If the answer is yes, then should we add that definition to the documentation? It seems the documentation currently only defines what well-formedness is by example. Also, I would consider replacing the term "well-formed" with "satisfies requires clauses" which is more verbose but self-explanatory.

MikaelMayer: Happy to make the documentation more user-meaningful with your help.

> Do I understand correctly that [expressions are] well-formed if [they satisfy] all requires clauses of the calls it makes, be it calls to native code such as indexing and division, or calls to user-code?

I would say yes, that's about that. That way, it's possible to define the expression as a mathematical expression without having to bother that it assumes something that is wrong.

Looking at the documentation, when it says `a.Length < 0 ==> a == null  // not well-formed`, it becomes evident that the formula to not be-well-formed requires that that there is no assumption before that, as if it was the first requires clause on a function, which we don't know out of context. Hence, I realize that the example `a.Length < 0 ==> a == null  // not well-formed` is not well-formed :-) and well-formedness also depends on the context.

> If the answer is yes, then should we add that definition to the documentation? It seems the documentation currently only defines what well-formedness is by example. Also, I would consider replacing the term "well-formed" with "satisfies requires clauses" which is more verbose but self-explanatory.

Let me work on that.

docs/DafnyRef/UserGuide.md
```diff
@@ -1153,13 +1153,58 @@ To understand how to control verification,
...
 
 For every method (or function, constructor, etc.), `dafny` extracts _assertions_.
-Assertions extracted from expressions are called *well-formedness assertions*.
-Assertions extracted from statements and clauses are called *correctness assertions*.
-Other than that, these are all assertions Dafny needs to verify that the code is correct,
-but that also your specification makes sense.
-Here is a non-exhaustive list of such extracted assertions:
+Assertions can roughly be sorted into two kinds: Well-formedness and correctness.
+
+- _Well-formedness_ assertions: All the implicit requirements
```
keyboardDrummer, line 1158: Nice and accurate explanation! Thanks

docs/DafnyRef/UserGuide.md
```diff
@@ -1153,13 +1153,58 @@ To understand how to control verification,
...
+} else {
+  assert f.requires(a); // Well-formedness
+}
+c := if b then 3/a else f(a);
+assert a != 0;       // Well-formedness
+assert c != 5/a;     // Correctness
+```
+
+Well-formedness is proved at the same time as correctness, except for
+[well-formedness of requires and ensures clauess](#sec-well-formedness-specifications)
```
keyboardDrummer, line 1200: clauses

docs/DafnyRef/UserGuide.md
```diff
@@ -1153,13 +1153,58 @@ To understand how to control verification,
...
+  assert a != 0;     // Well-formedness
+} else {
+  assert f.requires(a); // Well-formedness
+}
+c := if b then 3/a else f(a);
+assert a != 0;       // Well-formedness
+assert c != 5/a;     // Correctness
+```
+
+Well-formedness is proved at the same time as correctness, except for
```
keyboardDrummer, line 1199: What does at the same time mean? In the same VC ?

keyboardDrummer (dismissed):

docs/DafnyRef/UserGuide.md
MikaelMayer:
```suggestion
[well-formedness of requires and ensures clauses](#sec-well-formedness-specifications)
```

docs/DafnyRef/UserGuide.md
MikaelMayer: Yes, in the same VC indeed.

keyboardDrummer (dismissed):

keyboardDrummer (approved):

PR #4425, opened 2023-08-15
docs: remove erroneous syntax breaking table

SignalRichard:
Fixes type mapping table for C# Integration documentation here: [IntegrationCS](https://dafny.org/dafny/DafnyRef/integration-cs/IntegrationCS#types)


MikaelMayer (approved):
Indeed, I had not seen this file was broken, thanks for fixing it!

PR #4427, opened 2023-08-16
Various fixes related to prefix modules and the IDE

keyboardDrummer:
### Changes

- Correctly set the RangeToken of prefix modules, which enables nodes inside of these to be found using position based lookup
- Let `DafnyDocumentSymbolHandler` correctly work with prefix modules
- Add types to position based lookup which can improve the behavior if ranges are not correctly set, by using the types as additional information to filter the result.
- Rename `VerifyTask` to `VerifySymbol`

### Testing

Added two tests related to prefix modules, one for document symbol and one for verifying a symbol.


MikaelMayer (approved):
The fix looks good to me, thanks a lot.

EkanshdeepGupta (approved):

Issue #3855, opened 2023-04-10
{:ignore} attribute crashes dafny with a null pointer error  


kjx:
### Dafny version

4.0.0.5.0.3

### Code to produce this issue

```dafny
//SIMULADA 

//vh-v41.dfy - starting to transform to the venice memory - and stack
//             renamed LOTS of stuff, to get unique names across the lot
//vh-v40.dfy - runs on Dafny 4.0
//vh28i6.dfy - revised the fGC implementation
//vh28i5.dfy - v28i4 witout parallel seq "oid" model etc.
//vh28i4.dfy - exactly the same as v28i3 but with more comments!
//vh28i3.dfy - now verifies: the problem was evil precedence between "forall" vs "&&"
//vh28i2.dfy - DOES NOT VERIFY adds in a oidSeq parallel to oidMap and oidCcounter
//vh28i.dfy - trial with map & counter for object ids.  went through in about 25mins!
//vh28.dfy - the collated model of Verona(Mk1). verifies.
//vh27.dfy - documents attempts to formalise Verona access restrictions
//           including all the tables where things are got wrong.
//vh26.dfy - sketch of Verona memory. Mut objects have an owing iso... 
//           split "Kind" into "static" Mode and "dynamic" region
//df25.dfy - sketch of start of allocated/deallocated tracking...
//dh24.dfy - memory variant saying that immutables don't change
//dh23.dfy - further cleanup... 
//dh22.dfy - cleaned up & verified version
//dh21.dfy - verified version which I iwos
//dh20.dfy - slightly clean up version of dh19. removed lots of commented out versions of stuff 
//dh19.dfy - un-broke fNullify;  got dynMove running (thats the dynaically typed version)
//dh18.dfy - verifies...  worked on adding dynMove.
//    sleepily attmped to remove unnecessary asserts from fNullify while it was set to verify false
//    consequnelt fNullify broken (although most of the rst should work if you turn it back off)
//dh17.dfy - cleaned up version of dh16. verifies clean
//         - fInitialise handles Isos correctly - but requires 22s for verificatin rather than 4 or 5 wthout.
//dh16.dfy - messy but verified version of memory model with initialise
//

//////////////////////////////////////////////////////////////////////////////
//////////////////////////////////////////////////////////////////////////////
[...]

PR #4428, opened 2023-08-16
Optimize uses clauses

atomb:
(Note that this is @zafer-esen's work. I created the PR for permissions reasons.)

Here is an explanation for this PR, copied from the comment below.

Dafny uses the Boogie option `prune` by default. (Introduced in [Boogie PR #427](https://github.com/boogie-org/boogie/pull/427), although the semantics explained there is [out of date](https://github.com/boogie-org/boogie/pull/767#issuecomment-1668012521).).

The `prune` option, when enabled, runs a reachability analysis in Boogie, and prunes away functions, constants and axioms that are not reachable from root nodes. This greatly reduces resource counts, and leads to less brittleness; which is why pruning is intentionally made as strong as possible.

The semantics of pruning, at a high level, is as follows.

| | Axioms without quantifiers and axioms _without_ triggers | Axioms with both quantifiers and triggers |
| --- | --- | --- |
| Inside a `uses` clause of some symbol `s` | Do not prune if `s` is reachable. | Do not prune if (a) `s` is reachable _or_ (b) **all symbols in one of its triggers are reachable.** |
| Not inside a `uses` clause | Always pruned. | Do not prune if (b) **all symbols in one of its triggers are reachable.** |

If (a) holds but (b) does not, this leads to a weaker pruning if an axiom with quantifiers and triggers is inside a `uses` clause. (b) matches the semantics of how axioms with triggers are instantiated (AFAIK), so axioms with both quantifiers and triggers should not be inside `uses` clauses (with the exception of purely polymorphic quantifiers, more on this at the end of this comment).

Consider the [following lines](https://github.com/dafny-lang/dafny/blob/c31932b4b4f77a38e5da9c9f6a7689d8f57346bf/Source/DafnyCore/DafnyPrelude.bpl#L230-L290) from `DafnyPrelude.bpl`:
```
function $Is<T>(T,Ty): bool uses {           // no heap for now
    axiom(forall v : int  :: { $Is(v,TInt) }  $Is(v,TInt));
    axiom(forall v : real :: { $Is(v,TReal) } $Is(v,TReal));
    axiom(forall v : bool :: { $Is(v,TBool) } $Is(v,TBool));
    axiom(forall v : char :: { $Is(v,TChar) } $Is(v,TChar));
    axiom(forall v : ORDINAL :: { $Is(v,TORDINAL) } $Is(v,TORDINAL));
    [...] // more axioms for every type
 }
```

`$Is` is a function symbol and `TInt`, `TReal` etc. are constant symbols, i.e., the axioms shown are quantified and all have triggers with two function or constant symbols. Since they are inside the `uses` clause of `$Is`, none of these axioms can be pruned if `$Is` is reachable. However, if the second symbol inside the triggers, say `TReal` in the second axiom, is still unreachable, that axiom cannot be triggered despite being preserved (but will negatively affect resource counts).

This PR moves all quantified axioms with triggers outside of their `uses` clauses. The one exception is purely polymorphic quantifiers ([example](https://github.com/dafny-lang/dafny/blob/2a2e1c41af9b89c10437abc71cca92eb818e02a8/Source/DafnyCore/DafnyPrelude.bpl#L970-L972)), which should remain inside a `uses` clause, as these quantifiers disappear if monomorphization is used in Boogie.


zafer-esen:
Here is an explanation for this PR.

Dafny uses the Boogie option `prune` by default. (Introduced in [Boogie PR #427](https://github.com/boogie-org/boogie/pull/427), although the semantics explained there is [out of date](https://github.com/boogie-org/boogie/pull/767#issuecomment-1668012521).).

The `prune` option, when enabled, runs a reachability analysis in Boogie, and prunes away functions, constants and axioms that are not reachable from root nodes. This greatly reduces resource counts, and leads to less brittleness; which is why pruning is intentionally made as strong as possible.

The semantics of pruning, at a high level, is as follows.

| | Axioms without quantifiers and axioms _without_ triggers | Axioms with both quantifiers and triggers |
| --- | --- | --- |
| Inside a `uses` clause of some symbol `s` | Do not prune if `s` is reachable. | Do not prune if (a) `s` is reachable _or_ (b) **all symbols in one of its triggers are reachable.** |
| Not inside a `uses` clause | Always pruned. | Do not prune if (b) **all symbols in one of its triggers are reachable.** |

If (a) holds, this leads to a weaker pruning (preserving axioms that cannot be triggered). (b) matches the semantics of how triggers are instantiated (AFAIK), so axioms with both quantifiers and triggers should not be inside `uses` clauses (with the exception of purely polymorphic quantifiers, more on this at the end of this comment).

Consider the [following lines](https://github.com/dafny-lang/dafny/blob/c31932b4b4f77a38e5da9c9f6a7689d8f57346bf/Source/DafnyCore/DafnyPrelude.bpl#L230-L290) from `DafnyPrelude.bpl`:
```
function $Is<T>(T,Ty): bool uses {           // no heap for now
    axiom(forall v : int  :: { $Is(v,TInt) }  $Is(v,TInt));
    axiom(forall v : real :: { $Is(v,TReal) } $Is(v,TReal));
    axiom(forall v : bool :: { $Is(v,TBool) } $Is(v,TBool));
    axiom(forall v : char :: { $Is(v,TChar) } $Is(v,TChar));
    axiom(forall v : ORDINAL :: { $Is(v,TORDINAL) } $Is(v,TORDINAL));
    [...] // more axioms for every type
 }
```

`$Is` is a function symbol and `TInt`, `TReal` etc. are constant symbols, i.e., the axioms shown are quantified and have triggers with at least two function or constant symbol. Since they are inside the `uses` clause of `$Is`, none of these axioms can be pruned if `$Is` is reachable. However, if the second symbol inside the triggers, say `TReal` in the second axiom, is still unreachable, that axiom cannot be triggered despite being preserved (but will negatively affect resource counts).

This PR moves all quantified axioms with triggers outside of their `uses` clauses. The one exception is purely polymorphic quantifiers ([example](https://github.com/dafny-lang/dafny/blob/2a2e1c41af9b89c10437abc71cca92eb818e02a8/Source/DafnyCore/DafnyPrelude.bpl#L970-L972)), which should remain inside a `uses` clause, as these quantifiers disappear if monomorphization is used in Boogie.

Source/DafnyCore/DafnyPrelude.bpl
```diff
@@ -7,8 +7,9 @@
 // Copyright by the contributors to the Dafny Project
 // SPDX-License-Identifier: MIT
 
-const $$Language$Dafny: bool;  // To be recognizable to the ModelViewer as
-axiom $$Language$Dafny;        // coming from a Dafny program.
+const $$Language$Dafny: bool uses {  // To be recognizable to the ModelViewer as
+  axiom $$Language$Dafny;            // coming from a Dafny program.
```
zafer-esen, line 11: For your review: in addition to the described changes, I put this axiom inside a `uses` clause. Not sure of its implications.

Source/DafnyCore/DafnyPrelude.bpl
```diff
@@ -572,9 +565,9 @@ const unique allocName: NameFamily;
 // -- Arrays -----------------------------------------------------
 // ---------------------------------------------------------------
 
-function _System.array.Length(a: ref): int uses {
-  axiom (forall o: ref :: 0 <= _System.array.Length(o));
-}
+function _System.array.Length(a: ref): int;
+axiom (forall o: ref :: {_System.array.Length(o)} 0 <= _System.array.Length(o));
```
zafer-esen, line 569: I added a trigger here. Alternatively the trigger can be removed and the axiom moved back into the `uses` clause, if someone thinks this is a bad idea. `_System.array.Length` does not appear in any other triggers, so adding the trigger should not change the semantics I think.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -38,6 +38,15 @@ public partial class Translator {
...
 
     void AddOtherDefinition(Bpl.Declaration declaration, Axiom axiom) {
+      sink.AddTopLevelDeclaration(axiom);
+
+      // Axioms that have triggers and bound variables do not need to be inside
+      // uses clauses. Putting such axioms inside uses clauses weakens pruning
+      // when the trigger contains more than one function or constant symbol combined.
+      if (axiom.Expr is Microsoft.Boogie.QuantifierExpr qe && qe.Dummies != null && qe.Dummies.Any() &&
+          qe.Triggers != null && qe.Triggers.Tr != null && qe.Triggers.Tr.Any()) {
+        return;
```
keyboardDrummer, line 48: Can you add a comment that gives an example of when this return would trigger for Dafny?

Test/git-issues/git-issue-3855.dfy
```diff
@@ -666,26 +666,26 @@ method {:rlimit 30000} {:vcs_split_on_every_assert} fNullify(o : Object, f : str
   
       var nedge :=  Edge(o,f, o.fields[f]); //edge to be nullified
 
-      assert nedge in edges(objects);
+      //assert nedge in edges(objects);
```
keyboardDrummer, line 669: What's up with the commented out assertions?

Source/DafnyCore/DafnyPrelude.bpl
keyboardDrummer: How does the trigger affect the SMT solving? Was there a chance it would previously instantiate the quantifier even though there was no `_System.array.Length` call ?

zafer-esen: Z3 infers patterns if none are provided. In this case this is trivial, it would be the same pattern I manually added. This change should have no affect on SMT solving.

Test/git-issues/git-issue-3855.dfy
zafer-esen: This test seemed to be brittle and broke with my changes - it started failing with `:rlimit 30000`. Commenting out the assertions seemed to reduce the resource count, making it pass again. The assertions did not seem relevant for #3855.

Source/DafnyCore/DafnyPrelude.bpl
```diff
@@ -33,35 +34,35 @@ const unique TORDINAL  : Ty uses {
...
-  axiom (forall t, u: Ty :: { TMap(t,u) } Inv0_TMap(TMap(t,u)) == t);
-  axiom (forall t, u: Ty :: { TMap(t,u) } Inv1_TMap(TMap(t,u)) == u);
-  axiom (forall t, u: Ty :: { TMap(t,u) }    Tag(TMap(t,u))    == TagMap);
-}
-function TIMap(Ty, Ty) : Ty uses {
-  axiom (forall t, u: Ty :: { TIMap(t,u) } Inv0_TIMap(TIMap(t,u)) == t);
-  axiom (forall t, u: Ty :: { TIMap(t,u) } Inv1_TIMap(TIMap(t,u)) == u);
-  axiom (forall t, u: Ty :: { TIMap(t,u) }   Tag(TIMap(t,u))   == TagIMap);
-}
+function TBitvector(int) : Ty;
```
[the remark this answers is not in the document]
zafer-esen: We are already depending on the trigger inference of Boogie when triggers have multiple function / constant symbols, so something changing about trigger inference would already break the prelude. Moving axioms with only a single term out of their `uses` clauses would only amplify the brokenness :-) About the documentation aspect, don't triggers provide exactly the same information?

I think moving to `used_by` (or an alternative keyword without underscores), and disabling trigger inference completely would be the better approach, eventually. This takes away the possibility of misusing `uses` clauses (which would weaken pruning), and would be more robust as a result of not relying on automatic inference in Boogie.

Source/DafnyCore/Verifier/Translator.cs
zafer-esen: I just did.

I should note that, this will not return early when the outermost expression of an axiom is not quantified. This seems to happen in practice in axioms like `k < SomeConstant ==> forall ...`, so these axioms will still be in `uses` clauses. I do not really understand what happens at solver level with such axioms, so I did not want to change this.

Source/DafnyCore/DafnyPrelude.bpl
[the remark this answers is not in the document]
keyboardDrummer: > Overall, this PR make a remarkable improvement to the mean and standard deviation of both time and resource count on a large set of benchmarks.

I think it would be good to understand where the differences in pruning are.

> I don't know how much is attributable to this particular change, but it's conceivable that moving these out of the uses clause could lead to more pruning.

How do you conceive that? I think trigger based inference would infer the same dependencies.

> About the documentation aspect, don't triggers provide exactly the same information?

I don't think so, because triggers are nested inside axioms, and the axiom may be placed in a location distant from the owning declaration. I think the `uses` clause provides the clearest way to say that an axiom is only relevant for this declaration. Note that Boogie also has the concept of a `DefinitionAxiom` which is a single axiom that serves to define the function, with the syntax `function A(<parameters>): ReturnType { <definitionAxiom> }`. I think this serves a similar purpose as the `uses` clause but has the downside that it supports one axiom at most.

I would argue we should keep the `uses` clauses for now even if we add `used_by`, because I think >90% of axiom cases can be covered by the more documenting `uses`.

keyboardDrummer (approved):

Test/git-issues/git-issue-3855.dfy
keyboardDrummer: Could you remove the commented out code?

Source/DafnyCore/DafnyPrelude.bpl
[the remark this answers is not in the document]
zafer-esen: > I think it would be good to understand where the differences in pruning are.

The benefit arises in axioms where (a) in the following sentence holds, but (b) doesn't.

- "Do not prune if (a) s is reachable or (b) all symbols in one of its triggers are reachable."

By moving such axioms out of their `uses` clauses, we are strengthening it by removing the disjunction.

- "Do not prune if (b) all symbols in one of its triggers are reachable."

> How do you conceive that? I think trigger based inference would infer the same dependencies.

You're right, in the axioms this comment is referring to, both (a) and (b) would hold at the same time, because the axioms only have a single function symbol. In this case it doesn't matter besides the documentation aspect.

> Compared to trigger, I think the uses clause provides a clearer way to say that an axiom is only relevant for this declaration.

I think there are other ways to document such relations, such as using comments and bundling up related functions and axioms together, as done in the axiomatizations. Consider the example you gave, it currently makes no difference for pruning whether this axiom is inside a `uses` clause or not, so let's put it inside the `uses` clause for`$Box`:

```
function $Box(x: T) uses {
  axiom (forall<T> x : T :: { $Box(x) } $Unbox($Box(x)) == x);
}
```
Now let's update the trigger as you pointed out:

> Btw, should the trigger of this be updated to Unbox(Box(x)) ?

```
function $Box(x: T) uses {
  axiom (forall<T> x : T :: { $Unbox($Box(x)) } $Unbox($Box(x)) == x);
}
```
Ignoring that a `$Box`  might never appear without `$Unbox` in a program in practice, this will weaken pruning if `$Unbox` is not reachable: (a) holds and (b) doesn't. Having the axiom outside of any `uses` clauses would minimize the risk of this happening.

(Btw I am not sure if `$Unbox` should be added there. `$Box` would not be injective without this axiom. In any case, I think adjusting triggers is beyond the scope of this PR - this PR only aims to make unused axioms prunable.)

`Maybe we should replace uses clauses by definition axioms where possible`

I think this might make sense, does this happen when one provides a definition for a function and not provide the `:inline` attribute? One potential risk here is if `:inline` attribute is later added to such functions, which would break axioms where this function appears in a trigger.

> You can incorrectly use used_by just as well, can you not? I think one could argue for doing as much as possible based on the triggers so there's less possibility of specifying too many dependencies.

I agree, and I am starting to think it might be best if edge inference from triggers is preserved. The only correct way to use `used_by` might be to add exactly one tuple for each trigger of an axiom, which the current edge inference is already doing. Decoupling the semantics would make it possible to strengthen pruning, but am having difficulty seeing any cases where this would be useful. 

One potential application might be unquantified axioms where edge inference does not apply, and `used_by` allows for finer control over pruning logic than `uses` does, but I didn't see any cases in the prelude where this would benefit.

Test/git-issues/git-issue-3855.dfy
zafer-esen: I just noticed that commenting out those assertions led to less resource counts, putting it below the stated limit. I did not really check how stable it is currently with different seeds though, I can experiment and delete those comments completely if it is stable.

PR #4430, opened 2023-08-17
Fix: Triggers warnings correclty converted into errors with --warn-as-errors

### LINKED ISSUE #3358: "No trigger covering all quantified variables found" is still a warning with `/warningsAsErrors`
### Dafny version

master (f6bd1b8e13250ec31d6db0400ce8a32a9560fcad)

```dafny
datatype MyResult = Ok | Err(error: set<string>)
function method warningTriggerTest(sr: seq<MyResult>): set<string> {
  set i, err | 0 <= i < |sr| && err in (if sr[i].Err? then sr[i].error else {}) :: err
}
```

```code
% ./dafny /compile:0 /warningsAsErrors warning-trigger.dfy
.../warning-trigger.dfy(3,2): Warning: /!\ No trigger covering all quantified variables found.

```

### What happened?

I'd expect `/warningsAsErrors` to convert the warning to an error, or else the documentation should explain why it doesn't.

Linux

MikaelMayer:
This PR fixes #3358
I also verified that there was no other places where we were missing to convert warnings as errors by looking at all usages of "ErrorLevel.Warning".
Tests added


Source/DafnyCore/Triggers/QuantifiersCollection.cs
```diff
@@ -338,7 +339,7 @@ internal class QuantifierGroup {
           msg.Append(WARN).AppendLine("Suppressing loops would leave this expression without triggers.");
           FirstLetterCapitalOnNestedToken();
         } else if (suppressWarnings) {
-          errorLevel = ErrorLevel.Warning;
+          errorLevel = options.WarningsAsErrors ? ErrorLevel.Error : ErrorLevel.Warning;
```
keyboardDrummer, line 342: Isn't this the wrong location to check `options.WarningsAsErrors` ? Why wouldn't the reporter do that check? (as it already mostly does, but only when you call `reporter.Warning`, not when you call `reporter.Message`

MikaelMayer: Agreed. Implemented.

Source/DafnyCore/Triggers/QuantifiersCollection.cs
```diff
@@ -347,7 +347,11 @@ internal class QuantifierGroup {
 
       if (msg.Length > 0 && !Attributes.Contains(q.quantifier.Attributes, "auto_generated")) {
         var msgStr = msg.ToString().TrimEnd("\r\n ".ToCharArray());
-        reporter.Message(MessageSource.Rewriter, errorLevel, null, reportingToken, msgStr);
```
keyboardDrummer, line 350: Shouldn't we fix `reporter.Message` so it can't be used incorrectly?

keyboardDrummer (approved):

keyboardDrummer (approved):

keyboardDrummer (dismissed):

PR #4436, opened 2023-08-17
Multi-file support for  legacy symbol table

keyboardDrummer:
### Changes

1. Rename `LegacySymbolTable`, which is part of the non-LSP Dafny server, to `SuperLegacySymbolTable`
1. Rename `SignatureAndCompletionTable` to `LegacySignatureAndCompletionTable`
1. Let `SignatureAndCompletionTable` differentiate between positions in different files, so it doesn't get confused between symbols in different files when they have similar positions or ranges.
1. Only migrate state in `SignatureAndCompletionTable` that relates to the changed file

### Testing
- There could be a test added for 3 but I didn't add it.
- Merging this PR into this other PR (https://github.com/dafny-lang/dafny/pull/4419), reduces the time of the test `ManyFastEditsUsingLargeFilesAndIncludes` from 47 seconds to 8 seconds

### DIFF
--- a/Source/DafnyLanguageServer/Workspace/Migrator.cs
+++ b/Source/DafnyLanguageServer/Workspace/Migrator.cs
@@ -18,13 +18,13 @@ public class Migrator {
   private readonly ILogger<Migrator> logger;
   private readonly DidChangeTextDocumentParams changeParams;
   private readonly CancellationToken cancellationToken;
-  private readonly ILogger<SignatureAndCompletionTable> loggerSymbolTable;
+  private readonly ILogger<LegacySignatureAndCompletionTable> loggerSymbolTable;
 
   private readonly Dictionary<TextDocumentContentChangeEvent, Position> getPositionAtEndOfAppliedChangeCache = new();
 
   public Migrator(
     ILogger<Migrator> logger,
-    ILogger<SignatureAndCompletionTable> loggerSymbolTable,
+    ILogger<LegacySignatureAndCompletionTable> loggerSymbolTable,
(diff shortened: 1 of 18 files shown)


Source/DafnyLanguageServer/Workspace/Migrator.cs
```diff
@@ -112,26 +112,45 @@ CancellationToken cancellationToken
...
-      migratedDeclarations = ApplyDeclarationsChange(originalSymbolTable, migratedDeclarations, change, GetPositionAtEndOfAppliedChange(change));
+
+      if (migratedLocationsForUri != null) {
+        migratedLocationsForUri = ApplyDeclarationsChange(originalSymbolTable, migratedLocationsForUri, change, GetPositionAtEndOfAppliedChange(change));
+      }
+    }
+    if (migratedLookupTreeForUri != null) {
+      logger.LogTrace("migrated the lookup tree, lookup before={SymbolsBefore}, after={SymbolsAfter}",
+        originalSymbolTable.LookupTreePerUri.Count, migratedLookupTreeForUri.Count);
+      // TODO IntervalTree goes out of sync after any change and "fixes" its state upon the first query. Replace it with another implementation that can be queried without potential side-effects.
```
MikaelMayer, line 136: This TODO has been there for a long time. Any idea on how to solve it?

MikaelMayer (approved):
I'm very happy to see such a performance improvement change with only 100 LoC of changes. Thanks for doing it despite it not being obvious at first.

PR #4438, opened 2023-08-17
Implement unbounded ints/rationals in the Rust backend

shadaj:
Implement unbounded ints/rationals in the Rust backend

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT
license](https://github.com/dafny-lang/dafny/blob/master/LICENSE.txt).</small>

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/dafny-lang/dafny/pull/4438).
* #4450
* __->__ #4438


keyboardDrummer (commented):
One question that I have, which doesn't necessarily relate to this PR, but to the approach taken in `Compiler-dafny.cs`, is whether you've considered not just using `ConcreteSyntaxTree` instead of `string`, but to immediately generate the desired AST nodes.

I made a PR with a sort of prototype of that here: https://github.com/dafny-lang/dafny/pull/3396

The core idea is that `SinglePassCompiler` is made generic:

```c#
public abstract class GenericSinglePassCompiler<TExpression, TStatements, TStatement, TType, etc...>
```

and then the existing backends that want to use `ConcreteSyntaxTree`/`ICanRender` will inherit from a new type `ConcreteSinglePassCompiler`:

```c#
public abstract class ConcreteSinglePassCompiler : GenericSinglePassCompiler<ICanRender, ConcreteSyntaxTree, ICanRender, ICanRender>, Plugins.Compiler
```

shadaj:
@keyboardDrummer yeah, making `SinglePassCompiler` generic wrt `ConcreteSyntaxTree` would be quite nice. Even for our approach, there are situations where the compiler internally creates new `ConcreteSyntaxTree`s that cause trouble, so many of the refactoring steps have been to eliminate that. Would also be nice to eliminate assumptions about parenthesizing expressions, or how function calls are generated (right now the compiler unnecessarily assumes an order of receiver, then name, then args).

It's effectively what the Dafny-implemented backends look like, since there we have similar methods to generate statements/expressions/etc but the implementation is free to generate strings or its own representations of the target syntax.

Source/AutoExtern.Test/coverlet.runsettings
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<RunSettings>
+  <DataCollectionRunSettings>
+    <DataCollectors>
+      <DataCollector friendlyName="XPlat Code Coverage">
+        <Configuration>
+          <Exclude>[*]DAST.*,[*]DCOMP.*</Exclude>
```
shadaj, line 7: Because of the code blowup in the Dafny-to-C# compilation, instrumentation hangs and causes the CI to timeout. Since we ignore coverage of generated code anyways, this is safe.

PR #4439, opened 2023-08-18
Parser cancellation

keyboardDrummer:
### Changes
- Enable cancelling of parsing at the token granularity level, instead of only at the file level
- Do not open file streams used for parsing until actually parsing

### Testing
We don't have a good way of automatically testing such a performance change right now, but in one manual test run I saw runtime go down from 35 to 9s.

### DIFF
--- a/Source/DafnyDriver/DafnyDoc.cs
+++ b/Source/DafnyDriver/DafnyDoc.cs
@@ -50,6 +50,8 @@ class DafnyDoc {
       outputdir = DefaultOutputDir;
     }
 
+
+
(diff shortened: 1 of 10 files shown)


Source/DafnyDriver/DafnyDoc.cs
```diff
@@ -50,6 +50,8 @@ class DafnyDoc {
       outputdir = DefaultOutputDir;
     }
 
+
+
```
MikaelMayer, line 54:
```suggestion
```

MikaelMayer (approved):
I understand it makes the granularity at the file level, right? Approved.

PR #4440, opened 2023-08-18
feat: Read clauses on methods

robin-aws:
Implements https://github.com/dafny-lang/rfcs/pull/6.

This functionality is guarded by a new `--reads-clauses-on-methods` option, even though it's almost completely backwards compatible: the only wrinkle is function-by-methods, where valid code today might be rejected when the reads clause of the function definition is applied to the by-method body.

Note that "methods" here is interpreted as it is in the Dafny reference manual and similar documentation: it includes constructors and ghost methods, but NOT lemmas.

The `{:concurrent}` attribute, which previously only existed to generate an auditor warning that Dafny could not verify this property, now creates assertions that the `reads` and `modifies` clauses on the function or method are empty.

The core implementation strategy is relatively straightforward: enabling reads checks (via the `WFOptions` value) in the well-formedness check for statements as well as expressions. The existing `$_Frame` variable is now split into `$_ReadsFrame` and `$_ModifiesFrame`, since we now need both at once in method contexts. To help make reads checks optional based on the new option, and to optimize by not enabling them when in a `reads *` context (the default for methods), the `ExpressionTranslator.readsFrame` field may be null. I performed some mild refactoring to make this cleaner but open to suggestion on further improvements.

Documentation changes preview at https://robin-aws.github.io/dafny/.


keyboardDrummer (approved):
Whoa, so few changes are needed. Nice!

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -125,6 +125,10 @@ public class DafnyOptions : Bpl.CommandLineOptions {
       }
 
       RegisterLegacyUi(CommonOptionBag.AddCompileSuffix, ParseBoolean, "Compilation options", "compileSuffix");
+
+      RegisterLegacyUi(CommonOptionBag.ReadsClausesOnMethods, ParseBoolean, "Language feature selection", "readsClausesOnMethods", @"
+0 (default) - Reads clauses on methods are forbidden.
+1 - Reads clauses on methods are permitted.".TrimStart(), defaultValue: false);
```
keyboardDrummer, line 131: This is good, but it might be nice to also mention the `reads *` default, since the value/verbosity ratio is high.

Source/DafnyCore/Rewriters/ExpectContracts.cs
```diff
@@ -204,9 +204,17 @@ public class ExpectContracts : IRewriter {
     var post = new AttributedExpression(new BinaryExpr(tok, BinaryExpr.Opcode.Eq, r, fn) {
       Type = Type.Bool
     });
+    // If f.Reads is empty, replace it with an explicit `reads {}` so that we don't replace that
```
keyboardDrummer, line 207: I think this code becomes obsolete by assigning the default `reads *` during parsing

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -665,27 +674,51 @@ public partial class Translator {
...
-          CheckSubrange(e.tok, etran.TrExpr(e), e.Type, formal.Type, builder);
+        var readsCheckDelayer = new ReadsCheckDelayer(etran, null, localVariables, builderInitializationArea, builder);
 
-          if (formal.IsOld) {
-            Boogie.Expr wh = GetWhereClause(e.tok, etran.TrExpr(e), e.Type, etran.Old, ISALLOC, true);
-            if (wh != null) {
-              var desc = new PODesc.IsAllocated("default value", "in the two-state lemma's previous state");
-              builder.Add(Assert(e.tok, wh, desc));
+        // check well-formedness of any default-value expressions (before assuming preconditions)
+        readsCheckDelayer.WithDelayedReadsChecks(true, wfo => {
```
keyboardDrummer, line 680: AFAIK the prefix `With` is commonly used for pure methods, while this seems to be a side-effecting one. I suggest replacing `With` with `Add`.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -1334,7 +1370,58 @@ public partial class Translator {
       builder.Add(Assert(original.Tok, decrChk, new PODesc.TraitDecreases(original.WhatKind)));
     }
 
-    private void AddMethodOverrideSubsetChk(Method m, BoogieStmtListBuilder builder, ExpressionTranslator etran, List<Variable> localVariables,
+    // TODO: Worth sharing code between this and AddMethodOverrideModifiesSubsetChk?
```
keyboardDrummer, line 1373: TODO

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -1334,7 +1370,58 @@ public partial class Translator {
...
+    private void AddMethodOverrideReadsSubsetChk(Method m, BoogieStmtListBuilder builder, ExpressionTranslator etran, List<Variable> localVariables,
+      Dictionary<IVariable, Expression> substMap,
+      Dictionary<TypeParameter, Type> typeMap) {
+      //getting framePrime
```
keyboardDrummer, line 1377: I don't understand this comment.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -1334,7 +1370,58 @@ public partial class Translator {
       builder.Add(Assert(original.Tok, decrChk, new PODesc.TraitDecreases(original.WhatKind)));
     }
 
-    private void AddMethodOverrideSubsetChk(Method m, BoogieStmtListBuilder builder, ExpressionTranslator etran, List<Variable> localVariables,
+    // TODO: Worth sharing code between this and AddMethodOverrideModifiesSubsetChk?
+    private void AddMethodOverrideReadsSubsetChk(Method m, BoogieStmtListBuilder builder, ExpressionTranslator etran, List<Variable> localVariables,
```
keyboardDrummer, line 1374: `m` => `subject`/`target`/`method`

keyboardDrummer, line 1374: Could you clean up the code in the method `AddMethodOverrideReadsSubsetChk` a bit? It shows up as new in this PR but I see it was existing code.

Source/DafnyCore/AST/Members/Method.cs
```diff
@@ -213,6 +222,10 @@ public class Method : MemberDecl, TypeParameter.ParentType,
       formatter.SetAttributedExpressionIndentation(req, indentBefore + formatter.SpaceTab);
     }
 
+    foreach (var read in Reads) {
+      formatter.SetFrameExpressionIndentation(read, indentBefore + formatter.SpaceTab);
```
MikaelMayer, line 226: Good job formatting :-)

Source/DafnyCore/AST/Members/Method.cs
```diff
@@ -270,6 +283,29 @@ public class Method : MemberDecl, TypeParameter.ParentType,
...
+      var readsClausesOnMethodsEnabled = resolver.Options.Get(CommonOptionBag.ReadsClausesOnMethods);
+      if (readsClausesOnMethodsEnabled && !Reads.Any() && !IsLemmaLike) {
+        // Note that `reads *` is the right default for backwards-compatibility,
+        // but we may want to infer a sensible default like decreases clauses instead in the future.
+        Reads.Add(new FrameExpression(tok, new WildcardExpr(tok), null));
+      }
+      foreach (FrameExpression fe in Reads) {
+        resolver.ResolveFrameExpressionTopLevel(fe, FrameExpressionUse.Reads, this);
+        if (IsLemmaLike) {
+          resolver.reporter.Error(MessageSource.Resolver, fe.tok, "{0}s are not allowed to have reads clauses",
```
MikaelMayer, line 299:
```suggestion
          resolver.reporter.Error(MessageSource.Resolver, fe.tok, "{0}s do not have reads clauses but they can read anything they need",
```

Source/DafnyCore/DafnyOptions.cs
MikaelMayer: Yes I agree on that

Source/DafnyCore/Verifier/ProofObligationDescription.cs
```diff
@@ -1107,6 +1107,22 @@ public class AssignmentShrinks : ProofObligationDescriptionWithNoExpr {
...
 }
 
+public class ConcurrentFrameEmpty : ProofObligationDescriptionWithNoExpr {
+  public override string SuccessDescription =>
+    $"{frameName} is empty ({{:concurrent}} restriction)";
+
+  public override string FailureDescription =>
+    $"{frameName} might not be empty ({{:concurrent}} restriction)";
+
+  public override string ShortDescription => "assignment shrinks";
```
MikaelMayer, line 1117: Assignment shrinks? :-)

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
MikaelMayer: Good observation, it makes sense

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -535,12 +535,21 @@ public partial class Translator {
...
 
-      BoogieStmtListBuilder builder = new BoogieStmtListBuilder(this, options);
+      var builder = new BoogieStmtListBuilder(this, options);
+      var builderInitializationArea = new BoogieStmtListBuilder(this, options);
       builder.Add(new CommentCmd("AddMethodImpl: " + m + ", " + proc));
       var etran = new ExpressionTranslator(this, predef, m.tok);
+      // Only do reads checks for methods, not lemmas
+      // (which aren't allowed to declare frames and don't check reads and writes against them).
+      // Also don't do any reads checks if the reads clause is *,
+      // since all the checks will be vacuously true
```
MikaelMayer, line 545:
```suggestion
      // since all the checks will be trivially true
```

Vacuously = True because false is assumed
Trivially = True because it's an axiom that something belongs to everything

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -1334,7 +1370,58 @@ public partial class Translator {
...
+    // TODO: Worth sharing code between this and AddMethodOverrideModifiesSubsetChk?
+    private void AddMethodOverrideReadsSubsetChk(Method m, BoogieStmtListBuilder builder, ExpressionTranslator etran, List<Variable> localVariables,
+      Dictionary<IVariable, Expression> substMap,
+      Dictionary<TypeParameter, Type> typeMap) {
+      //getting framePrime
+      List<FrameExpression> traitFrameExps = new List<FrameExpression>();
+      List<FrameExpression> classFrameExps = m.Reads ?? new List<FrameExpression>();
+      if (m.OverriddenMethod.Reads != null) {
+        if (m.OverriddenMethod.Reads.Any(e => e.E is WildcardExpr)) {
+          // Vacuously true
```
MikaelMayer, line 1382:
```suggestion
          // Trivially true
```

MikaelMayer (commented):
Initial review, will continue later

Source/DafnyCore/Resolver/ModuleResolver.cs
```diff
@@ -792,9 +792,25 @@ public partial class ModuleResolver {
         new ActualBindings(f.Formals.ConvertAll(Expression.CreateIdentExpr)).ArgumentBindings,
         tok);
       var post = new AttributedExpression(new BinaryExpr(tok, BinaryExpr.Opcode.Eq, r, fn));
+      List<FrameExpression> reads;
+      if (Options.Get(CommonOptionBag.ReadsClausesOnMethods)) {
```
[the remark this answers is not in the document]
robin-aws: That would make it hard/impossible to reject explicit reads clauses if the option is not turned on though, since you couldn't distinguish between code that explicitly includes `reads *`.

Source/DafnyCore/AST/Members/Method.cs
```diff
@@ -22,6 +21,7 @@ public class Method : MemberDecl, TypeParameter.ParentType,
   public readonly List<Formal> Ins;
   public readonly List<Formal> Outs;
   public readonly List<AttributedExpression> Req;
+  public readonly List<FrameExpression> Reads;
```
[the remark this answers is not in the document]
robin-aws: https://xeiaso.net/blog/anything-message-queue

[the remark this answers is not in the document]
robin-aws: Agreed, but I couldn't see an easy way without a fair bit of refactoring.

Source/DafnyCore/AST/Members/Method.cs
robin-aws: Good idea, although I'm going with "(they are allowed to read all memory locations)" instead.

Source/DafnyCore/DafnyOptions.cs
robin-aws: Good call, I added this to the new CLI help string too.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -4343,48 +4385,56 @@ public enum StmtType { NONE, ASSERT, ASSUME };
       }
       builder.AddCaptureState(f.tok, false, "initial state");
 
-      DefineFrame(f.tok, f.Reads, builder, locals, null);
+      DefineFrame(f.tok, etran.ReadsFrame(f.tok), f.Reads, builder, locals, null);
       InitializeFuelConstant(f.tok, builder, etran);
 
+      // TODO: is this valid here given we haven't checked the well-formedness of the reads clauses yet?
```
MikaelMayer, line 4391: Please move this check after the well-formedness indeed so that the assertion won't implicitly assume well-formedness.

For example:
```
reads if a[0] then {} else {}
```

checking that the above is empty will automatically assume that a != null, which could be a soundness issue.

Source/DafnyDriver/DafnyDoc.cs
```diff
@@ -817,6 +817,11 @@ class DafnyDoc {
         details.Append(space4).Append(Keyword("requires")).Append(" ").Append(Code(req.E.ToString())).Append(br).Append(eol);
         some = true;
       }
+      if (m.Reads.Count > 0) {
+        var list = String.Join(", ", m.Reads.Select(e => Code(e.OriginalExpression.ToString() + (e.FieldName != null ? "`" + e.FieldName : ""))));
```
MikaelMayer, line 821: It feels like you can avoid some code duplication by creating at least a local function to render both reads and modifies clause, i.e. to extract `.Select(e => Code(e.OriginalExpression.ToString() + (e.FieldName != null ? "`" + e.FieldName : ""))));` both for the modifies and reads clause.
But as it's the doc tool, don't worry about that too much.

Source/DafnyCore/Dafny.atg
```diff
@@ -2072,15 +2073,17 @@ InvariantClause<. List<AttributedExpression> invariants.> =
...
             List<Expression> decreases, ref Attributes decAttrs, ref Attributes modAttrs, string caption, bool performThisDeprecatedCheck.>
 = (. Contract.Requires(cce.NonNullElements(req));
+     Contract.Requires(cce.NonNullElements(reads));
      Contract.Requires(cce.NonNullElements(mod));
      Contract.Requires(cce.NonNullElements(ens));
      Contract.Requires(cce.NonNullElements(decreases));
   .)
   SYNC
-  { ModifiesClause<ref mod, ref modAttrs, false, performThisDeprecatedCheck>
+  { ReadsClause<reads, false, false, true>
```
MikaelMayer, line 2085: It feeks weird that you parse reads clauses on method even if the flag --reads-clauses-on-method is set to false. Since we now have project files, I think you should at least emit a warning if you parse reads clauses but they are not verified.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
robin-aws: Fair point, although I do want to get across the idea of "execute this with this additional context/before and after effect". Perhaps `RunWithDelayedReadsChecks` (or is there a more idiomatic verb for executing an `Action<...>`?)

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
robin-aws: Yup I'll first seem if I can merge the two, and either way clean them up.

Test/dafny0/ConcurrentAttribute.dfy
```diff
@@ -0,0 +1,56 @@
...
+  method {:concurrent} GoodM(b: Box<int>) {
+  }
+
+  method {:concurrent} BadM(b: Box<int>)  // Error: reads clause might not be empty ({:concurrent} restriction)
+    reads b
+  {
+    var x := b.x;
+  }
+
+  method {:concurrent} WeirdButOkayM(b: Box<int>)
```
MikaelMayer, line 41: Please add a test where the reads clause has an empty object set but a designed field like reads {}`f (something that compiles of course)
Also, at least for future change detection, please add a test with a reads a `f and another with modifies a`f where f is a ghost field. You decided that it's not concurrent safe for this PR, but since I exhibited a sound concurrent execution model where ghost fields can be read and modified independently by threads, we should make sure we differentiate against that.

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
```diff
@@ -1334,7 +1370,58 @@ public partial class Translator {
...
+      IToken tok = m.tok;
+      // Declare a local variable $_Frame: <alpha>[ref, Field alpha]bool
+      // The next line is a throw-away expression, used only to extract the type and name of the $_ModifiesFrame variable
+      Boogie.IdentifierExpr traitFrame = etran.ReadsFrame(m.OverriddenMethod.tok);
+      traitFrame.Name = m.EnclosingClass.Name + "_" + traitFrame.Name;
+      Contract.Assert(traitFrame.Type != null);  // follows from the postcondition of ReadsFrame
+      Boogie.LocalVariable frame = new Boogie.LocalVariable(tok, new Boogie.TypedIdent(tok, null ?? traitFrame.Name, traitFrame.Type));
+      localVariables.Add(frame);
+      // $_Frame := (lambda<alpha> $o: ref, $f: Field alpha :: $o != null && $Heap[$o,alloc] ==> ($o,$f) in Reads-Clause);
+      Boogie.TypeVariable alpha = new Boogie.TypeVariable(tok, "alpha");
```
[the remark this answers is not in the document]
robin-aws: Again copied without cleaning up, will do if it makes sense (prefer to merge these two instead)

Test/dafny0/ConcurrentAttribute.dfy.expect
```diff
@@ -0,0 +1,5 @@
+ConcurrentAttribute.dfy(20,25): Error: reads clause might not be empty ({:concurrent} restriction)
```
MikaelMayer, line 1:
```suggestion
ConcurrentAttribute.dfy(20,25): Error: reads clause could not be proved to be empty ({:concurrent} restriction)
```
Several new PRs changed the wording might not => could not prove, so that the user knows what's the responsibility of Dafny: to prove things, not provide sound counter-examples.

Test/dafny0/ReadsOnMethods-Resolution.dfy.expect
```diff
@@ -0,0 +1,10 @@
+ReadsOnMethods-Resolution.dfy(7,28): Error: reads clauses on methods are forbidden without the --reads-clauses-on-methods option
```
MikaelMayer, line 1: Oh I thought it would be  parser warning. Forget my other comment then. An error during resolution is fine.
```suggestion
ReadsOnMethods-Resolution.dfy(7,28): Error: reads clauses on methods are forbidden without the command-line flag `--reads-clauses-on-methods`
```

MikaelMayer (changes requested):
I think that after my comments it's ready for approval!
Good job.

Test/dafny0/ReadsOnMethods.dfy
```diff
@@ -0,0 +1,513 @@
...
+  var myBox := new Box(42);
+  var x := GetBoxFn(myBox);
+  SetBox(myBox, 42);
+}
+
+// ---------- More complicated reads clauses -----------
+
+method ApplyLambda<T(!new), R>(f: T ~> R, t: T) returns (r: R) 
+  requires f.requires(t)
+  reads f.reads
```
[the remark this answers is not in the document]
robin-aws: It's still documented and working, and I didn't see any reason for method reads clauses to be any different.

Test/dafny0/ReadsOnMethods.dfy
```diff
@@ -0,0 +1,513 @@
...
+
+// And now for brand new freshly-baked tests!
+
+// ---------- Basic use cases -----------
+
+class Box<T> {
+  var x: T
+
+  constructor(x: T)
+    reads {}
```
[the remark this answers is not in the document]
robin-aws: Well `this` is fresh both before and after the `new;`.

But also note `this.x` is a LHS so it doesn't count as a read anyway.

Source/DafnyCore/Verifier/Translator.cs
robin-aws: I can't seem to create a test case that fails, and I suspect the ordering doesn't actually matter. But it does feel safer to reorder them so I'll still make the change.

Source/DafnyCore/Dafny.atg
robin-aws: They are already a hard error, see `ReadsOnMethods-Resolution.dfy`

Source/DafnyDriver/DafnyDoc.cs
robin-aws: Yeah there are small bits of duplication in this that weren't quite painful to bother with. But I am going to try to improve the larger bits.

docs/DafnyRef/Specifications.md
```diff
@@ -479,19 +485,32 @@ able to give invariants to preserve it in this case, it gets even more
 complex when manipulating data structures. In this case, framing is
 essential to making the verification process feasible.
 
-It is not just the body of a function that is subject to `reads`
+By default, methods are not required to list the things they read.
+However, there are use cases for restricting what memory methods can read as well.
```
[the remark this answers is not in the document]
robin-aws: Made it "what memory locations methods are..." instead

docs/DafnyRef/Specifications.md
```diff
@@ -479,19 +485,32 @@ able to give invariants to preserve it in this case, it gets even more
 complex when manipulating data structures. In this case, framing is
 essential to making the verification process feasible.
 
-It is not just the body of a function that is subject to `reads`
+By default, methods are not required to list the things they read.
```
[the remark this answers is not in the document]
robin-aws: Made this memory locations - "object or fields" felt a bit weird to me since you can only ever read object fields rather than objects directly, and the latter could imply things like reading the address of an object.

docs/DafnyRef/Specifications.md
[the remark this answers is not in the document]
robin-aws: Actually, went with your edit after all given the comment above.

Test/dafny0/ReadsOnMethods.dfy
```diff
@@ -0,0 +1,513 @@
...
+
+method ApplyLambda<T(!new), R>(f: T ~> R, t: T) returns (r: R) 
+  requires f.requires(t)
+  reads f.reads
+{
+  r := f(t);
+}
+
+method DependsOnAllocationState<T>(b: Box<T>) 
+  // TODO: Allowed but perhaps shouldn't be, since it isn't allowed on functions
```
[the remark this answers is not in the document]
robin-aws: Resolved this with a more firm decision and documented it on the test case

Source/DafnyCore/Verifier/Translator.ClassMembers.cs
robin-aws: Went with "DoWith.." :)

Source/DafnyCore/Resolver/ModuleResolver.cs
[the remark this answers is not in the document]
keyboardDrummer: Wouldn't you reject that code as part of parsing as well ?

Test/dafny0/ConcurrentAttribute.dfy.expect
```diff
@@ -1,5 +1,16 @@
...
-ConcurrentAttribute.dfy(35,23): Error: reads clause might not be empty ({:concurrent} restriction)
-ConcurrentAttribute.dfy(46,23): Error: modifies clause might not be empty ({:concurrent} restriction)
+ConcurrentAttribute.dfy(91,10): Warning: /!\ No terms found to trigger on.
+ConcurrentAttribute.dfy(92,13): Warning: /!\ No terms found to trigger on.
+ConcurrentAttribute.dfy(97,10): Warning: /!\ No terms found to trigger on.
+ConcurrentAttribute.dfy(98,13): Warning: /!\ No terms found to trigger on.
+ConcurrentAttribute.dfy(30,25): Error: reads clause could not be proved to be empty ({:concurrent} restriction)
+ConcurrentAttribute.dfy(48,23): Error: reads clause could not be proved to be empty ({:concurrent} restriction)
+ConcurrentAttribute.dfy(51,23): Error: reads clause could not be proved to be empty ({:concurrent} restriction)
+ConcurrentAttribute.dfy(55,23): Error: reads clause could not be proved to be empty ({:concurrent} restriction)
```
MikaelMayer, line 8: I see that you recently deleted {:concurrent} checks from the auditor and various places within the last 13 commits, however, you still have issues mentioning it. Can you please explain?

robin-aws: Ah which issues, I'm not sure I follow? I intended to delete all the references to `{:concurrent}` in the auditor context since I removed that case and have replaced it with verification obligations instead, but I could have missed some.

Source/DafnyCore/Resolver/ModuleResolver.cs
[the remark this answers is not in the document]
robin-aws: On revisiting this, yes, I think I could have assigned the default during parsing instead (with the same conditionals to avoiding doing that on lemmas etc.) I was wary of that at the time since I didn't see any other examples of adding default elements at parse time, and didn't want to risk more cases of errors that pointed to tokens that didn't exist in the original program, although in this particular case I don't think that's possible (since `reads *` can't cause errors AFAICT).

PR #4443, opened 2023-08-18
Fix: Top-level symbols are now all verified.

### LINKED ISSUE #4442: Methods with {:only} don't have the proper verification icon
### Dafny version

latest as of today

```dafny
method NotVerified() { // Should not be highlighted in green
  assert 1 == 0;
}

method {:only} Verified() { // Verified
  assert true;
}

method NotVerified2() { // Should not be highlighted in green
  assert 1 == 0;
}
```

```code
Paste in VSCode with the mode "Verification as tests"
```

### What happened?

image

I would have expected only the method with {:only} to be labelled as "verified", and the other two ignored.

Windows

MikaelMayer:
Fixes #4442
In https://github.com/dafny-lang/dafny/pull/4427, a bug was introduced which prevent multiple top-level verification symbols to be reported, only the first one would. This adds a test to cover the issue and fixes it. I verified that the test was failing before the fix, and now the test is passing. I am not adding a docs/dev/news entry because the bug was introduced after the 4.2.0 release.


Source/DafnyCore/Generic/NodeExtensions.cs
```diff
@@ -119,14 +119,13 @@ public static class NodeExtensions {
     }
 
     var newParent = new LList<INode>(node, parent);
+    var result = new List<LList<INode>>();
```
keyboardDrummer, line 122: On line 118, should it return `new LList<INode>[] {}` ?

Also, would `return node.Children.SelectMany(child => child.FindNodeChainsInUri(uri, newParent))` be more performant?

MikaelMayer: I will do the first two changes.
For the third, I love unit tests and I started to write some myself in DafnyCore, but I don't know exactly what this function does, why it does it, what I should require and ensure, and I don't have the bandwidth to investigate. At least, it's covered by a test in the language server.

keyboardDrummer (dismissed):

keyboardDrummer (approved):

PR #4444, opened 2023-08-19
Test Generation Auditor

Dargones:
This PR (number 4444:) introduces a test generation auditor that scans the program for anything that will cause test generation to fail (errors) or might lead to less helpful results (warnings). These are reported via `ConsoleErrorReporter` like Dafny would report any other problems. Dafny returns a non-zero exit code and terminates execution if any errors are present or if there are warnings and the `--warn-as-errors` flag is on. Here is the list of all errors/warnings that can be emitted. 

Errors:
- Cannot find a method or function annotated with {:testEntry} attribute
- Program is not wrapped in a module. Put your code inside \"module M {}\" or equivalent
- {:testInline} attribute on the method [METHOD] can only take one argument, which must be a positive integer specifying the recursion unrolling limit (absence of such an argument or 1 means no unrolling)
- Test Generation does not support abstract types, array types, and trait types as inputs. Consider disallowing values of type [TYPE] to be part of the input to {:testEntry}-annotated method [METHOD]
- Any error from parser or resolver is propagated up and reported as well

Warnings:
- Method [METHOD] will not be inlined because it is not reachable from any of the {:testEntry}-annotated methods or functions
- Test Generation does not fully support function types as inputs. Consider disallowing values of type [TYPE] to be part of the input to {:testEntry}-annotated method [METHOD]
- Cannot find witness for type [TYPE]. Please consider adding a witness to the declaration
- Method [METHOD] is annotated with {:timeLimit 300} but test generation is called with --verification-time-limit:20. Consider increasing the time limit for test generation.


Dargones:
As a side note, I noticed that several tests throughout the project use the `Utils.Parse` method from `DafnyTestGeneration` to parse the input. Is there perhaps a better place I should move it to?

Dargones:
Update: I brought the errors reporting mechanism in line with how it works in Dafny more generally - warnings are colored in yellow, errors in red, and -warn-as-errors flag treats warnings like errors

Source/DafnyCore/Generic/Reporting.cs
```diff
@@ -14,7 +14,7 @@ public enum ErrorLevel {
   }
 
   public enum MessageSource {
-    Parser, Cloner, RefinementTransformer, Rewriter, Resolver, Translator, Verifier, Compiler, Documentation
+    Parser, Cloner, RefinementTransformer, Rewriter, Resolver, Translator, Verifier, Compiler, Documentation, TestGenerationAuditor
```
atomb, line 17: I'm not sure that the `Auditor` term adds much here. I'd just call this `TestGeneration`. And the checks performed here are just the first phase of test generation.

Source/DafnyTestGeneration.Test/AuditorTests.cs
```diff
@@ -0,0 +1,266 @@
...
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Microsoft.Dafny;
+using Xunit;
+
+namespace DafnyTestGeneration.Test;
+
+public class AuditorTests : Setup {
```
atomb, line 16: I might call this `WarningTests` or something, to avoid confusion with the separate feature that uses the name "auditor".

atomb: It's great to have lots of tests of this, BTW. :)

Source/DafnyTestGeneration/Auditor.cs
```diff
@@ -0,0 +1,373 @@
...
+    NonZeroExitCode = Errors.Count() != 0;
+    return !Errors.Any() && (!Warnings.Any() || !options.WarningsAsErrors);
+  }
+
+  /// <summary>
+  /// Print all gathered warnings and errors
+  /// </summary>
+  private void PrintWarningsAndErrors(ErrorReporter errorReporter, Program program) {
+    if (Errors.Count() != 0) {
+      errorReporter.Message(MessageSource.TestGenerationAuditor, ErrorLevel.Error, "", program.StartToken,
```
atomb, line 72: Is there a reason you can't use the more concise `Error` method instead of the `Message` method?

Source/DafnyTestGeneration/Auditor.cs
```diff
@@ -0,0 +1,373 @@
...
+        $"Test generation auditor returned {Errors.Count()} errors. Fix them to proceed:");
+      foreach (var error in Errors) {
+        var errorId = error.ErrorId == "none" ? "Error" : error.ErrorId;
+        errorReporter.Message(MessageSource.TestGenerationAuditor, ErrorLevel.Error, errorId, error.Token,
+          error.Message);
+      }
+    }
+    if (Warnings.Count() != 0) {
+      errorReporter.Message(MessageSource.TestGenerationAuditor, ErrorLevel.Warning, "", program.StartToken,
```
atomb, line 81: Similarly to errors, can you use `Warning` here?

Source/DafnyTestGeneration/Auditor.cs
```diff
@@ -0,0 +1,373 @@
...
+      return true;
+    }
+    typesConsidered.Add(type);
+    var isSupported = false;
+    while (type is InferredTypeProxy inferred) {
+      type = inferred.T;
+    }
+    if (type is UserDefinedType userDefinedType) {
+      var genericMessage =
+        $"Consider disallowing values of type {userDefinedType} to be part of the input to " +
```
atomb, line 251: The word "disallowing" seems weird here. Maybe use terminology similar to what you use elsewhere about writing a wrapper to generate types that can't be constructed automatically out of types that can?

atomb (changes requested):
The implementation looks great, and I mostly had comments about terminology, both in messages and internal names. The key one is that I'm not sure that distinguishing this from the rest of the test generation code with the name "auditor" has a lot of purpose. It's just the first step of test generation, in my mind.

Source/DafnyTestGeneration.Test/AuditorTests.cs
Dargones: Done!

Source/DafnyCore/Generic/Reporting.cs
Dargones: This makes sense. Have made the change

Source/DafnyTestGeneration/Auditor.cs
Dargones: No, I think I just wasn't aware that can use Error/Message - have changed the code to use them now

Source/DafnyTestGeneration/Auditor.cs
Dargones: I have paraphrased this to "Consider modeling values of type TYPE with a datatype and passing them as input to the {:testEntry} annotated method". I agree that something along these lines should be better than "disallowing" since it tells the user what to do to prevent the error.

Dargones:
Thank you, @atomb! I have made all the changes and have also renamed `Auditor` to `FirstPass` to avoid confusion with the top-level `Auditor` functionality.

atomb (approved):
This looks great!

PR #4445, opened 2023-08-20
Documenting Test Generation in Reference Manual

Dargones:
This pull request documents the `generate-tests` command. 


docs/DafnyRef/UserGuide.md
```diff
@@ -563,27 +563,201 @@ The documentation for the root module is in `_.html`.
...
+```text
+program.dfy(5,9) is reachable.
+program.dfy(3,4):initialstate is reachable.
+program.dfy.dfy(5,9)#elseBranch is reachable.
+program.dfy.dfy(4,9)#thenBranch is potentially unreachable.
+Out of 4 basic blocks, 3 are reachable.
+```
+
+Dafny reports that the then branch of the condition is potentially unreachable because the verifier proves that no
+input can reach it. In this case, this is to be expected, since the product of two numbers can never be positive. In
```
atomb, line 747:
```suggestion
input can reach it. In this case, this is to be expected, since the product of two numbers can never be negative. In
```

atomb (changes requested):
I saw one small typo, but otherwise this looks great.

Dargones:
Thank you for catching this!

atomb (approved):

PR #4461, opened 2023-08-23
Basic implementation of proof dependency analysis

atomb:
This enables experimental support for tracking proof dependencies, currently not advertised and enabled only by using the Boogie `-trackVerificationCoverage` option.

When that option is enabled, the log produced by `/verificationLogger:text` or `--log-format text` now includes the proof dependencies required to prove each assertion batch, along with the list of assertions that were proved in that batch. Each dependency includes a source range rather than just a single location, which can identify cases where only a specific sub-expression was required for the proof. The tests includes examples of this.

Although the proper UI for proof dependencies isn't yet in place, a nice side effect of how this reports proof dependencies is that it's relatively easy to see vacuous proofs by looking for assertions that were proved but not included in the proof dependency list for their assertion batch. This will be automated in a future PR, but is already possible with a little manual work.


Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -125,6 +126,55 @@ public class TranslatorFlags {
...
+        var idString = $"id{proofDependencyIdCount}";
+        ProofDependenciesById[idString] = dep;
+        proofDependencyIdCount++;
+        return idString;
+      }
+
+      public void AddProofDependencyId(ICarriesAttributes boogieNode, IToken tok, ProofDependency dep) {
+        var idString = GetProofDependencyId(dep);
+        boogieNode.Attributes =
+          new QKeyValue(tok, "id", new List<object>() { idString }, boogieNode.Attributes);
```
keyboardDrummer, line 144: Why are we using a string `idString` instead of directly storing `dep` in the attribute, so you won't need the `ProofDependencyManager`? 

If the purpose is for the Boogie code to be printable and for a person to be able to map that print back to the original Dafny code, then I think you'll have to base the ids on Dafny source locations instead of generating them, and then you also wouldn't need a `ProofDependencyManager`.

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -125,6 +126,55 @@ public class TranslatorFlags {
...
+      }
+
+      // Get the full ProofDependency indicated by a compound ID string.
+      public ProofDependency GetFullIdDependency(string idString) {
+        var parts = idString.Split('$');
```
keyboardDrummer, line 149: I think string parsing should be reserved for working with the outside world, not for in-memory information processing.

Test/logger/ProofDependencyLogging.dfy
```diff
@@ -0,0 +1,424 @@
+// RUN: %baredafny verify --log-format:text --boogie -trackVerificationCoverage "%s" > "%t"
```
keyboardDrummer, line 1: I'm still a little puzzled on the UX.

One use-case that's clear to me is giving warnings whenever code is not used by any assertion batch. For that I wouldn't use this report, but instead turn on a warning `--warn-unused-by-verification`

Can you give an example of how how the information presented in this report would be used?

Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -125,6 +126,55 @@ public class TranslatorFlags {
     readonly Dictionary<string, Bpl.Constant> fieldConstants = new Dictionary<string, Constant>();
     readonly Dictionary<string, Bpl.Constant> tytagConstants = new Dictionary<string, Constant>();
 
+    public class ProofDependencyManager {
```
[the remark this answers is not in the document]
atomb: It was originally just a couple of fields that I ended up encapsulating. I think it makes sense to move it out of the translator class, and even into a separate file.

Source/DafnyCore/Verifier/ProofDependency.cs
```diff
@@ -0,0 +1,182 @@
...
+
+  private string comment;
+
+  public AssumptionDependency(string comment, Expression expr) {
+    this.comment = comment;
+    this.expr = expr;
+  }
+}
+
+public class InvariantDependency : ProofDependency {
```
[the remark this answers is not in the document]
atomb: This one, for invariant dependencies, is used in only one place (at the moment -- that could change). But most of the rest of the `ProofDependency` implementations are used in multiple locations. It feels more comprehensible to me to have them all in the same place.

Source/DafnyCore/Verifier/Translator.cs
atomb: The reason for this is that Boogie needs to have a string to send to Z3 -- that's the only way to specify the label on an expression you want to track for proof dependencies. Then Z3 returns a set of strings indicating which pieces it used. The notion of proof dependency that we have in Dafny is Dafny-centric, so it wouldn't make sense to have the `ProofDependency` class in Boogie. It's possible that we could have an additional layer of abstraction in the Boogie implementation, so we'd have 1) names as strings, 2) structured Boogie conceptions of those strings, and 3) structured Dafny conceptions that build on the Boogie ones. But I'm not sure that that additional layer is really worth it. Ultimately, the parsing we do needs to happen _somewhere_. It's just a question of whether it's in Boogie or Dafny.

Test/logger/ProofDependencyLogging.dfy
atomb: Yeah, this isn't really meant to be something that would be used by anyone other than the Dafny developers in its current state. It could be, but would require some expertise and manual effort. I'm building a more useful set of UI mechanisms on top of it, but I want to merge this now to get things done in pieces so there's not one monster PR to review. :) Think of it similarly to the type system refresh @RustanLeino  is doing.

Source/DafnyCore/Verifier/Translator.BoogieFactory.cs
```diff
@@ -197,6 +198,17 @@ enum BuiltinFunction {
       return attributes == null ? new Bpl.AssumeCmd(tok, expr) : new Bpl.AssumeCmd(tok, expr, attributes);
     }
 
+    public Bpl.AssumeCmd TrAssumeCmdWithDependencies(ExpressionTranslator etran, Bpl.IToken tok, Expression dafnyExpr, string comment = null, Bpl.QKeyValue attributes = null) {
+      return TrAssumeCmdWithDependenciesApp(etran, tok, dafnyExpr, e => e, comment, attributes);
+    }
+
+    public Bpl.AssumeCmd TrAssumeCmdWithDependenciesApp(ExpressionTranslator etran, Bpl.IToken tok, Expression dafnyExpr, Func<Bpl.Expr, Bpl.Expr> func, string comment = null, Bpl.QKeyValue attributes = null) {
```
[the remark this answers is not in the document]
atomb: I admit that this is a bit awkward. I'm not entirely happy with it, but I've tried a number of refactorings of it, including essentially what you suggest, without coming up with something that seems better. I'll give it another try, though.

Source/DafnyCore/Verifier/Translator.cs
keyboardDrummer: > The reason for this is that Boogie needs to have a string to send to Z3

Why not do the id generation with AST node mapping in Boogie? Seems better to let Boogie encapsulate SMT generation. Then you also don't need the method `GetFullIdDependency` in Dafny and references to the coverage related string constants. `VCResult.coveredElements` would then have type `IEnumerable<Absy>` instead of `IEnumerable<string>`.

Source/DafnyCore/Verifier/ProofDependency.cs
```diff
@@ -0,0 +1,209 @@
...
+namespace DafnyCore.Verifier;
+
+// A proof dependency represents a particular part of a Dafny program
+// that may be used in the process of proving its correctness. When
+// Boogie proves a particular verification condition, it returns a
+// list of strings, returned by the SMT solver, indicating which
+// program elements were used in completing the proof. After this, the
+// ProofDependencyManager can map each string to a more structured
+// ProofDependency.
+public abstract class ProofDependency {
```
keyboardDrummer, line 19: An instance of `ProofDependency` seems to be a description of an assumption or assertion in Boogie, together with a token. It serves the purpose of a source location, to tell the user where this assumption or assertion came from, even in cases where it originated from multiple locations, like `CallRequiresDependency`.

I think it would make more sense to use the field `Boogie.IToken tok` in `Absy` to store the 'origin' of a Boogie AST node, including textual description like `requires clause at` if need be.

To a large extend this is already done. `NestedToken` can store multiple locations in a single IToken, and has a `message` field that further explains what this location means.

Here's an example where `NestedToken` used to provide information similar to `CallRequiresDependency`:

```
              // check that the preconditions for the call hold
              foreach (AttributedExpression p in e.Function.Req) {
                Expression precond = Substitute(p.E, e.Receiver, substMap, e.GetTypeArgumentSubstitutions());
                var (errorMessage, successMessage) = CustomErrorMessage(p.Attributes);
                foreach (var ss in TrSplitExpr(precond, etran, true, out var splitHappened)) {
                  if (ss.IsChecked) {
                    var tok = new NestedToken(GetToken(expr), ss.Tok);
```

By using an IToken instead of a ProofDependency to describe an assume/assert, the description is not tied to usage tracking, but can also be used for error reporting, like reporting "for the call at location X, asserting the requires clause at location Y failed".

To summarize, I feel there's duplication between `ProofDependency` and `IToken`. What you could do is use the current `Absy.tok` field to generate the descriptions for the assumes/asserts, and leave tweaking the description of `Absy.tok` for later.

Test/logger/ProofDependencyLogging.dfy
keyboardDrummer: Okay, that works.

Source/DafnyCore/Verifier/Translator.cs
atomb: Although on the surface it seems like an `Absy` would encapsulate everything that we need, in practice this isn't the case because of how Boogie transforms the original program and how particular nodes are used in different ways in a proof (such as how a requires clause is an assumption when proving a definition itself, but a proof goal when invoking that definition). So, we'd need something like `ProofDependency` in Boogie, but it couldn't be quite the same thing, because those program constructs don't map cleanly back from Boogie to Dafny. So we'd need both a BoogeProofDependency and a DafnyProofDependency, and I don't think that extra layer of indirection is worthwhile at the moment.

Source/DafnyCore/Verifier/ProofDependency.cs
atomb: Although I can see what you're saying about the duplication, to me it feels like it would be less clear to use an `IToken`. I've always found the nested token notion somewhat awkward, because it doesn't clearly describe the relationship between the outer and inner token. This leads to things like "Related location" in error messages. Those related tokens can have descriptions themselves, but it leads to sort of convoluted messages. I'd instead be inclined to move away from using `NestedToken` (although that would be too substantial an effort to be worthwhile).

Source/DafnyCore/Verifier/Translator.cs
keyboardDrummer: > Although on the surface it seems like an Absy would encapsulate everything that we need, in practice this isn't the case because of how Boogie transforms the original program and how particular nodes are used in different ways in a proof (such as how a requires clause is an assumption when proving a definition itself, but a proof goal when invoking that definition)

I don't see that, since Dafny attaches each id to a unique `Absy` instance. Regarding your example, there a separate `Absy` for each pair of call and requires clause, and a separate one for each requires clause.

---

This PR lets Dafny attach ids to `Absy`s, while making an id -> `proofDependency` mapping, then it takes the ids returned by Boogie that gets them from the unsat core, and maps them back to `proofDependency`s.

Alternatively, you can have Boogie attach ids to the `Absy`s, map the ids coming out of the unsat core back to `Absy`s, and then have Dafny map the Absy's back to `proofDependency`s, so Dafny doesn't have to deal with the ids.

Source/DafnyCore/Verifier/ProofDependency.cs
keyboardDrummer: > I'd instead be inclined to move away from using NestedToken (although that would be too substantial an effort to be worthwhile).

I'm in favor of that, but also for not creating duplication in the meantime. My suggestion is to use NestedToken for now and improve it in a dedicated PR.

Source/DafnyCore/Verifier/Translator.cs
atomb: Dafny can't keep track of an Absy -> ProofDependency map because Boogie introduces new Absy nodes during internal transformations that Dafny knows nothing about. Boogie would end up returning Absy nodes that Dafny has never seen, and therefore can't be in the map Dafny created.

Source/DafnyCore/Verifier/ProofDependency.cs
atomb: Moving to NestedToken here would make this code much harder to comprehend, and although the `ProofDependency` type does overlap in purpose with `NestedToken`, it doesn't entirely.

One of the key goals of the current data structure is to allow, as an incremental improvement, embedding more information about the Dafny AST that led to the proof dependency in the first place, to allow for more informative messages. @MikaelMayer has begun doing similar things with `ProofObligationDescription` with very nice results. A `NestedToken` has no semantic content. It just tracks source ranges. Because of the complexity of the Dafny language, and the number of side conditions that come up, it's entirely possible for two source ranges, from two different "dependencies" to entirely overlap. A description string could disambiguate them, but awkwardly. Having the original AST node to disambiguate, instead, seems much more well-structured to me, and makes continually improving messages much more feasible.

atomb: The next step in this process will be to implement vacuity warnings, and having only `NestedToken` information to determine whether one of the proof dependencies is identical to one of the proof goals would be super fragile. As it is right now, though, it's possible to just look at the `ProofObligationDescripton` inside a `ProofObligationDependency`.

Source/DafnyCore/Verifier/Translator.cs
keyboardDrummer: > Dafny can't keep track of an Absy -> ProofDependency map

In what I described, it does not. Dafny does not keep track of anything. It retrieves an `IEnumerable<Absy>` from `VCResult.coveredElements` and those `Absy` contain `IToken`s with all the description that `ProofDependency` currently provides.

atomb: But `IToken` values don't (and I believe shouldn't) track all of the information that `ProofDependency` values do.

atomb: In particular, at least one of the `ProofDependency` types, and likely more in the future, keep the original _Dafny_ AST node that gave rise the dependency. That's only lightly used at the moment, but will need to be used more heavily in the upcoming changes I have planned.

atomb: It is also important to be able to print the Boogie files, as you suggest above, and encoding IDs in those that refer simply to Dafny source locations won't be enough, because Dafny source locations overlap too much. A single expression can result in multiple proof dependencies.

Source/DafnyCore/Verifier/Translator.ExpressionWellformed.cs
```diff
@@ -712,28 +712,52 @@ private record ReadsCheckDelayer(ExpressionTranslator etran, Function selfCallsA
...
-                foreach (var ss in TrSplitExpr(precond, etran, true, out var splitHappened)) {
-                  if (ss.IsChecked) {
-                    var tok = new NestedToken(GetToken(expr), ss.Tok);
-                    var desc = new PODesc.PreconditionSatisfied(errorMessage, successMessage);
-                    if (wfOptions.AssertKv != null) {
-                      // use the given assert attribute only
-                      builder.Add(Assert(tok, ss.E, desc, wfOptions.AssertKv));
-                    } else {
-                      builder.Add(AssertNS(tok, ss.E, desc));
+              // the check for .reads function must be translated explicitly: their declaration lacks
```
keyboardDrummer, line 715: This change is harder to review because it's part of a very large method, so it's not easy to find the method header and gather context about what this code is supposed to do.

I see this is part of a block to check the well-formedness of a call expression. Could we put that block in a separate method?

Also, could you see if you can come up with a method name that describes either the whole block between line 714 to 762 or one of the sub blocks, and if so put that in a separate method? Maybe something like `CheckReadsClauseSatisfied`

keyboardDrummer, line 715: `must be translated explicitly` is somewhat ambiguous to me, although I guess that you mean we need code specifically for checking that reads clauses are satisfied.

What does "there declaration lacks an explicit precondition" mean? Are reads functions declared? What would it look like if they had an explicit precondition?

Source/DafnyCore/Verifier/Translator.ExpressionWellformed.cs
```diff
@@ -712,28 +712,52 @@ private record ReadsCheckDelayer(ExpressionTranslator etran, Function selfCallsA
...
+              // an explicit precondition, which is added as an axiom in Translator.cs
+              if (e.Function.Name == "reads" && !e.Receiver.Type.IsArrowTypeWithoutReadEffects) {
```
keyboardDrummer, line 717: How is `IsArrowTypeWithoutReadEffects` relevant here?

Source/DafnyCore/Verifier/Translator.ExpressionWellformed.cs
```diff
@@ -712,28 +712,52 @@ private record ReadsCheckDelayer(ExpressionTranslator etran, Function selfCallsA
                 }
               }
               // check that the preconditions for the call hold
-              foreach (AttributedExpression p in e.Function.Req) {
```
keyboardDrummer: What's driving the changes at this location? I thought most of what you had to do was add extra attributes.

keyboardDrummer (approved):
Approved, but if you could either in this PR or in a follow-up address the remaining comments, that'd be much appreciated.

Source/DafnyCore/Verifier/Translator.ExpressionWellformed.cs
atomb: Hmm. I wonder whether this change and the others you comment on below are an artifact of the PR not being up-to-date with `master`? Because they aren't changes made by this PR. There's some recent work @jtristan made to resolve a soundness issue that touches the same files, and that I had to resolve conflicts to merge with, and I think the changes you're pointing to are from that. Were you reviewing an intermediate commit, perhaps?

PR #4464, opened 2023-08-23
Update and improve refreshed type system

RustanLeino:
This is an update of the type-system refresh. It only affects runs that use `/typeSystemRefresh:1`. The improves are:

* Pre-type inference of default-valued expressions
* Take module scopes into account (so that imported export-provides declarations don't see full definitions)
* Nested error messages (when variance of a type parameter causes a type-checking error)
* A type adjustment phase that defines and solves a set of "flows" to regain subset types (more details below)

With this PR, 260 of the 1500 test files in the test suite call the new resolver. Most of those 260 files call both the old resolver and the new resolver. Outside of this PR, I have temporarily hard-coded `/typeSystemRefresh:1` and run the entire test suite, which has resulted in fewer than 330 files that fail. There's work to be done to close the gap and get all test files to run with the new resolver.

For many experimental purposes, "traits on top of datatypes" are usable with this PR, when using `--general-traits --type-system-refresh` (`/generalTraits:1 /typeSystemRefresh:1` with the legacy CLI). Traits on top of co-datatypes and abstract types are also expected to work (again, modulo the the gap to be closed for the tests, as mentioned above). However, traits on top of `newtype` do not yet have full compilation support, so they will work only for type checking and verification.

# The type system refresh: pre-types and type adjustments

TBW


Source/DafnyCore/Resolver/PreType/Flows.cs
```diff
@@ -0,0 +1,443 @@
...
+using System.IO;
+using JetBrains.Annotations;
+
+namespace Microsoft.Dafny;
+
+record FlowContext(SystemModuleManager SystemModuleManager, ErrorReporter Reporter, bool DebugPrint) {
+  public TextWriter OutputWriter => SystemModuleManager.Options.OutputWriter;
+}
+
+abstract class Flow {
```
keyboardDrummer, line 20: Could you add a comment to this class?

Source/DafnyCore/Resolver/PreType/TypeAdjustor.cs
```diff
@@ -0,0 +1,278 @@
...
+  public void DebugPrint() {
+    systemModuleManager.Options.OutputWriter.WriteLine($"--------------------------- type-adjustment flows:");
+    foreach (var flow in flows) {
+      flow.DebugPrint(systemModuleManager.Options.OutputWriter);
+    }
+    systemModuleManager.Options.OutputWriter.WriteLine("------------------- (end of type-adjustment flows)");
+  }
+
+  protected override void PostVisitOneExpression(Expression expr, IASTVisitorContext context) {
+    if (expr is IdentifierExpr identifierExpr) {
```
keyboardDrummer, line 41: I would move these cases into the AST classes, so you don't forget to add a new case when you introduce a new AST class and to make it easier to read about and understand a subset of the language.

Source/DafnyCore/AST/Expressions/Expression.cs
```diff
@@ -60,6 +60,14 @@ public abstract class Expression : TokenNode {
       type = value.Normalize();
     }
   }
+  public Type UnnormalizedType {
```
keyboardDrummer, line 63: Could you add a comment? Also, if there is a way to avoid properties like this I think it would be great, since it's difficult to distinguish variations of a similar thing `UnnormalizedType, Type, Type.Normalize(), Type.NormalizeWrtScope()`

Also, is there a point after which types no longer need to be normalized? I see there are no Normalize calls from the code generators, but `Translator.cs` has one call to `.Normalize()`

jtristan (commented):
It would be nice to add concrete examples (maybe even a running example!) to the doc

Source/DafnyCore/Resolver/PreType/TypeAdjustments.cs
```diff
@@ -0,0 +1,274 @@
+//-----------------------------------------------------------------------------
```
jtristan, line 1: Is this code mostly utilities for the TypeAdjustor?

Source/DafnyCore/Resolver/PreType/TypeAdjustor.cs
jtristan: But it's so nice to have them all written like this, I find it more readable.

Source/DafnyCore/Resolver/PreType/Flows.cs
jtristan: +1

Source/DafnyCore/Resolver/PreType/Flows.cs
```diff
@@ -0,0 +1,443 @@
...
+
+  public FlowBetweenExpressions(Expression sink, Expression source, string description = "")
+    : base(sink, sink.tok, description) {
+    this.source = source;
+  }
+
+  protected override Type GetSourceType() {
+    return source.Type;
+  }
+}
```
jtristan, line 443: This is great code, and it seems like it can be made generic enough to be useful in other context. Should we have a directory where we define a generic abstract interpreter that we'll instantiate here and there?

Source/DafnyCore/Resolver/PreType/TypeAdjustor.cs
RustanLeino: I understand the trade-offs between these two styles. One makes it easier to understand a subset of the AST nodes, the other makes it easier to understand a subset of the operations applied to the AST. For many AST traversals in the new type system, I prefer this `ASTVisitor` style (which essentially uses type-case statements rather than dynamically dispatched methods on each class). However, setting up these flows seems like a good candidate for using the method-override-on-each-AST-node approach, because such a flow seems to be an intrinsic part of what an AST node is.

Nevertheless, I'd like to leave this organization as it is for now, until we get the full new type system in place. That will let us better separate the new passes over the AST from the old type system.

Source/DafnyCore/AST/Expressions/Expression.cs
RustanLeino: I would prefer that `.Type` just return whatever value is stored in the `.type` field. However, this seems to have changed at some point, so that `.Type` returns `.type.Normalize()`. I don't quite see the benefit of that, since most places will want to still call `NormalizeExpand()` to expand type synonyms. I assume there was some reason for the change to make `.Type` return the normalized type, so rather than changing that back, I defined the new `UnnormalizedType` field.

I added a comment to `UnnormalizedType`:
The new type inference includes a "type adjustment" phase, which determines the best subset types for a program. This phase works by adjusting (mutating) types in place, using `AdjustableType` type proxies. During that phase, it is necessary to obtain the unnormalized type stored in each AST node, which is what the `UnnormalizedType` property does. This property should only be used during the type adjustment phase. After type inference is complete, use `.Type` instead.

The answer to your question
> Is there a point after which types no longer need to be normalized so we can rely solely on `.Type`?
is that `.Type` should always be used after resolution pass 0. Since `.Type` returns normalized (that is, non-proxy) types, it is then not necessary to call `.Normalize()`. However, in most cases, it will still be necessary to call `.NormalizeExpand()` in order to expand type synonyms.

Source/DafnyCore/Resolver/PreType/TypeAdjustments.cs
RustanLeino: The code contains utilities for the PreType-to-Type phase. (That hadn't been clear to me when I started writing it.)

I split this file into two:
* `AdjustableType.cs` now contains the declaration of the class `AdjustableType` and the two `...TypePlaceholder` classes (which may change before this work is done)
* `PreType2TypeUtil.cs` contains class `TypeAdjustments` (which I have renamed `PreType2TypeUtil` and made into a `static` class)

Source/DafnyCore/Resolver/PreType/Flows.cs
RustanLeino: Yes, I like that idea. But I propose that we get this one instance of that abstract interpreter working first, and then consider generalizing it to support other abstract interpretation passes.

docs/TypeSystemRefresh.md
```diff
@@ -0,0 +1,73 @@
...
+
+Once names have been resolved and pre-types inferred, the type system converts the pre-types into types and then goes into a phase of _adjusting_ those types into types that involve subset types. For example, if a local variable `x` (without a user-supplied type) is inferred to have pre-type `int` and all assignments to `x` are from variables declared to have `nat`, then `x` will, at the end of the type adjustment phase, have type `nat`.
+
```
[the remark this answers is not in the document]
RustanLeino: I added some text and two examples.

docs/TypeSystemRefresh.md
```diff
@@ -0,0 +1,73 @@
...
+## Resolution phases
+
+Resolution proceeds in a number of phases. One group of these phases is called "pass 0" in the resolver (search for "`Pass 0`" in `ModuleResolver.cs`). This document is concerned with describe the phases in pass 0.
+
+The name "pass 0" is somewhat arbitrary, since resolution performs some important tasks before pass 0. Those initial tasks are quite delicate, since they don't even have pre-types available. It would be good to have a document that describes those initial tasks. For now, suffice it to mention some ingredients of those tasks: resolve import relations between modules, register the names of top-level declarations and member declarations in each module (that is, build a symbol table from names to declarations), resolve the types that occur in top-level declarations and member declarations (a tricky part is that `newtype`, subset type, and `const` declarations may omit such types, and those are instead later inferred).
```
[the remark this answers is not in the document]
RustanLeino: The three things I listed are indeed done in that order.

docs/TypeSystemRefresh.md
```diff
@@ -0,0 +1,73 @@
...
+
+For the new type system, pass 0 consists of the following phases:
+
+* Resolve names and infer pre-types. Ostensibly, this iterates over all declarations, computing the pre-type signature and body of each declaration. When one declaration uses another (that is, when one declaration has a name that resolves to another declaration), then it is only the pre-type signature of the other declaration that is needed. Because `newtype`, subset type, and `const` declarations can omit their type signatures and let them be inferred, the order in which declarations have to visited cannot be _a prioi_ determined (in particular, note that determining such an order requires names to be resolved, but this is the phase that resolves names!). Therefore, this phase keeps track of which declarations are in progress and which have finished, and so will visit the declarations on demand. If any cycles are detected, they are reported as errors.
```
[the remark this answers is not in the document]
RustanLeino: Just in the current module. I added words to make this clearer.

docs/TypeSystemRefresh.md
```diff
@@ -0,0 +1,73 @@
...
+
+The conversion from pre-types to types is coupled with the type-adjustment phase. The former needs to create adjustable types for the latter to adjust. Also, if there is a type that will not be adjusted, then the conversion phase should not set it to be a bottom type. For example, if the pre-types of expressions `x` and `y` are `int`, then the type of expression `x + y` is fixed to be `int`. That is, even if the types of `x` and `y` are eventually adjusted to, say, `nat`, the type of `x + y` is still defined to be `int`. For this situation, the type-adjustment phase does not define any flow from `x` and `y` to `x + y`, since it is not desirable to further adjust the type of `x + y`.
+
+## Solving constraints
+
+Pre-type inference builds up constraints that are required to hold among the pre-types to be inferred, and the type-adjustment phase builds up flows that are used to compute the final types as fix-points. At the moment, the data structures used in those phases are simple lists. Surely, there are far better ways to organize these constraints and flows. These can be improved and suitably optimized in the future.
```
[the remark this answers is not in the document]
RustanLeino: I added more documentation about this. (When this document will describe more details of pre-type inference, even more can be said. Since this PR is mostly about type adjustments, I'm postponing adding all those details.)

Source/DafnyCore/Resolver/PreType/TypeAdjustor.cs
keyboardDrummer: Let's leave it like it is for now. Seems like we need more discussion to agree on our preferred style anyways.

Source/DafnyCore/Resolver/PreType/Flows.cs
```diff
@@ -0,0 +1,454 @@
...
+using JetBrains.Annotations;
+
+namespace Microsoft.Dafny;
+
+record FlowContext(SystemModuleManager SystemModuleManager, ErrorReporter Reporter, bool DebugPrint) {
+  public TextWriter OutputWriter => SystemModuleManager.Options.OutputWriter;
+}
+
+/// <summary>
+/// A "Flow" is a puzzle piece in recomputing types. The "type adjustment" phase defines a set of flows and then
```
keyboardDrummer, line 21: Sounds like a constraint to me. Consider renaming to `SubsetPhaseConstraint`. Also, the name "type adjustment" doesn't have much meaning to me, considering a name that includes `SubsetType`, like "InferSubsetType"

jtristan (dismissed):
All my comments were addressed.

keyboardDrummer (dismissed):
Approving, changes for remaining comments can be made in a follow-up.

keyboardDrummer (dismissed):

jtristan (approved):

PR #4484, opened 2023-08-28
Allow cardinality to be treated as `long` in Java

### LINKED ISSUE #4483: Error when collection cardinality used as a `long` in Java
### Dafny version

4.2.0

```dafny
const TWO_TO_THE_64:  int := 0x1_00000000_00000000
newtype uint64 = x: int | 0 <= x < TWO_TO_THE_64

method Foo() returns (tuple: (uint64, uint64)) {
  var s := [1,2,3];
  var (foo, length) := (0, |s| as uint64);
  return (foo, length);
}
```

```code
$ dafny run -t:java longcardinality.dfy

example1-java/_System/__default.java:15: error: incompatible types: int cannot be converted to Long
    dafny.Tuple2<java.lang.Long, java.lang.Long> _let_tmp_rhs0 = dafny.Tuple2.<java.lang.Long, java.lang.Long>create(0L, (_0_s).cardinalityInt());
                                                                                                                                              ^
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error
```

### What happened?

It should have successfully compiled and run the Java code.

Mac

atomb:
Previously, using the cardinality of a collection in a context where it would be compiled to a `long` in Java led to a type error from the Java compiler. This change ensures that the necessary cast is inserted.

Fixes #4483


robin-aws (changes requested):
Fix certainly LGTM, but please do add the issue repro as a test case (especially as I'm curious to see if this fails in any other languages :)

robin-aws (approved):

PR #4486, opened 2023-08-29
Revisit reads checking

RustanLeino:
With some upcoming changes to `reads` checking (fixing soundness bugs in the encoding of function values, and adding opt-in `reads` clauses for methods), several issues were brought to the surface. These can broadly be placed in two groups:

* Eager versus delayed reads checking
* Reads checking for lambda expressions

For the former, one idea is to delay reads checking only when checking the well-formedness of preconditions. However, there are expressions (like quantifiers over function values where the quantifier's body restricts the function values to ones with a certain reads clause) that need delayed reads checking. So, it seems that a function's requires/reads/body should all do delayed reads checking. I think that's what's currently done, but it would be good to revisit this design point and to check that the implementation really does that.

Also for the former, default-value expressions that are subject to reads checking should do delayed reads checking. The reason is the same as above, namely that some expressions (like those quantifiers) need delayed checking. The general design for checking default-value expressions is that they should live up to the reads clause of the enclosing declaration--this way, it is never a surprise to a caller who implicitly uses the default-value expressions for the parameters.

For the second bullet above, the implementation seems to have two problems. One is that the lambda expression's precondition does not get assumed when the lambda expression's `reads` clause is checked. The other is that the reads checks for a lambda expression should be guided only by the `reads` clause of the lambda expression itself, not the enclosing context. Furthermore, when applying a function, reads checking should be done.

### DIFF
--- a/Source/DafnyCore/Verifier/Translator.ClassMembers.cs
+++ b/Source/DafnyCore/Verifier/Translator.ClassMembers.cs
@@ -668,7 +668,7 @@ namespace Microsoft.Dafny {
         // check well-formedness of any default-value expressions (before assuming preconditions)
         foreach (var formal in m.Ins.Where(formal => formal.DefaultValue != null)) {
           var e = formal.DefaultValue;
-          CheckWellformed(e, new WFOptions(null, false, false, true), localVariables, builder, etran);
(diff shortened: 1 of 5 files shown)


RustanLeino:
The good parts (the tests) of this PR have been rolled into other PRs.

PR #4493, opened 2023-08-30
Include more resolver-refresh tests

RustanLeino:
This PR includes fixes to the new resolver for

* datatype update expressions
* lambda expressions
* or-patterns with wildcards

As a result, the PR does

* convert more tests to include both resolvers

The PR also improves the following:

* include `DatatypeUpdateExpr.ResolvedCompiledExpression` in visitations
* improve `TryResolveTypeProxiesUsingKnownBounds` in the face of proxies
* do some debug printing for guarded constraints and post-inference confirmations


Test/VSComp2010/Problem2-Invert.dfy
```diff
@@ -26,57 +25,44 @@
 // connect the two.
 
 method M(N: int, A: array<int>, B: array<int>)
```
keyboardDrummer, line 27: Did the contents of this test need to change to allow `// RUN: %testDafnyForEachResolver "%s"` to be used? Why is that?

keyboardDrummer: Sorry, maybe my question wasn't specific enough. My question was: why did you need the changes on lines 29-81? I see that the change on line 1 triggers the new resolver to be used, but why can't the rest of the test stay the same?

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
```diff
@@ -404,10 +404,11 @@ public enum Type2PreTypeOption { GoodForInference, GoodForPrinting, GoodForBoth
     private class PreTypeInferenceModuleState {
       public readonly ISet<Declaration> StillNeedsPreTypeSignature;
       public readonly Stack<Declaration> InFirstPhase = new Stack<Declaration>();
-      public readonly Dictionary<string, TopLevelDecl> PreTypeBuiltins = new();
```
jtristan, line 407: Why these changes? Do we need to be able to use different PreTypeBuiltins Dictionnaries?

Source/DafnyCore/Resolver/PreType/PreTypeConstraints.cs
```diff
@@ -179,26 +241,40 @@ public class PreTypeConstraints {
...
         var bound = (fromSubBounds ? constraint.Sub : constraint.Super).Normalize() as DPreType;
-        if (proxy != null && bound != null) {
+        if (proxy != null) {
           if (!candidateHeads.TryGetValue(proxy, out var previousBest)) {
-            candidateHeads.Add(proxy, bound.Decl);
-            constraintOrigins.Add(proxy, constraint);
+            // we haven't seen this proxy before
+            if (bound != null) {
+              candidateHeads.Add(proxy, bound.Decl);
+              constraintOrigins.Add(proxy, constraint);
```
jtristan, line 263: Factor out that line after conditional for both branches?

jtristan (approved):
I left a few questions that could help me better understand the bigger picture, but looks great otherwise.

Test/VSComp2010/Problem2-Invert.dfy
RustanLeino: The rest could have stayed the same. But I took the opportunity to remove deprecated semi-colons, use a `for` loop instead `while`, use the display initialization of a new array, and remove some assert that wasn't needed for the proof.

Source/DafnyCore/AST/Expressions/Datatypes/DatatypeUpdateExpr.cs
```diff
@@ -43,14 +47,17 @@ public DatatypeUpdateExpr(IToken tok, Expression root, List<Tuple<IToken, string
...
+        Contract.Assert(ResolvedCompiledExpression == null);
         foreach (var preResolved in PreResolveSubExpressions) {
-
           yield return preResolved;
         }
       } else {
         foreach (var e in base.SubExpressions) {
           yield return e;
         }
+        if (ResolvedExpression != ResolvedCompiledExpression) {
```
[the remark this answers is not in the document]
RustanLeino: Yes, the previous version was incorrect. It had omitted the `ResolvedCompiledExpression` field.

Source/DafnyCore/AST/Expressions/Datatypes/DatatypeUpdateExpr.cs
```diff
@@ -26,7 +26,11 @@ public class DatatypeUpdateExpr : ConcreteSyntaxExpression, IHasUsages, ICloneab
       Members = original.Members;
       LegalSourceConstructors = original.LegalSourceConstructors;
       InCompiledContext = original.InCompiledContext;
-      ResolvedCompiledExpression = cloner.CloneExpr(original.ResolvedCompiledExpression);
+      if (original.ResolvedExpression == original.ResolvedCompiledExpression) {
```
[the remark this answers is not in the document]
RustanLeino: It's so that `ResolvedExpression` and `ResolvedCompiledExpression` can continue to be pointer equal, if they started off that way. And that property is, in turn, used in L58, which decides whether to include just the former or both in the `SubExpressions` (so that `ResolvedCompiledExpression` is not visited twice if it's identical).

Source/DafnyCore/Resolver/PreType/PreTypeConstraints.cs
```diff
@@ -56,9 +120,13 @@ public class PreTypeConstraints {
     }
 
     private bool TryMakeDecisions() {
-      if (TryResolveTypeProxiesUsingKnownBounds(true)) {
+      if (TryResolveTypeProxiesUsingKnownBounds(true, false)) {
```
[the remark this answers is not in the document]
RustanLeino: `TryResolveTypeProxiesUsingKnownBounds` used to have one parameter, and the old `TryMakeDecisions` called it with `true` and then `false`. Now, `TryResolveTypeProxiesUsingKnownBounds` takes two parameters, and `TryMakeDecisions` calls it with all four combinations of booleans, in a particular order.

Source/DafnyCore/Resolver/PreType/PreTypeResolve.cs
RustanLeino: It's so that `PreTypeBuiltins` can be defined in `SystemModuleManager` instead. In this way, all modules will use the same `PreTypeBuiltins`, rather than getting a different one for each module. And this, in turn, is needed in order for the pre-type resolver to properly understand, for example, that there's just one `int` type.

Source/DafnyCore/Resolver/PreType/PreTypeConstraints.cs
RustanLeino: The `constraintOrigins.Add` happens in 2 of the 3 branches. (Note that this is an `if ... else if`, not an `if ... else`.)

jtristan (approved):

PR #4495, opened 2023-08-31
Prevent the IDE from locking up when certain parse error occur

keyboardDrummer:
### Changes

Previously the IDE server would become completely unresponsive when a repeated comma occurred in a parameter, now it does not.

### Testing

Added test to ensure the behavior does not occur


MikaelMayer (approved):
Wow good finding and fix!

Issue #1367, opened 2021-08-22
Go: bug in multiple occurrences of the same key in map


sorawee:
```
method Main () 
{
  var e := map[1 := false, 1 := true];
  print e == e;
}
```

prints `false` in Go.

sorawee:
The data representation of `map` in Go makes multiple occurrences difficult to get right (or efficient, for that matter), since it's just a list of key value pairs, with no invariant that keys are distinct.

sorawee:
Actually, this also affects JS:

```
method Main() {
  print |map[1 := 2, 1 := 3]|;
}
```

prints `2`.

RustanLeino:
Thanks for pointing out this bug. I think the language ought not to allow this case. That is, I think the verifier should check that the keys given in the map display expressions are distinct.

ShubhamChaturvedi7:
This also leads to inconsistency when converting from `dafny.Map` to Go native `map` where the last value in the array is written into the Go map.

PR #4497, opened 2023-08-31
fix: Fix multiple occurrences of same key in map for Go and JS (reopened)

robin-aws:
(Re-opening @sorawee's PR: https://github.com/dafny-lang/dafny/pull/1370)

Attempting to enforce the invariant that keys are all distinct.

The distinction check changes the time complexity of a map display expression of size n to O(n^2). A better data structure is required to be more efficient.

(Robin's note: that's just fine IMO since most set and map operations are similarly inefficient already. This change at least addresses the soundness issue - the change to verification will be more involved and not backwards compatible, so I prefer to decouple the issue fixes.)

Fix https://github.com/dafny-lang/dafny/issues/1367

(Robin's second note: I just merged the latest master and added a bit more testing)


RustanLeino:
I think this is verifier bug (see https://github.com/dafny-lang/dafny/issues/1367#issuecomment-906754644), and should not require any change to the compilers.

robin-aws:
I disagree: yes it's a potential improvement to the language to require that map displays have unique keys, but that is a language definition change and a breaking one, not to mention incurring higher verification cost. It's not a verifier *bug* because the verifier is correct w.r.t. the current language definition.

The current semantics of the language are clear, and the reference manual does state that later entries override earlier ones, and Go and JS are not implementing those semantics correctly. It's cheap to fix that soundness issue, with the only downside being map displays in those languages are slower now (O(n^2) instead of O(n)). But given other map operations in those backends are already similarly poor, the impact will be minimal.

I say we do both: correct the compilers now, and then improve the language later!

RustanLeino:
> I say we do both: correct the compilers now, and then improve the language later!

Okay.

RustanLeino (approved):

PR #4499, opened 2023-09-01
Include both locations in duplicate module name error messages

### LINKED ISSUE #4289: Duplicate module error should give location of the other module
### Summary

It would be convenient, when reporting duplicate module names, to include the file and line number of the other module with the same name; at the moment it just reports the "current" duplicate module definition line number.

### Background and Motivation

When trying to verify Dafny code, I may have inadventently defined a module with the same name, or imported something twice. The fixes are, as far as I know:

- Rename the latter module
- Rename the former module
- Fix the double import problem

To determine which of the fixes is appropriate, one would like locations of both modules, possibly with a separate error message if it's a double include issue.

### Proposed Feature

instead of

```/local/home/some_location/SomeFile.dfy(3,38): Error: Duplicate module name: MyModule```

one could imagine:

```/local/home/some_location/SomeFile.dfy(3,38): Error: Duplicate module name: MyModule, conflicts with definition at /some_other_location/SomeOtherFile.dfy(4,12)```

or

```Error, file /some_location/SomeFile.dfy is included twice```

### Alternatives

_No response_

### PR DESCRIPTION
**Pull request by keyboardDrummer**
Fixes #4289

### DIFF
--- a/Source/DafnyLanguageServer/Workspace/ProjectManager.cs
+++ b/Source/DafnyLanguageServer/Workspace/ProjectManager.cs
@@ -221,7 +221,7 @@ public class ProjectManager : IDisposable {
     return await CompilationManager.LastDocument;
   }
 
-  public async Task<IdeState> GetSnapshotAfterParsingAsync() {
+  public async Task<IdeState> GetStateAfterParsingAsync() {
     try {
       var parsedCompilation = await CompilationManager.ParsedCompilation;
       logger.LogDebug($"GetSnapshotAfterParsingAsync returns compilation version {parsedCompilation.Version}");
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: Source/DafnyLanguageServer/Workspace/ProjectManager.cs [hunk 1]
```diff
@@ -221,7 +221,7 @@ public class ProjectManager : IDisposable {
     return await CompilationManager.LastDocument;
   }
 
-  public async Task<IdeState> GetSnapshotAfterParsingAsync() {
```
keyboardDrummer, line 224: Changes in this file are unrelated

stefan-aws (approved):
Thank you, this is very helpful.

PR #4529, opened 2023-09-08
Show assertions as hint diagnostics

keyboardDrummer:
### Changes
- Turned off by default: show implicit assertions as hint diagnostics, which can help a user understand where implicit assertions are coming from
 
image

- Show a warning when the error limit for verification has been hit

image

### Testing
- Added XUnit tests for the new features
- Many existing tests were updated so they don't depend on coincidental behavior, which could lead them to fail if showing assertion was turned on by default.


Source/DafnyLanguageServer.Test/Synchronization/VerificationDiagnostics.cs
```diff
@@ -0,0 +1,72 @@
...
+method ImplicitAssertions(x: int) {
+  if (*) {
+    assert false;
+  } else {
+    var x := 3 / 2;
+  }
+}".TrimStart();
+    var document = CreateAndOpenTestDocument(source, "ErrorLimitReached.dfy");
+    var diagnostics = await GetLastDiagnostics(document, DiagnosticSeverity.Hint);
+    Assert.Contains(diagnostics, d => d.Message.Contains("Implicit assertion: non-zero divisor"));
```
MikaelMayer, line 48: From the screenshot you shared, the hint is on 3 but it should be on the / symbol or even 2. I know it's not part of your PR to change that, but since your PR is making it visible to the user, we might want to fix it at the same time. I'd put the hint on the divisor since it's the operation that requires an implicit assertion

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
```diff
@@ -75,36 +75,37 @@ decreases y
...
-      var change2 = MakeChange(documentItem.Version + 4, new Range((0, 6), (0, 7)), "");
+      var change2 = MakeChange(documentItem.Version + 2, new Range((0, 6), (0, 7)), "");
+      documentItem = documentItem with { Version = change2.TextDocument.Version };
       client.DidChangeTextDocument(change2);
       // Save and don't wait, so the next save will interrupt and cancel verification
       client.SaveDocument(documentItem);
 
       // Do the previous again a few times. This seems to be what it takes to guarantee cancelling verification.
-      client.DidChangeTextDocument(MakeChange(documentItem.Version + 5, new Range((0, 6), (0, 6)), " "));
+      ApplyChange(ref documentItem, new Range((0, 6), (0, 7)), " ");
```
MikaelMayer, line 90: Good refactoring that will make it easier to maintain. Note that you can also remove the "new Range".

Source/DafnyLanguageServer/Server.cs
```diff
@@ -3,7 +3,9 @@
 using OmniSharp.Extensions.LanguageServer.Server;
 using Serilog;
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
```
MikaelMayer, line 8: Why are these two using now necessary in this file?

MikaelMayer (commented):
Minor clarification suggestions

Source/DafnyLanguageServer.Test/Various/ConcurrentInteractionsTest.cs
keyboardDrummer: I was surprised that this is possible. It works by implicitly converting the 2-tuples to Positions and then implicitly converting two Positions to a Range.

Source/DafnyLanguageServer.Test/Synchronization/VerificationDiagnostics.cs
MikaelMayer: I understand from your comment that you do consider like me that it would be better to place the hint on the "/", but you feel blocked by the current implementation that looses the information about the "/". If only the implementation causes problem, I suggest a 4-line fix:

- Look for `Translator.GetToken()`, and replace `node.RangeToken.ToToken();` with `node.RangeToken.ToToken(node.Tok);`
- Add an optional parameter "nameToken" to that method `.ToToken()`, set to `null` by default
- Replace `null` by `nameToken` in the body of `ToToken()`
- Check that, when you report the info, you are using the token itself and not the range.

keyboardDrummer: > I understand from your comment that you do consider like me that it would be better to place the hint on the "/"

I'm not sure !

> If only the implementation causes problem, I suggest a 4-line fix:

I don't fully understand the change, but it looks quite generic. I'd rather not do that as part of this PR.

MikaelMayer: > > I understand from your comment that you do consider like me that it would be better to place the hint on the "/"

> I'm not sure !

You are not sure that the hint about the implicit assertion about the non-zero divisor should better be on the "/" rather than the numerator "3" ?
What about the following:
```
assert x / z
          ^ Hint that it's divisor is non zero, twice. Looks like a duplicate. But second one is about p.
          / p;
          
```

MikaelMayer: That's a good solution too for that example and excellent remark as well for preferring the divisor over the operator itself.
It looks like a good compromise, given that / * and % are left associative and have the highest precedence among binary infix operators, so even if the divisor was an expression like `(y+y)`, the error would be at least reported on the opening parenthesis. Please also perform the same modification for reporting the "modulo is not zero" hint for %.

However, there are a few more binary operators that have an even higher precedence.

Consider the following examples:
```
assert x / y ^ z
        // ^ Hint: Divisor is always non zero
assert x / y | z
        // ^ Hint: Divisor is always non zero
assert x / y & z
        // ^ Hint: Divisor is always non zero
```
which are all parsed like a division of a divisor that is a binary operation. With your approach, the error and the issue would be both on `y`, as it's the start of each sub-expression `y ^ z` `y & z` or `y | z`, which is not very user-friendly.

Finally, all of this boils down to the fact that we report ranges, which is great to highlight errors on expressions, but hints happen only on a single token. For this simple token, we should revert to the old behavior of tokens that uniquely identify an expression of interest. Reporting only the first token is too ambiguous.

Therefore, if you merge your PR with this one that I just created, and instead of reporting the token, you just extract the `UniquelyIdentifyingToken` if it's not null when reporting info,
https://github.com/dafny-lang/dafny/pull/4621
the examples I'm giving will display the hint here:
```
assert x / y ^ z
          // ^ Hint: Divisor is always non zero
assert x / y | z
          // ^ Hint: Divisor is always non zero
assert x / y & z
          // ^ Hint: Divisor is always non zero
```
which is acceptable and non-ambiguous (the hint uniquely identifies the position of the divisor),
If we reported the hint on the division sign, it could clash as it identifies the whole expression instead of only the divisor, so your remark about reporting it on the divisor made the point that the hint should be only on the divisor as it's like a subset type check for the divisor.

keyboardDrummer: > assert x / y ^ z
          // ^ Hint: Divisor is always non zero
assert x / y | z
          // ^ Hint: Divisor is always non zero
assert x / y & z
          // ^ Hint: Divisor is always non zero

I would rather report on the `/` in that case.

MikaelMayer: You can. But note that the "/", if it identified an expression, identifies the entire division. We could also have a nested token that points at "/" and at the token of the division, but it's a bit overkill for a simple hint, and warnings are already on the entire division anyway to related range would not help.

I let you choose.

keyboardDrummer: Hey, I agree with you it would be better to show the `...` on `/` than on the start of the `/` expression, but do you want to block this PR on that? Since this PR is an optional feature off by default, I don't see how it could make things worse, and there's nothing stopping us from improving the UX in a follow-up.

MikaelMayer: I won't _block_ this PR, otherwise I would have put my status as "requests change". That's what I call blocking a PR. If another Dafny reviewer reviews this PR and approves it after reading this thread, I won't complain, your argument is legit that it's experimental and we can build on it.

My argument is legit too.
When I review, I'm always careful about the things I ask because I also prefer to move fast and iterate then. This is why, after your dubitative response, I tested to see if my request made sense, and it did https://github.com/dafny-lang/dafny/pull/4621 as all tests pass with only 7 lines modified. Taking my request into account equals to merging only 8 lines of code, after which I'll obviously approve it because it pass the minimal quality bar I set up for any feature, even experimental: It should be logical, have no obvious flaws, and consistent with the rest of Dafny, and no typos. Here we are talking about the consistency with the rest of Dafny, and having no obvious flaws.

And I'm saying that knowing I'm very much looking for having this feature (even turned on by default). We both agree it would be better to show the "..." on the division sign than the start of the expression. Since the cost is negligible (import my PR, add this one line of code), the ratio benefit/cost is now huge.
A small step for mankind, a huge step for Dafny. Well I'm a bit too enthusiastic here but you see the picture.

keyboardDrummer: For me, when authoring a PR, there's little difference between having a "request change" and a reviewer who intentionally has not approved it. I will only merge when there's unanimous agreement of all reviewers that it would be good to.

For this PR, I'd prefer to merge it and do the UI improvement as a follow-up. There's also some test refactoring in here that I'm missing in other PRs.

MikaelMayer: We are getting there. I'm waiting for your other PR with complete positions to merge, so that you can update this one, that way, we don't need to think of a follow-up.
You can always merge the test refactorings in your other PRs, the first one to merge will cancel the other ones.

Source/DafnyCore/Rewriters/RefinementTransformer.cs
```diff
@@ -58,7 +58,8 @@ public RefinementToken(IToken tok, ModuleDefinition m)
       }
 
       var rtok = tok as RefinementToken;
-      return rtok != null && rtok.InheritingModule == m;
+      var result = rtok != null && rtok.InheritingModule == m;
+      return result;
```
MikaelMayer: Note: when you are debugging, you can add a breakpoint on the return statement and step through. The debugger will display the returned value. That way, you don't accidentally commit temporary variable names

Test/logger/ProofDependencyLogging.dfy
```diff
@@ -3,7 +3,7 @@
 // CHECK: Results for RedundantAssumeMethod \(correctness\)
 // CHECK:     Proof dependencies:
 // CHECK:       ProofDependencyLogging.dfy\(177,12\)-\(177,16\): assume statement
-// CHECK:       ProofDependencyLogging.dfy\(178,14\)-\(178,14\): assertion always holds
+// CHECK:       ProofDependencyLogging.dfy\(178,12\)-\(178,16\): assertion always holds
```
MikaelMayer: Nice !

MikaelMayer (approved):
Well done, and thanks for your patience while I was reviewing everything. I saw that you reverted the hint on the / itself rather than the divisor to avoid having to fix all the tests, which I think is legit. Also, it shows the place of the division so it's ok.

Source/DafnyCore/Rewriters/RefinementTransformer.cs
keyboardDrummer: Thanks

PR #4533, opened 2023-09-11
Apply simplification to Imp and Iff connectives

zafer-esen:
Boogie applies simplification to `Bpl.And` and `Bpl.Or`, but [not to](https://github.com/boogie-org/boogie/blob/f088818450b0e9928a674e832435617a70c37ea7/Source/Core/AST/AbsyExpr.cs#L338-L352) `Bpl.Imp` and `Bpl.Iff`. Dafny provides [wrapper methods](https://github.com/dafny-lang/dafny/blob/37e00f70c3007002c4470a2d0129d66724473850/Source/DafnyCore/Verifier/Translator.BoogieFactory.cs#L772-L798) to apply these connectives which also apply simplification to `Imp` and `Iff`; however they do not seem to be used consistently and are missing some cases.

This PR replaces all direct usages of the Boogie methods `Bpl.And`, `Bpl.Or`, `Bpl.Imp` and `Bpl.Iff` with wrappers from Dafny, and adds code to avoid adding `true` axioms in cases where I noticed this happens. (A better solution might be to provide a single point for adding axioms where this check could be added.)


Source/IntegrationTests/TestFiles/LitTests/LitTest/dafny4/gcd.dfy
```diff
@@ -177,7 +177,7 @@ method EuclidGcd(X: pos, Y: pos) returns (gcd: pos)
 // ------------------------------------------------------------------------------------------------------
 // The alternative definitions that follow allow the two cases in the GCD algorithm to look more similar.
 
-lemma GcdSubtractAlt(x: pos, y: pos)
+lemma {:resource_limit "150e6"} GcdSubtractAlt(x: pos, y: pos)
```
keyboardDrummer, line 180: Why is this now needed?

keyboardDrummer (approved):

Source/IntegrationTests/TestFiles/LitTests/LitTest/dafny4/gcd.dfy
atomb: That example was just incredibly brittle, and for things in that state it seems that _any_ change, even one that overall should be a simplification, can throw things off the rails. Changing things as simple as the filenames passed in as arguments to the `:captureState` attribute was enough to change resource use by an order of magnitude. I've refactored that test, and one other, to be less brittle, and they no longer require raised resource limits.

The remaining failures will, I believe, be fixed by brittleness reductions included in #4597, so let's merge that one first, update this PR, and see if everything is good.

Source/DafnyStandardLibraries/src/Std/Arithmetic/LittleEndianNat.dfy
```diff
@@ -303,7 +303,7 @@ abstract module {:disableNonlinearArithmetic} Std.Arithmetic.LittleEndianNat {
 
   /* The nat representation of a sequence and its least significant position are
   congruent. */
-  lemma LemmaSeqLswModEquivalence(xs: seq<digit>)
+  lemma {:resource_limit "10e6"} LemmaSeqLswModEquivalence(xs: seq<digit>)
```
keyboardDrummer, line 306: Is this still needed?

keyboardDrummer (dismissed):

Source/DafnyStandardLibraries/src/Std/Arithmetic/LittleEndianNat.dfy
atomb: Oh, it looks like it isn't! Good catch.

keyboardDrummer (approved):

keyboardDrummer (approved):

PR #4541, opened 2023-09-12
Display proof dependency warnings in the IDE

atomb:
Follow-up from https://github.com/dafny-lang/dafny/pull/4542 to make the information available in the language server.
image

It requires 
- A Dafny file next to a `dfyconfig.toml` file containing 
```
[options]
warn-contradictory-assumptions = true
warn-redundant-assumptions = true
```
It displays
- Warnings about assertions proved by contradictory assumptions
- Warning about assumptions that were not used

Added tests for these warnings.


Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -27,10 +27,17 @@ public class DafnyProgramVerifier : IProgramVerifier {
...
       CompilationAfterResolution compilation,
       ModuleDefinition moduleDefinition,
       CancellationToken cancellationToken) {
+      var engine = boogieEngine;
+      if (compilation.Project.UsesProjectFile &&
+          (compilation.Program.Reporter.Options.Get(CommonOptionBag.WarnContradictoryAssumptions)
+           || compilation.Program.Reporter.Options.Get(CommonOptionBag.WarnRedundantAssumptions)
+          )) { // We need a separate Boogie engine
+        engine = new ExecutionEngine(compilation.Program.Reporter.Options, engine.Cache);
+      }
```
MikaelMayer, line 40: This is a crucial place: Before, the engine was never considering the options in dfyconfig.toml. These lines ensure that a new engine is spun if we use a project file. This is a temporary solution only for the experimental feature of warning about assumptions.
.
I think we might want a better long-term solution to avoid creating an execution engine for every thread; Did you consider that, @keyboardDrummer  ?

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -173,7 +173,8 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
         // Ok no assertion here. Maybe a method?
         if (node.Position.Line == position.Line) {
           areMethodStatistics = true;
-          return GetTopLevelInformation(node, orderedAssertionBatches);
+          return GetTopLevelInformation(node, orderedAssertionBatches) +
```
keyboardDrummer, line 176: Did we consider alternatives to Hover as a UI for this information?

Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
```diff
@@ -309,13 +308,40 @@ public class VerificationProgressReporter : IVerificationProgressReporter {
           targetMethodNode.StatusVerification = finalOutcome;
         }
 
+        var options = compilation.Program.Reporter.Options;
+        ReportVacuityAndRedundantAssumptionsChecks(compilation, implementation, verificationResult, options);
```
keyboardDrummer, line 312: Why not put this call here? https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Workspace/CompilationManager.cs#L402

I'd rather not add any code to `VerificationProgressReporter.cs`. I'm hoping to remove it after moving gutter icons to the client and updating hover so it doesn't need it.

Source/DafnyLanguageServer/Workspace/VerificationProgressReporter.cs
```diff
@@ -237,9 +236,9 @@ public class VerificationProgressReporter : IVerificationProgressReporter {
   /// Triggers sending of the current verification diagnostics to the client
   /// </summary>
   public void ReportRealtimeDiagnostics(CompilationAfterParsing compilation, Uri uri, bool verificationStarted) {
-    if (options.Get(ServerCommand.LineVerificationStatus)) {
+    if (compilation.Program.Reporter.Options.Get(ServerCommand.LineVerificationStatus)) {
```
keyboardDrummer, line 239: We can also add `Options` as a field to `Compilation`.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
keyboardDrummer: Instead of creating a new Boogie engine here, we should move the field `ProjectManagerDatabase.boogieEngine` to `ProjectManager`. ProjectManager is the longest lifetime in which the final options are available.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -235,12 +236,44 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
...
+      var depManager = program.ProofDependencyManager;
+      foreach (var assertionBatch in orderedAssertionBatches) {
+        var batchDependencies = assertionBatch.CoveredIds.Select(depManager.GetFullIdDependency);
+        var batchPotentialDependencies = depManager.GetPotentialDependenciesForDefinition(assertionBatch.VerboseName);
+        allPotentialDependencies.UnionWith(batchPotentialDependencies);
+        usedDependencies.UnionWith(batchDependencies);
+      }
+
+      if (usedDependencies.Any() && options.TrackVerificationCoverage) {
+        result += usedDependencies.Aggregate("\n\nProof dependencies:\n",
```
keyboardDrummer, line 264: Can you add screenshots with examples of this UI?

keyboardDrummer:
I find the message in the screenshot, "proved using contradictory assumptions: the assertion always holds", difficult to interpret. The second part, "the assertion always holds", is a success message which implies that the assertion was proven. What I think you want is a description of the assertion, but only in cases where it's an implicit one so the location is not enough to let the reader know what assertion you're referring to. Consider using `ShortDescription` instead of `Description` in `reporter.Warning(MessageSource.Verifier, "", dep.Range, $"proved using contradictory assumptions: {dep.Description}");`

keyboardDrummer:
> The behavior currently returns all the possible warnings

What do you mean by that?

keyboardDrummer (changes requested):
Left comments

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -235,12 +236,44 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
...
+          (acc, dep) => acc + $"* `{dep.RangeString()}: {dep.Description}`\n");
+      }
+
+      var unusedDependencies = allPotentialDependencies.Except(usedDependencies);
+      if (unusedDependencies.Any() && options.TrackVerificationCoverage) {
+        result += unusedDependencies.Aggregate("\n\nUnused program elements:\n",
```
keyboardDrummer, line 270: It seems strange to report at location X (the method header), about an unused program element that occurs at location Y. Why not only report at location Y ?

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
keyboardDrummer, line 264: Should we report which parts of the proof were used, if we're already reporting what wasn't used? Isn't that giving redundant information to the user?

I think it's bad if we put too much information in the hover screen, and actually I feel we already have too much in there.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -235,12 +236,44 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
...
       return information;
     }
 
+    private string GetProofDependencyInformation(IdeState state, TopLevelDeclMemberVerificationTree node, List<AssertionBatchVerificationTree> orderedAssertionBatches) {
+      var allPotentialDependencies = new HashSet<ProofDependency>();
+      var usedDependencies = new HashSet<ProofDependency>();
+      var program = (Program)state.Program;
+      var result = "";
+      if (program is null) {
+        return "null program";
```
keyboardDrummer, line 252: This seems like it shouldn't be shown to the user

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: It's an experimental interface to display the result of tracking proof dependencies. It won't be activated by default. Only users turning this option for their _project_ on will have this extra information on hover.
We need more return on experience about using this option, how customers use it, what it enables, before taking decisions that will make it to the default UI, assuming we activate this option.
For now, even though it's not the panacea, hover is the cheapest and easiest way to visualize this information, so I suggest we'll keep it that way until we revisit the whole experience of proof dependencies.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: It depends which question you want to answer:
* Is this element being used? In that case, you want to report that information on hover of location Y. We already do this on assumptions and requires, but not for specific expressions, as they are not explicitly part of the specifications, so I feel it would put too much noise on the body.
* Are there any elements being not used? In that case, you need a place to report that information that does not depend on the elements, so the method header at location X, not depending on any element, is a good choice for expressions without putting too much noise on the body.

MikaelMayer:
> I find the message in the screenshot, "proved using contradictory assumptions: the assertion always holds", difficult to interpret. The second part, "the assertion always holds", is a success message which implies that the assertion was proven. What I think you want is a description of the assertion, but only in cases where it's an implicit one so the location is not enough to let the reader know what assertion you're referring to. Consider using `ShortDescription` instead of `Description` in `reporter.Warning(MessageSource.Verifier, "", dep.Range, $"proved using contradictory assumptions: {dep.Description}");`

"which implies that the assertion was proven": But that's exactly the case. The proof dependency information is provided only if everything verifies.
I agree though that the format of the message is not always the best. It's generic enough so that we can replace "the assertion always hold" by any of the success messages that other `ProofObligationDescription` provide (like "the divisor is always non-zero", "the index is always in range", "lower bound is always lower than upper bound", etc. What about one of the following:

* "_the assertion always holds_, but that was proved using contradictory assumptions"
* "_the assertion always holds_; however, it's because Dafny was able to prove false before."
* "using contradictory assumptions, Dafny could prove that _the assertion always holds_"
* "Dafny proved using contradictor assumptions that _the assertion always holds_"
* "Beware: the assertion always hold, but it's because Dafny derived a contradiction from assumptions, so that code is unreachable."

The short description could be used but it's much less readable. If anything, I would prefer to have a generic message that does not refer to the description than refer to the short description, such as
* "At this point, Dafny has inferred a contradiction in the assumptions so that's why everything verifies"
* "Here, Dafny inferred false so everything will be proved"
* "Dafny is able to infer a contradiction in the assumptions at this state so it can prove everything"
* etc.

keyboardDrummer:
> > I find the message in the screenshot, "proved using contradictory assumptions: the assertion always holds", difficult to interpret. The second part, "the assertion always holds", is a success message which implies that the assertion was proven. What I think you want is a description of the assertion, but only in cases where it's an implicit one so the location is not enough to let the reader know what assertion you're referring to. Consider using `ShortDescription` instead of `Description` in `reporter.Warning(MessageSource.Verifier, "", dep.Range, $"proved using contradictory assumptions: {dep.Description}");`
> 
> "which implies that the assertion was proven": But that's exactly the case. 

It seems like we're discussing the semantics of what it means to say the assertion was proven. I think this is a case where it's controversial to say the assertion was proven, given that the SMT solver ignored the expression that is asserted.

I think a message like "Could not verify assertion because contradictory assumptions occur before it" would be more clear to me.

> * "_the assertion always holds_, but that was proved using contradictory assumptions"
> * "_the assertion always holds_; however, it's because Dafny was able to prove false before."
> * "using contradictory assumptions, Dafny could prove that _the assertion always holds_"
> * "Dafny proved using contradictor assumptions that _the assertion always holds_"
> * "Beware: the assertion always hold, but it's because Dafny derived a contradiction from assumptions, so that code is unreachable."

All of your suggestions state that the assertion holds, but why do you want to say that? To me it seems misleading: once you resolve the contradiction the assertion may no longer hold.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
keyboardDrummer: It might be easier to discuss with screenshots showing this UI

keyboardDrummer:
> > It seems like we're discussing the semantics of what it means to say the assertion was proven. I think this is a case where it's controversial to say the assertion was proven, given that the SMT solver ignored the expression that is asserted.
> 
> There is no controversy to say the assertion was _proven_, What you mean is that the assertion was not proven in a way that requires decomposing the expression or analyzing its sub-components. But that does not remove the validity of the proof. I'll explain this below.
> 
> > I think a message like "Could not attempt to verify assertion because contradictory assumptions occur before it" would be more clear to me.
> 
> I understand it's hard to swallow, but a message like the one you suggest wouldn't be mathematically correct. 

I think it's fair to say that the situation in which anything can be proven can easily lead to confusion. Sure it's true that the assertion was proven, but it's confusing to say so because it's easy to read such a statement as that it was proven without a contradictory assumption.

How about "Verification could not reach assertion because contradictory assumptions occur before it" ? Indeed for `assert false` this should not be reported, because that statement is only used when the user intends for it to be unreachable.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
atomb: I'm slightly confused by this suggestion. It seems that `ProjectManager` already has a `boogieEngine` field (and has since July), and the value of that field is exactly what's being passed in to `GetVerificationTasksAsync` here.

atomb:
I removed the proof dependency hover for now. It was a proof-of-concept placeholder that I didn't plan to merge. For now, warnings work, and I'd like to get those available. I'll create a future PR with a better display of full dependency information.

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
keyboardDrummer: Yes it has that field but it does not create the BoogieEngine, which is a problem because now the BoogieEngine does not use the options specific to that ProjectManager.

keyboardDrummer: The reason I only instantiated a single BoogieEngine is because each BoogieEngine creates a number of large stack threads, because our stack-dependent code overflows the stack when using regular threads, and thus consumes a lot of memory, and if you don't use project files, one BoogieEngine is created per file open in the IDE.

So.. what we also need is a way to pass a custom large stack thread pool to ExecutionEngine when creating it, so we can share those pools between the engines.

The following ExecutionEngine constructor from Boogie should have a `TaskScheduler largeStackThreadScheduler` argument:
```
    public ExecutionEngine(ExecutionEngineOptions options, VerificationResultCache cache) {
      Options = options;
      Cache = cache;
      checkerPool = new CheckerPool(options);
      verifyImplementationSemaphore = new SemaphoreSlim(Options.VcsCores);
      
      largeThreadScheduler = CustomStackSizePoolTaskScheduler.Create(16 * 1024 * 1024, Options.VcsCores);
      largeThreadTaskFactory = new(CancellationToken.None, TaskCreationOptions.None, TaskContinuationOptions.None, largeThreadScheduler);
    }
```

keyboardDrummer: We shouldn't create more Boogie `ExecutionEngine`s than we need, since it also has a `CheckerPool`, which is a pool of solvers it uses to verify things. This pool is specific to the DafnyOptions, since these options can configure the solver. So.. we should create 1 `ExecutionEngine` per `ProjectManager`

atomb: Alright, good. I like that. And I can make the change to do that as part of this PR (though I think the Boogie refactoring should wait for a later one).

keyboardDrummer: I see several things we could change in Boogie:
1. Allow passing a CustomStackSizePoolTaskScheduler to ExecutionEngine
1. Make CustomStackSizePoolTaskScheduler lazily create new threads
1. Make CustomStackSizePoolTaskScheduler scale up and down

I'd feel safer about creating 1 ExecutionEngine per ProjectManager if we do (1) as well, which should be a trivial change, then we don't consume any additional memory as part of this PR.

Source/DafnyLanguageServer/Workspace/GutterIconAndHoverVerificationDetailsManager.cs
```diff
@@ -309,13 +309,40 @@ public class GutterIconAndHoverVerificationDetailsManager : IGutterIconAndHoverV
           targetMethodNode.StatusVerification = finalOutcome;
         }
 
+        var options = compilation.Program.Reporter.Options;
+        ReportVacuityAndRedundantAssumptionsChecks(compilation, implementation, verificationResult, options);
```
keyboardDrummer, line 313: Could you move this out of this class, to where `ReportEndVerifyImplementation` is called?

Source/DafnyLanguageServer/Workspace/GutterIconAndHoverVerificationDetailsManager.cs
```diff
@@ -309,13 +309,40 @@ public class GutterIconAndHoverVerificationDetailsManager : IGutterIconAndHoverV
...
+
         targetMethodNode.PropagateChildrenErrorsUp();
         targetMethodNode.RecomputeAssertionBatchNodeDiagnostics();
         PublishGutterIcons(compilation, uri, true);
       }
     }
   }
 
+  private static void ReportVacuityAndRedundantAssumptionsChecks(CompilationAfterResolution compilation,
```
keyboardDrummer, line 322: Could you move this out of this class?

Source/DafnyLanguageServer/Workspace/GutterIconAndHoverVerificationDetailsManager.cs
```diff
@@ -309,13 +309,40 @@ public class GutterIconAndHoverVerificationDetailsManager : IGutterIconAndHoverV
...
+    Implementation implementation, VerificationResult verificationResult, DafnyOptions options) {
+    if (!options.Get(CommonOptionBag.WarnContradictoryAssumptions)
+        && !options.Get(CommonOptionBag.WarnRedundantAssumptions)
+       ) {
+      return;
+    }
+
+    var vcResults = verificationResult.VCResults.Select(r =>
+      new DafnyConsolePrinter.VCResultLogEntry(r.vcNum, r.startTime, r.runTime, r.outcome,
```
keyboardDrummer, line 331: I see duplication between line 330-337 here and `DafnyConsolePrinter.ReportEndVerifyImplementation`. Could you extract a reusable method from `DafnyConsolePrinter.ReportEndVerifyImplementation` to remove the duplication?

Source/DafnyLanguageServer/Workspace/GutterIconAndHoverVerificationDetailsManager.cs
```diff
@@ -309,13 +309,40 @@ public class GutterIconAndHoverVerificationDetailsManager : IGutterIconAndHoverV
...
+        r.asserts.Select(a => (a.tok, a.Description.SuccessDescription)).ToList(), r.coveredElements,
+        r.resourceCount)
+    ).ToList();
+    var res = new DafnyConsolePrinter.VerificationResultLogEntry(
```
keyboardDrummer, line 335: Something I'd like to do in the future is:

Use CompilationManager for the CLI as well, then the call to `WarnAboutSuspiciousDependencies` in `DafnyDriver` will be subsumed by the call to `WarnAboutSuspiciousDependenciesForImplementation` you have here.

Tweak `VerificationResultLogger` so it uses the `CompilationAfterResolution.ImplementationsPerVerifiable` instead of `DafnyConsolePrinter.VerificationResults` and remove `DafnyConsolePrinter.VerificationResults`

atomb: Yeah, I'd like to do that in the future, too!

keyboardDrummer (approved):

PR #4542, opened 2023-09-12
Proof dependency warnings

atomb:
Adds warnings for two cases of suspicious proof dependencies:

* Warn about a vacuous proof when the goal itself is not a dependency of the proof.
* Warn about any assumption that is not used by any proof in the context where it exists.

A few cases of vacuity are artifacts of how Dafny programs are encoded in Boogie. Several of these cases are automatically filtered out.


Test/logger/ProofDependencyLogging.dfy
```diff
@@ -2,170 +2,170 @@
...
 //
 // CHECK: Results for ObviouslyUnnecessaryRequiresMethod \(correctness\)
 // CHECK:     Proof dependencies:
-// CHECK:       ProofDependencyLogging.dfy\(244,24\)-\(244,24\): value always satisfies the subset constraints of 'nat'
-// CHECK:       ProofDependencyLogging.dfy\(244,47\)-\(244,47\): value always satisfies the subset constraints of 'nat'
+// CHECK:       ProofDependencyLogging.dfy\(244,25\)-\(244,25\): value always satisfies the subset constraints of 'nat'
+// CHECK:       ProofDependencyLogging.dfy\(244,48\)-\(244,48\): value always satisfies the subset constraints of 'nat'
 //
 // CHECK: Results for ObviouslyUnconstrainedCodeFunc \(well-formedness\)
 // CHECK:     Proof dependencies:
```
robin-aws, line 53: Also out of scope for this review, but can't help ask if we should label this more precisely to account for the fact that some proof dependencies may be "spurious", just as counter examples are really something like "potential counter examples"

Test/logger/ProofDependencyWarnings.dfy.expect
robin-aws, line 1: Nice that you got adequate testing coverage by reusing the existing test file!

Test/logger/ProofDependencyWarnings.dfy.expect
```diff
@@ -0,0 +1,20 @@
+ProofDependencyLogging.dfy(176,11): Warning: unnecessary assumption
+ProofDependencyLogging.dfy(186,13): Warning: vacuous proof: assertion always holds
```
robin-aws, line 2: Ah NOW I get my previous confusion - in this specific (and common) case, the `vacuous proof: <proof description>` template leads to a string that's too easy to misinterpret as `vacuous proof, or, as another way of saying 'vacuous proof', assertion always holds`. :)

Perhaps we could use a template like `vacuous proof of (<proof description>)` instead?

Source/DafnyCore/Verifier/ProofDependencyManager.cs
```diff
@@ -17,15 +18,30 @@ namespace Microsoft.Dafny {
...
+      }
+      currentSet.Add(dep);
       return idString;
     }
 
+    public void SetCurrentDefinition(string defName) {
+      currentDefinition = defName;
+    }
+
+    public IEnumerable<ProofDependency> GetPotentialDependenciesForDefinition(string defName) {
```
robin-aws, line 41: Ah there's the "potential" word I was looking for :)

robin-aws (commented):
Partial review, still have to have a look at the main file changes

Source/DafnyCore/DafnyMain.cs
```diff
@@ -203,5 +205,95 @@ public class DafnyMain {
       }
     }
 
+    public static void WarnAboutSuspiciousDependencies(DafnyOptions dafnyOptions, ErrorReporter reporter, ProofDependencyManager depManager) {
```
keyboardDrummer, line 208: I would put this code in a separate file. I think ideally this could would live in a plugin that could live in a separate assembly that hooks into some `PostSymbolVerification` event. Not needed for now but if we can make it look more like that I'd appreciate it.

Test/logger/ProofDependencyWarnings.dfy.expect
robin-aws: We could also consider not using "vacuous" since many users won't be familiar with that concept. Perhaps "Warning: proved from impossible assumptions: ..." ("contradictory assumptions" is appealing too, but then `false` by itself is impossible).

robin-aws (changes requested):
Awesome feature, looking forward to having it on by default in the future :)

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -248,6 +252,12 @@ public enum DefaultFunctionOpacityOptions {
     DafnyOptions.RegisterLegacyBinding(WarningAsErrors, (options, value) => { options.WarningsAsErrors = value; });
     DafnyOptions.RegisterLegacyBinding(VerifyIncludedFiles,
       (options, value) => { options.VerifyAllModules = value; });
+    DafnyOptions.RegisterLegacyBinding(WarnVacuity, (options, value) => {
```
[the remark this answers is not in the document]
atomb: Yeah, it's a bit of a tricky case where we have new, modern options but also need to set a Boogie option. I'd entirely be in favor of some new registry, but I think it's out of scope for this PR.

Source/DafnyDriver/DafnyDriver.cs
```diff
@@ -466,6 +468,10 @@ public enum CommandLineArgumentsResult {
         string baseName = cce.NonNull(Path.GetFileName(dafnyFileNames[^1]));
         var (verified, outcome, moduleStats) = await BoogieAsync(options, baseName, boogiePrograms, programId);
 
+        if (options.TrackVerificationCoverage && verified) {
```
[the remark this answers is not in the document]
atomb: Yeah, it's turned on by `--warn-vacuity` or `--warn-redundant-assumptions`. Dafny users will never need to see the terminology in this option.

Test/logger/ProofDependencyWarnings.dfy.expect
atomb: Yeah, so how about "proved using contradictory assumptions: <proof description>"? I could change the name of `--warn-vacuity`, too. Maybe `--warn-contradictory-assumptions`? That's nicely symmetric with `--warn-redundant-assumptions`.

Source/DafnyCore/DafnyMain.cs
atomb: Will do.

Source/DafnyCore/DafnyMain.cs
```diff
@@ -203,5 +205,95 @@ public class DafnyMain {
...
+          .Except(usedDependencies)
+          .OrderBy(dep => (dep.RangeString(), dep.Description));
+
+      var unusedObligations = unusedDependencies.OfType<ProofObligationDependency>();
+      var unusedRequires = unusedDependencies.OfType<RequiresDependency>();
+      var unusedEnsures = unusedDependencies.OfType<EnsuresDependency>();
+      var unusedAssumeStatements =
+        unusedDependencies
+          .OfType<AssumptionDependency>()
+          .Where(d => d.Description.Contains("assume statement"));
```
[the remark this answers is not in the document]
atomb: Yeah, I'll see whether there's a better way to do this.

Source/DafnyCore/DafnyMain.cs
```diff
@@ -203,5 +205,95 @@ public class DafnyMain {
...
+
+        // Similarly here
+        if (poDep.ProofObligation is MatchIsComplete or AlternativeIsComplete) {
+          return false;
+        }
+
+      }
+
+      // Ensures clauses are often proven vacuously during well-formedness checks.
+      if (verboseName.Contains("well-formedness") && dep is EnsuresDependency) {
```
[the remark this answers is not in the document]
atomb: Unfortunately, I don't know of a good way to change this, much as I don't like it. At this point, we only have access to the Boogie `Implementation` and it doesn't tell us where it came from.

robin-aws (dismissed):
Love the terminology change this led to. Thanks!

docs/dev/news/4542.feat
```diff
@@ -0,0 +1 @@
+Added two new options, `--warn-contradictory-assumptions` and `--warn-redundant-assumptions`, to detect potential problems with specifications that indicate that successful proof may be misleading.
```
robin-aws, line 1:
```suggestion
Added two new options, `--warn-contradictory-assumptions` and `--warn-redundant-assumptions`, to detect potential problems with specifications that indicate that successful verification may be misleading.
```

robin-aws (dismissed):

atomb:
I've noticed some odd behavior in this that I'd like to debug before merging it.

atomb:
The one thing that's clearly amiss is that it doesn't work well with Boogie's `BATCH_MODE` option. It's perhaps still worth merging if that's the only issue.

atomb:
There are also some cases where `--warn-contradictory-assumptions` can lead to false positives if a proof completes by being simplified away, before the solver sees it. Because of this, I've updated the release notes with a caveat and made the options hidden by default. I think it's worth merging, since the feature is still useful when keeping that caveat in mind, but I'll aim to fix it as soon as I can.

robin-aws (approved):

PR #4545, opened 2023-09-13
feat: Allow more assumptions in library backend

robin-aws:
The "library backend" which is selected using `--target:lib` and produces `.doo` files is currently more restrictive than the other backends. The common SinglePassCompiler logic rejects features that cannot be compiled such as `assume` statements without `{:axiom}`, but the library backend doesn't actually use that base class. It instead leverages the `Declaration.Assumptions()` method that the auditor uses to identify assumptions. `AssumptionDescription`s have a `allowedInLibraries` flag to indicate whether they should cause build errors when trying to build a `.doo` file, and the initial version was intentionally very conservative. 

This PR switches the value of this flag for a few cases that are very likely to appear in useful libraries and where the risk of misuse is zero or very low:

* `assume {:axiom} ...` - already allowed by other backends.
* `decreases *` - not really an assumption, more of a sound specification limitation, and we may even remove this from the auditor as well in the future.
* `{:extern}` declarations with `requires` or `ensures` - now allowed only if the declaration also has `{:axiom}`, and this case now creates auditor warnings as well.
* ~`{:termination false}` - allowed because there is no mitigation and traits are highly valuable to share in libraries.~ This is worth discussing, and the better alternative may be to support an assumption attribute on the class extending a trait instead, so that the assumption appears in the consumer instead of the library code. (Update: I reverted the change to allow this attribute in libraries and will implement the alternative instead in a separate PR)

Also corrected a typo in mitigation text suggesting `modifies *` when it should be `modifies {}` - `modifies *` doesn't even parse. :)


Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -25,22 +25,35 @@ bool allowedInLibraries
...
-    issue: "Declaration with [{:extern}] has a requires clause.",
-    mitigation: "Test any external callers (maybe with [/testContracts]).",
-    mitigationIETF: "MUST test any external callers.",
-    isExplicit: false,
-    allowedInLibraries: false);
-  public static AssumptionDescription ExternWithPostcondition = new(
-    issue: "Declaration with [{:extern}] has a ensures clause.",
-    mitigation: "Test external callee (maybe with [/testContracts]).",
-    mitigationIETF: "MUST test external callee.",
+    mitigation: "Turn into a [method] with [modifies {}] or test thoroughly for lack of side effects.",
```
atomb, line 28:
```suggestion
    mitigation: "Turn into a [method] with [modifies {}] and test thoroughly for lack of side effects.",
```

atomb: Turning it into a method allows for non-determinism that may occur, but if we have `modifies {}` then we'll really want to know that it has no side effects.

Source/DafnyCore/Auditor/Assumption.cs
```diff
@@ -25,22 +25,35 @@ bool allowedInLibraries
...
+    mitigationIETF: "SHOULD use [method] with [modifies {}] instead.",
     isExplicit: false,
     allowedInLibraries: false);
+
+  public static AssumptionDescription ExternWithPrecondition(bool hasAxiomAttribute) {
```
atomb, line 33: I know I originally suggested this design, but now I'm thinking this should  be treated as such:
* If an extern _does not_ have `{:axiom}`, have the auditor flag any pre- or post-conditions. Don't allow it in libraries.
* If an extern _does_ have `{:axiom}`, have the auditor flag it as an explicit (intentional) assumption, and allow it in libraries. Don't say anything about pre- or post-conditions.
This is more consistent with treatments elsewhere. Having `{:axiom}` on an enclosing scope (e.g., a declaration) silences mention of other assumptions associated with it (e.g., the contract of a declaration).

atomb (changes requested):

atomb (approved):
Looks good to me!

PR #4548, opened 2023-09-13
Fix help message for the --cores option

ChristopherBrix:
This option accepts both an absolute number and a percentage. The default is based on a percentage, so that should be made obvious.

This only changes what is shown when using `dafny verify --help`, nothing else is affected.


keyboardDrummer:
image

I get an exception if I run this code with a debugger attached.

Did you get any tests to pass with this change?

PR #4549, opened 2023-09-14
Fix gutter icons fields

### LINKED ISSUE #4540: Gutter icons not updating on "var" declarations
### Dafny version

latest nightly

```dafny
method Test() {
  assert true;
}

class A {
  ghost var B: seq<int> // elements between [consumeIndex .. produceIndex[

  var C: array<int>
  var D: int

  method Test() {
    assert true;
  }
}
```

```code
Paste in VSCode
```

### What happened?

image

The code in front of the `var` never says "verified", although since there is nothing to verify, gutter icons should state "verified" by default.

This is a regression as 4.2 did not have this issue, I just tested, so flagging it as release blocker. Too bad it was untested behavior.

Windows

keyboardDrummer:
Fixes #4540

### Changes

Do not create verification trees for non-constant fields that can never trigger verification

### Testing

Added an XUnit test


MikaelMayer (approved):
Excellent fix and test.

PR #4550, opened 2023-09-14
Add missing release note for 4374

keyboardDrummer:


docs/dev/news/4374.fix
```diff
@@ -0,0 +1 @@
+There was a small and a tiny difference between verification in the IDE and the CLI, which would only become apparent for proofs that Z3 had trouble verifying. The small difference has been resolved, while the tiny one still persists, so the IDE and CLI should behave (including resource counts) almost equivalently now.
```
atomb, line 1:
```suggestion
There were two differences between verification in the IDE and the CLI, one small and one tiny, which would only become apparent for proofs that Z3 had trouble verifying. The small difference has been resolved, while the tiny one still persists, so the IDE and CLI should behave almost equivalently now, including resource counts, on most code.  
```

atomb (changes requested):

atomb (approved):

PR #4557, opened 2023-09-15
Do not return exceptions when doing hover in a program with parse errors

keyboardDrummer:
### Changes
- Do not throw an exception when requesting hover in a program with parse errors

### Testing
- Add a test for the above


MikaelMayer (dismissed):
Wow that's a great one I hit every day I believe but wasn't aware of that.

MikaelMayer (approved):

PR #4558, opened 2023-09-15
Add test for prover log regression

keyboardDrummer:
### Changes

Add a test that notifies us when we make changes that change anything about verification, with the caveat that the Dafny input program is not using many Dafny features, so verification changes in specific Dafny features might be missed.


Source/DafnyPipeline.Test/DafnyPipeline.Test.csproj
```diff
@@ -43,5 +43,8 @@
 	  <Content Remove="_plugins/**" />
 	  <EmbeddedResource Include="_plugins/**" />
     <None Remove="_plugins\**" />
+    <None Update="expectedProverLog.txt">
```
atomb, line 46: It's a tiny thing, but I might call this `expectedProverLog.smt2` or `SomeProgram.smt2`.

atomb (approved):
Thanks for doing this! I left one comment, but it's not critical.

Source/DafnyPipeline.Test/ProverLogStability.cs
```diff
@@ -0,0 +1,109 @@
...
+  result := methodFormal == SomeFunc(42);
+}
+";
+
+    public ProverLogStabilityTest(ITestOutputHelper testOutputHelper) {
+      this.testOutputHelper = testOutputHelper;
+    }
+
+    [Fact]
+    public async Task ProverLogRegression() {
```
robin-aws, line 68: I think we need a sizable comment here to describe the purpose of the test. Something like "this test only provides data on when changes are made that change the encoding of Dafny programs to SMT programs. A failure is meant to provide this data, not block changes - if you are contributing a change that makes it fail...". Probably even with some background about the fundamental brittleness problem.

Source/DafnyPipeline.Test/ProverLogStability.cs
```diff
@@ -0,0 +1,109 @@
...
+namespace DafnyPipeline.Test {
+  // Main.Resolve has static shared state (TypeConstraint.ErrorsToBeReported for example)
+  // so we can't execute tests that use it in parallel.
+  [Collection("Singleton Test Collection - Resolution")]
+  public class ProverLogStabilityTest {
+
+    private readonly ITestOutputHelper testOutputHelper;
+
+    // All types of top level declarations.
+    readonly string originalProgram = @"
```
robin-aws, line 22: This is a good start at least, but I feel like it's likely one of the integration test files would cover more features, especially some non-trivial assertion verification.

robin-aws (changes requested):
VERY glad to have this, thanks. Blocking on deeply commenting the purpose here though - otherwise I worry some future contributor will update it blindly or even delete it, thinking it's just a poorly-written brittle unit test.

robin-aws (approved):
Thanks for adding the comment :)

robin-aws (approved):
🚀

PR #4563, opened 2023-09-18
feat: {:assume_concurrent} attribute

robin-aws:
This attribute complements the new verification support for `{:concurrent}`, to encode the assumption that a declaration is safe to execute concurrently despite having non-empty `reads` or `modifies` clauses.


Source/DafnyCore/AST/Members/Function.cs
```diff
@@ -74,6 +74,10 @@ public class Function : MemberDecl, TypeParameter.ParentType, ICallable, ICanFor
       yield return new Assumption(this, tok, AssumptionDescription.ExternWithPrecondition);
     }
 
+    if (Attributes.Contains(Reads.Attributes, Attributes.AssumeConcurrentAttributeName)) {
```
MikaelMayer, line 77: Because functions always rely on all their reads to be immutable from verification point of view, I would actually remove the ability to put {:concurrent} in front of a function.
Methods, on the other hand, can encode non determinism by calling other methods, so for concurrency support, I'd remove functions as potential entry point altogether to avoid pitfalls.

Test/dafny0/ConcurrentAttribute.dfy
```diff
@@ -98,4 +123,14 @@ class Concurrent {
...
+  method {:concurrent} AddToJournal(j: ConcurrentJournal<string>)
+    reads {:assume_concurrent} j
+    modifies {:assume_concurrent} j
+  {
+    j.Add("Today I added another test case.");
+  }
 }
+
+// Auditor warning for {:assume_concurrent}
+// TODO: Release notes!
```
MikaelMayer, line 136: TODO

MikaelMayer (changes requested):
A few details + a big request concerning functions that should not be allowed to be entry points for threads.

Source/DafnyCore/AST/Grammar/Printer.cs
```diff
@@ -1018,6 +1018,9 @@ public class Printer {
 
       int ind = indent + IndentAmount;
       PrintSpec("requires", method.Req, ind);
+      if (options.Get(CommonOptionBag.ReadsClausesOnMethods)) {
```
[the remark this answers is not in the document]
robin-aws: Agreed on both points.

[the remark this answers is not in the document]
robin-aws: Rustan ended up beating me to this one :) although I added an explicit `ReadsOnMethods_Printing.dfy` test to make sure that my check for the option wasn't necessary.

Source/DafnyCore/AST/Members/Function.cs
robin-aws: I'm totally willing to only support `{:assume_concurrent}` on methods, since as you say the risk of allowing non-empty `reads` and `modifies` on functions is greater, and I don't know of any use cases that want it on functions (e.g. `smithy-dafny` never generates any function entry points with `reads` clauses).

But I don't see any reason to not allow a `{:concurrent} function` with an empty `reads` clause, and `smithy-dafny` does generate `function` entry points for operations with the `@readonly` trait.

MikaelMayer (dismissed):

MikaelMayer (approved):

PR #4568, opened 2023-09-19
ci: Disable test reporter auto-discovery to speed up integration tests

robin-aws:
Addresses https://github.com/xunit/xunit/issues/2549 to reduce our integration test shard execution time from almost 90 mins in the worst case to under 30 mins.

The TL;DR is that xUnit has an extension mechanism that attempts to load every class in every available DLL to see if it extends a given interface. This manifests in a huge delay on startup before any tests begin to run, e.g. 52 mins in this case: https://github.com/dafny-lang/dafny/actions/runs/6241813664/job/16944685069#step:22:146 (make sure you turn on Show Timestamps!) This was steadily getting worse as we added more classes, especially https://github.com/dafny-lang/dafny/pull/4438 as it added a bunch of Dafny-generated code.

Fortunately xUnit 2.5.1 added a knob to switch off this discovery, which we're definitely not using.


keyboardDrummer (approved):
Amazing, many kuddos to you for this discovery!

PR #4569, opened 2023-09-20
Fix newtype of alias crash

keyboardDrummer:
### Changes
- Prevent crash from occurring when a newtype refers to an alias of a newtype

### Testing
- Add test that reproduces the crash


robin-aws (approved):
🎉

PR #4570, opened 2023-09-20
Fix thread leak

keyboardDrummer:
### Changes

- Clean up event loop threads to prevent the IDE from using more and more threads
- Some concurrency related hygiene improvements
- Fix the test `DoNotResendAfterNoopChange`, that for a reason I couldn't discover was passing when it shouldn't have.

### Testing

Added test that detects the issue


MikaelMayer (approved):
Good job for finding a test!

MikaelMayer (approved):
These changes look like an experimentation to enhance thread stability, so if the tests pass I'm ok with them.

MikaelMayer (approved):

PR #4571, opened 2023-09-20
Prevent modification exception when accessing parsed document

keyboardDrummer:
Due to a large rename commit, it might be best to review this PR by reviewing the individual commits.

### Changes
- Prevent the document symbol API from crashing, especially when prefix name modules are used. The server back-end API provides both a Dafny program after parsing and after resolution. Without this PR, the same program object is used, and the program you get after parsing will still be modified by resolution while you access it. This PR clones the program before it is used for resolution, so whether you get a program post parsing or post resolution, it won't be modified any more.

### Testing
- Update tests for document symbol API so they no longer with for the program to resolve, which prevented this bug from being detected. One of these tests now reliably fails without this fix.


robin-aws (approved):
Actual change LGTM. Slight nit pick with the test method renames, because to me now it suggests `CreateAndOpenTestDocument` only opens the document without even starting to resolve, when in reality it just doesn't wait for it to finish. You could consider renaming the other method `CreateOpenAndWaitForResolutionOfTestDocument`.

It also seems like the tests that only use `CreateAndOpenTestDocument` will be inherently flaky, by relying on racing to test the kind of bug you're fixing here. Is it possible to literally only open the document and then take other actions before triggering resolution?

Don't want to block merging on either point though.

keyboardDrummer:
> Slight nit pick with the test method renames, because to me now it suggests CreateAndOpenTestDocument only opens the document without even starting to resolve, when in reality it just doesn't wait for it to finish. You could consider renaming the other method CreateOpenAndWaitForResolutionOfTestDocument.

Did the rename

> It also seems like the tests that only use CreateAndOpenTestDocument will be inherently flaky, by relying on racing to test the kind of bug you're fixing here. 

The bug that this PR fixes is caused by a race condition, so there is no easy way to reliably detect it, although there was a test that seemed to detect it with a seemingly 100% chance.

> Is it possible to literally only open the document and then take other actions before triggering resolution?

No, opening the document will start its analysis

robin-aws (approved):

PR #4573, opened 2023-09-21
ci: Split off reusable nightly build workflow, run it for master and 4.3

robin-aws:
This addresses the fact that the nightly build only publishes prereleases out of `master`, even when we have a different branch containing the content of the next planned release. See comment on the top of `nightly-build.tml` for details.

Also addressed a couple of incidental issues, renaming `sha` to `ref` in a few places to clarify use, and passing the correct branch to `check-deep-tests-reusable.yml`.

Since these changes are notoriously difficult to test properly before merging, I'd recommend only merging this well ahead of the scheduled trigger time (14:30 UTC) and triggering it manually with the `nightly-build-manual.yml` workflow.

Feel free to update the target branch from `4.3` to `main-4.3` if that renaming happens before this is merged.

### DIFF
--- a/.github/workflows/doc-tests.yml
+++ b/.github/workflows/doc-tests.yml
@@ -19,7 +19,7 @@ jobs:
   check-deep-tests:
      uses: ./.github/workflows/check-deep-tests-reusable.yml
      with:
-       branch: master
+       branch: ${{ github.base_ref }}
 
   doctests:
     needs: check-deep-tests
--- a/.github/workflows/integration-tests-reusable.yml
+++ b/.github/workflows/integration-tests-reusable.yml
@@ -9,6 +9,9 @@ on:
       num_shards:
         required: true
         type: number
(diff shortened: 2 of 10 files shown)


keyboardDrummer (approved):
Sweet

keyboardDrummer:
> Feel free to update the target branch from 4.3 to main-4.3 if that renaming happens before this is merged.

Would renaming to `main-4.3` be to follow a convention, or is there some code relying on that pattern?

keyboardDrummer:
> Since these changes are notoriously difficult to test properly before merging, I'd recommend only merging this well ahead of the scheduled trigger time (14:30 UTC) and triggering it manually with the nightly-build-manual.yml workflow.

Given the current time and that I had to merge master into this PR so it's first rebuilding, I don't think the merge timing will work out well, but we can always revert and manually trigger nightly :-)

PR #4574, opened 2023-09-21
Allow for user-defined entity creation in test generation

### LINKED ISSUE #4543: Allow for user-defined entity creation in test generation
### Summary

Allow the user to specify how to generate a given argument for the test entry method.

### Background and Motivation

Often the tests are being generated in order to test a Dafny model against an external piece of software: if there is no branch in the model on a given attribute, we may miss an important discrepancy.

Having this kind of hook into the selection of constraints allows to see such possible "blind spots".

### Proposed Feature

The `generate-tests` option typically uses the "default" `witness` when trying to inhabit a type, if no additional constraints force it to be otherwise.

My recommendation is this: when given an input parameter $a: A$ to the harness, say

```
    method {:testEntry} Foo(a: A, b: B) {
        ...
     }
```

Allow for a function that produces a list of `A`s to be used at run-time, say

```
   method GenA() returns (out: seq<A>)
```

and

1. Generate tests as usual on the constraint side
2. Mark the entry as `method {:testEntry}{:useGen "a" GenA} Foo(a: A, b: B)` (or something)
3. At test emission time, replace `var a := whatever` by

```
    var aList := GenA();
    var a := PickUniform(aList);
```
where `PickUniform` is an `extern` which picks an element of the list uniformly at random (one can easily imagine fancier selection strategies of course).

I remember discussing such a feature with @Dargones at some point but I do not remember the status.

Finally, one may want an option to only invoke this generator if `a` is the "default" value for `A`, or generate this test *in addition* to all the other tests.

### Alternatives

_No response_

Dargones:
Fixes #4543

This PR lets users specify how to generate a given argument for the test entry method  using a `:testGenerators` attribute. Here is the description of the attribute I propose to put in the Reference Manual: 

This attribute instructs Dafny test generation to override an input parameter with a value obtained by calling the specified generator function or method. A method annotated with `{:testGenerators}` must always be a test entry method.

For example, consider the following Dafny code:

```dafny
module M {
  method {:extern} CoinFlip() returns (b:bool)
  method {:testEntry} {:testGenerators "b", "M.CoinFlip"} ToTest(b: bool) {
    // Do Something Here
  }
}
```

In the code snippet above, method `ToTest` is annotated with  `{:testGenerators "b", "M.CoinFlip"}` which means that Dafny test generation  will produce a test that calls the`M.CoinFlip` method to obtain a value for  the input parameter `b`. In particular, running  `dafny generate-tests Block FILE.dfy` on the file above, will produce the following:

```dafny
method {:test} Test0() {
  var bool0 : bool := M.flipCoin();
  M.m(bool0);
}
```


keyboardDrummer:
Why wouldn't you just write 

```
method {:test} Test0() {
  var bool0 : bool := M.flipCoin();
  M.m(bool0);
}
```

Instead of 
```
{:testGenerators "b", "M.CoinFlip"}
```

?

keyboardDrummer (changes requested):
Why wouldn't you just write 

```
method {:test} Test0() {
  var bool0 : bool := M.flipCoin();
  M.m(bool0);
}
```

Instead of 
```
{:testGenerators "b", "M.CoinFlip"}
```

?

keyboardDrummer:
> The goal is to not modify the code under test 

A wrapper test method such as

```
method {:test} Test0() {
  var bool0 : bool := M.flipCoin();
  M.m(bool0);
}
```

does not modify the code under test

> and at the same time to preserve `M.m` as an entry point for the tests (in particular so that it can be easily substituted with a similarly named method in the implementation to which Dafny is being compared). 

Why do want `M.m` to be the entry point? Even if it is not, could you not still substitute it with a similarly named method in the implementation to which Dafny is being compared?

robin-aws:
If I'm following correctly, the motivation is clearer once you have more than one input parameter and want to allow test generation to control what other arguments are passed in. Runnin `dafny generate-tests Block FILE.dfy` on this:

```dafny
module M {
  method {:extern} CoinFlip() returns (b:bool)
  method {:testEntry} {:testGenerators "b", "M.CoinFlip"} ToTest(b: bool, c: bool) {
    // Do Something Here
  }
}
```

Would produce something like:

```dafny
method {:test} Test0() {
  var bool0 : bool := M.flipCoin();
  M.m(bool0, true);
}

method {:test} Test1() {
  var bool0 : bool := M.flipCoin();
  M.m(bool0, false);
}
```

In other words, `{:testGenerators}` is useful for controlling how tests are generated, so you don't have to do a ton of post-processing on the generated tests to make them runnable/practical.

Dargones:
Using a wrapper method in this way makes sense to me when the goal is to test the Dafny code directly. 

That said, I do see a potential issue when the goal is to perform model-based testing, i.e. when we want to generate tests from the Dafny model to run on another implementation written in a different language. Ideally, we would want to (1) generate the tests, (2) switch the call to the method from which we generated the tests to the method that will be tested at runtime, and (3) run the tests. The use of a wrapper method complicates things in this scenario because now the switch has to happen within the wrapper method itself meaning that the user has to constantly modify the wrapper method between generating tests and running them. 

I do like the wrapper solution more because it does not require any new features but I am wondering to what extent it is feasible in practice in the model-based testing scenario. @codyroux, do you have a take on this?

PR #4584, opened 2023-09-26
Rename verification progress reporter

keyboardDrummer:
### Changes
- Rename a class and interface to make it clear what to use them for


MikaelMayer (approved):
Straightforward renaming that reduces the scope of the so-called verification progress reporter, but given that you did an amazing work with implementation views, it makes sense.

PR #4585, opened 2023-09-26
Use async main methods to avoid waiting for tasks at the top level

keyboardDrummer:
### Changes
- Update the legacy Dafny server code so it does not refer `Console.Out/Error/In`, so its can be run in parallel next to other Dafny tests.
- Change the `CoverageReport` code so it does not depend on a static counter, which could cause issues when running multiple Dafny instances in parallel.
- Change the legacy parser code so it does not print error message to `Console.Error`, which causes issues when running multiple Dafny instances in parallel.
- Add a timeout to individual littish tests of 5 minutes. Currently a stuck littish test will cause the entire test suite to hang until it times out at the Github actions level, which we've configured to 120 minutes, after which we get no indication of which test it was stuck on !!
- Refactoring of littish testing code
- Remove two occurrences of `#pragma warning disable VSTHRD002` by relying on async Main methods. An effect of this change might be that Dafny will use fewer threads. I was triggered to make this change when I was debugging Dafny and noticed some threads always waiting at the locations where this PR removes the Wait.

### Testing
This is a refactoring which is covered by existing tests.


robin-aws:
Comment as a spin-off from https://github.com/dafny-lang/dafny/pull/4798 - can we also add a link to an issue on the remaining pragmas this PR doesn't end up resolving?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -188,16 +188,79 @@ public class DafnyOptions : Bpl.CommandLineOptions {
...
+            } else {
+              UnknownSwitch(state);
+            }
+          }
+        } else {
+          AddFile(file, state);
+        }
+      }
+      if (state.EncounteredErrors) {
+        ErrorWriter.WriteLine("Use /help for available options");
```
MikaelMayer, line 257:
```suggestion
        ErrorWriter.WriteLine("Use --help for available options");
```
Since we can now customize this error message, I thought we could as well support the new CLI by default.

Source/DafnyDriver/Commands/TestCommand.cs
```diff
@@ -50,4 +50,6 @@ static class TestCommand {
     });
     return result;
   }
+
+  public static void EnsureStaticConstructorHasRun() { }
```
MikaelMayer, line 54: Wait, is it possible the static constructor to not run if you don't call this method?

Source/DafnyDriver/CoverageReporter.cs
```diff
@@ -31,7 +31,22 @@ public class CoverageReporter {
...
+  private readonly Dictionary<(CoverageReport, string), string> paths = new();
+
+  public string GetPath(CoverageReport report, string desiredPath) {
+    return paths.GetOrCreate((report, desiredPath), () => {
+      var index = 0;
+      var extension = Path.GetExtension(desiredPath);
+      var withoutExtension = Path.GetFileNameWithoutExtension(desiredPath);
+      var actualPath = desiredPath;
+      while (File.Exists(actualPath)) {
+        actualPath = withoutExtension + "_" + index + extension;
```
MikaelMayer, line 43: This looks like duplicate code from https://github.com/dafny-lang/dafny/pull/4585/files#diff-8f346d4cacb8d208b553e6e51abc34de1f262354c3901eff9416e148f0c6e19bR25
could you perhaps consolidate and define this code only once?

Source/IntegrationTests/TestFiles/LitTests/LitTest/logger/ProofDependencies.dfy_combined.html.expect
```diff
@@ -10,7 +10,7 @@
 <body onload="window['PR_TAB_WIDTH']=4">
 <div class="menu" id="menu">
     <a href="./index_combined.html">Index</a>
-    <a href="ProofDependencies.dfy_tests_expected_2.html" class="el_report">Expected Test Coverage (coverage_testing)</a><a href="ProofDependencies.dfy_verification_3.html" class="el_report">Verification coverage (coverage_verification)</a>
+    <a href="ProofDependencies.dfy_tests_expected.html" class="el_report">Expected Test Coverage (coverage_testing)</a><a href="ProofDependencies.dfy_verification.html" class="el_report">Verification coverage (coverage_verification)</a>
```
MikaelMayer, line 13: Oh nice, avoiding the global counter makes sure files with a different name don't get a useless suffice. Great!

MikaelMayer (commented):
Looks good!

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.Dafny;
 
 public class CommonOptionBag {
 
+  public static void EnsureStaticConstructorHasRun() { }
```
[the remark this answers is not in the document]
keyboardDrummer: What the name says it does. If you call a method in a class, dotnet will first ensure the class's static constructor was run.

Source/DafnyDriver/Commands/TestCommand.cs
keyboardDrummer: If you don't use a class, then it's static constructor is not run.

Source/Dafny/Dafny.cs
```diff
@@ -1,9 +1,10 @@
+using System.Threading.Tasks;
 using Microsoft.Dafny;
 
 namespace Dafny;
 
 public class Dafny {
-  public static int Main(string[] args) {
+  public static Task<int> Main(string[] args) {
```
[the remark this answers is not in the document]
keyboardDrummer: No idea ! I'm pretty sure it won't exist before the Task is done though :-D

Source/DafnyCore/DafnyOptions.cs
keyboardDrummer: I'm not sure. This is printed when you're already using the old CLI.

MikaelMayer (approved):
I was not expecting to delete the duplicate code, rather put it at the same place, but it looks like you figured out it was not necessary.

PR #4587, opened 2023-09-27
Make `GetLastDiagnostics` more reliable and reduce usage of `GetLastDocumentAsync`

keyboardDrummer:
### Changes
- Update `GetLastDiagnostics` so it doesn't depend on internal language server state, which makes it fragile. In another PR I ran into it not behaving well, but I didn't want to make in these changes with that one.
- Remove several calls to `GetLastDocumentAsync`, is a language server API that waits until all verification has completed, and then returns the diagnostics resulting from that. This API is primarily kept alive for the tests, and requires rather complicated logic on the server. The main driver of this change is that `GetLastDiagnostics` which uses `GetLastDocumentAsync` wasn't entirely reliable. I ran into trouble using it in another PR.

### Testing
Most tests didn't need to be updated for these changes


Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
```diff
@@ -87,18 +87,18 @@ public class NotificationPublisher : INotificationPublisher {
     }
 
     private CompilationStatus GetGlobalProgress(IdeState state) {
-      var hasResolutionDiagnostics = state.ResolutionDiagnostics.Values.SelectMany(x => x).
-        Any(d => d.Severity == DiagnosticSeverity.Error);
+      var errors = state.ResolutionDiagnostics.Values.SelectMany(x => x).
+        Where(d => d.Severity == DiagnosticSeverity.Error).ToList();
       if (state.Compilation is CompilationAfterResolution) {
-        if (hasResolutionDiagnostics) {
+        if (errors.Any(d => d.Source == MessageSource.Resolver.ToString())) {
```
MikaelMayer, line 93: Minor thing: Don't we have a way to cache all these .ToString() ? I hope they are compiled as constant strings because otherwise it feels it should be refactored in a constant somewhere.

MikaelMayer (approved):
Looks like minor refactorings, so tests passing are ok for me.

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
keyboardDrummer: > I hope they are compiled as constant strings

Seems likely to me

Issue #4322, opened 2023-07-21
Feature request: specifying the target language parent namespace when compiling


robin-aws:
Projects that want to embed compiled Dafny code inside libraries will frequently want to isolate the compiled code under a specific namespace. [For example](https://github.com/aws/aws-cryptographic-material-providers-library-java/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/Model/AwsCryptographyMaterialProvidersTypes.dfy#L9):

```dafny
module {:extern "software.amazon.cryptography.materialproviders.internaldafny.types" } AwsCryptographyMaterialProvidersTypes
```

The challenge with using `:extern` for this is that it's not realistic to expect a single symbol to make sense for all target languages. Several suggestions have been made to allow specifying extern names per language somehow:

* https://github.com/dafny-lang/dafny/issues/2397
* https://github.com/dafny-lang/dafny/issues/3030
* https://github.com/awslabs/smithy-dafny/issues/45

For the specific case of just relocating the compiled code, however, it would be a lot simpler to just provide a prefix namespace when building the code on a per-language basis. This would be adequate for `smithy-dafny`, since the user-facing target language symbols are generated from Smithy models instead anyway, so there is actually no need for precise control over extern attributes.

As a strawperson proposal, this could just be an option like `--compile-prefix "software.amazon.cryptography.materialproviders.internaldafny"`, although it may be poor UX to have a single term across all languages, even if the build or configuration of a project can provide different values for each target language.

PR #4591, opened 2023-09-29
Allow specifying an outer namespace when translating

keyboardDrummer:
Fixes https://github.com/dafny-lang/dafny/issues/4322

### Changes
- Add the option `--outer-namespace`: `Nest all code in this module. Can be used to customize generated code. Use dots as separators (foo.baz.zoo) for deeper nesting. The first specified module will be the outermost one.`

### Testing
- Add test that verifies the behavior of `--outer-namespace` for C#, Java, Go, Python and JavaScript
- For Python, modules are currently (already before this PR) emitted using a flat directory structure, even though more idiomatic might be to have every module define a folder, and put the module contents in `__init__.py`


Source/DafnyCore/Plugins/IExecutableBackend.cs
```diff
@@ -169,4 +171,10 @@ public abstract class IExecutableBackend {
   /// Instruments the underlying SinglePassCompiler, if it exists.
   /// </summary>
   public abstract void InstrumentCompiler(CompilerInstrumenter instrumenter, Program dafnyProgram);
+
+  public virtual string JoinNamespaceParts(IEnumerable<string> parts) => string.Join(".", parts);
```
robin-aws, line 175: One of many bits that goes away if each backend just takes a single-valued option instead of a sequence of parts.

Source/TestDafny/MultiBackendTest.cs
```diff
@@ -94,10 +96,10 @@ record ResolutionSetting(
     int ExpectedExitCode
   );
 
-  private int ForEachCompiler(ForEachCompilerOptions options) {
-    var parseResult = CommandRegistry.Create(TextWriter.Null, TextWriter.Null, TextReader.Null,
-      new string[] { "verify", options.TestFile! }.Concat(options.OtherArgs).ToArray());
-    var dafnyOptions = ((ParseArgumentSuccess)parseResult).DafnyOptions;
+  private async Task<int> ForEachCompiler(ForEachCompilerOptions options) {
```
robin-aws, line 99: Out of curiousity why all the new `async` introductions?

Source/DafnyCore/Plugins/IExecutableBackend.cs
keyboardDrummer: Well... actually something like this could reduce a lot of duplication in the back-ends, since they have a lot of `s1 + "." + s2` copied all over the place. But I've removed it now.

Source/TestDafny/MultiBackendTest.cs
keyboardDrummer: This called `Task<int> DafnyCli.Execute(..)` before, so it needed to, but even without that call this should be async style. Right now it's 50% using async which is confusing, but I have another PR that takes it all the way there, so I wouldn't worry too much about it.

Source/DafnyCore/Plugins/IExecutableBackend.cs
robin-aws: Oh I don't doubt the utility is useful in many other places, just the wrong place for it. :)

Test/comp/globalNamespace/globalNamespaceRunConfig.toml
```diff
@@ -0,0 +1,5 @@
+includes=["./globalNamespaceRun.dfy"]
+
+[options]
+outer-namespace=["my","happy"]
```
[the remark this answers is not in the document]
robin-aws: Apologies, I accidentally some important words in that sentence :). I meant "Also I'm assuming we'll need a way to specify options that are only provided when compiling to a given backend <**in project files**>"

> If you use a project file, then all options in the project file that apply to the command you're using are applied, so you can have

That's a bit dicey because it assumes the backend specific options are all uniquely named, but I bet more than one backend will end up naturally providing `--outer-module`. It also won't hold up as a general method for providing backend-specific options for the same reason.

As I said though I think it's fine to leave this out of scope for this PR - users can still use project files for common options but provide these backend specific ones when invoking the CLI, such as `dafny translate cs dfyconfig.toml --outer-namespace My.Happy`

Test/comp/ExternJavaString.dfy
```diff
@@ -3,8 +3,11 @@
 // In this example, the extern method obtains a Java string and returns it as such.
 // The Dafny code converts that Java string to a Dafny string.
 
-class {:extern "java.lang.String"} JavaString {
```
robin-aws, line 6: Did this have to change? Why?

Source/DafnyCore/Options/DafnyProject.cs
robin-aws, line 1: Just checking my understanding - looks like these changes are to allow array-valued options in project files, which is a good improvement to leave in the PR but not actually used by the new options any more (since they are back to single-values), correct?

Source/IntegrationTests/LitTests.cs
```diff
@@ -239,7 +239,7 @@ public class LitTests {
               return false;
             }
 
-            if (arguments.Any(arg => arg is "/compile:3" or "/compile:4" or "run" or "translate")) {
+            if (arguments.Any(arg => arg is "/compile:3" or "/compile:4" or "run")) {
```
robin-aws, line 242: This doesn't seem right to me, I added "translate" because there were important backend-specific tests that weren't being run for all backends before that.

Test/comp/globalNamespace/globalNamespaceRunConfig.toml
robin-aws, line 4: This should be "my.happy" now, shouldn't it?

robin-aws (changes requested):
Mostly looks great, and nice that the actual implementation is pretty simple (especially compared to #4601 :)

Looks like this is just for Java and C#, which is fine for now, but will will need the other backends at some point. Make sure we either keep #4322 open or cut fresh issues for the remaining backends. The PR description is also a bit out of date.

Source/DafnyCore/Options/DafnyProject.cs
keyboardDrummer: Indeed

Source/IntegrationTests/LitTests.cs
keyboardDrummer: How can you test `dafny translate` without back-end specific code? Testing `dafny translate` implies to me that you want to be back-end specific, which is also true for the tests in this PR.

Test/comp/globalNamespace/globalNamespaceRunConfig.toml
keyboardDrummer: Test no longer applies because you're can't specify --outer-namespace when doing `dafny run`. I know that MAYBE we want to enable this, but for now I won't since the use-case is underspecified.

Test/comp/ExternJavaString.dfy
keyboardDrummer: Reverted this change. Also had to redo the entire implementation for it :-D

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -470,7 +471,9 @@ public enum DefaultFunctionOpacityOptions {
       IncludeRuntimeOption,
       WarnContradictoryAssumptions,
       WarnRedundantAssumptions,
-      DefaultFunctionOpacity
+      DefaultFunctionOpacity,
+      JavaBackend.OuterPackage,
+      CsharpBackend.OuterNamespace
```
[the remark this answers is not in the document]
keyboardDrummer: I have trouble following. You think we MAY want to let something work in way X, so now we should already do X_part1 ? Why not wait until we know what we want? Right now `.doo` doesn't contain translated code or a derivative, so what sense is there in including translation options? If you change `.doo` so it contains more, than we can also add the options.

If you include `.doo` in something else that contains translated code, then you have no guarantee that the .doo file was generated with the same translation options as the generated code.

I would even go so far as to say translation options shouldn't be available when building a .doo file, so they will be ignored from your project file etc.

Source/IntegrationTests/LitTests.cs
robin-aws: No I couldn't convert them to `%testDafnyForEachCompiler` as you say, but I did manually add missing backend cases. https://github.com/dafny-lang/dafny/blob/master/Test/comp/separate-compilation/usesTimesTwo.dfy for example.

The name `NeedsConverting()` is a bit out of date, since the idea is just to identify tests that use backends but don't test all backends.

keyboardDrummer: What's the use of adding `// NONUNIFORM: using dafny translate` to all tests that exercise `dafny translate` ?

Source/DafnyCore/Options/CommonOptionBag.cs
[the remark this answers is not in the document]
robin-aws: That's totally fair. I'm just aiming to at least keep track of all the options that affect compilation output in a way that can break separate compilation, for when we do have the mechanism (and I'm working on that as part of adding standard libraries). Are you okay with putting these options in the `RegisterLibraryChecks` list with a comment for now?

I don't feel that strongly about it mind you. I should probably do another pass of reviewing all options anyway rather than trusting the existing lists.

Source/IntegrationTests/LitTests.cs
keyboardDrummer: I thought `NONUNIFORM` meant you tests some back-ends but not others, but now you're suggesting it will mean that you're not using `%testDafnyForEach...`

Source/DafnyCore/Options/CommonOptionBag.cs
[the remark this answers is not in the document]
keyboardDrummer: > I should probably do another pass of reviewing all options anyway rather than trusting the existing lists.

I'd prefer that if it's OK with you. "Design for today" was what my first programming mentor always said.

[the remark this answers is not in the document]
keyboardDrummer: Also I'm skeptical it makes sense to force translation options to be the same between different Dafny components that are integrating. There's also down-the-line compilation options, whatever you pass to other compilers, that might also influence something, but I feel that is outside of the scope of Dafny.

Dafny will ensure that Dafny's semantics are present when executing your program, but other things like the maximum stack size used when executing, are not Dafny's responsibility.

Source/IntegrationTests/LitTests.cs
robin-aws: It's just so that we can make the conscious choice to either switch to `%testDafnyForEachCompiler` (ideally) or document that the test can't be expressed in a way that automatically works across all backends. Adding `// NONUNIFORM` means the scan won't complain about the test any more, but we can periodically audit the non-uniform tests and at least manually add new backends, or find ways of making them uniform.

Source/DafnyCore/Options/CommonOptionBag.cs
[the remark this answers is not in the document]
robin-aws: Yup, not a problem. I'll take responsibility for the mechanism to track separate compilation compatibility as part of the standard library work.

Source/IntegrationTests/LitTests.cs
keyboardDrummer: > It's just so that we can make the conscious choice to either switch to %testDafnyForEachCompiler (ideally) or

For `translate` you can't switch to that so I don't see a choice.

> but we can periodically audit the non-uniform tests and at least manually add new backends, or find ways of making them uniform.

You can do that by looking for `dafny translate` !

keyboardDrummer:
Switched back to a unified option for all back-ends, since the new implementation guarantees a unified interface for all back-ends.

Source/IntegrationTests/LitTests.cs
robin-aws: > You can do that by looking for dafny translate !

The idea was to have `// NONUNIFORM` on all tests that don't automatically pick up new compilers, regardless of why. We could modify https://github.com/dafny-lang/dafny/blob/master/.github/workflows/msbuild.yml#L59-L60 to look for `dafny translate` as well if you want.

keyboardDrummer: > The idea was to have // NONUNIFORM on all tests that don't automatically pick up new compilers, regardless of why

What's the use? The reason you gave was that so you can "periodically audit the non-uniform tests and at least manually add new backends", but you don't need this semantic comment for that. Any test that specifies a target, either `--target`, or using `translate`, is non-uniform.

What I think is useful, is to complain about usages of `dafny run` or `dafny build` that don't use `--target`, because they are uniform and could be tested against all back-ends, so it's wasted tested opportunity not to do so.

Source/DafnyCore/Plugins/IExecutableBackend.cs
```diff
@@ -171,6 +183,17 @@ public abstract class IExecutableBackend {
   /// </summary>
   public abstract void InstrumentCompiler(CompilerInstrumenter instrumenter, Program dafnyProgram);
 
+  public static readonly Option<string> OuterModule =
+    new("--outer-module", "Nest all code in this module. Can be used to customize generated code. Use dots as separators (foo.baz.zoo) for deeper nesting. The first specified module will be the outermost one.");
```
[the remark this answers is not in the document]
keyboardDrummer: It was not. Doesn't it help to indicate which module will be the outermost one?

Source/IntegrationTests/LitTests.cs
keyboardDrummer: btw, do you know why a file like this one https://github.com/dafny-lang/dafny/blob/master/Test/c%2B%2B/extern.dfy#L1 isn't flagged by the check?

robin-aws: I don't offhand, I'll look into it separately from this PR. And that means for now I won't object to this change. :)

robin-aws: After offline conversation I realize this is a good change - // NONUNIFORM should only be used for tests of backend-specific behaviour, like ExternJavaString.dfy. as part of my action item I'll remove the tag from tests like that.

robin-aws (approved):

PR #4593, opened 2023-09-29
Chore: Updated the script docs/make-snapshot for Windows

MikaelMayer:
I executed the script step by step and fixed issues


robin-aws (approved):
Thanks for updating this. The changes look good - although I don't have high confidence that it will still work on non-windows systems, and it's unfortunate we don't have any automated testing, it seems better to make this change rather than know it doesn't work on Windows. Perhaps we can find a way to work this process into the nightly prerelease build as well in the future.

PR #4597, opened 2023-09-29
Monomorphic Field and collection types

zafer-esen:
Removes the type argument from `Field`, and from all collection types. All heap and collection elements are now always of type `Box`.

This has the effect of very slightly increasing brittleness on some particularly brittle examples, though it has minimal effect on most code. The benefits of the simplification (and the opportunities it opens for future optimization) seem worth the tradeoff, however.


RustanLeino:
Since we have the information, perhaps we can write the type in a comment. It might be useful during debugging some day in the future.

zafer-esen:
That's a good idea! A lot of the work in this PR went into trying to figure out what the types should be at each Box / Unbox location.

Another thing that can probably be improved is that there are too many functions for boxing & unboxing in `Translator.cs`  (and I introduced more in this PR); it is not very clear what the differences between them are.

Source/DafnyCore/DafnyPrelude.bpl
```diff
@@ -188,26 +190,26 @@ axiom (forall bx : Box ::
 // generated programmatically. Except, Bv0 is given here.
 axiom (forall bx : Box ::
     { $IsBox(bx, TBitvector(0)) }
-    ( $IsBox(bx, TBitvector(0)) ==> $Box($Unbox(bx) : Bv0) == bx && $Is($Unbox(bx) : Set Box, TBitvector(0))));
+    ( $IsBox(bx, TBitvector(0)) ==> $Box($Unbox(bx) : Bv0) == bx && $Is($Unbox(bx) : Bv0, TBitvector(0))));
```
MikaelMayer, line 193: This rewriting is surprising, I would have expected this instead:
```suggestion
    ( $IsBox(bx, TBitvector(0)) ==> $Box($Unbox(bx) : Bv0) == bx && $Is($Unbox(bx) : Set, TBitvector(0))));
```

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -239,25 +239,9 @@ public class DafnyModel {
     /// <summary>
     /// Return the name of a 0-arity type function that maps to the element if such
     /// a function exists and is unique. Return null otherwise.
-    /// If the name is also aliased by a type parameter, return the name of the concrete type. 
     /// </summary>
     private static string GetTrueTypeName(Model.Element element) {
-      string name = null;
```
MikaelMayer, line 245: What's happening  with the test `(name == null) || name.Contains("$") || name.StartsWith("#") || name.Contains("@"`? Why is it no longer necessary?

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
```diff
@@ -54,27 +54,27 @@ public class DafnyModel {
...
       };
       fSetSelect = ModelFuncWrapper.MergeFunctions(this, new List<string> { "MapType0Select", "MapType1Select" }, 2);
-      fSeqLength = new ModelFuncWrapper(this, "Seq#Length", 1, tyArgMultiplier);
-      fSeqBuild = new ModelFuncWrapper(this, "Seq#Build", 2, tyArgMultiplier);
-      fSeqAppend = new ModelFuncWrapper(this, "Seq#Append", 2, tyArgMultiplier);
-      fSeqDrop = new ModelFuncWrapper(this, "Seq#Drop", 2, tyArgMultiplier);
-      fSeqTake = new ModelFuncWrapper(this, "Seq#Take", 2, tyArgMultiplier);
-      fSeqIndex = new ModelFuncWrapper(this, "Seq#Index", 2, tyArgMultiplier);
-      fSeqUpdate = new ModelFuncWrapper(this, "Seq#Update", 3, tyArgMultiplier);
+      fSeqLength = new ModelFuncWrapper(this, "Seq#Length", 1, 0);
```
MikaelMayer, line 57: What is the tyArgMultipler?

Source/DafnyCore/Verifier/BoogieGenerator.ExpressionWellformed.cs
```diff
@@ -1537,7 +1537,7 @@ private record ReadsCheckDelayer(ExpressionTranslator etran, Function selfCallsA
         // assume (forall i0,i1,i2,... :: { nw[i0,i1,i2,...] }
         //            0 <= i0 < ... && ... ==> nw[i0,i1,i2,...] == init.requires(i0,i1,i2,...));
         var ai = ReadHeap(tok, etran.HeapExpr, nw, GetArrayIndexFieldName(tok, indices));
-        var ai_prime = UnboxIfBoxed(ai, elementType);
+        var ai_prime = TrType(elementType) == predef.BoxType ? ai : ApplyUnbox(tok, ai, TrType(elementType));
```
MikaelMayer, line 1540: I liked having only one function here, even if it was used only once. Would you consider re-adding it?

Source/IntegrationTests/TestFiles/LitTests/LitTest/dafny0/ReadsOnMethods.dfy.expect
```diff
@@ -16,7 +16,6 @@ ReadsOnMethods.dfy(183,25): Error: insufficient reads clause to read field
 ReadsOnMethods.dfy(245,9): Error: insufficient reads clause to read field
 ReadsOnMethods.dfy(304,33): Error: insufficient reads clause to read field
 ReadsOnMethods.dfy(308,22): Error: insufficient reads clause to invoke function
-ReadsOnMethods.dfy(313,13): Error: call might violate context's modifies clause
```
MikaelMayer, line 19: I wouldn't have expected that this error would go away. Can you explain a bit?

MikaelMayer (commented):
A few comments, otherwise, I like this simplification. Polymorphic types are not necessary for verification and often go against it. I hope this will make maintenance easier.

Source/DafnyCore/DafnyPrelude.bpl
atomb: Huh. That is weird. I'll give it a closer look.

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
atomb: It specifies how many arguments exist in the SMT-Lib function definitions for each type argument. At the moment, it's either 0 or 1, depending on the value of `TypeEncoding`. This seemed easier than having conditionals everywhere.

Source/IntegrationTests/TestFiles/LitTests/LitTest/dafny0/ReadsOnMethods.dfy.expect
atomb: Ultimately, the issue is that Boogie isn't guaranteed to report all errors in a given procedure. If it determines that some code is dead code (because a previous assertion was equivalent, in the current context, to `assert false`, for example), it may not even generate verification conditions for that part of the code.

atomb: I checked with @robin-aws, who wrote this test, and this error wasn't an essential one for the test to cover. There's another test, for the concurrent attribute, that I refactored so that the two different errors occurred in different definitions and were therefore guaranteed to be reported.

Source/DafnyLanguageServer/CounterExampleGeneration/DafnyModel.cs
atomb: Good question. I'll have to look at it more carefully. I'm not sure off the top of my head.

zafer-esen: I think it is no longer possible for `name` to have these special characters after monomorphization.  I remember asking @Dargones to review the changes in this part of the PR as they affect test generation, and if I recall correctly he mentioned that he tested them on his branch and were ok.

Source/DafnyCore/DafnyPrelude.bpl
zafer-esen: I thought `Set Box` looked like a copy-paste error, but maybe I am missing something...

MikaelMayer: You're probably right, but this mean that this particular type cast was probably untested then until now.

zafer-esen: Yes, maybe it makes more sense to have that as another PR with proper tests, not to have this as a hidden change here.

MikaelMayer: Not having a test for now is ok since it was not tested anyway, but if you had a test, indeed, you could move it to another PR. For now, let's piggy back on this PR to perform this bug fix if it's correct.

MikaelMayer (dismissed):

PR #4601, opened 2023-10-03
Use subcommands instead of an argument for the various back-ends that can be used with translate

keyboardDrummer:
### Changes
- Use subcommands instead of an argument for the various back-ends that can be used with `translate`, this allows each sub-command to get custom option. The syntax for using translate remains the same.
- Remove `ICommandSpec`, which was a wrapper around `System.Commandline.Command` that made it difficult to introduce commands that did something different than run the Dafny compilation pipeline. Commands that benefit from this are `dafny server`, `dafny format`, `dafny doc`, `dafny generate-tests`, `dafny find-dead-code`, `dafny merge-coverage-reports`
- Previously `DafnyDriver` did some CLI argument processing, and some orchestrating of the various steps of the Dafny pipeline. It has been renamed to `CompilerDriver` and now only does the latter. `CommandRegistry` was previously doing part of argument processing, but is now doing all argument processing and has been renamed to `DafnyCli`. There was some testing code related to CLI arguments in `DafnyDriver` that has been moved to a new class `DafnyCliTests`.
- No longer build the standalone Dafny server executable

### Help output
Default:
```
rwillems@bcd0745419f2 dafny % ./Scripts/dafny translate --help
Description:
  Translate Dafny sources to source and build files in a specified language.

Usage:
  Dafny translate [command] [options]

Options:
  -?, -h, --help   Show help and usage information
  --help-internal  Show help and usage information, including options designed for developing the Dafny language and toolchain.

Commands:
  cs <file>    Translate Dafny sources to C# source and build files.
  js <file>    Translate Dafny sources to JavaScript source and build files.
  go <file>    Translate Dafny sources to Go source and build files.
  java <file>  Translate Dafny sources to Java source and build files.
  py <file>    Translate Dafny sources to Python source and build files.
  cpp <file>   Translate Dafny sources to C++ source and build files. This back-end has various limitations (see Docs/Compilation/Cpp.md). This includes lack of support for BigIntegers (aka int), most higher order functions, and advanced features like traits or co-inductive types.

```

With hidden commands (Rust):
```
rwillems@bcd0745419f2 dafny % ./Scripts/dafny translate --help-internal
Description:
  Translate Dafny sources to source and build files in a specified language.

Usage:
  Dafny translate [command] [options]

Options:
  -?, -h, --help   Show help and usage information
  --help-internal  Show help and usage information, including options designed for developing the Dafny language and toolchain.

Commands:
  cs <file>    Translate Dafny sources to C# source and build files.
  js <file>    Translate Dafny sources to JavaScript source and build files.
  go <file>    Translate Dafny sources to Go source and build files.
  java <file>  Translate Dafny sources to Java source and build files.
  py <file>    Translate Dafny sources to Python source and build files.
  cpp <file>   Translate Dafny sources to C++ source and build files. This back-end has various limitations (see Docs/Compilation/Cpp.md). This includes lack of support for BigIntegers (aka int), most higher order functions, and advanced features like traits or co-inductive types.
  lib <file>   Translate Dafny sources to Dafny Library (.doo) source and build files.
  rs <file>    Translate Dafny sources to Rust source and build files.
```

### Testing
- Tests that were testing properties of the standalone language server process are now testing the Dafny CLI with the server command instead
- The rest of this PR is a refactoring that's covered by existing tests


Source/DafnyDriver/CompilerDriver.cs
```diff
@@ -24,365 +23,81 @@
...
-  public class DafnyDriver : IDisposable {
-    public DafnyOptions Options { get; }
-
+  /// <summary>
+  /// Calls into different phases of Dafny's compilation pipeline,
+  /// such as parsing, resolution, verification and code generation
+  /// 
+  /// Will be replaced by CompilationManager
+  /// </summary>
+  public class CompilerDriver : IDisposable {
```
robin-aws, line 37: Slightly uncomfortable with the new name, because it feels a bit subtle that this runs verification as well as the "compiler". It's especially weird that `VerifyCommand` calls `RunCompiler`, with options such that no "compilation" happens. :)

Perhaps `DafnyPipelineDriver` instead?

Source/DafnyDriver/Commands/DafnyCliTests.cs
```diff
@@ -0,0 +1,66 @@
+using System.Linq;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Dafny;
+
+public static class DafnyCliTests {
```
robin-aws, line 6: I definitely like having this extracted out separately :)

Source/DafnyDriver/Commands/TranslateCommand.cs
```diff
@@ -5,37 +5,42 @@
...
-    dafnyOptions.Compile = false;
-    var noVerify = dafnyOptions.Get(BoogieOptionBag.NoVerify);
-    dafnyOptions.CompilerName = context.ParseResult.GetValueForArgument(Target);
-    dafnyOptions.SpillTargetCode = noVerify ? 3U : 2U;
+    result.AddCommand(new Command("cs", "C#"));
+    result.AddCommand(new Command("go", "GoLang"));
+    result.AddCommand(new Command("js", "JavaScript"));
+    result.AddCommand(new Command("java"));
+    result.AddCommand(new Command("py", "Python"));
+    result.AddCommand(new Command("cpp", "C++. This back-end has various limitations (see Docs/Compilation/Cpp.md). This includes lack of support for BigIntegers (aka int), most higher order functions, and advanced features like traits or co-inductive types."));
```
robin-aws, line 25: This is probably the best we can do and I like the change to subcommands too much to complain about it. But do we have any integration tests that assert the output of `--help`?

Source/DafnyDriver/Commands/TranslateCommand.cs
```diff
@@ -5,37 +5,42 @@
...
-    result.AddArgument(ICommandSpec.FilesArgument);
-    return result;
-  }
 
-  public void PostProcess(DafnyOptions dafnyOptions, Options options, InvocationContext context) {
-    dafnyOptions.Compile = false;
-    var noVerify = dafnyOptions.Get(BoogieOptionBag.NoVerify);
-    dafnyOptions.CompilerName = context.ParseResult.GetValueForArgument(Target);
-    dafnyOptions.SpillTargetCode = noVerify ? 3U : 2U;
+    result.AddCommand(new Command("cs", "C#"));
```
robin-aws, line 20: Shouldn't the subcommands be dynamically populated based on the set of available backends? Or are you going to leave that for #4591? (That would be just fine if so)

Source/DafnyDriver/Commands/DafnyCli.cs
```diff
@@ -0,0 +1,514 @@
...
+    public IStandardStreamWriter Out => StandardStreamWriter.Create(OutWriter ?? TextWriter.Null);
+
+    public bool IsOutputRedirected => OutWriter != null;
+    public IStandardStreamWriter Error => StandardStreamWriter.Create(ErrWriter ?? TextWriter.Null);
+    public bool IsErrorRedirected => ErrWriter != null;
+    public bool IsInputRedirected => false;
+  }
+
+  public delegate Task<int> ContinueWithOptions(DafnyOptions dafnyOptions, InvocationContext context);
+  public static void SetHandlerUsingDafnyOptionsContinuation(Command command, ContinueWithOptions continuation) {
```
robin-aws, line 125: This doesn't feel like it justifies using "continuation" as in CPS. Perhaps just `TaskWithOptions` and `SetCommandHandlerImplementation`?

Test/c++/arrays.dfy
```diff
@@ -1,4 +1,4 @@
-// RUN: %testDafnyForEachCompiler "%s" --refresh-exit-code=0 -- --relax-definite-assignment --spill-translation --unicode-char:false
```
robin-aws, line 1: Why did this change?

Source/DafnyDriver/Commands/DafnyCli.cs
keyboardDrummer: From wiki:

> A function written in continuation-passing style takes an extra argument: an explicit "continuation"; i.e., a function of one argument. When the CPS function has computed its result value, it "returns" it by calling the continuation function with this value as the argument.

The function `SetHandlerUsingDafnyOptionsContinuation` takes an extra argument, the function `continuation`. When the handler has computed the DafnyOptions value, it "returns" it by calling the function `continuation` with this value as an argument.

To rename `ContinueWithOptions` to `TaskWithOptions` would be incorrect because `ContinueWithOptions` is not a task: it's a function that returns a task. `SetCommandHandlerImplementation` sounds like it does the same as `command.SetHandler`, which already exists.

Instead of `continuation` I can also use the term callback.

Source/DafnyDriver/Commands/TranslateCommand.cs
keyboardDrummer: There's a catch-22 there because plugins can add new back-ends. I think it would be better if plugins that expose new back-ends are new binaries that reference Dafny as a library. They could potentially then be called using `dafny translate <new-backend-name> by being available on path with the name `dafny-translate-<new-backend-name>`, similar to how `dotnet` does it.

@MikaelMayer

Source/DafnyDriver/Commands/TranslateCommand.cs
keyboardDrummer: What is there to complain about?

> But do we have any integration tests that assert the output of --help?

I don't think so. Would be a type of regression test, breaking every time the options are changed.

Source/DafnyDriver/CompilerDriver.cs
keyboardDrummer: Verification requires running the compiler 🤷‍♂️

Source/DafnyLanguageServer.Test/StandaloneServerTest.cs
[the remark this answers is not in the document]
keyboardDrummer: Moved to `DafnyDriver.Test` and was renamed to `LanguageServerProcessTest`

Test/c++/arrays.dfy
keyboardDrummer: Accidentally, but there is no harm in this change either. Reverted it.

Source/DafnyDriver/CompilerDriver.cs
robin-aws: Fair enough, I'll admit I'm biased by the fact that we've (incorrectly) used the word "compiler" for just the code generation phase of the overall compilation pipeline. :) Gradually fixing that will help.

> So translating Dafny to SMT is compilation

More importantly I agree with your point because the verification is just a heavyweight static analysis early on in the overall compilation pipeline (which happens to work by compiling to a different language), and there are good use cases for ONLY running that static analysis by itself.

Source/DafnyDriver/Commands/TranslateCommand.cs
robin-aws: > What is there to complain about?

It looked like you were cramming the disclaimer into a "full language name" property. But after pulling down your branch and exploring the `Command` types I see it's a "description" property, and the real shortcoming is the fact that the other subcommands aren't providing an actual description. And that leads to this:

```
dafny translate cs --help
Description:
  C#

Usage:
  Dafny translate cs [<file>...] [options]
...
```

So we should really make the descriptions things like "Translate Dafny sources to C# source and build files."

Source/DafnyDriver/Commands/TranslateCommand.cs
robin-aws: Ah that is an unfortunate circular dependency. That sounds like this PR breaks the workflow for compiler plugins though, because something like `dafny translate blort ... --plugin:BlortCompiler.dll` will no longer work?

Source/DafnyDriver/Commands/DafnyCli.cs
robin-aws: Okay I see it now, `SetHandlerUsingDafnyOptionsContinuation` itself isn't really in CPS style, but the handler beting set is.

> Instead of continuation I can also use the term callback.

That might be slightly better but I'm okay with this as is.

Source/DafnyDriver/Commands/TranslateCommand.cs
robin-aws: On a similar note what's up with the `[]`s?

```
Commands:
  cs <file>    C# []
...
```

Source/DafnyDriver/Commands/TranslateCommand.cs
MikaelMayer: The whole reason why plugins are provided to Dafny rather than the other way round is that it makes it possible to add these plugins to Dafny in VSCode too, including plugins that expose backends, because these backends could also serve as a basis for verifying things specifics to some backends in the future;

I suggest that, for custom back-ends, we simply break the circular dependency by having a special command that emulates the previous behavior, like "custom-backend" that will support an option like "--name:rs" and load a back-end from the set of plugins.

keyboardDrummer: > That sounds like this PR breaks the workflow for compiler plugins though, because something like dafny translate blort ... --plugin:BlortCompiler.dll will no longer work?

That's correct, but I don't think someone was doing this.

> The whole reason why plugins are provided to Dafny rather than the other way round is that it makes it possible to add these plugins to Dafny in VSCode too

That you can still do. `dafny server --plugin:AddWhateverYouLikeToTheIDE.dll`

> I suggest that, for custom back-ends, we simply break the circular dependency by having a special command that emulates the previous behavior, like "custom-backend" that will support an option like "--name:rs" and load a back-end from the set of plugins.

I suggest we consider adding that when there's a need, although then I would argue we shouldn't add it.

Source/DafnyDriver/Commands/TranslateCommand.cs
keyboardDrummer: Fixed both issues

Source/DafnyDriver/Commands/TranslateCommand.cs
MikaelMayer: > That's correct, but I can't imagine anyone was doing this.

To clarify: you don't believe anyone was doing this or you just don't know about this use case?

> That you can still do. `dafny server --plugin:AddWhateverYouLikeToTheIDE.dll`

But then they can't choose the back end. I know of at least one team that has two custom back-ends and use VSCode with their custom plugin. I'm pretty sure they will appreciate being able to right-click, and select dafny run, and with their custom arguments can choose which backend it would run on.

> I suggest we consider adding that when there's a need.

This PR is _removing_ what was plugins enabled in the first place, for which @cpitclaudel spent quite some time refactoring the codebase. In general, I would say that any PR that removes a feature without offering workarounds, when the author does not have evidence will not affect customers, should be approved by at least two team members, one of which participated in the creation of the said feature if possible.

robin-aws: +1 to what @MikaelMayer is saying, unfortunately. Is there any way we can special-case the `--plugin` option so it can be at least partially processed before resolving the subcommands? I doubt it without moderate hackery but I have to ask anyway. :P

Alternatively, how hard would it be to implement this idea instead? "They could potentially then be called using dafny translate <new-backend-name> by being available on path with the name dafny-translate-<new-backend-name>, similar to how dotnet does it." Then we would at least have a quick workaround to the breaking change.

keyboardDrummer: Made a related PR: https://github.com/dafny-lang/dafny/pull/4627

keyboardDrummer: > Is there any way we can special-case the --plugin option so it can be at least partially processed before resolving the subcommands?

We can but I don't think we should. Actually I think some of our current back-ends should become separate binaries at some point.

> Alternatively, how hard would it be to implement this idea instead? "They could potentially then be called using dafny translate by being available on path with the name dafny-translate-, similar to how dotnet does it." Then we would at least have a quick workaround to the breaking change.

The 'workaround' is to build a binary instead of a library when building your custom back-end. The users that I know whom are using a custom back-ends already do this. They also use a Dafny plugin but they don't use it for translation to a target language, only to customize resolution of the language server, which this PR still allows.

> But then they can't choose the back end. I know of at least one team that has two custom back-ends and use VSCode with their custom plugin. I'm pretty sure they will appreciate being able to right-click, and select dafny run, and with their custom arguments can choose which backend it would run on.

You can configure tasks in VS code to easily run a custom binary on the current particular file.

Source/IntegrationTests/LitTests.cs
```diff
@@ -535,7 +543,7 @@ class MultiBackendLitCommand : ILitCommand {
     public (int, string, string) Execute(TextReader inputReader,
       TextWriter outputWriter,
       TextWriter errorWriter) {
-      var exitCode = new MultiBackendTest(inputReader, outputWriter, errorWriter).Start(arguments.Prepend("for-each-compiler"));
+      var exitCode = new MultiBackendTest(inputReader, outputWriter, errorWriter).Start(arguments);
```
robin-aws, line 546: I assume you found this because of actually running tests with DAFNY_INTEGRATION_TESTS_IN_PROCESS=true? Thanks for the fix, and we should add some singleton tests for that mode to protect against regressions (not as part of this PR)

Source/TestDafny/MultiBackendTest.cs
```diff
@@ -94,10 +96,10 @@ record ResolutionSetting(
     int ExpectedExitCode
   );
 
-  private int ForEachCompiler(ForEachCompilerOptions options) {
-    var parseResult = CommandRegistry.Create(TextWriter.Null, TextWriter.Null, TextReader.Null,
-      new string[] { "verify", options.TestFile! }.Concat(options.OtherArgs).ToArray());
-    var dafnyOptions = ((ParseArgumentSuccess)parseResult).DafnyOptions;
+  private async Task<int> ForEachCompiler(ForEachCompilerOptions options) {
+    var pluginParseResult = CommonOptionBag.PluginOption.Parse(options.OtherArgs.ToArray());
```
robin-aws, line 100: i'm pleased that at least this context supports plugins :)

robin-aws (approved):
LGTM! I pushed a commit to add newlines to the cpp description as IMO it really improved the readability, and I say we should do that in general to all long command and argument descriptions.

You have a couple of legit test failures to deal with, but I'll happily re-approve once those are fixed.

stefan-aws (approved):

PR #4603, opened 2023-10-03
Clone reads for decreases

RustanLeino:
This PR does 3 things. The primary thing is to change

``` c#
Expression e = fe.E;
```

to

``` c#
Expression e = new Cloner(false, true).CloneExpr(fe.E);
```

in `InferDecreasesExpr.cs`. This prevents the given `reads` expression AST nodes from being shared in the generated `decreases` clause. (This is good, because it lets the new type inference assume that it will come across each variable declaration just once.)

The PR also introduces a method `SystemModuleManager.ObjectSetType()`, which returns the type `set<object?>`. This type was constructed in many places in the Dafny implementation, including in the code surrounding the primary change of this PR.

Finally, the PR rewrites some code surrounding the primary change to use more modern C#.


keyboardDrummer (approved):
Looks good

PR #4604, opened 2023-10-03
Fix spurious vacuity warnings when assertions converted to assumptions

atomb:
Some assertions are translated into assumptions in specific contexts.
Before this change, if those assertions were unused in the proof, Dafny
would warn that they may have been proved vacuously. Now it doesn’t.

Along the way, this also changes the behavior of proof dependency warnings
so that they're produced for any definition that successfully verifies, even
if some definitions in the program have errors.

By submitting this pull request, I confirm that my contribution is made under
the terms of the MIT license.

### DIFF
--- a/Source/DafnyCore/ProofDependencyWarnings.cs
+++ b/Source/DafnyCore/ProofDependencyWarnings.cs
@@ -81,8 +81,13 @@ public class ProofDependencyWarnings {
         return false;
       }
 
-      // Similarly here
-      if (poDep.ProofObligation is MatchIsComplete or AlternativeIsComplete) {
+      // Some proof obligations occur in a context that the Dafny programmer
+      // doesn't have control of, so warning about vacuity isn't helpful.
+      if (poDep.ProofObligation
+          is MatchIsComplete
+          or AlternativeIsComplete
+          or ValidInRecursion
+          or TraitDecreases) {
         return false;
       }
 
--- a/Source/DafnyCore/ProofDependencyWarnings.cs
+++ b/Source/DafnyCore/ProofDependencyWarnings.cs
@@ -1,6 +1,7 @@
 using System.Linq;
 using DafnyCore.Verifier;
 using Microsoft.Dafny.ProofObligationDescription;
+using VC;
 
 namespace Microsoft.Dafny;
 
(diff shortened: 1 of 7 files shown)


Source/DafnyCore/ProofDependencyWarnings.cs
```diff
@@ -81,8 +86,13 @@ public class ProofDependencyWarnings {
         return false;
       }
 
-      // Similarly here
-      if (poDep.ProofObligation is MatchIsComplete or AlternativeIsComplete) {
+      // Some proof obligations occur in a context that the Dafny programmer
+      // doesn't have control of, so warning about vacuity isn't helpful.
+      if (poDep.ProofObligation
```
keyboardDrummer, line 91: I think we should avoid having references to specific things from general code, so instead this `is or or or` check I'd rather see a `!ProofObligation.UserControlled` (or a better name)

Source/DafnyCore/ProofDependencyWarnings.cs
```diff
@@ -81,8 +86,13 @@ public class ProofDependencyWarnings {
         return false;
       }
 
-      // Similarly here
-      if (poDep.ProofObligation is MatchIsComplete or AlternativeIsComplete) {
+      // Some proof obligations occur in a context that the Dafny programmer
```
keyboardDrummer, line 89: What do you mean by doesn't have control of? How would a `MatchIsComplete` obligation be vacuous? Does that only occur if false is proven before the match is reached?

Source/DafnyCore/Verifier/ProofDependency.cs
```diff
@@ -66,6 +66,20 @@ public class ProofObligationDependency : ProofDependency {
   }
 }
 
+public class AssumedProofObligationDependency : ProofDependency {
+  public override RangeToken Range { get; }
+
+  public PODesc.ProofObligationDescription ProofObligation { get; }
+
+  public override string Description =>
```
keyboardDrummer, line 74: This text doesn't occur in new tests. What is it for?

keyboardDrummer (commented):
Could you add more tests? I don't think the new behavior is entirely covered.

keyboardDrummer (approved):

keyboardDrummer:
Approved and merged to unblock the nightly build, but would still appreciate it if you could address or reply to the comments.

Source/DafnyCore/ProofDependencyWarnings.cs
atomb: Sometimes Dafny encodes the proof obligations for these things in ways that include assumptions that don't show up in the original Dafny source, including sometimes something as explicit as `assume false`. These lead to warnings about vacuous proofs that the user can't do anything about. The proofs are still sound -- I think the encoding works that way just to make it simpler and more consistent -- but they're structured in such a way that they can involve vacuously-proved assertions.

Source/DafnyCore/Verifier/ProofDependency.cs
atomb: It's pretty difficult to cause the fact being assumed here to be necessary to complete a later proof (which is why it was leading to spurious warnings before!). Usually these assumptions come about as part of proving side conditions about partial operations. They usually get proven automatically, pretty easily, and therefore can be re-proved later without relying the assumption that they've already been proved. I've been working on a test that causes this message to show up, but I haven't managed to create one yet.

PR #4612, opened 2023-10-05
Use resolved syntax expressions for triggers

RustanLeino:
This PR makes sure that any automatically selected trigger expressions are of AST forms that can be resolved (by the new type system).

## More detail

The AST nodes `FunctionCallExpr` and `MemberSelectExpr` (henceforth referred to as "those two cases") are never produced by the parser. Instead, these are created during resolution, from some `ConcreteSyntaxExpression` after such a `ConcreteSyntaxExpression` has been resolved. This means that name-resolution/type-inference doesn't need to handle those two cases. In the old type system, there is nevertheless code for resolving these two AST nodes, which seems like dead code or at best redundant. The new type-system implementation instead asserts out if it's asked to resolve one of those two expressions. So, the new type system expects to never be asked to resolve these two kinds of AST nodes.

Indeed, most of the Dafny pipeline uses the more specific AST nodes (by looking up the `.Resolved` field of whatever AST node they see, which has the effect of going from any `ConcreteSyntaxExpression` to its resolved counterpart). So, most parts of the Dafny pipeline will use those two cases (and will always bypass any `ConcreteSyntaxExpression`). However, if, _after resolution_, Dafny extends the AST in any way, then the resulting AST may be cloned for any refinement module, in which case they will go through the resolver again for the refinement module.

This is the reason that the old type-system implementation is still willing to resolve a `FunctionCallExpr` or `MemberSelectExpr`. Something needs to be done for the new type-system implementation--should it also be willing to resolve those two AST nodes or should we hold the AST to a higher standard where those two cases are never used as input to the resolver? The design of the type-system implementation chooses the latter.

One place where expressions of those two kinds were being generated is in the auto-trigger machinery. This PR makes sure that any trigger that is generated and incorporated into the AST is not of those two kinds, but rather is a `ConcreteSyntaxExpression` (that the resolver will then turn into an expression of those two kinds).

## A detail

Evidently, the pretty printer includes any type parameters when asked to do an `/rprint`. For this reason, one of the `.expect` files in this PR changed to reflect that these parameters are shown in triggers.

### DIFF
--- a/Source/DafnyCore/AST/Expressions/Expression.cs
+++ b/Source/DafnyCore/AST/Expressions/Expression.cs
@@ -774,6 +774,18 @@ public abstract class Expression : TokenNode {
   }
 
   /// <summary>
+  /// If "expr" is an expression that exists only as a resolved expression, then wrap it in the usual unresolved structure.
+  /// </summary>
+  public static Expression WrapAsParsedStructureIfNecessary(Expression expr, SystemModuleManager systemModuleManager) {
+    if (expr is FunctionCallExpr functionCallExpr) {
+      return WrapResolvedCall(functionCallExpr, systemModuleManager);
+    } else if (expr is MemberSelectExpr memberSelectExpr) {
+      return WrapResolvedMemberSelect(memberSelectExpr);
+    }
+    return expr;
+  }
+
+  /// <summary>
   /// Create a resolved case expression for a match expression
   /// </summary>
   public static MatchCaseExpr CreateMatchCase(MatchCaseExpr old_case, Expression new_body) {
--- a/Source/DafnyCore/Rewriters/RewriterCollection.cs
+++ b/Source/DafnyCore/Rewriters/RewriterCollection.cs
@@ -22,7 +22,7 @@ public static class RewriterCollection {
 
     if (reporter.Options.AutoTriggers) {
       result.Add(new QuantifierSplittingRewriter(reporter));
-      result.Add(new TriggerGeneratingRewriter(reporter));
+      result.Add(new TriggerGeneratingRewriter(reporter, program.SystemModuleManager));
     }
 
     if (reporter.Options.TestContracts != DafnyOptions.ContractTestingMode.None) {
--- a/Source/DafnyCore/Rewriters/TriggerGeneratingRewriter.cs
+++ b/Source/DafnyCore/Rewriters/TriggerGeneratingRewriter.cs
@@ -3,8 +3,10 @@ using System.Diagnostics.Contracts;
 namespace Microsoft.Dafny;
 
 public class TriggerGeneratingRewriter : IRewriter {
-  internal TriggerGeneratingRewriter(ErrorReporter reporter) : base(reporter) {
+  private readonly SystemModuleManager systemModuleManager;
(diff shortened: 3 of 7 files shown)


keyboardDrummer:
> However, if, after resolution, Dafny extends the AST in any way, then the resulting AST may be cloned for any refinement module, in which case they will go through the resolver again for the refinement module.

I feel that cloning the AST is a performance problem, both for the Dafny compilation time but also for the performance of the generated code, because it can cause the generated code to have a size that does not linearly grow with the size of the Dafny program.

However, given that we are cloning, I'm OK with doing all of it before resolution (I think I argued against this in the past but apparently I changed my mind). If you want we can change RefinementTransformer so it overrides `PreResolve(Program)` instead of `PreResolve(ModuleDefinition)` so it does all the cloning before resolution.

jtristan (approved):

RustanLeino:
@keyboardDrummer Thanks for your thoughts. It's true that because of the cloning that refinement modules do, the size of the resulting program is not a linear function of the size of the textual program. Perhaps as some comfort, the module refinement features are designed to try to avoid the _verifier_ having to reverify those things in the parent module. But perhaps we can also devise some scheme to keep the textual form, rather than cloning it into refinement module.

If a module `B` refines a module `A`, then the contents of `A` must be made to be available in `B` at the time `B` is being resolved. This is currently done by cloning `A` as a _pre-resolution_ step when processing `B`. The problem that this PR solved was that the resolution of `A` added resolved-only AST nodes to `A`, and then (during pre-resolution of `B`) those AST nodes were cloned into `B`, and that led to that the resolution of `B` encountered unexpected things in the AST.

keyboardDrummer:
> The problem that this PR solved was that the resolution of A added resolved-only AST nodes to A, and then (during pre-resolution of B) those AST nodes were cloned into B, and that led to that the resolution of B encountered unexpected things in the AST.

I understood that, and I was trying to suggest we use a different way to resolve this problem than this PR does.

If we change RefinementTransformer so it overrides `PreResolve(Program)` (happens before any module is resolved) instead of `PreResolve(ModuleDefinition)` (happens before a particular module is resolved, after dependent modules have been resolved), then during pre-resolution of B, only unresolved nodes of A will be copied into B.

I believe this would avoid the needs for methods such as `WrapResolvedCall` which I think add a significant amount of complexity.

RustanLeino:
> overrides PreResolve(Program) (happens before any module is resolved) instead of PreResolve(ModuleDefinition)

Sorry, I missed this distinction. (In fact, I wasn't aware of the existence of `PreResolve(Program)`. Nice.) Yes, this would reduce complexity for the resolver.

keyboardDrummer:
I tried a little prototype but it's not as simple as I hoped. It seems refinement does require the module being refined to have already done some resolution. In particular the signature of the refinee is needed, and I think it needs to process the imports of the refinee as well so it's not just the refinee that's needs some level of resolution.

Looking at the state of things a while back (v3.0.0), module refinement was also being done after resolution of the refinee.

So, let's keep this as it is.

I hope to find some time, hopefully this year, to look into how other compilers store resolution state, and how they generate new code (and resolve it) while already being in the process of resolution.

PR #4615, opened 2023-10-05
Move lit tests

keyboardDrummer:
### Changes
Move Littish tests from `Test` to `Source/IntegrationTests/Test`. This way there's no symlink needed which causes Rider to behave annoyingly


robin-aws:
Great idea! 

Having trouble browsing/commenting on specific files since GitHub only loads the first 3000 files and my browser keeps hanging. :) I'll just make comments here instead:

1. Could you reformat the commits so that the big Test relocation is in its own dedicated commit? That way I can exclude it when viewing the changes.
2. I intended the `XUnitExtensions` project to be fairly generic and for some of it to apply to any tests with files as input data, not just littish ones. I suspect some of our language server tests could be streamlined a bit to use it as well and make it easier to add more test files for the same test harness. That's why the default location for a test's files is `$"TestFiles/{class name}/{test method name}"`. I strongly prefer to not hardcode that to `Test` instead. Instead we have two better options: we can either relocate the `Test` files under `IntegrationTests/TestFiles/LitTests/LitTest` directly, or you can override the `Directory` property of `FileDataAttribute` on the `LitTest` method to be `"Test"` instead. I prefer the former since "Test" as a subdirectory is pretty meaningless without the history behind it, but either would be fine.
3. Given how long the `Test` directory has existed in this project, I would keep it there with just a small README that says "These files have been relocated under ...", just to give contributors a hand if they hit nasty merge conflicts or get confused.

keyboardDrummer:
@robin-aws could you review the current diff? I think it addresses your comments. I've excluded the actual move for now since it's easier to review and then the PR doesn't accrue conflicts. Once you approve it I'll do the actual move, reapprove, and then merge.

robin-aws:
Looks great and sounds like a plan, thanks! :)

robin-aws (approved):

PR #4623, opened 2023-10-06
Rename Translator to BoogieGenerator

keyboardDrummer:
### Description
Rename Translator to BoogieGenerator. It is good to include Boogie in the name of this class, since its purpose is producing Boogie. For `Generator` we also considered `Translator`, but that makes it ambiguous whether Boogie is the input or the output.

### How has this been tested?
This is a refactoring that's covered by existing tests


RustanLeino (approved):
The name change seems fine.

Should `ExpressionTranslator` retain its name? `ExpressionGenerator` doesn't capture the same meaning, since the "Expression" in `ExpressionTranslator` is referring to the source of the translation. (I can't think of a better name than `ExpressionTranslator`, but am open to suggestions.)

PR #4627, opened 2023-10-07
Hide the plugin option

keyboardDrummer:
### Description
Hides the plugin option from the regular help. 

The plugin option was already marked experimental to indicate that the API is expected to change, and that without special circumstances, no system should depend on this option. Hiding the option means it won't show up when using `--help`, to make it extra clear that users should not depend on this option. Users who work with the Dafny maintainers to develop this feature can still use it, and the option is still shown when using `--help-internal`.

### How has this been tested?
Manually


MikaelMayer (approved):
I'm ok with that.

PR #4630, opened 2023-10-09
Store more source location information in Boogie AST

keyboardDrummer:
### Description
This is a refactoring that enables other features

### How has this been tested?
This refactoring is covered by existing tests


Source/DafnyCore/Verifier/Translator.cs
```diff
@@ -3471,8 +3471,15 @@ public enum StmtType { NONE, ASSERT, ASSUME };
...
-        return node.RangeToken.ToToken();
+        return new BoogieRangeToken(node.StartToken, node.EndToken, node.Tok);
       } else {
+        // The commented line is what we want, but it changes what is translated.
+        // Seems to relate to refinement and possibly RefinementToken.IsInherited and or ForceCheckToken
+        // It might be better to remove calls to RefinementToken.IsInherited from this file, and instead
+        // add generic attributes like {:verify false} in the refinement phases, so that refinement does not complicate
+        // translation,
+        //
+        // return new BoogieRangeToken(node.StartToken, node.EndToken, node.Tok);
```
MikaelMayer, line 3482: That's precisely what I encountered. I think there might be two mechanism for reporting related ranges: One in nested token, another one in auxiliary information retrieved from Boogie. That might create a conflict that we solve one way or the other, but it's incompatible with the correct code above.
At least, given that we have tests in place, it should be enough for us to still have a decent behavior.

MikaelMayer (approved):
That's great and very similar to what I suggested, so I approve it without hesitation. I will also reapprove as soon as you find how to fix the CI test. Thank you very much for the effort!

Source/DafnyCore/Rewriters/RefinementTransformer.cs
```diff
@@ -58,7 +58,8 @@ public RefinementToken(IToken tok, ModuleDefinition m)
       }
 
       var rtok = tok as RefinementToken;
-      return rtok != null && rtok.InheritingModule == m;
+      var result = rtok != null && rtok.InheritingModule == m;
+      return result;
```
MikaelMayer, line 62: Non-blocking: this refactoring introduces a variable that does not seem to provide meaning to the result

MikaelMayer (approved):
Good job for fixing the tests, it wasn't easy

MikaelMayer (dismissed):

stefan-aws (dismissed):

PR #4655, opened 2023-10-12
Dafny generator using project file

jtristan:
### Description
Changes to the generation of cs code from Dafny code in the compiler to use the project file and simplify and generalize the generation process.


keyboardDrummer (commented):
Seems like you should still update `https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Makefile#L20` to account for the fewer amount of generated files.

keyboardDrummer (approved):

PR #4657, opened 2023-10-12
chore: publish nightly prereleases from master instead of 4.3

alex-chew:
### Description

Now that 4.3.0 has been released, we should revert to publishing pre-releases from the master branch instead of the 4.3.0 branch.

This also removes the nightly builds for the 4.3.0 branch entirely, to save CI resources and since it's no longer needed.

### How has this been tested?

No manual test run, but can do so as described in https://github.com/dafny-lang/dafny/pull/4573.


robin-aws (approved):
Okay with not doing a manual test run since the edit is simple

PR #4658, opened 2023-10-12
feat: Populate runtimes with all built-in declarations

### LINKED ISSUE #511: Parser should not dynamically inject shared builtin program components
The advantage of dynamically generating elements such as tuple, array, and arrow types is that there's no maximum arity of such generic types. However, this interferes with being able to compile packages separately, since separate packages may decide to generate the same elements, leading to conflicts when loading or linking.

The easiest fix would be to pre-generate these types up to a maximum arity. Scala currently does this (with a cap of 22) but it looks like Scala 3 will no longer have this limit: https://github.com/lampepfl/dotty/pull/1758. It may be possible to have our cake and eat it too if we borrow some of those techniques.

robin-aws:
### Description
Fixes #511

Generalizes the approach used in #2284 to pre-generate all built-in declarations into the various runtimes:

* Adds an internal `--system-module` option to control how the system module is treated when translating:
  * Include (the current default) - emits code for the contents of the System module
  * Omit (the new default) - emits no code for the System module
  * Populate - emits ONLY the code for the System module
* The third option is used to generate part of each runtime from a `systemModulePopulator.dfy` file, previously named `tuplesForDafnyRuntime.dfy`. The output is now present in each runtime in a separate file/source location, and checked for consistency in CI just as other Dafny-generated committed source is.
* Also generalized how the runtime files are embedded in `DafnyPipeline` to support more than one file better (as Go already was).

TODOs:

* [x] Mechanism to check arities against the limit of 20 when --system-module is Omit
* [x] Remove `internal` from C# helper code (and tests to confirm it works)
* [x] Feature for pre-compiling built-ins (opt out for C++ and Rust for now)
* [x] Reference manual section on this topic (for feature to point to)

### How has this been tested?
Mostly existing tests, but also added an internal version of `--include-runtime` for `dafny run`, so that `%testDafnyForEachCompiler` can include testing on C# when using the separate `DafnyRuntime.dll` library, as that behaves significantly different in this case.


Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -110,8 +110,42 @@ public abstract class SinglePassCompiler {
...
+        UnsupportedFeatureError(Token.NoToken, Feature.ArrowsWithMoreThan16Arguments);
+      }
+      var maxArraysDims = systemModuleManager.arrayTypeDecls.Keys.Max();
+      if (maxArraysDims > 16) {
+        UnsupportedFeatureError(Token.NoToken, Feature.ArraysWithMoreThan16Dims);
+      }
+    }
+    
+    protected void CheckSystemModulePopulatedToCommonLimits(SystemModuleManager systemModuleManager) {
+      systemModuleManager.CheckHasAllTupleNonGhostDimsUpTo(20);
```
keyboardDrummer, line 144: Let's make sure the numbers in `CheckCommonSytemModuleLimits` match those in `CheckSystemModulePopulatedToCommonLimits`

Source/TestDafny/MultiBackendTest.cs
```diff
@@ -207,6 +207,17 @@ int ExpectedExitCode
         if (result != 0) {
           success = false;
         }
+
+        if (compiler.TargetId == "cs") {
+          // C# is a bit unusual in that the runtime behaves a little differently
+          // depending on whether it is included as source or referenced as DafnyRuntime.dll
+          // (because of "#ifdef ISDAFNYRUNTIMELIB" directives).
```
keyboardDrummer, line 214: I didn't get what these directives are for. Could you add documentation on that somewhere close to the C# runtime or compiler?

keyboardDrummer:
This is looking amazing already ! Awesome work

Source/DafnyCore/AST/TypeDeclarations/TupleTypeDecl.cs
```diff
@@ -49,6 +49,9 @@ private TupleTypeDecl(ModuleDefinition systemModule, List<TypeParameter> typeArg
       }
     }
     this.EqualitySupport = argumentGhostness.Contains(true) ? ES.Never : ES.ConsultTypeArguments;
+
+    // TODO: Correct?
```
keyboardDrummer, line 53: Please remove this TODO

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
```diff
@@ -119,8 +130,25 @@ public class CsharpCompiler : SinglePassCompiler {
...
+            CheckCommonSytemModuleLimits(systemModuleManager);
+            break;
+          }
+        case CommonOptionBag.SystemModuleMode.Populate: {
+            CheckSystemModulePopulatedToCommonLimits(systemModuleManager);
+            break;
+          }
+      }
+
+      // The declarations below would normally be omitted if we aren't compiling the system module,
```
keyboardDrummer, line 144: Seems like these declarations could be not-internal and only as part of the system module

Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -269,7 +267,7 @@ private enum JavaNativeType { Byte, Short, Int, Long }
       var wStmts = wr.Fork();
       var wrVarInit = wr;
       wrVarInit.Write($"java.util.ArrayList<{DafnyTupleClass(L)}<{tupleTypeArgs}>> {ingredients} = ");
-      AddTupleToSet(L);
+      // TODO: Assert that Tuple(L) exists in the _System module
```
keyboardDrummer, line 270: TODO

Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -370,7 +373,8 @@ private enum JavaNativeType { Byte, Short, Int, Long }
     protected override ConcreteSyntaxTree CreateModule(string moduleName, bool isDefault, bool isExtern, string /*?*/ libraryName, ConcreteSyntaxTree wr) {
       if (isDefault) {
         // Fold the default module into the main module
-        return wr;
+        // return wr;
```
keyboardDrummer, line 376: Please remove the comment

Source/DafnyCore/Compilers/Java/Compiler-java.cs
```diff
@@ -1840,6 +1836,12 @@ protected class ClassWriter : IClassWriter {
     }
 
     protected override IClassWriter/*?*/ DeclareDatatype(DatatypeDecl dt, ConcreteSyntaxTree wr) {
+      if (dt is TupleTypeDecl tupleTypeDecl) {
+        // CreateTuple() produces somewhat different code
```
keyboardDrummer, line 1840: I don't get the comment

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -226,6 +226,28 @@ public enum GeneralTraitsOptions {
...
+  /// </summary>
+  public static readonly Option<bool> InternalIncludeRuntimeOptionForExecution = new("--include-runtime", () => true,
+    "Include the Dafny runtime as source in the target language.") {
+    IsHidden = true
+  };
+
+  public enum SystemModuleMode {
+    Include,
+    Omit,
+    // TODO: better name? OmitOthers?
```
keyboardDrummer, line 242: The rule of these names seems to be to describe the effect, not the reason behind it, so `OmitAllOtherModules` or `OnlyIncludeSystem[Module]` matches well.

Source/DafnyRuntime/DafnyRuntime.cs
```diff
@@ -5,6 +5,11 @@
 //
 //-----------------------------------------------------------------------------
 
+// When --include-runtime is true, this file is directly prepended
```
keyboardDrummer, line 8: I'm a bit confused. Isn't this file generated? If so, how does this comment persist?

keyboardDrummer (approved):

keyboardDrummer:
Left a few minor comments. Feel free to merge first and address them as a follow-up

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
robin-aws: I tried that, but realized `FuncExtensions` in particular is tricky because they are extension methods on the `System.Func` family of delegates, and extension methods only ever apply within the current assembly. Given that I didn't bother spending the time moving the Array helper functions either. I'm adding a comment about this here though.

Source/DafnyCore/Compilers/Java/Compiler-java.cs
robin-aws: Expanded the comment.

MikaelMayer (approved):
I approve the last change :-)

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
keyboardDrummer: > extension methods only ever apply within the current assembly

I don't get it what you mean. You can use extensions methods defined in a library for sure.

PR #4663, opened 2023-10-13
Fix concurrency bug in a test case

### LINKED ISSUE #4661: Test generation relies on static state which can cause concurrency issues
Here's a test run failure because of a concurrent modification:
https://github.com/dafny-lang/dafny/actions/runs/6507844793/job/17676047373?pr=4630

Dargones:
### Description

Fixes #4661. Test Generation previously relied on maintaining a static cache of preprocessed program ids, which can cause a concurrency failure. This PR makes sure this field gets passed around and is not shared between runs.


keyboardDrummer (approved):

Dargones:
Apologies, @keyboardDrummer. Have accidentally pushed something that was intended for a different PR. Have revert the change now, so it should be good to go

keyboardDrummer (approved):

PR #4669, opened 2023-10-13
Add unused dependencies to text logger

atomb:
Make `--log-format text` output program elements unused in a proof, in addition to the existing proof dependencies.

These can be useful for deep dives into understanding how a proof was completed. 

### DIFF
--- a/Source/DafnyDriver/TextLogger.cs
+++ b/Source/DafnyDriver/TextLogger.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.Boogie;
 
 namespace Microsoft.Dafny;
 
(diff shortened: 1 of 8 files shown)


Source/DafnyDriver/TextLogger.cs
```diff
@@ -42,7 +43,8 @@ public class TextLogger {
           tw.WriteLine(
             $"      {cmd.Tok.filename}({cmd.Tok.line},{cmd.Tok.col}): {cmd.Description}");
         }
-        if (vcResult.CoveredElements.Any()) {
+
+        if (vcResult.CoveredElements.Any() && vcResult.Outcome == ProverInterface.Outcome.Valid) {
```
keyboardDrummer, line 47: Is this line change a bug fix?

keyboardDrummer (dismissed):

Source/DafnyDriver/TextLogger.cs
atomb: I added a test for this. I'm not sure whether I'd consider it a bug fix or just an instance of changing my mind about how the UI should work, but it's good to test either way.

keyboardDrummer (approved):

keyboardDrummer (approved):

PR #4673, opened 2023-10-16
Allow filtering of combined coverage reports

atomb:
Add a new `--only-label` option to `merge-coverage-reports` to include only one category of highlighting in the output. For example, merging coverage reports from test generation and verification using the option `--only-label NotCovered` will highlight only the regions not covered by either testing or verification.

### Description

### DIFF
--- a/Source/DafnyDriver/CoverageReporter.cs
+++ b/Source/DafnyDriver/CoverageReporter.cs
@@ -85,11 +85,15 @@ public class CoverageReporter {
         reports.Add(ParseCoverageReport(reportDir, $"{name} ({Path.GetFileName(reportDir)})", units, suffix));
       }
     }
+
+    var onlyLabel = options.Get(CoverageReportCommand.OnlyLabelOption);
     foreach (var report in reports) {
       foreach (var fileName in report.AllFiles()) {
         foreach (var span in report.CoverageSpansForFile(fileName)) {
           mergedReport.RegisterFile(span.Span.Uri);
-          mergedReport.LabelCode(span.Span, span.Label);
+          if ((onlyLabel ?? span.Label) == span.Label) {
+            mergedReport.LabelCode(span.Span, span.Label);
+          }
         }
       }
     }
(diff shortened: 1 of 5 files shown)


Source/DafnyDriver/CoverageReporter.cs
```diff
@@ -85,11 +85,15 @@ public class CoverageReporter {
...
+
+    var onlyLabel = options.Get(CoverageReportCommand.OnlyLabelOption);
     foreach (var report in reports) {
       foreach (var fileName in report.AllFiles()) {
         foreach (var span in report.CoverageSpansForFile(fileName)) {
           mergedReport.RegisterFile(span.Span.Uri);
-          mergedReport.LabelCode(span.Span, span.Label);
+          if ((onlyLabel ?? span.Label) == span.Label) {
+            mergedReport.LabelCode(span.Span, span.Label);
+          }
```
Dargones, line 96: Is the intent to display locations where both reports have the same label (i.e. intersection)? If yes, then there should be an else branch here that labels the code with the `NotApplicable` label. The report merged using the present approach will display the union between the two original reports.

Dargones (changes requested):

Source/DafnyDriver/CoverageReporter.cs
atomb: Not quite. The point is to highlight only the locations that, after merging (with union semantics), have the label specified on the command line. The expected output I've checked in looks the way I'd expect it to. The key use case I want to support is for `NotCovered`: highlight all locations that are not covered by either verification or tests.

atomb: Alternatively, you could see this code as taking the merged report and _removing_ all highlighting that isn't in the specified category. The conditional is arguably a bit confusing, given that it both checks whether the option has been specified and whether the current label matches the requested label in one expression.

Dargones (approved):

Source/DafnyDriver/CoverageReporter.cs
Dargones: Ah, I see! So you can get the intersection if you don't use `--only-label` and the union if you do. This makes sense!

PR #4675, opened 2023-10-17
Do not let documentSymbol API return a faulty line 

### LINKED ISSUE #446: :- syntax awkward for methods with multiple return values or no results
The error handling support in Dafny is currently centered around the "let or fail" statement, which attempts to `Extract()` a value and immediately returns the result of `PropagateFailure()` otherwise:

```
var x :- SomeMethod(...);
```

This doesn't compose well with methods that return multiple values, however, since `Extract()` must produce a single value, so you can never write this:

```
var x, y, z :- SomeMethod(...);
```

There is also use case for methods that operate by side-effects and return no values but may fail. I'm finding that I'm defining several methods that return a `Result<()>` of some kind and calling them like this:

```
var _ :- SomeMethod(...);
```

It's a minor annoyance but it adds up, especially when trying to implement tests. I'd suggest we support the unary variant of `:-` (currently only usable in functional expressions) for statements like this too:

```
:- SomeMethod(...);
```

keyboardDrummer:
Fixes [#446](https://github.com/dafny-lang/ide-vscode/issues/446)

### Description
Do not let documentSymbol API return a faulty line when parsing a symbol name has an error

### How has this been tested?
- Added language server test


MikaelMayer (approved):
That's sooo good thanks. I think there are a few more assumptions like that remaining in the code, but at least you are decreasing them by one !

stefan-aws (approved):
Previously approved modulo formatter.

PR #4677, opened 2023-10-17
feat: Standard library support (separate verification only for now) (WIP)

robin-aws:
### How has this been tested?


Source/DafnyStandardLibraries/scripts/check-examples
robin-aws, line 1: Copied over from https://github.com/dafny-lang/libraries/blob/master/Scripts/check-examples with two minor changes:

1. Replaced reference to `DafnyCore.dfy` with `--allow-standard-libraries`
2. Removed the default `--unicode-char:false` option

Source/DafnyCore/AST/Grammar/ProgramParser.cs
```diff
@@ -34,7 +34,7 @@ public class ProgramParser {
     CancellationToken cancellationToken) {
     var options = errorReporter.Options;
     var builtIns = new SystemModuleManager(options);
-    var defaultModule = new DefaultModuleDefinition(files.Where(f => !f.IsPreverified).Select(f => f.Uri).ToList());
+    var defaultModule = new DefaultModuleDefinition(files.Select(f => f.Uri).ToList());
```
robin-aws, line 37: Bug fix: otherwise the contents of a .doo file passed directly without `--library` doesn't get compiled.

Issue #4648, opened 2023-10-11
Standard library: MutableCollections


robin-aws:
Mutable versions of common collection types.

Existing source:

[dafny-lang/libraries/src/MutableMap](https://github.com/dafny-lang/libraries/tree/master/src/MutableMap) (partial)

robin-aws:
Here's more detail on the initial approach we're going to use and why its sound.

As of Dafny 4.4, the `{:concurrent}` attribute will verify that a concurrent entry point has effectively empty `reads` and `modifies` clauses, ensuring that there are no data races or incorrect verification w.r.t. shared mutable state. However, it is still possible to interact with mutable state held outside of Dafny's model of the heap, as long as that interaction is confined to `{:extern}` methods, which may exhibit non-determinism because they depend on this state.

For example, a sequential specification of a simple externally-implemented `Box<T>` type in Dafny might look like this:

```dafny
trait Box<T> {
 
  ghost var value: T

  static method {:extern} Make(t: T) returns (ret: Box<T>)
    ensures ret.value == t

  method {:extern} Get() returns (t: T)
    ensures value == t

  method {:extern} Put(t: T)
    ensures value == t
}
```

But if this box was accessed by multiple threads at once, even if there was no runtime errors, the post-conditions on the methods are no-longer valid: after a `Put(x)` there is no guarantee that a later `Get()` will return `x`.

To make such datatypes safe use use concurrently, we can weaken the specification to only enforce an invariant about the contents of the datatype:

```dafny
trait {:extern} AtomicBox<T> {

  ghost const inv: T -> bool

  static method {:extern} Make(ghost inv: T -> bool, t: T) returns (ret: AtomicBox<T>)
    requires inv(t)
    ensures ret.inv == inv

  method {:extern} Get() returns (t: T)
    ensures inv(t)

  method {:extern} Put(t: T)
    requires inv(t)
}
```

As long as the external implementation uses adequate synchronization to ensure atomic reads and writes (e.g. using Java's `volatile` keyword), Dafny verification is now sound under concurrent execution as well. Note the above trait is already used in the Dafny implementation itself: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntimeDafny/src/dafnyRuntime.dfy#L359

This technique can be extended to other containers as well:

```dafny
trait ConcurrentMutableMap<K(==), V(==)> {

  ghost const inv: (K, V) -> bool

  method Put(k: K, v: V)
[...]

PR #4678, opened 2023-10-18
feat: Standard library support + DafnyStdLibs.Wrappers

robin-aws:
### Description
This change adds support for "standard libraries" that are packaged together with the Dafny distribution, at least for now. They are automatically available for import with the option `--standard-libraries`, which will likely become true by default in Dafny 5. To avoid Dafny's footprint from getting larger and larger over time, some portion of them may be moved into separately distributed packages at a later date. They will remain optional and the Dafny language itself should not depend on them.

A later change will likely add pre-translated source to the runtimes as well, and a translation option such as `--include-standard-libraries` similar to `--include-runtime` to rely on this copy of them instead of translating them into every component.

Implementation notes:

* The primary implementation approach, which is mostly not user-visible, is to pre-build one or more `.doo` files and embed them in `DafnyPipeline.dll`, just as we already do with the various runtime files such as `DafnyRuntime.dll`. They are added as input `DafnyFiles`, using URIs of the form `dllresource://DafnyPipeline/<name>.doo`, so they are not re-verified but are in scope for compilation.
* I've fixed/improved a few things around `.doo` files, such as loosening the library checks for some options, and no longer checking options that only affect compilation (it may make sense to leverage these files for compilation compatibility checking in the future but that's not currently in scope).
* The current limitations of the standard library support are documented in `Source/DafnyStandardLibraries/CONTRIBUTING.md`, and should make it clear which pending libraries are still partially blocked.

### How has this been tested?
* The general standard library support is tested with the integration tests under `stdlibs` and `metatests/StdLibsOffByDefaultInTests.dfy`
* The Wrappers library is tested via `WrappersExamples.dfy`


Source/DafnyCore/AST/Grammar/ProgramParser.cs
```diff
@@ -35,7 +35,7 @@ public class ProgramParser {
     CancellationToken cancellationToken) {
     var options = errorReporter.Options;
     var builtIns = new SystemModuleManager(options);
-    var defaultModule = new DefaultModuleDefinition(files.Where(f => !f.IsPreverified).Select(f => f.Uri).ToList());
```
keyboardDrummer, line 38: Seems like the field RootSourceUris can be removed from `DefaultModuleDefinition` entirely, since it's only access on line 42 in this method.

Source/DafnyStandardLibraries/src/dfyconfig.toml
```diff
@@ -0,0 +1,8 @@
+
+[options]
+# TODO: This isn't getting applied for some reason?
+# I'd like to use this to set a hard upper limit on brittleness
+# and avoid having to use measure-complexity or dafny-reportgenerator.
+# Leaving this set ludicrously low to show the option is broken,
+# but hoping to set this to 1000 or even 100 once it's working.
+resource-limit = 1
```
keyboardDrummer, line 8: Note that 1 still means a 1000

Test/metatests/StdLibsOffByDefaultInTests.dfy
```diff
@@ -0,0 +1,17 @@
+
+// RUN: %exits-with 2 %verify "%s" > "%t"
```
keyboardDrummer, line 2: Can you add a test uses the standard libraries?

Source/DafnyStandardLibraries/src/dfyconfig.toml
keyboardDrummer:
```
/Users/rwillems/SourceCode/dafny/Binaries/Dafny verify --log-format=text /Users/rwillems/SourceCode/dafny/Source/DafnyStandardLibraries/src/dfyconfig.toml

Dafny program verifier finished with 4 verified, 0 errors

Results for DafnyStdLibs.Wrappers.Option.Extract (well-formedness)
  Overall outcome: Correct
  Overall time: 00:00:00.3802060
  Overall resource count: 7623
  Maximum assertion batch time: 00:00:00.0721625
  Maximum assertion batch resource count: 7623

  Assertion batch 0:
    Outcome: Valid
    Duration: 00:00:00.0721625
    Resource count: 7623

    Assertions:
      Wrappers.dfy(38,7): destructor 'value' is only applied to datatype values constructed by 'Some'

Results for DafnyStdLibs.Wrappers.Result.PropagateFailure (well-formedness)
  Overall outcome: Correct
  Overall time: 00:00:00.3731120
  Overall resource count: 8735
  Maximum assertion batch time: 00:00:00.0721694
  Maximum assertion batch resource count: 8735

  Assertion batch 0:
    Outcome: Valid
    Duration: 00:00:00.0721694
    Resource count: 8735

    Assertions:
      Wrappers.dfy(81,20): destructor 'error' is only applied to datatype values constructed by 'Failure'

Results for DafnyStdLibs.Wrappers.Result.Extract (well-formedness)
  Overall outcome: Correct
  Overall time: 00:00:00.3728960
  Overall resource count: 8627
  Maximum assertion batch time: 00:00:00.0721442
  Maximum assertion batch resource count: 8627

  Assertion batch 0:
    Outcome: Valid
    Duration: 00:00:00.0721442
    Resource count: 8627

    Assertions:
      Wrappers.dfy(88,7): destructor 'value' is only applied to datatype values constructed by 'Success'

Results for DafnyStdLibs.Wrappers.OutcomeResult.PropagateFailure (well-formedness)
  Overall outcome: Correct
  Overall time: 00:00:00.3730350
  Overall resource count: 12681
  Maximum assertion batch time: 00:00:00.0721833
  Maximum assertion batch resource count: 12681

  Assertion batch 0:
    Outcome: Valid
    Duration: 00:00:00.0721833
    Resource count: 12681

    Assertions:
      Wrappers.dfy(194,20): destructor 'error' is only applied to datatype values constructed by 'Fail''

```

keyboardDrummer: Smt log still shows:

```
...
...
(set-option :rlimit 1000)
(assert (not
 (=> (= (ControlFlow 0 0) 5) (let (($generated@@259  (=> (= (ControlFlow 0 3) (- 0 2)) ($generated@@39 $generated@@253))))
(let (($generated@@260 true))
(let (($generated@@261  (=> (and (and (= $generated@@254 ($generated@@112 $generated@@255 $generated@@256 $generated false)) ($generated@@34 $generated@@30 $generated@@253 ($generated@@31 $generated@@257) $generated@@256)) (and ($generated@@28 $generated@@257 $generated@@253) ($generated@@27 $generated@@257 $generated@@253))) (and (=> (= (ControlFlow 0 4) 1) $generated@@260) (=> (= (ControlFlow 0 4) 3) $generated@@259)))))
(let (($generated@@262  (=> (and ($generated@@35 $generated@@256) ($generated@@258 $generated@@256)) (=> (and (and ($generated@@29 $generated@@30 $generated@@253 ($generated@@31 $generated@@257)) ($generated@@34 $generated@@30 $generated@@253 ($generated@@31 $generated@@257) $generated@@256)) (and (= 1 $generated@@26) (= (ControlFlow 0 5) 4))) $generated@@261))))
$generated@@262)))))
))
(check-sat)
(pop 1)
; Valid
(get-info :rlimit)
```

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.dfy
```diff
@@ -0,0 +1,223 @@
...
+/** This module provides various kinds of failure-compatible datatypes, including Option.
+- Option<R> (Some or None) with a type parameter R that is the type of the Some value
+- Outcome<E> (Pass or Fail) with a type parameter E that is the type of a failure value
+- Result<R,E> which has both a value type and a failure type
+All three of these may be used with `:-` as Dafny failure-compatible types
+
+The module also defines two forms of `Need`, that check the truth of a predicate and
+return a failure value if false.
+ */
+module {:options "--function-syntax:4"} DafnyStdLibs.Wrappers {
```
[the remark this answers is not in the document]
robin-aws: No, I realized this myself and already removed it locally. :)

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.dfy
```diff
@@ -0,0 +1,223 @@
...
+
+  /** A helper function to ensure a requirement is true at runtime.
+      Example: `:- Need(5 == |mySet|, "The set MUST have 5 elements.")`
+  */
+  function Need<E>(condition: bool, error: E): (result: OutcomeResult<E>)
+  {
+    if condition then Pass' else Fail'(error)
+  }
+
+  // TODO: This doesn't really belong in Wrappers
```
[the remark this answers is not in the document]
robin-aws: I'm going to remove it for now and cut an issue - I'm fairly sure no one is using the copy at https://github.com/dafny-lang/libraries/blob/master/src/dafny/Wrappers.dfy#L211. The MutableMap library has a similar definition which is actually what I put here, but I'd rather leave it out and add as part of a later library import.

Test/metatests/StdLibsOffByDefaultInTests.dfy
robin-aws: This one does in fact test that it's possible when `--allow-standard-libraries` is provided, on the next LIT command line. But that's definitely not very clear, I will add a separate direct test.

Source/DafnyStandardLibraries/scripts/check-examples
```diff
@@ -0,0 +1,373 @@
+#! /usr/bin/env bash
```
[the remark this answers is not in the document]
robin-aws: I'd definitely like to, and more importantly make this a proper feature. Just cut https://github.com/dafny-lang/dafny/issues/4680 and will add a link here.

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.dfy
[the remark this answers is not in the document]
robin-aws: https://github.com/dafny-lang/dafny/issues/4648#issuecomment-1768760738

Source/DafnyStandardLibraries/src/dfyconfig.toml
robin-aws: From offline conversation, it seems like this behavior for low `rlimit` values is somewhat expected. Cut an issue to follow-up on it: https://github.com/boogie-org/boogie/issues/801

For this PR its not a blocker as I've upped the limit to what I actually intended (1M), which we know from experience definitely works.

Source/DafnyDriver/Commands/DafnyCli.cs
```diff
@@ -480,6 +482,14 @@ class WritersConsole : IConsole {
       //  "Usage:\ndafny format [--check] [--print] <file/folder> <file/folder>...\nYou can use '.' for the current directory");
       return ExitValue.PREPROCESSING_ERROR;
     }
+    
+    // Add standard library .doo files after explicitly provided source files,
+    // only because if they are added first, one might be used as the program name,
+    // which is not handled well.
+    if (options.Get(CommonOptionBag.AllowStdLibs)) {
+      options.CliRootSourceUris.Add(StandardLibrariesDooUri);
```
keyboardDrummer, line 490: Awkward that you need both lines.

keyboardDrummer, line 490: If you want this to work for `dafny server`, you will need to update `DafnyProject.GetRootSourceUris` so it returns `StandardLibrariesDooUri`.

Alternatively, I can do the refactoring to reuse more of the server code for the CLI and then this'll come for free.

robin-aws: Agreed, same goes for the existing `stdin:` protocol that I'm imitating.

robin-aws: I'd be very happy with that - I'm fine with the examples and tests in DafnyStandardLibraries not resolving in the IDE initially, since that doesn't block adding new libraries, but would love to fast-follow with fixing that. I suspect that refactoring will fix the awkwardness too since the DafnyFile will be added from the root URI automatically?

Source/DafnyCore/DafnyFile.cs
```diff
@@ -65,11 +73,26 @@ public class DafnyFile {
       IsPreverified = true;
       IsPrecompiled = false;
 
-      if (!File.Exists(filePath)) {
-        options.Printer.ErrorWriteLine(options.OutputWriter, $"*** Error: file {filePathForErrors} not found");
-        throw new IllegalDafnyFile(true);
+      DooFile dooFile;
+      if (uri.Scheme == "stdlibs") {
```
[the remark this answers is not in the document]
robin-aws: It's actually a lot like Java's built-in `jar:` protocol! .NET doesn't seem to have anything equivalent built in. I'm going to rename the protocol it to something a bit more general.

robin-aws:
Ready for a final review @keyboardDrummer, but I also intend to block on a second review of the README.md and CONTRIBUTING.md from @jtristan.

Source/DafnyCore/DooFile.cs
```diff
@@ -212,6 +221,30 @@ public class ManifestData {
     return false;
   }
 
+  // Checks that the library option ==> the local option.
```
keyboardDrummer, line 224: I think it's better to use `///` comments. IDE will then also move the comments if you move the definition.

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
```diff
@@ -0,0 +1,191 @@
...
+  var index: int :- convert(Find(s, 0, value));
+```
+
+### Combining different FC-types in methods
+
+The conversion functions used in the last section work syntactically because we had boxed values that were returned by expressions (function calls), to which the conversion functions could
+be applied. When a FC-value is returned by a method there is no place to call such a conversion function: the return value of the method must be captured by either `:=` or `:-`.
+So some new syntax will be needed --- what this will be is under design and discussion.
+
+The workaround, however, is straitforward: just capture the results using `:=` and make the conversion directly. Here the is example above restated using a method.
```
keyboardDrummer, line 155: give it to me STRAIT

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
```diff
@@ -0,0 +1,191 @@
+
+## The `Wrappers` module
+
+The Wrappers module holds some
+simple datatypes to support common patterns, such as optional values or the result of operations that can fail.
+These are additionally useful with Dafny's abrupt-termination-on-failure `:-` operator.
```
keyboardDrummer, line 6: I'm not a native English speaker but: additionally -> particularly?

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
```diff
@@ -0,0 +1,191 @@
...
+
+Any user datatype can serve this purpose, as long as it has an `IsFailure?` predicate 
+(in which case it is known as a [_failure-compatible_](https://dafny.org/latest/DafnyRef/DafnyRef#sec-failure-compatible-types) type --- an FC-type). 
+
+In this library module Dafny defines three such types:
+- `Option<R>` - which is either `Some` with a value of type `T` or a `None` with no information
+- `Outcome<E>` - which is either `Pass` with no informatino or `Fail` with an error value of type `E` (often a `string`) 
+- `Result<R,E>` - which is either `Success` with a value of type `R` or `Failure` with error value of type `E`
+
+These are common programming idioms. The main complication comes when they are mixed in the same program.
```
keyboardDrummer, line 16: Consider leaving out line 16

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
```diff
@@ -0,0 +1,191 @@
...
+The Wrappers module holds some
+simple datatypes to support common patterns, such as optional values or the result of operations that can fail.
+These are additionally useful with Dafny's abrupt-termination-on-failure `:-` operator.
+
+Any user datatype can serve this purpose, as long as it has an `IsFailure?` predicate 
+(in which case it is known as a [_failure-compatible_](https://dafny.org/latest/DafnyRef/DafnyRef#sec-failure-compatible-types) type --- an FC-type). 
+
+In this library module Dafny defines three such types:
+- `Option<R>` - which is either `Some` with a value of type `T` or a `None` with no information
+- `Outcome<E>` - which is either `Pass` with no informatino or `Fail` with an error value of type `E` (often a `string`)
```
keyboardDrummer, line 13: `informatino`

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
```diff
@@ -0,0 +1,191 @@
...
+  var value: int;
+  // do various calculations
+  var index: int :- Find(s, 0, value);
+  // more calculations
+  return Some(value);
+}
+```
+
+You could just capture the result of `Find` using `:=` in a `Option<int>` variable and inspect it. But if the `None` condition is 
+generally a rare error, it is easy to forget to always check that each operation was successful. Instead, the `:-` changes the
```
keyboardDrummer, line 48: I think I prefer this text not to have line breaks, so the viewer can take care of the appropriate breaking.

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+You could just capture the result of `Find` using `:=` in a `Option<int>` variable and inspect it. But if the `None` condition is 
+generally a rare error, it is easy to forget to always check that each operation was successful. Instead, the `:-` changes the 
+control flow so that if a `None` value is returned from `Find`, the method immediately aborts, with the output value (which has
+the `Option<int>` type) getting that returned `None` value. So this operates something like exceptions.
+See the [reference manual](https://dafny.org/latest/DafnyRef/DafnyRef#sec-update-with-failure-statement)
+for more on the similarities and differences with exceptions.
+
+### Outcome
```
keyboardDrummer, line 54: Is this used a lot? C#'s anonymous datatypes `Option<(int x, string y)>` seem nicer

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
```diff
@@ -0,0 +1,191 @@
...
+<!-- %no-check -->
+```dafny
+:- Need(IsEverythingOK(), "failure in method M");
+```
+ 
+### Result
+
+A `Result` carries both a success value and failure information, with two separate, specifiable types. Its use is then very similar to `Option`.
+
+### Combining different FC-types in expressions
```
keyboardDrummer, line 96: While the acronym FC has been defined before (once, and was not used afterwards), I would not use it in the title here.

keyboardDrummer (approved):
Amazing! Love it

Source/DafnyStandardLibraries/CONTRIBUTING.md
```diff
@@ -0,0 +1,135 @@
...
+### On brittleness
+
+There are two sides to brittleness relevant to this project:
+
+**Brittleness of the standard libraries**: This project uses the current 
+[best known defense against brittleness](https://dafny.org/dafny/DafnyRef/DafnyRef#sec-verification):
+enforcing a tight upper-bound on the resources needed to verify each batch of assertions.
+For simplicitly, this project just sets a direct `--resource-limit`,
+rather than relying on the second-pass approach of the `dafny-reportgenerator`.
+I have set this limit aggressively low in the hopes that we can maintain it
```
jtristan, line 91: I would use a passive voice there.

Source/DafnyStandardLibraries/CONTRIBUTING.md
```diff
@@ -0,0 +1,135 @@
...
+even as new libraries are added, which means some existing code
+will need to be refactored somewhat to be added here.
+
+The downside of this approach is that it can be challenging to know when code is actually correct
+but cannot be verified efficiently enough.
+Feel free to temporarily raise this limit when developing libraries,
+but we should be very resistant to raising the limit in the checked in copy.
+
+**Brittleness of code using the standard libraries**: This one is tougher.
+There is tension between two different use cases:
```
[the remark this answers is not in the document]
robin-aws: Have chatted with @jtristan about this a bit too. Cut https://github.com/dafny-lang/dafny/issues/4697 to follow up with ideas.

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
robin-aws: 🎸

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
robin-aws: My general preference is https://sembr.org/ personally. I didn't want to edit this too heavily so we can see a clean diff against the original, though: https://github.com/dafny-lang/libraries/blob/master/src/dafny/Wrappers.md

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
robin-aws: I actually disagree with the statement that this is the most common use case. It's more important that returning an Outcome instead of a Result means you can say `:- CheckSomeThings(x)` as opposed to `var _ :- CheckSomeThings(x)`, where in the latter case it's too easy to do `var _ := CheckSomeThings(x)` which has no effect!

Rewriting.

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
robin-aws: Agreed about the titles, although "FC" is used a handful of times in the content.

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
robin-aws: Ended up just flipping the order of use cases.

jtristan (approved):

Source/DafnyStandardLibraries/CONTRIBUTING.md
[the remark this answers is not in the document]
MikaelMayer: We can deal with this at the library side by refining modules and expose more or less data, and also define careful export sets.
That way, customizing opaqueness is done at the import level.

I also wish users should also be able to refine a library module while exposing everything, and then in their refining module they could hide everything they don't want.

Source/DafnyStandardLibraries/src/dafnystdlibs/Wrappers.md
keyboardDrummer: I don't get it. Why wouldn't you want regular text to look the same when writing as when reading? If we're soft-wrapping it with no explicit line breaks when reading, why not also write it like that?

PR #4679, opened 2023-10-18
Move some code out of BoogieGenerator

keyboardDrummer:
### Description
- Move code out of BoogieGenerator enough that it gets below 5k lines and Rider starts showing normal hints in it again
- This PR uses partial classes to make it easy to move code, but much better would be to use classes to allow using encapsulation to enforce code structure.
- The grouping of code is rather ambiguous

### How has this been tested?
This is purely code moves. No additional testing is needed


atomb (approved):
I haven't verified that these are pure refactorings, without functional changes, but assuming they are it looks good.

PR #4681, opened 2023-10-18
Add replaceable modules

keyboardDrummer:
### Description
- Move some C# code from `Dafny.atg` and `Parser.frame` to a new file named `ParserNonGeneratedPart.cs`, so C# IDEs can provide tooling for this code.

- Enable the following syntax
```dafny
replaceable module FooModule {
  method Foo() returns (i: int) 
    ensures i >= 2
}

module Nesting {
  module ConcreteFoo replaces FooModule {
    method Foo() returns (i: int) {
      return 3;
    }
  }
}

method Main() {
  var x := FooModule.Foo();
  print x; // Prints 3
}
```

Check out one the test-cases for an example of how you define a multi-target application using this feature: https://github.com/dafny-lang/dafny/blob/66c1b1ee414bb4a5060507dbb0dd333697968e02/Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/placeholders/complex/user.dfy

### Caveats

Refining the `replaceable ` module is not supported, and it will not be supported for some time. There's a conflict between how refinement works and how replaceable modules work. 

Replaceable modules are implemented as a post-resolution feature. It swaps out references to the replaceable with the replacement. It has to work like this because `dafny verify` should work without the replacement being part of the source, so refinement can not rely on the replacement module being present.

Refinement, as it is currently implemented, copies non-abstract parts of the refinee into the refiner during resolution, so you would need access to the replacement module during resolution, if you want to further refine it. 

We could change refinement so it does use the replacement module _if_ its available, to influence refinement, but I consider that detrimental to maintainability so I think we should not do this. Refinement being implemented using copying is already problematic because it increases compilation time and generated code size. I think we should change refinement so it uses references to the refinee's members instead of copying them, and then we can also implement refinement of placeholders.

### How has this been tested?
- Add a happy flow test containing:
  - A placeholder module that is replaced
  - A placeholder module that refines an abstract module, and is replaced
  - A placeholder module that replaces another placeholder module, and is replaced
  - Various abstract imports using placeholders
- Add complex tests that use different back-ends for a common library, showing off the main driver behind the feature.
- Add a test that exposes all added errors
  - A placeholder module can not be refined
  - A placeholder module was not replaced
  - A placeholder module was replaced more than once
  - Cycle of replacement


Source/DafnyCore/AST/Grammar/ParserNonGeneratedPart.cs
robin-aws, line 1: +1 and thank you, I've been tempted to do this myself a few times :)

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/placeholders/complex/user.dfy
```diff
@@ -0,0 +1,10 @@
+// RUN: %baredafny run %args %s --target=java --build=%S/Build/build --input %S/wrappers.dfy --input %S/random.dfy --input %S/randomJava.dfy --input %S/Interop/__default.java --spill-target-code > "%t"
+// RUN: %baredafny run %args %s --build=%S/Build/build --input %S/wrappers.dfy --input %S/random.dfy --input %S/randomCSharp.dfy --input %S/Interop.cs --spill-target-code >> "%t"
+// RUN: %diff "%s.expect" "%t"
+
+method Main() {
+  var x := DfyRandom.GetRandomNat(1000000);
+  var y := DfyRandom.GetRandomNat(1000000);
+  var areEqual := x == y;
+  print "areEqual: ", areEqual;
```
robin-aws, line 9: This is a good case for using `expect` instead of printing and comparing, I think.

Also I don't love the idea of a test that by design could non-deterministically fail, no matter how small the probability. :) How about `expect 0 <= x < 1000000` instead?

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/placeholders/placeholderExecutionErrors.dfy.expect
```diff
@@ -0,0 +1,4 @@
+
+Dafny program verifier finished with 3 verified, 0 errors
+placeholderExecutionErrors.dfy(14,7): Error: A placeholder module may only be replaced once. Other replacing module: placeholderExecutionErrors.dfy(20,7)
+placeholderExecutionErrors.dfy(4,19): Error: Placeholder module must be replaced somewhere in the program
```
robin-aws, line 4: I'd be a bit more specific that this only applies when compiling, just so the error makes more sense in isolation

docs/DafnyRef/UserGuide.md
```diff
@@ -1727,7 +1727,31 @@ Detailed description of the `dafny build` and `dafny run` commands and
 the `--input` option (needed when `dafny run` has more than one input file)
 is contained [in the section on command-line structure](#command-line).
 
-### 13.7.2. C\#
+### 13.7.2. Placeholder modules
+To allow defining behavior that depending on which target language Dafny is translated to, requires different Dafny code to be implemented, Dafny has placeholder modules. Here follows an example:
```
robin-aws, line 1731: nit: I'd like to tweak this a little to describe the abstract feature first, then the primary motivation. Something like like "To support customizing behavior differently at compile-time for different contexts, Dafny has placeholder modules... This is particularly useful for defining behavior that depends on which target language Dafny is translated to".

I really like that the feature is generic enough to apply to other use cases so let's not sell it short!

robin-aws: In fact I'd promote at least most of this section to the reference manual since it is in fact a language feature, not just a tool feature.

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/placeholders/placeholderHappyflow.dfy
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+placeholder module BarModule refines SuchAbstract {
+  method Bar() returns (i: int) 
+    ensures i >= 1
+}
+
+
+placeholder module BazModule replaces BarModule {
```
robin-aws, line 20: This is nice, the fact that you can have chains of replacement means I don't personally care much that you can't refine after you placehold.

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/placeholders/complex/randomJava.dfy
```diff
@@ -0,0 +1,57 @@
+
+module RandomJava replaces DfyRandom {
+  import opened Interop
+  import opened JavaLang
+  import opened JavaRest
+  
+  method GetRandomNat ... {
+    // Ignore the ceiling parameter for now
```
robin-aws, line 8:
```suggestion
```

:)

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/placeholders/complex/user.dfy
```diff
@@ -0,0 +1,10 @@
+// RUN: %baredafny run %args %s --target=java --build=%S/Build/build --input %S/wrappers.dfy --input %S/random.dfy --input %S/randomJava.dfy --input %S/Interop/__default.java --spill-target-code > "%t"
+// RUN: %baredafny run %args %s --build=%S/Build/build --input %S/wrappers.dfy --input %S/random.dfy --input %S/randomCSharp.dfy --input %S/Interop.cs --spill-target-code >> "%t"
```
robin-aws, line 2: You can definitely see how nice it would be to have the project file specify all this, and even better to have a default convention for the code layout of a Dafny project with extern code. :)

robin-aws (dismissed):
Loving this feature, and I'm getting pretty comfortable with the terminology/syntax, although I want to give other core maintainers the chance to object.

Just reviewed the feature design and tests/documentation for now, can circle back to review the actual implementation.

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/placeholders/complex/wrappers.dfy
```diff
@@ -0,0 +1,205 @@
...
+/** This module provides various kinds of failure-compatible datatypes, including Option.
+- Option<R> (Some or None) with a type parameter R that is the type of the Some value
+- Outcome<E> (Pass or Fail) with a type parameter E that is the type of a failure value
+- Result<R,E> which has both a value type and a failure type
+All three of these may be used with `:-` as Dafny failure-compatible types
+
+The module also defines two forms of `Need`, that check the truth of a predicate and
+return a failure value if false.
+ */
+module DafnyStdLibs.Wrappers {
```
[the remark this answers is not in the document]
keyboardDrummer: Totally, but your PR isn't merged yet ! I planned on removing this later.

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/placeholders/complex/user.dfy
keyboardDrummer: > Also I don't love the idea of a test that by design could non-deterministically fail

I changed the chance to 10^-30, so it'll fail after our planet disintegrates

keyboardDrummer: > This is a good case for using expect instead of printing and comparing, I think.

Not sure what the advantage is, but sure !

docs/DafnyRef/UserGuide.md
keyboardDrummer: > In fact I'd promote at least most of this section to the reference manual since it is in fact a language feature, not just a tool feature.

Not sure where to put it then, there is no "runtime behavior" section, and I think this is a pretty good spot for now.

MikaelMayer:
I love your new placeholder feature, I think you are doing great with the implementation and remarks about refinement, it makes sense to just talk about replacement. Thanks for explaining all of this.
I'm definitely going to help you in any way I can to make this feature live, because users will definitely need this feature and the faster the better, especially since you made it ready.

However, with the current design, it feels like a step backwards that makes me uncomfortable.
Before, we did have includes statements so that we could selectively include what module to inject, and the same file could be compiled with different backends by having different backends include common code. As of today, this is still possible and the recommended way of injecting modules.
Therefore, to compile/run, one would just create files that include all the correct files and run it. No big deal, it's persistent, configurable at will.

With project files, we finally removed the include statements, hurray. A project is now aware of all the dafny files in scope, amazing.
However, in your example, you are no longer using project files. And that's understandable: you want to selectively use which files to include, when in the project file, you have to decide globally for every configuration.
But then, it feels like all your --input on the CLI are just the migrated "include" that we no longer write, except they are no longer saved with the files.

Do you have any ideas on how I could feel more comfortable with your feature? Hacking my brain is not an option :slightly_smiling_face:
Here are options that would make me feel more comfortable. Can you think of other options?
- Have as part of your replace module a way to state for which case the compiler would choose this module over any other moduel to replace the placeholder. Similar to CSS, with media queries which describe configurations like (max-width:1200) or (orientation:portrait) or just print. That way, no more --input, the compiler would just find out what to replace. If two or more can replace the module in the given configuration, it would just pick the first one and emit a warning about the second.
- Describe a plan in the project file a way to include/exclude files based on certain compilation options. I feel less good with this option somehow, because most of the time, we know in the file when it should be included, and also it prevents being able to put everything in the same file, which we always allowed till today.

Even if you don't implement anything that would make me comfortable now, knowing what is your next plan and what you are going to commit to might help me feel more comfortable.

Also, whatever you do, please make sure you have tests that capture the following cases:
* Two modules claim to replace the same module in the same file, and in different files
* A placeholder module does not have any replacement
* A placeholder module does have a replacement but not in the language of interest.

keyboardDrummer:
> * Describe a plan in the project file a way to include/exclude files based on certain compilation options. I feel less good with this option somehow, because most of the time, we know in the file when it should be included, and also it prevents being able to put everything in the same file, which we always allowed till today.

If you were working with multiple target languages, you already had to supply different target-language source files for different targets, so I don't see a regression in UX in that regard when using the new feature.

Before:
```
dafny translate cs dafnyFileWithExterns.dfy externs.cs
dafny translate java dafnyFileWithExterns.dfy externs.java
```
Now
```
dafny translate cs dafnyFileWithExterns.dfy csReplacement.dfy
dafny translate java dafnyFileWithExterns.dfy javaReplacement.dfy
```

> Even if you don't implement anything that would make me comfortable now, knowing what is your next plan and what you are going to commit to might help me feel more comfortable.

I don't know what we'll do, but I think we have many options that can provide the different types of UX we've mentioned so far. I think we differ in our expectation of whether it will be ideal to specify configuration for multiple back-ends in a single file.

My feeling is that if you have a standalone dafny application, then you don't need the placeholder feature. However, if you have a Dafny library that you integrate in multiple other language applications, then you would have one `src/shared` folder containing a project file that does *not* specify the replacement files.

You would have another top level folder per target language `src/cs` that each contains another project file which references (or inherits from?) the shared project file, and specifies the additional Dafny files. That way, the IDE would be able to find a project file for each Dafny file is comes across, but none of these projects would contain files that do not occur together during compilation.

keyboardDrummer:
> More precisely, if we choose which Dafny files we can put on the command line, then you don't need a new placeholder feature. You can just make your library in `src/shared` abstract and import an abstract module (so that you can verify it independently)

You can't. Dafny will not allow you to do that. 

```dafny
abstract module Shared {
  method Foo() {

  }
}

module Consumer {
  import Shared // a compiled module (Consumer) is not allowed to import an abstract module (Shared)
  method Main() {
    Shared.Foo();
  }
}
```

MikaelMayer:
> > More precisely, if we choose which Dafny files we can put on the command line, then you don't need a new placeholder feature. You can just make your library in `src/shared` abstract and import an abstract module (so that you can verify it independently)
> 
> You can't. Dafny will not allow you to do that.
> 
> ```
> abstract module Shared {
>   method Foo() {
> 
>   }
> }
> 
> module Consumer {
>   import Shared // Error: a compiled module (Consumer) is not allowed to import an abstract module (Shared)
>   method Main() {
>     Shared.Foo();
>   }
> }
> ```
> 
> You might be able to do something like:
> 
> ```
> abstract module Shared {
>   method Foo()
> }
> 
> abstract module Consumer {
>   import X : Shared
>   method Main() {
>     X.Foo();
>   }
> }
> 
> module SharedJava refines Shared {
>   method Foo() {
> 
>   }
> }
> 
> module ConsumerJava refines Consumer {
>   import X = SharedJava
> }
> ```
> 

Great, I was about to come up with an very similar example:
```
abstract module FileIO {
  method openFile(n: string) returns (s: string)
}
abstract module Shared {
  import File : FileIO
  method Test() {
    var x := File.openFile("test.json");
    print x;
  }
}

module CSharpFileIO refines FileIO {
  method {:extern "System.file.open"} OpenThatFile(n: string) returns (s: string)
  method openFile(n: string) returns (s: string) {
    s := OpenThatFile(n);
  }
}

module CSharpConsumer refines Shared {
  import File = CSharpFileIO
  method Main() {
    Test();
  }
}
```

> But then you need a separate consumer for each target, and the goal was to verify all consumer code in a target-agnostic way.

The method above with abstract imports  still verifies all consumer code in a target agnostic way, so the goal is already accomplished, isn'it?
The difference between what we can do today and what your PR brings is that we would be able to omit, in a customer project
```
module Customer refines Shared {
  import Library = LibraryJava
}
```
which is only 3 lines of Dafny.

It looks like from this PR that you would still have a consumer/Dafny file for each target. In your example above, you are giving a body to SharedJava.Foo, but since you are giving a body in Shared.Foo you don't need to repeat this code at all.

keyboardDrummer:
> The method above with abstract imports still verifies all consumer code in a target agnostic way, so the goal is already accomplished, isn'it?

No because there is no target independent Consumer, there's only `CSharpConsumer`. You can't write all your consumer code, and then at some later point in time decide to run it against a new target XYZ by only supplying the library files for XYZ, without updating your consumer.

In the standard library case, with the solution from this PR, you only write your consumer code once against a target agnostic standard library, and can then run it against all back-ends.

With the approach you suggest, you need to write specific code for each back-end.

> The difference between what we can do today and what your PR brings is that we would be able to omit, in a customer project
> ```
> module Customer refines Shared {
>  import Library = LibraryJava
>}
> ```

Agreed ! And I think this makes all the difference, because now you don't have to think about targets when you write your code. You can have a huge Dafny ecosystem with tons of libraries all unaware of targets, that use a few target-aware libraries. Long after this ecosystem is written, you can introduce a new target, and call all the libraries in the ecosystem by supplying target-specific files for the few target-aware libraries that it uses.

It seems that you're worried about the UX of a Dafny standalone application that uses a target-aware library. Without this PR, that application will need target-specific code and target-specific CLI calls. With this PR, it will only need target-specific CLI calls. We can remove the need for the latter through package-management, but even before that, this PR already provides a completely target-agnostic experience when you're using the standard library, because the standard library will auto-select the right target-specific files depending on which target you're translating to.

RustanLeino:
Is there really a need to have the new keyword/concept `replaces`? Couldn't we just use the `refines` keyword for this purpose? That is, when a module (like `ConcreteFoo`) says it `refines` a module (here, `FooModule`), that the refined module (`FooModule`) is a `placeholder module`, then do whatever needs to be done.

Also, note that `ConcreteFoo` really does _refine_ `FooModule`, since it's filling in a method body. I imagine there will be other things it also wants to fill in, like type definitions. So, again, this feels like `refines` to me.

MikaelMayer:
> because the standard library will auto-select the right target-specific files depending on which target you're translating to.

This is what I want, but I don't see your PR doing anything like this. Did I miss something? Could you point me to the part where the standard library will auto-select the right target specific files depending on which target I am translating to?

MikaelMayer:
> Agreed ! And I think this makes all the difference, because now you don't have to think about targets when you write your code. You can have a huge Dafny ecosystem with tons of libraries all unaware of targets, that use a few target-aware libraries. Long after this ecosystem is written, you can introduce a new target, and call all the libraries in the ecosystem by supplying target-specific files for the few target-aware libraries that it uses.

I don't understand at all the "you don't have to think about targets when you write your code."

Let's say we have an ecosystem definiting the following library:
```
abstract module FileIO {
  method openFile(n: string) returns (s: string)
}
abstract module Shared {
  import File : FileIO
  method Test() {
    var x := File.openFile("test.json");
    print x;
  }
}
```
It also defines a few implementations for this library that people can load if they use C#
```
module CSharpFileIO refines FileIO {
  method {:extern "System.file.open"} OpenThatFile(n: string) returns (s: string)
  method openFile(n: string) returns (s: string) {
    s := OpenThatFile(n);
  }
}
```
Someone extending the library can do so without materializing the dependency.
```
abstract module SharedDecrypt refined Shared {
  method CallTest() {
    print "Decrypt";
    Test();
  }
}
```
Let's say we are in the year 2030 and there is a new language named "Blah" that someones has a backend for, but the library above was not written for blah. That's fine, that user will define the placeholder

```dafny
module BlahFileIO refines FileIO {
  method {:extern "/!!!++++-"} OpenThatFile(n: string) returns (s: string)
  method openFile(n: string) returns (s: string) {
    s := OpenThatFile(n);
  }
}
```

and then they can write their application as usual
```
module BlahConsumer refines DecryptShared {
  import FileIO = BlahFileIO
  method Main() {
    CallTest();
  }
}
```

It looks like you are trying to achieve that your library be not defined as "abstract" so that there is late-binding and linking by the compiler. This is fine and I can accept that. I'm not saying the above cannot be done with your feature, I'm just saying that introducing 2 keywords for that is a bit redundant to say the least, especially because it does not remove the boilerplace of all these files to include on the side.

Correct me if I'm wrong @robin-aws , but when we will include the standard library in Dafny itself, we will include files, based on what the user imports, not based on the compiler's mode. So ideally, our standard library has late bindings for all supported languages and the compiler will just pick the one that is the most relevant to the compilation mode, or say something like "language blah does not have a replacing module for X defined in Y. Please define one yourself."

This is where your PR could shine. It could make this process so easy by letting us define a placeholder and immediately variants for various languages/compiler configurations. This would make it possible to ship core Dafny libraries with wrappers around many languages. People could also use these wrappers directly if they wanted to.

keyboardDrummer:
> It could make this process so easy by letting us define a placeholder and immediately variants for various languages/compiler configurations. This would make it possible to ship core Dafny libraries with wrappers around many languages. People could also use these wrappers directly if they wanted to.

That's an interesting feature, but I think better to think about after we add the feature in this PR. We have one immediate problem which is enabling the standard library to be used in a target-agnostic way. Both the feature you describe and this PR enable solving that immediate problem. (where this PR also requires the compiler to include different target-specific standard library files based on the provided --target)

The feature you describe also solves the problem of having to specify target specific files on the CLI, and it doesn't require anything complex like package management, which is great. However, it only does this for the back-ends for which the target-aware code was published. I think we must have a way to specify target-specific code for a target-aware library, _after_ that library was written. For example, one author could work on the specification of FileIO, another author could work on FileIoForRust, and another on FileIoForJava.

I think it would be better to first resolve that must have, and only after that consider providing the UX feature that you describe. I think the syntax this PR provides could extend nicely to provide the feature you describe, through something like `module X replaces Y for <target>`

---

> Let's say we have an ecosystem definiting the following library:

In your example there's only one target-aware import `import File = BlahFileIO`, which is great. However, it's not clear to me that it's always possible to have only one such import in a program per target-aware library. In your example there's a chain of refinements, and only one usage of the target-aware library, which I think makes things simpler. Here's an example where I think the amount of target-aware imports required is bigger than the amount of target-aware libraries:

```dafny
// Blah is a target
// Blah_independent_ecosystem.dfy
abstract module FileIO {
  method openFile(n: string) returns (s: string)
}

abstract module FileIoBasedLibrary1 {
  import File : FileIO
  method Test() {
    var x := File.openFile("test.json");
    print x;
  }
}

abstract module FileIoBasedLibrary2 {
  import File : FileIO
  method Test() {
    var x := File.openFile("test.json");
    print x;
  }
}

// Blah_maintainer.dfy
module BlahFileIO refines FileIO {
  method {:extern "/!!!++++-"} OpenThatFile(n: string) returns (s: string)
  method openFile(n: string) returns (s: string) {
    s := OpenThatFile(n);
  }
}

// Consumer.dfy
module BlahFileIoBasedLibrary1 refines FileIoBasedLibrary1 {
  import File = BlahFileIO // First initialization of FileIO
}

module BlahFileIoBasedLibrary2 refines FileIoBasedLibrary2 {
  import File = BlahFileIO // Second initialization of FileIO
}

module Consumer {
  import BlahFileIoBasedLibrary1
  import BlahFileIoBasedLibrary2
  method Main() {
    BlahFileIoBasedLibrary1.Test();
    BlahFileIoBasedLibrary2.Test();
  }
}
```

keyboardDrummer:
> Is there really a need to have the new keyword/concept `replaces`? Couldn't we just use the `refines` keyword for this purpose? That is, when a module (like `ConcreteFoo`) says it `refines` a module (here, `FooModule`), that the refined module (`FooModule`) is a `placeholder module`, then do whatever needs to be done.

Are you suggesting that when an abstract module is refined exactly once, it would have the replacement semantics of this PR?

I think not having a keyword to distinguish replacement from refinement, would have the following downsides:
- If you have a file library.dfy containing `FooModule`, then the signature of the library changes depending on how often it refines `FooModule`, which makes it easy to accidentally change the library signature:
  - If it refines `FooModule` zero times, consumers can "replace" `FooModule` by refining it exactly once.
  - If it refines `FooModule` once, then consumers can't replace `FooModule`, but they can refer to it as if it was concrete.
  - If it refines `FooModule` more than once, consumers can't refer to it as if it was concrete.
- It would not be possible to refine the replacement of `FooModule` (`ConcreteFoo`) using `refines FooModule`
- It would not be possible to refine `ConcreteFoo`, because then `FooModule` would no longer have a unique refinement.

I think not having a keyword to distinguish a placeholder from an abstract module would have these downsides:
- It would not be possible to refine the replacement of `FooModule` (`ConcreteFoo`) using `refines FooModule` because that would already mean refining the `FooModule` module (which is useless, because it requires target-specific code to be implemented)
- Users can always refer to any `abstract module` as if it is concrete, while it might not be intended to be replaced.

> Note that `ConcreteFoo` really does _refine_ `FooModule`, since it's filling in a method body. I imagine there will be other things it also wants to fill in, like type definitions. So, again, this feels like `refines` to me.

You're right, it is refining it. The refinee `ConcreteFoo` can use the full feature set of refinement when replacing `FooModule`. However, during compilation it is also replacing all other references to `FooModule` with `ConcreteFoo`, so it's not just refining it.

I think an alternative that is very close to this PR is to reuse the abstract/refines syntax, which would help make the large overlap in behavior clear, but to add an additional keyword to resolve the downsides mentioned above. We could have `placeholder module` -> `unique abstract module` and `X replaced Y` -> `X uniquely refines Y`. 

However, I think that has these downsides:
- Requires typing more keywords.
- Makes it less clear what happens if you use regular refinement on a placeholder module.

While I think the advantage of making the link with refinement is tiny because users have to look up the behavior of the new keywords anyways.

An even closer alternative to this PR would be to use `replaceable module` instead of `placeholder module`, then we are only introducing ~1 keyword instead of 2. I don't have a strong preference for either.

RustanLeino (dismissed):
Nice work! I look forward to using "replaceable" modules for better foreign-language interop.

robin-aws:
> Correct me if I'm wrong @robin-aws , but when we will include the standard library in Dafny itself, we will include files, based on what the user imports, not based on the compiler's mode. So ideally, our standard library has late bindings for all supported languages and the compiler will just pick the one that is the most relevant to the compilation mode, or say something like "language blah does not have a replacing module for X defined in Y. Please define one yourself."

My plan is that the `DafnyStandardLibraries.doo` file that `--standard-libraries` implicitly adds for verification will just have the abstract/replacable `FileIO` module in it, let's say in a `FileIO.dfy` file. There will be other Dafny files defining the concrete module implementation for each target, `FileIOForJava.dfy` for example. And for completeness let's say there's a supporting Java file like `FileIO.java`. All three files will be compiled into the Java runtime: the Java files output from compiling `FileIO.dfy` and `FileIOForJava.dfy`, and `FileIO.java`. That means there will be a Makefile command something like this to run before building `DafnyRuntime.jar`:

```make
dafny translate java FileIO.dfy FileIOForJava.dfy ... --output DafnyStandardLibraries
cp -r DafnyStandardLibraries-java DafnyRuntimeJava/src/main/dafny-generated
```

I didn't set this up in the initial PR to add stdlib support because #511 means you hit duplicate definitions when you try to add Dafny-generated code to the runtimes, much worse than existing contexts, so I'm fixing that first. https://github.com/dafny-lang/dafny/pull/4658 already adds some similar make targets to pre-compile the `_System` module into the runtimes, though.

Having project file features for collecting the right input files into `dafny translate` would certainly be nice, but in the actual Makefile it's not too hard to collect them with globs. 

I think you're right that this can work without `replaceable` modules, by having consumers refer to an undefined `FileIO` module that is filed in at compile time by one of the target-specific versions. I can say from doing [something similar for the `UnicodeStrings` module](https://github.com/dafny-lang/libraries/blob/master/src/Unicode/UnicodeStrings.dfy) that it's a real pain when writing the code, because you have unresolved references the IDE can't deal with, so you're stuck with verifying using the CLI.

robin-aws:
To clarify the content of the files above:

```dafny
// FileIO.dfy
replaceable module DafnyStdLibs.FileIO {
  method OpenFile(n: string) returns (s: string)
}
```

```dafny
// FileIOForJava.dfy
module DafnyStdLibs.JavaFileIO replaces DafnyStdLibs.FileIO {
  method OpenFile(n: string) returns (s: string) {
    // Likely other validation/conversion to do here
    s := INTERNAL_openFileExtern(n);
  }

  method {:extern} INTERNAL_openFileExtern(n: string) returns (s: string)
}
```

```java
package DafnyStdLibs.JavaFileIO;

public class DafnyStdLibs.JavaFileIO {
  public static Dafny.Sequence<Character> INTERNAL_openFileExtern(Dafny.Sequence<Character> n) {
  }
}
```

MikaelMayer:
> To clarify the content of the files above:
> 
> ```
> // FileIO.dfy
> replaceable module DafnyStdLibs.FileIO {
>   method OpenFile(n: string) returns (s: string)
> }
> ```
> 
> ```
> // FileIOForJava.dfy
> module DafnyStdLibs.JavaFileIO replaces DafnyStdLibs.FileIO {
>   method OpenFile(n: string) returns (s: string) {
>     // Likely other validation/conversion to do here.
>     // This is close to the existing implementation that does all of that in Java code instead.
>     s := INTERNAL_openFileExtern(n);
>   }
> 
>   method {:extern} INTERNAL_openFileExtern(n: string) returns (s: string)
> }
> ```
> 
> ```java
> // FileIO.java
> package DafnyStdLibs.JavaFileIO;
> 
> public class DafnyStdLibs.JavaFileIO {
>   public static Dafny.Sequence<Character> INTERNAL_openFileExtern(Dafny.Sequence<Character> n) {
>     // ...actually read the file
>   }
> }
> ```
> 
> So Dafny user code depending on the standard library only sees what's in `FileIO.dfy`, but when compiled to Java ends up depending on the contents of all three, which are in `DafnyRuntime.jar`.

Thanks for explaining the details.

Right now you intend to compile FileIOForJava.dfy into java and include it in the runtime library so that it's never compiled/included twice, correct?
The problem I have with this approach is that it's no longer possible to replace the standard library shipped with a version of Dafny with another standard library (i.e. an updated version)

The problem of duplication essentially comes from if someones compiles two applications using Dafny into Java, and these two applications will thus define the same binaries. Then that user can no longer create an application that uses the previous two applications.
However, if that user delays the compilation and only perform compilation after including all the Dafny libraries and target-specific backend file, then no duplication exists.
This is a strong hint for me that a better way to ship the standard library for Dafny is to ship the files, stored in plain in Dafny.exe and regurgitated on demand, and not include them or precompile them in the Dafny Runtime.

robin-aws:
> The problem I have with this approach is that it's no longer possible to replace the standard library shipped with a version of Dafny with another standard library (i.e. an updated version)

I'm not too concerned with that since for now the standard libraries are tightly coupled to the Dafny distribution, and in some cases we'll take advantage of that to provide libraries like `Interop` that may need to change with each Dafny version.

> This is a strong hint for me that a better way to ship the standard library for Dafny is to ship the files, stored in plain in Dafny.exe and regurgitated on demand, and not include them or precompile them in the Dafny Runtime.

I do think it would be unfortunate to provide standard libraries but force users that need a single compiled version of them to build and distribute it themselves. We could consider distributing the compiled standard libraries as a separate package in each language instead of adding them to the runtimes, but that seems like more overhead for both us and customers without a lot of benefit right now.

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/replaceables/complex/Inputs/randomCSharp.dfy
```diff
@@ -0,0 +1,46 @@
...
+    Int32.MaxValueValue();
+    var resultInt32 := random.Next(IntToInt32(0).Extract(), ceilingInt32);
+    var resultInt := Int32ToInt(resultInt32);
+    return resultInt as nat;
+  }
+}
+
+module Interop {
+  import opened CSharpSystem
+  import opened DafnyStdLibs.Wrappers
```
MikaelMayer, line 19: Yay!

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/replaceables/complex/user.dfy
```diff
@@ -0,0 +1,17 @@
...
+  var x := DfyRandom.GetRandomNat(1000000);
+  var y := 0;
+  var i := 5;
+  while(x != y && i != 0)
+    decreases i
+  {
+    y := DfyRandom.GetRandomNat(1000000);
+    i := i - 1;
+  }
+  expect x != y;
```
MikaelMayer, line 16:
```suggestion
  expect i == 0 || x != y;
```
I hope you don't mind too much but I'd feel better for maintenance if the expect was always true :-)

docs/DafnyRef/UserGuide.md
```diff
@@ -1733,7 +1733,35 @@ Detailed description of the `dafny build` and `dafny run` commands and
...
+
+module ConcreteFoo replaces Foo {
+  method Bar() returns (i: int) {
+    return 3; // Main will print 3.
+  }
+}
+// ConcreteFoo can be swapped out for different replacements of Foo, to customize runtime behavior.
+```
+
+When replacing a placeholder module, the same rules apply as when refining an abstract module. However, unlike an abstract module, a placeholder module can be used as if it is a concrete module. When executing code, using for example `dafny run` or `dafny translate`, any program that contains a placeholder module must also contain a replacement of this placeholder. When using `dafny verify`, placeholder modules do not have to be replaced.
```
MikaelMayer, line 1760: I think I would love it if I could define a compilable replaceable module that has all its declaration bodies (e.g. a slow reference implementation that I can use to run my model and iterate on the design), and replaces module would actually swap it with e.g. a better implementation.

But that way, the replaceable module could be enough, it's just it would make it possible to make a dependency injection and allow for mocked testing for example.

How hard would it be/acceptable to let 100% defined replaceable modules be still compiled as usual if no replacement module is found?

docs/dev/news/4681.feat
```diff
@@ -0,0 +1 @@
+Introduce placeholder modules, which can be used to help define Dafny applications that translate to multiple target languages.
```
MikaelMayer, line 1:
```suggestion
Introduce replaceable modules, which can be used to help define Dafny applications that translate to multiple target languages.
```

MikaelMayer (commented):
Looks great! A few tweaks but otherwise the documentation is good, the tests are good (I would like to see at least one test where the replaceable module is already 100% defined by itself, whatever semantics you think you'd choose)

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/replaceables/complex/user.dfy
keyboardDrummer: :-( but then it doesn't test the randomness. Anyways, made the change.

docs/DafnyRef/UserGuide.md
keyboardDrummer: > How hard would it be/acceptable to let 100% defined replaceable modules be still compiled as usual if no replacement module is found?

For that we must enable module refinement to override bodied functions and methods.

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/replaceables/complex/user.dfy
MikaelMayer: hmmm. Ok. In that case, can you mark Main as `decreases *` and that way you can also mark the loop as `decreases *`. Then you are truly testing for randomness, but now it's no longer a probability on the test failing or not, it's how long it will take.

docs/DafnyRef/UserGuide.md
MikaelMayer: Oh, interesting. I thought replacement was meant to actually replace methods while keeping their signatures. To me that would be one of the biggest advantages of replaceable modules.

MikaelMayer (dismissed):
Good for me. Thanks.

Source/DafnyCore/Compilers/ExecutableBackend.cs
```diff
@@ -30,10 +30,36 @@ public abstract class ExecutableBackend : IExecutableBackend {
   public override string ModuleSeparator => Compiler.ModuleSeparator;
 
   public override void Compile(Program dafnyProgram, ConcreteSyntaxTree output) {
+    InstantiatePlaceholders(dafnyProgram);
```
MikaelMayer, line 33:
```suggestion
    InstantiateReplaceableModules(dafnyProgram);
```
One name per concept makes the code easier to read

PR #4682, opened 2023-10-18
Method reads fix

RustanLeino:
### Description

* Move initial processing of method `reads` clauses from `Method.Resolve` (which is called during name resolution + type inference) to later in the Resolver pipeline. This not only helps declutter the name resolution phase, but it also makes `reads` processing apply to both (legacy and refresh) resolvers.
* Fix various scoping issues of attribute resolution (e.g., include function result name in the scope).
* Remove some old code in new resolver for things that are now being checked later in the resolution pipeline.
* Make `{:warnShadowing}` be the same as `{:warnShadowing true}` (not `{:warnShadowing false}`)

Minor changes:

* fix: (new resolver) Resolve attributes of `:| assume`.
* fix: Reuse `.MethodCallInfo` if called to compute it again
* fix: Include method `reads` clauses in pretty print

### How has this been tested?

The `dafny0/AttributeChecks.dfy` and `dafny0/Shadows.dfy` tests have been updated.


Source/DafnyCore/Resolver/ModuleResolver.cs
```diff
@@ -1130,12 +1130,14 @@ public partial class ModuleResolver {
       // ---------------------------------- Pass 1 ----------------------------------
       // This pass does the following:
       // * desugar functions used in reads clauses
+      // * fills in "reads *" clauses on methods, when --reads-clauses-on-methods is used
```
robin-aws, line 1133: Just to confirm, do you think is this the "right" place to set defaults in general? From the original PR, we agreed it might have been cleaner to move in the other direction, to immediately fill in the default during parsing.

Source/DafnyCore/AST/Grammar/Printer.cs
```diff
@@ -1018,6 +1018,9 @@ public class Printer {
 
       int ind = indent + IndentAmount;
       PrintSpec("requires", method.Req, ind);
+      if (method.Reads.Expressions != null) {
```
robin-aws, line 1021: An annoying oversight on my part in the original PR. FYI I'd really like to switch the library backend on for `%testDafnyForEachCompiler` so that we get more automated testing of the pretty printer (since .doo files round trip programs through it)

robin-aws (approved):
LGTM, just questions and food for thought.

Do you think it's tractable to make a cheap AST visitor that asserts everything is resolved after the resolver finishes? That seems to be a common them with the fixes to resolving attributes here.

Source/DafnyCore/Resolver/ModuleResolver.cs
RustanLeino: (Responding a bit late.) If a default can be determined already in the parser, that may be a good place. At the other extreme, if the default matters only for verification (which I think is the case here), then we could choose never to update the AST with the default, and instead just take the appropriate action in the verifier. Here, it's now done in "Pass 1", which is alright. The one place we should avoid placing it, I think, is in the name resolution + type inference ("Pass 0"), because that pass is quite delicate.

Issue #3766, opened 2023-03-20
Recursive datatype has incorrectly initialized Java type descriptor


alex-chew:
### Dafny version

4.0.0.50303 (also occurs in 3.10)

### Code to produce this issue

```dafny
// Good

datatype Tree = Tree(ts: seq<Tree>)

datatype Nothing = Nothing
{
  predicate IsFailure() { false }
}

method Main() {
  var nothing: Nothing := Nothing;
  :- expect nothing;
}

// Bad

datatype Tree = Tree(ts: seq<Tree>)

datatype Nothing<T> = Nothing
{
  predicate IsFailure() { false }
}

method Main() {
  var nothing: Nothing<Tree> := Nothing;
  :- expect nothing;
}
```

### Command to run and resulting output

```code
# Good

$ dafny run --target:java Tmp.dfy
Dafny program verifier finished with 1 verified, 0 errors

$ echo $?
0

# Bad

$ dafny run --target:java Tmp.dfy
Dafny program verifier finished with 1 verified, 0 errors
Exception in thread "main" java.lang.ExceptionInInitializerError
        at _System.__default.Main(__default.java:16)
        at _System.__default.__Main(__default.java:21)
        at Tmp.lambda$main$0(Tmp.java:11)
        at dafny.Helpers.withHaltHandling(Helpers.java:279)
        at Tmp.main(Tmp.java:11)
Caused by: java.lang.NullPointerException: Cannot invoke "dafny.TypeDescriptor.newArray(int)" because "type" is null
        at dafny.ArrayDafnySequence.empty(DafnySequence.java:506)
        at dafny.DafnySequence.empty(DafnySequence.java:68)
        at _System.Tree.<clinit>(Tree.java:39)
        ... 5 more

$ echo $?
3
```

### What happened?

The bad program crashes with a `NullPointerException`, whereas the good program run without errors. The only difference between the two programs is that `Nothing` is parametrized by a recursive datatype in the bad program, but is non-generic in the good program.

### What type of operating system are you experiencing the problem on?

Mac

alex-chew:
The compiled `Tree` class includes the following snippet:

```java
public class Tree {
  // ...

  private static final Tree theDefault = _System.Tree.create(dafny.DafnySequence.<Tree> empty(Tree._typeDescriptor()));
  public static Tree Default() {
    return theDefault;
  }
[...]

PR #4694, opened 2023-10-20
Bug fix for Issue 3766

ssomayyajula:
### Description

Fix for #3766 with an explanation for the solution [here](https://github.com/dafny-lang/dafny/issues/3766#issuecomment-1771189785).

### How has this been tested?

Both tests from the original issue pass (`github-issue-3766-a.dfy` and `-b.dfy`) as well as a smaller one that doesn't use generics but still crashed before the fix (`-c.dfy`).


alex-chew (approved):
LGTM

PR #4706, opened 2023-10-23
chore: Always specify Java version when testing

fabiomadge:


atomb (approved):
Thanks for fixing this!

PR #4709, opened 2023-10-24
Relations

### LINKED ISSUE #4644: Standard library: Relations
Common properties of functions, such as Transitivity.

Existing source:

[dafny-lang/libraries/src/Relations.dfy](https://github.com/dafny-lang/libraries/blob/master/src/Relations.dfy)
[dafny-lang/libraries/src/Functions.dfy](https://github.com/dafny-lang/libraries/blob/master/src/Functions.dfy)

keyboardDrummer:
Fixed #4644

### Description
- Extract some methods in statement boogie generations, to improve readability.
- Fix a bug that prevented parsing .doo files that contain identifiers starting with underscores, which are sometimes generated as part of building a .doo file.
- Add modified version of the `Relations.dfy` and `Function.dfy` file from `https://github.com/dafny-lang/libraries` to `https://github.com/dafny-lang/dafny/tree/master/Source/DafnyStandardLibraries`

### How has this been tested?
Tests have been added to `Source/DafnyStandardLibraries/examples`


Source/DafnyStandardLibraries/src/DafnyStdLibs/Functions.dfy
```diff
@@ -0,0 +1,32 @@
...
+module DafnyStdLibs.Function {
+
+  ghost predicate Injective<X(!new), Y>(f: X --> Y)
+    reads f.reads
+    requires forall x :: f.requires(x)
+  {
+    forall x1, x2 :: f(x1) == f(x2) ==> x1 == x2
+  }
+
+  ghost predicate Commutative<T(!new),U(!new)>(f: (T,T) -> U)
```
robin-aws, line 19:
```suggestion
  ghost predicate Commutative<T(!new), U(!new)>(f: (T,T) -> U)
```

Source/IntegrationTests/TestFiles/LitTests/LitTest/git-issues/git-issue-1514.dfy
```diff
@@ -1,7 +1,6 @@
-// RUN: %testDafnyForEachCompiler "%s" --refresh-exit-code=0 -- --relax-definite-assignment
+// RUN: %testDafnyForEachCompiler "%s" --refresh-exit-code=0 -- --standard-libraries --relax-definite-assignment
```
robin-aws, line 1: Thanks for cleaning these up. I think limited use of `--standard-libraries` in tests is fine as long as we don't end up adding it everywhere, and for new tests I'd prefer just adding the bare minimum necessary for the test (e.g. just `Option<T> = None | Some(value: T)` rather than the whole thing from the stdlibs)

Source/DafnyStandardLibraries/src/DafnyStdLibs/Relations.dfy
robin-aws, line 1: We should bring over some subset of https://github.com/dafny-lang/libraries/blob/master/src/dafny/Relations.md here as well.

Source/DafnyCore/AST/Grammar/ProgramParser.cs
```diff
@@ -98,12 +97,13 @@ public class ProgramParser {
   }
 
   private DfyParseResult ParseFileWithErrorHandling(DafnyOptions options,
+    bool parseAsDooFile,
```
robin-aws, line 100: There's also the alternative of ensuring doo files contain only the original program text, before resolution/refinement/etc. I think this is moving in the right direction though, because eventually it would be great to not re-resolve doo files after parsing.

Source/DafnyLanguageServer/Language/CachingParser.cs
```diff
@@ -26,13 +27,18 @@ public class CachingParser : ProgramParser {
...
 
-  protected override DfyParseResult ParseFile(DafnyOptions options, Func<TextReader> getReader,
+  protected override DfyParseResult ParseFile(DafnyOptions options, bool parseAsDooFile, Func<TextReader> getReader,
     Uri uri, CancellationToken cancellationToken) {
     using var reader = getReader();
-    var (newReader, hash) = ComputeHashFromReader(uri, reader, HashAlgorithm.Create("SHA256")!);
+    
+    // Add NUL delimiter to avoid collisions (otherwise hash("A" + "BC") == hash("AB" + "C"))
+    var uriBytes = Encoding.UTF8.GetBytes(uri + "\0");
+    var startingBytes = new[] {Convert.ToByte(parseAsDooFile) }.Concat(uriBytes).ToArray();
```
robin-aws, line 36: Why was this necessary? Shouldn't we have different URIs for a .dfy and a .doo file, even if they contain identical content?

Source/DafnyStandardLibraries/src/DafnyStdLibs/Functions.dfy
```diff
@@ -0,0 +1,32 @@
...
+ *  SPDX-License-Identifier: MIT 
+ *******************************************************************************/
+/**
+A collection of predicates that establish whether commonly known properties on functions, 
+such as injectivity and commutativity, hold.
+These properties can only be applied to functions on value types.
+*/
+module DafnyStdLibs.Function {
+
+  ghost predicate Injective<X(!new), Y>(f: X --> Y)
```
robin-aws, line 12: Just FYI, the variants with `(!new)` are more convenient to use, but I expect we'll also want variants where the domain is explicitly passed in as an `iset`, so that they can be used with reference types too. 

I don't think that is blocking adding these versions now, only we might want to have a plan for how to name the variants intuitively. Perhaps this:

```dafny
ghost predicate InjectiveOn<X, Y>(f: X --> Y, ghost iset<X> domain)
  reads f.reads
  requires forall x <- domain :: f.requires(x)
  {
    forall x1 <- domain, x2 <- domain :: f(x1) == f(x2) ==> x1 == x2
  }
```

Or `InjectiveOver`?

robin-aws (changes requested):
Looks great, just some nits mostly. I'd update the PR comment to mention the doo file parsing bug you're fixing too.

Source/DafnyLanguageServer/Language/CachingParser.cs
keyboardDrummer: True, but at the parser level it's not guaranteed that the URI matches the `parseAsDooFile` bool. I could match on the URI and leave out the `parseAsDooFile` parameter altogether, but I'd prefer not to rely on the URI any more at this point.

Source/DafnyStandardLibraries/examples/WrappersExamples.dfy
```diff
@@ -1,4 +1,4 @@
-module Demo {
+module WrappersExamples {
```
robin-aws, line 1: Conflicts with #4713, and I think I prefer that PR's approach of just using the default module.

Source/DafnyStandardLibraries/src/DafnyStdLibs/Relations.md
```diff
@@ -0,0 +1,37 @@
...
+
+and then need to prove this lemma to use it in a sorting routine:
+<!-- %check-verify %use tmp-intlt.dfy -->
+```dafny
+  import opened DafnyStdLibs.Relations
+  lemma IntLTisStrictTotalOrder()
+    ensures StrictTotalOrdering(IntLT) {}
+```
+
+Fortunately, dafny proves this without aid.
```
robin-aws, line 35:
```suggestion
Fortunately, Dafny proves this without aid.
```

robin-aws (changes requested):
Just a couple of nits, and make sure you update https://github.com/dafny-lang/dafny/tree/master/Source/DafnyStandardLibraries#available-libraries too!

robin-aws (approved):

PR #4711, opened 2023-10-24
Fix client side crash for document symbol because of empty name 

keyboardDrummer:
### Description
- Prevent a client side crash on our malformed document symbol results, that would occur when code has `function` with a name.

### How has this been tested?
- Added a few XUnit tests


robin-aws (approved):
Nice fix. I'd suggest at some point in the future adding a test that simulates entering a program one character at a time, or perhaps one token at a time, since that would create all sorts of invalid states that might produce crashes.

PR #4713, opened 2023-10-25
Add BoundedInts to DafnyStdLibs

keyboardDrummer:
### Description
1. Add `BoundedInts.dfy`, `BoundedInts.md` and `BoundedIntExamples.dfy` from https://github.com/dafny-lang/libraries to https://github.com/dafny-lang/dafny/tree/master/Source/DafnyStandardLibraries
1. For Go, allows customizing the dummy value that is used for an `{:extern}` module, using a new undocumented, only internally used, `{:member <name>, <isType>}` attribute. This is necessary to let the `BoundedIntExamples.dfy` pass, because `FunctionExamples.dfy`, which is part of the same project, does not use the `{:extern}` module, so Go gives an unused import error. In the future I expect that we change our Go compiler so it only emits imports that are used, which makes the `{:member}` attribute obsolete and allows us to remove support for it from our codebase without breaking anyone.

### How has this been tested?
- The examples file serves as a test. It contains an extern method that demonstrates which native types the Dafny compiler picks when using `int32` in Dafny.


Source/DafnyStandardLibraries/examples/WrappersExamples.dfy
```diff
@@ -1,172 +1,168 @@
-module Demo {
+import opened DafnyStdLibs.Wrappers
```
robin-aws, line 1: Good change, but wow did GitHub ever mess up the diff something terrible. :)

Source/DafnyStandardLibraries/examples/BoundedIntExamples.dfy
```diff
@@ -0,0 +1,22 @@
...
+}
+
+method {:test} UseExterns() {
+  var squareOf8 := Externs.NonDefault.SquareNativeInt(8);
+  expect squareOf8 == 64;
+}
+
+module {:extern} Externs {
+  import opened DafnyStdLibs.BoundedInts
+  class {:extern} NonDefault {
```
robin-aws, line 19: Why "NonDefault"?

Source/DafnyStandardLibraries/Makefile
```diff
@@ -23,11 +23,11 @@ update-binary: build-binary
...
-	$(DAFNY) test -t:cs build/stdlibexamples.doo --output:build/stdlibexamples
-	$(DAFNY) test -t:java build/stdlibexamples.doo --output:build/stdlibexamples
-	$(DAFNY) test -t:go build/stdlibexamples.doo --output:build/stdlibexamples
-	$(DAFNY) test -t:py build/stdlibexamples.doo --output:build/stdlibexamples
-	$(DAFNY) test -t:js build/stdlibexamples.doo --output:build/stdlibexamples
+	$(DAFNY) test -t:cs build/stdlibexamples.doo examples/BoundedInts/NonDefault.cs --output:build/stdlibexamples
+	$(DAFNY) test -t:java build/stdlibexamples.doo examples/BoundedInts/Externs/NonDefault.java --output:build/stdlibexamples
+	$(DAFNY) test -t:go build/stdlibexamples.doo examples/BoundedInts/NonDefault.go --output:build/stdlibexamples
+	$(DAFNY) test -t:py build/stdlibexamples.doo examples/BoundedInts/NonDefault.py --output:build/stdlibexamples
+	$(DAFNY) test -t:js build/stdlibexamples.doo examples/BoundedInts/NonDefault.js --output:build/stdlibexamples
```
robin-aws, line 30: Fine for this change but probably worth replacing with a `find` of all files of the right extension in the next one, and before too long with project file support for this. :)

robin-aws (commented):
Looks great and I like the testing approach! Just a couple of thoughts for later and a question.

robin-aws:
Almost forgot, make sure you update https://github.com/dafny-lang/dafny/tree/master/Source/DafnyStandardLibraries#available-libraries too!

Might be worth adding a tiny `Functions.md` for that too.

Source/DafnyStandardLibraries/examples/BoundedIntExamples.dfy
keyboardDrummer: My impression is that for Java an extern should not be in the default class since there are no partial classes.

Source/IntegrationTests/TestFiles/LitTests/LitTest/wishlist/GoModule.dfy
```diff
@@ -2,7 +2,7 @@
 // RUN: %exits-with 3 %dafny /compile:3 /unicodeChar:0 /spillTargetCode:2 "%s" /compileTarget:go 2> "%t"
 // note: putting /compileTarget:go after "%s" overrides user-provided option
 // RUN: %OutputCheck --file-to-check "%t" "%s"
-// CHECK: GoModuleConversions.go:10:3: "net/url" imported and not used
+// CHECK: undefined: GoModuleConversions.ParseURL
```
robin-aws, line 5: Not clear on why this is the error we're getting now, but not a blocker since this is just a negative test.

Source/DafnyCore/Compilers/GoLang/Compiler-go.cs
```diff
@@ -202,11 +197,32 @@ private struct Import {
...
-          importDummyWriter.WriteLine("var _ = os.Args");
+      bool isType;
+      string memberName = null;
+      if (id == "os") {
+        memberName = "Args";
+        isType = false;
+      } else {
+        isType = true;
+        if (import.ExternModule != null) {
+          var attributes = Attributes.Find(import.ExternModule.Attributes, "dummyImportMember");
```
robin-aws, line 208: Definitely glad we're not documenting this officially, but it might be worth adding a note on this attribute to https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/CONTRIBUTING.md, just so anyone that might need to use it is at least aware of how it works.

Source/DafnyCore/Compilers/SinglePassCompiler.cs
```diff
@@ -167,12 +167,14 @@ public abstract class SinglePassCompiler {
     /// call to the instance Main method in the enclosing class.
     /// </summary>
     protected abstract ConcreteSyntaxTree CreateStaticMain(IClassWriter wr, string argsParameterName);
-    protected abstract ConcreteSyntaxTree CreateModule(string moduleName, bool isDefault, bool isExtern, string/*?*/ libraryName, ConcreteSyntaxTree wr);
+    protected abstract ConcreteSyntaxTree CreateModule(string moduleName, bool isDefault, ModuleDefinition externModule,
```
robin-aws, line 170: Definitely feels a little weird to have a parameter that is "the module being created, but only passed if it's an extern". Consider keeping `isExtern` and `theModule` as separate parameters.

Source/DafnyCore/AST/Members/ICodeContext.cs
robin-aws, line 1: Just for my own understanding, are the AST changes unrelated cleanup or necessary somehow?

robin-aws (approved):
Just a few more nits, nothing hard-blocking. Happy with the `{:dummyImportMember}` name. Do update the PR description so the commit message is accurate. :)

Source/DafnyCore/AST/Members/ICodeContext.cs
keyboardDrummer: Unrelated, yes

PR #4715, opened 2023-10-25
Fix spurious warning about vacuous proof

atomb:
There were two utility methods to translate the predicate associated with an assert statement. Both could translate it into an assumption in Boogie under certain circumstances. One avoided recording it as a proof obligation in this case, but the other didn’t. Now it does.

### How has this been tested?

The new test case, `git-issues/PR4715Test.dfy`, generates no warnings. (It did generate a warning prior to this PR.)

### DIFF
--- a/Source/IntegrationTests/TestFiles/LitTests/LitTest/git-issues/PR4715Test.dfy
+++ b/Source/IntegrationTests/TestFiles/LitTests/LitTest/git-issues/PR4715Test.dfy
@@ -1,0 +1,27 @@
+// RUN: %verify --warn-contradictory-assumptions "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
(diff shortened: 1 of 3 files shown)


Source/IntegrationTests/TestFiles/LitTests/LitTest/git-issues/PR4715Test.dfy
```diff
@@ -0,0 +1,27 @@
+// RUN: %verify --warn-contradictory-assumptions "%s" > "%t"
```
keyboardDrummer, line 1: Could you add a [line]comment to indicate what this is testing? Apparently the absence of a warning.

keyboardDrummer (dismissed):

keyboardDrummer (approved):

Issue #4656, opened 2023-10-12
Resolution errors no longer appearing in the gutter


MikaelMayer:
### Dafny version

4.3.0.

### Code to produce this issue

```dafny
method Test() {
  assert x == 1;
}
```

### Command to run and resulting output

```code
Paste in VSCode
```

### What happened?

Latest Dafny:
![image](https://github.com/dafny-lang/dafny/assets/3601079/2956bc5a-d821-49ce-b580-03ea9b113b77)

In Dafny 4.2. the warning icon was displayed correctly on the line.
![image](https://github.com/dafny-lang/dafny/assets/3601079/8a7e7069-130a-4035-b130-e0cc6960c920)

Not having this icon makes it much much harder to identify the location of resolution issues when there were previously verification issues.
Dafny nightly as of today has the same issue. I bisected and tracked back this issue to either one of the following commits 

99635a1f1494b47f4c19b85b870be3930f0cdad9 Let the IDE translate to and resolve Boogie only when
attempting to verify (#4378)
e26e146166a0a1dc490bb1e8624426e897d19f88 Fix backwards compatibility with regards to the Verify
OnMode option (#4412)

### What type of operating system are you experiencing the problem on?

Windows

PR #4732, opened 2023-11-01
Compilation events

keyboardDrummer:
#### Updates
`CompilationManager.CompilationUpdates` is renamed to `Updates`, and its type has been changed from `Observable<Compilation>` to `IObservable<ICompilationEvent>`. This means that the updates are now deltas that describe how the state of compilation has changed, instead of describing the full state of compilation. The full state of the compilation can still be queried through `CompilationManager.Program` and `CompilationManager.Resolution`, so the interface is strictly more powerful now.

Having a stream of deltas available in the interface is important for being able to use it for the CLI, since the CLI wants to immediately print diagnostics as they're found, which `CompilationManager` couldn't handle well without this change.

#### Gutter icons
All processing of gutter related data is now done based on `CompilationManager.CompilationUpdates`, instead of inside `CompilationManager`, which makes it possible to use `CompilationManager` for the CLI without computing gutter icons. This happens to also fix https://github.com/dafny-lang/dafny/issues/4656 for which a test was added.

#### Renames
- Rename `Compilation` to `CompilationInput`, `CompilationManager` to `Compilation`, 
- Rename `CompilationAfterResolution` to `Resolution`, remove its inheritance, move the mutable fields to `CompilationManager` and change it into a record

### How has this been tested?
Most of it is a refactoring, but also added the test `GitIssue4656GutterIconsResolutionError`


Source/DafnyLanguageServer.Test/Diagnostics/PublishedVerificationStatus.cs
```diff
@@ -0,0 +1,24 @@
...
+  public void CombineTest() {
+    Assert.Equal(PublishedVerificationStatus.Running,
+      NotificationPublisher.Combine(PublishedVerificationStatus.Stale, PublishedVerificationStatus.Running));
+    Assert.Equal(PublishedVerificationStatus.Running,
+      NotificationPublisher.Combine(PublishedVerificationStatus.Queued, PublishedVerificationStatus.Running));
+    Assert.Equal(PublishedVerificationStatus.Queued,
+      NotificationPublisher.Combine(PublishedVerificationStatus.Stale, PublishedVerificationStatus.Queued));
+    Assert.Equal(PublishedVerificationStatus.Stale,
+      NotificationPublisher.Combine(PublishedVerificationStatus.Stale, PublishedVerificationStatus.Error));
+    Assert.Equal(PublishedVerificationStatus.Running,
```
MikaelMayer, line 17: How much did you weight this one against `PublishedVerificationStatus.Error`? I understand why it's running, but for example in the gutter icons, errors prevail above everything, including running.

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
```diff
@@ -30,9 +32,14 @@ public class DafnyCounterExampleHandler : ICounterExampleHandler {
       try {
         var projectManager = await projects.GetProjectManager(request.TextDocument);
         if (projectManager != null) {
-          await projectManager.VerifyEverythingAsync(request.TextDocument.Uri.ToUri());
+          var uri = request.TextDocument.Uri.ToUri();
+          await projectManager.VerifyEverythingAsync(uri);
 
-          var state = await projectManager.GetIdeStateAfterVerificationAsync();
+          var state = await projectManager.States.Select(s => s.Value).Where(s =>
+            s.Status == CompilationStatus.ResolutionSucceeded &&
```
MikaelMayer, line 39: I feel like this could be written

```cs
s.EverythingFinished() // needs to be refactored.
```

what do you think?

Source/DafnyLanguageServer/Language/DafnyProgramVerifier.cs
```diff
@@ -28,21 +28,20 @@ public class DafnyProgramVerifier : IProgramVerifier {
     }
 
     public async Task<IReadOnlyList<IImplementationTask>> GetVerificationTasksAsync(ExecutionEngine boogieEngine,
-      CompilationAfterResolution compilation,
+      Resolution resolution,
```
MikaelMayer, line 31: I'm a bit hesitant about the "Resolution" name. Resolution is a process, while class names are usually objects.
What about:
* `ResolutionOutput`
* `ResolutionResult`
* `ResolutionOutcome`
?

Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -0,0 +1,461 @@
...
+
+namespace Microsoft.Dafny.LanguageServer.Workspace;
+
+public delegate Compilation CreateCompilation(
+  DafnyOptions options,
+  ExecutionEngine boogieEngine,
+  CompilationInput compilation);
+
+/// <summary>
+/// The compilation of a single document version.
```
MikaelMayer, line 27: How to differentiate this compilation from the compilation process when translating to target files?
It might help to use a different word. You could use
* `Verification` (more accurate given the end result)
* `VerificationPipeline`
* `CheckingPhases`
* `VerificationProcess`

Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -0,0 +1,461 @@
...
+  private TaskCompletionSource verificationCompleted = new();
+  private readonly DafnyOptions options;
+  public CompilationInput Input { get; }
+  private readonly ExecutionEngine boogieEngine;
+
+  private readonly Subject<ICompilationEvent> updates = new();
+  public IObservable<ICompilationEvent> Updates => updates;
+
+  private Program? programAfterParsing;
+  private Program? compiledProgram;
```
MikaelMayer, line 65: compiledProgram feels like a program that was emitted to the final target source code. What do you mean by compiledProgram here?

Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -0,0 +1,461 @@
...
+      });
+      if (updated) {
+        updates.OnNext(new CanVerifyPartsIdentified(canVerify,
+          implementationsPerVerifiable[canVerify].Values.Select(s => s.Task).ToList()));
+      }
+
+      // When multiple calls to VerifyUnverifiedSymbol are made, the order in which they pass this await matches the call order.
+      await ticket;
+
+      if (!onlyPrepareVerificationForGutterTests) {
```
MikaelMayer, line 262: Switch condition to reduce nesting.

Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -0,0 +1,461 @@
...
+    var result = Formatting.__default.ReindentProgramFromFirstToken(firstToken,
+      IndentationFormatter.ForProgram(program));
+
+    var lastToken = firstToken;
+    while (lastToken.Next != null) {
+      lastToken = lastToken.Next;
+    }
+    // TODO: https://github.com/dafny-lang/dafny/issues/3415
+    return new TextEditContainer(new TextEdit[] {
+      // TODO end position doesn't take into account trailing trivia
```
MikaelMayer, line 445: Move this comment up with the other TODO

Source/DafnyLanguageServer/Workspace/FinishedResolution.cs
```diff
@@ -0,0 +1,85 @@
...
+          l => l.Key,
+          l => MergeResults(l.Select(canVerify => MergeVerifiable(previousState, canVerify)))));
+    var signatureAndCompletionTable = LegacySignatureAndCompletionTable.Resolved ? LegacySignatureAndCompletionTable : previousState.SignatureAndCompletionTable;
+
+    return previousState with {
+      StaticDiagnostics = Diagnostics,
+      Status = status,
+      ResolvedProgram = ResolvedProgram,
+      SymbolTable = SymbolTable
+                    ?? previousState.SymbolTable, // TODO migration seems missing
```
MikaelMayer, line 48: TODO: Can you please create an issue and link the issue to that todo, or fix it?

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
```diff
@@ -66,11 +71,11 @@ public class NotificationPublisher : INotificationPublisher {
     }
 
     private void PublishGlobalProgress(IdeState previousState, IdeState state) {
-      foreach (var uri in state.Compilation.RootUris) {
+      foreach (var uri in state.Input.RootAndProjectUris) {
```
MikaelMayer, line 74: Do you know what the TODO below is about?

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```diff
@@ -352,21 +352,6 @@ Position Position
       }
     }
 
-    // If the verification never starts on this node, it means there is nothing to verify about it.
-    // Returns true if a status was updated
-    public bool SetVerifiedIfPending() {
-      if (StatusCurrent == CurrentStatus.Obsolete) {
```
MikaelMayer, line 358: Where did this code go?

MikaelMayer (commented):
Well done! It seems that the codebase is more maintainable with these refactorings, immutable records, renamings. A few suggestions and questions

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
keyboardDrummer: It's gone. Apparently it wasn't needed, or do you think there's a test missing?

Source/DafnyLanguageServer/Workspace/NotificationPublisher.cs
keyboardDrummer: Yes. Each file is owned by only one project, and files shouldn't publish notifications for files that they don't own. That ownership check is missing here, so one file could be "Parsing..." in one project and "Resolving..." and another, and both projects could publish competing notifications about that.

Source/DafnyLanguageServer/Workspace/Compilation.cs
keyboardDrummer: This class will also be used for the CLI, and also do translation to other languages

Source/DafnyLanguageServer/Workspace/Compilation.cs
keyboardDrummer: Given the context of this class, a Compilation, there is a single AST that is compiled throughout the compilation. This variable contains the AST at its currently most compiled point.

Source/DafnyLanguageServer/Workspace/Compilation.cs
keyboardDrummer: I don't see how it's possible to reduce nesting here by inverting, but I'll extract something to a separate method.

Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
MikaelMayer: It was probably obsolete, I vaguely remember changing the way we set verified on methods without verification condition, and I might have forgotten to delete a method.

Source/DafnyLanguageServer/Workspace/Compilation.cs
MikaelMayer: I was thinking of
```cs
if (onlyPrepareVerificationForGutterTests) {
  DecrementJobs();
  return;
}
<previously indented cod>
DecrementJobs();
```
but refactoring is a good idea as well.

Source/DafnyLanguageServer/Handlers/Custom/DafnyCounterExampleHandler.cs
keyboardDrummer: Not sure what you mean but I extracted some code into a method to make it more clear what it does.

Source/DafnyLanguageServer.Test/Diagnostics/PublishedVerificationStatus.cs
keyboardDrummer: I considered it. I think both have downsides:
- If you show running, then someone who's only watching the status bar to see whether errors are found, will wait longer than they have too. 
- If you show error, then someone might be surprised when new errors appear after it already went from running to error.

Source/DafnyLanguageServer/Workspace/Compilation.cs
keyboardDrummer: I can rename it to `program` but then I feel you might think it's the parsed program, without any changes. I can also rename it to `compilingProgram`, to indicate that's it not finished per se. Or `transformedProgram` or `transformingProgram`.

Renamed to `transformedProgram` for now

Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -241,90 +241,92 @@ CompilationInput input
...
-      verificationTickets.Enqueue(Unit.Default);
+        StatusUpdateHandlerFinally();
+      },
+      StatusUpdateHandlerFinally
+    );
+  }
+
+  private void StatusUpdateHandlerFinally() {
+    try {
+      var remainingJobs = Interlocked.Decrement(ref runningVerificationJobs);
```
MikaelMayer, line 321: A statement not expected to create an exception should generally not be in the try portion of a try catch statement if possible.

Source/DafnyLanguageServer/Workspace/ProjectManager.cs
```diff
@@ -290,18 +290,18 @@ public class ProjectManager : IDisposable {
       compilation.IncrementJobs();
       var resolution = await compilation.Resolution;
 
-      var verifiables = resolution.CanVerifies?.ToList();
-      if (verifiables == null) {
+      var canVerifies = resolution.CanVerifies?.ToList();
```
MikaelMayer, line 293: Seem like a weird name after all. Did you mean "VerifiableDeclarations"?

MikaelMayer (approved):
Thanks for taking my comments into account, it looks better. Two optional small improvements

Source/DafnyLanguageServer/Workspace/Compilation.cs
keyboardDrummer: I have removed this method in a branch I will publish after this PR is merged

Source/DafnyLanguageServer/Workspace/ProjectManager.cs
keyboardDrummer: That is what it means. CanVerifies is meant to the pural of CanVerify. Earlier I named these Verifiable and Verifiables. I'm OK with changing to `VerifiableDeclarations` although it is two words instead of one.

PR #4498, opened 2023-08-31
Removed fuel from encoding of Dafny opaque functions

EkanshdeepGupta:
At present, for encoding opaque functions, Dafny sets it with 0 fuel, and inserts a boolean flag which guards whether the body of the opaque function is available or not. These two mechanisms essentially perform the same function and thus are redundant.

This PR is an attempt to make the dafny encoding of opaque functions leaner by getting rid of the fuel in Dafn'y encoding of opaque functions. 

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/dafny-lang/dafny/blob/master/LICENSE.txt).</small>

PR #4180, opened 2023-06-14
Fix 4176 reveal crash

EkanshdeepGupta:
Fixes #4176 by making generated reveal lemmas be static.

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/dafny-lang/dafny/blob/master/LICENSE.txt).</small>

PR #4733, opened 2023-11-01
Allow static reveal of instance function in new resolver

RustanLeino:
PR https://github.com/dafny-lang/dafny/pull/4180 introduced the ability to `reveal` an instance function in a static context. In the recent PR https://github.com/dafny-lang/dafny/pull/4498, this has been implemented to put the instance function itself, without a receiver argument, into an attribute. The recent PR then special-cases the resolution of that attribute. I would call that a hack.

This PR mimics that hack for the new resolver. That makes `git-issue-4315.dfy` (see Issue https://github.com/dafny-lang/dafny/issues/4315) pass with both resolvers.

The hack introduced in this PR and in https://github.com/dafny-lang/dafny/pull/4498 is not a great idea, because a use of `this` is placed into the AST in a context where there is no `this`. If (rather, when) this gets fixed, it should be fixed in both resolvers.

### How has this been tested?

Test `git-issues/git-issue-4315.dfy` has been converted to run with both the old and new resolver.


MikaelMayer (approved):
It's unfortunate indeed, but it's the way the translator worked that made this change the most straightforward. The translator does not care about the "this" part, only the member and function, so "this.f" serves only as a pointer to the function. I'm not sure how we could store a pointer to a non-static function in a resolved expression otherwise.

Good job for making the new resolver up to date with that change.

If appropriate, can you please add an entry like doc/dev/4733.fix containing something like "The new type checker now also supports static reveals for instance functions" ?

MikaelMayer (approved):

PR #4735, opened 2023-11-01
feat: /standardLibraries flag in legacy CLI

RustanLeino:
This PR adds a legacy-CLI option equivalent for `--standard-libraries`. Using this legacy flag, the PR also updates the `lit` harness.


robin-aws (approved):
Change LGTM. I do think as long as we're bothering to maintain `lit` compatibility for some tests (and there are definitely some that don't work on it), we should either add at least one test to CI to confirm that, or (my preference) create the `xlit` replacement: https://github.com/dafny-lang/dafny/issues/2033

robin-aws (approved):

PR #4738, opened 2023-11-03
Enable integration tests to run with the Rust backend in CI

shadaj:
Enable integration tests to run with the Rust backend in CI


shadaj:
Do not update the branch for now, https://github.com/dafny-lang/dafny/pull/4658 breaks the Rust backend pretty badly becaus the `nat` newtype is never emitted.

shadaj:
Okay, maybe it's not https://github.com/dafny-lang/dafny/pull/4658; but something is definitely still wrong with the `_System` module never being emitted.

shadaj:
Okay, it is definitely #4658. Tests run fine locally but in CI this PR is merged with main so tests fail due to missing `_System`.

shadaj:
@jtristan ready to go assuming that CI passes!

jtristan (approved):

PR #4739, opened 2023-11-04
Catch copy error

keyboardDrummer:
### Description
Recently there was a change in MultibackendTest which causes `DafnyRuntime.dll` to be passed to Dafny as an input argument for C# runs. This triggers a potential crash in the C# back-end, which occurs when it fails to copy any input `.dlls` for any reason: https://github.com/dafny-lang/dafny/actions/runs/6755446930/job/18363885807

This PR changes the C# back-end so it simply swallows an IOException if copying the dependency fails. This isn't great because it leads to another crash if the copy fails for any reason other than that the correct .dll is already in that spot, and this may even be a security problem, but it does solve this immediate problem.

Some alternatives:
- Create a temporary directory to copy all the dlls to and from which to run
- I tried using the `dotnet` CLI option `--additionalprobingpath <PATH>` so we wouldn't have to copy the dependency, but so far had no luck in using it correctly. https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet#options-for-running-an-application

### How has this been tested?
Many of our tests execute this code


robin-aws:
How about changing `MultiBackendTest` to create a new temporary directory for each test and specifying `--out:...`, so that parallel tests don't interfere with each other? That seems like a Good Idea in general to avoid other similar issues.

robin-aws:
> Create a temporary directory to copy all the dlls to and from which to run

PR #4740, opened 2023-11-06
Follow-up changes related to BoundedInts

keyboardDrummer:
### Description
Contains some refactorings suggested by comments in https://github.com/dafny-lang/dafny/pull/4713

### How has this been tested?
No additional tests needed


Source/DafnyRuntime/DafnyRuntimeSystemModule.cs
robin-aws, line 1: Looks like you formatted this for whitespace but it should be excluded like the other Dafny-generated files :)

robin-aws (changes requested):
Thanks for following up, this is definitely better. Just a small nit.

PR #4744, opened 2023-11-06
Enable traits in the Rust backend

shadaj:
Enable traits in the Rust backend


shadaj:
Looks like this will need a bit more work (we don't properly handle variables with a trait type), so closing for now.

PR #4745, opened 2023-11-06
Syncing RDE Work

ssomayyajula:
### Description

Pretty printing RDE Dafny backend + Dafny Abstract Machine.

### How has this been tested?

Pretty printing tested on "Hello, World".


Source/DafnyCore/Compilers/ResolvedDesugaredExecutableDafny/Dafny-compiler-fdafny.dfy
```diff
@@ -2,15 +2,209 @@ module {:extern "ResolvedDesugaredExecutableDafnyPlugin"} ResolvedDesugaredExecu
   import opened DAST
   import PrettyPrinter
   import UnsupportedFeature
+  import DAM
 
   class COMP {
```
RustanLeino, line 7: Class `COMP` only has static members. Remove the class altogether and instead declare the members as module-level members.

Source/DafnyCore/Compilers/ResolvedDesugaredExecutableDafny/Dafny-compiler-fdafny.dfy
```diff
@@ -2,15 +2,209 @@ module {:extern "ResolvedDesugaredExecutableDafnyPlugin"} ResolvedDesugaredExecu
...
+        var next := EmitBlock(next);
+        st := DAM.While(guard, body, next);
+      
+      // Breaks, returns, and halts drop the rest of the block
+      case Break(lab) => {
+        match lab
+        case Some(lab) =>
+          st := DAM.Throw(DAM.Var(lab), DAM.Skip());
+        case None => UnsupportedFeature.Throw(); st := DAM.Skip();
+      }
```
RustanLeino, line 166: Instead of putting `{ ... }` around the entire `match lab ...` statement, put the `{ ... }` around the list of `case`s, like this:

``` dafny
match lab {
  case Some(lab) =>
    ...
  case None => ...
}
```

Source/DafnyCore/Compilers/Dafny/DAM.dfy
```diff
@@ -0,0 +1,654 @@
...
+  }
+
+  // CBPV distinguishes between expressions and statements of positive and negative type, respectively
+  module Syntax {
+    import opened Utils
+
+    type Field = string
+
+    datatype Pos =
+      Unit
```
RustanLeino, line 56: Add a prefix `|` to make all the alternatives look the same.

Source/DafnyCore/Compilers/Dafny/DAM.dfy
```diff
@@ -0,0 +1,654 @@
...
+          if CheckClosedExpr(s, arg, dom) then
+            SynthStack(s, cod, stack)
+          else
+            None
+        case _ => None
+      )
+    }
+
+    // S |- stack <= start- <= end-
+    predicate CheckStack(s: StoreTyping, start: Neg, stack: Stack, end: Neg) {
```
RustanLeino, line 353: The name `CheckStack` sounds operational, using the verb "check". Wouldn't it be more appropriate to think of a predicate is using an adjective, like `ValidStack`?

For that matter, it is also possible to rename the predicate just `Valid` and declare it as a member of the `StoreTyping` type.

RustanLeino (commented):
Here are a few style comments.

Source/DafnyCore/Compilers/Dafny/DAM.dfy
ssomayyajula: +1 for grouping checking/synthesis functions into each type definition

The terminology Check/Synth is a reference to the two modes of bidirectional typing (checking against a type vs. synthesis of a type).

Source/DafnyCore/Compilers/ResolvedDesugaredExecutableDafny/Dafny-compiler-fdafny.dfy
ssomayyajula: I think this should be fine as long as `Compiler-dafny.cs` does not depend on the fact that `COMP` is a class by way of its translation to C# in `GeneratedFromDafny.cs`.

PR #4752, opened 2023-11-08
Dynamic array

keyboardDrummer:
### Description
Add a library for dynamic arrays

### How has this been tested?
Added an example file that calls all public methods of DynamicArray.dfy


Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+      return true;
+    }
+
+    /**
+    Pop an element from the array in constant time
+     */
+    method PopFast()
+      requires Valid?()
+      requires size > 0
+      modifies `size, `items
```
RustanLeino, line 119: Use `modifies Repr` and some appropriate postconditions about the exported fields.

Same comment for the public methods below.

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+      var oldData, oldCapacity := data, capacity;
+      data, capacity := new A[newCapacity](_ => defaultValue), newCapacity;
+      CopyFrom(oldData, oldCapacity);
+      Repr := {this, data};
+    }
+
+    /**
+    For internal use
+     */
+    function DefaultNewCapacity(capacity: uint32) : uint32 {
```
RustanLeino, line 192: Style: No space before the `:`

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+      then 2 * capacity
+      else MAX_CAPACITY
+    }
+
+    /**
+    For internal use
+     */
+    method ReallocDefault() returns (s: bool)
+      requires Valid?()
+      modifies `capacity, `data, `Repr, data
```
RustanLeino, line 203: Note, here it's fine to mention these fields, since the method is not exported.

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+    Push an element onto the array in constant time if there is available capacity, 
+    and otherwise in time linear to the current size of the array.
+
+    Returns false only if it was not at all possible to push the element onto the array.
+     */
+    method Push(element: A) returns (s: bool)
+      requires Valid?()
+      modifies this, data
+      ensures Valid?()
+      ensures s == (old(size) < MAX_CAPACITY)
```
RustanLeino, line 154: If you change the `==` to a `<==>`, you can omit the parentheses (like you do on L205).

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+     */
+    method CopyFrom(newData: array<A>, count: uint32)
+      requires count as int <= newData.Length
+      requires count <= capacity
+      requires data.Length == capacity as int
+      modifies data
+      ensures data[..count] == newData[..count]
+      ensures data[count..] == old(data[count..])
+    {
+      for idx: uint32 := 0 to count
```
RustanLeino, line 234: No need to include `: uint32`, because it will be inferred.

RustanLeino: Actually, you're better off using a `forall` statement to do this copy:

``` dafny
forall i | 0 <= i < count {
  data[i] := newData[i];
}
```

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+      ensures Valid?()
+      ensures !s <==> old(capacity) == MAX_CAPACITY
+      ensures !s ==>
+                && unchanged(this)
+                && unchanged(data)
+      ensures s ==>
+                && fresh(data)
+                && old(capacity) < MAX_CAPACITY
+                && capacity == old(if capacity < MAX_CAPACITY_BEFORE_DOUBLING
+                                   then 2 * capacity else MAX_CAPACITY)
```
RustanLeino, line 213: For the RHS of this equality, use `old(DefaultNewCapacity(capacity))`.

RustanLeino (changes requested):

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+  but additionally allows growing and shrinking the size of the array. The cost of shrinking the array is constant time, 
+  while the cost of growing it by a single element is either constant time or linear in the size of the array. 
+  When growing the array multiple times, the amortized cost of growing the array by one element is constant time.
+
+  The downsize of using a DynamicArray is that it occupies up to twice the amount of memory as its size, 
+  and when elements are removed, its occupied memory does not decrease. 
+  */
+  class DynamicArray<A> {
+    ghost var items : seq<A>
+    ghost var Repr : set<object>
```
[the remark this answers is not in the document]
robin-aws: s/comma/space/? :)

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+
+    const defaultValue: A
+    var size: uint32
+    var capacity: uint32
+    var data: array<A>
+
+    const MAX_CAPACITY: uint32 := (TWO_TO_THE_32 - 1) as uint32
+    const MAX_CAPACITY_BEFORE_DOUBLING: uint32 := MAX_CAPACITY / 2
```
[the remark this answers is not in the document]
robin-aws: I agree with not modeling the max capacity and just assuming it always succeeds - since the runtime error when failing to allocate the storage will be unrecoverable in Dafny source code, it's not a soundness risk. If anything making client code transitively handle the failure makes it more likely that a program will recover from running out of memory and limp along when it shouldn't.

It's fine for now if the JSON library this came from continues this design internally when we import it over, though, especially since it already has to model runtime failure for various other reasons anyway: https://github.com/dafny-lang/libraries/blob/ae8708c091d32383235d5d8c15c08cff05613bbc/src/JSON/Errors.dfy

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+      capacity := initial_capacity;
+      data := new A[initial_capacity](_ => defaultValue);
+      Repr := {this, data};
+    }
+
+    /**
+    Retrieve the element at index 
+     */
+    function At(index: uint32) : (element: A)
+      reads this, this.data
```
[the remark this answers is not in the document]
keyboardDrummer: I see! Thanks

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,244 @@
...
+    {
+      && Repr == {this, data}
+      && capacity != 0
+      && data.Length == capacity as int
+      && size <= capacity
+      && size as int == |items|
+      && items == data[..size]
+    }
+
+    constructor(defaultValue: A, initial_capacity: uint32 := 8)
```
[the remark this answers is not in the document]
keyboardDrummer: > You can avoid this by using a ghost datatype constructor, see TEST/comp/Uninitialized.dfy

I don't understand how this trick works. Can I initialize a non-ghost array variable with ghost values? I tried `new MaybeInitialized<A>[initialCapacity]` but I get the message `unless an initializer is provided for the array elements, a new array of 'MaybeInitialized<A>' must have empty size`

If I use `new MaybeInitialized<A>[initialCapacity](_ => Uninitialized)`, then I get `ghost constructor is allowed only in specification contexts`

The test you linked worked because it does not use `--enforce-determinism`, while this file does use that.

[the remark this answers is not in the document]
RustanLeino: The trick uses the first of these, that is:

``` dafny
new MaybeInitialized<A>[initialCapacity]
```

without any further initializer. This means that the compiler is free to initialize the new array elements any way it wants, as long as the values it uses are of the right type. In fact, the compiler is also allowed to pick the `Uninitialized` value, which you are not allowed to mention yourself in non-ghost code.

However, if the `new` allocation does not give an initializer, then the statement is nondeterministic.

Why does this example use `--enforce-determinism`?

Source/DafnyStandardLibraries/examples/DynamicArrayExamples.dfy
```diff
@@ -0,0 +1,52 @@
+module DynamicArrayExamples {
+  import opened DafnyStdLibs.DynamicArray
+  import opened DafnyStdLibs.BoundedInts
+
+  method {:test} InitialCapacity() {
+    var arr := new DynamicArray<int>(0, 101);
+    for i: int := 0 to 100
```
RustanLeino, line 7: `: int` is not needed. It will be inferred.

RustanLeino: Btw, this test method only pushes 100 elements, so L6 could pass `100` (instead of `101`) to the constructor. In fact, I would argue that using `100` is a better test to make sure the specifications don't have an off-by-1 error somewhere.

Source/DafnyStandardLibraries/examples/DynamicArrayExamples.dfy
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  method {:test} Ensure() {
+    var arr := new DynamicArray<int>(0);
+    assert arr.size == 0;
+    arr.Ensure(101);
+    assert arr.capacity >= 101;
+    assert arr.size == 0;
+    for i: int := 0 to 100
```
RustanLeino, line 23: Same comments for this test. The `: int` in the `for` loop is unnecessary, and you could (if you want to) change the `101` to `100`.

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
```diff
@@ -0,0 +1,218 @@
...
+      items := items[..|items| - 1];
+    }
+
+    /**
+    Push an element onto the array in constant time
+     */
+    method PushFast(element: A)
+      requires Valid?()
+      requires size < capacity
+      modifies Repr, `size, `items
```
RustanLeino, line 142: Since `this` is already part of `Repr`, you don't need to list `` `size, `items``

RustanLeino (changes requested):
My suggestion is still to use a `MaybeInitialized` type with a `ghost` constructor. You pointed out that this does not work with `--enforce-determinism`, so my suggestion is to remove the `--enforce-determinism` option.

If it really isn't possible to remove that option, then--in order to remove the unexpected `defaultValue` parameter to the constructor--I suggest changing the design to start off with a capacity of 0. When there's a `Push(x)` operation and the underlying array needs to grow, then use the `x` to fill all new slots of the array (which you can do if you provide an initializer for the `new`-array allocation).

keyboardDrummer:
> If it really isn't possible to remove that option, then--in order to remove the unexpected `defaultValue` parameter to the constructor--I suggest changing the design to start off with a capacity of 0. When there's a `Push(x)` operation and the underlying array needs to grow, then use the `x` to fill all new slots of the array (which you can do if you provide an initializer for the `new`-array allocation).

That's only possible if we also remove the Ensure operation, but I think it's important to have it. Also I think it's important to be able to construct the array with a non-zero capacity.

The reason `--enforce-determinism` is on is to make the library compatible both with consumers that have it off, and that have it on (which requires the library to have it on as well).

RustanLeino:
I will think some more, but I like your idea of trying to special rule for arrays. How about the following: If the element type of the array is a datatype that contains a ghost constructor, then the nondeterminism can never be detected in a running program. Therefore, we could let the `--enforce-determinism` flag _not_ complain about that situation.

keyboardDrummer:
> I will think some more, but I like your idea of trying to special rule for arrays. How about the following: If the element type of the array is a datatype that contains a ghost constructor, then the nondeterminism can never be detected in a running program. Therefore, we could let the `--enforce-determinism` flag _not_ complain about that situation.

For now I've updated the API of this library, so such changes aren't needed.

Source/DafnyStandardLibraries/src/DafnyStdLibs/DynamicArray.dfy
RustanLeino: Same for method `Ensure`.

RustanLeino (approved):

PR #4753, opened 2023-11-08
Add the built-in format command instead of the discontinued dotnet-format tool

keyboardDrummer:
### Description
In CI and in the top-level Makefile, use the `dotnet format` command that comes with `dotnet`, instead of the discontinued dotnet-format tool.

### How has this been tested?
- Manually expected the formatting changes it makes. They look good.
- I also run the `style` and `analyzers` options but neither made any changes, so I've left them out of CI for now.
- Updated CI so it runs the newer format command
- Tried out `pre-commit` with the new command.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,5 +4,5 @@ repos:
     -   id: dotnet-format
         name: dotnet-format
         language: system
-        entry: dotnet tool run dotnet-format --folder -w --include
+        entry: dotnet format whitespace Source/Dafny.sln --include
         types_or: ["c#"]
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ z3-ubuntu:
 	chmod +x ${DIR}/Binaries/z3/bin/z3-*
 
 format:
-	dotnet tool run dotnet-format -w -s error Source/Dafny.sln --exclude DafnyCore/Scanner.cs --exclude DafnyCore/Parser.cs --exclude DafnyCore/GeneratedFromDafny.cs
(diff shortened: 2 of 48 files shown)


Makefile
```diff
@@ -62,7 +62,7 @@ z3-ubuntu:
 	chmod +x ${DIR}/Binaries/z3/bin/z3-*
 
 format:
-	dotnet tool run dotnet-format -w -s error Source/Dafny.sln --exclude DafnyCore/Scanner.cs --exclude DafnyCore/Parser.cs --exclude DafnyCore/GeneratedFromDafny.cs
+	dotnet format whitespace Source/Dafny.sln --exclude Source/DafnyCore/Scanner.cs --exclude Source/DafnyCore/Parser.cs --exclude Source/DafnyCore/GeneratedFromDafny.cs --exclude Source/DafnyRuntime/DafnyRuntimeSystemModule.cs
```
robin-aws, line 65: Thanks, missed this myself when adding this file :)

.pre-commit-config.yaml
```diff
@@ -4,5 +4,5 @@ repos:
     -   id: dotnet-format
         name: dotnet-format
         language: system
-        entry: dotnet tool run dotnet-format --folder -w --include
+        entry: dotnet format whitespace Source/Dafny.sln --include
```
robin-aws, line 7: Seems like this must not be used much since it's out-of-date w.r.t. the excluded files. Good idea to at least update it but we should consider dropping it in the future if it's not useful.

You said in the PR title you tried it and it worked? Is it possible the `--include` is just including nothing and it's a no-op? :)

robin-aws (dismissed):
Good call thanks. Curious about the pre-commit question but not blocking.

.pre-commit-config.yaml
keyboardDrummer: Using pre-commit on this file, it will format only the files that are currently staged. Existing files that require formatting but are not staged, or files that are added or changed but not staged, will not formatted.

stefan-aws (approved):
Previously approved. Only a small change since.

PR #4755, opened 2023-11-08
feat: Execution branch coverage report

### LINKED ISSUE #4671: Feature request: runtime branch coverage
This is partially implemented already, as we have a `/coverage` option implemented for most backends: https://github.com/dafny-lang/dafny/blob/master/Test/comp/BranchCoverage.dfy.

This option is really more of a branch coverage plugin interface, though: users have to provide their own coverage event callback handler, such as https://github.com/dafny-lang/dafny/blob/master/Test/comp/BranchCoverage2.cs. The feature that users will actually want is a simple switch to emit a coverage report, and the ability to set a minimum threshold for the coverage percentage.

I suggest providing a `--coverage-report` option for `dafny run` and `dafny test`, just as we already have for `dafny generate-tests` (which shows the expected runtime coverage). Since the runtime coverage emitted by every backend should be the same, we'll get a lot of benefit from just supporting the C# backend for now. We can hook up the existing `/coverage` instrumentation to the existing `CoverageReport` C# class.

robin-aws:
### Description

Implements an execution coverage report in the same format as the existing verification coverage and expected generated test coverage reports. Mainly this just meant using the existing low-level instrumentation provided by the old CLI `/coverage` option to write coverage data to a temporary file during execution, and then parsing this data and populating a `CoverageReport` object. I've only implemented this for C# for now, since the execution coverage SHOULD generally be identical across languages (the instrumentation is emitted from the common SinglePassCompiler) but it wouldn't take much work to hook it up for the other backends.

Also adds this to the `DafnyStandardLibraries` project, and augments the existing coverage report generation to handle `.doo` files and Uris somewhat better.

Also renames the existing verification option `--coverage-report` to `--verification-coverage-report`, since this option is almost always in scope (since most commands run verification), and for `dafny test` and `dafny run` "coverage report" more naturally means execution coverage instead. The former hasn't been released yet so this won't break anyone. :)

Resolves #4671

### How has this been tested?
`comp/CoverageReport.dfy`, similar to the existing `logger/CoverageReport.dfy`


robin-aws:
I've fixed the refman failure locally but don't want to push and trigger another CI run until I've also addressed any PR feedback

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
```diff
@@ -3435,5 +3443,30 @@ protected class ClassWriter : IClassWriter {
       TrStmt(recoveryBody, catchBlock);
     }
 
+    protected void EmitCoverageReportInstrumentation(Program program, ConcreteSyntaxTree wr) {
+      wr.WriteLine(@"
+namespace DafnyProfiling {
+  public class CodeCoverage {
```
atomb, line 3449: Could we put this in the runtime?

Source/DafnyCore/Compilers/CoverageInstrumenter.cs
```diff
@@ -82,4 +93,20 @@ public class CoverageInstrumenter {
...
   }
+
+  public void PopulateCoverageReport(CoverageReport coverageReport) {
+    var coverageReportDir = compiler.Options?.Get(CommonOptionBag.ExecutionCoverageReport);
+    if (coverageReportDir != null) {
+      var tallies = File.ReadLines(talliesFilePath).Select(int.Parse).ToArray();
+      foreach (var ((token, _), tally) in legend.Zip(tallies)) {
+        var label = tally == 0 ? CoverageLabel.NotCovered : CoverageLabel.FullyCovered;
+        // For now we only identify branches at the line granularity,
+        // which matches what `dafny generate-tests ... --coverage-report` does as well.
```
atomb, line 104: This is nice because then merging actual and expected coverage would work well. The `merge-coverage-reports` command could perhaps use a `--diff` flag that would produce a report indicating where two coverage reports differed.

Source/DafnyCore/Options/CommonOptionBag.cs
```diff
@@ -214,14 +214,18 @@ public enum GeneralTraitsOptions {
...
-    "Emit verification coverage report  to a given directory, in the same format as a test coverage report.") {
+  public static readonly Option<string> VerificationCoverageReport = new("--verification-coverage-report",
+    "Emit verification coverage report to a given directory, in the same format as a test coverage report.") {
     ArgumentHelpName = "directory"
   };
   public static readonly Option<bool> NoTimeStampForCoverageReport = new("--no-timestamp-for-coverage-report",
     "Write coverage report directly to the specified folder instead of creating a timestamped subdirectory.") {
     IsHidden = true
   };
+  public static readonly Option<string> ExecutionCoverageReport = new("--coverage-report",
```
atomb, line 225: You could call this `--execution-coverage-report`, since that term shows up in the variable name and the description text. :)

Source/DafnyStandardLibraries/examples/WrappersExamples.dfy
```diff
@@ -32,6 +32,20 @@ method {:test} TestMyMap() {
   expect greeting == "Hello\nDafny\n";
 }
 
+method {:test} TestGetGreetingSuccess() {
```
atomb, line 35: The coverage reporting is already bearing dividends. :)

docs/dev/news/4625.feat
```diff
@@ -1 +1 @@
-The new `--coverage-report` flag to `dafny verify` creates an HTML report highlighting which portions of the program were and were not necessary for verification. The format is the same as for `dafny generate-tests --coverage-report` and files from the two commands can be merged.
+The new `--verification-coverage-report` flag to `dafny verify` creates an HTML report highlighting which portions of the program were and were not necessary for verification. The format is the same as for `dafny generate-tests --coverage-report` and files from the two commands can be merged.
```
atomb, line 1: Thanks for being thorough about where you change the name of this flag!

atomb (approved):
This was remarkably simple! It changed a somewhat large number of lines, but all in very simple ways.

Source/DafnyCore/Compilers/CSharp/Compiler-Csharp.cs
robin-aws: Not without breaking the existing `/coverage` option, since the idea is that you could provide your own implementation of `DafnyProfilng.CodeCoverage` as per the `BranchCoverage.dfy` test.

Source/DafnyCore/Compilers/CoverageInstrumenter.cs
robin-aws: My thoughts exactly, and although I do suspect that the actual and expected coverage don't 100% agree on coverable things yet, I don't think it will take much to align them in the future.

Source/DafnyCore/Options/CommonOptionBag.cs
robin-aws: I did consider that, but then it would feel weird to not further qualify `dafny generate-tests --coverage-report` as well, perhaps as `--expected-execution-coverage-report`. Given the `generate-tests` one is already released I'm more inclined to leave both with the simpler name.

atomb (approved):

PR #4756, opened 2023-11-09
feat: import Collections library

alex-chew:
This PR imports the Collections libraries from `dafny-lang/libraries`, as a new `DafnyStdLibs` library.

Note that the `LittleEndianNat` and `LittleEndianNatConversions` libraries are not imported, as there is still work required to enable different verification flags in the standard libraries (particularly `--disable-nonlinear-arithmetic` as desired for these libraries).

### How has this been tested?


robin-aws (commented):
I think this is good to merge once the dependency Arithmetic is in, especially to unblock other libraries that depend on it. There are a couple of things to follow up on afterwards though:

* Obviously needs examples. Using #4755 to measure coverage will point out that they are missing so I'm not worried about forgetting about this.
* Some functions are ripe for performance improvements - probably quite a few need `{:tailrecursion}` to avoid stack overflows on large inputs, and perhaps some could use `by method` optimizations. I wouldn't block on that in general, but we did do a little bit of this to ensure this JSON stress test passes: https://github.com/dafny-lang/libraries/blob/master/src/JSON/Tests.dfy#L134-L136 If that regresses it means we've somehow lost an optimized version in transition.

Source/DafnyStandardLibraries/src/DafnyStdLibs/Collections/Seqs.dfy
```diff
@@ -0,0 +1,952 @@
...
+ *******************************************************************************/
+
+/**
+ This module defines useful properties and functions relating to the built-in `seq` type.
+ */
+module DafnyStdLibs.Collections.Seq {
+
+  import opened Wrappers
+  import opened Relations
+  import Math
```
robin-aws, line 21: Should be using `DafnyStdLibs.Arithmetic` instead.

Source/DafnyStandardLibraries/src/DafnyStdLibs/Math.dfy
robin-aws, line 1: Shouldn't be included since we have `DafnyStdLibs.Arithmetic` now

robin-aws (changes requested):

robin-aws (approved):
Changed my mind on my previous review - didn't realize that `Math` had definitions that aren't already in `Arithmetic`. The distinction is not common in standard libraries, but I think Dafny's allowed to be more precise than usual. :)

PR #4761, opened 2023-11-09
Import Arithmetic library

ssomayyajula:
### Description

Completed:
- Moves the NonlinearArithmetic `library` from the libraries repo into `Source/DafnyStandardLibraries`
- Removes `RUN` and `include` directives except for those in `Arithmetic/Internals/DivInternals.dfy` and `Mul.dfy` referencing `Arithmetic/Internals/*Nonlinear.dfy` (see those files for more details)
- Fix: proof of `LemmaPowPositive` in `Power.dfy`
- Style: Replaces `forall ensures` with `assert by` everywhere 
- Adds Robin’s build process for this library, edits it to account for `Internals/*Nonlinear.dfy` needing non-linear arithmetic (see the `.toml` files for more details)
- `LemmaDivAuto()` in `DivMod.dfy` brought under resource limit
- Standards: triggers added in `DivMod.dfy`

Holding off:
- Standards: `warn-redundant-assumptions` and `warn-contradictory-assumptions` actually breaks a one/two proofs each in `DivMod.dfy`, `Power.dfy`, and `ModInternals.dfy` (the latter breaks a proof in an straightforward-to-fix way, but the former may be flagging false positives with the way proofs are currently written)
- Style: expand short names everywhere

### How has this been tested?

Adding to `examples/ArithmeticExamples.dfy`.


Source/DafnyDriver/Commands/DafnyCli.cs
```diff
@@ -32,6 +32,7 @@ public static class DafnyCli {
   private static readonly RootCommand RootCommand = new("The Dafny CLI enables working with Dafny, a verification-aware programming language. Use 'dafny -?' to see help for the previous CLI format.");
 
   private static readonly Uri StandardLibrariesDooUri = new("dllresource://DafnyPipeline/DafnyStandardLibraries.doo");
+  private static readonly Uri StandardLibrariesArithmeticDooUri = new("dllresource://DafnyPipeline/DafnyStandardLibraries-arithmetic.doo");
```
robin-aws, line 35: I know this was from my own commit :) but I think we should use a more accurate suffix here, since this doo file only contains part of the Arithmetic library. Perhaps `-nla-disabled.doo`?

Source/DafnyStandardLibraries/src/DafnyStdLibs/Arithmetic/README.md
```diff
@@ -0,0 +1,11 @@
+# Overview
+
+Non-linear math is generally undecidable; hence Dafny (and Z3) rely on heuristics to prove such properties.
+While wonderful when they work, these heuristics can lead to unstable proofs.  Hence, in many projects,
```
robin-aws, line 4:
```suggestion
While wonderful when they work, these heuristics can lead to brittle proofs.  Hence, in many projects,
```

Source/DafnyStandardLibraries/src/dfyconfig-arithmetic.toml
```diff
@@ -0,0 +1,29 @@
...
+# Options that help support more values of consuming context options
+# (mostly turning on any more restrictive verification)
+track-print-effects = true
+enforce-determinism = true
+reads-clauses-on-methods = true
+
+# Options important for sustainable development
+# of the libraries themselves.
+resource-limit = 1000
+#warn-as-errors = true
```
robin-aws, line 27: I would keep this one on even if it's reasonable to remove the assumptions ones, since otherwise end users may see other warnings when they flip on `--standard-libraries`

Source/DafnyStandardLibraries/src/dfyconfig.toml
```diff
@@ -1,4 +1,7 @@
 
+# *Nonlinear.dfy in Arithmetic/Internals/ are subject to the options here
+excludes = ["**/Arithmetic/*.dfy", "**/Arithmetic/Internals/*Internals.dfy"]
```
robin-aws, line 3:
```suggestion
excludes = ["**/Arithmetic/*.dfy", "**/Arithmetic/**/*Internals.dfy"]
```

Just to keep the two project files 100% in sync. Not THAT critical because if we double-include or miss files, we'll get errors when testing the examples, but better to avoid the possibility. :)

Source/DafnyStandardLibraries/src/DafnyStdLibs/Arithmetic/DivMod.dfy
```diff
@@ -0,0 +1,1629 @@
...
+ *  
+ *  Modifications and Extensions: Copyright by the contributors to the Dafny Project
+ *  SPDX-License-Identifier: MIT 
+ *******************************************************************************/
+
+/* Every lemma comes in 2 forms: 'LemmaProperty' and 'LemmaPropertyAuto'. The
+former takes arguments and may be more stable and less reliant on Z3
+heuristics. The latter includes automation and its use requires less effort */
+
+module {:options "-functionSyntax:4"} DivMod {
```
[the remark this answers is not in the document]
ssomayyajula: Nitpicky question: do we prefer verbs or nouns, i.e., `DividesModulo` vs. `DivisionModulus`?

Source/DafnyDriver/Commands/DafnyCli.cs
ssomayyajula: Agreed; would hold off until PR is merged in the interest of time if that's OK with you.

robin-aws: Oh that's just fine, since this isn't user-facing

Source/DafnyStandardLibraries/src/DafnyStdLibs/Arithmetic/DivMod.dfy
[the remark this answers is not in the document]
robin-aws: I think that's fine, the nesting is the much more critical point anyway.

Source/DafnyStandardLibraries/examples/dfyconfig.toml
```diff
@@ -3,6 +3,7 @@
 [options]
 standard-libraries = true
 resource-limit = 1000
-warn-as-errors = true
+# TODO: necessary because Algorithms has warnings in it
+warn-as-errors = false
```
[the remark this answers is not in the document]
ssomayyajula: Should be back in the latest version!

Source/DafnyStandardLibraries/src/dfyconfig-arithmetic.toml
ssomayyajula: Should be back in the latest version!

Source/DafnyStandardLibraries/examples/dfyconfig.toml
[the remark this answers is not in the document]
robin-aws: Looks like it's still there? :)

robin-aws (approved):
Awesome work, this was one of the tougher ones!

PR #4767, opened 2023-11-10
fix: Escape tallies temporary file path properly

robin-aws:
### Description
#4755 broke the nightly build because I neglected to escape a string literal properly. Works just fine as long as your directory separator is `/`, but then Windows had to go and be all different. :)

I attempted to use `EmitStringLiteral` to escape correctly, but it turns out that this still tried to translate Dafny `\U` unicode escape sequences in a windows path like `C:\Users\...`. I suspect there might be a bug lurking there too but didn't want to block fixing the nightly build on it.

### How has this been tested?
Existing tests with `run-deep-tests`.


atomb (approved):
As long as the Windows tests pass, this looks good.

PR #4770, opened 2023-11-10
Standard lib support in `dafny server`

keyboardDrummer:
### Description
- Add support for the Dafny standard library to `dafny server`
- Add support for `library = [..]` in `dfyconfig.toml` to `dafny server`
- Refactor error handling for errors in dfyconfig.toml, so it is consistent between `dafny server` and other commands.
- Ensure `dafny server` only publishes notifications for Uris owned by a particular project. A project owns a Uri if it is the closest project to that Uri and it includes it.
- Moves code in `new Dafny` to `DafnyFile.CreateAndValidate`, so it's more clear that it contains logic.
- Refactor the logic in `DafnyFile` so it reports errors using an `ErrorReporter` instead of stdio, which makes it reusable for the language server.
- Stop using the exception `IllegalDafnyFile` to handle common scenarios.
- Replace obsolete usage of `Lazy<IdeState>` in `ProjectManager` with `IdeState`

### How has this been tested?
- Add a test for using the standard library with `dafny server`
- Add a test for using an unfound library file with `dafny server`
- Add a test for using a library file with `dafny server`


Source/DafnyLanguageServer/Workspace/Compilation.cs
```diff
@@ -92,18 +107,55 @@ CompilationInput input
   }
 
   public void Start() {
+    Project.Errors.CopyDiagnostics(errorReporter);
+    RootFiles = DetermineRootFiles();
+    var diagnosticsCopy = staticDiagnostics.ToImmutableDictionary(k => k.Key,
+      kv => kv.Value.Select(d => d.ToLspDiagnostic()).ToImmutableList());
+    updates.OnNext(new FoundFiles(Project, RootFiles!, diagnosticsCopy));
```
MikaelMayer, line 114: Can you please explain me in a few sentences why these lines? Or add a comment?

Source/DafnyLanguageServer/Workspace/FoundFiles.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Dafny.LanguageServer.Workspace.Notifications;
+using Microsoft.Extensions.Logging;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+namespace Microsoft.Dafny.LanguageServer.Workspace;
+
+record FoundFiles(
```
MikaelMayer, line 12: Why do we have these records?

Source/DafnyLanguageServer/Workspace/INotificationPublisher.cs
```diff
@@ -11,6 +11,6 @@ public interface INotificationPublisher {
     /// Publishes the diagnostics of the specified dafny document to the connected LSP client.
     /// </summary>
     /// <param name="state">The document whose diagnostics should be published.</param>
-    Task PublishNotifications(IdeState previousState, IdeState state);
+    void PublishNotifications(IdeState previousState, IdeState state);
```
MikaelMayer, line 14: Makes sense. Fire and forget.

Source/DafnyLanguageServer/Workspace/IdeState.cs
```diff
@@ -25,6 +25,7 @@ public record IdeVerificationResult(VerificationPreparationState PreparationProg
 /// </summary>
 public record IdeState(
   int Version,
+  ISet<Uri> OwnedUris, // TODO include project Uri
```
MikaelMayer, line 28: What is this TODO exactly?

Source/DafnyLanguageServer/Workspace/ProjectManager.cs
```diff
@@ -103,33 +107,25 @@ public class ProjectManager : IDisposable {
     options = DetermineProjectOptions(project, serverOptions);
     options.Printer = new OutputLogger(logger);
     boogieEngine = new ExecutionEngine(options, cache, scheduler);
-    var initialCompilation = GetCompilationInput();
+    var initialCompilation = new CompilationInput(options, version, Project);
```
MikaelMayer, line 110: Here I see CompilationInput. What has changed exaclty? What is Compilation vs. CompilationInput?

keyboardDrummer: `CompilationInput` is all the inputs of a compilation, so the DafnyProject and the DafnyOptions that result from combining the CLI argument with the DafnyProject.

`Compilation` is the class responsible for doing most of what our codebase does:
1. Based on the project file and files passed to the CLI, determine what the root files for parsing are, and whether they are preverified or precompiled
1. Parse files
1. Resolve modules
1. Verify declarations
1. Not yet: translate the program to a target language

Source/DafnyLanguageServer/Workspace/IdeState.cs
keyboardDrummer: It's resolved now, by this line: https://github.com/dafny-lang/dafny/pull/4770/files#diff-1c97a34c28a0bbbc6560c4c5b1649d777a9b200abf19099af19f73601e48228cR32

Source/DafnyLanguageServer/Workspace/FoundFiles.cs
keyboardDrummer: Is your question why is this a record instead of a class? Because it saves boilerplate in creating fields.

Or is your question, why is there a compilation event called FoundFiles? That's because the start of the compilation involves evaluating and validating the paths and globs specified in the CLI and in the project file, and this event exposes the results of that, which provides useful information for the IDE.

Source/DafnyLanguageServer/Language/DafnyLangParser.cs
```diff
@@ -37,25 +34,25 @@ public sealed class DafnyLangParser : IDafnyParser {
         : new ProgramParser(innerParserLogger, fileSystem);
     }
 
-    public Program Parse(CompilationInput compilation, ErrorReporter reporter, CancellationToken cancellationToken) {
+    public Program Parse(Compilation compilation, CancellationToken cancellationToken) {
```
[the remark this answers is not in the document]
keyboardDrummer: This PR moves the determination of what the root files for a particular compilation are from `ProjectManager` to `Compilation`. Since determining these files can discover user errors, its good to have this be part of Compilation so the handling of those errors works the same as other user errors that occur during compilation, such as parsing errors. Also, `Compilation` will be moved to `DafnyCore` and then the CLI and the server can re-use the same file discovery code.

Since the parser needs to know the root files to parse, which are now stored in the `Compilation`, but not in CompilationInput, it takes a `Compilation` instead of a `CompilationInput`.

Source/DafnyLanguageServer.Test/Various/CompilationStatusNotificationTest.cs
```diff
@@ -44,7 +44,6 @@ public class CompilationStatusNotificationTest : ClientBasedLanguageServerTest {
       var a4 = await compilationStatusReceiver.AwaitNextNotificationAsync(CancellationToken);
       var a5 = await compilationStatusReceiver.AwaitNextNotificationAsync(CancellationToken);
       var a6 = await compilationStatusReceiver.AwaitNextNotificationAsync(CancellationToken);
-      await CreateOpenAndWaitForResolve(producerSource, Path.Combine(directory, "producer.dfy"));
```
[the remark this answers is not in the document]
keyboardDrummer: Woops, brought it back

Source/DafnyLanguageServer/Workspace/FoundFiles.cs
MikaelMayer: My question was merely the second one. Thanks for answering both!

Source/DafnyDriver/Commands/DafnyCli.cs
```diff
@@ -489,11 +488,18 @@ class WritersConsole : IConsole {
     // which is not handled well.
     if (options.Get(CommonOptionBag.UseStandardLibraries)) {
       var reporter = new ConsoleErrorReporter(options);
+      if (options.CompilerName is null or "cs" or "java" or "go" or "py" or "js") {
```
MikaelMayer, line 491: Could you introduce a boolean name indicating why you perform this test? e.g. like "noTargetNeeded".

Source/IntegrationTests/TestFiles/LitTests/LitTest/stdlibs/StandardLibraries_Errors.dfy.expect
```diff
@@ -4,7 +4,9 @@ Dafny program verifier finished with 1 verified, 0 errors
 Dafny program verifier finished with 1 verified, 0 errors
 StandardLibraries_Errors.dfy(20,29): Error: module DafnyStdLibs does not exist (position 0 in path DafnyStdLibs.Wrappers)
 1 resolution/type errors detected in StandardLibraries_Errors.dfy
-*** Error: Cannot load DafnyStandardLibraries-notarget.doo: --unicode-char is set locally to False, but the library was built with True
+CLI: Error: cannot load DafnyStandardLibraries-notarget.doo: --unicode-char is set locally to False, but the library was built with True
```
MikaelMayer, line 7: Any way to remove the ":" between "CLI" and "Error" to avoid colon repetition?

MikaelMayer (approved):

Source/IntegrationTests/TestFiles/LitTests/LitTest/stdlibs/StandardLibraries_Errors.dfy.expect
keyboardDrummer: There's an existing method:

```
public string ErrorToString(ErrorLevel header, IToken tok, string msg) {
    return $"{tok.TokenToString(Options)}: {header.ToString()}: {msg}";
  }
```

that creates the double colon. We could remove the first one.

Source/DafnyDriver/Commands/DafnyCli.cs
robin-aws: This is from my PR (#4790), it's just moving in this one. I intend to clean this up together with some other stdlibs follow-up - I figure it would be better to add an optional `StdLibDooFile` property on backends instead.

keyboardDrummer: I copied this from DafnyCLI, but it checks whether a target specific .doo file exists. The block won't be necessary anymore once we merge the replaceable PR.

PR #4774, opened 2023-11-14
Remove Compilation.Finished

keyboardDrummer:
### Description
Remove some complicated code in `Compilation` that was only used by tests to wait for the IDE to be done with everything, but that code is no longer needed because the tests can use LSP notifications to wait for the same state.

### How has this been tested?
This is a refactoring and requires no additional tests


MikaelMayer (approved):
Cleanup feels good thanks

PR #4775, opened 2023-11-14
Test coverage for Arithmetic

ssomayyajula:
### How has this been tested?

See `examples/Arithmetic/*.dfy`.

Note: testing interesting properties about the relationship between pow/log and div/mod seems throws Verification Inconclusive, so they're omitted. That's odd!


ssomayyajula:
Note that the internal division and modulus are tested ad hoc since they differ from Dafny's own.

robin-aws:
> Note: testing interesting properties about the relationship between pow/log and div/mod seems throws Verification Inconclusive, so they're omitted. That's odd!

That doesn't surprise me too much, especially since the example project file has a pretty aggressive upper bound on resource usage. We may increase that a bit, especially on a per-proof basis, but at the same time there's also probably opportunities to make it cheaper to prove those properties if we provide alternate versions of some of the specifications. I talk about that a bit in [this section of CONTRIBUTING.md](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/CONTRIBUTING.md#on-brittleness).

It will be useful in the future to have a way to measure some notion of "specification coverage" driven by examples, so we can ensure we've demonstrated and tested typical ways of using the specifications. But for now since we DO have a way to measure execution coverage, and that's important too, it's good to focus on that first.

robin-aws (approved):

ssomayyajula:
Thanks! I wonder if there's actually some issue here: this shouldn't have any real verification burden, but the program fails to verify even with this high resource limit (Dafny just hangs forever). Of course, we can save this for way later.

```dafny
method {:test} {:rlimit 30000} TestDivMod() {
    for i := -10 to 10 {
      for j := -10 to 10 {
        expect i * DivRecursive(i, j) + ModRecursive(i, j) == j;
      }
    }
  }
```

PR #4776, opened 2023-11-14
100% branch coverage on Wrappers and Collections

robin-aws:
### Description
Also included some cleanup (it's amazing what you spot when you are writing tests for code and not just reviewing it in PR!):
* Renamed `Seq` to `Seqs`, since all other `Collections` modules use that naming pattern.
* Remove some duplication of `SortedBy` and `IsMinimal` between Seq and Relations
* Changed the order of parameters on a few functions/methods for better consistency across modules
* Removed `{:options "-functionSyntax:4"}` since we don't need that in this project compared to `dafny-lang/libraries`

### How has this been tested?
It's nothing but tests! :)


Source/DafnyStandardLibraries/src/DafnyStdLibs/Collections/Seqs.dfy
```diff
@@ -949,4 +916,11 @@ module DafnyStdLibs.Collections.Seq {
 
       [right[0]] + MergeSortedWith(left, right[1..], lessThanOrEq)
   }
+
+  lemma LemmaNewFirstElementStillSortedBy<T>(newFirst: T, s: seq<T>, lessOrEqual: (T, T) -> bool)
```
robin-aws, line 920: Moved this here since it's really only useful as a helper lemma for MergeSortedWith and not terribly general.

alex-chew (approved):

PR #4777, opened 2023-11-15
feat(Optimize Go imports): run goimports on Go target code

ShubhamChaturvedi7:
### Description

This PR adds support to optimize imports in the generated Go target code. It runs the `goimports` tool to do the optimization.
### How has this been tested?

* Manual Testing and the existing Integrating Tests


.github/workflows/integration-tests-reusable.yml
```diff
@@ -87,6 +87,22 @@ jobs:
       with:
         java-version: 18
         distribution: corretto
+    - name: Set up newest supported Go
+      if: matrix.target-language-version == 'newest'
```
robin-aws, line 91: Just FYI, we aren't currently ever populating `matrix.target-language-version`, because running the entire suite twice for the sake of a small percentage of tests wasn't frugal. See https://github.com/dafny-lang/dafny/issues/1983 for more background, especially my latest comment on how I think we should solve this problem.

Source/DafnyCore/Plugins/IExecutableBackend.cs
```diff
@@ -128,9 +128,9 @@ public abstract class IExecutableBackend {
   }
 
   /// <summary>
-  /// Perform any required cleanups after generating code with <c>Compile</c> and <c>EmitCallToMain</c>.
+  /// Perform any required processing after generating code with <c>Compile</c> and <c>EmitCallToMain</c>.
   /// </summary>
-  public virtual void OnPostCompile() { }
+  public abstract bool OnPostCompile(string dafnyProgramName, string targetFilename, TextWriter outputWriter);
```
robin-aws, line 133: I think it makes more sense to keep this as an empty virtual method, so implementing classes don't have to do anything if they don't need this hook.

.github/workflows/integration-tests-reusable.yml
robin-aws: This might be the reason the CI is still failing for you in fact.

Source/DafnyCore/Plugins/IExecutableBackend.cs
robin-aws: Also `targetFilename` should probably be `targetDirectory`, since you're passing `targetPaths.SourceDirectory` for it.

docs/Installation.md
```diff
@@ -205,7 +205,8 @@ To set up Dafny to compile to Go:
 * Install Go from [https://golang.org/dl](https://golang.org/dl) or
    * Linux: `sudo apt install golang`
    * Mac: `brew install golang`
-* Make sure `go` is on your path
+* Install `goimports` from [https://pkg.go.dev/golang.org/x/tools/cmd/goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports)
```
robin-aws, line 208: An explicit command to execute as well would be slightly friendlier

Source/DafnyRuntime/DafnyRuntimeGo/System_/System_.go
```diff
@@ -4,47 +4,47 @@
 package _System
 
 import (
-  _dafny "dafny"
-  os "os"
+	_dafny "dafny"
```
robin-aws, line 7: I take it these changes are because `goimports` applies auto-formatting as well? 

No issues with that and in fact that's kind of a nice side-effect of this change!

robin-aws (dismissed):
🚀

robin-aws (approved):

robin-aws (approved):

Issue #4778, opened 2023-11-15
Enable finer-grained control of redundant- and contradictory-assumption warnings


alex-chew:
### Background and Motivation

The `--warn-redundant-assumptions` and `--warn-contradictory-assumptions` options are useful for locating redundant and contradictory assumptions, but can generate false positives in nontrivial proofs. In some cases the proofs can be rewritten to avoid the warnings - but not always, it seems. It would be nice to be able to disable such warnings on a per-assertion (or per-declaration, or per-module) level.

Here is an excerpt from the Unicode libraries that generates contradictory-assumption warnings:
```dafny
  type CodeUnit
  type CodeUnitSeq = seq<CodeUnit>
  type MinimalWellFormedCodeUnitSeq = s: CodeUnitSeq
    | IsMinimalWellFormedCodeUnitSubsequence(s)
    witness *

  function IsMinimalWellFormedCodeUnitSubsequence(s: CodeUnitSeq): (b: bool)
    ensures b ==>
              && |s| > 0
              && forall i | 0 < i < |s| :: !IsMinimalWellFormedCodeUnitSubsequence(s[..i])
    decreases |s|

  /**
   * If minimal well-formed code unique subsequences `m1` and `m2` are prefixes of `s`, then they are equal.
   */
  lemma LemmaUniquePrefixMinimalWellFormedCodeUnitSeq(
    s: CodeUnitSeq, m1: MinimalWellFormedCodeUnitSeq, m2: MinimalWellFormedCodeUnitSeq
  )
    decreases |s|, |m1|, |m2|
    requires m1 <= s
    requires m2 <= s
    ensures m1 == m2
  {
    // Handle only the |m1| <= |m2| case explicitly
    if |m1| > |m2| {
      LemmaUniquePrefixMinimalWellFormedCodeUnitSeq(s, m2, m1);
    } else {
      assert m1 <= m2;
      assert m1 == m2 by {
        if m1 < m2 {
          // Here
          assert false by { assert m1 == m2[..|m1|]; }
        }
      }
    }
  }
```

The `assert false` near the end (labeled `// Here`) cannot be verified without the `by { ... }` portion. But the assertion within the `by { ... }` generates two warnings:

```
UnicodeEncodingForm.dfy(106,35): Error: proved using contradictory assumptions: assertion always holds
    |
106 |           assert false by { assert m1 == m2[..|m1|]; }
    |                                    ^^^^^^^^^^^^^^^^

[...]

PR #4780, opened 2023-11-15
feat: import Unicode libraries

alex-chew:
This PR imports the Unicode libraries from `dafny-lang/libraries`.

Notes:
- This disables redundant- and contradictory-assumptions warnings in `dfyconfig.toml`, as not all of the many false-positives can be resolved by rewriting existing proofs (as far as I can tell). We'll want to re-enable them in the future, for example if/when [finer-grained control](https://github.com/dafny-lang/dafny/issues/4778) of such warnings is possible.
- As a consequence of the above, verification performance of all the libraries are subject to perturbation. This PR also makes some non-Unicode-related changes to get other proofs to pass under the new set of options.
- Because the `DafnyStdLibs` only supports a single set of verification options right now, and we are proceeding with `--unicode-char:true`, the `UnicodeStringsWithoutUnicodeChar` module is not included in this PR.

TODO:
- [x] Examples + test coverage
- [x] Top-level module documentation

### How has this been tested?


Source/DafnyStandardLibraries/Makefile
```diff
@@ -10,6 +10,10 @@ DOO_FILE_ARITHMETIC_TARGET=binaries/DafnyStandardLibraries-arithmetic.doo
 
 all: check-binary test check-format check-examples
 
+verify:
```
robin-aws, line 13: Nice improvements! I had some of these as part of the FileIO import too, especially since the build will change per language too. I did find a slightly cleaner approach but I can apply that in my PR instead of insisting on it here.

Source/DafnyStandardLibraries/examples/Helpers.dfy
```diff
@@ -0,0 +1,19 @@
+/*******************************************************************************
+ *  Copyright by the contributors to the Dafny Project
+ *  SPDX-License-Identifier: MIT
+ *******************************************************************************/
+
+module Helpers {
+  import opened DafnyStdLibs.Wrappers
+    // TODO: consider tweaking /testContracts to support this use case better.
```
robin-aws, line 8: +1, and would prefer an issue to a TODO, but don't want to block on it

Source/DafnyStandardLibraries/src/DafnyStdLibs/Unicode/Unicode.dfy
```diff
@@ -0,0 +1,17 @@
...
+ This module defines concepts from the core specification
+ of the [Unicode Standard 15.1.0](https://www.unicode.org/versions/Unicode15.1.0/):
+ scalar values, code points, important kinds of code point sequences,
+ Unicode encoding forms and encoding schemes (including UTF-8 and UTF-16),
+ and helpful properties and functions for manipulation and conversion of Unicode strings.
+ */
+module DafnyStdLibs.Unicode {
+  // This top-level module contains no definitions.
+  // It exists to enable import between child modules,
+  // and so that its doc comment appears in generated documentation.
```
robin-aws, line 16: This is a good point and too bad we can't do the same for the top level `DafnyStdLibs` module since it has to appear in multiple doo files. But could hopefully address that somehow in the future.

robin-aws (approved):
LGTM, very nice testing coverage on this one

Source/DafnyStandardLibraries/examples/Helpers.dfy
alex-chew: I just copied this comment over from where `Helpers` was originally defined, but here's a new issue for it: https://github.com/dafny-lang/dafny/issues/4786

PR #4784, opened 2023-11-15
Strings

### LINKED ISSUE #4779: Character literals not encoded as CodePoints in map displays in Java
### Dafny version

4.1.0 (may be present in earlier versions)

```dafny
datatype Option<T> = Some(value: T) | None

function Get<X, Y>(m: map<X, Y>, x: X): Option<Y> {
  if x in m then Some(m[x]) else None
}

method {:test} SureIsSomeCharacter() {
  var m := map[2 := 'b'];
  assert Get(m, 2) == Some('b');
  expect Get(m, 2) == Some('b');
}
```

```code
% dafny test -t:java Scratch.dfy

SureIsSomeCharacter: FAILED
        Scratch.dfy(11,2): expectation violation
[Program halted] Scratch.dfy(2,0): Test failures occurred: see above.
```

### What happened?

Root cause seems to be encoding 'b' as a plain Java character literal instead of using the `dafny.CodePoint` wrapper as intended. Only affects `--unicode-char:true` mode. Normally this bug would result in a Java compilation error instead, but in this particular case the Java type checker doesn't catch the issue because of the generic method in the way.

Using the `testing-method: uniform-backend-testing` label since it was found when writing a test for the `DafnyStandardLibraries` project, and those tests are also run for all the stable backends by default.

Mac

### LINKED ISSUE #4791: Java compiler calls .value() on a CodePoint when it should not
The following runtime exception occurs because of the last call to `.value()` in this line `return ((__default.ToNat((str).take((java.math.BigInteger.valueOf((str).length())).subtract(java.math.BigInteger.ONE)))).multiply(__default.base())).add(((java.math.BigInteger)(java.lang.Object)((__default.charMap()).get(((dafny.CodePoint)((str).select(dafny.Helpers.toInt(((java.math.BigInteger.valueOf((str).length())).subtract(java.math.BigInteger.ONE)))))).value()))));`

which causes a map that has CodePoints as keys to be indexed using an Integer, thus returning `null`.

```
StringsExamples.TestToInt: Exception in thread "main" java.lang.NullPointerException: Cannot read field "signum" because "val" is null
        at java.base/java.math.BigInteger.add(BigInteger.java:1334)
        at DafnyStdLibs.Strings.DecimalConversion.__default.ToNat(__default.java:86)
        at DafnyStdLibs.Strings.DecimalConversion.__default.ToInt(__default.java:94)
        at DafnyStdLibs.Strings.__default.ToInt(__default.java:39)
        at StringsExamples.__default.TestToInt(__default.java:62)
        at _System.__default.__Test____Main__(__default.java:1219)
        at _System.__default.__Main(__default.java:1545)
        at stdlibexamples.lambda$main$0(stdlibexamples.java:118)
        at dafny.Helpers.withHaltHandling(Helpers.java:345)
        at stdlibexamples.main(stdlibexamples.java:118)
```

A similar problem can occur in a type-checked location, causing the Java compiler to emit an error:
```
/Users/rwillems/SourceCode/dafny2/Source/DafnyStandardLibraries/build/stdlibexamples-java/DafnyStdLibs/Strings/HexConversion/__default.java:34: error: incompatible types: int cannot be converted to CodePoint
      return dafny.DafnyMap.<dafny.CodePoint, java.math.BigInteger>update(__default.CreateCharMap((i).add(java.math.BigInteger.ONE), (s).drop(java.math.BigInteger.ONE)), ((dafny.CodePoint)((s).select(dafny.Helpers.toInt((java.math.BigInteger.ZERO))))).value(), i);
```

keyboardDrummer:
### Description
- Add the Strings library
- Add `LittleEndianNat.dfy` to the Arithmetic library
- Change the folder structure for the files in `DafnyStdLibs`, so every file is under the right `dfyconfig.toml` directory wise. This allows the IDE to work well when working on the library code.
- Fixes #4791 and fixes #4779, since they were exposed by testing on Java

### How has this been tested?
- Example file added for the Strings library
- Examples related to `LittleEndianNat.dfy` added to `ArithmeticExamples.dfy` 
- Testing coverage is showing up as lower than expected, likely because the execution test coverage instrumentation is under-reporting coverage for abstract modules. Will follow up with a fix to that first and then increase coverage of this code if necessary.


Source/DafnyStandardLibraries/examples/StringsExamples.dfy
```diff
@@ -0,0 +1,68 @@
...
+  method {:test} TestOfInt() {
+    expect OfInt(0) == "0";
+    expect OfInt(3) == "3";
+    expect OfInt(302) == "302";
+    expect OfInt(-3) == "-3";
+    expect OfInt(-302) == "-302";
+  }
+
+  // TODO The requires clauses of these calls cause the verifier to hang, even though they seem rather trivial
+  // method {:test} TestToInt() {
```
keyboardDrummer, line 14: To be more accurate, with the configured resource limit of 1000 they run out of resources, but if I increase the resources the verification does not respond after several minutes. The problematic part of the requires clause that does not want to verify is that each symbol from the input in a key in `charMap`. Even doing `assert '0' in DecimalConversion.charMap` causes verification to hang, which to me seems like an extremely tiny steps given the definition of `DecimalConversion.charMap`: `const charMap :=
      map[
        '0' := 0, '1' := 1, '2' := 2, '3' := 3, '4' := 4, '5' := 5, '6' := 6, '7' := 7, '8' := 8, '9' := 9
      ]

Source/DafnyStandardLibraries/binaries/DafnyStandardLibraries.zip
alex-chew, line 1: Is this file supposed to be checked in?

Source/DafnyStandardLibraries/binaries/DafnyStandardLibraries/program
alex-chew, line 1: Likewise, is this file supposed to be checked in?

Source/DafnyStandardLibraries/src/DafnyStdLibs/DisableNonLinearArithmetic/Arithmetic/DivMod.dfy
```diff
@@ -7,8 +7,8 @@
  *******************************************************************************/
 
 /* Every lemma comes in 2 forms: 'LemmaProperty' and 'LemmaPropertyAuto'. The
-former takes arguments and may be more stable and less reliant on Z3
-heuristics. The latter includes automation and its use requires less effort */
+former takes arguments and may be emore stable and less reliant on Z3
```
alex-chew, line 10:
```suggestion
former takes arguments and may be more stable and less reliant on Z3
```

Source/DafnyStandardLibraries/src/DafnyStdLibs/DisableNonLinearArithmetic/Strings.dfy
```diff
@@ -0,0 +1,279 @@
...
+    - the type of character used, Char
+    - a conversion from nat to Char and vice versa
+   */
+  abstract module ParametricConversion refines Arithmetic.LittleEndianNat {
+    import opened Wrappers
+
+    type Char(==)
+    type String = seq<Char>
+
+    type CharSeq = chars: seq<Char> | |chars| > 1 witness *
```
alex-chew, line 22: Should `CharSeq` also have the restriction that its elements are unique?

Source/DafnyStandardLibraries/binaries/DafnyStandardLibraries.zip
robin-aws: Definitely not, accidentally checked in debugging state

Source/DafnyStandardLibraries/src/DafnyStdLibs/DisableNonLinearArithmetic/Strings.dfy
robin-aws: The deeper point is that there's no specification that actually requires that! I think that's fine as having these utilities available in Dafny is important even if not all properties are proven, but I did cut https://github.com/dafny-lang/dafny/issues/4795 to track that for later.

alex-chew (approved):

PR #4788, opened 2023-11-16
fix: print error context snippets after their messages

### LINKED ISSUE #4787: Redundant- and contradictory-assumptions warnings are incorrectly formatted
### Dafny version

master (f81c316e1c3eb9de136c635f9d660bb1de3d3dbe)

```dafny
module M {
  opaque function ToSet(xs: seq<int>): set<int> {
    set x: int | x in xs
  }

  lemma LemmaCardinality(xs: seq<int>)
    ensures |ToSet(xs)| <= |xs|
  {}

  lemma LemmaEmpty(xs: seq<int>)
    requires xs == []
    ensures |ToSet(xs)| == 0
  {
    assume false;
  }
}
```

```code
$ dafny verify --warn-redundant-assumptions --warn-contradictory-assumptions tmp.dfy
tmp.dfy(8,2): Error: a postcondition could not be proved on this return path
  |
8 |   {}
  |   ^

tmp.dfy(7,12): Related location: this is the postcondition that could not be proved
  |
7 |     ensures |ToSet(xs)| <= |xs|
  |             ^^^^^^^^^^^^^^^^^^^

   |
12 |     ensures |ToSet(xs)| == 0
   |             ^^^^^^^^^^^^^^^^

tmp.dfy(12,12): Warning: ensures clause proved using contradictory assumptions
   |
11 |     requires xs == []
   |              ^^^^^^^^

tmp.dfy(11,13): Warning: unnecessary requires clause

```

### What happened?

The two errors and two warnings are all correct, but the warnings' messages are displayed below the code snippet context, whereas they should appear above the context.

Mac

alex-chew:
### Description
Fixes #4787.

### How has this been tested?

Added `git-issue-4787.dfy` to test the reported case.

### DIFF
--- a/Source/DafnyCore/Generic/Reporting.cs
+++ b/Source/DafnyCore/Generic/Reporting.cs
@@ -66,13 +66,13 @@ namespace Microsoft.Dafny {
           errorLine += "\n";
         }
 
+        Options.OutputWriter.Write(errorLine);
+
         if (Options.Get(DafnyConsolePrinter.ShowSnippets) && tok != Token.NoToken) {
           TextWriter tw = new StringWriter();
           new DafnyConsolePrinter(Options).WriteSourceCodeSnippet(tok.ToRange(), tw);
           Options.OutputWriter.Write(tw.ToString());
         }
-
-        Options.OutputWriter.Write(errorLine);
 
         if (Options.OutputWriter == Console.Out) {
           Console.ForegroundColor = previousColor;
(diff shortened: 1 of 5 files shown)


Source/DafnyCore/Generic/Reporting.cs
```diff
@@ -66,14 +66,14 @@ public class ConsoleErrorReporter : BatchErrorReporter {
           errorLine += "\n";
         }
 
+        Options.OutputWriter.Write(errorLine);
+
```
atomb, line 70: The fact that this was fixed by changing the core message reporting code makes me think that some of the other messages (which already came before the snippet) aren't using this and maybe should be? Not that we need to fix that now, it's just an inconsistency we might want to resolve at some point.

atomb (approved):

Source/DafnyCore/Generic/Reporting.cs
alex-chew: Yeah, I didn't get a chance to search for where the other messages (like normal verification failures) were bring printed from, but it's a good idea to unify them to use this logic.

PR #4798, opened 2023-11-17
Separate UI code and business logic for the resolve and verify commands

keyboardDrummer:
### Changes
At the heart of this PR is an architectural change: to lay the foundation for separating UI code from business logic. Part of this is to use Boogie purely as an API, instead of as a library that produces textual output for the CLI.

This change is not _necessary_, but I believe it is a good step towards giving us more control over the UI experience of our CLI, and to make it easier to work with our codebase. In the future section down below are some ideas of what this PR enables.

Because there is a lot of UI code intermingled with a lot of business code, making this architectural change produces a lot of code changes, so this PR only takes the first step: the pathway where UI and logic are separated is only used when using `dafny resolve` or `dafny verify`. I will update the remaining commands in follow-up PRs.

#### Refactoring
- Move `Compilation`, a UI agnostic back-end for driving the Dafny compiler pipeline, from the project `LanguageServer` to `DafnyCore`, so it can be used by the CLI as well. Reusing this code means more IDE code is tested using the CLI test, and you can see the effect of this additional testing from the IDE fixes down below, yay!
- Introduce `CliCompilation` which uses the CLI as a user interface to track the results of `Compilation`.
- Let the commands `dafny resolve` and `dafny verify` use `CliCompilation` instead of `CompilerDriver `.
- Due to the introduction of `CliCompilation`, rename `CompilerDriver` to `SynchronousCliCompilation`, even though there are many features `CliCompilation` does not support yet.
- Remove `DafnySymbolKind` from DafnyCore, which was a duplicate of an LSP enum that is now directly accessible in DafnyCore

#### Behavior
- IDE fix: enable comma's in the library flag to specify multiple libraries, as the CLI does
- IDE fix: enable having the same file as input multiple times, as the CLI does
- IDE fix: support verification for some code generating Dafny constructors, like iterators
- IDE fix: the option solver-path is respected by the IDE
- IDE fix: when verification can not start due to a technical issue, the IDE tells the user what happened
- Make it easy to update .expect files by setting the flag `DiffCommand.UpdateExpectFile` to true. This was broken before this PR.
- Configure littish tests so they time out after running for 120 minutes. They normally complete in less than 30m on all platforms, so 120m should be ample.
- Fix a bug in the cloning code, that was discovered because the CLI tests are using the server back-end code, which clones the program after parsing. (Is this good or bad? In the future I hope we'll have an immutable AST so then the question becomes redundant)
- Make the order in which verification error are shown on the CLI consistent. Now errors are always shown in the order in which they occur in the file

  Example diff:
```
- git-issue-4074.dfy(24,11): Error: assertion might not hold
git-issue-4074.dfy(5,9): Error: assertion might not hold
git-issue-4074.dfy(14,11): Error: assertion might not hold
+ git-issue-4074.dfy(24,11): Error: assertion might not hold
```

- Consistently let CLI diagnostics have a Warning/Error label. Example diff:
```
- github-issue-4804.dfy(19,8): Verification out of resource (Power.LemmaPowSubAddCancel (correctness))
+ github-issue-4804.dfy(19,8): Error: Verification out of resource (Power.LemmaPowSubAddCancel (correctness))
```
- The way in which custom error messages is shown is slightly different. Example diff:
```
CustomErrorMesage.dfy(42,63): Error: this loop invariant could not be proved on entry
- CustomErrorMesage.dfy(42,63): Related message: position variable out of range
+  Related message: position variable out of range
```

### Future
This PR paves the way for:
- Adding a `--filter` option that operates on Dafny symbols/files, and possibly also filters parsing and resolution
- Update the CLI output. 
  - Removed useless "related message" output. Example below this list.
  - Improve summary output. Dafny currently reports `Dafny program verifier finished with 21 verified, 0 errors` but the number 21 does not refer to anything visible to the user. It would be better to report the number of Dafny symbols that were verified.
  - Do not report `Dafny program verifier did not attempt verification` when running `dafny resolve`

```
Definedness.dfy(175,27): Error: this loop invariant could not be proved on entry
- Definedness.dfy(175,27): Related message: loop invariant violation
```

### How has this been tested?
- Updated existing tests to account for behavioral changes


Source/DafnyDriver/LegacyCliCompilation.cs
```diff
@@ -35,14 +35,23 @@ namespace Microsoft.Dafny {
...
     public static async Task<int> Run(DafnyOptions options) {
+
+      if (options.DafnyProject == null) {
+        var uri = options.CliRootSourceUris.FirstOrDefault() ?? new Uri(Directory.GetCurrentDirectory());
+        options.DafnyProject = new DafnyProject {
+          Includes = Array.Empty<string>(),
+          Uri = uri,
+          ImplicitFromCli = true
+        };
+      }
```
robin-aws, line 54: Didn't follow why this became necessary

robin-aws (changes requested):
Got through a high level review, will finish in a second review batch. Overall love the change, and the improvements in output consistency seem like a good sign we're moving in the right direction!

Regarding the excellent PR description, I'm just a bit unclear on why this refactoring makes the proposed Future changes easier.

Source/DafnyCore/ProofDependencyWarnings.cs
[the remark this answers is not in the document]
keyboardDrummer: Also they make the order of warnings match the line order, although it's not the only reason that proof dep warnings are reported in a better order.

Source/DafnyCore/AST/Grammar/Printer.cs
```diff
@@ -191,7 +191,9 @@ public class Printer {
     }
 
     public void PrintProgramLargeStack(Program prog, bool afterResolver) {
+#pragma warning disable VSTHRD002
```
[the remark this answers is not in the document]
keyboardDrummer: I don't know. I just wanted to get rid of the warning. I can remove the pragma from this PR if you prefer.

I have another PR that removes the need for these warnings, but it needs work to let the build pass: https://github.com/dafny-lang/dafny/pull/4585

Source/XUnitExtensions/Lit/DiffCommand.cs
```diff
@@ -33,8 +33,8 @@ public class DiffCommand : ILitCommand {
       TextWriter outputWriter, TextWriter errorWriter) {
       var actual = File.ReadAllText(ActualPath);
       if (UpdateExpectFile) {
-        var nonSymlinkedPath = Path.GetFullPath(ExpectedPath).Replace("Source/IntegrationTests/bin/Debug/net6.0/TestFiles/LitTests/LitTest", "Test");
-        File.WriteAllText(nonSymlinkedPath, actual);
+        var path = Path.GetFullPath(ExpectedPath).Replace("bin/Debug/net6.0/", "");
+        File.WriteAllText(path, actual);
```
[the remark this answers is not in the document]
keyboardDrummer: I'm not sure I understand what you're saying. Because of `.Replace("bin/Debug/net6.0/", "")`, it overwrites the `.expect` files in their original location.

Source/XUnitExtensions/Lit/DiffCommand.cs
```diff
@@ -10,7 +10,7 @@ namespace XUnitExtensions.Lit {
   /// because 'diff' does not exist on Windows.
   /// </summary>
   public class DiffCommand : ILitCommand {
-    private static readonly bool UpdateExpectFile = false;
+    public static readonly bool UpdateExpectFile = false;
```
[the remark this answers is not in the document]
keyboardDrummer: I'd prefer to do that in a separate PR. I just enabled this in this PR because I needed it.

Source/DafnyDriver/LegacyCliCompilation.cs
keyboardDrummer: Seems like its not. Might have been defensive coding on my part. I'll remove it.

Source/DafnyDriver/LegacyCliCompilation.cs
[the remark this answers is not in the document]
keyboardDrummer: > SynchronousCliCompilation

Sure. Renamed.

Source/IntegrationTests/TestFiles/LitTests/LitTest/logger/ProofDependencyWarnings.dfy
```diff
@@ -1,4 +1,4 @@
 // RUN: %baredafny verify --use-basename-for-filename --show-snippets false --verify-included-files --warn-contradictory-assumptions --warn-redundant-assumptions "%s" > "%t"
-// RUN: %diff "%t" "%s.expect"
+// RUN: %diff "%s.expect" "%t"
```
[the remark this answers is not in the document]
keyboardDrummer: Won't fix in this PR

Source/DafnyCore/AST/Grammar/Printer.cs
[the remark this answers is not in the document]
robin-aws: Oh let's leave the pragmas but add a comment linking to https://github.com/dafny-lang/dafny/pull/4585, that's exactly the context I was looking for.

Source/XUnitExtensions/Lit/DiffCommand.cs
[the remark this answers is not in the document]
robin-aws: Ah you're right, I thought it was relying on symlinking inside the output directory. The point stands since some day the "net6.0" part will change, but not worth worrying about now.

Source/DafnyCore/AST/Grammar/Printer.cs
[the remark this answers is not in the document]
keyboardDrummer: Actually that PR does not resolve all of these. In particular it does not resolve this one.

Do you mind if I leave changes to this for a follow-up PR ? I'm not sure about the trade-off between the current approach and using `JoinableTaskFactory.Run`

Source/DafnyCore/Pipeline/Compilation.cs
```diff
@@ -123,11 +133,19 @@ CompilationInput input
     }
 
     foreach (var uri in Options.CliRootSourceUris) {
-      var file = DafnyFile.CreateAndValidate(errorReporter, fileSystem, Options, uri, Token.Cli);
+      var shortPath = Path.GetRelativePath(Directory.GetCurrentDirectory(), uri.LocalPath);
+      var file = DafnyFile.CreateAndValidate(errorReporter, fileSystem, Options, uri, Token.Cli,
+        $"command-line argument '{shortPath}' is neither a recognized option nor a Dafny input file (.dfy,.doo,.toml).");
```
robin-aws, line 138:
```suggestion
        $"command-line argument '{shortPath}' is neither a recognized option nor a Dafny input file (.dfy, .doo, or .toml).");
```

Source/DafnyCore/Generic/ReporterExtensions.cs
```diff
@@ -36,26 +40,27 @@ public static class ErrorReporterExtensions {
...
   private static string FormatRelated(string related) {
     return $"Could not prove: {related}";
   }
 
-  public static IEnumerable<DafnyRelatedInformation> CreateDiagnosticRelatedInformationFor(IToken token, string? message) {
+  public static IEnumerable<DafnyRelatedInformation> CreateDiagnosticRelatedInformationFor(IToken token, string? message, bool usingSnippets) {
     var (tokenForMessage, inner, newMessage) = token is NestedToken nestedToken ? (nestedToken.Outer, nestedToken.Inner, nestedToken.Message) : (token, null, null);
     var dafnyToken = BoogieGenerator.ToDafnyToken(true, tokenForMessage);
-    if (dafnyToken is RangeToken rangeToken) {
+    if (!usingSnippets && dafnyToken is RangeToken rangeToken) {
```
robin-aws, line 52: Isn't it still useful to get the related information even when using snippets? Or are you just maintaining the current behaviour here?

Source/IntegrationTests/TestFiles/LitTests/LitTest/doofiles/Test2.dfy.expect
```diff
@@ -1 +1 @@
-*** Error: BadDoo.doo: End of Central Directory record could not be found.
+CLI: Error: malformed doo file BadDoo.doo
```
robin-aws, line 1: Definite improvement, thanks :)

Source/IntegrationTests/TestFiles/LitTests/LitTest/cli/projectFile/projectFile.dfy.expect
```diff
@@ -7,24 +7,25 @@ Error: file web.toml not found
...
 Only one project file can be used at a time. Both dfyconfig.toml and dfyconfig.toml were specified
 Error: The Dafny project file dfyconfig.toml contains the following errors: (1,1): the property includdddes does not exist.
 
 Warning: only Dafny project files named dfyconfig.toml are recognised by the Dafny IDE.
 Warning: option 'does-not-exist' that was specified in the project file, is not a valid Dafny option.
+invalidOption.toml(1,0): Warning: only Dafny project files named dfyconfig.toml are recognised by the Dafny IDE.
 
 Dafny program verifier did not attempt verification
 Error: file doesNotExist.toml not found
-Error: Could not parse value '3' for option 'warn-shadowing' that has type 'Boolean'
```
robin-aws, line 19: Also a nice improvement

Source/DafnyDriver/DafnyBackwardsCompatibleCli.cs
robin-aws, line 1: Off topic, but I'm not familiar with the history of this class: what past change was this backwards compatible for?

Source/DafnyLanguageServer/Language/Symbols/ISymbolTableFactory.cs
```diff
@@ -1,22 +1,15 @@
...
   public interface ISymbolTableFactory {
-    /// <summary>
-    /// Initializes a new symbol table with the given compilation unit.
-    /// </summary>
-    /// <param name="program">The parsed dafny program.</param>
-    /// <param name="compilationUnit">The compilation to create the symbol table of.</param>
-    /// <param name="cancellationToken">A token to cancel the update operation before its completion.</param>
-    /// <returns>A symbol table representing the provided compilation unit.</returns>
-    /// <exception cref="System.OperationCanceledException">Thrown when the cancellation was requested before completion.</exception>
-    /// <exception cref="System.ObjectDisposedException">Thrown if the cancellation token was disposed before the completion.</exception>
```
robin-aws, line 18: Why delete this comment?

Source/DafnyLanguageServer/Workspace/INotificationPublisher.cs
```diff
@@ -7,10 +7,6 @@ namespace Microsoft.Dafny.LanguageServer.Workspace {
   /// of a <see cref="CompilationInput"/> to the LSP client.
   /// </summary>
   public interface INotificationPublisher {
-    /// <summary>
-    /// Publishes the diagnostics of the specified dafny document to the connected LSP client.
-    /// </summary>
-    /// <param name="state">The document whose diagnostics should be published.</param>
```
robin-aws, line 13: Ditto

robin-aws (commented):
Reviewed most of the rest, just need to have a final look at the core changes in a third batch.

Source/DafnyDriver/CliCompilation.cs
```diff
@@ -0,0 +1,228 @@
...
+            Compilation.Reporter.Error(MessageSource.Verifier, ResolutionErrors.ErrorId.none, canVerify.Tok,
+              $"Internal error occurred during verification: {e.Message}\n{e.StackTrace}");
+          }
+        }
+      }
+
+      SynchronousCliCompilation.WriteTrailer(options, /* TODO ErrorWriter? */ options.OutputWriter, statSum);
+
+    } catch (TaskCanceledException) {
+    }
```
robin-aws, line 226: When does this happen? Why not catch OperationCanceledException instead?

Source/DafnyCore/Pipeline/Compilation.cs
```diff
@@ -153,19 +172,28 @@ CompilationInput input
...
+    // Allow specifying the same file twice on the CLI
+    return result.DistinctBy(d => d.Uri).ToList();
   }
 
   private async Task<Program> ParseAsync() {
     try {
       await started.Task;
+      if (HasErrors) {
+        throw new OperationCanceledException();
+      }
```
robin-aws, line 193: The name of this exception always bothered/confused me a little, and it feels worse using it in the CLI context now: it sounds like the user cancelled the pipeline, when it's really that the pipeline stoped early because of errors. 

Is this usage idiomatic for C# in general when hitting an "early return" as opposed to a parallel cancellation request?

robin-aws (commented):
A few more for today. :)

Source/DafnyCore/Pipeline/Compilation.cs
```diff
@@ -123,11 +133,19 @@ CompilationInput input
...
       if (file != null) {
         result.Add(file);
+      } else {
+        return result;
       }
     }
+    if (Options.UseStdin) {
+      var uri = new Uri("stdin:///");
+      result.Add(DafnyFile.CreateAndValidate(errorReporter, fileSystem, Options, uri, Token.Cli));
+    }
```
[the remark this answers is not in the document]
keyboardDrummer: You can't. `StdIn` is not an IDE option.

Source/DafnyDriver/DafnyBackwardsCompatibleCli.cs
keyboardDrummer: Introduction of the command based CLI. This class enables the non-command one.

docs/HowToFAQ/Errors-CommandLine.md
```diff
@@ -75,7 +75,7 @@ in an unrecognized form (e.g., with no filename extension).
 
 <!-- %check-cli -->
 ```bash
-dafny resolve z.zzz
+dafny build z.zzz
```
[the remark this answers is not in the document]
keyboardDrummer: `dafny resolve` does not take auxiliary files, so it no longer returns this error message. `dafny build` does take aux files and still returns this message.

Source/DafnyCore/Generic/ReporterExtensions.cs
keyboardDrummer: This block will append some source code to the error message, similar to what snippets do but less pretty.

Source/DafnyLanguageServer/Language/Symbols/ISymbolTableFactory.cs
keyboardDrummer: Was it useful? I would argue it's not. Even the interface `ISymbolTableFactory` is unused.

Source/DafnyDriver/CliCompilation.cs
keyboardDrummer: Added a comment and changed to `OperationCanceledException`. Happens when there's a parsing error so it can not resolve. I could change the interface of `Compiler` so it uses `null` instead of `OperationCanceledException` when it can't return a Task<ResolutionResult` due to a user program error. This means inside `Compilation` there's a need for more `if (resultIDependOn == null) { return null }` since we don't coast on the monadic behavior of async/await, but it might be more readable.

Source/DafnyCore/Pipeline/Compilation.cs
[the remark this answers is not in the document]
robin-aws: Sure, I'm just pointing out that technically you could add `--stdin` to the Language Server Launch Args, and now it seems like it would actually do something rather than being ignored. I assume it's harmless but asking just in case.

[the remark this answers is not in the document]
keyboardDrummer: If you use the CLI to specify an option to a command to which it doesn't apply, then the CLI will return an error. If you pass it through the project file, the CLI won't error but it'll be ignored.

Source/DafnyCore/Pipeline/Compilation.cs
keyboardDrummer: See above answer

Source/DafnyLanguageServer/Language/Symbols/ISymbolTableFactory.cs
robin-aws: Given the amount of refactoring going on I just wanted to make sure it wasn't dropped by accident. If you removed it because it wasn't useful that's just fine (and looking closer I agree it seems like "I'm being forced to document every method" rather than "I have something useful to say beyond the signature here" :)

Source/DafnyLanguageServer/Language/Symbols/DafnyLangSymbolResolver.cs
```diff
@@ -23,54 +22,41 @@ public class DafnyLangSymbolResolver : ISymbolResolver {
...
-    public DafnyLangSymbolResolver(ILogger<DafnyLangSymbolResolver> logger, ILogger<CachingResolver> innerLogger, ITelemetryPublisher telemetryPublisher) {
+    public DafnyLangSymbolResolver(ILogger<DafnyLangSymbolResolver> logger, ILogger<CachingResolver> innerLogger, TelemetryPublisherBase telemetryPublisher) {
       this.logger = logger;
       this.innerLogger = innerLogger;
       this.telemetryPublisher = telemetryPublisher;
     }
 
     private readonly ResolutionCache resolutionCache = new();
-    public CompilationUnit ResolveSymbols(DafnyProject project, Program program, CancellationToken cancellationToken) {
+    public void ResolveSymbols(Compilation compilation, Program program, CancellationToken cancellationToken) {
```
robin-aws, line 34: It feels redundant to be passing in a program separately from the compilation here (and probably other similar places). Does it have to be the same as `compilation.programAfterParsing`?

Source/DafnyLanguageServer/Language/Symbols/SymbolTableFactory.cs
```diff
@@ -16,20 +16,27 @@
...
-      var program = compilationUnit.Program;
+    public LegacySignatureAndCompletionTable CreateFrom(CompilationInput input,
+      ResolutionResult resolutionResult,
+      CancellationToken cancellationToken) {
+
+      var beforeLegacyServerResolution = DateTime.Now;
+      var compilationUnit = resolutionResult.HasErrors
+        ? new CompilationUnit(input.Project.Uri, resolutionResult.ResolvedProgram)
+        : new SymbolDeclarationResolver(logger, cancellationToken).ProcessProgram(input.Project.Uri, resolutionResult.ResolvedProgram);
+      // telemetryPublisher.PublishTime("LegacyServerResolution", project.Uri.ToString(), DateTime.Now - beforeLegacyServerResolution); TODO
```
robin-aws, line 33: TODO

Source/DafnyLanguageServer/Workspace/IdeState.cs
```diff
@@ -139,4 +161,379 @@ public record IdeState(
   public IEnumerable<Uri> GetDiagnosticUris() {
     return StaticDiagnostics.Keys.Concat(VerificationResults.Keys);
   }
+
```
robin-aws, line 164: I assume all the code from this point on is relocated rather than brand new? If not, how's the testing coverage?

Source/IntegrationTests/TestFiles/LitTests/LitTest/logger/ProofDependencyLogging.dfy
```diff
@@ -111,7 +111,7 @@
 // CHECK:       ProofDependencies.dfy\(337,12\)-\(337,16\): requires clause
 // CHECK:       ProofDependencies.dfy\(338,11\)-\(338,15\): ensures clause
 // CHECK:       ProofDependencies.dfy\(341,3\)-\(341,39\): function call result
-// CHECK:       ProofDependencies.dfy\(341,3\)-\(341,3\): function precondition satisfied
```
robin-aws, line 114: Not seeing why this location changed, was it actually incorrect before?

docs/dev/news/4798.feat
```diff
@@ -0,0 +1,4 @@
+- Fix: verification in the IDE no longer fails for iterators
+- Fix: the IDE now provides feedback when verification fails to run, for example due to a bad solver path
+- Fix: let the IDE correctly use the solver-path option when it's specified in a project file
+- Feat: improve the order of verification diagnostics emitted by the Dafny CLI, so that they now always follow the line order of the program.
```
robin-aws, line 4: :heart:

robin-aws (commented):
Last batch!

Source/DafnyDriver/CliCompilation.cs
```diff
@@ -0,0 +1,229 @@
...
+      new(factory.CreateLogger<Compilation>(), fileSystem,
+        new TextDocumentLoader(factory.CreateLogger<ITextDocumentLoader>(),
+          new DafnyLangParser(options, fileSystem, telemetryPublisher,
+            factory.CreateLogger<DafnyLangParser>(),
+            factory.CreateLogger<CachingParser>()),
+          new DafnyLangSymbolResolver(factory.CreateLogger<DafnyLangSymbolResolver>(), factory.CreateLogger<CachingResolver>(), telemetryPublisher)),
+        new DafnyProgramVerifier(factory.CreateLogger<DafnyProgramVerifier>()), engine, input);
+  }
+
+  public async Task VerifyAllAndPrintSummary() {
```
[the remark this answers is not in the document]
keyboardDrummer: In theory you can first create the Task, then call start, and then await the task, although I don't see why you would do that.

Source/DafnyDriver/CliCompilation.cs
```diff
@@ -0,0 +1,229 @@
...
+      }
+
+      var canVerifies = resolution.CanVerifies?.DistinctBy(v => v.Tok).ToList();
+
+      if (canVerifies != null) {
+        var orderedCanVerifies = canVerifies.OrderBy(v => v.Tok.pos).ToList();
+        foreach (var canVerify in orderedCanVerifies) {
+          canVerifyResults[canVerify] = new CliCanVerifyResults();
+          await Compilation.VerifyCanVerify(canVerify, false);
+        }
```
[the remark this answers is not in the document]
keyboardDrummer: I have to await before I do the next call, to ensure that the previous call has been added to any queues. Otherwise, the order in which things get verified becomes non-deterministic.

Source/DafnyLanguageServer/Workspace/IdeState.cs
keyboardDrummer: It is relocated, yes

robin-aws (approved):
A lot of work but well worth it, thanks!

PR #4800, opened 2023-11-17
Split `fresh` predicate to reduce brittleness

atomb:
### Description

When translating `fresh(e)` for `e` of type `set<T>`, Dafny previously generated a formula of the following form:

    forall $o: ref :: blah ==> $o != null && !old($Heap)[$o,alloc]

This was causing some seemingly trivial proofs not to go through.

Now it instead generates:

    (forall $o: ref :: blah ==> $o != null) &&
    (forall $o: ref :: blah ==> !old($Heap)[$o,alloc])

Each side of this conjunction has a separate trigger matching the shape of its consequent.

### How has this been tested?

The functionality was already well-tested. The fact that this passes the existing test suite shows evidence for its correctness. It also allows at least one verification from the [`smithy-dafny`](https://github.com/smithy-lang/smithy-dafny) repository to pass when it previously didn't.


MikaelMayer (approved):
Nice optimization. Indeed, it feels like being able to have two different triggers could help reduce brittleness, and you showed me that it has unnoticeable differences on codebases of interest.

PR #4807, opened 2023-11-20
Make customBoogie.patch work with recent versions

atomb:
The previous version of the patch would successfully apply, but would not build against a recent Boogie source tree.


customBoogie.patch
```diff
@@ -57,12 +57,22 @@ diff --git a/Source/DafnyCore/DafnyCore.csproj b/Source/DafnyCore/DafnyCore.cspr
...
 --- a/Source/DafnyCore/DafnyCore.csproj
 +++ b/Source/DafnyCore/DafnyCore.csproj
-@@ -31,5 +31,5 @@
+@@ -32,5 +32,15 @@
        <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
        <PackageReference Include="System.Runtime.Numerics" Version="4.3.0" />
        <PackageReference Include="System.Collections.Immutable" Version="1.7.0" />
 -      <PackageReference Include="Boogie.ExecutionEngine" Version="3.0.5" />
 +      <ProjectReference Include="..\..\boogie\Source\ExecutionEngine\ExecutionEngine.csproj" />
++      <ProjectReference Include="..\..\boogie\Source\BaseTypes\BaseTypes.csproj" />
```
keyboardDrummer, line 66: Why do we need to reference all of these instead of just ExecutionEngine?

atomb: I'm not 100% sure why they're necessary, but I've observed that local build fails without them but succeed with them.

keyboardDrummer: Are all of them necessary? I think we could prune this list

atomb: I created the list by adding dependencies that the build claimed to need until it succeeded. Local builds, at least for me, don't succeed with any of them removed.

keyboardDrummer: OK, sorry. I'd like to change Boogie at some point so the projects that Dafny conceptually does not use at all, like Houdini or Concurrency, don't need to be included.

PR #4815, opened 2023-11-22
feat: Concurrent collections

fabiomadge:
### How has this been tested?


Source/DafnyStandardLibraries/src/DafnyStdLibs_Concurrent.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def Get(self):
+        return self.boxed
+        
+    def Put(self, t):
+        self.boxed = t
+
+class Lock:
+    def __init__(self) -> None:
+        self.lock = threading.Lock()
```
robin-aws, line 80: How much overhead does this incur? It's good to have an actual lock by default, but there will also be lots of use cases where it's safe to assume Python is only single-threaded.

We can leave this out of scope for this PR, but I think it would be great to have some kind of configuration for multi-target Dafny projects to say "when translating to Python, make the concurrency datatypes no-ops" (and probably for other languages too)

Source/DafnyStandardLibraries/examples/ConcurrentExamples.dfy
```diff
@@ -0,0 +1,177 @@
...
+
+    method EverythingElse()
+      requires Valid()
+      modifies primary
+      ensures Valid()
+    {
+      assert primary.inv(0,0);
+      primary.Remove(0);
+      var keys := primary.Keys();
+      print keys, "\n";
```
robin-aws, line 87: We should replace the print statements with expects, since nothing looks at the output in this project.

Source/DafnyStandardLibraries/examples/ConcurrentExamples.dfy
```diff
@@ -0,0 +1,177 @@
+module ConcurrentExamples {
+
+  import opened DafnyStdLibs.Concurrent
+  import opened DafnyStdLibs.Wrappers
+  import opened Helpers
+
+  const p1: nat -> bool := _ => true
+  const p2: (nat, nat) -> bool := (_, _) => true
```
robin-aws, line 8: If it doesn't make verification a lot harder, we should make these not completely degenerate, to illustrate a bit more why it's useful to have an invariant.

Source/DafnyStandardLibraries/src/DafnyStdLibs/TargetSpecific/Concurrent-java.dfy
```diff
@@ -0,0 +1,54 @@
+module {:extern "DafnyStdLibsExterns.Concurrent"} DafnyStdLibs.Concurrent refines ConcurrentInterface {
+
+  class {:extern "MutableMap"} MutableMap<K(==), V(==)> ... {
+
+    constructor {:extern} {:axiom} (ghost inv: (K, V) -> bool)
```
robin-aws, line 5: Why was `{:axiom}` necessary everywhere here?

Source/DafnyStandardLibraries/src/DafnyStdLibs/TargetSpecific/Concurrent-js.dfy
```diff
@@ -0,0 +1,125 @@
+module DafnyStdLibs.Concurrent refines ConcurrentInterface {
+
+  class MutableMap<K(==), V(==)> ... {
+
+    var internal: map<K, V>
```
robin-aws, line 5: Ghost?

Source/DafnyStandardLibraries/src/DafnyStdLibs/TargetSpecific/ConcurrentInterface-notarget-java-cs-js-go-py.dfy
```diff
@@ -0,0 +1,73 @@
+abstract module DafnyStdLibs.ConcurrentInterface {
```
robin-aws, line 1: I think this is the right file to put documentation on the types and why their interfaces are what they are (which is counterintuitive for Dafny). Again I'm happy to provide text.

robin-aws (changes requested):
Great job getting all of this working - looks like even more fun than FileIO was. :)

One high-level comment on testing: it's worth covering multiple potential types for the type parameters, because there can easily be runtime issues with specific types. I'm aware of at least two gotchas:

1. On C# and Java IIRC, there are restrictions around using `null` for either a key or value.
2. For Go and JS, not all Dafny types will have a hashing function defined. That's why the runtime representations of `set` and `map` just using linear lookups instead of something based on hashtables. That means in this case types like `Concurrent.MutableMap<D, D>` where `D` is a `datatype` won't work correctly.

It might be tractable to raise an error on construction for types that aren't supported, if there's enough information in type descriptors. But at a minimum we should document these limitations and make and effort to find any others in tests.

Source/DafnyStandardLibraries/examples/ConcurrentExamples.dfy
```diff
@@ -0,0 +1,185 @@
...
+    assert mmap.inv(1,1);
+    mmap.Remove(1);
+    b := mmap.HasKey(0);
+    expect(b);
+    mmap.Remove(0);
+    b := mmap.HasKey(0);
+    expect(!b);
+  }
+
+  method {:test} TestCardinality() {
```
robin-aws, line 150: Copy and paste typo I think, doesn't test Cardinality

fabiomadge: Yes! Thanks!

Source/DafnyStandardLibraries/src/DafnyStdLibs/TargetSpecific/Concurrent-java.dfy
robin-aws: Ah right, because otherwise the library backend complains about externs

keyboardDrummer:
> I realized that the `AtomicBox` and `MutableMap` specifications still had non-empty `reads` and `modifies` clauses, which we didn't intend. Part of that was probably because we couldn't use the native Dafny implementation for Javascript otherwise. I fixed the specs and also added `{:concurrent}` to some of the test methods. That does mean we need another solution for Javascript - I got partway through compiling the Dafny and using it as an extern, it but I think it might be simpler to just implement it in Javascript instead.
> 
> The other problem I found/caused :) was that declaring `modifies {}` on these collections means having to add `reads-clauses-on-methods` to the source project file. But that means having to add the same option when using the standard libraries, since the doo file option check rejects mixing the value of those options. The best solution here might be to allow setting that attribute at the module level as well - @keyboardDrummer any chance that would be easy to add as we did for disabling nonlinear arithmetic?

Why is `reads-clauses-on-methods` an option instead of always being allowed? The default is backwards compatible anyways.

robin-aws:
Unfortunately not completely, there are function-by-methods that are valid without it but break when it is switched on.

robin-aws (approved):
Happy with this now that #4854 is in, and updated the TODO comment on examples/dfyconfig.toml accordingly (and other integration tests like `StandardLibrary.dfy` ensure that they can be loaded even when `reads-clauses-on-methods` is off in the consuming context).

Awesome work!

PR #4824, opened 2023-11-28
Fix: Ability to cast a datatype to its trait when overriding functions

### LINKED ISSUE #4823: New type system refresh issue
### Dafny version

latest-nightly

```dafny
trait Test<T> {
  function Cast(t: T): Test<T>
}

datatype Impl extends Test<Impl> = ImplConstructor()
{
  function Cast(t: Impl): Test<Impl> { t }
}
```

```code
Run this with

dafny verify --type-system-refresh --general-traits=datatype example.dfy
```

### What happened?

Unhandled exception: System.InvalidCastException: Unable to cast object of type 'Microsoft.Dafny.IndDatatypeDecl' to type 'Microsoft.Dafny.ClassLikeDecl
'.
   at Microsoft.Dafny.BoogieGenerator.AddFunctionOverrideReqsChk(Function f, BoogieStmtListBuilder builder, ExpressionTranslator etran, Dictionary`2 sub
stMap, Dictionary`2 typeMap) in dafny\Source\DafnyCore\Verifier\BoogieGenerator.Methods.cs:line 1185

Windows

### LINKED ISSUE #4989: Can't rebuild standard library files on Windows
Follow up from https://github.com/dafny-lang/dafny/pull/4824

The `check-binary` Make targets work by rebuilding the doo files, then comparing them to the checked in copies. They do this by unzipping and comparing the contents. Because the currently committed doo files are generally built on Macs, this means if a developer makes changes and rebuilds them on a Windows machine, the comparison fails because of the difference in line endings.

I propose fixing this by standardizing line endings inside doo files. This does imply altering program content when serializing to doo files, but this is already happening (merging multiple original source files into one, not yet maintaining comments or formatting of expressions in general) and needs to be allowed in general in order to optimize doo files in the future.

MikaelMayer:
This PR fixes #4823 (but not #4982 yet)

There were two issues there:
- The function override check was casting to ClassLikeDecl, when it should now be TopLevelDeclWithMembers as IndDatatypeDecl can also override trait declarations.
- The generated axiom used to rely on types being unboxed, but since now we might compare a trait with a datatype value, we need to ensure correct boxing for Boogie to type-check.

I also ensured that the generation of doo files only use \n lines when creating the source (fixes #4989)


Source/DafnyCore/Verifier/BoogieGenerator.Functions.cs
```diff
@@ -569,6 +569,8 @@ public partial class BoogieGenerator {
     }
   }
 
+  private int Cutoff = 0;
```
RustanLeino, line 572: It seems this field is not used.

RustanLeino (changes requested):
Looks good, except for the unused declared of `Cutoff`.

RustanLeino (dismissed):

RustanLeino (approved):
It seems issue 4823 was fixed between Dafny 4.4 and 4.5.

MikaelMayer:
Indeed! The fix itself I had originally published got ousted in the merge commit, but at least this PR adds a test.

PR #4826, opened 2023-11-29
Let more tests use the new CLI

keyboardDrummer:
### Description
- Replace %dafny /compile:0 "%s" with %verify "%s"
- TODO

### How has this been tested?
This is a change to tests


jtristan (commented):
Any reason why we're doing this for "most" verification tests and not all of them?

keyboardDrummer:
> Any reason why we're doing this for "most" verification tests and not all of them?

Some rely on features that don't exist in the new CLI. An example is verification snapshots.

jtristan (approved):

PR #4832, opened 2023-12-01
Different uri for doo file

keyboardDrummer:
### Description
- Prefix Uris to .doo files with the `doo://` scheme. This can be used by the LSP client to implement custom behavior for opening these files

### How has this been tested?
- Added the test StandardLibraryTest.GotoDefinition


Source/DafnyCore/DafnyFile.cs
```diff
@@ -23,6 +24,13 @@ public class DafnyFile {
 
   public static DafnyFile CreateAndValidate(ErrorReporter reporter, IFileSystem fileSystem,
     DafnyOptions options, Uri uri, IToken origin) {
+
+    if (uri.Scheme == "doo") {
```
keyboardDrummer, line 28: VSCode only allows you to have custom handlers for custom schemes, which seems fair to me, but it means we have to introduce a `doo:` scheme, which is at odds with that we already handle the .doo extension as being special, even if no `doo:` scheme is used. That's why the code looks a bit funky now.

If we were more strict, only allowing doo files being used as libraries, and only allowing dfy files being used as sources, then we wouldn't need to rely on the extension checking, but that seems like something to consider for the future.

Source/DafnyCore/DafnyFile.cs
```diff
@@ -16,13 +18,30 @@ public class DafnyFile {
   public string BaseName { get; private set; }
   public bool IsPreverified { get; set; }
   public bool IsPrecompiled { get; set; }
-  public bool IsPrerefined { get; private set; }
```
robin-aws, line 19: Off topic but when did this appear and what was it used for?

Source/DafnyCore/DooFile.cs
```diff
@@ -131,7 +130,7 @@ public class ManifestData {
     }
 
     var success = true;
-    var relevantOptions = currentCommand.Options.ToHashSet();
+    var relevantOptions = options.Options.OptionArguments.Keys.ToHashSet();
```
robin-aws, line 133: Is this still correct, given the comment immediately below? i.e. is `OptionArguments.Keys` the set of valid parameters for the current command, or just the options explicitly provided on the CLI or in project files?

Source/DafnyCore/DafnyFile.cs
```diff
@@ -16,13 +18,30 @@ public class DafnyFile {
...
   public Func<TextReader> GetContent { get; set; }
-  public Uri Uri { get; }
+  public Uri Uri { get; private set; }
   [CanBeNull] public IToken Origin { get; }
 
+  private static readonly Dictionary<Uri, Uri> ExternallyVisibleEmbeddedFiles = new();
+
+  public static Uri ExposeInternalUri(string externalName, Uri internalUri) {
+    var externalUri = new Uri("dafny:" + externalName);
```
robin-aws, line 28: Why not just use `$"dafny:{internalUri}"`? That would be simpler and leaves the door open for other schemes. AFAICT attaching a fake Dafny file name doesn't help the IDE but perhaps I'm missing something.

robin-aws (commented):
Thanks for the change, I definitely like the `dafny:` version better. Just some questions.

Source/DafnyCore/DafnyFile.cs
keyboardDrummer: Because we expose the externalUri and the internalUri contains implementation details. Is there are a downside to having this mapping, except for it being slightly more work?

Source/DafnyCore/DooFile.cs
keyboardDrummer: This is still correct, yes

Source/DafnyCore/DafnyFile.cs
keyboardDrummer: It was needed when .doo files still contained resolved code instead of just parsed code.

Source/DafnyCore/DafnyFile.cs
robin-aws: No real objection to the complexity/work, but a little worried about exposing "DafnyStandardLibraries.dfy" when that string/file name isn't actually meaningful. Eventually I expect the IDE will be able to actually open the doo file, and we shouldn't hide the fact that the source is coming from a doo file (especially to be clear it's read-only)

keyboardDrummer: Actually I added the dot dfy at the end so the IDE can recognize it's Dafny code and provide ide tooling on it.

I think from the user perspective, opening a dotdfy file to view the sources inside a dotdoo file, seems good

robin-aws: Okay, I can live with it :)

robin-aws (approved):

Source/DafnyCore/DafnyFile.cs
keyboardDrummer: We can still change the part after the `dafny:` in a follow-up PR, so let's not get hung up on this

I think, as part of adding `.doo` files, we should have added IDE navigation for .doo files as well, and then we would have already solved this. I think it would help with correctly designing `.doo` files as well. For example, we could have considered not making it dotdoo a compressed file and instead having it be a single text file, so the IDE can open it directly, without the need for a `dafny doo source` command. Not saying we should have gone for that option, but at least we would have had a reason to consider it.

Thinking about dotdoo support in the IDE more, and given the way dotdoo files are currently structured, the `dafny doo source` command should support both on-disk files and files internal to the CLI, so it does need to take a URI, and then the `dafny:` uris exposed to the IDE should probably contain a URI after the `dafny:`. The IDE will chop off the `dafny:` part and pass the rest to `dafny doo source`

Source/DafnyCore/Rewriters/PrecedenceLinter.cs
```diff
@@ -5,21 +5,18 @@
...
 
 namespace Microsoft.Dafny {
 
   public class PrecedenceLinter : IRewriter {
-    private CompilationData compilation;
+    private readonly CompilationData compilation;
     // Don't perform linting on doo files in general, since the source has already been processed.
     internal override void PostResolve(ModuleDefinition moduleDefinition) {
-      if (moduleDefinition.tok.Uri != null && moduleDefinition.tok.Uri.LocalPath.EndsWith(".doo")) {
+      if (moduleDefinition.tok.Uri != null && !moduleDefinition.ShouldVerify(compilation)) {
```
robin-aws, line 19: For the record I think this is a step backwards in terms of accuracy: not linting doo files is more about the fact that the content is read-only to consumers and not their responsibility to lint, nothing to do with verification.

I don't want to block this change on it, but perhaps we could at least add a comment in the next related PR.

robin-aws (approved):

Source/DafnyCore/DafnyFile.cs
robin-aws: > For example, we could have considered not making dotdoo a compressed file and instead having it be a single text file, so the IDE can open it directly, without the need for a dafny doo/library source command.

I definitely intended to iterate on the file format as we added features, and hence why the manifest has a `doo_file_version` property. :) But I don't think I'd want to make it a single text file, since then the manifest information would have to be embedded in comments or something like that, which would make it a bit tougher to read robustly and harder to hide that implementation detail from users.

> Thinking about dotdoo support in the IDE more, and given the way dotdoo files are currently structured, the dafny doo/library source command should support both on-disk files and files internal to the CLI, so it does need to take a URI, and then the dafny: uris exposed to the IDE should probably contain a URI after the dafny:. The IDE will chop off the dafny: part and pass the rest to dafny doo source

Yup, this is one of the reasons I thought `$"dafny:{internalUri}"` would be smart, but you expressed it better than I managed at the time. :)

Source/DafnyCore/Rewriters/PrecedenceLinter.cs
keyboardDrummer: I'm wondering whether `ShouldVerify` should be renamed to `ShouldCheck`.

The reasoning behind not verifying a file seem to apply to other correctness checks as well.

PR #4837, opened 2023-12-01
Chore: Explain how to reason with functions by methods

MikaelMayer:
### How has this been tested?
I added the tests files for each Dafny entry


docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,89 @@
+---
+title: I can't prove the equivalence between the method part of a function by method and the function itself
```
codyroux, line 2: should this be "the method part of a `function by method`"?

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,89 @@
...
+You might be tempted to immediately write the following equivalence,
+with your best guess of an invariant
+and a hint at the end to help Dafny realize that the result is correct.
+However you cannot prove the invariant you just wrote.
+
+```
+{% include_relative FAQFunctionByMethodProof1.dfy %}
+```
+
+Let's step back for a moment. Let's unroll the computation of the function on a sequence of 3 elements.
```
codyroux, line 22: I'd remove "let's step back for a moment".

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,89 @@
...
+The function will compute:
+
+`s[0] + (s[1] + (s[2] + 0))`
+
+But the loop will compute
+
+`((0 + s[0]) + s[1]) + s[2]`
+
+In the function, we add the first element to the sum of the remaining.
+In the by method, we add the last element to the sum of the initial terms.
```
codyroux, line 32: (to the accumulator)

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,89 @@
...
+Let's step back for a moment. Let's unroll the computation of the function on a sequence of 3 elements.
+The function will compute:
+
+`s[0] + (s[1] + (s[2] + 0))`
+
+But the loop will compute
+
+`((0 + s[0]) + s[1]) + s[2]`
+
+In the function, we add the first element to the sum of the remaining.
```
codyroux, line 31: (to the result)

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,89 @@
...
+## Improved invariant
+
+You can change the invariant to be `invariant result + Sum(s[i..]) == Sum(s)`, as below:
+
+```
+{% include_relative FAQFunctionByMethodProof2.dfy %}
+```
+
+This works because, with `result'` being the result at the end of the loop and `+` being associative,
+the lemma `IntermediateProperty()` demonstrates the underlying reasoning that helps prove that the invariant
```
codyroux, line 50: "demonstrates" is a frenchism, I suspect. "Proves" or "shows" probably.

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,89 @@
...
+is indeed invariant. Dafny can prove it automatically, so the lemma is not needed in the loop body.
+And the nice thing is that we can get rid of the final hint.
```
codyroux, line 52: Avoid starting a sentence with "And".

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,89 @@
...
+In the function, we add the first element to the sum of the remaining.
+In the by method, we add the last element to the sum of the initial terms.
+If it was not the case that the addition was associative and 0 was a neutral element, there would be no immediate way to prove that the two computations are equivalent.
+
+There are essentially three solutions around that:
+
+* Improve the invariant, knowing `+` is associative
+* Change the order of computation of the function so that it matches the by method
+* Change the order of computation of the by method so that it matches the function
+
```
codyroux, line 40: One can imagine a more "intrusive" solution: make the function tail-call, and remove the while loop altogether.

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,89 @@
...
+addition that will be computed by the method. As mentioned previously, the first addition is the one performed at the bottommost level of recursion: the first addition is `s[|s|-1] + 0`,
+so the loop has to actually be in reverse, like this:
+
+```
+{% include_relative FAQFunctionByMethodProof4.dfy %}
+```
+
+Note that this approach results in the smallest invariant that actually closely matches the functio itself.
+Also, instead of adding to the right of `result`, adding to the left ensures the same order is kept,
+in case the operation was not commutative (so that it would work for sequence append)
```
codyroux, line 89: sequence append -> the sequence append operation

docs/HowToFAQ/FAQFunctionByMethodProof2.dfy
```diff
@@ -0,0 +1,27 @@
...
+  requires result + Sum(s[i..]) == Sum(s)
+  ensures result' + Sum(s[i+1..]) == Sum(s)
+{
+  calc {
+        result + Sum(s[i..])             == Sum(s);
+  <==>  result + (s[i]  + Sum(s[i+1..])) == Sum(s); // Definition of Sum()
+  <==> (result +  s[i]) + Sum(s[i+1..])  == Sum(s); // Associativity
+  <==>  result'         + Sum(s[i+1..])  == Sum(s); // Definition of result'
+  }
+}
```
codyroux, line 27: Newline?

docs/HowToFAQ/FAQFunctionByMethodProof3.dfy
```diff
@@ -0,0 +1,17 @@
...
+  result := 0;
+  for i := 0 to |s|
+    invariant result == Sum(s[0..i])
+  {
+    assert s[0..i+1][0..i] == s[0..i];
+     // Minimum hint needed for proof
+    result := result + s[i];
+  }
+  assert s[0..|s|] == s;
+} // One last hint
```
codyroux, line 17: Newline etc.

docs/HowToFAQ/FAQFunctionByMethodProof2.dfy
MikaelMayer: That would add an extra newline in the rendered block of code that is not necessary.

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -61,17 +63,13 @@ be the first addition performed by the function, it has to be at the bottommost
...
 ```
 {% include_relative FAQFunctionByMethodProof3.dfy %}
 ```
 
-As you can see, it requires some proof hint so that it works.
-Many times, however, it's not practical to change the function because it might change proofs that
-depend on that function's body.
-Moreover, most of the time, the goal of a function by method body is to give a more efficient implementation
-of a function, and the method body itself will be opaque.
+Not only this approach requires more proof hint, but it might also break the proofs that dependend on the shape of the function, while everything we wanted in the first place was to give a more efficient implementation to the function.
```
codyroux, line 72: nit: Not only does this approach require more proof hints, but...

codyroux: while everything we wanted -> while all we wanted

docs/HowToFAQ/onepage.md
```diff
@@ -2355,9 +2357,9 @@ You can change the invariant to be `invariant result + Sum(s[i..]) == Sum(s)`, a
 ```
 
 This works because, with `result'` being the result at the end of the loop and `+` being associative,
-the lemma `IntermediateProperty()` demonstrates the underlying reasoning that helps prove that the invariant
+the lemma `IntermediateProperty()` shows the underlying reasoning that helps prove that the invariant
 is indeed invariant. Dafny can prove it automatically, so the lemma is not needed in the loop body.
-And the nice thing is that we can get rid of the final hint.
+Ohe nice thing is that we can get rid of the final hint.
```
codyroux, line 2362: Typo: Ohe -> One

docs/HowToFAQ/FAQFunctionByMethodProof.md
```diff
@@ -0,0 +1,87 @@
...
+Let's unroll the computation of the function on a sequence of 3 elements.
+The function will compute:
+
+`s[0] + (s[1] + (s[2] + 0))`
+
+But the loop will compute
+
+`((0 + s[0]) + s[1]) + s[2]`
+
+In the function, we add the first element to the result of computing the sum of the remaining elements.
```
ssomayyajula, line 31: I would indicate that the first element is `s[0]`, the "sum of the remaining elements" is `s[1] + (s[2] + 0)`, the last element is `s[2]`, and the "sum of the initial terms" is `(0 + s[0]) + s[1]` so that the reader can connect the text explanation to the source code.

docs/HowToFAQ/FAQFunctionByMethodProof2.dfy
```diff
@@ -0,0 +1,27 @@
...
+  }
+} // No hint needed at the end
+
+lemma IntermediateProperty(result: int, result': int, i: int, s: seq<int>)
+  requires 0 <= i < |s|
+  requires result' == result + s[i]
+  requires result + Sum(s[i..]) == Sum(s)
+  ensures result' + Sum(s[i+1..]) == Sum(s)
+{
+  calc {
```
ssomayyajula, line 21: Do you think the proof is easier to read this way?

```dafny
  calc {
    Sum(s);
    result  + Sum(s[i..]);
    result  + (s[i]  + Sum(s[i+1..])); // Definition of Sum()
    (result +  s[i]) + Sum(s[i+1..]);  // Associativity
    result'          + Sum(s[i+1..]);  // Definition of result'
  }
```

docs/HowToFAQ/onepage.md
```diff
@@ -2307,6 +2307,92 @@ method mmm<K(==),V(==)>(m: map<K,V>, k: K, v: V) {
...
+Let's unroll the computation of the function on a sequence of 3 elements.
+The function will compute:
+
+`s[0] + (s[1] + (s[2] + 0))`
+
+But the loop will compute
+
+`((0 + s[0]) + s[1]) + s[2]`
+
+In the function, we add the first element to the result of computing the sum of the remaining elements.
```
ssomayyajula, line 2339: Same comment as in the original text.

ssomayyajula (commented):
Just a few prose/code style comments - otherwise LGTM.

docs/HowToFAQ/onepage.md
MikaelMayer: This is generated text so don't worry

ssomayyajula (approved):

PR #4841, opened 2023-12-04
Use {:disable_nonlinear_arithmetic} to simplify standard library

keyboardDrummer:
### Description
- Use `{:disableNonlinearArithmetic}` on modules to simplify the definition and usage of the standard library
- Use `{:z3ArithmeticSolver 6}` on some modules to let their verification pass more reliably

### How has this been tested?
- This is a refactoring so no addition tests were needed


atomb (changes requested):
I love this. The simplification is substantial. Before we merge it, though, we should figure out how to remove the `{:axiom}` annotations and make the other failing tests pass.

keyboardDrummer:
> Before we merge it, though, we should figure out how to remove the `{:axiom}` annotations and make the other failing tests pass.

If the SMT solver's non linear arithmetic support is not doing the definitional proofs of the operators for us, then what else can we do but add definitional axioms for these operators?

keyboardDrummer:
Comparing bruned Boogie files doesn't reveal any reason why the the NLA lemma's would suddenly fail.

Looking a the suddenly failing lemma `FundamentalDivMod`, and the pruned Boogie code that's sent to the solver.

Here it is with the change from this PR, causing the lemma not to verify:
[LemmaFundamentalDivModPrunedNoPass.txt](https://github.com/dafny-lang/dafny/files/13550306/LemmaFundamentalDivModPrunedNoPass.txt)

If I add `includes = ["**/*/ModInternalsNonlinear.dfy"]` to the `dfyconfig.toml`, I get an almost exactly the same pruned .bpl file, but verification will fail:
[LemmaFundamentalDivModPrunedPasses.txt](https://github.com/dafny-lang/dafny/files/13550307/LemmaFundamentalDivModPrunedPasses.txt)

The only difference is in the generated numbers.

atomb:
This might be a case where it's worth exposing another use of the `:smt_option` attribute added to Boogie to support `:disable_nonlinear_arithmetic`. I've confirmed that all the definitions marked `{:axiom}` in your branch verify successfully with`smt.arith.solver=6`. It's something I hope most users wouldn't have to do themselves, but for the standard library it seems reasonable.

keyboardDrummer:
> This might be a case where it's worth exposing another use of the `:smt_option` attribute added to Boogie to support `:disable_nonlinear_arithmetic`. I've confirmed that all the definitions marked `{:axiom}` in your branch verify successfully with`smt.arith.solver=6`. It's something I hope most users wouldn't have to do themselves, but for the standard library it seems reasonable.

Works, thanks

Source/DafnyCore/AST/Members/Method.cs
```diff
@@ -46,7 +46,7 @@ public class Method : MemberDecl, TypeParameter.ParentType,
       yield return a;
     }
 
-    if (Body is null && HasPostcondition && EnclosingClass.EnclosingModuleDefinition.ModuleKind == ModuleKindEnum.Concrete && !HasExternAttribute) {
+    if (Body is null && HasPostcondition && EnclosingClass.EnclosingModuleDefinition.ModuleKind == ModuleKindEnum.Concrete && !HasExternAttribute && !HasAxiomAttribute) {
```
atomb, line 49: This change seems unrelated to the others. Was it intentionally included?

Source/DafnyStandardLibraries/src/DafnyStdLibs/Arithmetic/Internal/DivInternalsNonlinear.dfy
```diff
@@ -6,7 +6,14 @@
...
  *******************************************************************************/
 
-module DafnyStdLibs.Arithmetic.DivInternalsNonlinear {
+/**
+Several lemmas in this module are bodyless and marked as axiom.
+These lemmas could previously be proven with an empty body, but only if this file was verified by itself,
+not with the rest of the standard library.
+If we want to remove the {:axiom} annotation from these lemmas, 
+we will have to investigate why they only verify under specific conditions.   
+*/
```
atomb, line 15: This comment is obsolete now. You could comment on the arithmetic solver setting, though.

Source/DafnyStandardLibraries/src/DafnyStdLibs/Arithmetic/Internal/ModInternalsNonlinear.dfy
```diff
@@ -6,7 +6,14 @@
...
  *******************************************************************************/
 
-module DafnyStdLibs.Arithmetic.ModInternalsNonlinear {
+/**
+Several lemmas in this module are bodyless and marked as axiom.
+These lemmas could previously be proven with an empty body, but only if this file was verified by itself,
+not with the rest of the standard library.
+If we want to remove the {:axiom} annotation from these lemmas, 
+we will have to investigate why they only verify under specific conditions.   
+*/
```
atomb, line 15: Another obsolete comment.

Source/IntegrationTests/TestFiles/LitTests/LitTest/verification/nonLinearArithmetic.dfy
```diff
@@ -28,7 +28,7 @@ module Power0 {
 
 }
 
-module {:disable_nonlinear_arithmetic} Power1 {
+module {:disable-nonlinear-arithmetic} Power1 {
```
atomb, line 31: I think this should be `disableNonlinearArithmetic`.

atomb (changes requested):
Thanks for adding both new attributes! I noticed a few small inconsistencies and obsolete comments, but otherwise I think it looks good.

As to the final test failures, I've noticed that using `--boogie-filter "*LemmaModAuto*"` allows me to reproduce them locally (similarly for `DivAuto`). Both lemmas are kind of enormous. My inclination would be to take each of the conjuncts in the `DivAuto` and `ModAuto` predicates, prove each one in a completely separate lemma, and then call that collection of lemmas (and nothing else) from the top-level lemma.

Source/DafnyCore/AST/Members/Method.cs
keyboardDrummer: Yes, although it could be removed.

Source/IntegrationTests/TestFiles/LitTests/LitTest/auditor/TestAuditor.dfy-ietf.md.expect
```diff
@@ -10,8 +10,6 @@
 
 * Declaration has explicit `{:axiom}` attribute. SHOULD provide a proof or test.
 
-* Ghost declaration has no body and has an ensures clause. MUST provide a body or add `{:axiom}`.
-
```
atomb, line 14: This seems like something we want to continue reporting in the auditor.

keyboardDrummer: But it has `{:axiom}` on it. The comment seemed incorrect.

atomb: Ah! Got it. I didn't look at the source code it was reporting on. :)

atomb (dismissed):

atomb:
Looks like there are still some test failures.

robin-aws (approved):
Previously approved (but also wanted to say big thanks for this, it cleans up the source very nicely!)

keyboardDrummer:
> Previously approved (but also wanted to say big thanks for this, it cleans up the source very nicely!)

Thanks!

Next up, getting rid of all the target specific .doo files

PR #4843, opened 2023-12-04
fix: The Python compiler emits reserved names for datatypes

### LINKED ISSUE #4830: [Python] Dafny can generate invalid class names for datatypes with reserved keywords
### Dafny version

4.3.0

```dafny
module AnyModule {

    datatype None = | None ()  
    
    method anyMethod() {
        var a := None();
    }

}
```

```code
dafny -compileTarget:py Index.dfy

*** Error compiling './any_module/Index-py/AnyModule.py'...
  File "./any_module/Index-py/AnyModule.py", line 24
    class None:
          ^^^^
SyntaxError: invalid syntax
```

### What happened?

Dafny generates a class `None` for the `datatype None`. `None` is a reserved keyword in Python, so this class is invalid.

I noticed that Dafny seems to already handle the reserved keyword elsewhere in the file, so I suspect it should handle reserved keywords in the class definition as well.

Later in the generated file, Dafny appends an underscore to the reserved keyword:

```
class None_None(None, NamedTuple('None_', [])):
```

Here, `None_` inside the `NamedTuple` has an underscore. Dafny generates an underscore here for reserved keywords (True, False, del, etc.), but does not if this is not a reserved keyword.

Given this, I would expect that the underscore would also be appended to the class definition and any generated references to that class.

Mac

fabiomadge:
Fixes #4830


docs/dev/news/4843.fix
```diff
@@ -0,0 +1 @@
+Python compiler emits reserved names for datatypes
```
robin-aws, line 1:
```suggestion
The Python compiler no longer emits reserved names for datatypes
```
?

Source/IntegrationTests/TestFiles/LitTests/LitTest/git-issues/git-issue-4830.dfy
```diff
@@ -0,0 +1,3 @@
+// RUN: %testDafnyForEachCompiler "%s"
+
+datatype None = None
```
robin-aws, line 3: How the heck did the standard libraries not trigger this?

docs/dev/news/4843.fix
robin-aws: Okay sure. It hasn't been our pattern, but since it ends up under a "Bug fixes" header in the release notes this style works too.

robin-aws (approved):

PR #4851, opened 2023-12-06
More granular proof dependencies

### LINKED ISSUE #4719: Dafny reports a warning on an `assert` statement, saying it is an unnecessary assumption
Dafny reports a warning on an `assert` statement, saying it is an unnecessary assumption.

It's confusing that Dafny calls an assert statement an assumption, especially Dafny also has assume statements. I suggest changing the text to "unnecessary assertion".

Alternatively Dafny could be more verbose and say that this assert statement was not required to prove other assert statements or ensures clauses.

atomb:
# Description

This PR refactors the proof dependency data types and the warning generation code to do several things:

* Distinguish between the various types of assumption that can show up. We currently report redundancies only for assumptions that come directly from assume and assert statements and for requires clauses. Other types are difficult for the user to control, but will likely be useful to treat separately in the future.
* Distinguish between the split parts of `requires` and `ensures` clauses.
* Clarify that unnecessary assumptions and assertions may be only partly unnecessary. A next step will be to indicate which part is unnecessary, but that turns out to require changing verification conditions in a way that may affect performance and brittleness, so I'm going to leave that for a separate PR that I hope to land after we release 4.4.

Fixes #4719 

# How has this been tested?
This changes the expected output of a few existing tests.


Source/DafnyCore/Verifier/ProofDependency.cs
```diff
@@ -162,24 +168,40 @@ public class CallDependency : ProofDependency {
   }
 }
 
+public enum AssumptionForm {
```
keyboardDrummer, line 171: This enum barely seems used. The only check on its value I saw was

```
if (assumeDep.Form != AssumptionForm.AssumeStatement && assumeDep.Form != AssumptionForm.AssertStatement)
```

In which case I think you could have used a bool instead of an enum. I imagine you have future intentions for this enum but I suggest only writing code that is useful for the current set of tests, and adding more later.

Also, the enum seems to mirror fields of AST node types, so consider using references to those instead of an enum, for example like this:

```c#
interface IHasAssumptions : INode { 
  public X AnswerQuestionAboutAssumptionField(string fieldName);
}

class ComprehensionExpr : IHasAssumptions { 
  public X AnswerQuestionAboutAssumptionField(string fieldName) {
    if (fieldName == nameof(Range)) { 
      ..
    }
  }
}

record Assumption(IHasAssumptions Container, string FieldName)
```

Source/DafnyCore/ProofDependencyWarnings.cs
```diff
@@ -135,6 +135,10 @@ public class ProofDependencyWarnings {
           lit == true) {
         return false;
       }
+
+      if (assumeDep.Form != AssumptionForm.AssumeStatement && assumeDep.Form != AssumptionForm.AssertStatement) {
```
keyboardDrummer, line 139: Generally I'm wary of code that is in a generic location (unused specification warnings) and points to specific locations (specific assumption form enum values, which map to AST node types).

An alternative implementation would be to put the knowledge of whether these forms can generate warnings in the various assumption forms, and then this code would be `if (!assumeDep.Form.IsExplicit) { return false; }` 

I've randomly picked the name `IsExplicit`, since I don't know what the condition is upon which you decide which forms should return false here. That you can decide on this name would be another advantage of this code style: it explains better what the property is that makes you decide to return false.

Source/DafnyCore/Verifier/ProofDependency.cs
atomb: It's true, that enum is defined as it is in preparation for broader use in the future. I don't think a boolean would be appropriate here, because there are a number of different decisions I expect we may make based on the type of assumption in the future, and that would lead to the need for substantial refactoring in the future.

I would leave out the unused alternatives in the assumption for now except that it needs to be filled in at every creation site. Given that there is an `Other` alternative, I could leave out all of the other alternatives that are (currently) never checked and use `Other` for all of them.

Source/DafnyCore/ProofDependencyWarnings.cs
atomb: Yeah, I think putting the check as a member of the `AssumptionForm` type would be a distinct improvement. Good catch.

Source/DafnyCore/Verifier/ProofDependency.cs
atomb: It's probably also worth noting that, although these are very similar to AST nodes, they don't map directly. For example, `ForallEnsures` and `ForallRange` describe aspects of a `forall` statement that are, themselves, only `Expression` values. If we were to use the `ForallStmt` node, we wouldn't know which part of it this assumption referred to. If we were to use the ensures or range expressions, we wouldn't know where those expressions came from.

Source/DafnyCore/ProofDependencyWarnings.cs
atomb: I ended up putting it in `AssumptionDependency`, which also seems like a reasonable location, since `enum` types in C# can't have members. (You can use extension methods, but that seems messier).

Source/DafnyCore/Verifier/ProofDependency.cs
MikaelMayer: I'm fine with an enum, it's like a named boolean. Thanks for explaining that.

MikaelMayer (commented):
I like this PR, but I would love to see a test or two about the behavior described in the PR description, namely assertions that are unused. With this PR, I would have expected some Dafny file to be updated to add a useless assertion that would not trigger anything, right?

atomb:
> I like this PR, but I would love to see a test or two about the behavior described in the PR description, namely assertions that are unused. With this PR, I would have expected some Dafny file to be updated to add a useless assertion that would not trigger anything, right?

This PR doesn't change _what_ unused things are reported, just _how_ they're reported. If you look at the very last line of the PR diff, you'll see that a message that previously referred to an assumption now refers to an assert statement (which gives rise to an assumption).

Source/DafnyCore/Verifier/ProofDependency.cs
keyboardDrummer: > It's probably also worth noting that, although these are very similar to AST nodes, they don't map directly. For example, ForallEnsures and ForallRange describe aspects of a forall statement that are, themselves, only Expression values. If we were to use the ForallStmt node, we wouldn't know which part of it this assumption referred to. If we were to use the ensures or range expressions, we wouldn't know where those expressions came from.

I agree, that's why I said the enum values seems to mirror _fields_ of AST node types, and why the record I suggested 
```
record Assumption(IHasAssumptions Container, string FieldName)
```
includes a FieldName property.


> I don't think a boolean would be appropriate here, because there are a number of different decisions I expect we may make based on the type of assumption in the future, and that would lead to the need for substantial refactoring in the future.

Given the "I expect", are you saying you do think a boolean would be appropriate right now?

What do you mean with substantial refactoring? Changing a boolean into something else? What's substantial about that? Writing code is cheap. However, how can I review code that is meant to do things not specified or tested in the current PR? And how will you ask a future reviewer to review the whole set of changes that were accumulated of several PRs, once it has finally reached a state where each piece is fully used?

atomb:
> > > I like this PR, but I would love to see a test or two about the behavior described in the PR description, namely assertions that are unused. With this PR, I would have expected some Dafny file to be updated to add a useless assertion that would not trigger anything, right?
> > 
> > 
> > This PR doesn't change _what_ unused things are reported, just _how_ they're reported. If you look at the very last line of the PR diff, you'll see that a message that previously referred to an assumption now refers to an assert statement (which gives rise to an assumption).
> 
> Oh I see. I misread the description, but the last line is what I was looking for anyway; Thanks for pointing it out. I'm a bit worried about the "false negative" aspect of recognizing assertions might be partly useless. Is it meaning that there exists at least a conjunct that is not used in the following proofs? If there is only exactly one conjunct, will we still have this "partly" wording or will it be removed? If the answer to the first one is yes but the second one is no, then I'd suggest considering being more precise in the message and say something like: "Warning: At least one conjunct of this assertion/assumption is useless" in the first case and, if we can know whether there is exactly one conjunct, have the optimized message as before "Warning: this assertion/assumption is useless" or something like that.

Yeah, I agree. I have code that allows the "partly" to be removed, but it also changes what Boogie we generate (because that's the only way to track it precisely), so it may have an impact on brittleness that still needs to be evaluated.

MikaelMayer (approved):

Source/DafnyCore/Verifier/ProofDependency.cs
atomb: The "substantial refactoring" I mentioned has to do with putting behavior in the AST nodes now and moving it back out later, not just changing the type of a field. But I think that may have been a misunderstanding of what you were suggesting. For the moment, I'd be happy to move to `WarnUnused` and `IgnoreUnused`, or something like that.

keyboardDrummer: > For the moment, I'd be happy to move to WarnUnused and IgnoreUnused, or something like that.

Thanks !

keyboardDrummer (approved):

PR #4854, opened 2023-12-07
Resolve doo file with its options

keyboardDrummer:
### Description
- Parse and resolve the contents of doo files using the options that the doo file was created with. This is safe because of the existing doo file validation
- The code used for this enables overriding arbitrary options for particular modules. We could see for which options this is useful and add syntax for doing so

### How has this been tested?
- Added the test `useReadsOnMethodsLibrary.dfy`


robin-aws (approved):
🚀 

A few comments but nothing blocking.

PR #4855, opened 2023-12-07
Code navigation when resolution errors

keyboardDrummer:
### Description
- Allows code navigation to work even when there are resolution errors, but only if there are no resolution errors in module dependencies.

### How has this been tested?
- Added several XUnit tests


stefan-aws (approved):
I did some initial (successful) tests that involved "go to definition". This is very helpful, thanks.

PR #4867, opened 2023-12-11
feat: Add a few more Std.Collections.Seq functions

robin-aws:
### Description
These are functions I got used to having around in [THIS "standard library"](https://github.com/aws/aws-cryptographic-material-providers-library-dafny/tree/main/StandardLibrary) and didn't spot they weren't in OUR standard library until now. :)

Also removed `Std.Strings.Concat` (already covered more generically with `Std.Collections.Seq.Flatten`) and moved `Std.Strings.Join` to `Std.Collections.Seq`: these were only referenced by tests and their existence contradicted the documentation on the purpose of the `Strings` library.

Also commented out the remaining print statements in JSONTests.dfy to keep the testing output cleaner.

### How has this been tested?
DafnyStandardLibraries tests.


Source/DafnyStandardLibraries/src/Std/Collections/Seq.dfy
```diff
@@ -282,12 +282,22 @@ module Std.Collections.Seq {
     ensures if o.Some? then o.value < |xs| && xs[o.value] == v &&
                             forall j {:trigger xs[j]} :: 0 <= j < o.value ==> xs[j] != v
             else v !in xs
+  {
+    IndexByOption(xs, x => x == v)
+  }
+
+  /* Returns Some(i), if an element satisfying p occurs at least once in a sequence, and i is
+    the index of the first such occurrence. Otherwise the return is None. */
+  function {:opaque} IndexByOption<T(==)>(xs: seq<T>, p: T -> bool): (o: Option<nat>)
```
robin-aws, line 291: I'd also be happy with `IndexWithOption` as a naming scheme.

Source/DafnyStandardLibraries/Makefile
```diff
@@ -8,7 +8,8 @@ DOO_FILE_TARGET=binaries/DafnyStandardLibraries.doo
 all: check-binary test check-format check-examples
 
 verify:
-	$(DAFNY) verify src/DafnyStdLibs/dfyconfig.toml
```
atomb, line 11: Was this just missed in the `DafnyStdLibs` -> `Std` renaming?

atomb (approved):
This is a nice improvement! I like having those functions generic to `seq<T>` rather than specialized to `string`.

Source/DafnyStandardLibraries/Makefile
robin-aws: Yup, forgot to call it out in the PR description

PR #4876, opened 2023-12-13
Replaceable module separate compilation

keyboardDrummer:
### Description
- Make separate compilation work reliably for replaceable modules

### Caveats
- There is a still a check that should be added but is missing in this PR, which is to emit an error if an `{:extern}` attribute is used to overwrite the compiled name of a replaced definition. I would've added that check to this PR but I'm saving it for the future since I'm not yet sure what the best way is.
- This PR adds the restriction that you can't overwrite names using `{:extern}` inside modules that replace another module. This takes some of the fun out of using replaceable modules, since now users may have to create not one but two modules in their target specific code: 
  - one that uses `{:extern}` with the desired name overrides
  - another that replaces the replaceable module and delegates its implementation to the `{:extern}` one. 

  I can imagine that we add an attribute `{:wrapped}` that automatically generates this indirection so that module using `replaces` can once again overwrite names using `{:extern}`, but I'm leaving that as future work.

### How has this been tested?
- Updated the test `replaceables/complex/user.dfy` to make the C# case separately compiled
- Added a test `externErrors.dfy` that captures a new restriction

### DIFF
--- a/Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/replaceables/externErrors.dfy
+++ b/Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/replaceables/externErrors.dfy
@@ -1,0 +1,12 @@
+// RUN: ! %run %s > %t
+// RUN: %diff "%s.expect" "%t"
+
+replaceable module {:extern "FooNameOverride"} Foo {
+  method {:extern "ZazOverride1"} Zaz() returns (i: int) 
+    ensures i >= 2
+}
+
+module {:extern "BarNameOverride1"} Bar replaces Foo { 
+  method {:extern "BarNameOverride2", "ZazOverride2"} Zaz() returns (i: int) 
+    ensures i >= 2
+}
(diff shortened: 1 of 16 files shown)


Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/replaceables/externErrors.dfy
```diff
@@ -0,0 +1,23 @@
...
+    ensures i >= 2
+}
+
+replaceable module {:extern "FaaNameOverride"} Faa {
+  method {:extern "ZazOverride1"} Zaz() returns (i: int) 
+    ensures i >= 2
+}
+
+module {:extern} NoErrors replaces Faa { 
+  method {:extern "ZazOverride2"} Zaz() returns (i: int)
```
alex-chew, line 21: If I understood your PR description correctly, we're also missing an error here right? If so I think it would be worthwhile to put a comment here too.

alex-chew (approved):
LGTM, just one minor comment.

Source/IntegrationTests/TestFiles/LitTests/LitTest/comp/replaceables/externErrors.dfy
keyboardDrummer: You are correct ! Fix: https://github.com/dafny-lang/dafny/pull/4889

PR #4890, opened 2023-12-15
Allow including a test runner using 'dafny translate'

### LINKED ISSUE #3818: Support generating test running code when using `dafny translate`
### Summary

Add an option `--with-test-runner` to `dafny translate` so that it generates a main method that runs all the tests, similar to what `dafny test` does.

### Background and Motivation

Currently Dafny programs can run tests using `dafny test`. However, for Dafny programs that use external code with external dependencies, it might not be possible to compile them using `dafny test`.

### Proposed Feature

See summary

### Alternatives

Alternative 0: by default, let `dafny translate` generate a `runTests` file that contains a main method that will run the tests. Generating annotations for a particular target language testing framework, such as is done for C# and XUnit now, will only be done when using a back-end specific option.

Alternative 1: Increase the functionality of `dafny test` so that is support builds with external code and external dependencies.

Alternative 2: enable `dafny translate` to always generate code for tests that can be executed using a target language testing framework, like is currently done for the C# back-end which generates XUnit tests.

### LINKED ISSUE #4888: Parity between `dafny` and `dafny translate` in CLI
### Summary

The sets of supported compile flags for `dafny` and `dafny translate` are different.
This is an issue if I want to combine `outer-module` and `runAllTests`.
`dafny translate` supports `outer-module`, but not `runAllTests` (nor something like `--run-all-tests`).
`dafny` supports `runAllTests`, but not `outer-module` (nor something like `-outerModule`).

Fixing this will allow Smithy-Dafny to generate tests for code that requires the `--outer-module` flag.

This is currently low-priority; I'm not sure if Smithy-Dafny will end up using `outer-module`, so I may not end up using these flags together. I just want to cut this for tracking.

### Background and Motivation

Using `outer-module` in conjunction with `runAllTests` is not possible. I encountered this when trying to build tests for a project that would use `outer-module`.

### Proposed Feature

Either/or are supported:

`dafny ... -outerModule:path.to.module`

or

`dafny translate [lang] ... --run-all-tests:1`

### Alternatives

_No response_

keyboardDrummer:
Fixes #3818
Fixes #4888

### Description
Add the option `--include-test-runner` to `dafny translate`, to enable getting the results of `dafny test` when doing manual compilation.

### How has this been tested?
Updated `RunAllTestsOption.dfy` so it uses this new feature.


alex-chew (approved):
LGTM

PR #4891, opened 2023-12-15
Let `Std.` example code, and user code using `Std.`, depend on target agnostic interfaces instead of target specific ones

keyboardDrummer:
### Description
- Use replaceable modules to let Std example code, and user code, depend on target agnostic interfaces instead of target specific ones. This prevents differences in behavior for different targets.
- Add missing constructors to classes in the target agnostic concurrent collection code (a bug detected by the previous change)
- Because a replacing module currently can not override external names, some external code had to be updated (and now looks a little worse).

### How has this been tested?
This is purely a refactoring.


Source/DafnyStandardLibraries/src/Std_Concurrent.py
```diff
@@ -10,6 +10,9 @@
 
 
 class MutableMap:
+    def ctor__(self):
+        pass
```
alex-chew, line 14: Why are no-op `ctor__` methods needed?

alex-chew (approved):
LGTM, just added one question for my own understanding.

Source/DafnyStandardLibraries/src/Std_Concurrent.py
keyboardDrummer: Dafny constructors are encoded using these.

PR #4906, opened 2023-12-19
Chore: Ensuring error messages appear before successes

MikaelMayer:
In the IDE, when there are many assertions on the same range, hovering over them reveals information about them in the order they were verified. However, the information most relevant to users is the one about verifications errors. This PR moves them first.

I will add tests shortly in case no test fails.

### DIFF
--- a/Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
+++ b/Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
@@ -170,7 +170,7 @@ namespace Microsoft.Dafny.LanguageServer.Handlers {
         var biggerResourceCountFirst =
           Comparer<(string content, long resources)>.Create(
           (left, right) =>
-            MapLongToInt(right.resources) - MapLongToInt(left.resources)
+            right.resources.CompareTo(left.resources)
         );
         errors.Sort(biggerResourceCountFirst);
         other.Sort(biggerResourceCountFirst);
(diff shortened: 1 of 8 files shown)


Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
```diff
@@ -150,18 +154,29 @@ class AssertionBatchIndexComparer : IComparer<AssertionBatchIndex> {
...
             }
 
             assertionIndex++;
           }
         }
 
+        var biggerResourceCountFirst =
+          Comparer<(string content, long resources)>.Create(
+          (left, right) =>
+            (int)(right.resources / 1000000) - (int)(left.resources / 1000000)
```
atomb, line 173: Is there any need to do this division just to compare resource counts?

atomb (changes requested):
This seems great to me, though there's one place where I think it could be simplified.

Source/DafnyLanguageServer/Handlers/DafnyHoverHandler.cs
MikaelMayer: Good question, the comparison metric requires an "int", when the resource count is a long. I have seen resource counts exceeding the range of an int in practice and display a negative result, but they should be displayed first.
I divided by one million to ensure we were practically within the int range, but I decided to do better. In the newest version, I'm mapping algorithmically long to int and added a comment for that.

atomb: I think the only significance of the `int` result is whether it's less than, equal to, or greater than zero. So I think `right.resources.CompareTo(left.resources)` would work.

MikaelMayer: Oh yeah did not know this method existed, and I could have implemented it myself. Thanks for the simplification.

atomb (approved):
This will be a nice help!

PR #4909, opened 2023-12-20
Relax 'check-deep-tests' workflow to make CI incorrectly fail less often

keyboardDrummer:
### Description
The 'check-deep-tests' step of the CI worflow often fails because the nightly is somewhere in the process of running, but apparently not in the `in_progress` state. This change makes sure that check only fails if nightly actually failed

### How has this been tested?
Careful code review


MikaelMayer (approved):
Good job spotting this tiny issue! how did you come up with that?

keyboardDrummer:
> Good job spotting this tiny issue! how did you come up with that?

I run into this issue every week

fabiomadge:
The previous behavior was in fact intentional

PR #4910, opened 2023-12-20
Fix extern contract checking on abstract modules

### LINKED ISSUE #4845: Crash when running `-testContracts` on refinement of abstract module
### Dafny version

4.3.0

https://github.com/aws/aws-cryptographic-material-providers-library-java/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Index.dfy

```code
$ dafny -noVerify \
      -vcsCores:1 \
      -compileTarget:java \
      -spillTargetCode:3 \
      -compile:0 \
      -optimizeErasableDatatypeWrapper:0 \
      -compileSuffix:1 \
      -quantifierSyntax:3 \
      -unicodeChar:0 \
      -functionSyntax:3 \
      -useRuntimeLib \
      -testContracts:Externs \
      -library:StandardLibrary/src/Index.dfy \
      AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Index.dfy

...
Unhandled exception. System.AggregateException: One or more errors occurred. (One or more errors occurred. (One or more errors occurred. (Object reference not set to an instance of an object.)))
 ---> System.AggregateException: One or more errors occurred. (One or more errors occurred. (Object reference not set to an instance of an object.))
 ---> System.AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.)
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Dafny.ModuleResolver.ResolveCallStmt(CallStmt s, ResolutionContext resolutionContext, Type receiverType)
   at Microsoft.Dafny.ModuleResolver.ResolveStatement(Statement stmt, ResolutionContext resolutionContext)
   at Microsoft.Dafny.ModuleResolver.ResolveStatementWithLabels(Statement stmt, ResolutionContext resolutionContext)
   at Microsoft.Dafny.ModuleResolver.ResolveBlockStatement(BlockStmt blockStmt, ResolutionContext resolutionContext)
   at Microsoft.Dafny.Method.Resolve(ModuleResolver resolver)
   at Microsoft.Dafny.ModuleResolver.ResolveClassMemberBodies(TopLevelDeclWithMembers cl)
   at Microsoft.Dafny.ModuleResolver.ResolveNamesAndInferTypesForOneDeclaration(TopLevelDecl topd)
   at Microsoft.Dafny.ModuleResolver.ResolveNamesAndInferTypes(List`1 declarations, Boolean initialRound)
   at Microsoft.Dafny.ModuleResolver.ResolveTopLevelDecls_Core(List`1 declarations, Graph`1 datatypeDependencies, Graph`1 codatatypeDependencies, String moduleName, Boolean isAnExport)
   at Microsoft.Dafny.ModuleDefinition.Resolve(ModuleSignature sig, ModuleResolver resolver, Boolean isAnExport)
   at Microsoft.Dafny.LiteralModuleDecl.Resolve(ModuleResolver resolver, CompilationData compilation)
   at Microsoft.Dafny.ModuleResolver.ResolveModuleDeclaration(CompilationData compilation, ModuleDecl decl)
   at Microsoft.Dafny.ProgramResolver.ResolveModuleDeclaration(CompilationData compilation, ModuleDecl decl)
   at Microsoft.Dafny.ProgramResolver.Resolve(CancellationToken cancellationToken)
   at Microsoft.Dafny.DafnyMain.<>c__DisplayClass5_0.<Resolve>b__0()
   at System.Threading.Tasks.Task.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__272_0(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at Microsoft.Dafny.DafnyMain.Resolve(Program program)
   at Microsoft.Dafny.DafnyDriver.ProcessFilesAsync(IReadOnlyList`1 dafnyFiles, ReadOnlyCollection`1 otherFileNames, DafnyOptions options, Boolean lookForSnapshots, String programId)
   at Microsoft.Dafny.DafnyDriver.ThreadMain(TextWriter outputWriter, TextWriter errorWriter, TextReader inputReader, String[] args)
   at Microsoft.Dafny.DafnyDriver.<>c__DisplayClass11_0.<MainWithWriters>b__0()
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__272_0(Object obj)
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
   at Microsoft.Dafny.DafnyDriver.MainWithWriters(TextWriter outputWriter, TextWriter errorWriter, TextReader inputReader, String[] args)
   at Microsoft.Dafny.DafnyDriver.Main(String[] args)
   at Dafny.Dafny.Main(String[] args)
...
```

### What happened?

It should have successfully generated target code without throwing an exception.

The exception comes from trying to dereference `callee`, asserted to be non-null [here](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Resolver/NameResolutionAndTypeInference/NameResolutionAndTypeInference.cs#L3137)

(I used a release build above, so it's a later null dereference that fails. With a debug build, the assertion fails.)

The crash is happening on wrapper code generated by the implementation of `testContracts`, but I've done some debugging of the generation, and the `Method` property is set to a non-null value there when running on thins example. So somewhere later, perhaps during cloning, property becomes null.

Mac

atomb:
### Description

Previously, compiling an abstract module with an `{:extern}` method using `--test-assumptions Externs` would cause an assertion failure. This was due to interleaving between the contract checking wrapper generator and the module refinement rewriter. This PR fixes the problem by generating contract checking wrappers for the whole program after resolution is finished, rather than as each module finishes resolution.

Fixes #4845

### How has this been tested?

Tested by `contract-wrappers/RefineContract.dfy`.


Source/DafnyCore/Rewriters/ExpectContracts.cs
```diff
@@ -222,31 +221,36 @@ public class ExpectContracts : IRewriter {
   }
 
   /// <summary>
-  /// Add wrappers for certain top-level declarations in the given module.
-  /// This runs after the first pass of resolution so that it has access to
-  /// ghostness information, attributes and call targets.
+  /// Add wrappers for certain top-level declarations in the given program.
```
MikaelMayer, line 224:
```suggestion
  /// Adds wrappers for certain top-level declarations in the given program.
```
Usually third person to talk about what a function does

MikaelMayer (commented):
Sounds reasonable. Thanks for writing the tests.

keyboardDrummer (changes requested):
> Previously, compiling an abstract module with an {:extern} method using --test-assumptions Externs would cause an assertion failure. This was due to interleaving between the contract checking wrapper generator and the module refinement rewriter.

What is the problem caused by interleaving?

> This PR fixes the problem by generating contract checking wrappers for the whole program after resolution is finished, rather than as each module finishes resolution.

Resolution caching is done on a per-module level. Any operation done in `internal override void PostResolve(Program program) {` is not cached, so we shouldn't have such methods unless they don't relate to a particular module, like `RunAllTestsMainMethod` which adds a single main method for the whole program.

atomb:
> Resolution caching is done on a per-module level. Any operation done in `internal override void PostResolve(Program program) {` is not cached, so we shouldn't have such methods unless they don't relate to a particular module, like `RunAllTestsMainMethod` which adds a single main method for the whole program.

It seems right to me that this shouldn't be cached as part of resolution. I'd consider it to essentially be part of the compiler, and not something that it would make sense to enable in the IDE, for example. It really is a whole-program transformation, in essence, because it introduces wrappers in one module that are then called (by redirection) from other modules.

atomb:
> What is the problem caused by interleaving?

The contract checking code creates a resolved wrapper in the abstract module, which is then copied, with resolution information removed, to the refining module, which hasn't yet been resolved. The removal of the resolution information causes the resolver to crash when checking the refining module. That, on its own, could perhaps be fixed in other ways. But even if it was, the contract checking code would then try to create that wrapper _again_ after resolving the refining module. That could be fixed by adding a special case, but that would additionally complicate the code. By running after all modules are revolved, the approach can be straightforward, without complicating special cases that we could easily get wrong.

keyboardDrummer:
> It seems right to me that this shouldn't be cached as part of resolution. I'd consider it to essentially be part of the compiler, and not something that it would make sense to enable in the IDE, for example. 

Fair, if you do it in `internal override void PostResolve(Program program)` though, it will be part of resolution and be run by the IDE.

> It really is a whole-program transformation, in essence, because it introduces wrappers in one module that are then called (by redirection) from other modules.

In my experience the term whole-program analysis is used when the analysis of a definition (often a callee, but in this case an imported module) is affected by the usages (often the caller, but in this case an importing module), which I think is not the case here. Case in point: leaving out the refinement related bug, contract-wrapping is already working and it is done one module at a time.

> It seems right to me that this shouldn't be cached as part of resolution. I'd consider it to essentially be part of the compiler

But it should be cached! Everything should be cached. We should design the whole compilation pipeline (by compilation pipeline I mean everything from parsing to code generation) with caching in mind. I'll admit that caching is less important for code generation, but I would still argue that it is better if it designed so that it could be cached. If code generation can be done for modules independently, then it could also be run in parallel.

No need to take action for this now though.

> The contract checking code creates a resolved wrapper in the abstract module, which is then copied, with resolution information removed, to the refining module, which hasn't yet been resolved. The removal of the resolution information causes the resolver to crash when checking the refining module. That, on its own, could perhaps be fixed in other ways.

Seems like a general bug not tied to contract-checking

> But even if it was, the contract checking code would then try to create that wrapper again after resolving the refining module. That could be fixed by adding a special case, but that would additionally complicate the code. By running after all modules are revolved, the approach can be straightforward, without complicating special cases that we could easily get wrong.

Refinement modules in general collide with meta-programming transformations. Attributes such as `{:autocontracts}` and `{:autoReq}` have a similar problem.

They contains code such as 

```csharp
if (member.RefinementBase != null) {
  // member is inherited from a module where it was already processed
  continue;
}
```
to deal with this.

That said, those contracts _need_ to be part of refinement, and contract wrapping does not. I think it is indeed easier to do the wrapping transformation _after_ all refinement is done. Could you move the transformation to the compilers/code-generators? It seems there's already a handle `void PostVerification(Program program)` that allows this.

atomb:
> > It seems right to me that this shouldn't be cached as part of resolution. I'd consider it to essentially be part of the compiler, and not something that it would make sense to enable in the IDE, for example.
> 
> Fair, if you do it in `internal override void PostResolve(Program program)` though, it will be part of resolution and be run by the IDE.

But it won't do anything unless the option is enabled, which only makes sense when you're compiling.

> > It really is a whole-program transformation, in essence, because it introduces wrappers in one module that are then called (by redirection) from other modules.
> 
> In my experience the term whole-program analysis is used when the analysis of a definition (often a callee, but in this case an imported module) is affected by the usages (often the caller, but in this case an importing module), which I think is not the case here. Case in point: leaving out the refinement related bug, contract-wrapping is already working and it is done one module at a time.

It processes one module at a time, but it generates calls that go between modules and need to be consistent between them.
 
> That said, those contracts _need_ to be part of refinement, and contract wrapping does not. I think it is indeed easier to do the wrapping transformation _after_ all refinement is done. Could you move the transformation to the compilers/code-generators? It seems there's already a handle `void PostVerification(Program program)` that allows this.

That seems reasonable. Part of the process already happens in `PostVerification(Program)`, and the rest could be moved. I think the fact that it's separate is, IIRC, an artifact of the fact that it used to run pre-resolution in an attempt to have resolution help detect cases where we generate badly-structured code. But we already had to move it after resolution to get ghostness information, so that's no longer a consideration.

When it comes to finishing this PR, is moving the code to all run in `PostVerification` the one change you'd concretely like to see in place?

keyboardDrummer:
> But it won't do anything unless the option is enabled, which only makes sense when you're compiling.

I see. That helps but I'd rather not depend on that.

> When it comes to finishing this PR, is moving the code to all run in PostVerification the one change you'd concretely like to see in place?

Yes

atomb:
> > When it comes to finishing this PR, is moving the code to all run in PostVerification the one change you'd concretely like to see in place?
> 
> Yes

Got it. The latest commit makes that change. Let me know what you think.

Source/DafnyCore/Rewriters/ExpectContracts.cs
```diff
@@ -222,34 +221,42 @@ public class ExpectContracts : IRewriter {
...
   /// <summary>
-  /// Add wrappers for certain top-level declarations in the given module.
-  /// This runs after the first pass of resolution so that it has access to
-  /// ghostness information, attributes and call targets.
+  /// Adds wrappers for certain top-level declarations in the given
+  /// program and redirects callers to call those wrappers instead of
+  /// the original members.
+  ///
+  /// This runs after resolution so that it has access to ghostness
+  /// information, attributes and call targets. It runs on the entire
```
keyboardDrummer, line 229: The sentence 'it runs on the entire program' doesn't make sense to me.

I think this is a transformation where modules are only affected by modules they import, but not the other way around, so you can do it module by module, which the existing loops indicate. I think you can cache the results of processing a module, even though we don't need to do that now.

Source/DafnyCore/Rewriters/ExpectContracts.cs
```diff
@@ -222,34 +221,42 @@ public class ExpectContracts : IRewriter {
...
-
-    moduleDefinition.CallRedirector = new(Reporter);
-
-    // Find module members to wrap.
-    foreach (var topLevelDecl in moduleDefinition.TopLevelDecls.OfType<TopLevelDeclWithMembers>()) {
-      foreach (var decl in topLevelDecl.Members) {
-        if (ShouldGenerateWrapper(decl)) {
-          membersToWrap.Add((topLevelDecl, decl));
+  /// <param name="program">The program to generate wrappers for and in.</param>
+  public override void PostVerification(Program program) {
```
keyboardDrummer, line 234: Should

```
if (Reporter.Options.TestContracts != DafnyOptions.ContractTestingMode.TestedExterns) {
      return;
    }
```

on line 270, be moved to the top of the method?

Source/DafnyCore/Rewriters/ExpectContracts.cs
```diff
@@ -222,34 +221,42 @@ public class ExpectContracts : IRewriter {
...
+    // Create wrappers
+    foreach (var moduleDefinition in program.Modules()) {
```
keyboardDrummer, line 236: Sometimes I see `program.CompileModules` being used and sometimes `program.Modules()`. Is this difference intentional? They're very similar but one also includes the system module.

Source/DafnyCore/Rewriters/ExpectContracts.cs
keyboardDrummer, line 234: There are three loops in this method that all iterate over modules. Can they be merged into one loop?

Source/DafnyCore/Rewriters/ExpectContracts.cs
atomb: I guess that comment isn't as relevant as it was before. When running after resolution, there are two possible methods to override, and this was an explanation for why the code overrides the `Program` version instead of the `Module` version. After verification, there's only one option, though.

Source/DafnyCore/Rewriters/ExpectContracts.cs
atomb: Not without more complex logic that would be easy to get wrong. It needs to ensure that wrappers exist before they're called. The most straightforward way to do that is to construct them all before redirecting calls.

atomb: No. The first two loops should run regardless of that option.

keyboardDrummer: But the modules returned by `.Modules()` are already sorted in topological order of the import graph. How would having a single loop get the order wrong?

keyboardDrummer: I see.

atomb: Ah, right! I hadn't realized it was already sorted. So, yes, the first two loops can be merged. I'll do that.

The final loop is checking for whether any extern exists that hasn't been called by a test, and a test could always show up in a later module in the dependency chain, so I think it still needs to run last.

keyboardDrummer: Sounds good

Issue #4836, opened 2023-12-01
Project files prevent formatting from happening correctly


MikaelMayer:
### Dafny version

latest-nightly

### Code to produce this issue

```dafny
When opening a large project with a Dafny file, right clicking and selecting format produces unexpected random formatting issues.
```

### What happened?

I guess that there are shared variables that should not be shared between files, like the indentation per file. That might solve the issue.

### What type of operating system are you experiencing the problem on?

Other

MikaelMayer:
Since I work on a project file now and want to format, I have the same issue now.

It all boils down to the following:

- In the command line, "dafny format" does not load the project file currently, only parses it, so that it can reformat it correctly.
- when there is a project file, "dafny format" has no longer a working method to determine if the token it's trying to format is part of the file or not, and it will try to apply the formatting of all the files on the current file.

So it's good that the methodology guarantees we print the same tokens. It's just that for some reason "format" is launched on all the files of the project at the same time while all the formatting happens on the current file, which triggers random formatting issues.

keyboardDrummer:
Thanks for fixing https://github.com/dafny-lang/dafny/pull/4918 !!!

PR #4918, opened 2023-12-27
Fix: Formatting on VSCode and CLI

MikaelMayer:
This PR makes sure StreamReaders are closed after use, which prevented formatting from overwriting the files randomly.
I don't think I can write a test case for that.

I manually tested and before, the "dafny format" command was throwing exceptions, and now it no longer throws an exception.

This does not yet solves the issue that when I create a blank Dafny file and try to save it, it will complain with "Failed to save 'todelete.dfy': Unable to write file", but I guess I'll continue searching this one for now.

### DIFF
--- a/Source/DafnyCore/AST/Grammar/IFileSystem.cs
+++ b/Source/DafnyCore/AST/Grammar/IFileSystem.cs
@@ -48,7 +48,10 @@ public class OnDiskFileSystem : IFileSystem {
   }
 
   public TextReader ReadFile(Uri uri) {
-    return new StreamReader(uri.LocalPath);
+    var reader = new StreamReader(uri.LocalPath);
+    var str = reader.ReadToEnd();
+    reader.Close();
+    return new StringReader(str);
   }
 
   public bool Exists(Uri path) {
--- a/Source/DafnyDriver/CompilerDriver.cs
+++ b/Source/DafnyDriver/CompilerDriver.cs
@@ -596,6 +596,7 @@ namespace Microsoft.Dafny {
       if (moreText != null) {
         target.Write(moreText);
       }
+      target.Close();
     }
 
     private static void CheckFilenameIsLegal(string filename) {
--- a/Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
+++ b/Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
@@ -17,15 +17,12 @@ module {:extern "DCOMP"} DCOMP {
   }
 
   function replaceDots(i: string): string {
-    if |i| == 0 then (
-                       ""
-                     ) else (
-                              if i[0] == '.' then (
-                                                    "_" + replaceDots(i[1..])
-                                                  ) else (
-                                                           [i[0]] + replaceDots(i[1..])
-                                                         )
-                            )
+    if |i| == 0 then
+      ""
+    else if i[0] == '.' then
+      "_" + replaceDots(i[1..])
+    else
+      [i[0]] + replaceDots(i[1..])
   }
 
   function escapeIdent(i: string): string {


Source/DafnyCore/AST/Grammar/IFileSystem.cs
```diff
@@ -48,7 +48,10 @@ public class OnDiskFileSystem : IFileSystem {
   }
 
   public TextReader ReadFile(Uri uri) {
-    return new StreamReader(uri.LocalPath);
+    var reader = new StreamReader(uri.LocalPath);
+    var str = reader.ReadToEnd();
+    reader.Close();
+    return new StringReader(str);
```
keyboardDrummer, line 54: This is not OK for performance reasons. We need to design our code towards parsing in a streaming fashion.

keyboardDrummer (changes requested):
Blocking for performance reasons. Details in comments.

Source/DafnyDriver/CompilerDriver.cs
```diff
@@ -596,6 +596,7 @@ private record TargetPaths(string Directory, string Filename) {
       if (moreText != null) {
         target.Write(moreText);
       }
+      target.Close();
```
keyboardDrummer, line 599: Does the above using directive not already guarantee that target is closed? This line seems obsolete.

Source/DafnyCore/AST/Grammar/TokenNewIndentCollector.cs
```diff
@@ -83,8 +83,9 @@ private class Indentations {
   public int binOpIndent = -1;
   public int binOpArgIndent = -1;
 
-  internal TokenNewIndentCollector(Program program) {
+  internal TokenNewIndentCollector(Program program, Uri uri) {
```
keyboardDrummer, line 86: What uri is this? My impression is that you want to format one file at a time, which is fine, but let's make that explicit by renaming `uri` to `fileToFormat`

Source/DafnyCore/AST/Grammar/IndentationFormatter.cs
```diff
@@ -23,11 +23,14 @@ public class IndentationFormatter : IIndentationFormatter {
...
   /// by immediately processing all nodes and assigning indentations to most structural tokens 
   /// </summary>
-  public static IndentationFormatter ForProgram(Program program, bool reduceBlockiness = true) {
-    var tokenNewIndentCollector = new TokenNewIndentCollector(program) {
+  public static IndentationFormatter ForProgram(Program program, Uri uri, bool reduceBlockiness = true) {
+    var tokenNewIndentCollector = new TokenNewIndentCollector(program, uri) {
       ReduceBlockiness = reduceBlockiness
     };
     foreach (var child in program.DefaultModuleDef.PreResolveChildren) {
+      if (child.Tok.line != 0 && child.Tok.Uri != uri) {
```
keyboardDrummer, line 31: `child.Tok.line != 0` what is the meaning of this check? Could you assign the expression to a variable to clarify?

keyboardDrummer:
Thanks for fixing #4836 !!!

Source/DafnyCore/AST/Grammar/IFileSystem.cs
MikaelMayer: I usually worry more about correctness than performance, but I think I might find something that suits both of our needs and I'll put it up in a separate PR.

MikaelMayer: I was about to write a StreamReader that auto-closes the file it reads when done, since I haven't found any. However, after analzying the code further, I don't think it is necessary.

Here is an analysis to consider:

* `IFileSystem.cs:50  public TextReader ReadFile(Uri uri) {` is called by
  * `DafnyProject.cs:53` which has a `ReadToEndAsync` but does not closes the stream
  * `IFileSystem.cs:25   public TextReader ReadFile(Uri uri) {` which has the same usage as its parents
  * `LanguageServerFileSystem.cs:41 existingText = OnDiskFileSystem.Instance.ReadFile(uri).ReadToEnd();` which reads to the end but does not closes the stream
  * `LanguageServerFileSystem:92` which has the same usage as its parents
  * `DafnyFile.cs:96` store the call as a closure in the `GetContent` of a field of a `DafnyFile`. This field is read in different places:
    * DafnyConsolePrinter.cs:72` `var reader = file.GetContent(); lines = Util.Lines(reader).ToList();`. Similarly, we read to the end to get all the lines but it does not close the stream
    * `ProgramParser.cs:269` `var text = SourcePreprocessor.ProcessDirectives(getReader(), new List<string>());` which returns a source file after preprocessing. as a single string, but does not close the stream
    * `CoverageReporter.cs:321` `var source = dafnyFile.GetContent().ReadToEnd();` reads to the end but does not close the stream
    * `FormatCommand.cs:75` `using var content = dafnyFile.GetContent(); var originalText = await content.ReadToEndAsync();` This one also reads to the end and close the stream appropriately (I don't remember doing that, probably you did when you migrate to the new file system, so thanks)

In short, we currently never rely on streaming files and just forget to close it in almost every place. C# does perform auto-closing but is lazy and for big files it won't do it fast enough, which crashes the formatter on the command-line randomly.
So my recommandation is just to do this read and close it immediately so that it becomes possible to write after read. I'll create a separate PR for that.

MikaelMayer: https://github.com/dafny-lang/dafny/pull/4950

keyboardDrummer (approved):
Awesome, thanks

Source/DafnyCore/AST/Grammar/IFileSystem.cs
keyboardDrummer: > DafnyProject.cs:53 which has a ReadToEndAsync but does not closes the stream

What makes you say it does not close the stream? It uses a `using` directive:
```
using var textReader = fileSystem.ReadFile(uri);
```

Good finds though. Let's fix them!

PR #4921, opened 2023-12-29
refactor: Extract constant folding into separate file

RustanLeino:
This PR does a refactoring of the code that partially evaluates expressions--that is, that performs constant folding. It also fixes a null dereference error along the way.


keyboardDrummer (approved):
This is great, thanks

PR #4928, opened 2024-01-03
fix: Check for correct use of type characteristics in ghosts and default expressions

### LINKED ISSUE #4926: Missing (!new) check
### Dafny version

4.4.0

```dafny
function F<A(!new)>(): int

class Cell { }

lemma Lemma()
  ensures F<Cell>() < F<Cell>() + 1  // BUG: should complain that A must be (!new)
{
  // assert F<Cell>() < F<Cell>() + 1;
}
```

```code
% dafny verify test.dfy

```

### What happened?

The type parameter of function `F` is declared as `A(!new)`, which restricts it to types that do not contain references. Thus, the call `F<Cell>()` should be flagged as illegal, but it isn't.

Curiously, Dafny does flag the assertion, if uncommented, as illegal. So, it seems this check is just missing in the postcondition (or perhaps in specifications in general).

Here is another test example:

``` dafny
codatatype InfiniteTree<!A> = Node(value: int, next: A -> InfiniteTree)

greatest predicate Bisimilar<A(!new)>(t: InfiniteTree, u: InfiniteTree) {
  && t.value == u.value
  && forall a: A :: Bisimilar(t.next(a), u.next(a))
}

greatest lemma SelfSimilar<A>(t: InfiniteTree<A>)
  ensures Bisimilar<A>(t, t) // BUG: should complain that A must be (!new)
{
}

class Class { }

lemma SelfSimilarForClass(t: InfiniteTree<Class>)
  ensures Bisimilar<Class>(t, t) // BUG: should complain that A must be (!new)
{
  SelfSimilar<Class>(t);
}

lemma SelfSimilarForClassHash(o: ORDINAL, t: InfiniteTree<Class>)
  ensures Bisimilar#<Class>[o](t, t) // BUG: should complain that A must be (!new)
{
  SelfSimilar#<Class>[o](t);
}
```

Mac

RustanLeino:
This PR adds checks for correct instantiations of type parameters that were previously missing. Previously, these checks were only done in non-ghost expressions (historically, it seems the code that performs these checks was designed for checking `(==)`, which applies only in ghost expressions). This PR adds the checks also specifications (which are ghost) and default expressions for parameters). 

The new test file `git-issue-4926.dfy` reports 38 errors, whereas before this PR, Dafny reported only 8.

It also extracts the code for these checks into a separate file, `TypeCharacteristicChecker.cs` (in addition to the existing file `CheckTypeCharacteristics_Visitor`).

Fixes #4926


keyboardDrummer (dismissed):
Thanks for the refactoring as well ! Note it's easier to review if you put the code move into a separate commit.

keyboardDrummer (approved):

docs/dev/news/4928.fix
```diff
@@ -0,0 +1 @@
+Check for correct usage of type characteristics in specifications and other places where they were missing
```
keyboardDrummer, line 1: If I understand correctly than this PR is a breaking change: it will cause Dafny to error in some locations where it previously did not. Might be good to note that in the release notes.

keyboardDrummer (approved):

PR #4934, opened 2024-01-04
Added OS defaults for Python executable

### LINKED ISSUE #4922: Cannot do `dafny run target:py` on Windows
### Dafny version

4.4.0+707b18acee078b3aa4d84c0590a980966bf22428

```dafny
method Main()
{
    print "hello, world!";
}
```

```code
dafny run --target:py python_noworky.dfy
```

### What happened?

When VSCode runs the above, the following is printed to the console:
```
Error: Unable to start python3: An error occurred trying to start process 'python3' with working directory 'C:\Users\Username\source\playground\program_proofs'. The system cannot find the file specified.
```

I would expect to see the same output as when doing `dafny run python_noworky.dfy` which is `hello, world!`

It seems from this perspective that the current implementation depends on the system-wide implementation being accessible globally through the command `python3`, which is not typically true on Windows. Additionally, there does not seem to be a way to retarget which interpreter is used, which would seem prudent to me.

Suggestions:
- Allow command-line option to override the target interpreter.
- Allow environmental variable to override the target interpreter.
- Add `dfyconfig.toml` option for target interpreter.

Windows

alexporter8013:
### Description

Fixes #4922 
Adds default python executable name by detected OS.

### How has this been tested?

Tested manually, locally. 


Source/DafnyCore/Compilers/Python/PythonBackend.cs
```diff
@@ -46,6 +47,16 @@ public class PythonBackend : ExecutableBackend {
...
   }
 
+  private static readonly Dictionary<OSPlatform, string> PlatformDefaults = new() {
+    {OSPlatform.Linux, "python3"},
+    {OSPlatform.Windows, "python"},
+    {OSPlatform.FreeBSD, "python3"},
+    {OSPlatform.OSX, "python3"},
+  };
+  private static string DefaultPythonCommand {
+    get => PlatformDefaults.Single(kv => RuntimeInformation.IsOSPlatform(kv.Key)).Value;
```
keyboardDrummer, line 57: It would be good to have a fallback value in case the platform value is not among the hardcoded list, instead of throwing an exception.

keyboardDrummer (dismissed):

Source/DafnyCore/Compilers/Python/PythonBackend.cs
alexporter8013: Good call :P I felt since I exhausted the options from `OSPlatform` it wasn't much of an issue, but it can't hurt to use `SingleOrDefault` instead of `Single`.

alexporter8013:
Would I be able to get some help on the failed unit test? On looking, it has to do with file ownership stuff. Aside from not believing my changes would affect that, I also can't debug on OSX since I only have Windows currently plus whatever VM I can install.

keyboardDrummer:
> Would I be able to get some help on the failed unit test? On looking, it has to do with file ownership stuff. Aside from not believing my changes would affect that, I also can't debug on OSX since I only have Windows currently plus whatever VM I can install.

It's an unstable test. I'm sorry to say we have a few at the moment.

Source/DafnyCore/Compilers/Python/PythonBackend.cs
```diff
@@ -46,6 +47,17 @@ public class PythonBackend : ExecutableBackend {
     return Path.GetExtension(externFilename) == ".py" ? Path.GetFileNameWithoutExtension(externFilename) : null;
   }
 
+  private static readonly Dictionary<OSPlatform, string> PlatformDefaults = new() {
+    {OSPlatform.Linux, "python3"},
+    {OSPlatform.Windows, "python"},
+    {OSPlatform.FreeBSD, "python3"},
+    {OSPlatform.OSX, "python3"},
+  };
+  private static string DefaultPythonCommand => PlatformDefaults.SingleOrDefault(
```
keyboardDrummer, line 56: I think you could have done `PlatformDefaults.SingleOrDefault(kv => RuntimeInformation.IsOSPlatform(kv.Key)) ?? "python3"` as well, to drop the unused `OSPlatform.Linux` value

keyboardDrummer (approved):

Source/DafnyCore/Compilers/Python/PythonBackend.cs
alexporter8013: Good catch. I was spinning my wheels on this for a few minutes lol. Your suggestion maybe isn't completely correct because the result of the single call would end up being a KV pair if it succeeded, I think. So I think the full "gudder" way would have been `blah.SingleOrDefault(P)?.Value ?? "python3"`. Is this correct? I usually use Python at work so I don't get a lot of C# road time. IIRC `.?` is the null coalescing accessor which would propogate the null of the `SingleOrDefault`, then the `??` would drop in the `"python3"` if the LHS is null. Please correct me if I'm wrong, like I said, I primarily do Python.

P. S. Apologies on laziness on the shortened parts, I'm on my phone.

keyboardDrummer: Yes, you might be right !

PR #4941, opened 2024-01-05
Fix bug in check deep tests script

keyboardDrummer:
### Description
- Fix the nightly build check so it fails if the last completed build was a failure or timed out.

### How has this been tested?
- Read the code twice
- I will do a manual test once this is merged

### DIFF
--- a/.github/workflows/check-for-workflow-run.js
+++ b/.github/workflows/check-for-workflow-run.js
@@ -16,13 +16,18 @@ module.exports = async ({github, context, core, workflow_id, sha, ...config}) =>
   const runFilterDesc = sha ? `${workflow_id} on ${sha}` : workflow_id
   for (const run of result.data.workflow_runs) {
     if ((!sha || run.head_sha === sha)) {
-      if (run.conclusion === "success") {
-        // The SHA is fully tested, exit with success
-        console.log(`Last run of ${runFilterDesc} succeeded: ${run.html_url}`)
-        return
-      } else if (run.status === "failure" || run.status === "timed_out") {
-        core.setFailed(`Last run of ${runFilterDesc} did not succeed: ${run.html_url}`)
-        return
+      // The status property can be one of: “queued”, “in_progress”, or “completed”.
+      if (run.status === "completed") {
+        // The conclusion property can be one of:
+        // “success”, “failure”, “neutral”, “cancelled”, “skipped”, “timed_out”, or “action_required”.
+        if (run.conclusion === "success") {
+          // The SHA is fully tested, exit with success
+          console.log(`Last run of ${runFilterDesc} succeeded: ${run.html_url}`)
+          return
+        } else if (run.conclusion === "failure" || run.conclusion === "timed_out") {
+          core.setFailed(`Last run of ${runFilterDesc} did not succeed: ${run.html_url}`)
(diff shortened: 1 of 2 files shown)


.github/workflows/check-for-workflow-run.js
```diff
@@ -16,13 +16,18 @@ module.exports = async ({github, context, core, workflow_id, sha, ...config}) =>
...
-        core.setFailed(`Last run of ${runFilterDesc} did not succeed: ${run.html_url}`)
-        return
+      // The status property can be one of: “queued”, “in_progress”, or “completed”.
+      if (run.status === "completed") {
+        // The conclusion property can be one of:
+        // “success”, “failure”, “neutral”, “cancelled”, “skipped”, “timed_out”, or “action_required”.
+        if (run.conclusion === "success") {
+          // The SHA is fully tested, exit with success
+          console.log(`Last run of ${runFilterDesc} succeeded: ${run.html_url}`)
+          return
```
fabiomadge, line 26: You could pull out the `return` again, if you were so inclined.

keyboardDrummer: Two things:
- I'm not sure what you mean by pulling out the return
- It's not clear to me whether you want me to change anything. From your phrasing, "if you are so inclined" it sounds like you don't have an opinion about the change you're describing. If you don't desire any change, could you approve the PR, and if you do, could you be more explicit?

fabiomadge (approved):

.github/workflows/check-for-workflow-run.js
```diff
@@ -16,13 +16,18 @@ module.exports = async ({github, context, core, workflow_id, sha, ...config}) =>
...
+        } else if (run.conclusion === "failure" || run.conclusion === "timed_out") {
+          core.setFailed(`Last run of ${runFilterDesc} did not succeed: ${run.html_url}`)
+          return
+        }
       }
```
fabiomadge, line 31:
```suggestion
      if (run.status === "completed") {
        // The conclusion property can be one of:
        // “success”, “failure”, “neutral”, “cancelled”, “skipped”, “timed_out”, or “action_required”.
        if (run.conclusion === "success") {
          // The SHA is fully tested, exit with success
          console.log(`Last run of ${runFilterDesc} succeeded: ${run.html_url}`)
        } else if (run.conclusion === "failure" || run.conclusion === "timed_out") {
          core.setFailed(`Last run of ${runFilterDesc} did not succeed: ${run.html_url}`)
        } else {
          break
        }
        return
      }
```
More or less what I had in mind.

.github/workflows/check-for-workflow-run.js
fabiomadge: I don't have a strong opinion, so I left it vague. I was also hoping for a simpler change, but I added a suggestions that should be workable.

.github/workflows/check-for-workflow-run.js
keyboardDrummer: Is having a single return statement something you desire in code? I'm not so familiar with that style, but I see it makes the code more like that written in languages that don't have statements.

I'll merge this as is because I feel the status quo of our codebase is to also use a multiple return statement coding style.

PR #4945, opened 2024-01-05
Update documentation for Dafny server

### LINKED ISSUE #4938: DafnyServer does not support project files
### Dafny version

4.4.0+707b18acee078b3aa4d84c0590a980966bf22428

```dafny
// example.dfy
function method foo(): int
{
  0
}
```
```toml
[options]
function-syntax = 3
```
`dafnyserver_input.json`:
```json
{
  "args": [],
  "source": "example.dfy",
  "sourceIsFile": true,
  "filename": "example.dfy"
}
```

```code
$ dafny verify dfyconfig.toml

```
```
$ echo -e "verify\n$(base64 dafnyserver_input.json)" |  DafnyServer
example.dfy(1,0): Error: the phrase 'function method' is not allowed when using --function-syntax:4; to declare a compiled function, use just 'function' (ID: p_deprecating_function_method)

From Dafny 4 on, the phrases `function method` and `predicate method` are no
longer accepted. Use `function` for compiled, non-ghost functions and
`ghost function` for non-compiled, ghost functions, and similarly for predicates.
See [the documentation here](https://dafny.org/latest/DafnyRef/DafnyRef#sec-function-syntax).
Verification completed successfully!
[SUCCESS] [[DAFNY-SERVER: EOM]]
```

### What happened?

As can be seen above, `DafnyServer` doesn't respect settings in `dfyconfig.toml` - at least not in a way that I could find.

The [documentation](https://dafny.org/latest/DafnyRef/DafnyRef#sec-dafny-server) states:
>The Dafny Server is still supported, but we recommend using the [Language Server](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) instead.

So is this considered a bug in `DafnyServer` or should the documentation mention that some things aren't supported?

Linux

keyboardDrummer:
Fixes #4938

Update documentation for Dafny server

### DIFF
--- a/docs/DafnyRef/VSCodeIDE.md
+++ b/docs/DafnyRef/VSCodeIDE.md
@@ -73,7 +73,7 @@ the position of the parse or resolution error.
 
 ## 14.3. The Dafny Server {#sec-old-dafny-server}
 
-Before Dafny [implemented](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer) the official [Language Server Protocol](https://microsoft.github.io/language-server-protocol/), it implemented its own protocol for [Emacs](https://github.com/boogie-org/boogie-friends), which resulted in a project called [DafnyServer](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyServer).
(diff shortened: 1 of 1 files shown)


atomb (approved):
Good clarification.

PR #4950, opened 2024-01-08
Fix: Formatting on the command line works again

MikaelMayer:
This PR makes sure StreamReaders are closed after use, which prevented formatting from overwriting the files randomly.

Before, the "dafny format" command was throwing exceptions on large files, and now it no longer throws an exception (test added for that)

Even better, this solved the issue that when I create a blank Dafny file and try to save it, it used to complain with "Failed to save 'todelete.dfy': Unable to write file". (Fixes https://github.com/dafny-lang/ide-vscode/issues/452)

### DIFF
--- a/Source/DafnyCore/AST/Grammar/IFileSystem.cs
+++ b/Source/DafnyCore/AST/Grammar/IFileSystem.cs
@@ -48,7 +48,10 @@ public class OnDiskFileSystem : IFileSystem {
   }
 
   public TextReader ReadFile(Uri uri) {
-    return new StreamReader(uri.LocalPath);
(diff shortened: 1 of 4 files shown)


keyboardDrummer:
> The reason why to do this despite an obvious memory increase is that in https://github.com/dafny-lang/dafny/pull/4918#discussion_r1444530522, we actually already read to the end and don't make use of the streaming.

I'm aware of that, but to me that doesn't mean we should not design for the future in which we will not. I think if we stop reading from a stream now, we'll stop being disciplined about closing the stream, and have more work later when we switch to parsing from a stream.

keyboardDrummer (approved):

Issue #4946, opened 2024-01-05
Resolver crashes on possibly empty subset type of trait


ssomayyajula:
### Dafny version

4.4.0

### Code to produce this issue

```dafny
// Trampoline.dfy

trait ProgramTrait {
  method Compute() returns (r: Result)
}

type Program = ProgramTrait | true

datatype Result =
| Bounce(next: Program)
| Done()

datatype Trivial extends ProgramTrait =
  Trivial()
{
  method Compute() returns (r: Result)
  {
    return Done();
  }
}
```

### Command to run and resulting output

```code
$ dafny verify --type-system-refresh --general-traits:datatype Trampoline.dfy

Unhandled exception: System.AggregateException: One or more errors occurred. (Unable to cast object of type 'Microsoft.Dafny.InferredTypeProxy' to type 'Microsoft.Dafny.UserDefinedType'.)
 ---> System.InvalidCastException: Unable to cast object of type 'Microsoft.Dafny.InferredTypeProxy' to type 'Microsoft.Dafny.UserDefinedType'.
   at Microsoft.Dafny.ModuleResolver.CheckCanBeConstructed(Type type, ISet`1 typeParametersUsed) in /com.docker.devenvironments.code/Source/DafnyCore/Resolver/ModuleResolver.cs:line 2920
   at Microsoft.Dafny.ModuleResolver.CheckCanBeConstructed(Type type, ISet`1 typeParametersUsed) in /com.docker.devenvironments.code/Source/DafnyCore/Resolver/ModuleResolver.cs:line 2951
   at Microsoft.Dafny.ModuleResolver.ComputeGroundingCtor(IndDatatypeDecl dt) in /com.docker.devenvironments.code/Source/DafnyCore/Resolver/ModuleResolver.cs:line 2879
   at Microsoft.Dafny.ModuleResolver.SccStratosphereCheck(IndDatatypeDecl startingPoint, Graph`1 dependencies) in /com.docker.devenvironments.code/Source/DafnyCore/Resolver/ModuleResolver.cs:line 2841
   at Microsoft.Dafny.ModuleResolver.ResolveTopLevelDecls_Core(List`1 declarations, Graph`1 datatypeDependencies, Graph`1 codatatypeDependencies, String moduleDescription, Boolean isAnExport) in /com.docker.devenvironments.code/Source/DafnyCore/Resolver/ModuleResolver.cs:line 1260
[...]

PR #4956, opened 2024-01-09
fix: Fix parsing error and a crash with non-reference-trait base type

RustanLeino:
This PR fixes three bugs.

One bug was that the new resolver didn't check that the base type of a subset type was determined.

Another bug was that, even after reporting the error in the first bug, the resolver could crash downstream.

The third bug was that, even with a properly formed `x: BaseType`, the verifier would crash when trying to construct a default witness (if the base type is a non-reference trait and there is no explicit `witness` clause).

Fixes https://github.com/dafny-lang/dafny/issues/4946

Note: I previously (and erroneously) thought the bug was a problem in the parser. But we _do_ allow the syntax with just one identifier before the `|` in a subset-type declaration. The problem was that the new resolver didn't check to see that the type of that identifier was properly inferred.


keyboardDrummer (approved):

ssomayyajula (approved):
LGTM!

ssomayyajula (approved):

ssomayyajula (approved):

PR #4971, opened 2024-01-10
Rework warn-as-error option

keyboardDrummer:
The existing option `--warn-as-error` causes everything that would otherwise be a warning, to be recorded as an error. Because the Dafny compiler often uses the presence of errors to determine whether to continue, using `--warn-as-error` causes compilation to potentially halt earlier when it's turned on. This can hide errors until warnings that occur earlier in the pipeline have been resolved. In addition, using this option makes it impossible for the user to see what was originally a warning instead of an error.

### Changes
- Replace the option `--warn-as-error` with the option `--succeed-on-warnings`. Besides having the opposite meaning, this option no longer turns warnings into errors, nor does it reduce the amount of diagnostics collected in compilation. When turned off, the default, any warnings will cause the CLI to return an error exit code and prevent the execution of the Dafny code. All forms of execution, whether directly executing it or through the creation of build artifacts such as generated code (cs/java/go/etc) or further compiled generated code (dll/jar/etc), are prevented.
- A doo file that was created using `--succeed-on-warnings` causes a _warning_ if used by a consumer that does not use it.
- Change the behavior of `--warn-deprecation=false` so that deprecation diagnostics are reported as information, which means they show up in the IDE
- Flip the behavior of `--warn-deprecation` and change the name to `--allow-deprecation`, so the default is now false, which is common for boolean options. It's slightly nicer for user to type `--allow-deprecation` instead of `--warn-deprecation=false`.

### Caveats
- When a user is developing and wants to quickly run their code to test something, but the code has warnings, the user must add `--succeed-on-warnings` to their `dafny run` invocation.
- This change is not backwards compatible. 
  - Dafny users must remove any usages of `--warn-as-error`
  - If users have builds that were passing with warnings, they have to add `--succeed-on-warnings` to allow them to still pass
  - If users upgrade to a new Dafny version, and are not using `--succeed-on-warnings`, and do not want to migrate off of deprecated feature, they will have to use `--allow-deprecation`. In the future we will introduce `dafny migrate` to enable use to automatically move off of deprecated features, allowing them to instantly upgrade to newer Dafny versions without having to use `--allow-deprecation` or `--succeed-on-warnings`.
  - Any instance of `--warn-deprecation false` must be replaced with `--allow-deprecation`

### How has this been tested?
- Updated tests


robin-aws:
Just commenting on the description rather than the implementation.

I'm definitely in favour of no longer converting warnings to errors and instead having the option to fail on warnings.

I don't think it makes sense to fail on warnings by default though. IMO the whole point of a "warning" as opposed to an "error" is that it isn't fatal by default, and if we have warnings that should fail the pipeline by default we should upgrade them to errors with ways to quickly fix/suppress them if necessary.

I suspect based on other conversations that the underlying reason for wanting this is that Dafny currently has things it only conditionally flags as errors based on command, which we want to make warnings instead. For example, `assume` statements without `{:axiom}` or methods without bodies: those are allowed by `dafny verify` but errors in `dafny translate`. The problem with making those warnings is they will be just noise when using Dafny as a proof assistant and never translating or executing the code.

I'd suggest we try to tackle supporting these different use cases better more directly instead. Perhaps there should be a project file property indicating whether the Dafny code will only be used for verification and not execution, similar to the `<OutputType>` property in `.csproj` files that can be `Exe` or `Library` and so on. That way if a project was tagged as "executable", `dafny verify` could produce errors during resolution if it finds execution issues such as methods without bodies.

keyboardDrummer:
> I don't think it makes sense to fail on warnings by default though. IMO the whole point of a "warning" as opposed to an "error" is that it isn't fatal by default, and if we have warnings that should fail the pipeline by default we should upgrade them to errors with ways to quickly fix/suppress them if necessary.

It's unclear to me how you are positioning warnings. It sounds to me like you're saying warnings are never fatal, which implies that they are allowed to proliferate in a codebase, which I think makes them more annoying than useful.

I would like to position a warning as something that is fatal by default in CI, but not fatal during development.

Running Dafny during development is done by calling `dafny run --pass-on-warnings`. If you are calling `dafny verify` during development, you likely don't care about the exit code and you have no need to use `--pass-on-warnings`. Also, if you are using `dafny server`, there is no relevant exit code and you don't need `--pass-on-warnings`.

I have considered letting `dafny run` have `--pass-on-warnings=true` by default, since I believe `dafny run` is mostly used during development. However, I want to be defensive and not make that change, because I can also imagine a future where `dafny run` is used for production purposes, and for example there is a `dafny debug` that's meant for development.

> I suspect based on other conversations that the underlying reason for wanting this is that Dafny currently has things it only conditionally flags as errors based on command, which we want to make warnings instead. For example, assume statements without {:axiom} or methods without bodies: those are allowed by dafny verify but errors in dafny translate. The problem with making those warnings is they will be just noise when using Dafny as a proof assistant and never translating or executing the code.

I disagree. I think even as a proof assistant you want to differentiate between an `assume` that you are adding temporarily as part of developing your proof, and one that you are checking into your codebase for the long term: `assume {:axiom}`. I think this is unrelated to whether you're also executing code.

robin-aws:
Okay I get where you're coming from: you want all warnings to always block committing changes to a project. I don't think that's the only valid definition of "warning" but I do like the idea of being that strict about it. Really it makes warnings "non-fatal errors" - ones that ultimately mean the program isn't valid but that don't block doing useful local development with it such as executing it.

If we're going that route though, I think we should review all our warnings to look for ones that shouldn't block CI before we make failing on warnings the default. There are definitely ones that should not be interpreted as "non-fatal errors" and aren't necessarily quick to address. For example: "a forall statement with no bound variables is deprecated; use an 'assert by' statement instead". Similarly, although I think it's reasonable to make `assume` without `{:axiom}` a warning, we shouldn't do the same for things that are only errors for translation/execution like bodiless methods (and it's possible my suggestion for catching those errors by communicating the intent of a project might be useful orthogonally).

It might also be to have more severity levels: "warning" for things that may or may not be issues, "error" for definite problems that shouldn't be allowed in CI but don't block development, and "fatal" for things where the pipeline can't proceed any further.

robin-aws:
Minor quibble: I'd suggest `--fail-on-warnings:false` instead of `--pass-on-warnings` for pure readability, just because "fail" feels a little more intuitive in the context of CLI exit codes. Or perhaps `--succeed-on-warnings`?

keyboardDrummer:
> I think we should review all our warnings to look for ones that shouldn't block CI before we make failing on warnings the default

What do you think would be missing from the test-suite to help with that review?

> For example: "a forall statement with no bound variables is deprecated; use an 'assert by' statement instead"

One change that I forgot to list in the PR details is that the option `--warn-deprecation` is off by default, and that deprecation diagnostics are reported as "information" when this option is turned off, meaning they still show up in the IDE.

> Similarly, although I think it's reasonable to make assume without {:axiom} a warning, we shouldn't do the same for things that are only errors for translation/execution like bodiless methods

I think we should let these two behave the same, so a bodyless method will produce a warning whether you are executing or not, unless you annotate it with {:axiom}, similar to an `assume` statement.

> It might also be to have more severity levels: "warning" for things that may or may not be issues, "error" for definite problems that shouldn't be allowed in CI but don't block development, and "fatal" for things where the pipeline can't proceed any further.

I'm not sure I understand what you're suggesting. The labels we have at our disposal through LSP are "error", "warning" and "information". I think the common way to map these to behavior are:
"error" => always returns an error exit code
"warning" => returns an error exit code depending on CLI options
"information" => never returns an error exit code

> Minor quibble: I'd suggest --fail-on-warnings:false instead of --pass-on-warnings for pure readability, just because "fail" feels a little more intuitive in the context of CLI exit codes. Or perhaps --succeed-on-warnings

It's nicer for boolean options to have false be the default, so you can turn them on using just `--option-name`, but I agree `--fail-on-warnings` is an easier to understand name. If `--succeed-on-warnings` is clearer than `--pass..` to you, then I'd prefer that.

robin-aws:
> > I think we should review all our warnings to look for ones that shouldn't block CI before we make failing on warnings the default
> 
> What do you think is missing from the test-suite to complete that review?

I did a quick scan myself - I just wanted to check for any warnings that were obviously unreasonable to block on, and the point about not warning on deprecation any more helps.

The only one that gave me pause was the one in DafnyProject about the IDE only recognizing files named `dfyconfig.toml`. I'm surprised that one didn't cause a test failure.

> > Similarly, although I think it's reasonable to make assume without {:axiom} a warning, we shouldn't do the same for things that are only errors for translation/execution like bodiless methods
>
> I think we should let these two behave the same, so a bodyless method will produce a warning whether you are executing or not, unless you annotate it with {:axiom}, similar to an `assume` statement.

Ah fair enough, having to add `{:axiom}` is a reasonable way to suppress that one.

> > It might also be to have more severity levels: "warning" for things that may or may not be issues, "error" for definite problems that shouldn't be allowed in CI but don't block development, and "fatal" for things where the pipeline can't proceed any further.
> 
> I'm not sure I understand what you're suggesting. The labels we have at our disposal through LSP are "error", "warning" and "information". I think the common way to map these to behavior are: "error" => always returns an error exit code "warning" => returns an error exit code depending on CLI options "information" => never returns an error exit code

Damn, didn't realize the LSP was prescriptive about that, never mind.

> > Minor quibble: I'd suggest --fail-on-warnings:false instead of --pass-on-warnings for pure readability, just because "fail" feels a little more intuitive in the context of CLI exit codes. Or perhaps --succeed-on-warnings
> 
> It's nicer for boolean options to have false be the default, so you can turn them on using just `--option-name`, but I agree `--fail-on-warnings` is an easier to understand name. If `--succeed-on-warnings` is clearer than `--pass..` to you, then I'd prefer that.

Sure, let's go with `--succeed-on-warnings` then!

Source/DafnyCore/Compilers/CSharp/Synthesizer-Csharp.cs
```diff
@@ -179,11 +179,12 @@ public class CsharpSynthesizer {
     var methodName = method.GetCompileName(Options);
 
     if (((Function)method).Ens.Count != 0) {
-      compiler.Error(CompilerErrors.ErrorId.c_possibly_unsatisfied_postconditions, lastSynthesizedMethod.tok,
```
robin-aws, line 182: I'd be happier if this stayed as an error, does it have to change?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -720,7 +720,7 @@ public enum IncludesModes {
           }
 
         case "warningsAsErrors":
```
robin-aws, line 722: The release notes should mention that `/warningsAsErrors` is changing behavior too.

Source/DafnyCore/DooFile.cs
```diff
@@ -269,10 +273,12 @@ public class ManifestData {
...
   }
 
-  private static bool OptionValuesImplied(Option option, object first, object second) {
-    var lhs = (bool)first;
-    var rhs = (bool)second;
-    return !lhs || rhs;
+  public static bool OptionValuesImplied(Option option, object first, object second) {
+    try {
+      return !(bool)first || (bool)second;
+    } catch (NullReferenceException) {
```
robin-aws, line 279: I'd prefer not to swallow this as it shouldn't happen with released Dafny versions: AFAICT you can only be missing options in doo files if they were built with an older version of Dafny, which is explicitly blocked.

Source/DafnyDriver/DafnyBackwardsCompatibleCli.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Dafny;
 public interface ILegacyParseArguments { }
 
 // TODO: Refactor so that non-errors (NOT_VERIFIED, DONT_PROCESS_FILES) don't result in non-zero exit codes
-public enum ExitValue { SUCCESS = 0, PREPROCESSING_ERROR, DAFNY_ERROR, COMPILE_ERROR, VERIFICATION_ERROR, FORMAT_ERROR }
+public enum ExitValue { SUCCESS = 0, PREPROCESSING_ERROR = 1, DAFNY_ERROR = 2, COMPILE_ERROR = 3, VERIFICATION_ERROR = 4, FORMAT_ERROR = 5 }
```
robin-aws, line 14: Ah does C# not guarantee this indexing by default?

Source/DafnyRuntime/DafnyRuntimeGo/System_/System_.go
robin-aws, line 1: What happened here? it looks like `goimports` didn't get applied, but CI should be catching the mismatch.

Source/IntegrationTests/TestFiles/LitTests/LitTest/c++/arrays.dfy
```diff
@@ -1,4 +1,4 @@
-// RUN: %testDafnyForEachCompiler --refresh-exit-code=0 "%s" -- --relax-definite-assignment --spill-translation --unicode-char:false
```
robin-aws, line 1: Why this change?

Source/IntegrationTests/TestFiles/LitTests/LitTest/git-issues/git-issue-1151-more.dfy.expect
```diff
@@ -6,19 +6,15 @@ git-issue-1151-more.dfy(13,5): Error: Abstract type ('_module.Opaque') cannot be
 
 Dafny program verifier did not attempt verification
 git-issue-1151-more.dfy(13,5): Error: Abstract type ('_module.Opaque') cannot be compiled; perhaps make it a type synonym or use :extern.
-Wrote textual form of partial target program to git-issue-1151-more-java/git_issue_1151_more.java
```
robin-aws, line 9: Was it intentional to lose these outputs?

docs/dev/news/4971.feat
```diff
@@ -0,0 +1,7 @@
+- Flip the behavior of `--warn-deprecation` and change the name to `--allow-deprecation`, so the default is now false, which is standard for boolean options.
+- When using `--allow-deprecation`, deprecated code is shown using tooltips in the IDE, and on the CLI when using `--show-tooltips`.
+- Replace the option `--warn-as-error` with the option `--succeed-on-warnings`. The new option, when false, the default value, causes Dafny to stop generating executable output and return a failure exit code, when warnings occur in the program. Contrary to the previous `--warn-as-error` option, warnings are always reported as warnings.
+  - During development, users must use `dafny run --succeed-on-warnings` if they want to run their Dafny code when it contains warnings.
```
robin-aws, line 4: This is worth mentioning in the user guide as well somewhere.

robin-aws (changes requested):
Implementation looks great, just asking for a couple of tweaks.

I did realize that `--succeed-on-warnings` will be confusing for `dafny run` and `dafny test`, since it's only the pipeline that is "succeeding" and not the process' exit code. I suggest `--allow-warnings` instead, assuming it's just an easy search-and-replace.

Source/DafnyCore/DooFile.cs
keyboardDrummer: Note that I return `false` here, so it will prevent the doo file from being used.

It also sounds like you're saying that you'd like a version check when using Doo files, and that it's currently not there, otherwise I wouldn't have been able to hit this exception.

Source/DafnyDriver/DafnyBackwardsCompatibleCli.cs
keyboardDrummer: It does, but sometimes you have to know what the numbers are when looking at test output, so I found this easier to read.

Source/DafnyRuntime/DafnyRuntimeGo/System_/System_.go
keyboardDrummer: O is this when it does _not_ get applied? I couldn't judge by the changes. It looked to me like it was now suddenly applying the formatting :-D

keyboardDrummer: Made a change since this gets reverted.

Source/IntegrationTests/TestFiles/LitTests/LitTest/c++/arrays.dfy
keyboardDrummer: Just cleanup. The test does not need that option.

Source/IntegrationTests/TestFiles/LitTests/LitTest/git-issues/git-issue-1151-more.dfy.expect
robin-aws: Completely agree with that and was tempted to do this myself in the past, just didn't spot the connection - thanks for that change then!

Source/DafnyCore/DooFile.cs
robin-aws: No the version check totally exists! https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/DooFile.cs#L129-L133. I think we're missing an integration test for that though - I recall writing one that depended on checking in a doo file with an older version, but I think we had to pull it out for some reason.

However that only kicks in when we bump the Dafny version number, which we usually don't do until just before releasing. So it's not hard to hit the NRE in development if you change the set of recorded options and have doo files lying around from before the change.

It might be worth defensively emitting a clearer error message, even if this would be an internal error that shouldn't happen on released Dafny versions. But I definitely don't want to silently return false since it would indicate something else has gone wrong.

Source/DafnyCore/Compilers/CSharp/Synthesizer-Csharp.cs
robin-aws: Got it. I don't love downgrading the severity of the issue but we should still support the use case. I cut https://github.com/dafny-lang/dafny/issues/5075 for follow up instead.

Source/DafnyCore/DooFile.cs
keyboardDrummer: > It might be worth defensively emitting a clearer error message, even if this would be an internal error that shouldn't happen on released Dafny versions. But I definitely don't want to silently return false since it would indicate something else has gone wrong.

I think I disagree, but I've made the change you suggested.

docs/DafnyRef/UserGuide.md
```diff
@@ -233,7 +233,7 @@ implementation.
 
 The options relevant to this command are
 - those relevant to the command-line itself
-   - `--warn-as-errors` --- turn all warnings into errors, which alters [dafny's exit code](#sec-exit-codes)
+   - `--allow-warnings` --- return a success [exit code](#sec-exit-codes), even when there are warnings
```
robin-aws, line 236: Not true though since `dafny run` or `dafny test` still could return a non-zero exit code.

Issue #4204, opened 2023-06-21
Dafny passes method attributes to Boogie raw, causing unsoundness


Dargones:
### Dafny version

4.1.0

### Code to produce this issue

```dafny
function {:inline 1} b(i:int):int {1/i}
function a():int { b(0)/1 }
```

### Command to run and resulting output

```code
> dafny FILE.dfy
Dafny program verifier finished with 1 verified, 0 errors
Compiled assembly into FILE.dll
```

### What happened?

As far as I understand, Dafny is not intended to support the `:inline` attribute and I doubt anyone uses it (search for `"function {:inline 1}" AND language:dafny` does not return anything on GitHub), so the bug probably has no impact on Dafny users. That said, it appears that Dafny performs no checks on which attributes it passes on to Boogie, so I am wondering if there are other attributes that can similarly introduce unsoundness when interpreted by Boogie.

The solution that makes most sense to me would be to maintain in `DafnyOptions` a list of allowed attributes that would be passed on to Boogie (if there are such attributes), with an option for a Dafny extension to modify this list if necessary.

The reason Dafny verifies the code in this particular case is, I think, that Boogie ignores the `CheckWellformed$$_module.__default.b` procedure emitted by Dafny for function `b` because it has the `{:inline 1}` attribute on it. The corresponding procedure for `a` simply assumes that it can call `b(0)` because `b` has no preconditions.

### What type of operating system are you experiencing the problem on?

Mac

MikaelMayer:
Another attribute that is passed raw, causing unsoundness:

```dafny
method {:selective_checking} Test() {
  assert false;
}
```

MikaelMayer:
Because some users have their own compilers for Dafny that make use of annotations, it would be cumbersome to ask them to write extensions for Dafny to not complain about their custom attributes.

What I think we could do better is the following:
1) Whitelist a set of options for methods that can be passed to Boogie
2) Any other option is removed for verification by Boogie
3) If an option is given like "{:boogie-X}", then it is kept and rewritten into "{:X}" for Boogie verification.

That way, we will never accidentally add an attribute that is not recognized, while not blocking users that make use of custom boogie attributes.

I think that we should do the following:
1) Have a list of attributes

Dargones:
> That way, we will never accidentally add an attribute that is not recognized, while not blocking users that make use of custom boogie attributes.

I don't know enough about other extensions of Dafny but this would definitely work for test generation purposes!

[...]

PR #4972, opened 2024-01-10
Use resource limit in most places in test suite

atomb:
A few exceptions remain, but this will allow testing the general idea

### How has this been tested?


Source/DafnyDriver/Commands/DafnyCliTests.cs
```diff
@@ -42,8 +42,11 @@ public static class DafnyCliTests {
     // We do not want output such as "Compiled program written to Foo.cs"
     // from the compilers, since that changes with the target language
     "/compileVerbose:0",
-      
-    // Set a default time limit, to catch cases where verification time runs off the rails
+
+    // Set a default resource limit, to catch cases where verification runs off the rails
+    "/rlimit:50000000",
```
robin-aws, line 47: Shouldn't this be `50000` instead, since this one does the X1000?

Source/DafnyDriver/Commands/DafnyCliTests.cs
```diff
@@ -57,10 +60,13 @@ public static class DafnyCliTests {
     // We do not want absolute or relative paths in error messages, just the basename of the file
     "--use-basename-for-filename",
 
-    // Set a default time limit, to catch cases where verification time runs off the rails
-    "--verification-time-limit=300",
+    // Set a default resource limit, to catch cases where verification runs off the rails
+    "--resource-limit:50000000",
```
robin-aws, line 64:
```suggestion
    "--resource-limit:50e6",
```
?

(and likewise elsewhere)

Source/DafnyLanguageServer.Test/Various/CancelVerificationTest.cs
```diff
@@ -80,7 +80,7 @@ public class CancelVerificationTest : ClientBasedLanguageServerTest {
     Ack(m - 1, Ack(m, n - 1))
 }
 
-method {:timeLimit 10} test() {
+method {:rlimit 10000} test() {
```
robin-aws, line 83: I get that with the legacy CLI we only have `/rlimit` as an option, but for attributes we should only use the non-deprecated `{:resource_limit}` attribute IMO.

robin-aws (changes requested):

Source/DafnyDriver/Commands/DafnyCliTests.cs
atomb: Good catch.

Source/DafnyDriver/Commands/DafnyCliTests.cs
atomb: Yeah, I was debating doing that. It's probably better. :)

Source/DafnyDriver/Commands/DafnyCliTests.cs
atomb: Oh, actually, no. Boogie no longer does the multiplication, and it handles this flag directly. It's one of many reasons to deprecate it. I think I'll implement the deprecation warning as part of fixing #4204.

robin-aws: Oh I didn't realize that was true for the CLI flag. Should we update https://github.com/dafny-lang/dafny/blob/master/docs/dev/news/4975.fix to call both out?

robin-aws (approved):

Source/DafnyDriver/Commands/DafnyCliTests.cs
atomb: Yep, I'll do that in a separate PR.

Issue #4882, opened 2023-12-14
Class datatypes


MikaelMayer:
### Summary

I would like to declare a class with a datatype syntax. 

### Background and Motivation

Datatypes are extremely useful because of their short syntax to define, to build and to deconstruct.
Classes are much more verbose but it makes it possible to reason about the fact that they consume heap memory, which datatypes are extremely good at hiding.
The problem with datatypes in the compiled code is that they can be duplicated or reused infinitely. In a garbage-collected language, it's fine. In an ownership language like Rust, it requires at minimal reference counting, if we don't want to change the language to add memory annotations about types.

If we had a way to have allocated datatypes (that is, whose allocation actually creates a fresh object), it would make it possible down the road to reason fully about the memory consumption about a program, and have the compiler deallocate them when no longer needed.
We might not need this immediately, but I'm working on some memory-related features and datatypes makes it impossible to reason about the actually memory actually used by a Dafny program (the Dafny heap model does not care about datatypes), and users would be afraid to move to classes because they are verbose and don't have pattern matching.

To summarize, I want a way to allocate datatype-like values on the heap, reason about their freshness and unicity (so that deallocation would not necessarily need reference counting or automated garbage collection), and still have pattern matching.

### Proposed Feature

Just some thoughts:

```dafny
class datatype X = C1(i: int, y: int) | C2(i: int, s: string) {
  ...members
}
```

Instead of X being a value, it would be just an allocated reference, so that two instances with the same fields are not necessary equal in reference.

The desugaring could look like what we currently generate datatypes for

```dafny
trait X extends object {
  const C1?: bool
  const C2?: bool
  function i(): int
    requires C1? || C2?

  function y(): int
    requires C1?
  
  function s(): string
    requires C2?
}
class C1 extends X {
  const i: int
  const y: int
  constructor(i: int, y: int) ensures C1? && !C2? && i == this.i && y == this.y {
    this.i := i;
    this.y := y;
    C1? := true;
    C2? := false
  }
  ...members
}
```

With that, we would be able to support pattern matching on such classes out of the box:
```
method Print() {
  var x := new C1(1, 2);
  match x
  case C1(i, x) => print i, y;
  case _ =>
}
```
would be desugared to
```
method Print(x: X) {
  var x := new C1(1, 2);
  if x.C1? {
    var i := x.i();
    var y := x.y();
    print i, y;
  }
}
```

### Alternatives

[...]

PR #4974, opened 2024-01-10
Feat: Dafny-to-Rust code indentation and identifiers

MikaelMayer:
# Improvements of the DAST to Rust code generator:

* The code generator emits a Rust-like AST that is then formatted with proper indentation
* Ensured Rust-like identifiers in Dafny stay the same in Rust (like "one_variable")
* Added Initial implementation of sequences and maps in the DafnyRuntimeRust, following the implementation of the Go
* Added module to perform conversions from and to Dafny types to and from native types

# Code enhancement

* Added support for using the standard library's Option type rather than the hard-coded Optional type.

# Tests

* Added proof that the identifier mapping is fully reversible and won't conflict with other internal constructs
* Added a Rust test module for the DafnyRuntimeRust and wrote tests for testing the encoding Dafny/Rust

# Tooling improvement

- The makefile now works even if there is a space in the path
- `make dfydev` does not perform verification and formatting of the resulting C# file, but `make dfyprod` does it. Both build Dafny afterwards
' `make dfydevinit` and `make dfyprodinit` only perform the conversion from dafny files to C# files, without rebuilding Dafny.
 


Source/DafnyCore/Compilers/Dafny/AST.dfy
```diff
@@ -70,9 +75,9 @@ module {:extern "DAST"} DAST {
     Neq(referential: bool, nullable: bool) |
     Div() | EuclidianDiv() |
     Mod() | EuclidianMod() |
-    Implies() |
+    Implies() | // TODO: REplace by Not Or
```
robin-aws, line 78: That seems like a good idea to me, but it assumes that the goal of the DAST is to be as simple as possible, rather than retaining extra information that might produce better output in some contexts. I think that's a good design goal personally, but this file should be more explicit about documenting that.

robin-aws (commented):
Just some high-level comments as I'm finding my bearings in the new DafnyCompiler/DafnyWrittenCompiler world, more to come!

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+        pub fn dafny_string_to_string(s: &Rc<Sequence<u16>>) -> String {
+            let characters = s.to_array();
+            String::from_utf16_lossy(&characters)
+        }
+    }
+    
+}
+
+#[allow(dead_code)]
+enum Sequence<T>
```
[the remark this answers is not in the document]
MikaelMayer: If will be feasible to use the compilation of that source directly instead, but it's more likely if we have [sealed traits (or class datatypes)](https://github.com/dafny-lang/dafny/issues/4882). as it's not possible in Dafny to declare an allocated datatype, which we would need here to compile to an enum.
At best, we would be able to generate a `dyn Sequence` is Sequence was a trait, but it would not fit the conversion I'm envisioning.
But even then, the Dafny to Rust compiler does not support consistently the notion of traits anyway yet.

What did you do with this --bootstrapping flag exactly?

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+  }
+  // replaceDots is not reversible generally speaking.
+  // Indeed, both _d and . get translated to _d.
+  lemma replaceDotsNotReversible()
+    ensures replaceDots("_d") == replaceDots(".")
+    ensures !isDafnyEncodedIdTail("_d")
+  {}
+
+  // However, if we restrict ourself to a Dafny encoding, it's reversible
+  lemma {:rlimit 500}
```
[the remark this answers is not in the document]
MikaelMayer: Added a comment about doing that migration for the next PR, thanks

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+      } else if s[0] == '?' { // "_q" + dafnyEscape(s[1..]) else
+      } else if s[0] == '\'' { // "_k" + dafnyEscape(s[1..]) else
+      } else if s[0] == '#' { // "_h" + dafnyEscape(s[1..]) else=
+      } else { // [s[0]] + dafnyEscape(s[1..])=
+        dafnyEscapeCorrect(s[1..]);
+      }
+    }
+  }
+  // replaceDots is not reversible generally speaking.
+  // Indeed, both _d and . get translated to _d.
```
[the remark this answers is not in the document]
MikaelMayer: If you change `_d` to `__d`, you'll have to change `_` to `__` too as there could be a `i__d` somewhere that comes from a dafny identifier `i_d`. I prefer to keep transformations simple, as I haven't written replaceDots, I might be able to have a better transformation down the road.
`dafnyEscape` is only a modeling function to model the fact that Dafny itself does not escape dots and might put dots in identifiers names, like module names. I have no control over that.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+  {
+    if |s| == 0 {
+    } else {
+      dafnyEscapeBijective(s[1..]);
+      if s[0] == '_' {
+      } else if s[0] == '?' {
+      } else if s[0] == '\'' {
+      } else if s[0] == '#' {
+      } else {
+      }
```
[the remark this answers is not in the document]
MikaelMayer: It does reduces the proof cost :-)

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+  ghost predicate IsDafnyId(s: string) {
+    && |s| > 0 && s[0] in "aqkhd" &&
+       IsDafnyIdTail(s[1..])
+  }
+  lemma IsDafnyIdImpliesDafnyIdTail(s: string)
+    requires IsDafnyId(s)
+    ensures IsDafnyIdTail(s)
+  {}
+
+  function dafnyEscape(s: string): string {
```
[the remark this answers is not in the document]
MikaelMayer: Conversion implemented in the next PR.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+  //     - a dafny-generated identifier that starts with a "_" like "_module"
+  //     - A rust keyword
+  //       ==> use "r#" + rust keyword
+  //     - Anything else
+  //       ==> use "r#_" + replaceDot
+  // - The reverse function is such that
+  //   - If the input is a Dafny-encoded Id or a tuple numeric,
+  //     it can find the original identifier based in the output
+
+  lemma better_tuple_builder_name_not_rsharp(i: string)
```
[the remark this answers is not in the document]
MikaelMayer: I was just borrowing the convention used in the existing Dafny file... will change that in the next PR :-)

Source/DafnyCore/Compilers/Dafny/AST.dfy
MikaelMayer: Ideally I want:
- The Compiled Dafny AST to be minimal
- We want the generated code to look idiomatic and close to the original Dafny file if possible.

I was able to accomplish both goals the following way, as this step has already been accomplished in the next PR. The design is the following:
1) We keep only the minimum amount of operators (so `BinaryOp(Implies, x, y)` becomes `BinaryOp(Or, Not(x), y)`, same for all the variants of <= and < and != and == etc.
2) I added a "format" attribute to `BinaryOp` and `UnaryOp` so that compilers can decide to enhance the rendering of the operators based on original Dafny information, but that should only be an optional idiomatic rendering step or a transformation of the target language AST to another more idiomatic target language AST.
This ensures conversions are done minimally .

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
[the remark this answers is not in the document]
robin-aws: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/Compilers/GoLang/Compiler-go.cs#L79-L83

Source/DafnyCore/DafnyGeneratedFromDafny.sh
```diff
@@ -19,13 +35,20 @@ else
   output="GeneratedFromDafny"
 fi
 
-../../Scripts/dafny translate cs dfyconfig.toml --output $output.cs
+../../Scripts/dafny translate cs dfyconfig.toml --output $output.cs $noverify
+
+# We will remove all the namespaces Std.* except Std.Wrappers
```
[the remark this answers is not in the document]
robin-aws: Ah got it, the comment is just misleading. I'd just rephrase.

Source/DafnyCore/Compilers/Dafny/AST.dfy
robin-aws: > Ideally I want:
> * The Compiled Dafny AST to be minimal
> * We want the generated code to look idiomatic and close to the original Dafny file if possible.

Both good goals to have, but there will be times when they conflict. :)

Sounds like you're saying in this case we might annotate the AST further if we decided we need to know when the original Dafny source was `A ==> B` as opposed to `!A || B`. I think that's a reasonable general approach: keep the AST itself minimal so getting to simple correct translation is easier, but add more additional information when you feel the need to improve the output.

I think it's worth stating that tenet in or around this file somewhere.

Source/DafnyCore/DafnyGeneratedFromDafny.sh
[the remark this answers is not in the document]
MikaelMayer: I'll put in the new version:
```dafny
# We remove all the standard libraries except Std.Wrappers, Std.Strings, Std.Collections.Seq, Std.Arithmetic and Std.Math
```

Source/DafnyCore/Compilers/Dafny/AST.dfy
MikaelMayer: I just added to the top of Dafny-compiler-rust.dfy for the next PR:

```dafny
// Dafny to Rust compilation tenets:
// - The Compiled Dafny AST should be minimal
// - The generated code should look idiomatic and close to the original Dafny file if possible
```

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+  function dafnyEscape(s: string): string {
+    if |s| == 0 then "" else
+    if s[0] == '_' then "__" + dafnyEscape(s[1..]) else
+    if s[0] == '?' then "_q" + dafnyEscape(s[1..]) else
+    if s[0] == '\'' then "_k" + dafnyEscape(s[1..]) else
+    if s[0] == '#' then "_h" + dafnyEscape(s[1..]) else
+    [s[0]] + dafnyEscape(s[1..])
+  }
+
+  function dafnyEscapeReverse(s: string): string {
```
robin-aws, line 33: `dafnyUnescape` would be a more familiar name IMHO

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
+
+
+
+
+
+include "../Dafny/AST.dfy"
+
+module {:extern "DCOMPProofs"} {:compile false} DCOMPProofs refines DCOMP {
+
+  ghost predicate IsDafnyIdTail(s: string) {
```
robin-aws, line 10: nit: declaring this after `IsDafnyId` would make it much clearer what "tail" means

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+        assert reverseReplaceDots(replaceDots(i)) == i;
+      } else {
+        replaceDotsReversible(i[1..]);
+        assert reverseReplaceDots(replaceDots(i)) == i;
+      }
+    }
+  }
+
+  // The reason why escapeIndent is correct is that
+  // - It receives either
```
robin-aws, line 149: This sounds like a big old implicit precondition we should make explicit (since it's defined in Dafny)

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+        var s := "_" + idiomatic_rust(i[2..]);
+        assert s[0] == '_';
+      } else {
+        var s := [i[0]] + idiomatic_rust(i[1..]);
+        assert s[0] == i[0];
+      }
+    }
+  }
+
+  lemma {:fuel has_special, 2, 3} escapeIndentExamples()
```
robin-aws, line 190: Discussion: should we put "test lemmas" outside of the source files, like tests?

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+      idiomatic_rust_reversible(i[2..]);
+      assert 2 <= |i| && i[1] == '_';
+      var ss := ['_'] + idiomatic_rust(i[2..]);
+      assert ss[0] == '_' && |ss| != 0;
+    } else {
+      idiomatic_rust_reversible(i[1..]);
+    }
+  }
+
+  function reverseEscapeIdent(s: string): string {
```
robin-aws, line 226: Likewise consider "unescape" instead

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
```diff
@@ -0,0 +1,362 @@
...
+    requires !is_tuple_numeric(s)
+    requires !is_tuple_builder(s)
+    requires s !in reserved_rust
+    requires !is_idiomatic_rust(s)
+    requires is_dafny_generated_id(s)
+    ensures reverseEscapeIdent(escapeIdent(s)) == s
+  {
+  }
+
+  lemma {:rlimit 800} everythingElseReversible(i: string)
```
robin-aws, line 314: Either "Unescapable" or "Invertable" would be better than "Reversible" in general IMO

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
robin-aws, line 1: Just remarking that these are well worth the effort, since failing to escape identifiers completely in all contexts has been a frequent source of compiler bugs in the past.

Source/DafnyCore/DafnyGeneratedFromDafny.sh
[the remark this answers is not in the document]
robin-aws: This would be a good reason to start actually splitting up the standard libraries into separate units. But we need more packaging support first.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+pub fn deallocate<T : ?Sized>(pointer: *const T) {
+    unsafe {
+        // Takes ownership of the reference,
+        // so that it's deallocated at the end of the method
+        let _ = Box::from_raw(pointer as *mut T);
+    }
+}
+
+pub fn is_instance_of<T: ?Sized + Any, U: 'static>(theobject: *const T) -> bool {
+    unsafe { &*theobject }.as_any().downcast_ref::<U>().is_some()
```
robin-aws, line 356: Because Dafny is pretty restrictive about the types on either side of an `is`, it seems like it's possible to not resort to `as_any()`, as much as that is still correct. How much of an efficiency hit is it?

robin-aws (commented):
Another batch

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
MikaelMayer: Renamed to `DafnyUnescape` in the next PR

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
MikaelMayer: I exchanged the two predicates in the next PR

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
MikaelMayer: You mean, adding this as a precondition to the function `escapeIdent` itself?
Since the reception is total ("anything else" is the last case), the precondition of `escapeIdent` is actually true all the time.
Without this being an objection, the disjunction in the precondition of that function is there only to figure out a particular type of bijection (I love writing this -ctions :-) )
so writing a precondition would make sense if I split the function "escapeIdent" itself, which at this point is not necessary. Unless you have another idea to format it?

I also renamed "escapeIndent" to "escapeIdent" in the next PR since it's a typo.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
MikaelMayer: Based on your discussion topic, I added the following at the top of this file this comment, because I should make it explicit that:
```
/*This module does not contain any compiled code because it
  only proves properties about DCOMP. In a sense, it's a test file. */
```
So you are already viewing here a file that only contain test lemmas. It's just that the others are more like abstract tests, while this one is more concrete.

I could also have made it a method and expect statements to not overload the verifier. Using the verifier to perform computation is not something to advise in general. Here I kept my examples to use the least amount of fuel (the fuel indicated is the minimum + 1) and this can help to design quickly any new bijection feature, and also it makes it possible to not compile this module until we have a way to filter out test modules for compilation for production code.

I also renamed this lemma to `escapeIdentExamples` (fixed the typo)

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
MikaelMayer: I replaced every "Reversible" by "Invertible" in this file.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
MikaelMayer: As soon as we have newtypes for strings as well, I'll use them to ensure I'm not forgetting any case. I have found already two instances in which identifiers were not escaped in the source code.
And indeed, generating identifiers that can collide could be sources of issues, thanks for appreciating :-)

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Based on my current understanding, Rust is one of this few languages where references are not "transparent". Transparency means that the representation of the reference is the same whether the object in the original code was a class or an interface.
Contrary to Java, that means if a struct `C` implement `trait A` and `trait B`, a pointer of type `*const C` does not have the same representation as a pointer of type `*const dyn A` or of type `*const dyn B`. This is because a `dyn` pointer not only stores the address of the object data, but also the address of the object's methods (that's why there is a keyword "dyn", to indicate it stores two pointers). A pointer of type `*const C` therefore is not immediately a pointer of type `*const dyn A`, and a pointer of type `*const dyn A` requires an explicit conversion to a pointer of type `*const dyn B` because the methods are not at the same address. This is why the method `as_any()` makes it possible to go back on our feet.

By the way, I'm not using this is_instance_of yet but I will shortly not in the next PR, but in a PR afterwards. So this can still evolve.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
robin-aws: Oh I missed the "everything else" case and therefore thought there was a non-trivial precondition, ignore :)

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust-proofs.dfy
robin-aws: Ah that is illuminating. I had assumed these proofs would be necessary dependencies for future higher-level proofs of correctness, and therefore more like source code. If that becomes true later we can always relocate/reclassify them.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+
+// Three elements
+// ArraySequence(var isString: bool, nodeCount: SizeT, immutable_array)
+// ConcatSequence(var isString: bool, nodeCount: SizeT, left: Sequence, right: Sequence)
+// LazySequence(length: BigInteger, box: RefCell<Rc<Sequence<T>>>)
+// We use the named version using {...}, and use snake_case format
+
+// The T must be either a *const T (allocated) OR a Reference Counting (immutable)
+
+#[allow(dead_code)]
```
robin-aws, line 34: Where does this occur and how hard is it to avoid?

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+
+    pub fn dafny_sequence_to_vec<T, X>(s: &Sequence<T>, elem_converter: fn(&T) -> X) -> Vec<X>
+      where T: Clone
+    {
+        let mut array: Vec<T> = Vec::with_capacity(s.cardinality());
+        Sequence::<T>::append_recursive(&mut array, s);
+        array.iter().map(|x| elem_converter(x)).collect()
+    }
+
+    // Used for external conversions
```
robin-aws, line 59: Only this one and not the other direction?

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+    pub fn bigint_to_dafny_int(i: &BigInt) -> Rc<BigInt> {
+        Rc::new(i.clone())
+    }
+
+    pub fn dafny_sequence_to_vec<T, X>(s: &Sequence<T>, elem_converter: fn(&T) -> X) -> Vec<X>
+      where T: Clone
+    {
+        let mut array: Vec<T> = Vec::with_capacity(s.cardinality());
+        Sequence::<T>::append_recursive(&mut array, s);
+        array.iter().map(|x| elem_converter(x)).collect()
```
robin-aws, line 56: Wouldn't it be more efficient (if slightly less simple) to create a `Vec<X>` and populate it with converted elements directly? That way you'd avoid the intermediate `Vec<T>`.

I don't know Rust iterators well enough yet to know the best way to implement that, but it's probably worth figuring out as a follow-up, since this conversion method is going to be used a lot on potentially large data.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+
+    // Used for external conversions
+    pub fn vec_to_dafny_sequence<T, X>(array: &Vec<X>, elem_converter: fn(&X) -> T) -> Rc<Sequence<T>>
+      where T: Clone
+    {
+        let mut result: Vec<T> = Vec::with_capacity(array.len());
+        for elem in array.iter() {
+            result.push(elem_converter(elem));
+        }
+        Sequence::<T>::new_array_sequence(&Rc::new(result))
```
robin-aws, line 67: Just to double check, the resulting Sequence will reference `result` directly rather than creating a copy, correct?

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+            cache: RefCell::new(None)
+        })
+    }
+
+    // --unicode-chars:true
+    pub mod unicode_chars_true {
+        use crate::Sequence;
+        use std::rc::Rc;
+
+        pub fn string_to_dafny_string(s: &str) -> Rc<Sequence<char>> {
```
robin-aws, line 100: Definitely try to introduce those aliases I mentioned next PR if possible, `dafny_string` instead of `Rc<Sequence<char>>`

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+            Sequence::new_array_sequence_is_string(&Rc::new(s.chars().collect()), true)
```
robin-aws, line 101: [SHOULD fix for this PR]: I think you've got the is_string values backwards here, assuming you've also imitated how `print` handles strings from `dafnyRuntime.dfy`. See: https://github.com/dafny-lang/dafny/blob/8bf0879f69f2227fe7f8a27941084c0a9e6e0f12/Source/DafnyRuntime/DafnyRuntimeDafny/src/dafnyRuntime.dfy#L381-L388

There should be tests that assert how strings should be printed in various cases where you can check that Rust is consistent, UnicodeStrings.dfy for example

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+    // --unicode-chars:false
+    pub mod unicode_chars_false {
+        use crate::Sequence;
+        use std::rc::Rc;
+        pub fn string_to_dafny_string(s: &str) -> Rc<Sequence<u16>> {
+            Sequence::new_array_sequence_is_string(&Rc::new(s.encode_utf16().collect()), false)
+        }
+        pub fn dafny_string_to_string(s: &Rc<Sequence<u16>>) -> String {
+            let characters = s.to_array();
+            String::from_utf16_lossy(&characters)
```
robin-aws, line 118: Wouldn't it be better to panic on invalid UTF16 instead?

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+                array.push(value.clone());
+              },
+            Sequence::ConcatSequence { left, right, .. } =>
+              // Let's create an array of size length and fill it up recursively
+              {
+                Sequence::<T>::append_recursive(array, left);
+                Sequence::<T>::append_recursive(array, right);
+              }
+            Sequence::LazySequence { boxed, .. } =>
+            Sequence::<T>::append_recursive(array, boxed.borrow().as_ref()),
```
robin-aws, line 228: Indentation

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+    
+}
+
+#[allow(dead_code)]
+enum Sequence<T>
+  where T: Clone,
+{
+    ArraySequence {
+        is_string: bool,
+        node_count: usize,
```
robin-aws, line 130: Is the node_count actually used by anything at runtime? IIRC it's only in dafnyRuntime.dfy for the benefit of termination proofs in Dafny

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+            Sequence::<T>::append_recursive(&mut array, self);
+            Rc::new(array)
+        } else if let Sequence::LazySequence { boxed, is_string, ..  } = self {
+            let result = boxed.borrow().to_array();
+            // Put the value back into boxed
+            boxed.replace(Rc::clone(
+                &Sequence::<T>::new_array_sequence_is_string(&result, *is_string)));
+            result
+        } else {
+            panic!("This should never happen")
```
robin-aws, line 210: Why is this necessary here but not in the `match this` cases?

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+  where K: Clone + Eq + std::hash::Hash, V: Clone
+{
+    data: Rc<Sequence<(K, V)>>,
+    // Any time we explicitly access this map, we index the data
+    cache: RefCell<Option<HashMap<K, V>>>
+}
+
+#[allow(dead_code)]
+impl <K, V> Map<K, V>
+  where K: Clone + Eq + std::hash::Hash, V: Clone
```
robin-aws, line 279: One cross-cutting issue to keep in mind - there's an assumption that the image of all Dafny types implement `Hash`. What level of testing will support that? Do we have confidence eventually turning Rust on for all the uniform compiler tests will catch missing cases because the Rust won't compile?

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+    fn compute_hashmap(&self) {
+        let mut cache = self.cache.borrow_mut();
+        if cache.is_none() {
+            *cache = Some(self.to_hashmap_owned(
+                |x| x.clone(),
+                |x| x.clone()
+            ));
+        }
+    }
+    // Dafny will normally guarantee that the key exists.
```
robin-aws, line 311: Which means the `unwrap()`s should never panic, right?

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+struct Map<K, V> 
+  where K: Clone + Eq + std::hash::Hash, V: Clone
+{
+    data: Rc<Sequence<(K, V)>>,
+    // Any time we explicitly access this map, we index the data
+    cache: RefCell<Option<HashMap<K, V>>>
+}
+
+#[allow(dead_code)]
+impl <K, V> Map<K, V>
```
robin-aws, line 278: Ah now this is interesting. On the one hand great, ALL backends should have a smarter implementation of immutable maps. On the other hand the code we want to support with this backend has already had to work around the lack of this in the existing backends, so this optimization is not going to help them - they will use https://github.com/dafny-lang/libraries/tree/master/src/MutableMap instead.

I'd say don't spend time on this optimization here, go with the simple wrapping of a mutable map like the other backends. At some point we should do what you're doing here, but in `dafnyRuntime.dfy` instead so that we have a shot at getting consistent runtime performance across all backends.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
```diff
@@ -9,6 +16,346 @@ pub use num::NumCast;
...
+
+// Generic function to allocate and return a raw pointer immediately
+#[inline]
+pub fn allocate<T>(value: Box<T>) -> *const T {
+    Box::into_raw(value)
+}
+
+// Generic function to safely deallocate a raw pointer
+#[inline]
+pub fn deallocate<T : ?Sized>(pointer: *const T) {
```
robin-aws, line 347: Is this used anywhere already?

robin-aws (commented):
Another batch, just Dafny-compiler-rust.dfy and the Rust tests to go

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
...
+  const reserved_rust := {
+    "as","async","await","break","const","continue",
+    "crate","dyn","else","enum","extern","false","fn","for","if","impl",
+    "in","let","loop","match","mod","move","mut","pub","ref","return",
+    "Self","self","static","struct","super","trait","true","type","union",
+    "unsafe","use","where","while","Keywords","The","abstract","become",
+    "box","do","final","macro","override","priv","try","typeof","unsized",
+    "virtual","yield"}
+
+  predicate is_tuple_numeric(i: string) {
```
robin-aws, line 327: Should we be making all functions opaque by default?

More importantly, it looks like we're not applying any mechanisms to fight brittleness in the Dafny code in DafnyCore. We need to either use the dafny-reportgenerator or set a resource-limit as we did for the standard libraries.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
...
+  {
+    "_T" + i[8..]
+  }
+
+  predicate is_dafny_generated_id(i: string) {
+    && |i| > 0 && i[0] == '_' && !has_special(i[1..])
+    && (|i| >= 2 ==> i[1] != 'T') // To avoid conflict with tuple builders _T<digits>
+  }
+
+  predicate is_idiomatic_rust(i: string) {
```
robin-aws, line 386: Perhaps `is_idiomatic_rust_id` instead?

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -150,38 +555,93 @@ module {:extern "DCOMP"} DCOMP {
...
-      var ptrPartialEqImpl := "impl " + typeParams + " ::std::cmp::PartialEq for " + escapeIdent(c.name) + typeParams + " {\n";
-      ptrPartialEqImpl := ptrPartialEqImpl + "fn eq(&self, other: &Self) -> bool {\n";
-      ptrPartialEqImpl := ptrPartialEqImpl + "::std::ptr::eq(self, other)";
-      ptrPartialEqImpl := ptrPartialEqImpl + "\n}\n}\n";
-
-      var identEraseImpls := "impl " + constrainedTypeParams + " ::dafny_runtime::DafnyErasable for " + escapeIdent(c.name) + typeParams + " {\n" + "type Erased = " + escapeIdent(c.name) + erasedParams + ";\n}\n";
-      identEraseImpls := identEraseImpls + "impl " + constrainedEraseParams + " ::dafny_runtime::DafnyUnerasable<" + escapeIdent(c.name) + unerasedParams + "> for " + escapeIdent(c.name) + typeParams + " {}\n";
-
-      s := s + "\n" + defaultImpl + "\n" + printImpl + "\n" + ptrPartialEqImpl + "\n" + identEraseImpls;
+      var d := R.ImplFor(
```
robin-aws, line 580: Consider occasionally adding high-level comments to show the rough template for what we're generating, as we do in the C# implementation of rewriters and code generators. The string manipulation version could have already benefited from that, and I think the benefit is even greater when you start manipulating AST nodes instead.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -2100,12 +2703,14 @@ module {:extern "DCOMP"} DCOMP {
 
     static method Compile(p: seq<Module>) returns (s: string) {
       s := "#![allow(warnings, unconditional_panic)]\n";
+      s := s + "#![allow(nonstandard_style)]\n";
```
robin-aws, line 2706: Kind of sums up the overall philosophy of our code generators nicely doesn't it. :)

Source/DafnyRuntime/DafnyRuntimeRust/src/tests/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+        *y = !x;
+        *z = !x || *y;
+    }
+    fn Test3() { 
+      let mut y = true; let mut z = true;
+      Test2(true, &mut y, &mut z);
+      let mut i; let mut j;
+      i = y;
+      j = z;
+    }
```
robin-aws, line 383: These look like random code written to experiment with semantics, and AFAICT they aren't actually part of any tests, so we should remove them.

Source/DafnyRuntime/DafnyRuntimeRust/src/tests/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+    #[test]
+    fn test_apply1Native() {
+        let y: Rc<BigInt> = Rc::new(BigInt::one());
+        let y_copy = Rc::clone(&y); // Create a cloned BigInt
+        let f = Rc::new(
+            move |x: &Rc<BigInt>|y_copy.eq(x));
+        assert_eq!(f.as_ref()(&Rc::new(BigInt::zero())), false);
+    }
+}
+// Struct containing two reference-counted fields
```
robin-aws, line 411: ?

robin-aws (commented):
Got through the rest. Massive progress, nicely done.

My only real concern is integration-level testing. I'm very happy with the Rust testing of the Rust runtime as that's already above the level of testing of some of the other runtimes. But because we're using the `RunAllTests` feature as a catch-all escape for the Dafny-implemented compilers, it's basically impossible for me to know how many more actual Dafny programs we support after this change, or if we're even regressing on what was already working.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Yeah, no reflection is needed because any class that can be upcasted must implement the Any trait, otherwise the approach is unsound.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: I removed all these annotations in the new PR. They were there because, contrary to Dafny, functions are private to the module by default, so if they are not prefixed by "pub", then it's dead code. By adding "pub" on every function like of Dafny (which I had done since), I now realized that this annotation of allowing dead code is no longer used, so I removed the 13 of them.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: The other direction is already implemented for the next PR.

MikaelMayer: I might have forgotten it for this PR, but the other direction is implemented in the next version.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Yes, indeed. Rc::new() takes ownership of the result.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: I introduced them two weeks ago in the upcoming work, definitely a good idea.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: It looks like the name in DafnyRuntimeDafny is an antonym compared to what this boolean represents.
Can I suggest we change the name in DafnyRuntimeDafny to "is_seq_char" instead?

MikaelMayer: Fixed in this PR.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Since dafny strings are built only from valid u16 that are UTF-16, this problem will never arise, but I'll do the change in this PR.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Because Rust's if sentences are like if-then-else in Dafny, they are types, so they always require an else (there is no return keyword, only the last value is returned). panic returning the empty type, it type checks.

In the newest PR, this if then else is transformed back to a match, as I'm getting more used to Rust's syntax I was able to make it work.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Since I already completed this optimization, let's keep it for now as it is strictly better in terms of complexity than wrapping a hashmap. Wrapping a hashmap is O(n) when you add a key/value and O(1) when accessing it, whereas this is O(1) when adding a key/value pair or even combining maps, and amortized O(1) when accessing it.

But either optimizations are not even the panacea, as if there is a loop where someone tests if something is in the map before adding it, then it defeats totally the optimization as the Hashmap will be realized at every step anyway.

My vision is that for those who really want performance, they will use real mutable hashmap for which maps will only be ghost specification.
In other terms, I believe that classes and arrays should be the most optimized, whereas built-in immutable types are mostly there for prototyping and can tolerate non-optimal complexities.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Correct!

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Only in the test/mod.rs for now.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
robin-aws: That's all reasonable. Let me refine my advice:

This does look like a nice optimization without making the implementation a lot more complex. But it is still slightly more complex. In particular, you have to worry about `data` and `cache` getting out of sync when used by different concurrent executions. That problem is solved for the lazy sequences in `dafnyRuntime.dfy` by using an `AtomicBox`, and for Rust it looks like the equivalent solution is a `RefCell`, so perhaps you need that here.

All I'm saying is that you still have more work TODO to complete this implementation anyway, so I'm still not convinced that this optimization is worth the non-zero risk. I think it would be more worth it if multi-targeting Dafny code could rely on those [amortized] O(1) costs regardless of the backend.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
 include "../Dafny/AST.dfy"
 
+// Rust AST
+module RAST {
+  import opened Std.Wrappers
+  import opened Std.Strings
+
+  const IND := "  "
+  // Indentation level
+  /* trait ModDecl {} */
```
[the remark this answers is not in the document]
MikaelMayer: Removed. I want to use general-traits datatype down the road, but for now I have not done the switch yet. I got many issues with this general traits datatypes in other projects that I'm now a bit burned to use it again (for example, it was impossible to downcast traits to datatypes, an issue has been cut but it's not resolved yet)

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
...
+
+  datatype Mod /*extends ModDecl*/ =
+      // Rust modules
+    | Mod(name: string, body: seq<ModDecl>)
+    | ExternMod(name: string)
+  {
+    function ToString(ind: string): string
```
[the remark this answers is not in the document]
MikaelMayer: I'm unfortunately not playing with --general-traits anymore until a few issues have been fixed. Moreover, traits are useful only if I need to abstract datatypes away, which here we don't.
I'll add a comment about ToString

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
...
+          "mod " + name + " {" + "\n" + ind + IND +
+          SeqToString(
+            body,
+            (modDecl: ModDecl) requires modDecl < this =>
+              modDecl.ToString(ind + IND), "\n" + ind + IND)
+          + "\n" + ind + "}"
+      }
+    }
+  }
+  function SeqToString<T>(s: seq<T>, f: T --> string, separator: string := ""): string
```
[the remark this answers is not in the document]
MikaelMayer: Indeed I thought for that, but I don't want to compute the intermediate seq<seq<char>>, so I'll stick with this SeqToString function for now. But I have used other standard library methods somewhere else :-).

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
...
+  {
+    function ToString(ind: string): string {
+      if FnDecl? then
+        (if pub == PUB then "pub " else "") + fun.ToString(ind)
+      else assert RawImplMember?; content
+    }
+  }
+  newtype VISIBILITY = x: int | 0 <= x < 2
+  const PUB := 1 as VISIBILITY
+  const PRIV := 0 as VISIBILITY
```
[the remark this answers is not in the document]
MikaelMayer: I don't need the index, so I'll write just `datatype Visibility = PUB | PRIV`

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
...
+  datatype AssignIdentifier =
+    AssignIdentifier(identifier: string, rhs: Expr)
+  {
+    function ToString(ind: string): string {
+      identifier + ": " + rhs.ToString(ind + IND)
+    }
+  }
+
+  // When a raw stream is given, try to put some indentation on it
+  function AddIndent(raw: string, ind: string): string {
```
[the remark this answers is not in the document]
MikaelMayer: Yep, although I might actually use this one later on if I decide that the indentation I asked for is not enough, i.e. because the string is too long and I want to add a newline, I'd add some indentation.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
...
+        case Some(body) =>
+          " {\n" + ind + IND +
+          body.ToString(ind + IND) +
+          "\n" + ind + "}"
+      }
+    }
+  }
+}
+
 module {:extern "DCOMP"} DCOMP {
```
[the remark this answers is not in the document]
MikaelMayer: Renamed in the next PR.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
```diff
@@ -1,126 +1,508 @@
...
+    else
+      [i[0]] + replaceDots(i[1..])
+  }
+
+  predicate is_tuple_builder(i: string) {
+    && |i| >= 9
+    && i[..8] == "___hMake"
+    && i[8] in "0123456789"
+    && (|i| == 9 || (|i| == 10 && i[9] in "0123456789"))
+  }
```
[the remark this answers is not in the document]
MikaelMayer: Good catch ! I added the comment `// A tuple builder identifier looks like ___hMake0 to ___hMake99`

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
MikaelMayer: Because of the current design, everything verifies quite fast as I'm abiding by the rule that I should not use preconditions or postconditions in functions. Thus, there is little to verify in Dafny and the few lines that Dafny verifies cannot be reduced.
Since there are several competing ways to fight brittleness, I'm making a note to remember that when the time comes. I don't think we need to care about this right now.
What bothers me more is that resolution takes 5 minutes the second I make a new change (and is much faster when I reload the IDE), and verification 20 seconds afterwards.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
MikaelMayer: Renamed in the next PR

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
MikaelMayer: I might not do this immediately as I'm constantly evolving the representation, but please remind me of doing that when the implementation is more stabilized if I have not done so yet.

Source/DafnyCore/Compilers/Rust/Dafny-compiler-rust.dfy
MikaelMayer: Note that smithy-rust does change CamelCase to snake_case, we could have a compiler option for doing that as well :-)

Source/DafnyRuntime/DafnyRuntimeRust/src/tests/mod.rs
MikaelMayer: I removed the entire file from the index and kept it locally until we use it for proper unit testing.

Source/DafnyRuntime/DafnyRuntimeRust/src/lib.rs
MikaelMayer: Concurrency is a good thing to consider, but consider this:
For this code to work under concurrent execution, we would need anyway to use the slower Arc instead of Rc, because Rc is not concurrent-friendly. And I'm not sure we want to so that, we still need to evaluate that need.

But regarding your remark, how can data and cache get ouf of sync? I'm already using RefCell
Data is only ever written once during construction, and cache could theoretically be updated concurrently but all result only depend only on the data so at worst it will be populated twice with the same value.

robin-aws (approved):
Will follow up on lots of comments in the next PR, but nothing blocking. Ship it!

PR #4975, opened 2024-01-11
Implement `:resource_limit` attribute

### LINKED ISSUE #4962: Remove multiplication of resource count by 1000
### Dafny version

4.4.0

```dafny
method {:rlimit 10} M(x: int) returns (r: int)
  ensures r >= 0
{
    return x * x;
}
```

```code
$ dafny verify --solver-option VERBOSITY=2 rlimit.dfy
...
[SMT-INP-0] (set-option :timeout 0)
[SMT-INP-0] (set-option :rlimit 10000)
[SMT-INP-0] (set-option :auto_config false)
[SMT-INP-0] (set-option :type_check true)
[SMT-INP-0] (set-option :smt.qi.eager_threshold 100)
[SMT-INP-0] (set-option :smt.delay_units true)
[SMT-INP-0] (set-option :smt.case_split 3)
[SMT-INP-0] (set-option :smt.mbqi false)
[SMT-INP-0] (set-option :model.compact false)
[SMT-INP-0] (set-option :model.v2 true)
[SMT-INP-0] (set-option :pp.bv_literals false)
[SMT-INP-0] (set-option :smt.arith.solver 2)
[SMT-INP-0] (assert (not  (=> (= (ControlFlow 0 0) 4) (let ((...
[SMT-INP-0] (get-info :name)
[SMT-OUT-0] (:name "Z3")
[SMT-INP-0] (check-sat)
[SMT-INP-0] (get-info :name)
[SMT-OUT-0] unsat
[SMT-OUT-0] (:name "Z3")
[SMT-INP-0] (get-info :rlimit)
[SMT-INP-0] (get-info :name)
[SMT-OUT-0] (:rlimit 3431)
[SMT-OUT-0] (:name "Z3")
[SMT-INP-0] (pop 1)
[SMT-INP-0] ; Valid
[SMT-INP-0] (get-info :name)
[SMT-OUT-0] (:name "Z3")

```

### What happened?

Note the `(set-option :rlimit 10000)`. The values sent to Z3, printed with `--boogie-option -trace`, and stored in verification log files are in terms of Z3's conception of rlimit, not the value given with the `{:rlimit ...}` attribute. This may be a source of confusion.

We should get rid of the multiplication. However, for backward compatibility, we may want to do it by introducing a new attribute. 

Mac

atomb:
### Description

This is an alternative to `:rlimit` that doesn't multiply by 1000 before sending the value to Z3. With the addition of this new attribute, `:rlimit` is deprecated.

The `--resource-limit` flag now also passes through the value unchanged, without a multiplication. This flag is currently not used widely, so the impact of this breaking change is minimal. Verification pipelines that are affected will break immediately, and be easy to fix.

The `:resource_limit` attribute and `--resource-limit` flag now accept exponential notation to allow shorter, easier-to-read limits.

Fixes #4962.

### How has this been tested?

The integration tests have been updated to use `:resource_limit` wherever they used `:rlimit` previously. Some `--resource-limit` and `:resource_limit` uses have been updated to use exponential notation.


Source/DafnyCore/DafnyOptions.cs
```diff
@@ -358,6 +358,27 @@ public enum IncludesModes {
       ).ToArray();
     }
 
+    public static bool TryParseResourceCount(string value, out uint number) {
```
keyboardDrummer, line 361: Could you rename the parameter `number` to `result` ?

Source/DafnyCore/DafnyOptions.cs
```diff
@@ -358,6 +358,27 @@ public enum IncludesModes {
...
+      uint multiplier = 1;
+      if (value.EndsWith("G")) {
```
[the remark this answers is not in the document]
atomb: Oh, that might be nice. Thanks for pointing it out!

docs/dev/news/4975.fix
```diff
@@ -0,0 +1 @@
+The `{:rlimit N}` attribute, which multiplied `N` by 1000 before sending it to Z3, is deprecated in favor of the `{:resource_limit N}` attribute, which can accept string arguments with exponential notation for brevity.
```
robin-aws, line 1: I'd also mention the change in behavior of `--resource-limit`

Source/DafnyCore/Options/BoogieOptionBag.cs
```diff
@@ -63,7 +63,18 @@ public static class BoogieOptionBag {
...
+      var value = result.Tokens[^1].Value;
+
+      if (DafnyOptions.TryParseResourceCount(value, out var number)) {
+        return number;
+      }
+
+      result.ErrorMessage = $"Cannot parse resource limit: {value}";
+      return 0;
+    },
+    false,
```
robin-aws, line 76: What parameter is this? Consider using a named argument or a comment.

Source/DafnyCore/Verifier/BoogieGenerator.ExpressionTranslator.cs
```diff
@@ -1959,6 +1959,19 @@ public ExpressionTranslator(ExpressionTranslator etran, Boogie.Expr heap)
               BigNum.FromUInt(Boogie.Util.BoundedMultiply((uint)litExpr.asBigNum.ToIntSafe, 1000)),
               litExpr.Immutable);
           }
+
+          // Do this after the above multiplication because :resource_limit should not be multiplied.
```
robin-aws, line 1963: We should probably just reject using both attributes at once. At a minimum that check would be here, but perhaps there's a cleaner spot in resolution to catch it earlier?

docs/DafnyRef/Attributes.md
```diff
@@ -393,12 +393,15 @@ Print effects are enforced only with `--track-print-effects`.
...
 
-### 11.2.14. `{:rlimit}` {#sec-rlimit}
+### 11.2.14. `{:resource_limit}` and `{:rlimit}` {#sec-rlimit}
 
-`{:rlimit N}` limits the verifier resource usage to verify the method or function at `N * 1000`.
-This is the per-method equivalent of the command-line flag `/rlimit:N`.
+`{:resource_limit N}` limits the verifier resource usage to verify the method or function to `N`.
+
+This is the per-method equivalent of the command-line flag `/rlimit:N` or `--resource-limit N`.
 If using [`{:vcs_split_on_every_assert}`](#sec-vcs_split_on_every_assert) as well, the limit will be set for each assertion.
```
robin-aws, line 401: Hey while you're adding attribute aliases, how about `{:isolate_assertions}`? ;)

(Not literally in this PR of course)

robin-aws (changes requested):
Definitely in favor of this cleanup ASAP.

> This flag is currently not used widely, so the impact of this breaking change is minimal.

What also makes me okay with this is that it's a hard breakage: if someone upgrades without being aware of it their verification will likely fail because their limit is suddenly 1/1000th what it was. And if it doesn't then yay, they are better protected against brittleness! :)

atomb:
> > This flag is currently not used widely, so the impact of this breaking change is minimal.
> 
> What also makes me okay with this is that it's a hard breakage: if someone upgrades without being aware of it their verification will likely fail because their limit is suddenly 1/1000th what it was. And if it doesn't then yay, they are better protected against brittleness! :)

I'll update the description (and merge message) to mention that. :)

robin-aws (approved):

PR #4979, opened 2024-01-12
fix: Don’t flag extern functions with {:axiom} as assumptions

### LINKED ISSUE #4964: Standard Library fails to build to target:lib (doo)
### Dafny version

4.4

```dafny
Line 25, UTF8.dfy in Standard Library

function method {:extern "Encode"} Encode(s: string): (res: Result<ValidUTF8Bytes, string>)
   // US-ASCII only needs a single UTF-8 byte per character
   ensures IsASCIIString(s) ==> res.Success? && |res.value| == |s|
   // The following MUST be true for any correct implementation of Encode
   // If it weren't, then data would be lost.
   ensures res.Success? ==> Decode(res.value).Success? && Decode(res.value).value == s
```

```code
Command:

build
--target:lib
--function-syntax:3
--quantifier-syntax:3

Output:

Error: Function with [{:extern}] attribute.
   |
25 |   function method {:extern "Encode"} Encode(s: string): (res: Result<ValidUTF8Bytes, string>)
   |                                      ^^^^^^
```

### What happened?

Got the below error. Same code builds with `target:go`.
```
Error: Function with [{:extern}] attribute.
   |
25 |   function method {:extern "Encode"} Encode(s: string): (res: Result<ValidUTF8Bytes, string>)
   |                                      ^^^^^^
```

Mac

robin-aws:
### Description
Fixes #4964, more directly and specifically than the general improvements @keyboardDrummer is making as per https://github.com/dafny-lang/dafny/issues/4964#issuecomment-1884898031 (which are also good improvements for the future). Pre/post-conditions on extern functions are already allowed in doo files with the `{:axiom}` attribute, so this just allows that to suppress the additional implicit requirement that the extern implementation behaves like a deterministic function.

### How has this been tested?
Existing tests of the auditor and doo files (removes an arguably duplicative assumption on `function {:extern} {:axiom} ...` declarations)

### DIFF
--- a/Source/DafnyCore/AST/Members/Function.cs
+++ b/Source/DafnyCore/AST/Members/Function.cs
@@ -65,15 +65,14 @@ public class Function : MemberDecl, TypeParameter.ParentType, ICallable, ICanFor
       yield return new Assumption(this, tok, AssumptionDescription.NoBody(IsGhost));
     }
 
-    if (HasExternAttribute) {
+    if (HasExternAttribute && !HasAxiomAttribute) {
       yield return new Assumption(this, tok, AssumptionDescription.ExternFunction);
-      if (HasPostcondition && !HasAxiomAttribute) {
+      if (HasPostcondition) {
         yield return new Assumption(this, tok, AssumptionDescription.ExternWithPostcondition);
       }
-    }
-
-    if (HasExternAttribute && HasPrecondition && !HasAxiomAttribute) {
-      yield return new Assumption(this, tok, AssumptionDescription.ExternWithPrecondition);
+      if (HasPrecondition) {
+        yield return new Assumption(this, tok, AssumptionDescription.ExternWithPrecondition);
+      }
     }
 
     foreach (var c in this.Descendants()) {
--- a/Source/IntegrationTests/TestFiles/LitTests/LitTest/auditor/TestAuditor.dfy-ietf.md.expect
+++ b/Source/IntegrationTests/TestFiles/LitTests/LitTest/auditor/TestAuditor.dfy-ietf.md.expect
@@ -86,8 +86,6 @@
 
 * Declaration has explicit `{:axiom}` attribute. SHOULD provide a proof or test.
 
-* Function with `{:extern}` attribute. SHOULD use `method` with `modifies {}` instead.
-
 ## Type `T`
 
 * Trait method calls may not terminate (uses `{:termination false}`). MUST remove `{:termination false}` or justify.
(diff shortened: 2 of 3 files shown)


Source/DafnyCore/AST/Members/Function.cs
```diff
@@ -65,15 +65,14 @@ public class Function : MemberDecl, TypeParameter.ParentType, ICallable, ICanFor
       yield return new Assumption(this, tok, AssumptionDescription.NoBody(IsGhost));
     }
 
-    if (HasExternAttribute) {
+    if (HasExternAttribute && !HasAxiomAttribute) {
```
keyboardDrummer, line 68: I believe the auditor wants to report on everything that has `{:axiom}` on it. The way I understood the auditor is that you don't run it as part of CI, but use it as part of a whole-program code review that you do at particular times in your development workflow, like before a release or when onboarding a new customer.

atomb: Yes, that's my intention for it as well. It's for when you want to say "this thing is fully verified, but what risks remain?". For that, we really want to keep all extern declarations in the list of things reported by the auditor.

I think this approach is a good one for silencing warnings during the normal development process. In the context of the auditor, I think having `{:axiom}` should convert an extern declaration from an implicit assumption to an explicit assumption. Explicit assumptions are those that have been acknowledged as assumptions through some extra annotation and are therefore considered intentional. The final goal for a verified project is, in my mind, for all remaining assumptions to be explicit, and to have documented reasons for why you think that each one is warranted.

robin-aws: On thing that might not be obvious here: the state before this change is that `function {:extern} {:axiom} ...` would always produce two Assumptions: one for the presence of `{:axiom}` in general, and one for the fact that you're assuming that an extern actually behaves like a non-deterministic function. The first is considered an explicit assumption and the second an implicit one. The intent is that doo files only allow explicit assumptions, so the first was allowed but the second caused an error.

It was already true that adding `{:axiom}` to an extern method or function would replace the implicit assumptions for any post-conditions by an explicit one just saying "you have an axiom here" instead. The effect of this change is just to let the explicit `{:axiom}` assumption suppress the additional implicit one about the function-ness of your extern too. Hence just adding `&& !HasAxiomAttribute` makes sense for both the auditor and building doo files.

keyboardDrummer (approved):

Source/DafnyCore/AST/Members/Function.cs
keyboardDrummer: Makes sense. I had trouble seeing that the explicit assumption is still flagged.

atomb: Ditto. I agree in that broader context.

Issue #4983, opened 2024-01-12
Boogie crash: Error: invalid argument types (Box and DatatypeType) to binary operator ==


selig:
### Dafny version

4.4.0

### Code to produce this issue

```dafny
trait Foo {}
datatype Bar extends Foo = Bar

function getFoo() : Foo { Bar }
```

### Command to run and resulting output

```code
dafny verify --type-system-refresh --general-traits:datatype

tmpg7YYPr.tmp.dfy(4,9): Error: invalid argument types (Box and DatatypeType) to binary operator ==
tmpg7YYPr.tmp.dfy(4,9): Error: invalid argument types (Box and DatatypeType) to binary operator ==
2 type checking errors detected in /tmp/tmpg7YYPr.tmp__module.bpl

*** Encountered internal translation error - re-running Boogie to get better debug information

/tmp/tmpg7YYPr.tmp__module.bpl(2838,34): Error: invalid argument types (Box and DatatypeType) to binary operator ==
/tmp/tmpg7YYPr.tmp__module.bpl(2843,34): Error: invalid argument types (Box and DatatypeType) to binary operator ==
2 type checking errors detected in /tmp/tmpg7YYPr.tmp__module.bpl
```

### What happened?

Expected it to verify; it crashed.

### What type of operating system are you experiencing the problem on?

Other

PR #4997, opened 2024-01-19
fix: Make datatype cycle detection independent of auto-init

### LINKED ISSUE #4939: False cyclic dependency with datatype extending trait
### Dafny version

4.4.0

```dafny
// Trampoline.dfy

trait Program {
  method Compute() returns (r: Result)
}

datatype Result =
| Bounce(next: Program)
| Done()

datatype Trivial extends Program =
  Trivial()
{
  method Compute() returns (r: Result) {
    return Done();
  }
}

datatype Seq extends Program =
  Seq(left: Program, right: Program)
{
  method Compute() returns (r: Result) {
    var result := left.Compute();
    match result
    case Done() => r := right.Compute();
    case Bounce(next) =>
      var next' := Seq(next, right);
      assert next' is Program;
      return Bounce(next');
  }
}
```

```code
dafny verify --type-system-refresh --general-traits:datatype Trampoline.dfy
```

### What happened?

The typechecker raises:

```
Trampoline.dfy(19,9): Error: because of cyclic dependencies among constructor argument types, no instances of datatype 'Seq' can be constructed
   |
19 | datatype Seq extends Program =
   |          ^^^
```

likely b/c of the self-reference in `Seq.Compute()` despite `Seq(Trivial(), Trivial())` because one such instance.

Mac

RustanLeino:
This PR makes the "datatype has no instances" checks more liberal.

Fixes #4939 

## Description

An inductive datatype is allowed to be defined as an empty type. For example, in

``` dafny
predicate P(x: int) { false }
type Subset = x: int | P(x) witness *
datatype Record = Record(Subset)
```

`Record` is an empty type, because `Subset` is, since `P(x)` is always `false`. But if `P(x)` was instead defined to be `true` for some `x`'s, then `Record` would be nonempty. Determining whether or not `Record` is empty goes well beyond the syntactic checks of the type system.

However, if a datatype is empty because of some "obvious" cycle among datatype definitions, then that is both detectable by syntactic checks and likely unintended by the programmer. By this PR, Dafny will search for such type declarations and give error messages if an inductive datatype is necessarily involved in such an "obvious" cycle.

## Previous behavior

Dafny had attempted this check before, too, but then the rule was too strict. The check had been implemented as part of the machinery that computed "grounding constructors". That search consulted the auto-init status of the types that a datatype depends on. Since traits are of the "possibly empty" auto-init category, if every constructor of a datatype depended on a trait (that is, the constructors had a parameter whose type is a trait), then Dafny would report an error.

This PR removes the error reporting done by the machinery that computes grounding constructors. It adds a new, separate check in the resolver to determine the "obvious" cycles.

## Grounding constructors

The name "grounding constructor" was originally to mean "a constructor that can be used to construct an instance of the datatype".  This name is rather outdated, because Dafny's syntactic checks do not ensure datatypes to be nonempty anyway (case in point: type `Record` above). A name like "initializing constructor" would be more to the point--the compiler uses this constructor in a generated method for creating an arbitrary value of the type. (At run time, this method will never be called if the datatype really is empty, but since the compiler does not know this, it still has to generate valid target code for this method.)

## Alternative designs

One alternative design would be to change the new "because of a cycle, this datatype has no instances" error message to be just a warning. This alternative has some merit, since Dafny already supports empty types.

Another alternative would be to not perform the check at all, and to report neither an error nor a warning for datatypes with no instances. This would give more flexibility. However, since there are simple ways to write an empty type, a program where a datatype cycle causes a type to have no instances is most likely a mistake that the programmer would want to know about and fix.


Source/DafnyCore/Resolver/ModuleResolver.cs
```diff
@@ -1272,6 +1272,26 @@ public partial class ModuleResolver {
...
+      // not Record is empty goes well beyond the syntactic checks of the type system.
+      //
+      // However, if a datatype is empty because of some "obvious" cycle among datatype definitions, then
+      // that is both detectable by syntactic checks and likely unintended by the programmer. Therefore,
+      // we search for such type declarations and give error messages if something is found.
+      if (reporter.Count(ErrorLevel.Error) == prevErrorCount) {
+        foreach (var dtd in declarations.ConvertAll(decl => decl as IndDatatypeDecl).Where(dtd => dtd != null && dtd.Ctors.Count != 0)) {
+          if (AreThereAnyObviousSignsOfEmptiness(UserDefinedType.FromTopLevelDecl(dtd.tok, dtd), new HashSet<IndDatatypeDecl>())) {
+            reporter.Error(MessageSource.Resolver, dtd,
+              $"because of cyclic dependencies among constructor argument types, no instances of datatype '{dtd.Name}' can be constructed");
```
ssomayyajula, line 1290: Should this be a warning instead?

Source/DafnyCore/AST/TypeDeclarations/IndDatatypeDecl.cs
```diff
@@ -1,17 +1,33 @@
...
 
   public override DatatypeCtor GetGroundingCtor() {
-    return GroundingCtor;
+    return GroundingCtor ?? Ctors.FirstOrDefault(ctor => ctor.IsGhost, Ctors[0]);
   }
 
-  public bool[] TypeParametersUsedInConstructionByGroundingCtor;  // set during resolution; has same length as the number of type arguments
+  private bool[] typeParametersUsedInConstructionByGroundingCtor;
+
+  public bool[] TypeParametersUsedInConstructionByGroundingCtor {
```
ssomayyajula, line 17: I'm assuming this is here in case the type in question is empty on purpose.

RustanLeino: Correct.

Source/DafnyCore/Resolver/ModuleResolver.cs
RustanLeino: Yes, that's one of the alternative designs. Thanks for your vote for it. I'll change it to be a warning.

ssomayyajula (approved):

RustanLeino:
This PR also fixes #4983. Tests for that issue have been added in PR #5058.

PR #5000, opened 2024-01-22
Improve trigger related warnings

keyboardDrummer:
### Description
- Refactor trigger related code. Sorry for not having this be a separate PR. I can still try to split it if that helps a lot.
- Add a section to the reference manual to help debug trigger related warnings
- Improve diagnostics generated for triggers.
  - Trigger generation related warnings now point to the relevant section in the reference manual
  - Trigger generation related warnings now mention how to silence them
  - Splitting of quantifiers is reported more explicitly
  - UI tweaks to the diagnostics, best inspected through the diff of the expect files

### How has this been tested?
- Updated expect files of existing tests


docs/DafnyRef/Topics.md
```diff
@@ -1146,3 +1146,24 @@ method TestD(dd: D) {
...
+
+Here follows the approach Dafny uses to generate triggers based on a quantifier. Dafny finds terms in the quantifier body where a quantified variable is used in an operation, such as in a function application `P(x)`, array access `a[x]`, member accesses `x.someField`, or set membership tests `x in S`. To find a trigger, Dafny must find a set of such terms so that each quantified variable is used. You can investigate which triggers Dafny finds by hovering over quantifiers in the IDE and looking for 'Selected triggers', or by using the options `--show-tooltips` when using the LCI.
+
+There are particular expressions which, for technical reasons, Dafny can not use as part of a trigger. Among others, these expression include: match, let, arithmetic operations and logical connectives. For example, in the quantifier `forall x :: x in S ⇐⇒ f(x) > f(x+1)`, Dafny will use `x in S` and `f(x)` as trigger terms, but will not use `x+1` or any terms that contain it. You can investigate which triggers Dafny can not use by hovering over quantifiers in the IDE and looking for 'Rejected triggers', or by using the options `--show-tooltips` when using the LCI.
```
MikaelMayer, line 1157:
```suggestion
There are particular expressions which, for technical reasons, Dafny can not use as part of a trigger. Among others, these expression include: [match](#sec-match-expression), [let](#sec-let-expression), [arithmetic operations and logical connectives](#sec-expressions). For example, in the quantifier `forall x :: x in S ⇐⇒ f(x) > f(x+1)`, Dafny will use `x in S` and `f(x)` as trigger terms, but will not use `x+1` or any terms that contain it. You can investigate which triggers Dafny can not use by hovering over quantifiers in the IDE and looking for 'Rejected triggers', or by using the options `--show-tooltips` when using the LCI.
```

docs/DafnyRef/Topics.md
```diff
@@ -1146,3 +1146,24 @@ method TestD(dd: D) {
...
+
+Besides not finding triggers, another problematic situation is when Dafny was able to generate triggers, but believes the triggers it found may still cause useless instantiations because they create matching loops. Dafny emits a warning when this happens, indicating the Dafny program should be changed so Dafny can find triggers for this quantifier that do not cause matching loops.
+
+To understand matching loops, one needs to understand how triggers are used. During verification, Dafny maintains a set of expression which it believes to be true, which we call ground terms. Given a univeral quantifier that's a ground term, Dafny will use its triggers to pattern match on other ground terms. If the pattern matches, that ground term is used to instantiate the quantifier.
```
MikaelMayer, line 1161: Could you explain a bit more with an example what a ground term is? I'm not sure I follow. Is it a global set of expressions, a local one, a tree one?

docs/DafnyRef/Topics.md
```diff
@@ -1146,3 +1146,24 @@ method TestD(dd: D) {
...
+
+Even existential quantifiers need triggers. This is because when Dafny determines an existential quantifier is false, for example in `requires !exists x :: f(x) == 2`, Dafny will use a logical rewrite rule to change this existential into a universal quantifier, so it becomes `requires forall x :: f(x) != 2`. Before verification, Dafny can not determine whether quantifiers will be determined to be true or false, so it must assume any quantifier may turn into a universal quantifier, and thus they all need triggers.
+
+Besides quantifiers, comprehensions such as set and map comprehension also need triggers, since these are modelled using universal quantifiers.
```
MikaelMayer, line 1167:
```suggestion
Besides quantifiers, comprehensions such as set and map comprehension also need triggers, since these are modeled using universal quantifiers.
```
US English is used throughout the documentation (see Color)

docs/DafnyRef/Topics.md
```diff
@@ -1146,3 +1146,24 @@ method TestD(dd: D) {
...
+
+Dafny may report 'Quantifier was split into X parts'. This occurs when Dafny determines it can only generate good triggers for a quantifier by splitting it into multiple smaller quantifiers, whose disjunction is logically equivalent to the original one. To maintain logical equivalence, Dafny has to generate more triggers than if the split had been done in source. One may consider doing the split in source to generate a leaner set of triggers.
```
MikaelMayer, line 1169: I would love to see an example of that, otherwise that paragraph leaves me with unsatisfied thirst.

For example, you can say `exists x :: P(x) || Q(x)` is split into  `exists x :: P(x)` and `exists x :: Q(x)`. And you can use that example to comment on the trigger generation as well. Also, the disjunction is proper to existential quantifiers, not forall quantifiers. For them, it's a conjunction.

Source/IntegrationTests/TestFiles/LitTests/LitTest/triggers/splitting-triggers-recovers-expressivity.dfy.expect
```diff
@@ -6,18 +6,20 @@ splitting-triggers-recovers-expressivity.dfy(26,10): Info: Selected triggers: {Q
...
-   Selected triggers: {Q(j)}
-   Rejected triggers: {P(j)} (may loop with "P(j + 1)")
+splitting-triggers-recovers-expressivity.dfy(44,10): Info: Quantifier was split into 2 parts. Better verification performance and error reporting may be obtained by splitting the quantifier in source. For more information, see the section quantifier instantiation rules in the reference manual.
+splitting-triggers-recovers-expressivity.dfy(44,10): Info: Part #0 is 'P(i) || !Q(i)'
+ Selected triggers:
+   {Q(i)}, {P(i)}
+splitting-triggers-recovers-expressivity.dfy(44,10): Info: Part #1 is 'P(i + 1)'
+ Selected triggers: {Q(i)}
+ Rejected triggers: {P(i)} (may loop with "P(i + 1)")
+splitting-triggers-recovers-expressivity.dfy(49,11): Info: Quantifier was split into 2 parts. Better verification performance and error reporting may be obtained by splitting the quantifier in source. For more information, see the section quantifier instantiation rules in the reference manual.
```
MikaelMayer, line 16: I'm not sure I understood with the documentation what it means by "splitting the quantifier in source". You mean, like the exists quantifier example I give below?

MikaelMayer (commented):
I skimmed through the refactored parts and redundant message changing but I like it! A few items to address but otherwise it's in good shape.

Source/IntegrationTests/TestFiles/LitTests/LitTest/triggers/splitting-triggers-recovers-expressivity.dfy.expect
keyboardDrummer: I mean that you would manually perform the split so Dafny does not have to

docs/DafnyRef/Topics.md
```diff
@@ -1158,12 +1158,16 @@ There are particular expressions which, for technical reasons, Dafny can not use
...
 
-To understand matching loops, one needs to understand how triggers are used. During verification, Dafny maintains a set of expressions which it believes to be true, which we call ground terms. Given a univeral quantifier that's a ground term, Dafny will use its triggers to pattern match on other ground terms. If the pattern matches, that ground term is used to instantiate the quantifier. 
+To understand matching loops, one needs to understand how triggers are used. During a single verification run, such as verifying a method or function, Dafny maintains a set of expressions which it believes to be true, which we call the ground terms. For example, in the body of a method, Dafny knows the requires clauses of that method hold, so the expressions in those will be ground terms. When Dafny steps through the statements of the body, the set of ground terms grows. For example, when an assignment `var x := 3` is evaluated, a ground term `x == 3` will be added. Given a universal quantifier that's a ground term, Dafny will try to pattern match its triggers on other ground terms. If the pattern matches, that ground term is used to instantiate the quantifier.
```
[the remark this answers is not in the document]
keyboardDrummer: I think if we introduce a new word we should ensure it is used often enough. Knowledge base is used just once. Regarding 'trigger' and 'ground terms', these concepts come from SMT. Even though Dafny encapsulates SMT, there are some holes in that encapsulation so I think reusing some of those terms makes sense at this point.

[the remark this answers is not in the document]
MikaelMayer: Oh ok. I'm fine with ground terms then. I'll just have a minor suggestion then, because pattern matching is done not just at the level of ground terms, but sub-expressions
```suggestion
To understand matching loops, one needs to understand how triggers are used. During a single verification run, such as verifying a method or function, Dafny maintains a set of expressions which it believes to be true, which we call the ground terms. For example, in the body of a method, Dafny knows the requires clauses of that method hold, so the expressions in those will be ground terms. When Dafny steps through the statements of the body, the set of ground terms grows. For example, when an assignment `var x := 3` is evaluated, a ground term `x == 3` will be added. Given a universal quantifier that's a ground term, Dafny will try to pattern match its triggers on sub-expressions of other ground terms. If the pattern matches, that sub-expression is used to instantiate the quantifier. 
```

MikaelMayer (approved):

PR #5008, opened 2024-01-24
Filter option

### LINKED ISSUE #4816: Add `--filter` option to `dafny verify`
### Summary

There should be an easy command `--filter` to which I can pass which files/lemmas I want to Dafny to verify.

### Background and Motivation

Before project files, people could run `dafny verify onefile.dfy` to only verify a subset of the project. With a project file, this is no longer possible if people have removed `include`s from their files. Instead they have to use `--boogie-filter`, which is less ergonomic and less documented. Related discussion in #4624.

### Proposed Feature

Add a command line option `dafny verify --filter` to restrict the set of files/lemmas that Dafny verifies. It might use `--boogie-filter` under the hood, but provides a more ergonomic interface to the user and has good documentation.

### Alternatives

One alternative for users is not to use project files. But it would be a shame if the lack of such a feature deterred people from using project files.

For restricting verification to just one function, users can use the `{:only}` attribute (thanks @seebees for pointing this out!). But I'm not aware of an equivalent for restricting verification to just one file.

keyboardDrummer:
Fixes #4816 

### Description
Add an option `--filter-position` to the `dafny verify` command, that allows verifying a single function or method, based on a file path and a line number. For example, `dafny verify dfyconfig.toml --filter-position=source1.dfy:5` will only verify something if it ranges over line 5 in the file `source1.dfy`.

### How has this been tested?
Add a littish test `filter.dfy`


Source/DafnyDriver/CliCompilation.cs
```diff
@@ -228,4 +234,28 @@ public class CliCompilation {
...
     }
   }
+
+  private List<ICanVerify> FilterCanVerifies(List<ICanVerify> canVerifies) {
+    var filterPosition = options.Get(VerifyCommand.FilterPosition);
+    if (filterPosition == null) {
+      return canVerifies;
+    }
+
+    var regex = new Regex(@"(.+)(?::(\d+))?", RegexOptions.RightToLeft);
```
ssomayyajula, line 244: I'll trust you on this, haha.

Source/DafnyDriver/CliCompilation.cs
```diff
@@ -228,4 +234,28 @@ public class CliCompilation {
...
+    }
+    var filePart = result.Groups[1].Value;
+    string? linePart = result.Groups.Count > 2 ? result.Groups[2].Value : null;
+    var fileFiltered = canVerifies.Where(c => c.Tok.Uri.ToString().EndsWith(filePart)).ToList();
+    if (string.IsNullOrEmpty(linePart)) {
+      return fileFiltered;
+    }
+
+    var line = int.Parse(linePart);
+    return fileFiltered.Where(c =>
```
ssomayyajula, line 258: To confirm: you are trusting that the delimited portion of the file is valid standalone Dafny code?

ssomayyajula (approved):

Source/DafnyDriver/CliCompilation.cs
keyboardDrummer: No, the entire Dafny project is parsed and resolved regardless. After resolution the verifiable parts are extracted, and this filter is applied to those to select what exactly to verify.

ssomayyajula (approved):

davidcok:
I would find this much easier to use (and less prone to breaking when files are edited) if the designation of what to verify took a possibly-qualified method name instead of file/lie numbers

keyboardDrummer:
> I would find this much easier to use (and less prone to breaking when files are edited) if the designation of what to verify took a possibly-qualified method name instead of file/lie numbers

I would like to add that option too, that's why I've named this one `--filter-position` instead of just `--filter`. This option is one that users have requested as well. You can use it to verify a whole file, and in the future you'll be able to use to verify a single assertion, which the option that you describe won't be able to.

